PR #285, opened 2019-11-30
HDDS-2655. Use pre-compiled Pattern in NetUtils#normalize

adoroszlai:

Replace `String#replaceAll` in `NetUtils#normalize` with a pre-compiled `Pattern`, which is better for performance.

https://issues.apache.org/jira/browse/HDDS-2655

## How was this patch tested?

Added unit test (`TestNetUtils`).


adoroszlai:
Thanks @ayushtkn and @elek for the review, and @elek for committing it.

PR #286, opened 2019-12-02
HDDS-2475. Unregister ContainerMetadataScrubberMetrics on thread exit

adoroszlai:

Unregister `ContainerMetadataScrubberMetrics` when `ContainerMetadataScanner` thread exits to avoid leak.

https://issues.apache.org/jira/browse/HDDS-2475

## How was this patch tested?

Added unit test case.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/ozoneimpl/ContainerMetadataScanner.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/ozoneimpl/ContainerMetadataScanner.java
@@ -66,6 +66,8 @@ public class ContainerMetadataScanner extends Thread {
           metrics.resetNumContainersScanned();
         }
       }
+    } catch (Exception e) {
+      LOG.error("{} exiting because of exception ", this, e);
     } finally {
       if (metrics != null) {
         metrics.unregister();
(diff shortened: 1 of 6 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/ozoneimpl/ContainerMetadataScanner.java
```diff
@@ -58,11 +58,17 @@ public void run() {
...
-        metrics.resetNumUnhealthyContainers();
-        metrics.resetNumContainersScanned();
+    try {
+      while (!stopping) {
+        runIteration();
+        if (!stopping) {
+          metrics.resetNumUnhealthyContainers();
+          metrics.resetNumContainersScanned();
+        }
+      }
```
xiaoyuyao: Can we add a log entry if any chance the runIteration() throw? Similar logic can be found in ContainerDataScanner#run().

anuengineer:
+1. @xiaoyuyao  Thanks for the review. @adoroszlai  Thanks for the contribution. I will file a follow up JIRA for the other issues mentioned by Xiaoyu.

anuengineer:
Filed https://issues.apache.org/jira/browse/HDDS-2664 to track the other metric issue and add logging in this path.

adoroszlai:
Thanks @xiaoyuyao for the review and @anuengineer for committing it.

I'll resolve HDDS-2664, since all followup issues were addressed in 889664c06a19520705027d9903bfd5cbec62948b and 1233fec96c07372ae02430f2c6a3b73b2723604e.

PR #287, opened 2019-12-02
HDDS-2654. Conditionally enable profiling at the kernel level

adoroszlai:

Extend `entrypoint.sh` to set the kernel parameters required for profiling if the `ASYNC_PROFILER_HOME` environment variable (used by `ProfileServlet`) is defined.

https://issues.apache.org/jira/browse/HDDS-2654

## How was this patch tested?

```
$ cd hadoop-ozone/dist/target/ozone-0.5.0-SNAPSHOT/compose/ozone
$ docker-compose up -d
$ docker-compose logs scm | grep 'Enabled profiling'
scm_1       | Enabled profiling in kernel
# wait for OM
$ open http://localhost:9874/prof
```


adoroszlai:
Thanks @elek for reviewing and committing this.

PR #288, opened 2019-12-03
HDDS-2506. Remove keyAllocationInfo and replication info from the auditLog

adoroszlai:

Remove key location info (and replication info) from audit log params.  The same information can be obtained (for debug purposes) from Ozone Insight.  This improves performance of audit logging, formatting a deeply nested proto message structure is expensive.  It also improves audit log structure, as each messages is expected to be a single line (except for stack traces).

https://issues.apache.org/jira/browse/HDDS-2506

## How was this patch tested?

Ran `ozonesecure` acceptance test.  Verified audit messages:

```
2019-12-02 12:10:43,616 | INFO  | OMAudit | user=HTTP/scm@EXAMPLE.COM | ip=172.31.0.4 | op=ALLOCATE_KEY {volume=vol-0-29689, bucket=bucket-0-44465, key=key-0-79381, dataSize=10240} | ret=SUCCESS |
2019-12-02 12:10:46,042 | INFO  | OMAudit | user=HTTP/scm@EXAMPLE.COM | ip=172.31.0.4 | op=COMMIT_KEY {volume=vol-0-29689, bucket=bucket-0-44465, key=key-0-79381, dataSize=10240} | ret=SUCCESS |
```

https://github.com/adoroszlai/hadoop-ozone/runs/330684634

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyArgs.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyArgs.java
@@ -147,12 +147,6 @@ public final class OmKeyArgs implements Auditable {
     auditMap.put(OzoneConsts.BUCKET, this.bucketName);
     auditMap.put(OzoneConsts.KEY, this.keyName);
     auditMap.put(OzoneConsts.DATA_SIZE, String.valueOf(this.dataSize));
-    auditMap.put(OzoneConsts.REPLICATION_TYPE,
-        (this.type != null) ? this.type.name() : null);
-    auditMap.put(OzoneConsts.REPLICATION_FACTOR,
-        (this.factor != null) ? this.factor.name() : null);
-    auditMap.put(OzoneConsts.KEY_LOCATION_INFO,
-        (this.locationInfoList != null) ? locationInfoList.toString() : null);
     return auditMap;
   }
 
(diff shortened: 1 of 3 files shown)


elek:
Thanks @adoroszlai I totally agree (+1), for me "audit" is audit of security and not audit for debugging. I think it's better to use "ozone insight" for debug and include only the key information (WHO did WHAT with a resource)

But will wait at least one day to get some feedback from @bharatviswa504

adoroszlai:
Thanks @dineshchitlangia and @bharatviswa504 for the review.

> having replication information would be a good idea right, to know what type of replication factor is used for creating keys.

I think replication factor/type are fine to keep in the audit log, they are of limited size.  Let me know if I should add them back.

bharatviswa504:
> Thanks @dineshchitlangia and @bharatviswa504 for the review.
> 
> > having replication information would be a good idea right, to know what type of replication factor is used for creating keys.
> 
> I think replication factor/type are fine to keep in the audit log, they are of limited size. Let me know if I should add them back.

+1 to add replication information

anuengineer:
> +1 to add replication information

Why? it is a Audit log. Why do we need Replica Info? Audit is used to find out who made access and when. There is a very delicate trade-off between making the size of Audit log larger and finding critical information when we need it. Typically, an HDFS cluster would produce GBs of Audit logs per day, and processing all that information is lots of work. The replication in my mind is a debug tool, plus, you can always get that info from the OM.

We have a set of Queries that we support in the Audit log tool. If we are able to write at least one query, them I am +1 for adding this to Audit Log. In fact, that should be the test for adding any field to the Audit log; we cannot consume the field what is the point of having it in the log. 

Even a far-fetched hypothetical scenario is ok, but we must have some use from the auditing and security point of view to add something to the audit log.

anuengineer:
+1. I am good to commit this change. I will leave open till end of the day to see if anyone would like to comment on this. Thanks

anuengineer:
Since there are no follow up comments, I will commit this now. If we need to make any changes to this, please feel free to file a follow-up JIRA.

adoroszlai:
Thanks @anuengineer for committing it.

PR #291, opened 2019-12-04
HDDS-1574 Average out pipeline allocation on datanodes and add metrcs/test

timmylicheng:
1. Fix pipeline allocation in non-topology env and prevent pipelines from sharing the same set of datanodes.
2. Add metrics and logs to record when pipeline policy chooses two same set of datanodes for different pipeline for future reference.
3. Add tests for basic reports and mock what pipeline allocation would do given different number of nodes.

https://issues.apache.org/jira/browse/HDDS-1574

## How was this patch tested?
UT 

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManager.java
@@ -133,6 +133,13 @@ class PipelineStateManager {
       pipeline = pipelineStateMap
           .updatePipelineState(pipelineId, PipelineState.OPEN);
     }
+    // Amend nodeIdsHash if needed.
+    if (pipeline.getType() == ReplicationType.RATIS &&
+        pipeline.getFactor() == ReplicationFactor.THREE &&
+        pipeline.getNodeIdsHash() == 0) {
+      pipeline.setNodeIdsHash(RatisPipelineUtils
+          .encodeNodeIdsOfFactorThreePipeline(pipeline.getNodes()));
+    }
     return pipeline;
   }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
@@ -157,7 +157,6 @@ public class RatisPipelineProvider implements PipelineProvider {
     }
 
     List<DatanodeDetails> dns;
-
     switch(factor) {
     case ONE:
       dns = pickNodesNeverUsed(ReplicationFactor.ONE);
@@ -170,12 +169,18 @@ public class RatisPipelineProvider implements PipelineProvider {
       throw new IllegalStateException("Unknown factor: " + factor.name());
     }
 
+    int nodeIdHash = 0;
+    if (factor == ReplicationFactor.THREE) {
+      nodeIdHash = RatisPipelineUtils.encodeNodeIdsOfFactorThreePipeline(dns);
+    }
+
     Pipeline pipeline = Pipeline.newBuilder()
         .setId(PipelineID.randomId())
         .setState(PipelineState.ALLOCATED)
         .setType(ReplicationType.RATIS)
         .setFactor(factor)
         .setNodes(dns)
+        .setNodeIdsHash(nodeIdHash)
         .build();
 
     // Send command to datanodes to create pipeline
@@ -197,12 +202,17 @@ public class RatisPipelineProvider implements PipelineProvider {
   @Override
   public Pipeline create(ReplicationFactor factor,
                          List<DatanodeDetails> nodes) {
+    int nodeIdHash = 0;
+    if (factor == ReplicationFactor.THREE) {
+      nodeIdHash = RatisPipelineUtils.encodeNodeIdsOfFactorThreePipeline(nodes);
+    }
     return Pipeline.newBuilder()
         .setId(PipelineID.randomId())
         .setState(PipelineState.ALLOCATED)
         .setType(ReplicationType.RATIS)
         .setFactor(factor)
         .setNodes(nodes)
+        .setNodeIdsHash(nodeIdHash)
         .build();
   }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
@@ -157,6 +157,8 @@ public class RatisPipelineProvider implements PipelineProvider {
     }
 
     List<DatanodeDetails> dns;
+    int nodeIdHash = 0;
+
     switch(factor) {
     case ONE:
       dns = pickNodesNeverUsed(ReplicationFactor.ONE);
@@ -164,14 +166,10 @@ public class RatisPipelineProvider implements PipelineProvider {
     case THREE:
       dns = placementPolicy.chooseDatanodes(null,
           null, factor.getNumber(), 0);
+      nodeIdHash = RatisPipelineUtils.encodeNodeIdsOfFactorThreePipeline(dns);
       break;
     default:
       throw new IllegalStateException("Unknown factor: " + factor.name());
-    }
-
-    int nodeIdHash = 0;
-    if (factor == ReplicationFactor.THREE) {
-      nodeIdHash = RatisPipelineUtils.encodeNodeIdsOfFactorThreePipeline(dns);
     }
 
     Pipeline pipeline = Pipeline.newBuilder()
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
@@ -18,9 +18,15 @@
 package org.apache.hadoop.hdds.scm.pipeline;
 
 import java.io.IOException;
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Collectors;
 
+import org.apache.commons.collections.comparators.ComparableComparator;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.client.HddsClientUtils;
 import org.apache.hadoop.hdds.ratis.RatisHelper;
@@ -34,7 +40,6 @@ import org.apache.ratis.rpc.SupportedRpcType;
 import org.apache.ratis.util.TimeDuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 
 /**
  * Utility class for Ratis pipelines. Contains methods to create and destroy
@@ -98,6 +103,47 @@ public final class RatisPipelineUtils {
             requestTimeout)) {
       client.groupRemove(RaftGroupId.valueOf(pipelineID.getId()),
           true, p.getId());
+    }
+  }
+
+  static int encodeNodeIdsOfFactorThreePipeline(List<DatanodeDetails> nodes) {
+    if (nodes.size() != HddsProtos.ReplicationFactor.THREE.getNumber()) {
+      return 0;
+    }
+    List<UUID> nodeIds = nodes.stream()
+        .map(DatanodeDetails::getUuid).distinct()
+        .collect(Collectors.toList());
+    nodeIds.sort(new ComparableComparator());
+    // Only for Factor THREE pipeline.
+    return new HashCodeBuilder()
+        .append(nodeIds.get(0).toString())
+        .append(nodeIds.get(1).toString())
+        .append(nodeIds.get(2).toString())
+        .toHashCode();
+  }
+
+  /**
+   * Return first existed pipeline which share the same set of datanodes
+   * with the input pipeline.
+   * @param stateManager PipelineStateManager
+   * @param pipeline input pipeline
+   * @return first matched pipeline
+   */
+  static Pipeline checkPipelineContainSameDatanodes(
+      PipelineStateManager stateManager, Pipeline pipeline) {
+    List<Pipeline> matchedPipelines = stateManager.getPipelines(
+        HddsProtos.ReplicationType.RATIS,
+        HddsProtos.ReplicationFactor.THREE)
+        .stream().filter(p ->
+            // For all OPEN or ALLOCATED pipelines
+            (p.getPipelineState() == Pipeline.PipelineState.OPEN ||
+                p.getPipelineState() == Pipeline.PipelineState.ALLOCATED) &&
+                p.getNodeIdsHash() == pipeline.getNodeIdsHash())
+        .collect(Collectors.toList());
+    if (matchedPipelines.size() == 0) {
+      return null;
+    } else {
+      return matchedPipelines.stream().findFirst().get();
     }
   }
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
@@ -134,9 +134,9 @@ public final class RatisPipelineUtils {
     List<Pipeline> matchedPipelines = stateManager.getPipelines(
         HddsProtos.ReplicationType.RATIS,
         HddsProtos.ReplicationFactor.THREE)
-        .stream().filter(p ->
-            // For all OPEN or ALLOCATED pipelines
-            (p.getPipelineState() == Pipeline.PipelineState.OPEN ||
+        .stream().filter(p -> !p.equals(pipeline) &&
+            (// For all OPEN or ALLOCATED pipelines
+                p.getPipelineState() == Pipeline.PipelineState.OPEN ||
                 p.getPipelineState() == Pipeline.PipelineState.ALLOCATED) &&
                 p.getNodeIdsHash() == pipeline.getNodeIdsHash())
         .collect(Collectors.toList());
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
@@ -19,11 +19,8 @@ package org.apache.hadoop.hdds.scm.pipeline;
 
 import java.io.IOException;
 import java.util.List;
-import java.util.UUID;
 import java.util.stream.Collectors;
 
-import org.apache.commons.collections.comparators.ComparableComparator;
-import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
@@ -110,16 +107,9 @@ public final class RatisPipelineUtils {
     if (nodes.size() != HddsProtos.ReplicationFactor.THREE.getNumber()) {
       return 0;
     }
-    List<UUID> nodeIds = nodes.stream()
-        .map(DatanodeDetails::getUuid).distinct()
-        .collect(Collectors.toList());
-    nodeIds.sort(new ComparableComparator());
-    // Only for Factor THREE pipeline.
-    return new HashCodeBuilder()
-        .append(nodeIds.get(0).toString())
-        .append(nodeIds.get(1).toString())
-        .append(nodeIds.get(2).toString())
-        .toHashCode();
+    return nodes.get(0).getUuid().hashCode() ^
+        nodes.get(1).getUuid().hashCode() ^
+        nodes.get(2).getUuid().hashCode();
   }
 
   /**
@@ -134,7 +124,7 @@ public final class RatisPipelineUtils {
     List<Pipeline> matchedPipelines = stateManager.getPipelines(
         HddsProtos.ReplicationType.RATIS,
         HddsProtos.ReplicationFactor.THREE)
-        .stream().filter(p -> !p.equals(pipeline) &&
+        .stream().filter(p -> !p.getId().equals(pipeline.getId()) &&
             (// For all OPEN or ALLOCATED pipelines
                 p.getPipelineState() == Pipeline.PipelineState.OPEN ||
                 p.getPipelineState() == Pipeline.PipelineState.ALLOCATED) &&
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
@@ -177,13 +177,15 @@ public class SCMPipelineManager implements PipelineManager {
         metrics.incNumPipelineCreated();
         metrics.createPerPipelineMetrics(pipeline);
       }
-      if (RatisPipelineUtils.checkPipelineContainSameDatanodes(
-          stateManager, pipeline) != null) {
+      Pipeline overlapPipeline = RatisPipelineUtils
+          .checkPipelineContainSameDatanodes(stateManager, pipeline);
+      if (overlapPipeline != null) {
         metrics.incNumPipelineContainSameDatanodes();
         //TODO remove until pipeline allocation is proved equally distributed.
-        LOG.info("Pipeline contains same datanodes as previous pipeline." +
-            " PipelineId:" + pipeline.getId().toString() +
-            " nodeIds: " + pipeline.getNodes().get(0).getUuid().toString() +
+        LOG.info("Pipeline: " + pipeline.getId().toString() +
+            " contains same datanodes as previous pipeline: " +
+            overlapPipeline.getId().toString() + " nodeIds: " +
+            pipeline.getNodes().get(0).getUuid().toString() +
             ", " + pipeline.getNodes().get(1).getUuid().toString() +
             ", " + pipeline.getNodes().get(2).getUuid().toString());
       }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
@@ -128,6 +128,19 @@ public class SCMPipelineManager implements PipelineManager {
     pipelineFactory.setProvider(replicationType, provider);
   }
 
+  private int computeNodeIdHash(Pipeline pipeline) {
+    if (pipeline.getType() != ReplicationType.RATIS) {
+      return 0;
+    }
+
+    if (pipeline.getFactor() != ReplicationFactor.THREE) {
+      return 0;
+    }
+
+    return RatisPipelineUtils.
+        encodeNodeIdsOfFactorThreePipeline(pipeline.getNodes());
+  }
+
   private void initializePipelineState() throws IOException {
     if (pipelineStore.isEmpty()) {
       LOG.info("No pipeline exists in current db");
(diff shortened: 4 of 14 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -162,6 +183,15 @@ public synchronized Pipeline createPipeline(ReplicationType type,
...
         metrics.incNumPipelineCreated();
         metrics.createPerPipelineMetrics(pipeline);
+        for (int i = 0; i < checkPipelineContainSameDatanodes(pipeline); i++) {
+          metrics.incNumPipelineContainSameDatanodes();
+          //TODO remove until pipeline allocation is proved equally distributed.
+          LOG.info("Pipeline contains same datanodes as previous pipeline." +
+              " PipelineId:" + pipeline.getId().toString() +
+              " nodeIds: " + pipeline.getNodes().get(0).getUuid().toString() +
+              ", " + pipeline.getNodes().get(1).getUuid().toString() +
+              ", " + pipeline.getNodes().get(2).getUuid().toString());
```
ChenSammi: Suggest also log the overlapped pipeline ID, besides the current pipeline ID.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/MockNodeManager.java
```diff
@@ -93,16 +93,17 @@
...
     this.staleNodes = new LinkedList<>();
     this.deadNodes = new LinkedList<>();
     this.nodeMetricMap = new HashMap<>();
     this.node2PipelineMap = new Node2PipelineMap();
     this.node2ContainerMap = new Node2ContainerMap();
     this.dnsToUuidMap = new ConcurrentHashMap<>();
-    aggregateStat = new SCMNodeStat();
-    clusterMap = new NetworkTopologyImpl(new Configuration());
+    this.aggregateStat = new SCMNodeStat();
+    this.clusterMap =clusterMap;
```
ChenSammi: add a space between after '='

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
```diff
@@ -0,0 +1,118 @@
...
+        createdPipelineCount++;
+      }
+    } catch(SCMException e) {
+      end = true;
+    } catch (IOException e) {
+      end = true;
+      // Should not throw regular IOException.
+      Assert.fail();
+    }
+    end = false;
```
ChenSammi: end seems have no effect.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
```diff
@@ -0,0 +1,118 @@
...
+
+  private int checkPipelineContainSameDatanodes(
+      PipelineStateManager stateManager, Pipeline pipeline) {
+    List<Pipeline> pipelines = stateManager.getPipelines(
+        HddsProtos.ReplicationType.RATIS,
+        HddsProtos.ReplicationFactor.THREE,
+        Pipeline.PipelineState.OPEN);
+    return (int)pipelines.stream().filter(p -> p.getNodeIdsHash()
+        == pipeline.getNodeIdsHash()).count();
+  }
```
ChenSammi: Move to RatisPipelineUtil.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
```diff
@@ -0,0 +1,118 @@
...
+    int healthyNodeCount = nodeManager
+        .getNodeCount(HddsProtos.NodeState.HEALTHY);
+    int intersectionCount = 0;
+    int createdPipelineCount = 0;
+    try {
+      while (!end && createdPipelineCount <= healthyNodeCount * nodeHeaviness) {
+        Pipeline pipeline = provider.create(HddsProtos.ReplicationFactor.THREE);
+        stateManager.addPipeline(pipeline);
+        nodeManager.addPipeline(pipeline);
+        intersectionCount +=
```
ChenSammi: intersectionCount seems larger than expected. Please re-consider the algorithm.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestRatisPipelineProvider.java
```diff
@@ -130,6 +140,49 @@ public void testCreatePipelineWithNodes() {
...
+
+    Pipeline pipeline1 = provider.create(
+        HddsProtos.ReplicationFactor.THREE, healthyNodes);
+    Pipeline pipeline2 = provider.create(
+        HddsProtos.ReplicationFactor.THREE, healthyNodes);
+
+    Assert.assertTrue(pipeline1.getNodes().parallelStream()
+        .allMatch(pipeline2.getNodes()::contains));
+    Assert.assertEquals(pipeline1.getNodeIdsHash(), pipeline2.getNodeIdsHash());
+  }
```
ChenSammi: Add one more test, to assert different datanodes formed pipelines have different hash.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineFactory.java
```diff
@@ -55,6 +55,10 @@ void setProvider(ReplicationType replicationType,
     providers.put(replicationType, provider);
   }
 
+  PipelineProvider getProvide(ReplicationType type) {
+    return providers.get(type);
+  }
+
```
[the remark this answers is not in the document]
timmylicheng: Removed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeManager.java
```diff
@@ -125,6 +125,13 @@
    */
   int getPipelinesCount(DatanodeDetails datanodeDetails);
 
+  /**
+   * Get the count of pipelines associated to datanode id.
+   * @param dnid datanodeId
+   * @return number of pipelines associated with it
+   */
+  int getPipelinesCount(UUID dnid);
```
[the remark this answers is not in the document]
timmylicheng: Removed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManager.java
```diff
@@ -133,6 +133,10 @@ Pipeline openPipeline(PipelineID pipelineId) throws IOException {
       pipeline = pipelineStateMap
           .updatePipelineState(pipelineId, PipelineState.OPEN);
     }
+    if (pipeline.getType() == ReplicationType.RATIS) {
+      pipeline.setNodeIdsHash(RatisPipelineUtils
+          .encodeNodeIdsOfFactorThreePipeline(pipeline.getNodes()));
+    }
```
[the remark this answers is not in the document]
timmylicheng: This is also called during recovery. I add some check to avoid duplicated action.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
```diff
@@ -197,12 +196,17 @@ public Pipeline create(ReplicationFactor factor) throws IOException {
   @Override
   public Pipeline create(ReplicationFactor factor,
                          List<DatanodeDetails> nodes) {
+    int nodeIdHash = 0;
+    if (factor == ReplicationFactor.THREE) {
+      nodeIdHash = RatisPipelineUtils.encodeNodeIdsOfFactorThreePipeline(nodes);
+    }
```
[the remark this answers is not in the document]
timmylicheng: The method is refactored in recent master. Update to accommodate.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -143,11 +156,19 @@ private void initializePipelineState() throws IOException {
...
       nodeManager.addPipeline(pipeline);
     }
   }
 
+  private int checkPipelineContainSameDatanodes(Pipeline pipeline) {
+    List<Pipeline> pipelines = stateManager.getPipelines(ReplicationType.RATIS,
+        ReplicationFactor.THREE, Pipeline.PipelineState.OPEN);
+    return (int)pipelines.stream().filter(p -> p.getNodeIdsHash()
+        == pipeline.getNodeIdsHash()).count();
+  }
```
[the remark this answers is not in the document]
timmylicheng: Refactored

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/MockNodeManager.java
timmylicheng: Done

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
timmylicheng: Refactored the loop.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
timmylicheng: Refactored the loop. See the updates.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestRatisPipelineProvider.java
timmylicheng: Done in another test.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
timmylicheng: Refactored the logic. Please check the updates.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
```diff
@@ -292,6 +292,15 @@ public int getPipelinesCount(DatanodeDetails datanodeDetails) {
...
 
+  /**
+   * Get the count of pipelines associated to datanode id.
+   * @param dnid datanodeId
+   * @return number of pipelines associated with it
+   */
+  public int getPipelinesCount(UUID dnid) {
+    return node2PipelineMap.getPipelinesCount(dnid);
+  }
+
```
ChenSammi: Not used. Can be removed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -143,11 +156,19 @@ private void initializePipelineState() throws IOException {
...
       nodeManager.addPipeline(pipeline);
     }
   }
 
+  private boolean checkPipelineContainSameDatanodes(Pipeline pipeline) {
+    List<Pipeline> pipelines = stateManager.getPipelines(ReplicationType.RATIS,
+        ReplicationFactor.THREE, Pipeline.PipelineState.OPEN);
+    return (int)pipelines.stream().filter(p -> p.getNodeIdsHash()
+        == pipeline.getNodeIdsHash()).count() > 0;
+  }
```
ChenSammi: Not used.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -162,6 +183,16 @@ public synchronized Pipeline createPipeline(ReplicationType type,
       if (pipeline.isOpen()) {
         metrics.incNumPipelineCreated();
         metrics.createPerPipelineMetrics(pipeline);
+        if (RatisPipelineUtils.checkPipelineContainSameDatanodes(
```
ChenSammi: Returned previous pipeline is not recorded.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
```diff
@@ -0,0 +1,121 @@
...
+
+    System.out.println("Among total " +
+        stateManager.getPipelines(HddsProtos.ReplicationType.RATIS,
+            HddsProtos.ReplicationFactor.THREE).size() + " created pipelines" +
+        " with " + healthyNodeCount + " healthy datanodes and " +
+        nodeHeaviness + " as node heaviness, " +
+        intersectionCount + " pipelines has same set of datanodes.");
+  }
+
+  private boolean checkPipelineContainSameDatanodes(
```
ChenSammi: not used.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
ChenSammi: There are another case of pipeline open, through openPipeline API. You should add the same logic there.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
timmylicheng: Removed

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
timmylicheng: Removed

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
timmylicheng: Removed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
timmylicheng: Updated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
```diff
@@ -170,12 +169,18 @@ public Pipeline create(ReplicationFactor factor) throws IOException {
       throw new IllegalStateException("Unknown factor: " + factor.name());
     }
 
+    int nodeIdHash = 0;
+    if (factor == ReplicationFactor.THREE) {
```
xiaoyuyao: NIT: this condition check can be combined into the switch(factor) after line 165.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
```diff
@@ -100,4 +105,45 @@ static void destroyPipeline(DatanodeDetails dn, PipelineID pipelineID,
...
   }
+
+  static int encodeNodeIdsOfFactorThreePipeline(List<DatanodeDetails> nodes) {
+    if (nodes.size() != HddsProtos.ReplicationFactor.THREE.getNumber()) {
+      return 0;
+    }
+    List<UUID> nodeIds = nodes.stream()
+        .map(DatanodeDetails::getUuid).distinct()
+        .collect(Collectors.toList());
+    nodeIds.sort(new ComparableComparator());
```
xiaoyuyao: Can we use orderless hash here to avoid sorting? E.g., xor the hash of all the nodeId hash?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
```diff
@@ -100,4 +105,45 @@ static void destroyPipeline(DatanodeDetails dn, PipelineID pipelineID,
...
+  }
+
+  /**
+   * Return first existed pipeline which share the same set of datanodes
+   * with the input pipeline.
+   * @param stateManager PipelineStateManager
+   * @param pipeline input pipeline
+   * @return first matched pipeline
+   */
+  static Pipeline checkPipelineContainSameDatanodes(
```
xiaoyuyao: Should we move this to a non-static PipelineStateManager method?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
```diff
@@ -100,4 +105,45 @@ static void destroyPipeline(DatanodeDetails dn, PipelineID pipelineID,
...
+      PipelineStateManager stateManager, Pipeline pipeline) {
+    List<Pipeline> matchedPipelines = stateManager.getPipelines(
+        HddsProtos.ReplicationType.RATIS,
+        HddsProtos.ReplicationFactor.THREE)
+        .stream().filter(p -> !p.equals(pipeline) &&
```
xiaoyuyao: Need to check the override of equals() in Pipeline object, the list of nodes may not in the same order. This requires us to use getNodesInOrder instead of getNodes().

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -128,6 +128,19 @@ public void setPipelineProvider(ReplicationType replicationType,
...
+    if (pipeline.getType() != ReplicationType.RATIS) {
+      return 0;
+    }
+
+    if (pipeline.getFactor() != ReplicationFactor.THREE) {
+      return 0;
+    }
+
+    return RatisPipelineUtils.
+        encodeNodeIdsOfFactorThreePipeline(pipeline.getNodes());
```
xiaoyuyao: unordered node list can produce different hash for pipelines with the same datanodes.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -163,6 +177,18 @@ public synchronized Pipeline createPipeline(ReplicationType type,
         metrics.incNumPipelineCreated();
         metrics.createPerPipelineMetrics(pipeline);
       }
+      Pipeline overlapPipeline = RatisPipelineUtils
+          .checkPipelineContainSameDatanodes(stateManager, pipeline);
+      if (overlapPipeline != null) {
+        metrics.incNumPipelineContainSameDatanodes();
+        //TODO remove until pipeline allocation is proved equally distributed.
+        LOG.info("Pipeline: " + pipeline.getId().toString() +
```
xiaoyuyao: can we put this with if (LOG.isDebugEnabled()) and change the log level to DEBUG as this is could affect write performance?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
timmylicheng: I will use PipelineID then

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
timmylicheng: The log only prints when we find overlapped pipelines who share same datanodes. From TestPipelineDatanodesIntersection that I observe, overlap shows up rarely (like 0 in 62 pipelines, 1 in 73 pipelines, 4 in 242 pipelines). 
I set up this log to collect overlap cases in regular running in order to collect information for potential overlapping. My major purpose is to see if random allocation of datanodes considering rack and topology is good enough to spread nodes up in pipelines. If the stats proved otherwise, I will try to implement other allocation algorithm. I talk with Sammi and we feel like we may not be able to collect enough info in DEBUG mode.
So the log will only print few and while it prints, it could give us a good idea on how pipelines are spread across all nodes.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
timmylicheng: I put this into Utils for all classes to access. I tried PipelineStateManager first and Sammi advised to put it into Utils for some test classes who don't have pipelineStateManager initialized.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
timmylicheng: That's why I sort the datanode list. I will see if XOR could resolve it.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
timmylicheng: I will use xor then

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
timmylicheng: Good catch. Updated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineUtils.java
xiaoyuyao: Both should work just non-static seems cleaner. But if there are too much additional work to fix the unit tests. I'm OK with current approach.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
xiaoyuyao: Talked offline. OK to have this at INFO level for now.

timmylicheng:
The acceptance test is fixed in https://issues.apache.org/jira/browse/HDDS-2650

timmylicheng:
testCloseContainerEventWithRatis doesn't see the same error locally as on CI.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
```diff
@@ -0,0 +1,124 @@
...
+      } catch (IOException e) {
+        end = true;
+        // Should not throw regular IOException.
+        Assert.fail();
+      }
+    }
+
+    end = false;
+
+    System.out.println("Among total " +
```
xiaoyuyao: can we change to a log4j output instead of console output?

xiaoyuyao:
Thanks @timmylicheng  for the update. The latest change LGTM. 

Wrt. the failures in testCloseContainerEventWithRatis. I spend some time and debug it.
After we allow background pipeline thread to create more than one pipelines per datanode, the OZONE_SCM_PIPELINE_NUMBER_LIMIT should always be set properly in the test. Otherwise, this will run indefinitely and timeout the even queue processing for some of those with tight timeout (e.g., 1s in this case). Also, when the limit is reached, we throw exceptions which ends up with a lot of false alarm error logs and the pipeline failure counter incremented. We can fix that in follow up JIRAs.      

Here is a proposed fix for testCloseContainerEventWithRatis, which you can include in this PR.  

```
TestCloseContainerEventHandler.java

@@ -24,6 +24,7 @@
 import org.apache.hadoop.hdds.HddsConfigKeys;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
+import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.TestUtils;
 import org.apache.hadoop.hdds.scm.pipeline.MockRatisPipelineProvider;
 import org.apache.hadoop.hdds.scm.pipeline.PipelineProvider;
@@ -67,22 +68,25 @@ public static void setUp() throws Exception {
         .getTestDir(TestCloseContainerEventHandler.class.getSimpleName());
     configuration
         .set(HddsConfigKeys.OZONE_METADATA_DIRS, testDir.getAbsolutePath());
+    configuration.setInt(
+        ScmConfigKeys.OZONE_SCM_PIPELINE_NUMBER_LIMIT, 16);
+
     nodeManager = new MockNodeManager(true, 10);
     eventQueue = new EventQueue();
     pipelineManager =
         new SCMPipelineManager(configuration, nodeManager, eventQueue);
     PipelineProvider mockRatisProvider =
         new MockRatisPipelineProvider(nodeManager,
-            pipelineManager.getStateManager(), configuration);
+            pipelineManager.getStateManager(), configuration, eventQueue);
     pipelineManager.setPipelineProvider(HddsProtos.ReplicationType.RATIS,
         mockRatisProvider);
     containerManager = new
         SCMContainerManager(configuration, nodeManager,
         pipelineManager, new EventQueue());
-    pipelineManager.triggerPipelineCreation();
     eventQueue.addHandler(CLOSE_CONTAINER,
         new CloseContainerEventHandler(pipelineManager, containerManager));
     eventQueue.addHandler(DATANODE_COMMAND, nodeManager);
+    pipelineManager.triggerPipelineCreation();
     // Move all pipelines created by background from ALLOCATED to OPEN state
     Thread.sleep(2000);
     TestUtils.openAllRatisPipelines(pipelineManager);
@@ -93,6 +97,9 @@ public static void tearDown() throws Exception {
     if (containerManager != null) {
       containerManager.close();
     }
+    if (pipelineManager != null) {
+      pipelineManager.close();
+    }
     FileUtil.fullyDelete(testDir);
   }
```

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineDatanodesIntersection.java
timmylicheng: Sure.

timmylicheng:
> +    if (pipelineManager != null) {
> +      pipelineManager.close();
> +    }

Thanks for the efforts! 
Looks like this test is not caught in previous PR when I enabled PipelinePlacementPolicy. I'm able to finish the test locally now.

xiaoyuyao:
Thanks @timmylicheng  for the update. The latest change LGTM, +1. Not sure why the acceptance test is failing here. Will take another look tomorrow before merge.

Also, can you open two follow up JIRAs for the other issues?

timmylicheng:
> Thanks @timmylicheng for the update. The latest change LGTM, +1. Not sure why the acceptance test is failing here. Will take another look tomorrow before merge.
> 
> Also, can you open two follow up JIRAs for the other issues?

Are you referring to:
1. Use either OZONE_SCM_PIPELINE_NUMBER_LIMIT or OZONE_DATANODE_MAX_PIPELINE_ENGAGEMENT to limit pipeline creation for all tests.

2. Use other method to notify pipeline creation failure due to exceeding max limit than throwing an exception?

timmylicheng:
> Thanks @timmylicheng for the update. The latest change LGTM, +1. Not sure why the acceptance test is failing here. Will take another look tomorrow before merge.
> 
> Also, can you open two follow up JIRAs for the other issues?

https://issues.apache.org/jira/browse/HDDS-2756 is creating to track logging issue.

xiaoyuyao:
bq. Use either OZONE_SCM_PIPELINE_NUMBER_LIMIT or OZONE_DATANODE_MAX_PIPELINE_ENGAGEMENT to limit pipeline creation for all tests.

Yes. I think the default value of 0 will be problematic as it will keep sending createPipeline command to DNs without restriction. A better default should be provided for it to use in production.

timmylicheng:
> OZONE_DATANODE_MAX_PIPELINE_ENGAGEMENT

https://issues.apache.org/jira/browse/HDDS-2772 is created to track this.

xiaoyuyao:
The acceptance test failures are unrelated tracked by HDDS-2774.

PR #292, opened 2019-12-04
HDDS-2645. Refactor MiniOzoneChaosCluster to a different package to add filesystem tests.

mukul1987:

Moves MiniOzone Chaos Cluster to a new module so that dependency to ozone-filesystem can be added. Also updated some of the logging utilities for the test.

https://issues.apache.org/jira/browse/HDDS-2645

## How was this patch tested?
Ran MiniOzoneChaos test locally after moving it to a new module and also verified that the files were created in the correct locations.

### DIFF
--- a/hadoop-ozone/fault-injection-test/mini-chaos-tests/pom.xml
+++ b/hadoop-ozone/fault-injection-test/mini-chaos-tests/pom.xml
@@ -18,19 +18,36 @@
 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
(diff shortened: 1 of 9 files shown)


anuengineer:
@mukul1987  There are 2 checkstyle issues -- 
` [ERROR] src/test/java/org/apache/hadoop/ozone/TestMiniChaosOzoneCluster.java:[92] (sizes) LineLength: Line is longer than 80 characters (found 81).
[ERROR] src/test/java/org/apache/hadoop/ozone/utils/LoadBucket.java:[63,23] (whitespace) ParenPad: '(' is followed by whitespace.
`

PR #293, opened 2019-12-04
HDDS-2530. Sonar : refactor verifyResourceName in HddsClientUtils to fix Sonar errors

supratimdeka:
https://issues.apache.org/jira/browse/HDDS-2530

refactored method to reduce cognitive complexity score (CC) + 2 other minor Sonar issues.
Added a unit test.

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
@@ -120,15 +120,7 @@ public final class HddsClientUtils {
         .toInstant().toEpochMilli();
   }
 
-  /**
-   * verifies that bucket name / volume name is a valid DNS name.
-   *
-   * @param resName Bucket or volume Name to be validated
-   *
-   * @throws IllegalArgumentException
-   */
-  public static void verifyResourceName(String resName)
-      throws IllegalArgumentException {
+  private static void doNameChecks(String resName) {
     if (resName == null) {
       throw new IllegalArgumentException("Bucket or Volume name is null");
     }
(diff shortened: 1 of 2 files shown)


hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
```diff
@@ -150,6 +142,49 @@ public static void verifyResourceName(String resName)
       throw new IllegalArgumentException("Bucket or Volume name "
           + "cannot end with a period or dash");
     }
+  }
+
+  private static boolean isLowercaseAlphanumeric(char c) {
+    return (Character.toString(c).matches("[a-z0-9]"));
```
adoroszlai: I think this can be improved:

1. `doNameChecks` can verify that `resName` `matches()` the pattern `[a-z0-9.-]+`, instead of checking each character separately in `doCharacterChecks`
2. `Pattern` can be pre-compiled as a constant

xiaoyuyao: We can avoid the string conversion and pattern match with the following:
return Character.isLowerCase(c) || Character.isDigit(c)

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
```diff
@@ -150,6 +142,49 @@ public static void verifyResourceName(String resName)
...
+  }
+
+  private static boolean isSupportedCharacter(char c) {
+    return (c == '.' || c == '-' || isLowercaseAlphanumeric(c));
+  }
+
+  private static void doCharacterChecks(char currChar, char prev) {
+    if (currChar > 'A' && currChar < 'Z') {
```
xiaoyuyao: Can we use Character.isUpperCase(c)

supratimdeka: Done

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
supratimdeka: Thanks for the review and suggestion @adoroszlai . I have removed the pattern match code altogether. Made the change as suggested in the comment from @xiaoyuyao

adoroszlai: Sure, that's fine, too.

anuengineer:
+1. LGTM. I am going to commit this.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
anuengineer: Unless Java inlines the isLower or isDigit, from the performance point of view, I bet that checking if (char < a || char > 9) is way more faster. If someone can profile and it prove that I am wrong, I would be very happy to learn how Java works. But from a gut perspective I think this is actually slower -- but then again, it really does not matter, and if it improves code reading I am all for it. Just commenting here for posterity's sake; so that someone profiles this and tells us the perf argument did not hold.

PR #296, opened 2019-12-04
HDDS-2659. KeyValueHandler#handleCreateContainer should log the exception on container creation failure

cxorm:
Let the ```StorageContainerException``` cause of ```KeyValueHandler#handleCreateContaine``` be logged.

https://issues.apache.org/jira/browse/HDDS-2659

## How was this patch tested?
Just build and compile.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/ContainerUtils.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/ContainerUtils.java
@@ -143,10 +143,10 @@ public final class ContainerUtils {
   public static ContainerCommandResponseProto logAndReturnError(
       Logger log, StorageContainerException ex,
       ContainerCommandRequestProto request) {
-    log.info("Operation: {} , Trace ID: {} , Message: {} , Result: {}",
-        request.getCmdType().name(), request.getTraceID(),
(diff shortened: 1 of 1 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/ContainerUtils.java
```diff
@@ -146,6 +146,7 @@ public static ContainerCommandResponseProto logAndReturnError(
     log.info("Operation: {} , Trace ID: {} , Message: {} , Result: {}",
         request.getCmdType().name(), request.getTraceID(),
         ex.getMessage(), ex.getResult().getValueDescriptor().getName());
+    log.error("StorageContainerException Occurred!", ex);
```
xiaoyuyao: We already have a log.info just above it with exception message and the result. If more information about the exception is needed, can we consolidate it into the previous message.

cxorm: Thanks @xiaoyuyao for the comment.
Updated.

anuengineer:
Not sure why the checks failed. This patch should not cause it. I have requested a re-run of the tests.

cxorm:
Thanks @xiaoyuyao, @anuengineer and @bharatviswa504 for the reviews.
And thanks @mukul1987 for the merge.

PR #297, opened 2019-12-04
HDDS-2661. Use field based Config annotation instead of method based

elek:

HDDS-2413 proposes an additional usage of the @Config annotation: to set configuration based on an existing configuration class.

But as of now we annotate the setters instead of the fields. To avoid annotation duplication (we need to read the values from the getters or the fields) I propose to switch to use field based annotations instead of setter based annotation.

I think it's more readable and additional validation (even the class level validations) can be done in a @PostConstruct method.

## How is it implemented

To support the annotation in private field we need to update our logic to check the fields of the parent class(es). It was not necessary with the setter based approach as the setter methods are public inherited. The patch updates both the annoation processor and the reflection based API (OzoneConfiguration).

With field based annotation injection it's not possible to put validation logic to the setters, therefor the standard `@PostConstruct` annotation is used to provide class level validation (or any other execution).

https://issues.apache.org/jira/browse/HDDS-2661

## How was this patch tested?

Mainly with unit tests. New unit test is introduced in configuration project and the TestOzoneConfiguration is also improved.


elek:
>  I would love if complexity of getObject could be reduced...

Sure, thanks the review / feedback. I moved a few steps to separated methods to improve the readability (like calling the `@PostConstruct` annotated method) but it's not a full solution. Let me know if you have more powerful idea and we can do it in follow up steps.

As a next step I will implement HDDS-2413 (the opposite direction for unit testing...(

adoroszlai:
> I moved a few steps to separated methods to improve the readability (like calling the `@PostConstruct` annotated method)

Thank you very much for improving it.

elek:
Thanks the review @adoroszlai. I will commit it after a new CI build (Just to me sure if it's still stable).

adoroszlai:
Thanks @elek for the new CI run, looks good.

PR #299, opened 2019-12-04
HDDS-2070. Create insight point to debug one specific pipeline

elek:

Wit the first implementation of ozone insight tool we had a demo insight-point to debug Ratis pipelines. It was not stable enough to include in the first patch, this patch is about fixing it.

The goal is to implement a new insight point (eg. datanode.pipeline) which can show information about one pipeline.

It can be done with retrieving the hosts of the pipeline and generate the loggers metrics (InsightPoint.getRelatedLoggers and InsightPoint.getMetrics) based on the pipeline information (same loggers should be displayed from all the three datanodes.

The pipeline id can be defined as a filter parameter which (in this case) should be required.

https://issues.apache.org/jira/browse/HDDS-2070

## How was this patch tested?

 1. Start a docker-compose cluster with three datanodes.
 2. Get the id of the Ratis/THREE pipeline: `ozone scmcli pipeline list | grep THREE`
 3. Check the logs of this specific pipeline: `ozone insight logs -v -f=pipeline=1e0dce39-98fe-43b9-b8fe-68fd003f321f datanode.pipeline`

### DIFF
--- a/hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/datanode/RatisInsight.java
+++ b/hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/datanode/RatisInsight.java
@@ -49,12 +49,13 @@ public class RatisInsight extends BaseInsightPoint implements InsightPoint {
   @Override
   public List<LoggerSource> getRelatedLoggers(boolean verbose,
       Map<String, String> filters) {
-    if (filters == null || !filters.containsKey(PIPLINE_FILTER)) {
-      throw new IllegalArgumentException(PIPLINE_FILTER
-          + " filter should be specified (-f pipeline=<pipelineid)");
+    if (filters == null || !filters.containsKey(PIPELINE_FILTER)) {
+      throw new IllegalArgumentException(PIPELINE_FILTER
+          + " filter should be specified (-f " + PIPELINE_FILTER
+          + "=<pipelineid)");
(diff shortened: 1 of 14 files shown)


hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/datanode/RatisInsight.java
```diff
@@ -39,31 +39,43 @@
...
-      try (ScmClient scmClient = createScmClient(conf)) {
-        pipeline = scmClient.listPipelines()
-            .stream()
-            .filter(d -> d.getNodes().size() > 1)
-            .findFirst();
+
+    try (ScmClient scmClient = createScmClient(conf)) {
+      Optional<Pipeline> pipelineSelection = scmClient.listPipelines()
+          .stream()
+          .filter(d -> d.getNodes().size() > 1)
```
adoroszlai: Do we need the size filter, given that pipeline ID is required?

elek: Not sure if this insight is useful for Ratis/ONE but let's remove it. If we would like to forbid the usage for Ratis/One it should be done with a more explicit check.

Line is removed.

adoroszlai:
Thanks @elek for the contribution.

PR #300, opened 2019-12-04
HDDS-2662. Update gRPC and datanode protobuf version in Ozone.

mukul1987:
This jira updates the gRPC and protobuf version to be same as the ratis-thirdparty versions.
This is needed after HDDS-2637

https://issues.apache.org/jira/browse/HDDS-2662

## How was this patch tested?
Ran MiniOzoneChaosCluster and verified that read/writes were working correctly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -174,8 +174,8 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
 
     <!-- Maven protoc compiler -->
     <protobuf-maven-plugin.version>0.5.1</protobuf-maven-plugin.version>
-    <protobuf-compile.version>3.5.0</protobuf-compile.version>
-    <grpc.version>1.10.0</grpc.version>
+    <datanode.protobuf-compile.version>3.10.0</datanode.protobuf-compile.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -174,8 +174,8 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
 
     <!-- Maven protoc compiler -->
     <protobuf-maven-plugin.version>0.5.1</protobuf-maven-plugin.version>
-    <protobuf-compile.version>3.5.0</protobuf-compile.version>
-    <grpc.version>1.10.0</grpc.version>
+    <datanode.protobuf-compile.version>3.10.0</datanode.protobuf-compile.version>
```
anuengineer: is this version that is used in Hadoop trunk?

elek: Hadoop 3.2 uses 3.7 https://github.com/apache/hadoop/blob/7c869b4df2503555de6a0e376aab4a98c655fc94/hadoop-project/pom.xml#L88

elek:
> Is it true? Is there any specific blocker issue or just a safety upgrade/

Discussed it offline with Mukul. It's mainly for safety but newer version also can provide additional performance benefit.

It started a new PR check to have a green acceptance test will merge it after...

PR #301, opened 2019-12-04
HDDS-2636. Refresh pipeline information in OzoneManager lookupFile call.

nandakumar131:

lookupFile call in OzoneManager doesn't refresh the pipeline information. As a result of this, wrong pipeline information is returned to the client and the client fails eventually. This change brings in support to refresh pipeline in lookupFile call.

https://issues.apache.org/jira/browse/HDDS-2636

## How was this patch tested?
* Added unit test to validate the change.
* Manually installed a 3 node cluster and tested it.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestKeyManagerUnit.java
+++ b/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestKeyManagerUnit.java
@@ -23,6 +23,8 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestKeyManagerUnit.java
```diff
@@ -177,4 +194,128 @@ private OmMultipartInfo initMultipartUpload(KeyManagerImpl omtest,
...
+        .setType(ReplicationType.RATIS)
+        .setFactor(ReplicationFactor.THREE)
+        .setState(Pipeline.PipelineState.OPEN)
+        .setLeaderId(dnOne.getUuid())
+        .setNodes(Arrays.asList(dnOne, dnTwo, dnThree))
+        .build();
+
+    final PipelineID pipelineTwoId = PipelineID.randomId();
+    final Pipeline pipelineTwo = Pipeline.newBuilder()
+        .setId(pipelineOneId)
```
avijayanhwx: Should this be pipelineTwoId?

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestKeyManagerUnit.java
```diff
@@ -177,4 +194,128 @@ private OmMultipartInfo initMultipartUpload(KeyManagerImpl omtest,
...
+
+    final OmKeyArgs.Builder keyArgs = new OmKeyArgs.Builder()
+        .setVolumeName("volumeOne")
+        .setBucketName("bucketOne")
+        .setKeyName("keyOne")
+        .setFactor(ReplicationFactor.THREE)
+        .setAcls(Collections.emptyList());
+
+    final OpenKeySession session = manager.createFile(keyArgs.build(),
+        false, false);
```
bharatviswa504: Write requests in old code will be soon removed. Can we use similar API's which are used to create bucket/volume or directly use OmMetadataManager keyTable put to add key?

nandakumar131: Agreed, even the read path in old code will also be removed, this change is just to make sure that we are not breaking any functionality in the old code path. Both the KeyManagerImpl and the corresponding test class will be removed eventually.

bharatviswa504: I think it will be better to add a similar API to create Key in TestOMRequestUtils, which will be helpful in future tests. (As right now there is no API to add the key to DB which takes OmKeyInfo as the argument.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestKeyManagerUnit.java
nandakumar131: Thanks for the catch.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestKeyManagerUnit.java
bharatviswa504: > even the read path in old code will also be removed

 We have not done any work for that, Read requests are using old code path. I am not sure there is any plan for rewrite read requests, as they can be used as-is for OM HA/non-HA

bharatviswa504: `make sure that we are not breaking any functionality in the old code path`
For old code path for write requests which is no more used, I think it is okay to not to use these API's and use an API to create a key in DB by adding a new API in TestOmRequestUtils. As this is a newly added test, it is okay to do this way.

bharatviswa504:
Thank You @nandakumar131 for the contribution and @avijayanhwx for the review.

PR #302, opened 2019-12-04
HDDS-2656. Prefer execute() over submit() if the returned Future is ignored

adoroszlai:

`Future submit(Runnable)` and `void execute(Runnable)` in `ExecutorService` have the same result.  If the returned `Future` is ignored, `execute` can be used instead of `submit` to avoid creating some objects.

https://issues.apache.org/jira/browse/HDDS-2656

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/333081135


anuengineer:
The acceptance tests all pass, but for some reason the script is returning a non-zero result. I am going to commit this patch.

adoroszlai:
Thanks @anuengineer for committing it.

PR #304, opened 2019-12-05
HDDS-1993. Merge OzoneManagerRequestHandler and OzoneManagerHARequest…

bharatviswa504:

Merge HARequestHandler and non-HA request handler.

https://issues.apache.org/jira/browse/HDDS-1993

## How was this patch tested?

Ran a few integration tests which execute this code path.


hanishakoneru:
LGTM. +1.
What do you think of having the similar structure for ReadRequests as we have for WriteRequests?

bharatviswa504:
> What do you think of having the similar structure for ReadRequests as we have for WriteRequests?

Do you mean to have their own request classes or not sure if this is what you are suggesting?

hanishakoneru:
Yes.. They can have their own request classes extending a parent ReadRequest class and override handleReadRequest() method. It was just a thought. 
I am +1 for this PR. Ee can merge this and if you agree open a newbie Jira for the above.

bharatviswa504:
> Yes.. They can have their own request classes extending a parent ReadRequest class and override handleReadRequest() method. It was just a thought.
> I am +1 for this PR. Ee can merge this and if you agree open a newbie Jira for the above.

Ya it is a good suggestion, but for read requests as we can use same code, no work is done in this area. But to be in a similar structure with write requests we can do that.

bharatviswa504:

PR #305, opened 2019-12-05
HDDS-2578. Handle InterruptedException in RandomKeyGenerator

dineshchitlangia:
Fix sonar issue.
Re-interrupt current thread when InterruptedException was encountered.

https://issues.apache.org/jira/browse/HDDS-2578

## How was this patch tested?
mvn install to ensure it builds cleanly.

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/RandomKeyGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/RandomKeyGenerator.java
@@ -1061,7 +1061,7 @@ public final class RandomKeyGenerator implements Callable<Void> {
             is.close();
           }
         } catch (IOException ex) {
-          LOG.error("Exception while validating write: " + ex.getMessage());
+          LOG.error("Exception while validating write.", ex);
         } catch (InterruptedException ex) {
           Thread.currentThread().interrupt();
         }
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/RandomKeyGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/RandomKeyGenerator.java
@@ -1060,11 +1060,10 @@ public final class RandomKeyGenerator implements Callable<Void> {
             }
             is.close();
           }
-        } catch (IOException | InterruptedException ex) {
+        } catch (IOException ex) {
           LOG.error("Exception while validating write: " + ex.getMessage());
-          if(ex instanceof InterruptedException) {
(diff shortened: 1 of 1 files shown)


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/RandomKeyGenerator.java
```diff
@@ -1062,6 +1062,9 @@ public void run() {
           }
         } catch (IOException | InterruptedException ex) {
           LOG.error("Exception while validating write: " + ex.getMessage());
+          if(ex instanceof InterruptedException) {
```
adoroszlai: The `instanceof` check triggers another warning:

> Exception types should not be tested using "instanceof" in catch blocks
>
> Code smell    Major    squid:S1193
>
> Multiple catch blocks of the appropriate type should be used instead of catching a general exception, and then testing on the type.

dineshchitlangia: @adoroszlai Addressed review comments in latest commit. 
Did you run the sonar check locally to discover this?
I want to understand so in future I can take care of these before I open a PR :)

adoroszlai: Thanks for updating the patch.

> Did you run the sonar check locally to discover this?

Yes, using [SonarLint](https://www.sonarlint.org/) plugin.  But this one I also remembered from another bug.

> I want to understand so in future I can take care of these before I open a PR :)

Thanks for that. :)

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/RandomKeyGenerator.java
```diff
@@ -1060,8 +1060,10 @@ public void run() {
             }
             is.close();
           }
-        } catch (IOException | InterruptedException ex) {
+        } catch (IOException ex) {
           LOG.error("Exception while validating write: " + ex.getMessage());
```
xiaoyuyao: should we change line 1064 like:
LOG.error("Exception while validating write.", ex);

dineshchitlangia: @xiaoyuyao addressed in latest commit.

bharatviswa504:
Thank You @dineshchitlangia for the contribution and @xiaoyuyao and @adoroszlai for the review.

dineshchitlangia:
Thanks everyone.

PR #306, opened 2019-12-05
HDDS-2667. Promethues reports invalid metric type

adoroszlai:

Normalize RocksDB metrics record name, which comes from DB file names (eg. `om.db -> Rocksdb_om.db`), because periods are not allowed in Prometheus metrics names.

https://issues.apache.org/jira/browse/HDDS-2667

## How was this patch tested?

Changed existing unit test.

Tested using `ozoneperf` environment after enabling RocksDB metrics in `docker-config`:

```
OZONE-SITE.XML_ozone.metastore.rocksdb.statistics=ALL
```

and creating test data with Freon:

```
docker-compose exec scm ozone freon ockg -t 1 -n 1
```

Verified that Prometheus shows all datanodes, OM, SCM as up.  Verified that `/prom` endpoint and Prometheus show RocksDB metrics with normalized names, eg:

```
# TYPE rocksdb_om_db_wal_file_bytes counter
rocksdb_om_db_wal_file_bytes{hostname="198e740b5533"} 1823
```

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/PrometheusMetricsSink.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/PrometheusMetricsSink.java
@@ -110,10 +110,11 @@ public class PrometheusMetricsSink implements MetricsSink {
   public String prometheusName(String recordName,
       String metricName) {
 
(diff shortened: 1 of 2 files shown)


adoroszlai:
Thanks @elek for committing it.

PR #310, opened 2019-12-05
HDDS-2542. Race condition between read and write stateMachineData.

lokeshj1703:

The write payload (the chunk itself) is sent to the Ratis as an external, binary byte array. It's not part of the LogEntry and saved from an async thread with calling ContainerStateMachine.writeStateMachineData

As it's an async thread it's possible that the stateMachineData is not yet written when the data should be sent to the followers in the next heartbeat.

By design a cache is used to avoid this issue but there are multiple problems with the cache.

First, the current cache size is chunkExecutor.getCorePoolSize() which is not enough. By default it means 60 executor threads and a cache with size 60. But in case of one very slow and 59 very fast writer the cache entries can be invalidated before the write.
In tests (freon datanode-chunk-writer-generator) @elek saw missed cache hits even with cache size 5000.

Second: as the readStateMachineData and writeStateMachien data are called from two different thread there is a race condition independent from the the cache size. It's possible that the write thread has not yet added the data to the cache but the read thread needs it.

The proposal is to replace the cache with an internal implementation of a Map interface. The default map would have limits so that map does not exceed a specified byte limit and specified limit on the number of entries.

https://issues.apache.org/jira/browse/HDDS-2542

## How was this patch tested?

There was a test added for the default map implementation. The CI runs should pass with the patch.


hanishakoneru:
@lokeshj1703 , patch LGTM. 
I am trying to understand what happens if the writeChunkFuture completes exceptionally? Do we not want to remove the entry from the stateMachineDataMap? Or will marking the stateMachine unhealthy take care of it?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/Map.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hdds.utils;
+
+public interface Map<K, V> {
```
bshashikant: Let's change the interface name to ResourceLimitMap instead. Map interface seems too generic and confusing.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/ResourceLimitMap.java
```diff
@@ -0,0 +1,69 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hdds.utils;
+
+
+import java.util.function.BiFunction;
+
+public class ResourceLimitMap<K, V> implements Map<K, V> {
```
bshashikant: ResourceLimitMap name can be change=d to stateMachhineCache.

bshashikant:
Thanks @lokeshj1703 for working on this. If the cache is full and entry cannot be added to the map because of no available permits, should we also add a timeout so the writeChunk executor thread gets released and does not block for significantly longer?

lokeshj1703:
@adoroszlai @hanishakoneru @mukul1987 @bshashikant Thanks for reviewing the PR! I have added a commit which addresses the review comments.

lokeshj1703:
@bshashikant I have renamed the classes to Cache and ResourceLimitCache.

lokeshj1703:
> I tried it with very low request count limit and checked behavior with `freon ockg`. With `ratis.leader.num.pending.requests=2` and `-t 1` the client could make progress, but with `-t 2` both threads were waiting indefinitely at `BlockOutputStream.waitOnFlushFutures`. The same happened with `ratis.leader.num.pending.requests=4` and `-t 4`.
@adoroszlai Here the issue is in Ratis. I will create new jiras and post them in the conversation.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java
```diff
@@ -588,10 +593,14 @@ private ByteString readStateMachineData(
...
-  private ByteString getCachedStateMachineData(Long logIndex, long term,
-      ContainerCommandRequestProto requestProto) throws ExecutionException {
-    return stateMachineDataCache.get(logIndex,
-        () -> readStateMachineData(requestProto, term, logIndex));
+  private ByteString getStateMachineData(Long logIndex, long term,
+      ContainerCommandRequestProto requestProto)
+      throws IOException {
+    ByteString data = stateMachineDataMap.get(logIndex);
+    if (data == null) {
+      data = readStateMachineData(requestProto, term, logIndex);
```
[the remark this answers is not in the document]
lokeshj1703: @mukul1987 I have created HDDS-2701 to address this issue.

lokeshj1703:
> If the cache is full and entry cannot be added to the map because of no available permits, should we also add a timeout so the writeChunk executor thread gets released and does not block for significantly longer?

This should not happen in our case as number of pending requests in Ratis is equal to entry limit for our stateMachineDataCache.

lokeshj1703:
> I am trying to understand what happens if the writeChunkFuture completes exceptionally? Do we not want to remove the entry from the stateMachineDataMap? Or will marking the stateMachine unhealthy take care of it?

@hanishakoneru Making stateMachine unhealthy should be sufficient as it would lead to closing of pipeline. However if we do change that behavior in future we would need to handle cache eviction. I have added comments in the code accordingly.

lokeshj1703:
@adoroszlai I have created https://issues.apache.org/jira/browse/RATIS-776 which addresses the issue you pointed out.

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/utils/TestResourceLimitCache.java
```diff
@@ -0,0 +1,82 @@
...
+    // permits out of 10
+    CompletableFuture future = CompletableFuture.supplyAsync(() -> {
+      try {
+        return resourceCache.put(1, "a");
+      } catch (InterruptedException e) {
+        e.printStackTrace();
+      }
+      return null;
+    });
+    Thread.sleep(100);
```
elek: Thank you very much @lokeshj1703 to address this issue.

I didn't check the full pach, but usually the Thread.sleep(...) cause test intermittency as in the CI environment it's possible that we have slower cpu than the local cpu.

Using GenericTestUtil.waitFor (with longer timeout and shorter checking period) can be more safe (IMHO).

lokeshj1703: @elek I have added a commit which addresses the review comments.
There are three places in the test where we call Thread.sleep(100). For two instances we are checking that a condition c is true before and after 100ms. These are not required to be replaced by GenericTestUtil.waitFor. The rest I have changed.

bshashikant:
> > If the cache is full and entry cannot be added to the map because of no available permits, should we also add a timeout so the writeChunk executor thread gets released and does not block for significantly longer?
> 
> @bshashikant This should not happen in our case as number of pending requests in Ratis is equal to entry limit for our stateMachineDataCache.

@lokeshj1703 , ideally we will be having different configs to tune this up and in the worst case scenario (for example in case of misconfigured cluster), we should atleast be able to free the thread instead of just waiting for it to be released. But this can be added as a separate task/jira altogether.

hanishakoneru:
Thanks @lokeshj1703 for working on this.
+1

adoroszlai:
> I have created https://issues.apache.org/jira/browse/RATIS-776 which addresses the issue you pointed out.

Thank you very much.  I'm +1 on the patch.

Should we wait for new Ratis snapshot/release with the fix for RATIS-776?  My concern is that if we hit this Ratis bug when reaching the given resource limit, and so we need to set the limit sufficiently high to avoid it, then we are in effect using an unlimited cache.

Another question: can we use `ResourceSemaphore` from Ratis instead of duplicating it?  I understand that current Ratis in Ozone has an older version of this class.  But is it feasible to replace the duplicate in a followup, when Ozone is updated to next Ratis snapshot/release?

lokeshj1703:
> Should we wait for new Ratis snapshot/release with the fix for RATIS-776? 

I agree. I tried running OzoneClientKeyGenerator with RATIS-776 and it works.

> Another question: can we use `ResourceSemaphore` from Ratis instead of duplicating it? 

Yes. I made some other changes to ResourceSemaphore class for the ContainerStateMachine specific use case. I will try and make them inside Ratis.

mukul1987:
@lokeshj1703, RATIS-776 is now resolved, can this be committed as well ?

lokeshj1703:
@mukul1987 I will create a ratis snapshot with RATIS-776 and RATIS-781 and update it in this jira.

lokeshj1703:
I have updated the branch to include RATIS-776. There is one change missing in RATIS-781. I will create a new jira to replace ResourceSemaphore.

lokeshj1703:
@bshashikant @adoroszlai Can you please take a look?

lokeshj1703:
Thanks everyone for the reviews! I have committed the PR to master branch.

PR #311, opened 2019-12-05
HDDS-2532. Sonar : fix issues in OzoneQuota

cxorm:
Fix 1.
```IllegalArgumentException``` extends ```RuntimeException```.
We can not throws it cause the kind of ```exception``` would not be checked while compiling.

Fix 2.
When boxed type ```Boolean``` is used as an expression it will throw ```NullPointerException``` if the value is null.
It is safer to avoid such conversion.

https://issues.apache.org/jira/browse/HDDS-2532

## How was this patch tested?
Ran on my standalone cluster.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
@@ -92,11 +92,8 @@ public class OzoneQuota {
    * @param quotaString Quota String
    *
    * @return OzoneQuota object
-   *
-   * @throws IllegalArgumentException
    */
-  public static OzoneQuota parseQuota(String quotaString)
-      throws IllegalArgumentException {
+  public static OzoneQuota parseQuota(String quotaString) {
 
     if ((quotaString == null) || (quotaString.isEmpty())) {
       throw new IllegalArgumentException(
@@ -136,10 +133,11 @@ public class OzoneQuota {
       found = true;
     }
 
-    if (!found) {
-      throw new IllegalArgumentException(
-          "Quota unit not recognized. Supported values are BYTES, MB, GB and " +
-              "TB.");
+    if (Boolean.FALSE.equals(found)) {
+      String exceptionInfo =
+          "Quota unit not recognized. " +
+          "Supported values are BYTES, MB, GB and TB.";
+      throw new IllegalArgumentException(exceptionInfo);
     }
 
     nSize = Integer.parseInt(size);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java [hunk 1]
```diff
@@ -136,10 +133,11 @@ public static OzoneQuota parseQuota(String quotaString)
       found = true;
     }
 
-    if (!found) {
-      throw new IllegalArgumentException(
-          "Quota unit not recognized. Supported values are BYTES, MB, GB and " +
-              "TB.");
+    if (Boolean.FALSE.equals(found)) {
```
adoroszlai: Since `found` is never `null`, can we change it to `boolean` instead?

anuengineer: and also, might be because I have not caught up with the new hip code styles; I find !found to be far easier to read than Boolean.FALSE.equals(found)

cxorm: Thanks @adoroszlai and @anuengineer for the reviews.
Yeah ```found``` is never ```null```, I converted its type to primitive type, and the ```if-statement``` is more readable.
Fixed.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -137,9 +134,10 @@ public static OzoneQuota parseQuota(String quotaString)
     }
 
     if (!found) {
-      throw new IllegalArgumentException(
-          "Quota unit not recognized. Supported values are BYTES, MB, GB and " +
-              "TB.");
+      String exceptionInfo =
+          "Quota unit not recognized. " +
+          "Supported values are BYTES, MB, GB and TB.";
+      throw new IllegalArgumentException(exceptionInfo);
```
dineshchitlangia: Let us avoid creating the variable `exceptionInfo` as it is not used after the next line.
We could still beautify the current line as:
``` suggestion
      throw new IllegalArgumentException("Quota unit not recognized." +
             "Supported values are BYTES, MB, GB and TB.");
```

cxorm: Thanks @dineshchitlangia for the advice.
Fixed.

cxorm:
Thanks @adoroszlai and @anuengineer for the reviews,
and thanks @dineshchitlangia for the commit.

PR #313, opened 2019-12-05
HDDS-2242. Avoid unnecessary rpc needed to discover the pipeline leader.

swagle:
Used the leaderId in the pipeline proto to connect directly to the Leader in XcieverClientRatis, this is a follow up from HDDS-1868. 

https://issues.apache.org/jira/browse/HDDS-2242

## How was this patch tested?
New integration tests added to test the changes.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
@@ -24,6 +24,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
@@ -130,6 +131,25 @@ public final class Pipeline {
    */
   public List<DatanodeDetails> getNodes() {
     return new ArrayList<>(nodeStatus.keySet());
+  }
+
+  /**
+   * Returns the leader if found else defaults to closest node.
+   *
+   * @return {@link DatanodeDetails}
+   */
+  public DatanodeDetails getLeaderNodeIfPresent() throws IOException {
+    if (nodeStatus.isEmpty()) {
+      throw new IOException(String.format("Pipeline=%s is empty", id));
+    }
+    Optional<DatanodeDetails> datanodeDetails =
+        nodeStatus.keySet().stream().filter(d ->
+            d.getUuid().equals(leaderId)).findFirst();
+    if (datanodeDetails.isPresent()) {
+      return datanodeDetails.get();
+    } else {
+      return getClosestNode();
+    }
   }
 
   public DatanodeDetails getFirstNode() throws IOException {
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
@@ -138,7 +138,7 @@ public final class Pipeline {
    *
    * @return {@link DatanodeDetails}
    */
-  public DatanodeDetails getLeaderNodeIfPresent() throws IOException {
+  public DatanodeDetails getLeaderNode() throws IOException {
     if (nodeStatus.isEmpty()) {
       throw new IOException(String.format("Pipeline=%s is empty", id));
     }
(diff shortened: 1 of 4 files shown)


swagle:
Acceptance test error if unrelated: [ERROR] Plugin org.jacoco:jacoco-maven-plugin:0.8.3 or one of its dependencies could not be resolved

adoroszlai:
> Acceptance test error if unrelated: [ERROR] Plugin org.jacoco:jacoco-maven-plugin:0.8.3 or one of its dependencies could not be resolved

True, but that doesn't mean it would pass otherwise. ;)

Can you please push an empty commit to re-trigger the checks?  (Or someone with write access can trigger it on the [Checks](https://github.com/apache/hadoop-ozone/pull/313/checks?check_run_id=337290486) page.)

```
git commit --allow-empty -m 'trigger new CI check' && git push
```

swagle:
@adoroszlai  thanks for suggestion!

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
```diff
@@ -132,6 +133,25 @@ void setLeaderId(UUID leaderId) {
     return new ArrayList<>(nodeStatus.keySet());
   }
 
+  /**
+   * Returns the leader if found else defaults to closest node.
+   *
+   * @return {@link DatanodeDetails}
+   */
+  public DatanodeDetails getLeaderNodeIfPresent() throws IOException {
```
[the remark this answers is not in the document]
swagle: Hi @nandakumar131, renamed to getLeaderNode() to fit in with similarly named methods for this class.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/hdds/scm/TestRatisPipelineLeader.java
```diff
@@ -0,0 +1,124 @@
...
+    // Verify client connects to Leader without NotLeaderException
+    XceiverClientRatis xceiverClientRatis =
+        XceiverClientRatis.newXceiverClientRatis(ratisPipeline.get(), conf);
+    GenericTestUtils.LogCapturer logCapturer =
+        GenericTestUtils.LogCapturer.captureLogs(GrpcClientProtocolClient.LOG);
+    xceiverClientRatis.connect();
+    logCapturer.stopCapturing();
+    Assert.assertFalse("Client should connect to pipeline leader on first try.",
+        logCapturer.getOutput().contains(
+            "org.apache.ratis.protocol.NotLeaderException"));
```
adoroszlai: Can this assertion ever fail?  No matter how I tried setting log level for `GrpcClientProtocolClient.LOG`, output was always empty.  Even changing `getLeaderNode()` to deliberately return a non-leader node didn't trigger any log.  Am I missing something?

swagle: Hi @adoroszlai, good catch, this was from memory, it seems we have since changed where this gets logged and the level. I have verified by returning non-leader on the newly added method that we get failure:

`2019-12-10 21:28:23,433 [ForkJoinPool.commonPool-worker-2] DEBUG client.GrpcClientProtocolService (GrpcClientProtocolService.java:sendReply(354)) - 0-OrderedRequestStreamObserver0: sendReply seq=1, RaftClientReply:client-C8FCD8002CD8->1f4d9226-afd7-40f5-9dd3-db27236750e8@group-A49D14340303, cid=0, FAILED org.apache.ratis.protocol.NotLeaderException: Server 1f4d9226-afd7-40f5-9dd3-db27236750e8@group-A49D14340303 is not the leader f12ee13d-5e47-4dc1-87ca-f330c773f86f:192.168.1.15:50058, logIndex=0, commits[1f4d9226-afd7-40f5-9dd3-db27236750e8:c-1, f12ee13d-5e47-4dc1-87ca-f330c773f86f:c0, 6f12f4a3-71e9-41ff-8fdb-714053a8f799:c0]`

adoroszlai: Thanks @swagle for checking.

swagle:
> The code will become a lot simpler if the leader node in `Pipeline` is represented as `DatanodeDetails` instead of just `UUID`.

Hi, @nandakumar131 this was changed based on your suggestion in HDDS-1868 :-)
Although it isn't affecting performance, can we refactor later, this patch was meant as downstream work and I was hoping not to change proto structure for it.

nandakumar131:
> > The code will become a lot simpler if the leader node in `Pipeline` is represented as `DatanodeDetails` instead of just `UUID`.
> 
> Hi, @nandakumar131 this was changed based on your suggestion in HDDS-1868 :-)
> Although it isn't affecting performance, can we refactor later, this patch was meant as downstream work and I was hoping not to change proto structure for it.

@swagle Even in HDDS-1868 the suggestion was to use `DatanodeDetails` instead of just `UUID`.
We can do it in a follow-up jira.

I will merge this shortly.

PR #314, opened 2019-12-05
HDDS-2555. Handle InterruptedException in XceiverClientGrpc

dineshchitlangia:
Sonar: Interrupted current thread to handle InterruptedException

https://issues.apache.org/jira/browse/HDDS-2555

## How was this patch tested?
Local build was successful.

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
@@ -234,6 +234,7 @@ public class XceiverClientGrpc extends XceiverClientSpi {
     } catch (ExecutionException e) {
       throw new IOException("Failed to execute command " + request, e);
     } catch (InterruptedException e) {
+      LOG.error("Command execution was interrupted ", e);
       Thread.currentThread().interrupt();
       return null;
     }
@@ -250,6 +251,7 @@ public class XceiverClientGrpc extends XceiverClientSpi {
     } catch (ExecutionException e) {
       throw new IOException("Failed to execute command " + request, e);
     } catch (InterruptedException e) {
+      LOG.error("Command execution was interrupted ", e);
       Thread.currentThread().interrupt();
       return null;
     }
@@ -345,6 +347,7 @@ public class XceiverClientGrpc extends XceiverClientSpi {
         ioException = new IOException(e);
         responseProto = null;
       } catch (InterruptedException e) {
+        LOG.error("Command execution was interrupted ", e);
         Thread.currentThread().interrupt();
         responseProto = null;
       }
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
@@ -231,8 +231,11 @@ public class XceiverClientGrpc extends XceiverClientSpi {
     try {
       return sendCommandWithTraceIDAndRetry(request, null).
           getResponse().get();
-    } catch (ExecutionException | InterruptedException e) {
+    } catch (ExecutionException e) {
       throw new IOException("Failed to execute command " + request, e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      return null;
     }
   }
 
@@ -244,8 +247,11 @@ public class XceiverClientGrpc extends XceiverClientSpi {
       XceiverClientReply reply;
       reply = sendCommandWithTraceIDAndRetry(request, validators);
       return reply.getResponse().get();
-    } catch (ExecutionException | InterruptedException e) {
+    } catch (ExecutionException e) {
       throw new IOException("Failed to execute command " + request, e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      return null;
     }
   }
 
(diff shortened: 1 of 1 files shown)


hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
```diff
@@ -231,8 +231,11 @@ public ContainerCommandResponseProto sendCommand(
     try {
       return sendCommandWithTraceIDAndRetry(request, null).
           getResponse().get();
-    } catch (ExecutionException | InterruptedException e) {
+    } catch (ExecutionException e) {
       throw new IOException("Failed to execute command " + request, e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
xiaoyuyao: Can we log an error here? Same apply to other places?

dineshchitlangia: Would you prefer something like:
`LOG.error("Execution was interrupted", e)` ?

dineshchitlangia: @xiaoyuyao addressed review comment in recent commit.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
```diff
@@ -244,8 +248,12 @@ public ContainerCommandResponseProto sendCommand(
       XceiverClientReply reply;
       reply = sendCommandWithTraceIDAndRetry(request, validators);
       return reply.getResponse().get();
-    } catch (ExecutionException | InterruptedException e) {
+    } catch (ExecutionException e) {
       throw new IOException("Failed to execute command " + request, e);
+    } catch (InterruptedException e) {
```
bharatviswa504: Now when Interrupted Exception happens, we return null. But I see in some places callers do not have a check for not null.

For instance
**ContainerProtocolCalls.java L 117:**
```
    ContainerCommandResponseProto response =
        xceiverClient.sendCommand(request, getValidatorList());
```

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
bharatviswa504: Same as the below comment. Can you verify once?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
dineshchitlangia: Thank you for catching this @bharatviswa504 .
It seems that in all such callers, the response is assumed to be non-null.

We can address this in following ways:
1. Avoid returning NULL and instead rethrow the InterruptedException
2. Return empty `response` instance
3. Return NULL and put a null check in caller method.

I am more inclined to option 1 as it keeps the code base simple.

What are your thoughts ?

bharatviswa504: Yes, we can rethrow with IOException

dineshchitlangia: Updated PR. Thanks @bharatviswa504 !

dineshchitlangia:
Findbug issue reported, I will fix it and update PR again.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
```diff
@@ -231,8 +232,13 @@ public ContainerCommandResponseProto sendCommand(
     try {
       return sendCommandWithTraceIDAndRetry(request, null).
           getResponse().get();
-    } catch (ExecutionException | InterruptedException e) {
+    } catch (ExecutionException e) {
       throw new IOException("Failed to execute command " + request, e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw (IOException) new InterruptedIOException(
```
bharatviswa504: We can directly throw IOException here, instead of casting

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
```diff
@@ -244,8 +250,13 @@ public ContainerCommandResponseProto sendCommand(
       XceiverClientReply reply;
       reply = sendCommandWithTraceIDAndRetry(request, validators);
       return reply.getResponse().get();
-    } catch (ExecutionException | InterruptedException e) {
+    } catch (ExecutionException e) {
       throw new IOException("Failed to execute command " + request, e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw (IOException) new InterruptedIOException(
```
bharatviswa504: Same as above

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
bharatviswa504: This is still missing?

dchitlangia:
@xiaoyuyao Your comments are addressed. Do you have further recommendations or is this good to be merged? Thanks.

xiaoyuyao:
Thanks @dineshchitlangia  for the update. LGTM, +1.

PR #316, opened 2019-12-05
HDDS-2680. Fix updating lastAppliedIndex in OzoneManagerStateMachine.

bharatviswa504:

This Jira is to fix a bug that has been caused by HDDS-2637.

Now we have applyTransactionMap where we put all entries of ratisTransactionIndex before complete.

And when updating lastAppliedIndex, now there is a chance that we updateLastAppliedIndex before DoubleBuffer flush has committed transactions to DB.

Let's take 1-10 are apply transaction entries which are not flushed, 11th is indexUpdate a metadata/conf entry transaction, now when notifyIndexUpdate is called with 11, we updateLastAppliedIndex to 11. (Which we should not do) This Jira is to fix this issue.

https://issues.apache.org/jira/browse/HDDS-2680

## How was this patch tested?

Trying to add some UT's to test this.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
@@ -446,4 +446,9 @@ public class OzoneManagerStateMachine extends BaseStateMachine {
     HadoopExecutors.shutdown(executorService, LOG, 5, TimeUnit.SECONDS);
     HadoopExecutors.shutdown(installSnapshotExecutor, LOG, 5, TimeUnit.SECONDS);
   }
+
+  @VisibleForTesting
+  void addApplyTransactionTermIndex(long term, long index) {
+    applyTransactionMap.put(index, term);
+  }
 }
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
@@ -22,6 +22,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.protobuf.ServiceException;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
@@ -148,7 +149,7 @@ public class OzoneManagerStateMachine extends BaseStateMachine {
     // placeHolderIndex (when a node becomes leader, it writes a conf entry
     // with some information like its peers and termIndex). So, calling
     // updateLastApplied updates lastAppliedTermIndex.
-    updateLastAppliedIndex(index, currentTerm, null, false);
+    computeAndUpdateLastAppliedIndex(index, currentTerm, null, false);
     snapshotInfo.updateTerm(currentTerm);
   }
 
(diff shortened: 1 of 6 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -349,22 +357,51 @@ private Message runCommand(OMRequest request, long trxLogIndex) {
...
+          appliedIndex = i;
+          final Long removed = applyTransactionMap.remove(i);
+          appliedTerm = removed;
+        } else if (ratisTransactionMap.containsKey(i)) {
+          final Long removed = ratisTransactionMap.remove(i);
+          appliedTerm = removed;
+          appliedIndex = i;
+        } else {
+          break;
+        }
```
hanishakoneru: @bharatviswa504, I am trying to think of cases in which this approach might fail. This scenario might not be possible so correct me if I am wrong.
Lets say we have multiple executors and the DoubleBuffer flushes the transactions as below.
Flush 1 -> Txns 1, 2, 4
Flush 2 -> Txns 3, 5, 6
Flush 3 -> Txns 7, 8, 9, 10

Now, when the 1st flush happens and it calls updateLaspAppliedIndex, for txn 1 and 2 there will not be a problem. Then it will look for txn 3 which is not a part of either flushedEpochs or ratisTransactionMap. Hence, the loop will break and the SM lastAppliedIndex will be updated to 2.
The next flush has txns 3, 5 and 6. The same scenario will be repeated and the SM lastAppliedIndex will be updated to 3.
Now when the 3rd flush calls updateLastAppliedIndex, the SM lastAppliedIndex will not be updated as the we have no record of txns 4, 5 and 6 being flushed to disk.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -349,22 +357,51 @@ private Message runCommand(OMRequest request, long trxLogIndex) {
...
-  public synchronized void updateLastAppliedIndex(long lastFlushedIndex) {
-    Long appliedTerm = null;
-    long appliedIndex = -1;
-    for(long i = getLastAppliedTermIndex().getIndex() + 1;
-        i <= lastFlushedIndex; i++) {
-      final Long removed = applyTransactionMap.remove(i);
-      if (removed == null) {
-        break;
+  private synchronized void updateLastAppliedIndex(long lastFlushedIndex,
+      long currentTerm, List<Long> flushedEpochs, boolean checkMap) {
```
hanishakoneru: Can we rename this method to something else? This is very confusing with BaseStateMachine's updateLastAppliedTermIndex.
Maybe computeAndUpdateLastAppliedIndex?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
bharatviswa504: I think with multiple executors this can certainly possible. Thanks for catching it. Fixed to handle this issue in latest commits.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
bharatviswa504: Agreed. Done.

hanishakoneru:
Thank you @bharatviswa504 for working on this. 
Latest changes LGTM. +1.

PR #320, opened 2019-12-06
HDDS-2683. Ratis MaxBuffer should be the same size as the segment size.

mukul1987:

As noticed with RATIS-767, the write buffer is allocated with 32MB size whereas the raft log segment is 1MB. The rest of the buffer will not be used and the memory is wasted allocation.

This jira sets the max buffer size same as segment size.

https://issues.apache.org/jira/browse/HDDS-2683

## How was this patch tested?
Ran MiniOzoneChaosCluster with the new config to ensure data read/write are working fine.


lokeshj1703:
@mukul1987 Thanks for the contribution! I have merged the PR to master branch.

PR #321, opened 2019-12-06
HDDS-2685. Fix Rename API in BasicOzoneFileSystem

ayushtkn:

In the Rename API :
1. This doesn't work if one of the path contains URI and other doesn't.

    if (src.equals(dst)) {
      return true;
    }

2. This check is suppose to be done only for directories, but is done for Files too, it can be moved after getting the FileStatus and checking the type.

    // Cannot rename a directory to its own subdirectory
    Path dstParent = dst.getParent();
    while (dstParent != null && !src.equals(dstParent)) {
      dstParent = dstParent.getParent();
    }
    Preconditions.checkArgument(dstParent == null,
        "Cannot rename a directory to its own subdirectory");

3. This too doesn't work (similar to 1.)

    if (srcStatus.isDirectory()) {
      if (dst.toString().startsWith(src.toString() + OZONE_URI_DELIMITER)) {
        LOG.trace("Cannot rename a directory to a subdirectory of self");
        return false;
      }

4. Rename is even success if the URI provided is of different FileSystem.
In general HDFS/Other FS shall throw IllegalArgumentException if the path doesn't belong to the same FS.

https://issues.apache.org/jira/browse/HDDS-2685

## How was this patch tested?

Added UT.

### DIFF
--- a/hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFsRenameDir.java
+++ b/hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFsRenameDir.java
@@ -18,7 +18,6 @@
 
 package org.apache.hadoop.fs.ozone;
 
-import java.io.IOException;
 
 import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
 import org.apache.hadoop.fs.FileSystem;
--- a/hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFsRenameDir.java
+++ b/hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFsRenameDir.java
@@ -29,7 +29,7 @@ import org.apache.hadoop.ozone.MiniOzoneCluster;
 import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.TestDataUtil;
 import org.apache.hadoop.ozone.client.OzoneBucket;
-
+import org.apache.hadoop.test.LambdaTestUtils;
 import org.junit.After;
 import static org.junit.Assert.assertTrue;
 import org.junit.Before;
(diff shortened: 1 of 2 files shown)


ayushtkn:
Thanx @adoroszlai 
Have Fixed CheckStyle and Updated the description.

hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFsRenameDir.java
```diff
@@ -98,5 +98,13 @@ public void testRenameDir() throws IOException {
     // sub-directories of the renamed directory have also been renamed.
     assertTrue("Keys under the renamed direcotry not renamed",
         fs.exists(new Path(dest, "sub_dir1")));
+
```
steveloughran: typo worth fixing just above

steveloughran:
a test for rename under self would round this out. it might be best in AbstractContractRenameTest for all to check for

hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFsRenameDir.java
ayushtkn: fixed.

ayushtkn:
Thanx @steveloughran for the review, I have fixed the typo.

> it might be best in AbstractContractRenameTest 
Makes sense to add, AbstractContractRenameTest is in Hadoop-Common, Should I create a new PR with this same JIRA ID to the other Repo, or should I raise a new JIRA for this. Or there is some other way. Let me know the process, I am not sure about of it. 

Out of Context : Observed while fixing the typo, same was there in BasicOzoneFileSystem, in Delete API, so fixed that too, 

There is a logic to prevent Root from getting deleted. Just couple of lines above.
`
      if (key.equals("/")) {
        return false;
      }`

I don't think it works, if it is root the key will be empty not "/", As pathToKey() will remove the "/" and make it "" and there after addTrailingSlashIfNeeded() will not add a slash as it has a check !isEmpty(key) before adding slash.

hanishakoneru:
Thanks @ayushtkn for working on this.
LGTM. +1. I will merge shortly.
We can create a new Jira for adding rename to self test in AbstractContractRenameTest.
Also, do you want to open a new Jira to prevent root from getting deleted as you pointed?

PR #271, opened 2019-11-26
HDDS-1812. Du while calculating used disk space reports that chunk files are file not found

adoroszlai:
## What changes were proposed in this pull request?

1. Copy `DU` from `hadoop-common` and customize it to ignore temporary chunk files.  Unfortunately the original implementation could not be simply extended.  Note that this means temp chunk files will not be counted towards HDDS space usage.  One possible alternative is ignoring errors from `du` (which returns with error due to the missing files), but that might make it more difficult to spot other problems.
2. Introduce an interface for creating various "space usage" implementations.  `hadoop-common` allows a single implementation to be configured, which is used for all volumes.  The factory interface allows creating different implementations or different parameters for different volumes.  Currently 3 factories are implemented:
   1. pure `du`-based
   2. pure `df`-based (this is actually based on `java.io` in both Hadoop Common and HDDS)
   3. mock for testing -- this also helps avoid test failures due to "disk out of space" errors when allocating containers
   Hybrid or other more intelligent implementation could be a future improvement.  Eg. one that would use `du` for shared volumes and `df` for dedicated ones.
3. Introduce an interface for the object that persists space usage info (to a file named `scmUsed`) to be able to:
   1. skip persisting info from cheaper sources (ie. `df`)
   2. use in-memory persistence for unit testing
4. Decouple caching and periodic refresh of space usage from its source.  Use a `ScheduledExecutorService` for refreshing disk usage info.  Use a "same thread executor" for unit tests.
5. Let `container-service` tests use code from HDDS `common` tests, which provide mock implementations of the new interfaces.  This also required:
   1. Avoid using real config properties (eg. `ozone.scm.client.bind.host`) for testing config file generation.  The generated config file is picked up by other tests and causes failures.
   2. Rename Log4J2 config file used by audit logger test to avoid creating an untracked `audit.log` (similar to HDDS-2063).  This would happen if some test starts components which use audit logger, and it picks up `log4j2.properties` by default.

https://issues.apache.org/jira/browse/HDDS-1812

## How was this patch tested?

Added several unit tests, even for logic extracted from existing classes.  On compose-based cluster tested configuring each of the real factory implementations and configuring different refresh periods.  Tested `du`-based implementation by continuously creating/renaming files from shell.

PR #322, opened 2019-12-06
HDDS-2684. Refactor common test utilities to hadoop-hdds/common

adoroszlai:

Expose test code from `hadoop-hdds/common` to other modules.  Move some "common" test utilities (random `DatanodeDetails` and `Pipeline` creation).  The goal is to be able to write unit tests for `client` module without code duplication.

This was partly extracted from #271:

 * Avoid using real config properties (eg. `ozone.scm.client.bind.host`) for testing config file generation.  The generated config file is picked up by other tests and causes failures.
 * Rename Log4J2 config file used by audit logger test to avoid creating an untracked `audit.log` (similar to [HDDS-2063](https://issues.apache.org/jira/browse/HDDS-2063)).  This would happen if some test starts components which use audit logger, and it picks up `log4j2.properties` by default.

https://issues.apache.org/jira/browse/HDDS-2684

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/336769633


hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/conf/SimpleConfiguration.java
```diff
@@ -22,7 +22,7 @@
 /**
  * Example configuration to test the configuration injection.
  */
-@ConfigGroup(prefix = "ozone.scm.client")
+@ConfigGroup(prefix = "fake.scm.client")
```
xiaoyuyao: can we name the prefix like test.scm.client?

hadoop-hdds/container-service/pom.xml
```diff
@@ -33,6 +33,11 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-hdds-common</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-hdds-common</artifactId>
+      <type>test-jar</type>
```
xiaoyuyao: Do we need hadoop-hdds-common as a test-jar type for container-service?

hadoop-hdds/pom.xml
```diff
@@ -186,6 +193,13 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
         <version>${junit.jupiter.version}</version>
         <scope>test</scope>
       </dependency>
+
+      <dependency>
+        <groupId>org.mockito</groupId>
+        <artifactId>mockito-core</artifactId>
+        <version>2.2.0</version>
```
xiaoyuyao: NIT: can we replace the hard-coded package version?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestSCMNodeManager.java
```diff
@@ -1113,7 +1115,7 @@ public void testScmRegisterNodeWith4LayerNetworkTopology()
     try (SCMNodeManager nodeManager = createNodeManager(conf)) {
       DatanodeDetails[] nodes = new DatanodeDetails[nodeCount];
       for (int i = 0; i < nodeCount; i++) {
-        DatanodeDetails node = TestUtils.createDatanodeDetails(
+        DatanodeDetails node = MockDatanodeDetails.createDatanodeDetails(
```
xiaoyuyao: NIT: MockDatanodeDetails can be removed as static import is added.

hadoop-hdds/container-service/pom.xml
adoroszlai: Yes, AFAIK that makes the test code from `hadoop-hdds-common` available to the other project.

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/conf/SimpleConfiguration.java
adoroszlai: Good idea, thanks.

hadoop-hdds/pom.xml
adoroszlai: I took this one from `container-service`.  It turns out we have quite a few different version numbers.  So I defined a property for the latest 2.x in root `pom.xml` in 0ebb158fcc793651aa19b3a43bc906fcaa43ae59.

I left Mockito 1.x usage and versions as is, because it's not API-compatible with 2.x.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestSCMNodeManager.java
adoroszlai: The static import was previously only for another method, but added this one as well.

adoroszlai:
Thanks @xiaoyuyao for reviewing this.  I have updated the change with your suggestions.

adoroszlai:
@xiaoyuyao may I ask you to take another look?

arp7:
+1

Looks like Xiaoyu's comments are addressed. Xiaoyu - if not we can address in a followup Jira.

adoroszlai:
Thanks @arp7 for reviewing and committing it.

PR #323, opened 2019-12-06
HDDS-2686. Use protobuf 3 instead of protobuf 2

elek:

Protobuf2 is 4.5 years old, Hadoop trunk already upgraded to use 3.x protobuf.

Would be great to use recent protobuf version which can also provide performance benefit and using new features.

https://issues.apache.org/jira/browse/HDDS-2686

## How was this patch tested?

Run a few acceptance test locally (secure/unsecure) and they are passed. Cluster worked well.

Waiting for a full acceptance run.


pom.xml
```diff
@@ -149,7 +149,7 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
 
     <!-- ProtocolBuffer version, used to verify the protoc version and -->
     <!-- define the protobuf JAR version                               -->
-    <protobuf.version>2.5.0</protobuf.version>
+    <protobuf.version>3.10.0</protobuf.version>
     <protoc.path>${env.HADOOP_PROTOC_PATH}</protoc.path>
```
mukul1987: protoc path will not really be required after this changed right ? Can this we rmeoved ?

hadoop-hdds/common/src/main/proto/SCMSecurityProtocol.proto
```diff
@@ -21,6 +21,7 @@
  * Please see http://wiki.apache.org/hadoop/Compatibility
  * for what changes are allowed for a *unstable* .proto interface.
  */
+syntax = "proto2";
```
mukul1987: I didnt' understand it fully, if we are compiling the files using proto3, can we also move the format to proto3 as well ?

hadoop-hdds/pom.xml
```diff
@@ -229,6 +229,16 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <artifactId>picocli</artifactId>
       <version>3.9.6</version>
     </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-stub</artifactId>
```
mukul1987: What is the requirement of this grpc dependency ? We already have grpc dependency for datanode client proto files.

pom.xml
```diff
@@ -149,7 +149,7 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
 
     <!-- ProtocolBuffer version, used to verify the protoc version and -->
     <!-- define the protobuf JAR version                               -->
-    <protobuf.version>2.5.0</protobuf.version>
+    <protobuf.version>3.10.0</protobuf.version>
```
mukul1987: Can we rename this to something like hadoop-protobuf-compile version ?

mukul1987: Also can this be the same as the "<protobuf-compile.version>3.5.0</protobuf-compile.version>"

hadoop-hdds/common/src/main/proto/SCMSecurityProtocol.proto
anuengineer: Compiling only. So no, we cannot move to Protoc3 in this patch. That would require some serious testing and work. But it is an amazing goal to have. I have probably try to change the DN protocol since it is on GRPC and independent of Hadoop RPC libraries. Eventually, we will get there, but baby steps.

elek: Agree. proto3 syntax is different. Easier to upgrade the compilation library first and update the proto files to syntax 3 in the next step.

pom.xml
elek: Sure. Removed.

hadoop-hdds/pom.xml
elek: As we have service entries in the proto files, the service stubs are also generated which require these jar files (I got class not found exception without these dependencies).

I tried to use the ratis grpc (with search and replace all the package names similar what we have in the datanode) but it didn't work very well as some common package names (like ServiceException) are also renamed (=package rename). But Hadoop rpc couldn't work with the renamed exceptions.

Finally it turned out the simplest way just to add these dependencies.

And I would like to experiment with grpc for the heartbeat protocol which requires these jar files anyway.

I think it's safe to keep them as of now.

pom.xml
elek: In #300 `protobuf-compile.version` has been disappeared and we have only `datanode.protobuf-compile.version`. I reorganized this setting to be more similar:

```
   <!-- Maven protoc compiler -->
    <protobuf-maven-plugin.version>0.5.1</protobuf-maven-plugin.version>
    <hadooprpc.protobuf-compile.version>3.10.0</hadooprpc.protobuf-compile.version>
    <datanode.protobuf-compile.version>3.10.0</datanode.protobuf-compile.version>
 ```

But let me know if you have any better idea...

hadoop-hdds/common/src/main/proto/SCMSecurityProtocol.proto
mukul1987: Thanks makes sense.

pom.xml
mukul1987: This sounds perfect to me.

elek:
> but there are acceptance test failures.

Yes, the problem is in the mr test with hadoop-32. With hadoop 2.7 - 3.1 we have no problems as we use properly isolated classpath environments ("legacy" ozonesf jar) but until now we were compatible with hadoop 3.2. 

With this change the `current` jar will be compatible with the (unreleased) hadoop 3.3 and we need to use the classpath separation for 3.2.

I updated the tests and created a new mr test for 3.3 (based on a custom docker image which can be replaced after the release with the official apache image).

Note: shading is not working for the protobuf only the classpath isolation. With the `current` jar we are not shading the hadoop API which includes references to the protobuf GenericMessage. Without shading the Hadoop API we can't shade the protobuf files, therefore the only option is to use the classpath isolation.

mukul1987:
Hi @elek , there are still acceptance test failure with the patch. Can you please take a look.

elek:
Closing this PR temporary as this patch requires to fix the classpath isolation in secure environments. Without isolation we can't bump the PB version.

PR #324, opened 2019-12-07
HDDS-2688. Refactor container response builders to hadoop-hdds/common

adoroszlai:

Move `ContainerCommandResponseProto` builder methods scattered across various utility classes to `ContainerCommandResponseBuilders` in `hadoop-hdds/common`.  These would be useful for client-side unit tests, but `client` does not depend on `container-service` since the interfaces and messages it needs are defined in `common`.

Also change the signature of few of these builder methods to use proto types instead of helper counterparts (`BlockData` and `ChunkInfo`).  This lets us avoid some duplicated proto message building.

https://issues.apache.org/jira/browse/HDDS-2688

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/338003076

`TestContainerMetrics` and `TestContainerSmallFile` from integration tests.


adoroszlai:
@lokeshj1703 please review

adoroszlai:
Thanks @bshashikant for reviewing and committing this.

PR #328, opened 2019-12-09
HDDS-2671 Have NodeManager.getNodeStatus throw NodeNotFoundException

sodonnel:

Currently, the SCM node manager method getNodeStatus catches any NodeNotFoundException and returns null.

```
  /**
   * Returns the node status of a specific node.
   *
   * @param datanodeDetails Datanode Details
   * @return NodeStatus for the node
   */
  @Override
  public NodeStatus getNodeStatus(DatanodeDetails datanodeDetails) {
    try {
      return nodeStateManager.getNodeStatus(datanodeDetails);
    } catch (NodeNotFoundException e) {
      // TODO: should we throw NodeNotFoundException?
      return null;
    }
  }
```

This should throw the exception to ensure downstream code does not need to perform a null check each time it is called.

https://issues.apache.org/jira/browse/HDDS-2671

## How was this patch tested?

Tested by existing unit tests which make calls to this API.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -576,7 +576,7 @@ public class ReplicationManager implements MetricsSource {
         LOG.warn("Cannot replicate container {}, no healthy replica found.",
             container.containerID());
       }
-    } catch (IOException | RuntimeException ex) {
+    } catch (IOException | IllegalStateException ex) {
       LOG.warn("Exception while replicating container {}.",
           container.getContainerID(), ex);
     }
@@ -789,7 +789,7 @@ public class ReplicationManager implements MetricsSource {
 
   /**
    * Wrap the call to nodeManager.getNodeStatus, catching any
-   * NodeNotFoundException and instead throwing a RuntimeException.
+   * NodeNotFoundException and instead throwing an IllegalStateException.
    * @param dn The datanodeDetails to obtain the NodeStatus for
    * @return NodeStatus corresponding to the given Datanode.
    */
@@ -797,7 +797,7 @@ public class ReplicationManager implements MetricsSource {
     try {
       return nodeManager.getNodeStatus(dn);
     } catch (NodeNotFoundException e) {
-      throw new RuntimeException("Unable to find NodeStatus for "+dn, e);
+      throw new IllegalStateException("Unable to find NodeStatus for "+dn, e);
     }
   }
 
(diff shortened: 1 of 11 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -785,6 +787,20 @@ private void sendDeleteCommand(final ContainerInfo container,
...
+   * Wrap the call to nodeManager.getNodeStatus, catching any
+   * NodeNotFoundException and instead throwing a RuntimeException.
+   * @param dn The datanodeDetails to obtain the NodeStatus for
+   * @return NodeStatus corresponding to the given Datanode.
+   */
+  private NodeStatus getNodeStatus(DatanodeDetails dn) {
+    try {
+      return nodeManager.getNodeStatus(dn);
+    } catch (NodeNotFoundException e) {
+      throw new RuntimeException("Unable to find NodeStatus for "+dn, e);
```
adoroszlai: I think this should be more specific, eg. `IllegalStateException` seems to be a good candidate.  If you agree, please remember to update the corresponding `catch`, too.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/SimpleMockNodeManager.java
```diff
@@ -87,7 +87,8 @@ public void setPipelines(DatanodeDetails dd, int count) {
    *         Inservice and Healthy NodeStatus.
    */
   @Override
-  public NodeStatus getNodeStatus(DatanodeDetails datanodeDetails) {
+  public NodeStatus getNodeStatus(DatanodeDetails datanodeDetails)
+      throws NodeNotFoundException {
```
adoroszlai: Shouldn't this method really throw `NodeNotFoundException` if `dni` is not found, instead of returning a healthy status, to more closely reflect actual `NodeManager` behavior?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: Yea that makes sense. This means that is a "real" runtime exception occurs, the catch will not stop it propagating up the stack and instead we only catch this more specific exception. I have made this change.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/SimpleMockNodeManager.java
sodonnel: Yea it would make more sense for this method do thow NodeNotFoundException, however, I cheated somewhat when using this class in an earlier patch within TestReplicationManager. Replication manager only calls the "getNodeStatus()" method of the NodeManager, and to avoid having to change too many existing tests I defaulted this to return "Healthy + Inservice" for any node not registered so the existing tests passed without modification.

I have an existing Jira HDDS-2673 to refactor and merge MockNodeManager and SimpleMockNodeManager into one, so perhaps we could improve this issue as part of that?

adoroszlai: > perhaps we could improve this issue as part of that?

Sure, that's fine, too.

sodonnel: I added a comment to HDDS-2673 to remind about this when I get to working on it.

PR #329, opened 2019-12-09
HDDS-2423. Add the recover-trash command client side code

cxorm:
This PR adds the client side of ```recover-trash ```.

Fix 1. Add the new fields on protobuf.
Fix 2. Add the corresponding method on ```RpcClient```.
Fix 3. Add the corresponding method on ```OzoneManager``` and the method would be completed on following PR.

https://issues.apache.org/jira/browse/HDDS-2423

## How was this patch tested?
Build, compile and checkstyle simply.
UTs would be updated on following JIRA.

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
@@ -721,6 +721,14 @@ public class RpcClient implements ClientProtocol {
   }
 
   @Override
+  public boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationName) throws IOException {
+
+    return ozoneManagerClient.recoverTrash(volumeName, bucketName, keyName,
+        destinationName);
+  }
+
+  @Override
   public OzoneKeyDetails getKeyDetails(
       String volumeName, String bucketName, String keyName)
       throws IOException {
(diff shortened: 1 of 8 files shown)


hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
```diff
@@ -720,6 +720,14 @@ public void renameKey(String volumeName, String bucketName,
         keyPrefix, maxKeys);
   }
 
+  @Override
+  public boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationName) throws IOException {
```
xiaoyuyao: NIT: can we rename destinationName  to destinationBucket to be consistent?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2238,6 +2238,15 @@ public void deleteKey(OmKeyArgs args) throws IOException {
     }
   }
 
+  // TODO: HDDS-2424. recover-trash command server side handling.
+  @Override
+  public boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationBucket) throws IOException {
+
+    boolean recoverOperation = true;
+    return recoverOperation;
```
xiaoyuyao: Even this is a partial work, is IT possible to add a unit test for it, which can be easily extended later for HDDS-2424.

cxorm: Thanks @xiaoyuyao for the review.

I updated the UT that is prototype for now,
and it can be extended for the following server-side development.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
cxorm: Thanks @xiaoyuyao for this NIT.
Fixed.

cxorm:
Thanks @xiaoyuyao for the review, 
and thanks @elek for the commit.

PR #331, opened 2019-12-09
HDDS-2188. Implement LocatedFileStatus & getFileBlockLocations to pro…

avijayanhwx:
…vide node/localization information to Yarn/Mapreduce.

For applications like Hive/MapReduce to take advantage of the data locality in Ozone, Ozone should return the location of the Ozone blocks. This is needed for better read performance for Hadoop Applications.
`
        if (file instanceof LocatedFileStatus) {
          blkLocations = ((LocatedFileStatus) file).getBlockLocations();
        } else {
          blkLocations = fs.getFileBlockLocations(file, 0, length);
        }
`

https://issues.apache.org/jira/browse/HDDS-2188

## How was this patch tested?
Manually tested.


hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
```diff
@@ -677,6 +677,7 @@ message KeyArgs {
     // request type.
     optional uint64 modificationTime = 13;
     optional bool sortDatanodes = 14;
+    optional bool refreshPipeline = 15;
```
bharatviswa504: Do we need this flag at all? As from the code it looks like we always want to return with the latest block information or are there any case we want to send it to false?. So, in OzoneManagerRequestHandler, we can just set omKeyArgs.set(true) right?

Similar is done for other requests like lookupFile.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
```diff
@@ -1077,6 +1077,7 @@ private GetFileStatusResponse getOzoneFileStatus(
         .setVolumeName(keyArgs.getVolumeName())
         .setBucketName(keyArgs.getBucketName())
         .setKeyName(keyArgs.getKeyName())
+        .setRefreshPipeline(keyArgs.getRefreshPipeline())
```
bharatviswa504: Do we need to set refreshPipeline to true in list status also?

avijayanhwx: @bharatviswa504 Yes, FS listStatus API is invoked by MapReduce jobs to get the located file blocks.

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
avijayanhwx: Yes, I agree. I can make the change.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
bharatviswa504: So is that missing in OzoneManagerRequestHandler.java?

avijayanhwx: Yes, it is not being set to true in OzoneManagerRequestHandler#listStatus currently.

avijayanhwx:
The acceptance test failures are due to this patch. I am looking into it.

avijayanhwx:
@elek I have fixed the acceptance test failures by refactoring the code based on our discussion.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1683,6 +1683,9 @@ public OzoneFileStatus getFileStatus(OmKeyArgs args) throws IOException {
           volumeName, bucketName, keyName);
       OmKeyInfo fileKeyInfo = metadataManager.getKeyTable().get(fileKeyBytes);
       if (fileKeyInfo != null) {
+        if (args.getRefreshPipeline()) {
```
bharatviswa504: Missing usage of refreshPipeline in listStatus

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -440,7 +450,62 @@ private FileStatusAdapter toFileStatusAdapter(OzoneFileStatus status) {
...
         status.getGroup(),
-        status.getPath()
+        status.getPath(),
+        getBlockLocations(status)
     );
   }
+
+  /**
+   * Helper method to get List of BlockLocation from OM Key info.
+   * @param omKeyLocationInfoGroups OM Key locations.
```
bharatviswa504: Incorrect parameters in javadoc

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
avijayanhwx: @bharatviswa504 listStatus calls getFileStatus internally. Hence I used it only in getFileStatus.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
avijayanhwx: Will fix it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
bharatviswa504: I see that is not the case always, can you once check. Not sure if I am missing something.

avijayanhwx: Yes, you are right. Thanks for the review!

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -440,7 +450,61 @@ private FileStatusAdapter toFileStatusAdapter(OzoneFileStatus status) {
...
+
+    int i = 0;
+    for (OmKeyLocationInfo omKeyLocationInfo :
+        omKeyLocationInfoGroup.getBlocksLatestVersionOnly()) {
+      List<String> hostList = new ArrayList<>();
+      List<String> nameList = new ArrayList<>();
+      omKeyLocationInfo.getPipeline().getNodes()
+          .forEach(dn -> {
+            hostList.add(dn.getHostName());
+            int port = dn.getPort(
```
bharatviswa504: Here hostList and name list they need to be Set, as the same Pipeline can be across multiple OmKeyLocationInfo (blocks). In the above case, list can have duplicates.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -440,7 +450,61 @@ private FileStatusAdapter toFileStatusAdapter(OzoneFileStatus status) {
...
+    }
+    List<OmKeyLocationInfoGroup> omKeyLocationInfoGroups =
+        keyInfo.getKeyLocationVersions();
+    if (CollectionUtils.isEmpty(omKeyLocationInfoGroups)) {
+      return new BlockLocation[0];
+    }
+
+    OmKeyLocationInfoGroup omKeyLocationInfoGroup =
+        omKeyLocationInfoGroups.iterator().next();
+    BlockLocation[] blockLocations = new BlockLocation[
```
bharatviswa504: And also when a key is overrided, there will be multiple versions, and in the latest version, it has the blocks of (latest and before version). In those kind of scenario's what needs to be returned, right now every time we get the first version. I think we should get the last version from omKeyLocationInfoGroups.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -440,7 +450,61 @@ private FileStatusAdapter toFileStatusAdapter(OzoneFileStatus status) {
...
+            if (port == 0) {
+              port = configuredDnPort;
+            }
+            nameList.add(dn.getHostName() + ":" + port);
+          });
+
+      String[] hosts = hostList.toArray(new String[hostList.size()]);
+      String[] names = nameList.toArray(new String[nameList.size()]);
+      BlockLocation blockLocation = new BlockLocation(
+          names, hosts, 0, fileStatus.getLen());
```
bharatviswa504: Here length is set file length right now, or do we need to set block length?

bharatviswa504: From documentation looks like we need to set block size.

>    * BlockLocation(offset: 0, length: BLOCK_SIZE,
>    *   hosts: {"host1:9866", "host2:9866, host3:9866"})
>    * BlockLocation(offset: BLOCK_SIZE, length: BLOCK_SIZE,
>    *   hosts: {"host2:9866", "host3:9866, host4:9866"})

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -783,5 +797,11 @@ private FileStatus convertFileStatus(
         fileStatusAdapter.getPath()
     );
 
+    BlockLocation[] blockLocations = fileStatusAdapter.getBlockLocations();
+    if (blockLocations == null || blockLocations.length == 0) {
```
bharatviswa504: Why we need this special handling. Even in this case, we can return LocatedFileStatus with an empty right, like using default BlockLocation constructor. Or is there any reason for doing this way.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
avijayanhwx: I agree, will make the change.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
avijayanhwx: When we send LocatedFileStatus without the locations, I felt that it seemed wrong. If we cannot get block locations, we go to default behavior of base filesystem class.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
avijayanhwx: Will change it to pass in block length and offset.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
bharatviswa504: QQ: If block location is empty it is 0-byte key. So, sending as before will help anyway?

adoroszlai:
Acceptance test failure caused by:

```
2019-12-20 03:55:08 INFO  Job:1380 - Job job_1576813928823_0001 failed with state FAILED due to: Job commit failed: org.apache.hadoop.ipc.RemoteException(java.lang.NullPointerException):
java.lang.NullPointerException
        at org.apache.hadoop.ozone.om.KeyManagerImpl.refreshPipeline(KeyManagerImpl.java:690)
        at org.apache.hadoop.ozone.om.KeyManagerImpl.listStatus(KeyManagerImpl.java:2027)
        at org.apache.hadoop.ozone.om.OzoneManager.listStatus(OzoneManager.java:2878)
        at org.apache.hadoop.ozone.protocolPB.OzoneManagerRequestHandler.listStatus(OzoneManagerRequestHandler.java:595)
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -686,31 +687,35 @@ public OmKeyInfo lookupKey(OmKeyArgs args, String clientAddress)
...
-        // The production should always have containerClient initialized.
-        if (scmClient.getContainerClient() != null) {
-          try {
-            if (!containerWithPipelineMap.containsKey(k.getContainerID())) {
-              ContainerWithPipeline containerWithPipeline = scmClient
-                  .getContainerClient()
-                  .getContainerWithPipeline(k.getContainerID());
-              containerWithPipelineMap.put(k.getContainerID(),
-                  containerWithPipeline);
+    if (value != null &&
```
bharatviswa504: I just want to understand why we need not null check here. Because in DB we never store for a key null value as KeyInfo.

xiaoyuyao:
@fapifta can you take a look?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
avijayanhwx: It may be possible that the location info is null, and not the key info itself.

bharatviswa504: Figured out why it can be null, as when directory entry fileStatus will have OmKeyInfo null. So, we need a null check over there in refreshPipeline.

PR #332, opened 2019-12-09
HDDS-2539. Sonar: Fix sonar warnings in SCMContainerManager class.

swagle:
Fix sonar warnings in SCMContainerManager. The synchronized on pipeline parameter is required for correctness.

https://issues.apache.org/jira/browse/HDDS-2539

## How was this patch tested?
No logical modifications made.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
@@ -83,7 +83,6 @@ public class SCMContainerManager implements ContainerManager {
    * @param pipelineManager - PipelineManager
    * @throws IOException on Failure.
    */
-  @SuppressWarnings("unchecked")
   public SCMContainerManager(final Configuration conf,
       PipelineManager pipelineManager) throws IOException {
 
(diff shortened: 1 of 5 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -90,8 +85,7 @@
    */
   @SuppressWarnings("unchecked")
```
adoroszlai: Only if you need to update the patch anyway: this `SuppressWarnings` is unnecessary, can be removed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -79,19 +77,14 @@
    * Constructs a mapping class that creates mapping between container names
    * and pipelines.
    *
-   * @param nodeManager - NodeManager so that we can get the nodes that are
-   * healthy to place new
-   * containers.
    * passed to LevelDB and this memory is allocated in Native code space.
    * CacheSize is specified
    * in MB.
```
xiaoyuyao: missing Javadoc for conf

swagle: Addressed @xiaoyuyao, thanks.

PR #333, opened 2019-12-10
HDDS-2279. Ozone S3 CLI commands not working on HA cluster

bharatviswa504:

Make Ozone S3 commands work on OM HA cluster

https://issues.apache.org/jira/browse/HDDS-2279

## How was this patch tested?

Without this patch used to get an error **Service ID or host name must not be omitted when ozone.om.service.ids is defined.**

With this patch on non-secure OM HA cluster
**bash-4.2$ ozone s3 getsecret --om-service-id id1
This command is not supported in unsecure clusters.** 
From this message, we can confirm the client is able to create.

As we don't have any secure om ha docker-compose cluster, not added any new test in this PR. Opened a new Jira HDDS-2700 to have a secure OM HA cluster where this can be tested end to end.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/OzoneAddress.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/OzoneAddress.java
@@ -29,6 +29,8 @@ import org.apache.hadoop.ozone.client.OzoneClientFactory;
 
 import static org.apache.hadoop.ozone.OzoneConsts.OZONE_HTTP_SCHEME;
 import static org.apache.hadoop.ozone.OzoneConsts.OZONE_RPC_SCHEME;
+import static org.apache.hadoop.ozone.om.OMConfigKeys.OZONE_OM_SERVICE_IDS_KEY;
+
 import org.apache.http.client.utils.URIBuilder;
 
 /**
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/OzoneAddress.java
```diff
@@ -124,6 +126,41 @@ public OzoneClient createClient(OzoneConfiguration conf)
...
+      if (OmUtils.isOmHAServiceId(conf, omServiceID)) {
+        return OzoneClientFactory.getRpcClient(omServiceID, conf);
+      } else {
+        throw new OzoneClientException("Service ID specified does not match" +
+            " with " + OZONE_OM_SERVICE_IDS_KEY + " defined in the " +
+            "configuration. Configured " + OZONE_OM_SERVICE_IDS_KEY + " are" +
+            conf.getTrimmedStringCollection(OZONE_OM_SERVICE_IDS_KEY));
+      }
+    } else {
+      // If om service id is not specified, consider it as a non-HA cluster.
```
xiaoyuyao: Can we handle the HA case where om service ID is not provided by CLI argument but local conf has a valid service ID without the Exception?

bharatviswa504: I too thought that, but later I got a doubt that if conf has service ID, but it is of the remote cluster. (I think this can happen when conf has remote HA cluster details and the local non-HA cluster) . So, thrown exception and have marked this mandatory for HA cluster. Let me know your view on this.

bharatviswa504: And also it can have multiple serviceID's when it has conf of local HA cluster and remote HA cluster.

arp7: I think this approach makes sense.

arp7:
The changes look fine to me. @xiaoyuyao what do you think?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/OzoneAddress.java
xiaoyuyao: agree. Thanks for the context.

bharatviswa504:
Thank You @xiaoyuyao and @arp7 for the review.

PR #337, opened 2019-12-10
HDDS-2706. Add class CRLCodec - used for certificate revocation list.

abhishekaypurohit:

When revoking SCM certs, DNs local cache is not refreshed. This is the first patch for that functionality. 

Note: This is not integrated yet so it's safe. However, I am going to add Unit Tests with the next patch. 

https://issues.apache.org/jira/browse/HDDS-2706

## How was this patch tested?

mvn clean install


elek:
Thank you very much the patch @abhishekaypurohit Is there any related design doc which can help to understand the goal and review the change? I didn't find it in the Jira but I might miss it...

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/utils/TestCRLCodec.java
```diff
@@ -0,0 +1,141 @@
...
+    JcaContentSignerBuilder contentSignerBuilder =
+        new JcaContentSignerBuilder("SHA256WithRSAEncryption");
+
+    contentSignerBuilder.setProvider("BC");
+    PrivateKey privateKey = keyPair.getPrivate();
+    X509CRLHolder cRLHolder =
+        builder.build(contentSignerBuilder.build(privateKey));
+
+    CRLCodec crlCodec = new CRLCodec(securityConfig);
+    crlCodec.writeCRL(cRLHolder, "pemcertificate.crl", true);
```
anuengineer: can we please rename this file RevocationList.crl. Thanks

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/utils/TestCRLCodec.java
```diff
@@ -0,0 +1,141 @@
...
+  public void testWriteCRL() throws IOException, OperatorCreationException {
+
+    X500Name issuer = x509CertificateHolder.getIssuer();
+    Date now = new Date();
+    X509v2CRLBuilder builder = new X509v2CRLBuilder(issuer, now);
+    builder.addCRLEntry(x509CertificateHolder.getSerialNumber(), now,
+                        CRLReason.cACompromise);
+
+    JcaContentSignerBuilder contentSignerBuilder =
+        new JcaContentSignerBuilder("SHA256WithRSAEncryption");
```
anuengineer: Can we add this "SHA256WithRSAEncryption" to the security config class if it is not already there? if you are adding this new, you might want to replace this string the certificate signing class too in a different Jira.

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/utils/TestCRLCodec.java
```diff
@@ -0,0 +1,141 @@
...
+
+    contentSignerBuilder.setProvider("BC");
```
anuengineer: we can use SecurityConfig.getProvider() call here.

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/utils/TestCRLCodec.java
```diff
@@ -0,0 +1,141 @@
+package org.apache.hadoop.hdds.security.x509.certificate.utils;
```
adoroszlai: License header is needed in all files.  Example:

https://github.com/apache/hadoop-ozone/blob/050d58b6d0f59e251519a4108930fa78f850b231/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/scm/net/TestNetUtils.java#L1-L17

abhishekaypurohit: thanks @adoroszlai and thanks to @dineshchitlangia for explaining this stuff.

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/utils/TestCRLCodec.java
abhishekaypurohit: Done.

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/utils/TestCRLCodec.java
abhishekaypurohit: Done. Getting from securityConfig now.

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/utils/TestCRLCodec.java
abhishekaypurohit: Done

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsConfigKeys.java
```diff
@@ -164,10 +164,13 @@
   public static final String HDDS_X509_FILE_NAME = "hdds.x509.file.name";
   public static final String HDDS_X509_FILE_NAME_DEFAULT = "certificate.crt";
 
+  public static final String HDDS_X509_CRL_NAME = "hdds.x509.CRL.name";
```
adoroszlai: @abhishekaypurohit @anuengineer Sorry for flagging this too late: config properties should be also added to `ozone-default.xml` (so that it can be exposed to users), or as an exception in `TestOzoneConfigurationFields#addPropertiesNotInXml` (if it's for developers only).  We have an integration test to catch such config mismatches (`TestOzoneConfigurationFields`), but it's currently disabled with all other integration tests...

Filed https://issues.apache.org/jira/browse/HDDS-2873 for this new property.  Please take a look.

PR #339, opened 2019-12-11
HDDS-2553. Iterator.next() methods should throw NoSuchElementException

iamabug:

throw NoSuchElementException in Iterator.next()

https://issues.apache.org/jira/browse/HDDS-2553

## How was this patch tested?

mvn -Dtest=TestChunkBuffer test


hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBufferImplWithByteBuffer.java
```diff
@@ -53,6 +54,9 @@ public boolean hasNext() {
       @Override
       public ByteBuffer next() {
         final ByteBuffer duplicated = buffer.duplicate();
+        if (!buffer.hasRemaining()) {
```
xiaoyuyao: Change LGTM. Should we move this before Line 56? Can you add a unit test for repro and verification to avoid regression in future?

iamabug: Moving to Line 56 is ok, but could you give some advice on uni test here ?

iamabug: And is rebase required here ?

iamabug: @xiaoyuyao  Please review my unit test

adoroszlai: Rebase is not required while GitHub says:

> This branch has no conflicts with the base branch

Even if there are conflicts, a merge from `master` is preferred over rebase, to avoid the need for force push.

Regarding unit test, I would tweak the existing one, adding to the end of https://github.com/apache/hadoop-ozone/blob/050d58b6d0f59e251519a4108930fa78f850b231/hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/common/TestChunkBuffer.java#L111-L134

these:

```
    Assert.assertFalse(i.hasNext());
    Assertions.assertThrows(NoSuchElementException.class, i::next);
```

`TestChunkBuffer` tests both implementations of `ChunkBuffer` without code duplication.

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/common/TestChunkBufferImplWithByteBuffer.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.NoSuchElementException;
+
+public class TestChunkBufferImplWithByteBuffer {
+
+    @Test(expected = NoSuchElementException.class)
+    public void testNoSuchElementException() {
+        ChunkBuffer impl = ChunkBuffer.allocate(10);
+        Iterator<ByteBuffer> iterator = impl.iterate(1).iterator();
+        while (true) {
+            iterator.next();
```
adoroszlai: Wouldn't this be an infinite loop with the buggy implementation?

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/common/TestChunkBufferImplWithByteBuffer.java
```diff
@@ -0,0 +1,19 @@
+package org.apache.hadoop.ozone.common;
```
adoroszlai: License header is needed in all files.  Example:

https://github.com/apache/hadoop-ozone/blob/050d58b6d0f59e251519a4108930fa78f850b231/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/scm/net/TestNetUtils.java#L1-L17

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBufferImplWithByteBuffer.java
iamabug: Thanks for the detailed explanation. And how can I run checks on my own fork ?

adoroszlai: I think they should be run automatically when you push to your fork.  Check your repo settings at https://github.com/iamabug/hadoop-ozone/settings/actions

iamabug:
@adoroszlai @xiaoyuyao Thanks for your reviews and valuable opinion.

elek:
As this patch got two +1 (thanks the review @xiaoyuyao and @adoroszlai) I am merging it to the master.

Thanks the contribution @iamabug

PR #343, opened 2019-12-11
HDDS-2607 DeadNodeHandler should not remove replica for a dead maintenance node

sodonnel:

Normally, when a node goes dead, the DeadNodeHandler removes all the containers and replica associated with the node from the ContainerManager.

If a node is IN_MAINTENANCE and goes dead, then we do not want to remove its replica. They should remain present in the system to prevent the container being marked as under-replicated.

We also need to consider the case where the node is dead, and then maintenance expires automatically. In that case, the replica associated with the node must be removed and the affected containers will become under-replicated.

This PR resolves this issue, by ensuring stale, dead and healthy node events are fired as normal by the NodeStateManager when the node health state changes. This is works as before, prior to any decommission related changes.

To ensure the the container replicas are not removed for a node IN_MAINTENANCE, the DeadNodeHandler has been modified to check the node state and skip removing the replicas if the node is IN_MAINTENANE. For all other states, the dead node handling is unchanged.

Then, for any changes in the node operational state, the same health events are fired when the operational state is changed. This means that a DECOMMISSIONED or IN_MAINTENANCE node that is heartbeating and healthy will fire the HEALTHY node event when it is moved to IN_SERVICE, which triggers pipeline creation.

Similarly, if a node was IN_MAINTENANCE and is DEAD, when it transitions to IN_SERVICE, the dead node event will be trigger, causing the container replicas to be removed. Some of these events will have no effect, but the volume of changes to the node operational state will be minimal, as it will only be changed by the decommission monitor or someone triggering or cancelling decommission and maintenance.

https://issues.apache.org/jira/browse/HDDS-2607

## How was this patch tested?

Additional and modified unit tests

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/DeadNodeHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/DeadNodeHandler.java
@@ -74,12 +74,14 @@ public class DeadNodeHandler implements EventHandler<DatanodeDetails> {
        * To be on a safer side, we double check here and take appropriate
        * action.
        */
-
       destroyPipelines(datanodeDetails);
       closeContainers(datanodeDetails, publisher);
 
-      // Remove the container replicas associated with the dead node.
-      removeContainerReplicas(datanodeDetails);
+      // Remove the container replicas associated with the dead node unless it
+      // is IN_MAINTENANCE
+      if (!nodeManager.getNodeStatus(datanodeDetails).isInMaintenance()) {
+        removeContainerReplicas(datanodeDetails);
+      }
 
     } catch (NodeNotFoundException ex) {
       // This should not happen, we cannot get a dead node event for an
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
@@ -93,11 +93,6 @@ public class NodeStateManager implements Runnable, Closeable {
    */
   private final StateMachine<NodeState, NodeLifeCycleEvent> nodeHealthSM;
   /**
-   * StateMachine for node operational state.
-   */
-  private final StateMachine<HddsProtos.NodeOperationalState,
-      NodeOperationStateEvent> nodeOpStateSM;
-  /**
    * This is the map which maintains the current state of all datanodes.
    */
   private final NodeStateMap nodeStateMap;
@@ -112,7 +107,7 @@ public class NodeStateManager implements Runnable, Closeable {
   /**
    * Maps the event to be triggered when a node state us updated.
    */
-  private final Map<NodeStatus, Event<DatanodeDetails>> state2EventMap;
+  private final Map<NodeState, Event<DatanodeDetails>> state2EventMap;
   /**
    * ExecutorService used for scheduling heartbeat processing thread.
    */
@@ -162,10 +157,7 @@ public class NodeStateManager implements Runnable, Closeable {
     this.state2EventMap = new HashMap<>();
     initialiseState2EventMap();
     Set<NodeState> finalStates = new HashSet<>();
-    Set<HddsProtos.NodeOperationalState> opStateFinalStates = new HashSet<>();
     this.nodeHealthSM = new StateMachine<>(NodeState.HEALTHY, finalStates);
-    this.nodeOpStateSM = new StateMachine<>(
-        NodeOperationalState.IN_SERVICE, opStateFinalStates);
     initializeStateMachines();
     heartbeatCheckerIntervalMs = HddsServerUtil
         .getScmheartbeatCheckerInterval(conf);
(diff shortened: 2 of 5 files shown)


sodonnel:
This change partly tested by modifying the single test in TestDeadNodeHandler(), however due to HDDS-2508 this test is Ignored (passes locally, failing under CI). Therefore I ran the test manually a few times and it always passed and the test should be re-enabled when the mentioned issue is resolved.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestDeadNodeHandler.java
```diff
@@ -58,11 +58,7 @@
 import org.apache.hadoop.security.authentication.client
     .AuthenticationException;
 import org.apache.hadoop.test.GenericTestUtils;
-import org.junit.After;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.Ignore;
-import org.junit.Test;
+import org.junit.*;
```
swagle: Generally, explicit imports are preferred over *

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
```diff
@@ -238,36 +228,6 @@ private void initializeStateMachines() {
         NodeState.STALE, NodeState.HEALTHY, NodeLifeCycleEvent.RESTORE);
     nodeHealthSM.addTransition(
         NodeState.DEAD, NodeState.HEALTHY, NodeLifeCycleEvent.RESURRECT);
-
-    nodeOpStateSM.addTransition(
-        NodeOperationalState.IN_SERVICE, NodeOperationalState.DECOMMISSIONING,
-        NodeOperationStateEvent.START_DECOMMISSION);
```
adoroszlai: Now `enum NodeOperationStateEvent` is unused.  Can it be removed?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/DeadNodeHandler.java
```diff
@@ -74,12 +74,14 @@ public void onMessage(final DatanodeDetails datanodeDetails,
...
        */
-
       destroyPipelines(datanodeDetails);
       closeContainers(datanodeDetails, publisher);
 
-      // Remove the container replicas associated with the dead node.
-      removeContainerReplicas(datanodeDetails);
+      // Remove the container replicas associated with the dead node unless it
+      // is IN_MAINTENANCE
+      if (!nodeManager.getNodeStatus(datanodeDetails).isInMaintenance()) {
```
adoroszlai: Is it possible to get this message for a node `ENTERING_MAINTENANCE`?  If so, shouldn't it be ignored here as well?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestNodeStateManager.java
```diff
@@ -202,6 +202,58 @@ public void testNodeOpStateCanBeSet()
...
+    // First set the node to decommissioned, then run through all op states in
+    // order and ensure the non_healthy_to_healthy event gets fired
+    nsm.setNodeOperationalState(dn,
+        HddsProtos.NodeOperationalState.DECOMMISSIONED);
+    for (HddsProtos.NodeOperationalState s :
+        HddsProtos.NodeOperationalState.values()) {
+      eventPublisher.clearEvents();
+      nsm.setNodeOperationalState(dn, s);
+      assertEquals("NON_HEALTHY_TO_HEALTHY_NODE",
+          eventPublisher.getLastEvent().getName());
```
adoroszlai: I think it would be safer to compare the event, not its name:

```
      assertEquals(SCMEvents.NON_HEALTHY_TO_HEALTHY_NODE,
          eventPublisher.getLastEvent());
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/DeadNodeHandler.java
sodonnel: If a node goes dead while "entering maintenance" then it should be handled as a dead node. The reason, is that we have not yet determine if all the containers on the host are sufficiently replicated. If it goes dead before that check has completed and the node has moved into IN_MAINTENANCE, then it must be handled as a dead node as normal.

This is the same for a node which is DECOMMISSIONING and not yet reached DECOMMISSIONED.

Unless the node has reached its end state (IN_MAINTENANCE or DECOMMISSIONED), if it goes dead the workflow is aborted and it is handled just like any other dead node.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestNodeStateManager.java
sodonnel: True. I was pretty much copying existing code when I added this test. I have made this change here and also applied it to the other occurrences in the class to keep in consistent.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
sodonnel: Correct, this can be removed. I have made this change.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestDeadNodeHandler.java
sodonnel: Ah, caught out by Intellij again! I have fixed this, and also, hopefully fixed my Intellij settings to stop it doing this.

adoroszlai:
Thanks @sodonnel for the contribution and @swagle for the review.

PR #345, opened 2019-12-11
HDDS-2703. OzoneFSInputStream to support ByteBufferReadable

cxorm:
This PR makes the ```OzoneFSInputStream``` implement ```ByteBufferReadable``` and override ```int read(ByteBuffer)```
(Note: Cause the current dependency version of Hadoop in Ozone is 3.2.0, I don't implement ```ByteBufferPositionedReadable``` for now)

https://issues.apache.org/jira/browse/HDDS-2703

## How was this patch tested?
UT is updated.
Ran on my standalone cluster, and test it with python.
(Without ```UnsupportedOperationException``` like [HDDS-2443](https://issues.apache.org/jira/browse/HDDS-2443
))
Without ```UnsupportedOperationException```


hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
```diff
@@ -76,4 +79,21 @@ public boolean seekToNewSource(long targetPos) throws IOException {
...
+  /**
+   * @param buf the ByteBuffer to receive the results of the read operation.
+   * @return the number of bytes read, possibly zero.
+   * @throws IOException if there is some error performing the read
+   */
+  @Override
+  public int read(ByteBuffer buf) throws IOException {
+
+    int bufInitPos = buf.position();
+    while (buf.hasRemaining() && (inputStream.available() > 0)) {
```
bharatviswa504: Instead of reading byte by byte, can we use inputStream.read(byte b[], int off, int len)?

Something in similar terms
```
    int readLen = Math.min(buf.remaining(), inputStream.available());
    byte[] readData = new byte[readLen];

    inputStream.read(readData, bufInitPos, readLen);
    buf.put(readData);
```

hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java
```diff
@@ -134,4 +135,20 @@ public void testO3FSMultiByteRead() throws IOException {
     Assert.assertTrue(Arrays.equals(value, data));
     inputStream.close();
   }
+
+  @Test
+  public void testO3FSByteBufferRead() throws IOException {
+    FSDataInputStream inputStream = fs.open(filePath);
```
jojochuang: i would really love to see this wrapped in a try with block to ensure resource is closed properly. But I see that all test cases in the file follows the same pattern so maybe we can have a separate jira to correct that.

cxorm: Thanks @jojochuang for the suggest.
JIRA created for this issue. ([HDDS-2723](https://issues.apache.org/jira/browse/HDDS-2723))

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
```diff
@@ -76,4 +79,21 @@ public boolean seekToNewSource(long targetPos) throws IOException {
   public int available() throws IOException {
     return inputStream.available();
   }
+
+  /**
+   * @param buf the ByteBuffer to receive the results of the read operation.
+   * @return the number of bytes read, possibly zero.
```
jojochuang: According to the ByteBufferReadable javadoc, it should return -1 if end of file.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
```diff
@@ -76,4 +81,39 @@ public boolean seekToNewSource(long targetPos) throws IOException {
...
+  /**
+   * Query the stream for a specific capability.
+   *
+   * @param capability string to query the stream support for.
+   * @return True if the stream supports capability.
+   */
+  @Override
+  public boolean hasCapability(String capability) {
+    switch (StringUtils.toLowerCase(capability)) {
+    case "in:readbytebuffer":
```
cxorm: This case should be ```StreamCapabilities.READBYTEBUFFER``` that is supported in Hadoop 3.3.0 for now.
I use ```in:readbytebuffer``` here cause the dependency version of hadoop in Ozone is 3.2.0.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
cxorm: Thanks @bharatviswa504 for the review.
Fixed, and work well.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
cxorm: Thanks @jojochuang for the review.
Fixed the description, and updated the implementation.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
cxorm: Should we create a JIRA for this ?

bharatviswa504: Sorry missed this, what is this change because Jira title says to support ByteBufferReadable.

bharatviswa504: If it is something we want to support, can we address this in new Jira.

cxorm:
> Missed one change related to StreamCapabilities, If it is something we want to support, can we address this in a new Jira. As this Jira purpose is to support ByteBufferReadable

Thanks @bharatviswa504 for the comments.
I revert the ```Make OzoneFSInputStream support StreamCapabilities```, and create a Jira for this purpose. ([HDDS-2750](https://issues.apache.org/jira/browse/HDDS-2750))

bharatviswa504:
Thank You @cxorm for the contribution and @jojochuang for the review.

cxorm:
Thanks @bharatviswa504 and @jojochuang for the review.
And thanks @bharatviswa504 for the merge.

PR #349, opened 2019-12-12
HDDS-2719. Rename S3Token to S3AuthInfo

mohammadjkhan:
Renamed references of S3TOKEN to S3AUTHINFO

https://issues.apache.org/jira/browse/HDDS-2719

## How was this patch tested?
Ran project unit tests

### DIFF
--- a/hadoop-hdds/docs/content/security/SecureOzone.md
+++ b/hadoop-hdds/docs/content/security/SecureOzone.md
@@ -60,12 +60,12 @@ Block tokens allow a client to read or write a block. This is needed so that
 data nodes know that the user/client has permission to read or make
 modifications to the block.
 
-### S3Token ###
+### S3AuthInfo ###
 
 S3 uses a very different shared secret security scheme. Ozone supports the AWS Signature Version 4 protocol,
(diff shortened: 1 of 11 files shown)


xiaoyuyao:
@bharatviswa504 We prefer to use ozone tokens for hadoop delegation tokens such as OM token, block token, and SCM/DN tokens (To be done in Security phase III).
 
The current S3 token is more like a standard S3 access key/secret, which we prefer to use standard S3 terms.

hadoop-hdds/docs/content/security/SecureOzone.md
```diff
@@ -60,12 +60,12 @@ Block tokens allow a client to read or write a block. This is needed so that
 data nodes know that the user/client has permission to read or make
 modifications to the block.
 
-### S3Token ###
+### S3AuthInfo ###
 
 S3 uses a very different shared secret security scheme. Ozone supports the AWS Signature Version 4 protocol,
 and from the end users perspective Ozone's s3 feels exactly like AWS S3.
```
xiaoyuyao: s3 => S3

mohammadjkhan: Done. Updated

PR #351, opened 2019-12-13
HDDS-1989. Fix ApplyTransaction error handling in OzoneManagerStateMachine.

bharatviswa504:

Terminate OM, after a critical error in apply transaction. Right now terminate for INTERNAL_ERROR and METADATA_ERROR.

https://issues.apache.org/jira/browse/HDDS-1989

## How was this patch tested?

We need to see if any test for this failure scenario can be added. Mostly will try to come up with UT once HDDS-2682 merges in. Since the code in applyTransaction is running in an executor thread, not able to come up with a test for this failure scenario.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
@@ -190,7 +190,8 @@ public final class OzoneManagerRatisServer {
         } else {
           // Current Ratis is setting cause, this is an safer side check.
           LOG.error("StateMachine exception cause is not set");
-          omResponse.setStatus(OzoneManagerProtocolProtos.Status.INTERNAL_ERROR);
+          omResponse.setStatus(
+              OzoneManagerProtocolProtos.Status.INTERNAL_ERROR);
         }
 
         if (LOG.isDebugEnabled()) {
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
@@ -183,9 +183,16 @@ public final class OzoneManagerRatisServer {
         OMResponse.Builder omResponse = OMResponse.newBuilder();
         omResponse.setCmdType(omRequest.getCmdType());
         omResponse.setSuccess(false);
-        omResponse.setMessage(stateMachineException.getCause().getMessage());
-        omResponse.setStatus(
-            exceptionToResponseStatus(stateMachineException.getCause()));
+        if (stateMachineException.getCause() != null) {
+          omResponse.setMessage(stateMachineException.getCause().getMessage());
+          omResponse.setStatus(
+              exceptionToResponseStatus(stateMachineException.getCause()));
+        } else {
+          // Current Ratis is setting cause, this is an safer side check.
+          LOG.error("StateMachine exception cause is not set");
+          omResponse.setStatus(OzoneManagerProtocolProtos.Status.INTERNAL_ERROR);
+        }
+
         if (LOG.isDebugEnabled()) {
           LOG.debug("Error while executing ratis request. " +
               "stateMachineException: ", stateMachineException);
@@ -215,11 +222,6 @@ public final class OzoneManagerRatisServer {
    */
   private OzoneManagerProtocolProtos.Status exceptionToResponseStatus(
       Throwable cause) {
-    if (cause == null) {
-      // Current Ratis is setting cause, this is an safer side check.
-      LOG.error("StateMachine exception cause is not set");
-      return OzoneManagerProtocolProtos.Status.INTERNAL_ERROR;
-    }
     if (cause instanceof OMException) {
       return OzoneManagerProtocolProtos.Status.values()[
           ((OMException) cause).getResult().ordinal()];
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
@@ -81,6 +81,7 @@ import org.apache.ratis.server.protocol.TermIndex;
 import org.apache.ratis.thirdparty.com.google.protobuf.ByteString;
 import org.apache.ratis.util.LifeCycle;
 import org.apache.ratis.util.SizeInBytes;
+import org.apache.ratis.util.StringUtils;
 import org.apache.ratis.util.TimeDuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -192,6 +193,8 @@ public final class OzoneManagerRatisServer {
           LOG.error("StateMachine exception cause is not set");
           omResponse.setStatus(
               OzoneManagerProtocolProtos.Status.INTERNAL_ERROR);
+          omResponse.setMessage(
+              StringUtils.stringifyException(stateMachineException));
         }
 
         if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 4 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -210,23 +209,24 @@ private OMResponse processReply(OMRequest omRequest, RaftClientReply reply)
...
-    if (errorMessage.contains(STATUS_CODE)) {
-      String errorCode = errorMessage.substring(
-          errorMessage.indexOf(STATUS_CODE) + STATUS_CODE.length());
-      LOG.debug("Parsing error message for error code " +
-          errorCode);
-      return OzoneManagerProtocolProtos.Status.valueOf(errorCode.trim());
+  private OzoneManagerProtocolProtos.Status exceptionToResponseStatus(
+      Throwable cause) {
+    if (cause == null) {
+      // Current Ratis is setting cause, this is an safer side check.
```
hanishakoneru: I think we would never hit this code. In line 188 (before calling this function) we call "stateMachineException.getCause().getMessage()". If the cause was null, we would get a NullPointerException there itself.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -210,23 +209,24 @@ private OMResponse processReply(OMRequest omRequest, RaftClientReply reply)
...
+      LOG.error("StateMachine exception cause is not set");
+      return OzoneManagerProtocolProtos.Status.INTERNAL_ERROR;
+    }
+    if (cause instanceof OMException) {
+      return OzoneManagerProtocolProtos.Status.values()[
+          ((OMException) cause).getResult().ordinal()];
     } else {
+      LOG.error("Unknown error occurs", cause);
       return OzoneManagerProtocolProtos.Status.INTERNAL_ERROR;
```
hanishakoneru: Is there a reason for changing the parsing logic? If I remember correctly, this was added so that even if OMException is wrapped in a RemoteException or a ServiceException, the error code could still be parsed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -221,16 +225,63 @@ public TransactionContext startTransaction(
...
+          } else {
+            // For all other errors which are not critical, we can complete
+            // future normally.
+            ratisFuture.complete(OMRatisHelper.convertResponseToMessage(
+                omResponse));
+          }
+        } else {
+          ratisFuture.complete(OMRatisHelper.convertResponseToMessage(
+              omResponse));
+        }
```
hanishakoneru: We can combine the 2 "ratisFuture.complete()". If the terminate is not called, we can complete the ratisFuture.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -221,16 +225,63 @@ public TransactionContext startTransaction(
...
+   * @param resultCode
+   */
+  private void terminate(OMResponse omResponse,
+      OMException.ResultCodes resultCode) {
+    OMException exception = new OMException(omResponse.getMessage(),
+        resultCode);
+    String errorMessage = "OM Response has received unrecoverable " +
+        "error, terminating OM. Error Response is" + omResponse;
+    ExitUtils.terminate(1, errorMessage, exception, LOG);
+  }
```
hanishakoneru: This is a critical error message. Can we expand on the error message to give as much information as possible? I am thinking we should mention that we received an unrecoverable error from OM Ratis server and to avoid DB corruption, we are terminating the OM. 
One question is do we have to terminate here in case of single node OM ratis also?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
bharatviswa504: Previously we used to have some code which used to run in startTransaction for allocateBlock, in that when exception happens, we used to get OMException wrapped in StateMachine Exception and to make parsing easier, we changed errorMessage as STATUS_CODE="<<code>>" Now that logic is removed. So, without current modification, we shall not throw a correct exception status when the cause is OMException.

bharatviswa504: And also now this is called when we receive StateMachineException in RaftClientReply

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
bharatviswa504: > One question is do we have to terminate here in case of single node OM ratis also?
Good question, I am also not completely sure, but looks like in single node looks like we can continue. If it is really critical error like rocksdb error due to disk failure, anyway it will fail the user operations. But I am not completely sure which will be the correct way for single node OM. Thoughts?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
bharatviswa504: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
bharatviswa504: Fixed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
bharatviswa504: Addressed other comment on changing the log message.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -184,9 +183,17 @@ private OMResponse processReply(OMRequest omRequest, RaftClientReply reply)
...
-        omResponse.setStatus(parseErrorStatus(
-            stateMachineException.getCause().getMessage()));
+        if (stateMachineException.getCause() != null) {
+          omResponse.setMessage(stateMachineException.getCause().getMessage());
+          omResponse.setStatus(
+              exceptionToResponseStatus(stateMachineException.getCause()));
+        } else {
+          // Current Ratis is setting cause, this is an safer side check.
+          LOG.error("StateMachine exception cause is not set");
+          omResponse.setStatus(
```
hanishakoneru: We should set the error message also here, right? Might break something else to not have a error message.

bharatviswa504: The downstream client only checks status. As here cause is null, so skipped the setMessage.

If for sure, we want to set error message, I see 2 options.
1. 
if (stateMachineException.getMessage() != null) 
omResponse.setMessage(stateMachineException.getMessage())

2.
 omResponse.setMessage(org.apache.hadoop.util.StringUtils.stringifyException(stateMachineException);

Let me know what you prefer?

hanishakoneru: I am ok with either.

bharatviswa504: Done. Used 2nd approach.

hanishakoneru:
Thanks @bharatviswa504 .
+1

bharatviswa504:

PR #353, opened 2019-12-13
HDDS-2623. Expose SCMDatanodeProtocolServer RPC endpoint through Recon.

avijayanhwx:
Recon should expose the 'SCMDatanodeProtocol' RPC protocol server as an endpoint. This is the first step in sending DN reports to Recon. The goal is to reuse SCM's code as much as possible.

**NOTE** : This patch has intentionally left the various subsystems (Node, Pipeline, Container) of the SCM server in Recon incomplete. Those will be completed one by one.

https://issues.apache.org/jira/browse/HDDS-2623

## How was this patch tested?
Started ozone/recon docker and verified that the RPC server starts up.

```
2019-12-13 00:02:18,906 [main] INFO org.apache.hadoop.ozone.recon.scm.ReconStorageContainerManager: Recon ScmDatanodeProtocol RPC server
 is listening at localhost/127.0.0.1:9961
2019-12-13 00:02:18,906 [main] INFO org.apache.hadoop.hdds.scm.server.SCMDatanodeProtocolServer: RPC server for DataNodes is listening a
t localhost/127.0.0.1:9961
2019-12-13 00:02:18,907 [IPC Server listener on 9961] INFO org.apache.hadoop.ipc.Server: IPC Server listener on 9961: starting
2019-12-13 00:02:18,907 [IPC Server Responder] INFO org.apache.hadoop.ipc.Server: IPC Server Responder: starting
```

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMDatanodeProtocolServer.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMDatanodeProtocolServer.java
@@ -152,8 +152,7 @@ public class SCMDatanodeProtocolServer implements
                 new StorageContainerDatanodeProtocolServerSideTranslatorPB(
                     this, protocolMessageMetrics));
 
-    InetSocketAddress datanodeRpcAddr =
-        HddsServerUtil.getScmDataNodeBindAddress(conf);
+    InetSocketAddress datanodeRpcAddr = getDataNodeBindAddress(conf);
 
     datanodeRpcServer =
         startRpcServer(
@@ -406,6 +405,10 @@ public class SCMDatanodeProtocolServer implements
         .replaceAll(System.lineSeparator(), " ")
         .trim()
         .replaceAll(" +", " ");
+  }
+
+  protected InetSocketAddress getDataNodeBindAddress(OzoneConfiguration conf) {
+    return HddsServerUtil.getScmDataNodeBindAddress(conf);
   }
 
   /**
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMDatanodeProtocolServer.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMDatanodeProtocolServer.java
@@ -115,14 +115,15 @@ public class SCMDatanodeProtocolServer implements
    */
   private final RPC.Server datanodeRpcServer;
 
-  private final StorageContainerManager scm;
+  private final StorageContainerManagerInterface scm;
   private final InetSocketAddress datanodeRpcAddress;
   private final SCMDatanodeHeartbeatDispatcher heartbeatDispatcher;
   private final EventPublisher eventPublisher;
   private final ProtocolMessageMetrics protocolMessageMetrics;
 
   public SCMDatanodeProtocolServer(final OzoneConfiguration conf,
-      StorageContainerManager scm, EventPublisher eventPublisher)
+                                   StorageContainerManagerInterface scm,
+                                   EventPublisher eventPublisher)
       throws IOException {
 
     Preconditions.checkNotNull(scm, "SCM cannot be null");
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMDatanodeProtocolServer.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMDatanodeProtocolServer.java
@@ -115,14 +115,14 @@ public class SCMDatanodeProtocolServer implements
    */
   private final RPC.Server datanodeRpcServer;
 
-  private final StorageContainerManagerInterface scm;
+  private final OzoneStorageContainerManager scm;
   private final InetSocketAddress datanodeRpcAddress;
   private final SCMDatanodeHeartbeatDispatcher heartbeatDispatcher;
   private final EventPublisher eventPublisher;
   private final ProtocolMessageMetrics protocolMessageMetrics;
 
   public SCMDatanodeProtocolServer(final OzoneConfiguration conf,
-                                   StorageContainerManagerInterface scm,
+                                   OzoneStorageContainerManager scm,
                                    EventPublisher eventPublisher)
       throws IOException {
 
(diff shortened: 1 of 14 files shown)


avijayanhwx:
cc @swagle

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMDatanodeProtocolServer.java
```diff
@@ -115,14 +115,15 @@
    */
   private final RPC.Server datanodeRpcServer;
 
-  private final StorageContainerManager scm;
+  private final StorageContainerManagerInterface scm;
```
swagle: It's uncanny to call something an interface, why not use IDE to refactor the StorageContainerManager into and interface? It would not be too big a change for callers.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
```diff
@@ -77,10 +78,9 @@ protected void configureServlets() {
       LOG.info("Starting Recon server");
       httpServer.start();
 
-      //Start Ozone Manager Service that pulls data from OM.
-      OzoneManagerServiceProvider ozoneManagerServiceProvider = injector
-          .getInstance(OzoneManagerServiceProvider.class);
-      ozoneManagerServiceProvider.start();
+      getOzoneManagerServiceProvider().start();
```
swagle: Service providers generally should not have a life cycle method. Let's start a service provided by the OzoneManagerServiceProvider, or refactor to an inner class of it.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
```diff
@@ -103,4 +103,13 @@ void stop() throws Exception {
         .getInstance(OzoneManagerServiceProvider.class);
     ozoneManagerServiceProvider.stop();
   }
+
+  OzoneManagerServiceProvider getOzoneManagerServiceProvider() {
```
swagle: If using DI properly this should never be needed, unless used in non-DI code path

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
```diff
@@ -103,4 +103,13 @@ void stop() throws Exception {
...
+    return injector.getInstance(OzoneManagerServiceProvider.class);
+  }
+
+  ReconStorageContainerManager getReconStorageContainerManager() {
```
swagle: Same as above.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconStorageContainerManager.java
```diff
@@ -0,0 +1,114 @@
...
+   *  needed.
+   * @param configuration configuration object.
+   * @return same configuration object with possible added elements.
+   */
+  private OzoneConfiguration getReconScmConfiguration(
+      OzoneConfiguration configuration) {
+    Map<String, String> reconScmConfigs =
+        configuration.getPropsWithPrefix(RECON_SCM_CONFIG_PREFIX);
+    for (Map.Entry<String, String> entry : reconScmConfigs.entrySet()) {
+      configuration.set(entry.getKey(), entry.getValue());
```
swagle: 1. Setting it in the original config object is it safe? Maybe when running inside same JVM with SCM or some corner case, why not return a clone?
2. Shouldn't we remove the prefix and set?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMDatanodeProtocolServer.java
avijayanhwx: I can change the name of the interface. I just did not want to move everything to an interface now, since we are not sure what all we will need to override. At the end of this, we can expect the interface to have everything Recon needs to override.

swagle: But the point is after your patch, anybody changing SCM will change the interface and thereby Recon will get the change too. It's a good thing that way. From Recon point of view, empty method behavior is also ok until we use those methods.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
avijayanhwx: The class should be renamed to something like "ReconOzoneManager" since its role is more akin to a manager than a service provider. I can do that in another patch since it will touch all files.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
avijayanhwx: Yes, I agree. I had created them for readability. I can move it to private.  We don't expect external consumers for these other than tests maybe.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconStorageContainerManager.java
avijayanhwx: 1. I did see some other examples of setting subsets of configuration in HDDS. I also wanted to create a new configuration object from properties, but the Hadoop configuration does not seem to provide a straightforward way to do that. I will look again.
2. configuration.getPropsWithPrefix() removes the prefix.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMDatanodeProtocolServer.java
avijayanhwx: Done.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
avijayanhwx: Done.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconStorageContainerManager.java
avijayanhwx: Created new Configuration instance and modified it instead of the global config object.

swagle:
+1 LGTM

PR #354, opened 2019-12-13
HDDS-2724. Sonar : Fix sonar issues in package 'org.apache.hadoop.ozo…

avijayanhwx:
…ne.recon.tasks'.

Fix sonar issues in package 'org.apache.hadoop.ozone.recon.tasks'.

https://issues.apache.org/jira/browse/HDDS-2724

## How was this patch tested?
mvn clean test

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/FileSizeCountTask.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/FileSizeCountTask.java
@@ -162,10 +162,10 @@ public class FileSizeCountTask implements ReconDBUpdateTask {
           updateUpperBoundCount(omKeyInfo, DELETE);
           break;
 
-        default: LOG.trace("Skipping DB update event : " + omdbUpdateEvent
-                  .getAction());
+        default: LOG.trace("Skipping DB update event : {}",
+            omdbUpdateEvent.getAction());
         }
-      } catch (IOException e) {
+      } catch (Exception e) {
         LOG.error("Unexpected exception while updating key data : {} {}",
                 updatedKey, e.getMessage());
         return new ImmutablePair<>(getTaskName(), false);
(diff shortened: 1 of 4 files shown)


hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/OMDBUpdateEvent.java
```diff
@@ -107,7 +107,7 @@ OMUpdateEventBuilder setSequenceNumber(long sequenceNumber) {
      * @return OMDBUpdateEvent
      */
     public OMDBUpdateEvent build() {
-      return new OMDBUpdateEvent<KEY, VALUE>(
+      return new OMDBUpdateEvent<K1, V1>(
```
bharatviswa504: Can we leave the Generic Types to actual KEY, VALUE instead of K1, V1. I feel it is more readable than single letters.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/FileSizeCountTask.java
```diff
@@ -162,10 +162,10 @@ private void updateCountFromDB() {
           updateUpperBoundCount(omKeyInfo, DELETE);
           break;
 
-        default: LOG.trace("Skipping DB update event : " + omdbUpdateEvent
-                  .getAction());
+        default: LOG.trace("Skipping DB update event : {}",
+            omdbUpdateEvent.getAction());
         }
-      } catch (IOException e) {
+      } catch (Exception e) {
```
bharatviswa504: Why Exception is changed here, any reason?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/OMDBUpdateEvent.java
avijayanhwx: Thanks for the review, done.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/FileSizeCountTask.java
avijayanhwx: The exception type was wrong. IOException was not being thrown. And in this method, we have to capture all Exceptions.

bharatviswa504:
Thank You @avijayanhwx for the contribution.

PR #355, opened 2019-12-13
HDDS-2728. Remove methods of internal representation from DatanodeAdminMontor interface 

elek:
* `DatanodeAdminMonitorInterface` has some settesr and getters but they are not part of the contract they are used by the implementation. An other implementation of the interface may require different setters. Therefore the interface should contain only the required fields.
 *  `DatanodeAdminMonitorInterface` can be renamed to `DatanodeAdminMonitor` to follow the naming convention in the code (*Impl)
 * `PipelineManager` is unused field can be removed from `DatanodeAdminMonitorImpl`

https://issues.apache.org/jira/browse/HDDS-2728

## How was this patch tested?

With the existing Unit tests.


elek:
@sodonnel please review.

sodonnel:
This change LGTM and is what we discussed previously. I assume no logic has changed in DatanodeAdminMonitorImpl aside from the constructor, dropped the getters and removing the unused field? As the file was renamed the diff just shows it as all new code.

PR #356, opened 2019-12-13
HDDS-2590 Integration tests for Recon with Ozone Manager

shwetayakkali:
Integration tests for Recon with Ozone Manager

https://issues.apache.org/jira/browse/HDDS-2590

## How was this patch tested?
Added Integration tests for adding Recon to MiniOzoneCluster. Tested via TestMiniOzoneCluster unit test, manual testing via API.


hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/OzoneManagerServiceProviderImpl.java
```diff
@@ -205,12 +205,12 @@ public void start() {
...
     }
     long initialDelay = configuration.getTimeDuration(
-        RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY,
-        RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY_DEFAULT,
+        configuration.get(RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY),
+        configuration.get(RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY_DEFAULT),
         TimeUnit.MILLISECONDS);
     long interval = configuration.getTimeDuration(
-        RECON_OM_SNAPSHOT_TASK_INTERVAL,
-        RECON_OM_SNAPSHOT_TASK_INTERVAL_DEFAULT,
```
avijayanhwx: Given that we are setting the Recon OM snapshot interval value in MiniOzoneClusterImpl, this should just be conf.get(RECON_OM_SNAPSHOT_TASK_INTERVAL). We should not fall back to default.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneClusterImpl.java
```diff
@@ -685,5 +756,23 @@ private void configureTrace() {
...
+          tempNewFolder.getAbsolutePath());
+
+      File reconOmDbDir = tempFolder.newFolder();
+      conf.set(OZONE_RECON_OM_SNAPSHOT_DB_DIR, reconOmDbDir
+          .getAbsolutePath());
+
+      conf.set(OZONE_RECON_SQL_DB_JDBC_URL, "jdbc:sqlite:"+
+          tempNewFolder.getAbsolutePath()+"/ozone_recon_sqlite.db");
+
+      conf.set(RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY_DEFAULT, "2s");
```
avijayanhwx: Why are we setting a config value with the "DEFAULT" value constant as key? We should set RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY and RECON_OM_SNAPSHOT_TASK_INTERVAL. Those are the config keys.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestRecon.java
```diff
@@ -0,0 +1,385 @@
...
+    LinkedTreeMap linkedTreeMap = (LinkedTreeMap) map.get("data");
+    long reconMetadataKeyCount = (long)(double)linkedTreeMap.get("totalCount");
+
+    //verify count of keys after full snapshot
+    Assert.assertEquals(omMetadataKeyCount, reconMetadataKeyCount);
+
+    //verify if Recon Metadata captures vol0/bucket0/key0 info in container0
+    ArrayList containers = (ArrayList) linkedTreeMap.get("containers");
+    LinkedTreeMap nestedContainerMap = (LinkedTreeMap)containers.get(0);
+    Assert.assertEquals(1.0, nestedContainerMap.get("NumberOfKeys"));
```
avijayanhwx: Can we also verify containerId if 0?

avijayanhwx: If we don't verify containerId being what we expect it to be, we are not testing the public ContainerId -> Key mapping functionality in Recon.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestRecon.java
```diff
@@ -0,0 +1,385 @@
...
+      throw new IOException("Unexpected exception when trying to reach Ozone " +
+          "Manager, " + EntityUtils.toString(entity));
+    } else {
+      throw new IOException("Unexpected null in http payload," +
+          " while processing request");
+    }
+  }
+
+  @Test
+  public void testReconServer() throws Exception {
```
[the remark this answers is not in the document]
shwetayakkali: For testing Recon restart, I would have to initialize keys and check for reprocess which will be done as part of testing reprocess and delta updates for Recon, so to avoid re-testing the same, I opted for consolidating the whole scenario together.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestRecon.java
```diff
@@ -0,0 +1,395 @@
...
+    Thread.sleep(pauseInterval);
+
+    // HTTP call to /api/containers
+    String containerResponse = makeHttpCall(containerKeyServiceURL);
+    Map map = new Gson().fromJson(containerResponse, HashMap.class);
+    LinkedTreeMap linkedTreeMap = (LinkedTreeMap) map.get("data");
+    long reconMetadataKeyCount = (long)(double)linkedTreeMap.get("totalCount");
+
+    //verify count of keys after full snapshot
+    Assert.assertEquals(omMetadataKeyCount, reconMetadataKeyCount);
```
avijayanhwx: Here, we are actually comparing number of keys in OM and number of containers in Recon containers API endpoint. In this case it happens to be true because we associate each key with its own container. Is there a way we can compare keys to keys?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestRecon.java
```diff
@@ -0,0 +1,395 @@
...
+    //verify sequence number after full snapshot
+    Assert.assertEquals(omLatestSeqNumber, reconLatestSeqNumber);
+
+    //add 4 keys to check for delta updates
+    addKeys(1, 5);
+    omKeyValueTableIterator = metadataManager.getKeyTable().iterator();
+    omMetadataKeyCount = getTableKeyCount(omKeyValueTableIterator);
+
+    //verify OM Metadata has all 5 keys
+    Assert.assertEquals(5, omMetadataKeyCount);
```
avijayanhwx: We don't need to assert OM behavior here.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestRecon.java
```diff
@@ -0,0 +1,395 @@
...
+    omLatestSeqNumber = ((RDBStore) metadataManager.getStore())
+        .getDb().getLatestSequenceNumber();
+
+    taskStatusResponse = makeHttpCall(taskStatusURL);
+    reconLatestSeqNumber = getReconMetadataKeyCount(taskStatusResponse, 2);
+
+    //verify sequence number after Delta Updates
+    Assert.assertEquals(omLatestSeqNumber, reconLatestSeqNumber);
+
+    File reconDbDir =
```
avijayanhwx: We can use task status API to do this check instead.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestRecon.java
```diff
@@ -0,0 +1,395 @@
...
+    //restart Recon
+    cluster.restartReconServer();
+
+    //add 5 more keys to OM
+    addKeys(5, 10);
+    omKeyValueTableIterator = metadataManager.getKeyTable().iterator();
+    omMetadataKeyCount = getTableKeyCount(omKeyValueTableIterator);
+
+    //verify 10 keys in OM
+    Assert.assertEquals(10, omMetadataKeyCount);
```
avijayanhwx: We don't need to assert OM behavior here.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestRecon.java
```diff
@@ -0,0 +1,395 @@
...
+
+    // HTTP call to /api/task/status
+    omLatestSeqNumber = ((RDBStore) metadataManager.getStore())
+        .getDb().getLatestSequenceNumber();
+
+    taskStatusResponse = makeHttpCall(taskStatusURL);
+    reconLatestSeqNumber = getReconMetadataKeyCount(taskStatusResponse, 2);
+
+    //verify sequence number after Delta Updates
+    Assert.assertEquals(omLatestSeqNumber, reconLatestSeqNumber);
```
avijayanhwx: We do the same set of checks every time -> Key count, specific key presence, sequence number. Can we move the steps into a helper method and call them from the test?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestRecon.java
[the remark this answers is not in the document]
avijayanhwx: Ok, we can keep it as 1 test then. But, I have added a suggestion in my review that will decrease the number of lines in this test.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneCluster.java
```diff
@@ -127,6 +128,14 @@ static Builder newHABuilder(OzoneConfiguration conf) {
    */
   List<HddsDatanodeService> getHddsDatanodes();
 
+  /**
+   * Returns the list of {@link ReconServer} which are part of this
```
vivekratnavel: It returns a single instance of ReconServer and not a list.

avijayanhwx:
LGTM +1, we can refine this later.

PR #357, opened 2019-12-13
HDDS-2729. Enable multilingual Hugo features in ozone docs

elek:

We need to reconfigure hugo to support multilingual site (to support other languages)

https://gohugo.io/content-management/multilingual/

I added three type of fake translations to test the templates of the Hugo.
 
 * translation of the main page
 * translation of a section (concept)
 * translation of a page (concept / datanode) 

The fake translation pages can be removed later or replaced during the translation process.

The name "Mandarin" can be replaced with the proper Chinese characters in the follow-up patches (`config.yaml` in the `hadoop-hdds/docs` folder)

https://issues.apache.org/jira/browse/HDDS-2729

## How was this patch tested?

```
cd hadoop-hdds/docs
hugo serve
firefox localhost:1313
```

You should see a "mandarin" text on the top right corner which points to the translated page

### DIFF
--- a/hadoop-hdds/docs/config.yaml
+++ b/hadoop-hdds/docs/config.yaml
@@ -21,7 +21,7 @@ languages:
     languageName: English
     weight: 1
   zh:
-    languageName: Mandarin
+    languageName: 中文
     weight: 2
 title: "Ozone"
 theme: "ozonedoc"
--- a/hadoop-hdds/docs/config.yaml
+++ b/hadoop-hdds/docs/config.yaml
@@ -16,6 +16,13 @@
 
 languageCode: "en-us"
 DefaultContentLanguage: "en"
+languages:
+  en:
+    languageName: English
+    weight: 1
+  zh:
+    languageName: Mandarin
+    weight: 2
 title: "Ozone"
 theme: "ozonedoc"
 pygmentsCodeFences: true
--- a/hadoop-hdds/docs/content/concept/_index.zh.md
+++ b/hadoop-hdds/docs/content/concept/_index.zh.md
@@ -1,3 +1,11 @@
+---
+title: Concepts
+date: "2017-10-10"
+menu: main
+weight: 6
+
+---
+
 <!---
   Licensed to the Apache Software Foundation (ASF) under one or more
   contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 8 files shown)


hadoop-hdds/docs/config.yaml
```diff
@@ -16,11 +16,18 @@
 
 languageCode: "en-us"
 DefaultContentLanguage: "en"
+languages:
+  en:
+    languageName: English
+    weight: 1
+  zh:
+    languageName: Mandarin
```
iamabug: how about using a Chinese word for language name here so that Chinese users can recognize it easilier, i.e., Mandarin -> 中文

hadoop-hdds/docs/themes/ozonedoc/layouts/partials/languages.html
```diff
@@ -0,0 +1,24 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<div class="pull-right">
+    {{ range .AllTranslations }}
+    {{ if ne $.Language .Language}}
```
iamabug: Thanks for the work. @elek

hadoop-hdds/docs/content/concept/_index.zh.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Concepts
```
iamabug: I have replaced the content with Chinese on my laptop and it seems to work out just fine. Except that the language switch button is not exactly on the top right, I am not sure if it's a problem. Screenshot:

on the other hand, index page seems to be okay on this subject:

hadoop-hdds/docs/config.yaml
elek: Yes, definitely. Just didn't know how can I write it. But I copy pasted this to replace Mandarin (thanks a lot).

cxorm:
Thanks @elek for the patch, LGTM.

hadoop-hdds/docs/content/concept/_index.zh.md
xiaoyuyao: I think it is on the right top of the content frame.

xiaoyuyao:
LGTM, +1. I will merge it shortly.

PR #366, opened 2019-12-16
HDDS-2753. Test OM Volume and Bucket operations are idempotent.

hanishakoneru:

On OM restarts, it is possible that already applied ratis logs will be replayed. Hence, we need to ensure that all OM write operations are idempotent.
This Jira aims to add unit tests for testing that OM volume and bucket related operations are idempotent.

https://issues.apache.org/jira/browse/HDDS-2753

## How was this patch tested?

This Jira is for adding a unit test.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMIdempotence.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMIdempotence.java
@@ -1,0 +1,372 @@
+package org.apache.hadoop.ozone.om;
+
+import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


cxorm:
Thanks @hanishakoneru for the PR.
Some violations detected by ```checkstyle```.

bshashikant:
Thanks @hanishakoneru for working on this. Instead of exposing a public API to force set last AppliedTermIndex on stateMachine, we can just do some operations, trigger a ratis snapshhot , do some more operations and restart. It will ensure replay will happen from lastSnapshotIndex?

Is there any specific need to add an api to force set lastAppliedIndex?

elek:
/pending Waiting for the answer of the author

PR #367, opened 2019-12-16
[WIP] HDDS-2665. Implement new Ozone Filesystem scheme ofs://

smengcl:

Implement a new scheme for Ozone Filesystem where all volumes (and buckets) can be access from a single root.

https://issues.apache.org/jira/browse/HDDS-2665

## How was this patch tested?

This is work-in-progress. Hasn't been completely tested yet. Still writing new test cases.

## Notes
1. So far the new FS class is named as `OFileSystem` (will later change into `OzoneRootedFileSystem` or something else);
2. A bunch of TODOs are noted in commit messages and within code.

### DIFF
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
@@ -215,7 +215,7 @@ public class BasicOFileSystem extends FileSystem {
   }
 
   protected void incrementCounter(Statistic statistic) {
-    //don't do anyting in this default implementation.
+    //don't do anything in this default implementation.
   }
 
   @Override
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
@@ -49,8 +49,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 import java.util.StringTokenizer;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static org.apache.hadoop.fs.ozone.Constants.LISTING_PAGE_SIZE;
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
@@ -87,9 +87,10 @@ public class BasicOFileSystem extends FileSystem {
   private String adapterPath;
 
   private static final String URI_EXCEPTION_TEXT =
-      "Ozone file system URL should be one of the following formats: " +
-      "ofs://om-host.example.com/volume/bucket/[key]  OR " +
-      "ofs://om-host.example.com:5678/volume/bucket/[key]";
+      "URL should be one of the following formats: " +
+      "ofs://om-service-id/  OR " +
+      "ofs://om-host.example.com/  OR " +
+      "ofs://om-host.example.com:5678/";
 
   @Override
   public void initialize(URI name, Configuration conf) throws IOException {
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
@@ -159,11 +159,25 @@ public class BasicOFileSystem extends FileSystem {
     }
   }
 
-  protected OzoneClientAdapter createAdapter(OFSPath ofsPath)
+  /**
+   * Check path and create client adapter accordingly.
+   * @param ofsPath
+   * @throws IOException
+   */
+  protected void checkAndCreateAdapter(OFSPath ofsPath)
       throws IOException {
-
-    adapterPath = ofsPath.getNonKeyParts();
-    return createAdapter(this.gConf, ofsPath.getBucketName(),
+    // Check if an adapter is already initialized.
+    if (this.adapter != null) {
+      // Sanity check.
+      assert(this.adapterPath != null);
+      // Close the existing adapter.
+      // TODO: do so only when volume/bucket changed from previous op for perf.
+      this.adapter.close();
+      this.adapter = null;
+      this.adapterPath = null;
+    }
+    this.adapterPath = ofsPath.getNonKeyParts();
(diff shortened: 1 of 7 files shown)


hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
```diff
@@ -0,0 +1,934 @@
...
+  private int omPort;
+  private Configuration conf;
+  private boolean isolatedClassloader;
+
+  private OzoneClientAdapter adapter;
+  private String adapterPath;
+
+  private static final String URI_EXCEPTION_TEXT =
+      "Ozone file system URL should be one of the following formats: " +
+      "ofs://om-host.example.com/volume/bucket/[key]  OR " +
```
bharatviswa504: For this kind of file system, URI should be only ofs://om-host.example.com/ or do we want to support both in the same filesystem like ofs://om-host.example.com and ofs://om-host.example.com/volume/bucket/

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
```diff
@@ -0,0 +1,934 @@
...
+
+    this.conf = conf;
+    omHost = null;
+    omPort = -1;
+    // Parse hostname and port
+    String[] parts = authority.split(":");
+    if (parts.length > 2) {
+      throw new IllegalArgumentException(URI_EXCEPTION_TEXT);
+    }
+    omHost = parts[0];
```
bharatviswa504: Do we need to have HA check here, as for HA port will not be specified?

bharatviswa504:
Thank You @smengcl for the PR.

I tried mkdir ozone fs -mkdir ofs://om:9862/vol10/buck10 

mkdir: Volume vol10 is not found

I am seeing the above message, I see the reason for this, in OFS we have parsed path and used BasicOzoneClientAdapterImpl, but in our case we should try to create volume and bucket and if it has more keys we should create a directory inside that volume/bucket. As we are using the old implementation of adapter I see we are getting this error.

And also one more thing instead of targetting all the API's lets get basic implementation of adapter and mkdir in first jira, and then we can open further jira's to fix remaning API's, as for each API we need to handle similarly.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
smengcl: This text is not finalized. It supports both URIs in the same FileSystem instance.

smengcl:
> Thank You @smengcl for the PR.
> 
> I tried mkdir ozone fs -mkdir ofs://om:9862/vol10/buck10
> 
> mkdir: Volume vol10 is not found
> 
> I am seeing the above message, I see the reason for this, in OFS we have parsed path and used BasicOzoneClientAdapterImpl, but in our case we should try to create volume and bucket and if it has more keys we should create a directory inside that volume/bucket. As we are using the old implementation of adapter I see we are getting this error.
> 
> And also one more thing instead of targetting all the API's lets get basic implementation of adapter and mkdir in first jira, and then we can open further jira's to fix remaning API's, as for each API we need to handle similarly.

mkdir is expected to not work at the moment. The conversion from mkdir -> volume/bucket creation is not implemented, yet.

Yeah for mkdir to implicitly create volume and bucket we definitely need to rewrite the adapter. But for path inside buckets I believe we can safely use the old adapter?

bharatviswa504:
> > Thank You @smengcl for the PR.
> > I tried mkdir ozone fs -mkdir ofs://om:9862/vol10/buck10
> > mkdir: Volume vol10 is not found
> > I am seeing the above message, I see the reason for this, in OFS we have parsed path and used BasicOzoneClientAdapterImpl, but in our case we should try to create volume and bucket and if it has more keys we should create a directory inside that volume/bucket. As we are using the old implementation of adapter I see we are getting this error.
> > And also one more thing instead of targetting all the API's lets get basic implementation of adapter and mkdir in first jira, and then we can open further jira's to fix remaning API's, as for each API we need to handle similarly.
> 
> mkdir is expected to not work at the moment. The conversion from mkdir -> volume/bucket creation is not implemented, yet.
> 
> Yeah for mkdir to implicitly create volume and bucket we definitely need to rewrite the adapter. But for path inside buckets I believe we can safely use the old adapter?

Yes, if provided bucket and volume exists, it will basically boil down to old implementation. This change is needed for most of API's I believe like open,mkdir and list. So, in my view to target this feature.
1. Lets do basic implementation of adapter which can do what is requiref for new file system.
2. Fix mkdir to use the new implementation and sufficient tests to cover all scenarios.
3. As once basic framework we have for further set of API's it will be easy to implement.

Any thoughts?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOFileSystem.java
bharatviswa504: Not understood what do you mean by both, as because having fs as ofs://om:host, then 2nd one should work by default, as now with this we are working on the root file system. Not sure if there is any reason why ofs need to support mounting at bucket/volume as for this old file system is already there. Not sure if I am missing something here.

smengcl:
> > > Thank You @smengcl for the PR.
> > > I tried mkdir ozone fs -mkdir ofs://om:9862/vol10/buck10
> > > mkdir: Volume vol10 is not found
> > > I am seeing the above message, I see the reason for this, in OFS we have parsed path and used BasicOzoneClientAdapterImpl, but in our case we should try to create volume and bucket and if it has more keys we should create a directory inside that volume/bucket. As we are using the old implementation of adapter I see we are getting this error.
> > > And also one more thing instead of targetting all the API's lets get basic implementation of adapter and mkdir in first jira, and then we can open further jira's to fix remaning API's, as for each API we need to handle similarly.
> > 
> > 
> > mkdir is expected to not work at the moment. The conversion from mkdir -> volume/bucket creation is not implemented, yet.
> > Yeah for mkdir to implicitly create volume and bucket we definitely need to rewrite the adapter. But for path inside buckets I believe we can safely use the old adapter?
> 
> Yes, if provided bucket and volume exists, it will basically boil down to old implementation. This change is needed for most of API's I believe like open,mkdir and list. So, in my view to target this feature.
> 
> 1. Lets do basic implementation of adapter which can do what is required for new file system like parsing path, create/get volume/bucket if needed, and use them in further operations.
> 2. Fix mkdir to use the new implementation and sufficient tests to cover all scenarios.
> 3. As once basic framework we have for further set of API's it will be easy to implement.
> 4. (Mainly it also helps reviewers)
> 
> Any thoughts?

Sounds good on the incremental change.

Also I'm going to rebase my commits to make reviewing a bit easier.

smengcl:
Closing this PR as work will be done in subtasks' PRs.

PR #369, opened 2019-12-17
HDDS-2755. Compare transactionID and updateID of Volume operations to avoid replaying transactions

hanishakoneru:

To ensure that volume operations are idempotent, we should first compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

After this patch, the replay behavior for volume operations will be as follows:
Lets say we have the following transactions:
Tx 10 - CreateVolume (Vol1, objectID = 10, updateID = 10, owner = o1, quota = 100)
Tx 12 - SetOwner (Vol1, objectID = 10, updateID = 12, owner = o2, quota = 100)
Tx 20 - SetQuota (Vol1, objectID = 10, updateID = 20, owner = o2, quota = 200)
Tx 28 - DeleteVolume (Vol1)
Tx 40 - CreateVolume (Vol1, objectID = 40, updateID = 40, owner = o1, quota = 100)
Tx 50 - SetOwner (Vol1, objectID = 40, updateID = 50, owner = o3, quota = 100)

On replaying these transactions:
Replay Tx 10 - Vol1 already exists => Tx will throw VOLUME_ALREADY_EXISTS when trying to createVolume
Replay Tx 12 - UpdateID (50) is more than equal to TxID (12) => Tx will not update DB (return ReplayResponse in validateAndUpdateCache)
Replay Tx 20 - UpdateID (50) is more than equal to TxID (20) => Tx will not update DB (return ReplayResponse in validateAndUpdateCache)
Replay Tx 28 - UpdateID (50) is more than equal to TxID (28) => Tx will not update DB (return ReplayResponse in validateAndUpdateCache)
Replay Tx 40 - Vol1 already exists => Tx will throw VOLUME_ALREADY_EXISTS when trying to createVolume
Replay Tx 50 - UpdateID (50) is more than equal to TxID (28) => Tx will not update DB (return ReplayResponse in validateAndUpdateCache)

So none of the transactions will effect any change in the DB.

https://issues.apache.org/jira/browse/HDDS-2755

## How was this patch tested?

Will add unit test in next commit

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
@@ -132,11 +132,6 @@ public class OMVolumeCreateRequest extends OMVolumeRequest {
         }
       }
 
-      // A replay transaction for CreateVolume can reach here only if the
-      // volume has been deleted in later transactions. Hence, we can
-      // continue with this request irrespective of whether it is a replay or
-      // not.
-
       // acquire lock.
       acquiredVolumeLock = omMetadataManager.getLock().acquireWriteLock(
           VOLUME_LOCK, volume);
@@ -160,12 +155,13 @@ public class OMVolumeCreateRequest extends OMVolumeRequest {
 
         omResponse.setCreateVolumeResponse(CreateVolumeResponse.newBuilder()
             .build());
-        omClientResponse = new OMVolumeCreateResponse(omVolumeArgs, volumeList,
-            omResponse.build());
+        omClientResponse = new OMVolumeCreateResponse(omResponse.build(),
+            omVolumeArgs, volumeList);
         LOG.debug("volume:{} successfully created", omVolumeArgs.getVolume());
       } else {
         // Check if this transaction is a replay of ratis logs.
-        if (isReplay(dbVolumeArgs.getUpdateID(), transactionLogIndex)) {
+        if (isReplay(ozoneManager, dbVolumeArgs.getUpdateID(),
+            transactionLogIndex)) {
           // Replay implies the response has already been returned to
           // the client. So take no further action and return a dummy
           // OMClientResponse.
@@ -181,7 +177,7 @@ public class OMVolumeCreateRequest extends OMVolumeRequest {
 
     } catch (IOException ex) {
       exception = ex;
-      omClientResponse = new OMVolumeCreateResponse(null, null,
+      omClientResponse = new OMVolumeCreateResponse(
           createErrorOMResponse(omResponse, exception));
     } finally {
       if (omClientResponse != null) {
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
@@ -132,7 +132,10 @@ public class OMVolumeCreateRequest extends OMVolumeRequest {
         }
       }
 
-      UserVolumeInfo volumeList = null;
+      // A replay transaction for CreateVolume can reach here only if the
+      // volume has been deleted in later transactions. Hence, we can
+      // continue with this request irrespective of whether it is a replay or
+      // not.
 
       // acquire lock.
       acquiredVolumeLock = omMetadataManager.getLock().acquireWriteLock(
@@ -146,6 +149,7 @@ public class OMVolumeCreateRequest extends OMVolumeRequest {
       OmVolumeArgs dbVolumeArgs =
           omMetadataManager.getVolumeTable().get(dbVolumeKey);
 
+      UserVolumeInfo volumeList = null;
       if (dbVolumeArgs == null) {
         String dbUserKey = omMetadataManager.getUserKey(owner);
         volumeList = omMetadataManager.getUserTable().get(dbUserKey);
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
@@ -164,9 +164,19 @@ public class OMVolumeCreateRequest extends OMVolumeRequest {
             omResponse.build());
         LOG.debug("volume:{} successfully created", omVolumeArgs.getVolume());
       } else {
-        LOG.debug("volume:{} already exists", omVolumeArgs.getVolume());
-        throw new OMException("Volume already exists",
-            OMException.ResultCodes.VOLUME_ALREADY_EXISTS);
+        // Check if this transaction is a replay of ratis logs.
+        if (isReplay(dbVolumeArgs.getUpdateID(), transactionLogIndex)) {
+          // Replay implies the response has already been returned to
+          // the client. So take no further action and return a dummy
+          // OMClientResponse.
+          LOG.debug("Replayed Transaction {} ignored. Request: {}",
+              transactionLogIndex, createVolumeRequest);
+          return new OMVolumeCreateResponse(createReplayOMResponse(omResponse));
+        } else {
+          LOG.debug("volume:{} already exists", omVolumeArgs.getVolume());
+          throw new OMException("Volume already exists",
+              OMException.ResultCodes.VOLUME_ALREADY_EXISTS);
+        }
       }
 
     } catch (IOException ex) {
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/OMClientResponse.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/OMClientResponse.java
@@ -42,6 +42,14 @@ public abstract class OMClientResponse {
   }
 
   /**
+   * For error or replay cases, check that the status of omResponse is not OK.
+   */
+  public void checkStatusNotOK() {
+    Preconditions.checkArgument(!omResponse.getStatus().equals(
+        OzoneManagerProtocolProtos.Status.OK));
+  }
+
+  /**
    * Check if omResponse status is OK. If yes, add to DB.
    * For OmResponse with failure, this should do nothing. This method is not
    * called in failure scenario in OM code.
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/OMClientResponse.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/OMClientResponse.java
@@ -23,6 +23,7 @@ import java.util.concurrent.CompletableFuture;
 
 import com.google.common.base.Preconditions;
 import org.apache.hadoop.ozone.om.OMMetadataManager;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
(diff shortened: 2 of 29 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/volume/OMVolumeSetQuotaResponse.java
```diff
@@ -36,22 +36,22 @@
...
       BatchOperation batchOperation) throws IOException {
 
-    // For OmResponse with failure, this should do nothing. This method is
-    // not called in failure scenario in OM code.
-    if (getOMResponse().getStatus() == OzoneManagerProtocolProtos.Status.OK) {
-      omMetadataManager.getVolumeTable().putWithBatch(batchOperation,
-          omMetadataManager.getVolumeKey(omVolumeArgs.getVolume()),
-          omVolumeArgs);
-    }
+    omMetadataManager.getVolumeTable().putWithBatch(batchOperation,
```
bharatviswa504: We should not read from DB/cache without lock, so that is the reason for having OmVolumeArgs as an argument to the constructor. Same applies to other places also

hanishakoneru: We need to set omVolumeArgs when the status is OK. Otherwise, DB update is not done.
The 2nd constructor is for the error/ replay cases. I will add a check that the status is not OK in this case to make sure addToDBBatch is not called.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
```diff
@@ -132,7 +132,10 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
         }
       }
 
-      UserVolumeInfo volumeList = null;
+      // A replay transaction for CreateVolume can reach here only if the
+      // volume has been deleted in later transactions. Hence, we can
```
bharatviswa504: We can come here without delete request also right. Example CreateVolume T1,  but lastAppliedIndex is at zero, still it it not set to t1(As snapshot has not been taken) and OM restart. Not sure If i am missing something here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
```diff
@@ -160,9 +164,19 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+          // the client. So take no further action and return a dummy
+          // OMClientResponse.
+          LOG.debug("Replayed Transaction {} ignored. Request: {}",
+              transactionLogIndex, createVolumeRequest);
+          return new OMVolumeCreateResponse(createReplayOMResponse(omResponse));
+        } else {
+          LOG.debug("volume:{} already exists", omVolumeArgs.getVolume());
+          throw new OMException("Volume already exists",
+              OMException.ResultCodes.VOLUME_ALREADY_EXISTS);
+        }
```
bharatviswa504: Below when an exception is caught, we can use the new OMVolumeCreateResponse  constructor which can be used for error responses.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/OMClientResponse.java
```diff
@@ -40,6 +41,21 @@ public OMClientResponse(OMResponse omResponse) {
...
 
+  /**
+   * Check if omResponse status is OK. If yes, add to DB.
+   * For OmResponse with failure, this should do nothing. This method is not
+   * called in failure scenario in OM code.
+   * @param omMetadataManager
+   * @param batchOperation
+   * @throws IOException
+   */
+  public void checkAndUpdateDB(OMMetadataManager omMetadataManager,
```
bharatviswa504: I like this approach of not checking getStatus check to OK, but I see we have some special case in S3MultipartUploadCommitPartResponse. I think in that case we need to override this method in S3MultipartUploadCommitPartResponse.
https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/s3/multipart/S3MultipartUploadCommitPartResponse.java#L70

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/volume/OMVolumeSetOwnerResponse.java
```diff
@@ -53,29 +54,35 @@ public OMVolumeSetOwnerResponse(String oldOwner,
     this.newOwnerVolumeArgs = newOwnerVolumeArgs;
   }
 
+  /**
+   * For when the request is not successful or it is a replay transaction.
+   * For a successful request, the other constructor should be used.
+   */
+  public OMVolumeSetOwnerResponse(@Nonnull OMResponse omResponse) {
+    super(omResponse);
+    Preconditions.assertTrue(!omResponse.getStatus().equals(
```
bharatviswa504: We can have a method for this precheck in a base class, which can be used in all the response classes.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/volume/OMVolumeSetQuotaResponse.java
```diff
@@ -29,29 +29,36 @@
...
 /**
  * Response for set quota request.
  */
 public class OMVolumeSetQuotaResponse extends OMClientResponse {
   private OmVolumeArgs omVolumeArgs;
 
-  public OMVolumeSetQuotaResponse(OmVolumeArgs omVolumeArgs,
-      @Nonnull OMResponse omResponse) {
+  public OMVolumeSetQuotaResponse(@Nonnull OMResponse omResponse,
+      OmVolumeArgs omVolumeArgs) {
```
bharatviswa504: Can we add @NonNull check for other params too, as now we have another constructor for error/replay scenario. The same comment applies to all classes.

bharatviswa504:
> Replay Tx 40 - Vol1 already exists => Tx will throw VOLUME_ALREADY_EXISTS when trying to createVolume

This also will be considered as replay right?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
hanishakoneru: You are right. I had updated the check in line 168 but forgot to remove this comment here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/OMClientResponse.java
hanishakoneru: Yes. I will handle this when I get to Multipart idempotence. I noticed that in some requests, we missed checking status OK. So thought we should have it in the abstract class itself as it is common code also.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/volume/OMVolumeSetOwnerResponse.java
hanishakoneru: Sure. Will update it.

bharatviswa504:
> > Replay Tx 40 - Vol1 already exists => Tx will throw VOLUME_ALREADY_EXISTS when trying to createVolume
> 
> This also will be considered as replay right?

No. This is not replay. This is another request after the restart. In non-HA after every restart ratisTransactionIndex starts from zero, we have not persisted the value.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/OMClientResponse.java
bharatviswa504: But until we reach there, when this is merged current trunk is not correct right. As now in OzoneManagerDoubleBuffer, we call addToDBBatch only when the status is OK.

hanishakoneru:
> > > Replay Tx 40 - Vol1 already exists => Tx will throw VOLUME_ALREADY_EXISTS when trying to createVolume
> > 
> > 
> > This also will be considered as replay right?
> 
> No. This is not replay. This is another request after the restart. In non-HA after every restart ratisTransactionIndex starts from zero, we have not persisted the value.

Sorry I didnt understand. In non-HA case, we have to handle differently by setting the transactionIndex to LastAppliedIndex. 
In the example I gave, this is a replay of transaction 40. I have not considered the requests which are received after the restart.

bharatviswa504:
> > > > Replay Tx 40 - Vol1 already exists => Tx will throw VOLUME_ALREADY_EXISTS when trying to createVolume
> > > 
> > > 
> > > This also will be considered as replay right?
> > 
> > 
> > No. This is not replay. This is another request after the restart. In non-HA after every restart ratisTransactionIndex starts from zero, we have not persisted the value.
> 
> Sorry I didnt understand. In non-HA case, we have to handle differently by setting the transactionIndex to LastAppliedIndex.
> In the example I gave, this is a replay of transaction 40. I have not considered the requests which are received after the restart.

No, what I mean is after OM restart, let's say setQuotaRequest comes with transactionindex 0,  but in OM DB that volume has updateID has 100, we consider them as a replay. So, in non-HA this is not a replay just a normal transaction, we should perform SetQuota normally.

bharatviswa504:
> > > > Replay Tx 40 - Vol1 already exists => Tx will throw VOLUME_ALREADY_EXISTS when trying to createVolume
> > > 
> > > 
> > > This also will be considered as replay right?
> > 
> > 
> > No. This is not replay. This is another request after the restart. In non-HA after every restart ratisTransactionIndex starts from zero, we have not persisted the value.
> 
> Sorry I didnt understand. In non-HA case, we have to handle differently by setting the transactionIndex to LastAppliedIndex.
> In the example I gave, this is a replay of transaction 40. I have not considered the requests which are received after the restart.


`I have not considered the requests which are received after the restart.`
Replay means after restart only right. Sorry not understand what you are referring here.

hanishakoneru:
> Replay means after restart only right. Sorry not understand what you are referring here.
Replay means the transaction was already executed once and it is being executed again as the logs are replayed.

hanishakoneru:
> No, what I mean is after OM restart, let's say setQuotaRequest comes with transactionindex 0, but in OM DB that volume has updateID has 100, we consider them as a replay. So, in non-HA this is not a replay just a normal transaction, we should perform SetQuota normally.

Agree. Will post an update to fix non-HA case by updating the transactionID on restart.

bharatviswa504:
> > Replay means after restart only right. Sorry not understand what you are referring here.
> > Replay means the transaction was already executed once and it is being executed again as the logs are replayed.

One question, log replay can happen with out a restart also?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/OMClientResponse.java
hanishakoneru: Understood now. Thanks for catching this. I will update it.

bharatviswa504:
LGTM. Test failures and check style issues are related to the patch.

bharatviswa504:
CI failures look related and caused by the latest commit of making addDBToBatch protected.

hanishakoneru:
Thanks @bharatviswa504 . I added some unit tests as you suggested and fixed the compilation error.

bharatviswa504:

PR #374, opened 2019-12-18
HDDS-2774. Hadoop 3.1 acceptance test fails with apk command not found

adoroszlai:

Update `ozone-mr/hadoop31` test with recent changes made in `hadoop:3.1.2` image:

 * It is based on CentOS, not Alpine, so `apk` doesn't work.
 * `pip` and `robotframework` are both pre-installed.
 * But `python` needs to point to `python3`.

Also check if `apk` is present, in case the new image is reverted or otherwise updated to be Alpine-based again.

https://issues.apache.org/jira/browse/HDDS-2774

## How was this patch tested?

Tested `ozone-mr/hadoop31` locally with both old and new image.

Successful run in fork: https://github.com/adoroszlai/hadoop-ozone/runs/354300796

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop31/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop31/test.sh
@@ -24,7 +24,7 @@ source "$COMPOSE_DIR/../../testlib.sh"
 start_docker_env
 
 #rm is the container name (resource manager) and not the rm command
-execute_command_in_container rm bash -c "if test -e /sbin/apk; then sudo apk add --update py-pip; sudo pip install robotframework; else sudo alternatives --set python /usr/bin/python3; fi"
+execute_command_in_container rm bash -c "if test -e /sbin/apk; then sudo apk add --update py-pip; sudo pip install robotframework; fi"
 
 execute_robot_test scm createmrenv.robot
 
--- a/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop31/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop31/test.sh
@@ -23,12 +23,10 @@ source "$COMPOSE_DIR/../../testlib.sh"
 
 start_docker_env
 
-execute_robot_test scm createmrenv.robot
-
-
 #rm is the container name (resource manager) and not the rm command
-execute_command_in_container rm sudo apk add --update py-pip
-execute_command_in_container rm sudo pip install robotframework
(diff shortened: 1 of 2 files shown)


hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop31/test.sh
```diff
@@ -23,12 +23,10 @@ source "$COMPOSE_DIR/../../testlib.sh"
 
 start_docker_env
 
-execute_robot_test scm createmrenv.robot
-
-
 #rm is the container name (resource manager) and not the rm command
-execute_command_in_container rm sudo apk add --update py-pip
-execute_command_in_container rm sudo pip install robotframework
+execute_command_in_container rm bash -c "if test -e /sbin/apk; then sudo apk add --update py-pip; sudo pip install robotframework; fi"
```
xiaoyuyao: Do we need to add this check for hadoop27/test.sh as well?

When /sbin/apk does not exist, we skip pip install, etc via apk. How does the test pass without a robotframework install in the new image? If the new image moving forward already contain the roboframework, can we remove the logic?

should we use "-f" instead of "-e" to test the existence of a file?

adoroszlai: > Do we need to add this check for hadoop27/test.sh as well?

The docker image for 2.7 hasn't been updated to CentOS, so not needed there.

> When /sbin/apk does not exist, we skip pip install, etc via apk. How does the test pass without a robotframework install in the new image? If the new image moving forward already contain the roboframework, can we remove the logic?

Yes, the new image has both `pip` and `robotframework`.  The conditional logic to install them can be safely removed, as long as Alpine is gone for good.

> should we use "-f" instead of "-e" to test the existence of a file?

My intention was to make it work with symlinks.  But after some testing it turns out both `-f` and `-e` return true for links that point to files, too.  So `-f` is probably better, but it only matters in case `/sbin/apk` happens to be a dir.

adoroszlai:
Thanks @xiaoyuyao for the review.  I'd rather make any further tweaks in a follow-up, to unblock CI builds.

hanishakoneru:
Thanks @adoroszlai for working on this. LGTM. +1.

xiaoyuyao:
bq. I'd rather make any further tweaks in a follow-up, to unblock CI builds.

Make sense to me. Let's do the improvement in follow up JIRAs.

adoroszlai:
Thanks @hanishakoneru for the review and @xiaoyuyao for the commit.

PR #377, opened 2019-12-19
HDDS-2773. Configure Goldmark renderer

adoroszlai:

Enable `unsafe` option for the new [Goldmark renderer](https://gohugo.io/getting-started/configuration-markup/#goldmark) introduced in Hugo 0.60.

> **unsafe**
> By default, Goldmark does not render raw HTMLs and potentially dangerous links. If you have lots of inline HTML and/or JavaScript, you may need to turn this on.

Also fix couple of incorrectly formatted links (no space allowed between `]` and `(`).

https://issues.apache.org/jira/browse/HDDS-2773

## How was this patch tested?

Built docs locally with Hugo 0.61.0.

### DIFF
--- a/hadoop-hdds/docs/config.yaml
+++ b/hadoop-hdds/docs/config.yaml
@@ -31,3 +31,7 @@ relativeURLs: true
 disableKinds:
 - taxonomy
 - taxonomyTerm
+markup:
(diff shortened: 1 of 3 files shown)


adoroszlai:

PR #378, opened 2019-12-19
HDDS-2721. Allow wrapping list of ByteBuffers with ChunkBuffer

adoroszlai:

Introduce a new `ChunkBuffer` implementation that wraps an existing list of `ByteBuffer` instances.  

For now the only usage is in `WriteChunk` (and `PutSmallFile`) where it wraps the chunk data buffer list.

https://issues.apache.org/jira/browse/HDDS-2721

## How was this patch tested?

Added unit test.

https://github.com/adoroszlai/hadoop-ozone/runs/356252796

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBufferImplWithByteBufferList.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBufferImplWithByteBufferList.java
@@ -1,0 +1,215 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBufferImplWithByteBufferList.java
```diff
@@ -0,0 +1,215 @@
...
+
+    final List<ByteBuffer> duplicates = new ArrayList<>(buffers.size());
+    int min = 0;
+    for (final ByteBuffer buf : buffers) {
+      final int max = min + buf.limit();
+      final int pos = relativeToRange(newPosition, min, max);
+      final int lim = relativeToRange(newLimit, min, max);
+
+      final ByteBuffer duplicate = buf.duplicate();
+      duplicate.position(pos).limit(lim);
```
lokeshj1703: I think position and limit are set with buf.duplicate(). We might not need the computation of min, max and lim in that case.

adoroszlai: Calculation of min/max/lim is required because `ChunkBuffer#duplicate` combines operation of `duplicate` and `position`/`limit` (by accepting `newLimit` and `newPosition` parameters).

adoroszlai:
Thanks @bshashikant and @lokeshj1703 for reviewing it.

> Should we replace ChunkBufferImplWithByteBuffer with ChunkBufferImplWithByteBufferList by including implementation for singleton list?

I'd rather keep `ChunkBufferImplWithByteBuffer` to avoid any potential runtime penalties.

PR #379, opened 2019-12-19
HDDS-2750. OzoneFSInputStream to support StreamCapabilities

cxorm:
This PR was created to let client probe for `read(ByteBuffer)` ability. 
(Cause the interface `StreamCapabilities` doesn't has constant string `READBYTEBUFFER` for Hadoop 3.2.0, I use `in:readbytebuffer` and this case will be updated after the new version of Hadoop is applied.)

#### Proposed fix.
We couldn't implement `StreamCapabilities` from `OzoneFSInputStream` directly cause the older version Hadoop can't find the class.

So we added `CapableOzoneFSInputStream` that extends `OzoneFSInputStream` and implements `StreamCapabilities`.

Then we maked
`BasicOzoneFileSystem#createFSInputStream(InputStream)` return `OzoneFSInputStream` and 
`OzoneFileSystem#createFSInputStream(InputStream)` return `CapableOzoneFSInputStream` that implements `StreamCapabilities`.

With these fixes, the older version Hadoop couldn't report `NoClassDefFoundError`.

https://issues.apache.org/jira/browse/HDDS-2750

## How was this patch tested?
UT is updated, and [Github Actions](https://github.com/apache/hadoop-ozone/runs/388775928).

### DIFF
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
@@ -38,7 +38,7 @@ import org.apache.hadoop.fs.Seekable;
  */
 @InterfaceAudience.Private
 @InterfaceStability.Evolving
-public final class OzoneFSInputStream extends FSInputStream
+public class OzoneFSInputStream extends FSInputStream
     implements ByteBufferReadable {
 
   private final InputStream inputStream;
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneStreamCapabilities.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneStreamCapabilities.java
@@ -1,0 +1,38 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.fs.ozone;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Utility class to query streams for supported capabilities of Ozone.
+ * Capability strings must be in lower case.
+ */
+final class OzoneStreamCapabilities {
+
+  private OzoneStreamCapabilities() {
+  }
+
+  /**
+   * Stream read(ByteBuffer) capability implemented by
+   * {@link OzoneFSInputStream#read(ByteBuffer)}.
+   *
+   * TODO: If Hadoop dependency is upgraded, this string can be removed.
+   */
+  static final String READBYTEBUFFER = "in:readbytebuffer";
+}
--- a/hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java
+++ b/hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java
@@ -32,6 +32,7 @@ import java.util.function.IntFunction;
 
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 /**
  * Tests for {@link OzoneFSInputStream}.
@@ -111,6 +112,17 @@ public class TestOzoneFSInputStream {
       assertEquals(-1, bytesRead);
       assertEquals(0, buf.position());
     }
+  }
+
+  @Test
+  public void testStreamCapability() {
+    final OzoneFSInputStream subject = createTestSubject(emptyStream());
+    final CapableOzoneFSInputStream capableOzoneFSInputStream =
+        new CapableOzoneFSInputStream(subject,
+            new FileSystem.Statistics("test"));
+
+    assertTrue(capableOzoneFSInputStream.
+        hasCapability(OzoneStreamCapabilities.READBYTEBUFFER));
   }
 
   private static OzoneFSInputStream createTestSubject(InputStream input) {
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -20,6 +20,7 @@ package org.apache.hadoop.fs.ozone;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.EnumSet;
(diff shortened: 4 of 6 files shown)


hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
```diff
@@ -98,4 +100,24 @@ public int read(ByteBuffer buf) throws IOException {
...
+   * @return True if the stream supports capability.
+   */
+  @Override
+  public boolean hasCapability(String capability) {
+    switch (StringUtils.toLowerCase(capability)) {
+
+    /*TODO: Update this case to READBYTEBUFFER when
+     *  dependency version of Hadoop is 3.3.0 or later.
+     */
+    case "in:readbytebuffer":
```
adoroszlai: Would it make sense to introduce an `OzoneStreamCapabilities` interface extending Hadoop `StreamCapabilities` to hold this constant?  It could be used Ozone-wide (maybe only here and the test case), and constants could be dropped when Hadoop dependency is upgraded.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneStreamCapabilities.java
```diff
@@ -0,0 +1,69 @@
...
+    private final String capability;
+
+    StreamCapability(String value) {
+      this.capability = value;
+    }
+
+    public final String getValue() {
+      return capability;
+    }
+  }
```
adoroszlai: No need to add this deprecated enum.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneStreamCapabilities.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Query the stream for a specific capability.
+   *
+   * @param capability string to query the stream support for.
+   * @return True if the stream supports capability.
+   */
+  boolean hasCapability(String capability);
```
adoroszlai: No need to duplicate the method from parent interface.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneStreamCapabilities.java
```diff
@@ -0,0 +1,69 @@
...
+ * to define their own capabilities.
+ */
+@InterfaceAudience.Public
+@InterfaceStability.Evolving
+public interface OzoneStreamCapabilities extends StreamCapabilities {
+
+
+  /**
+   * Stream read(ByteBuffer) capability implemented by
+   * {@link ByteBufferReadable#read(java.nio.ByteBuffer)}.
```
adoroszlai: This `@link` to `ByteBufferReadable` needs an import, since they are not in the same package.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneStreamCapabilities.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.hadoop.classification.InterfaceAudience;
+import org.apache.hadoop.classification.InterfaceStability;
+import org.apache.hadoop.fs.StreamCapabilities;
+/**
+ * Interface to query streams for supported capabilities of Ozone.
+ *
+ * Capability strings must be in lower case.
+ *
+ * Constant strings are chosen over enums in order to allow other file systems
+ * to define their own capabilities.
```
adoroszlai: No need to duplicate complete doc from parent interface.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneStreamCapabilities.java
```diff
@@ -0,0 +1,69 @@
...
+@InterfaceAudience.Public
+@InterfaceStability.Evolving
+public interface OzoneStreamCapabilities extends StreamCapabilities {
+
+
+  /**
+   * Stream read(ByteBuffer) capability implemented by
+   * {@link ByteBufferReadable#read(java.nio.ByteBuffer)}.
+   */
+  String READBYTEBUFFER = "in:readbytebuffer";
```
adoroszlai: Please mention in the doc that this is from newer Hadoop, can be removed if dependency is upgraded.

hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java
```diff
@@ -152,4 +154,12 @@ public void testO3FSByteBufferRead() throws IOException {
       Assert.assertArrayEquals(value, buffer.array());
     }
   }
+
+  @Test
+  public void testHasCapability() throws IOException {
+    try (OzoneFSInputStream inputStream =
+             new OzoneFSInputStream(fs.open(filePath))) {
```
adoroszlai: There is a compile error in CI:

```
2019-12-23T18:38:08.9765142Z [ERROR] /github/workspace/hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java:[161,14] constructor OzoneFSInputStream in class org.apache.hadoop.fs.ozone.OzoneFSInputStream cannot be applied to given types;
2019-12-23T18:38:08.9775744Z   required: java.io.InputStream,org.apache.hadoop.fs.FileSystem.Statistics
2019-12-23T18:38:08.9804410Z   found: org.apache.hadoop.fs.FSDataInputStream
2019-12-23T18:38:08.9816806Z   reason: actual and formal argument lists differ in length
```

To fix it, please merge latest master (which introduced the new `Statistics` parameter in `OzoneFSInputStream` constructor) into the source branch.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneStreamCapabilities.java
cxorm: Thanks @adoroszlai for reviewing.

I would violate the ```checkstyle``` with ```[ERROR] src/main/java/org/apache/hadoop/fs/ozone/OzoneStreamCapabilities.java:[30] (design) InterfaceIsType: interfaces should describe a type and hence have methods.```  when removing the method.

I think the we could disable the ```InterfaceIsType``` by removing this [line](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/dev-support/checkstyle/checkstyle.xml#L180), but I don't know whether it is proper to do this.

How do you think about this ?

adoroszlai: Nice catch.  In that case I think it can be a class (only for holding the constant), referencing the Hadoop `StreamCapabilities` interface in javadoc.  Or we can drop the idea and define the constant directly in `OzoneFSInputStream` (in that case sorry about the dead-end suggestion).

cxorm: Thanks @adoroszlai for helping and reviewing it.
I converted ```OzoneStreamCapabilities``` to utility class, and updated related string with it.

hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java
cxorm: Thanks @adoroszlai for this info.
Rebase and fixed.

adoroszlai:
One question though: if `OzoneFSInputStream` does not implement `StreamCapabilities`, how will the new method `hasCapability` be used?

cxorm:
> One question though: if `OzoneFSInputStream` does not implement `StreamCapabilities`, how will the new method `hasCapability` be used?

Thanks @adoroszlai for the question.

I think that we still should implement interface `StreamCapabilities` or `OzoneStreamCapabilities`, while both cause `NoClassDefFoundError` in CI.
I'm trying to fix it.

How do you think about this solution ?

adoroszlai:
@cxorm see 7d204553b31720ce72ae343d55695847a2940e21 for a possible implementation.  I have run all three `ozone-mr` acceptance tests with it successfully.

cxorm:
Sorry for missing here for a while.
Thanks @adoroszlai for the suggestion of implementation.
I’m going to fix this issue soon.

cxorm:
Thanks @adoroszlai for the help.
Fixed and github CI passed.

adoroszlai:
Thanks @cxorm for the contribution.

cxorm:
Thanks @adoroszlai for the help and merge.

PR #380, opened 2019-12-19
HDDS-2704. Source tar file is not created during the relase build.

bharatviswa504:

Source tar ball is not generating

https://issues.apache.org/jira/browse/HDDS-2704

## How was this patch tested?

Ran the build command, now I see tar ball.

This PR is taken from @elek original PR #334. Thank You @elek for PR.

### DIFF
--- a/hadoop-ozone/dist/pom.xml
+++ b/hadoop-ozone/dist/pom.xml
@@ -388,6 +388,45 @@
       </dependencies>
 
     </profile>
+
+    <profile>
+      <id>src</id>
+      <activation>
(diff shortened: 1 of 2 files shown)


anuengineer:
Thank you for fixing this. I confirmed that we are able to build scr.tars with this patch committed. 
@dineshchitlangia , @elek , @arp7  FYI. @bharatviswa504 has confirmed that this patch enables the BigTop builds also.

arp7:
Thanks for the quick fix @bharatviswa504, and thanks @anuengineer for providing the additional context on this.

PR #381, opened 2019-12-19
HDDS-2781. Add ObjectID and updateID to BucketInfo to avoid replaying transactions

hanishakoneru:

This Jira has 2 objectives:
1. Add objectID and updateID to BucketInfo proto persisted to DB.
2. To ensure that bucket operations are idempotent, compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

https://issues.apache.org/jira/browse/HDDS-2781

## How was this patch tested?

Will add unit tests in the another commit.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
@@ -40,7 +40,7 @@ import com.google.common.base.Preconditions;
 /**
  * A class that encapsulates Bucket Info.
  */
-public final class OmBucketInfo extends WithMetadata implements Auditable{
+public final class OmBucketInfo extends WithMetadata implements Auditable {
   /**
    * Name of the volume in which the bucket belongs to.
    */
@@ -220,7 +220,7 @@ public final class OmBucketInfo extends WithMetadata implements Auditable{
 
   /**
    * Set the Object ID. If this value is already set then this function throws.
-   * There is a reason why we cannot use the final here. The OmVolumeArgs is
+   * There is a reason why we cannot use the final here. The OMBucketInfo is
    * deserialized from the protobuf in many places in code. We need to set
    * this object ID, after it is deserialized.
    *
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
@@ -266,8 +266,6 @@ public final class OmBucketInfo extends WithMetadata implements Auditable {
     auditMap.put(OzoneConsts.STORAGE_TYPE,
         (this.storageType != null) ? this.storageType.name() : null);
     auditMap.put(OzoneConsts.CREATION_TIME, String.valueOf(this.creationTime));
-    auditMap.put(OzoneConsts.OBJECT_ID, String.valueOf(this.getObjectID()));
-    auditMap.put(OzoneConsts.UPDATE_ID, String.valueOf(this.getUpdateID()));
     return auditMap;
   }
 
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
@@ -66,6 +66,16 @@ public final class OmBucketInfo extends WithMetadata implements Auditable{
    * Creation time of bucket.
    */
   private final long creationTime;
+  /**
+   * ObjectIDs are unique and immutable identifier for each object in the
+   * System.
+   */
+  private long objectID;
+  /**
+   * UpdateIDs are monotonically increasing values which are updated
+   * each time there is an update.
+   */
+  private long updateID;
 
   /**
    * Bucket encryption key info if encryption is enabled.
@@ -90,6 +100,8 @@ public final class OmBucketInfo extends WithMetadata implements Auditable{
                        boolean isVersionEnabled,
                        StorageType storageType,
                        long creationTime,
+                       long objectID,
+                       long updateID,
                        Map<String, String> metadata,
                        BucketEncryptionKeyInfo bekInfo) {
     this.volumeName = volumeName;
@@ -98,6 +110,8 @@ public final class OmBucketInfo extends WithMetadata implements Auditable{
     this.isVersionEnabled = isVersionEnabled;
     this.storageType = storageType;
     this.creationTime = creationTime;
+    this.objectID = objectID;
+    this.updateID = updateID;
     this.metadata = metadata;
     this.bekInfo = bekInfo;
   }
@@ -181,6 +195,22 @@ public final class OmBucketInfo extends WithMetadata implements Auditable{
   }
 
   /**
+   * Returns objectID.
+   * @return long
+   */
+  public long getObjectID() {
+    return objectID;
+  }
+
+  /**
+   * Returns updateID.
+   * @return long
+   */
+  public long getUpdateID() {
+    return updateID;
+  }
+
+  /**
    * Returns bucket encryption key info.
    * @return bucket encryption key info
    */
@@ -188,6 +218,30 @@ public final class OmBucketInfo extends WithMetadata implements Auditable{
     return bekInfo;
   }
 
+  /**
+   * Set the Object ID. If this value is already set then this function throws.
+   * There is a reason why we cannot use the final here. The OmVolumeArgs is
+   * deserialized from the protobuf in many places in code. We need to set
+   * this object ID, after it is deserialized.
+   *
+   * @param obId - long
+   */
+  public void setObjectID(long obId) {
+    if(this.objectID != 0) {
+      throw new UnsupportedOperationException("Attempt to modify object ID " +
+          "which is not zero. Current Object ID is " + this.objectID);
+    }
+    this.objectID = obId;
+  }
+
+  /**
+   * Sets the update ID. For each modification of this object, we will set
+   * this to a value greater than the current value.
+   * @param updateID  long
+   */
+  public void setUpdateID(long updateID) {
+    this.updateID = updateID;
+  }
 
   /**
    * Returns new builder class that builds a OmBucketInfo.
@@ -212,6 +266,8 @@ public final class OmBucketInfo extends WithMetadata implements Auditable{
     auditMap.put(OzoneConsts.STORAGE_TYPE,
         (this.storageType != null) ? this.storageType.name() : null);
     auditMap.put(OzoneConsts.CREATION_TIME, String.valueOf(this.creationTime));
+    auditMap.put(OzoneConsts.OBJECT_ID, String.valueOf(this.getObjectID()));
+    auditMap.put(OzoneConsts.UPDATE_ID, String.valueOf(this.getUpdateID()));
     return auditMap;
   }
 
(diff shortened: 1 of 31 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
```diff
@@ -188,6 +218,30 @@ public BucketEncryptionKeyInfo getEncryptionKeyInfo() {
...
+  /**
+   * Set the Object ID. If this value is already set then this function throws.
+   * There is a reason why we cannot use the final here. The OmVolumeArgs is
+   * deserialized from the protobuf in many places in code. We need to set
+   * this object ID, after it is deserialized.
+   *
+   * @param obId - long
+   */
+  public void setObjectID(long obId) {
+    if(this.objectID != 0) {
```
bharatviswa504: Do we have guarranty that ratisTransactionIndex starts from 1?

Becoz lets take a scenario, 0 - createBucket 1 - setBucketProperty, then when setBucketProperty calls I can still call setObjectID and update objectID.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
```diff
@@ -188,6 +218,30 @@ public BucketEncryptionKeyInfo getEncryptionKeyInfo() {
     return bekInfo;
   }
 
+  /**
+   * Set the Object ID. If this value is already set then this function throws.
+   * There is a reason why we cannot use the final here. The OmVolumeArgs is
```
bharatviswa504: Do you mean here OmBucketInfo?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
```diff
@@ -212,6 +266,8 @@ public static Builder newBuilder() {
     auditMap.put(OzoneConsts.STORAGE_TYPE,
         (this.storageType != null) ? this.storageType.name() : null);
     auditMap.put(OzoneConsts.CREATION_TIME, String.valueOf(this.creationTime));
+    auditMap.put(OzoneConsts.OBJECT_ID, String.valueOf(this.getObjectID()));
```
bharatviswa504: I don't think we need to add these things to audit logs, as these are internal info to OM.
As in the audit log, if we have info of user passed information, we should be good.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketCreateRequest.java
```diff
@@ -120,7 +120,9 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
     OMMetadataManager metadataManager = ozoneManager.getMetadataManager();
 
-    BucketInfo bucketInfo = getBucketInfoFromRequest();
+    CreateBucketRequest createBucketRequest = getOmRequest()
```
bharatviswa504: Any reason for removing method, and adding that logic here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/bucket/S3BucketCreateRequest.java
```diff
@@ -214,7 +214,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
               transactionLogIndex));
 
       OMBucketCreateResponse omBucketCreateResponse =
-          new OMBucketCreateResponse(omBucketInfo, omResponse.build());
```
bharatviswa504: One question why there is no such logic of replay for S3Bucket Create Request.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/s3/bucket/S3BucketCreateResponse.java
```diff
@@ -59,19 +59,15 @@ public S3BucketCreateResponse(
   public void addToDBBatch(OMMetadataManager omMetadataManager,
```
bharatviswa504: addToDBBatch should be marked as protected, as in base class it is marked private, and it has below javadoc.
```
  /**
   * Implement logic to add the response to batch. This function should be
   * called from checkAndUpdateDB only.
   * @param omMetadataManager
   * @param batchOperation
   * @throws IOException
   */
  protected abstract void addToDBBatch(OMMetadataManager omMetadataManager,
      BatchOperation batchOperation) throws IOException
```

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
hanishakoneru: RatisTransactionIndex starts from 0 but the first index is always taken up by the conf change for leader election. I will confirm this.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
hanishakoneru: Yes. Thanks. Fixed it

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
hanishakoneru: I added as we have this information in Audit Log for VolumeInfo. Do you think this information will not be useful for debugging lets say?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketCreateRequest.java
hanishakoneru: Needed the createBucketRequest for logging in case of replay transaction. Hence, moved it from the function to avoid duplication.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/bucket/S3BucketCreateRequest.java
hanishakoneru: Missed it. Thanks for catching. I will update it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/s3/bucket/S3BucketCreateResponse.java
hanishakoneru: Done.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
bharatviswa504: Yaa agreed, but recently in one of PR a discussion came up that audit log is for to find information who did what operation, but not for debugging. So, if needed during debugging, we can log this information during performing the actual operation. 

Having precise/required information in audit logs will help when flushing audit logs and also the size of the audit log. So, I feel having this information is not really required. Thoughts?

hanishakoneru:
> I see for comments, it is marked as done/updated, but they are missing in the PR. Just letting you know about this.

Sorry I meant I fixed it and will post in the next commit. I am working on unit tests.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
hanishakoneru: Sure. I will remove this info from audit log.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketDeleteRequest.java
```diff
@@ -135,12 +151,16 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
 
       // Add to double buffer.
-      omClientResponse = new OMBucketDeleteResponse(volumeName, bucketName,
-          omResponse.build());
+      omClientResponse = new OMBucketDeleteResponse(omResponse.build(),
+          volumeName, bucketName);
+      LOG.debug("Deleted bucket:{} in volume:{}", bucketName, volumeName);
     } catch (IOException ex) {
+      omMetrics.incNumBucketDeleteFails();
+      LOG.error("Delete bucket failed for bucket:{} in volume:{}", bucketName,
```
bharatviswa504: Can we move back logging to before?

Why that is done like that way is, logging to be done outside the lock.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
```diff
@@ -168,12 +191,17 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
       omResponse.setSetBucketPropertyResponse(
           SetBucketPropertyResponse.newBuilder().build());
-      omClientResponse =
-          new OMBucketSetPropertyResponse(omBucketInfo, omResponse.build());
+      omClientResponse = new OMBucketSetPropertyResponse(
+          omResponse.build(), omBucketInfo);
+      LOG.debug("Setting bucket property for bucket:{} in volume:{}",
+          bucketName, volumeName);
     } catch (IOException ex) {
+      LOG.error("Setting bucket property failed for bucket:{} in volume:{}",
```
bharatviswa504: Same here, can we move logging outside the lock.

hanishakoneru: done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketDeleteRequest.java
hanishakoneru: done

hanishakoneru:
Thanks for the review @bharatviswa504 . I have addressed your comments in the new commit.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
```diff
@@ -266,8 +266,6 @@ public static Builder newBuilder() {
     auditMap.put(OzoneConsts.STORAGE_TYPE,
         (this.storageType != null) ? this.storageType.name() : null);
     auditMap.put(OzoneConsts.CREATION_TIME, String.valueOf(this.creationTime));
-    auditMap.put(OzoneConsts.OBJECT_ID, String.valueOf(this.getObjectID()));
-    auditMap.put(OzoneConsts.UPDATE_ID, String.valueOf(this.getUpdateID()));
```
anuengineer: I am in 2 minds about this change. This is useful if someone deletes and re-creates a bucket with same name it would be easy for us to identify that information. No need to fix in this, just curious why you thought this is a bad idea.

anuengineer: Does this break the Audit Parser code or Audit tests ?

hanishakoneru: @bharatviswa504  pointed out that audit logs are actually used for finding out what operations were performed on the cluster, as was discussed in  some PR. And for debugging, we can use the debug logs if required. 
I am ok to go either way.

bharatviswa504:
Thank You @hanishakoneru for the contribution and @anuengineer for the review.

PR #382, opened 2019-12-20
HDDS-2777. Add bytes read statistics to Ozone FileSystem implementation

fapifta:
NOTE: please forgive me and be aware the fact that I have created the branch with the wrong JIRA ID in the name, and the commit also contains a wrong JIRA ID, please fix it when commiting or if needed let me know and I recreate the PR...

The main aim is to account the number of bytes read FileSystem statistics.
Besides that the PR adds tests around the number of bytes read and number of bytes write, the number of read operations, and the number of write operations statistics.
Also there is a fix for the BasicOzoneFileSystem#open call which was accounted as a write operation before the patch.

https://issues.apache.org/jira/browse/HDDS-2777

## How was this patch tested?

See added JUnit tests.


fapifta:
Hi Attila,

thank you for the review, somehow I did not went down in the file enough and haven't noticed that method... However I tried to review all the implementations, I missed this one.

I have pushed the discussed modifications to the PR. And also added two new tests to cover the method as well.

bharatviswa504:
Thank You @fapifta for the contribution and @adoroszlai for the review.

PR #387, opened 2019-12-23
HDDS-2787. Use try-with-resources or close resource in finally block

vivekratnavel:

Sonarlint fixes for BasicOzoneClientAdapterImpl, OzoneClientAdapterFactory, DatanodeChunkGenerator, RandomKeyGenerator

https://issues.apache.org/jira/browse/HDDS-2787

## How was this patch tested?

Verified that the project compiles successfully.

### DIFF
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
@@ -203,10 +203,9 @@ public class BasicOzoneClientAdapterImpl implements OzoneClientAdapter {
   public OzoneFSOutputStream createFile(String key, boolean overWrite,
       boolean recursive) throws IOException {
     incrementCounter(Statistic.OBJECTS_CREATED);
-    try {
-      OzoneOutputStream ozoneOutputStream = bucket
-          .createFile(key, 0, replicationType, replicationFactor, overWrite,
-              recursive);
+    try (OzoneOutputStream ozoneOutputStream = bucket
+        .createFile(key, 0, replicationType, replicationFactor, overWrite,
+            recursive)) {
       return new OzoneFSOutputStream(ozoneOutputStream.getOutputStream());
     } catch (OMException ex) {
       if (ex.getResult() == OMException.ResultCodes.FILE_ALREADY_EXISTS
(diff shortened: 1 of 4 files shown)


vivekratnavel:
@elek @anuengineer @bharatviswa504 @adoroszlai @dineshchitlangia Please review

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -203,10 +203,9 @@ protected void incrementCounter(Statistic objectsRead) {
   public OzoneFSOutputStream createFile(String key, boolean overWrite,
       boolean recursive) throws IOException {
     incrementCounter(Statistic.OBJECTS_CREATED);
-    try {
-      OzoneOutputStream ozoneOutputStream = bucket
-          .createFile(key, 0, replicationType, replicationFactor, overWrite,
-              recursive);
+    try (OzoneOutputStream ozoneOutputStream = bucket
+        .createFile(key, 0, replicationType, replicationFactor, overWrite,
+            recursive)) {
```
adoroszlai: This stream is returned to the caller for writing.  It cannot be closed in this method.

vivekratnavel: @adoroszlai Thanks for the review. You are right! I fixed this in the latest patch.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/DatanodeChunkGenerator.java
```diff
@@ -95,45 +95,47 @@ public Void call() throws Exception {
...
+            .findFirst()
+            .orElseThrow(() -> new IllegalArgumentException(
+                "Pipeline ID is NOT defined, and no pipeline " +
+                    "has been found with factor=THREE"));
+        LOG.info("Using pipeline {}", pipeline.getId());
+      }
 
-    timer = getMetrics().timer("chunk-write");
+      xceiverClientSpi =
+          new XceiverClientManager(ozoneConf).acquireClient(pipeline);
```
adoroszlai: Turns out that in addition to the client, the `XceiverClientManager` should be closed, too.

vivekratnavel: Good catch! Will fix it in the next patch!

PR #390, opened 2019-12-24
HDDS-2737. Insight point should provide service type to get correct config

cxorm:
This PR was created for providing correct service type in command 
```ozone insight config [-hV] <insightName>``` to get correct config.

I split the ```insightName``` to get the service name, and set the correct component in ```showConfig(Class, Type)```.

https://issues.apache.org/jira/browse/HDDS-2737

## How was this patch tested?
Build and compiled.
Ran on standalone cluster.

### DIFF
--- a/hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/ConfigurationSubCommand.java
+++ b/hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/ConfigurationSubCommand.java
@@ -52,20 +52,7 @@ public class ConfigurationSubCommand extends BaseInsightSubCommand
             + ")");
     System.out.println();
 
-    Type type;
-    switch (insightName.split("\\.")[0]) {
-    case "scm":
-      type = Type.SCM;
-      break;
-    case "om":
-      type = Type.OM;
-      break;
-    case "datanode":
-      type = Type.DATANODE;
-      break;
-    default:
-      throw new RuntimeException("No such component " + insightName);
-    }
+    Type type = Type.valueOf(insightName.split("\\.")[0].toUpperCase());
 
     for (Class clazz : insight.getConfigurationClasses()) {
       showConfig(clazz, type);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/ConfigurationSubCommand.java [hunk 1]
```diff
@@ -51,16 +51,31 @@ public Void call() throws Exception {
         "Configuration for `" + insightName + "` (" + insight.getDescription()
             + ")");
     System.out.println();
-    for (Class clazz : insight.getConfigurationClasses()) {
-      showConfig(clazz);
 
+    Type type;
+    switch (insightName.split("\\.")[0]) {
```
maobaolong: How about change the switch expression to the following way?

```java
try {
  Type type =Type.valueOf(insightName.split("\\.")[0]);
} catch (IllegalArgumentException e) {
   throw new RuntimeException("No such component " + insightName);
}
```

cxorm: Thanks @maobaolong for the suggestion.

I found we wouldn't catch the ```IllegalArgumentException``` that would be detected in L49 ```getInsight()``` if it is illegal.
Rest fixed.

hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/ConfigurationSubCommand.java
```diff
@@ -51,16 +51,19 @@ public Void call() throws Exception {
         "Configuration for `" + insightName + "` (" + insight.getDescription()
             + ")");
     System.out.println();
-    for (Class clazz : insight.getConfigurationClasses()) {
-      showConfig(clazz);
 
+    Type type;
```
maobaolong: Would you like to combine L55 and L56? I think it would be better.

cxorm: Thanks @maobaolong for this.
Fixed.

cxorm:
Thanks @adoroszlai and @maobaolong for helping and reviewing.
And thanks @elek for committing it.

PR #392, opened 2019-12-26
HDDS-2805. Fix sudo reset the environment variables about proxy

runzhiwang:

How to reproduce the problem:

When run the  compose/ozone-mr/hadoop27/test.sh on computer which connect outer network by proxy, it fails to execute `sudo apk add --update py-pip` in docker container. Because `sudo` will reset all environment variables including the proxy, so it fails to download  http://dl-cdn.alpinelinux.org/alpine/v3.8/community/x86_64/APKINDEX.tar.gz.

How to fix:

 By `sudo -E`, it will execute the command with the environment of the user, thus the proxy can be valid. 

https://issues.apache.org/jira/projects/HDDS/issues/HDDS-2805

## How was this patch tested?

Execute  compose/ozone-mr/hadoop27/test.sh on computer which connect outer network by proxy .

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/test.sh
@@ -27,8 +27,8 @@ execute_robot_test scm createmrenv.robot
 
 
 #rm is the container name (resource manager) and not the rm command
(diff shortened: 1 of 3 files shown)


runzhiwang:
@elek Could you help review this PR ? Thank you very much.

PR #398, opened 2019-12-27
HDDS-2814. Fix fail to connect s3g in docker container with network proxy

runzhiwang:

How to reproduce the problem:

1. edit ~/.docker/config.json as follow image to connect outer network by proxy
image

2. run compose/ozones3-haproxy/test.sh, it fails to `curl http://scm:9876` as the image of log.html show. Because `curl` try to resolve http://scm:9876 by the network proxy, which caused fail.  Actually, http://scm:9876 is the local realm name used by the docker container which should not resolved by proxy.

image

How to fix:

config `no_proxy=om,scm,s3g,kdc,localhost,127.0.0.1` in docker-config

https://issues.apache.org/jira/browse/HDDS-2814

## How was this patch tested?

Execute compose/ozones3-haproxy/test.sh on computer which connect outer network by proxy .


runzhiwang:
@adoroszlai  Could you help review this PR ? It's also related to proxy. Thank you very much.

PR #399, opened 2019-12-28
HDDS-2424. Add the recover-trash command server side handling.

cxorm:
This PR was created for adding the server side of recover-trash.

Components updated including (in propagation order.)
```OzoneManager``` and ```OzoneManagerRequestHandler```
```KeyManager``` and its implementation
```OMMetadataManager``` and its implementation

Other fixes would be completed in **HDDS-2425** and **HDDS-2426**
(Including ```startKey``` and ```prefix```)

#### Note
Cause `recoverTrash` is write request, we should handle the request with [OMHA](https://issues.apache.org/jira/browse/HDDS-505).

With this [doc](https://issues.apache.org/jira/secure/attachment/12973260/Handling%20Write%20Requests%20with%20OM%20HA.pdf), we use `late validation` to handle write request.

This PR mainly has parts including updating `OzoneManager` and
fixing `OzoneManagerProtocolClientSideTranslatorPB` as well as
`OMTrashRecoverRequest` and `OMTrashRecoverResponse` to handle request of write type.

https://issues.apache.org/jira/browse/HDDS-2424

## How was this patch tested?
Cause only adding the server side handling.
Just tested the propagation and ran the UT.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
@@ -500,7 +500,7 @@ public interface OzoneManagerProtocol
    * @param bucketName - The bucket name.
    * @param keyName - The key user want to recover.
    * @param destinationBucket - The bucket user want to recover to.
-   * @return The recoverTrash
+   * @return The result of recovering operation is success or not.
    * @throws IOException
    */
   boolean recoverTrash(String volumeName, String bucketName, String keyName,
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
@@ -503,7 +503,9 @@ public interface OzoneManagerProtocol
    * @return The result of recovering operation is success or not.
    * @throws IOException
    */
-  boolean recoverTrash(String volumeName, String bucketName, String keyName,
-      String destinationBucket) throws IOException;
+  default boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationBucket) throws IOException {
+    return false;
+  }
 
 }
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
@@ -170,18 +170,6 @@ public interface KeyManager extends OzoneManagerFS, IOzoneAcl {
       String startKeyName, String keyPrefix, int maxKeys) throws IOException;
 
   /**
-   * Recover trash allows the user to recover the keys
-   * that were marked as deleted, but not actually deleted by Ozone Manager.
-   * @param volumeName - The volume name.
-   * @param bucketName - The bucket name.
-   * @param keyName - The key user want to recover.
-   * @param destinationBucket - The bucket user want to recover to.
-   * @return The result of recovering operation is success or not.
-   */
-  boolean recoverTrash(String volumeName, String bucketName,
-      String keyName, String destinationBucket) throws IOException;
-
-  /**
    * Returns a list of pending deletion key info that ups to the given count.
    * Each entry is a {@link BlockGroup}, which contains the info about the
    * key name and all its associated block IDs. A pending deletion key is
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
@@ -170,6 +170,18 @@ public interface KeyManager extends OzoneManagerFS, IOzoneAcl {
       String startKeyName, String keyPrefix, int maxKeys) throws IOException;
 
   /**
+   * Recover trash allows the user to recover the keys
+   * that were marked as deleted, but not actually deleted by Ozone Manager.
+   * @param volumeName - The volume name.
+   * @param bucketName - The bucket name.
+   * @param keyName - The key user want to recover.
+   * @param destinationBucket - The bucket user want to recover to.
+   * @return The result of recovering operation is success or not.
+   */
+  boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationBucket) throws IOException;
+
+  /**
    * Returns a list of pending deletion key info that ups to the given count.
    * Each entry is a {@link BlockGroup}, which contains the info about the
    * key name and all its associated block IDs. A pending deletion key is
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -887,19 +887,6 @@ public class KeyManagerImpl implements KeyManager {
   }
 
   @Override
-  public boolean recoverTrash(String volumeName, String bucketName,
-      String keyName, String destinationBucket) throws IOException {
-
-    Preconditions.checkNotNull(volumeName);
-    Preconditions.checkNotNull(bucketName);
-    Preconditions.checkNotNull(keyName);
-    Preconditions.checkNotNull(destinationBucket);
-
-    return metadataManager.recoverTrash(volumeName, bucketName, keyName,
-        destinationBucket);
-  }
-
-  @Override
   public List<BlockGroup> getPendingDeletionKeys(final int count)
       throws IOException {
     return  metadataManager.getPendingDeletionKeys(count);
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -887,6 +887,19 @@ public class KeyManagerImpl implements KeyManager {
   }
 
   @Override
+  public boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationBucket) throws IOException {
+
+    Preconditions.checkNotNull(volumeName);
+    Preconditions.checkNotNull(bucketName);
+    Preconditions.checkNotNull(keyName);
+    Preconditions.checkNotNull(destinationBucket);
+
+    return metadataManager.recoverTrash(volumeName, bucketName, keyName,
+        destinationBucket);
+  }
+
+  @Override
   public List<BlockGroup> getPendingDeletionKeys(final int count)
       throws IOException {
     return  metadataManager.getPendingDeletionKeys(count);
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
@@ -804,9 +804,6 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
     return deletedKeys;
   }
 
-  /**
-   * @param userName volume owner, null for listing all volumes.
-   */
   @Override
   public boolean recoverTrash(String volumeName, String bucketName,
       String keyName, String destinationBucket) throws IOException {
@@ -819,6 +816,9 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
     return recoverOperation;
   }
 
+  /**
+   * @param userName volume owner, null for listing all volumes.
+   */
   @Override
   public List<OmVolumeArgs> listVolumes(String userName,
       String prefix, String startKey, int maxKeys) throws IOException {
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
@@ -808,6 +808,16 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
    * @param userName volume owner, null for listing all volumes.
    */
   @Override
+  public boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationBucket) throws IOException {
+
+    // TODO: core logic stub would be added in later patch.
+
+    boolean recoverOperation = true;
+    return recoverOperation;
+  }
+
+  @Override
   public List<OmVolumeArgs> listVolumes(String userName,
       String prefix, String startKey, int maxKeys) throws IOException {
 
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
@@ -811,7 +811,9 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
   public boolean recoverTrash(String volumeName, String bucketName,
       String keyName, String destinationBucket) throws IOException {
 
-    // TODO: core logic stub would be added in later patch.
+    /* TODO: HDDS-2425 and HDDS-2426
+        core logic stub would be added in later patch.
+     */
 
     boolean recoverOperation = true;
     return recoverOperation;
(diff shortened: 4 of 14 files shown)

### CODE REVIEW

File: hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java [hunk 1]
```diff
@@ -871,6 +871,19 @@ private boolean isKeyEmpty(OmKeyInfo keyInfo) {
      startKeyName, keyPrefix, maxKeys);
   }
 
+  @Override
+  public boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationBucket) throws IOException {
+
+    Preconditions.checkNotNull(volumeName);
```
maobaolong: Not sure the @Nonnull is a better way? I saw the Nonnull annotation in the  hadoop-ozone project already, and findbugs will find null argument during build.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
```diff
@@ -804,6 +804,16 @@ public boolean isBucketEmpty(String volume, String bucket)
     return deletedKeys;
   }
 
+  @Override
+  public boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationBucket) throws IOException {
+
+    // TODO: core logic stub would be added in later patch.
```
maobaolong: Maybe we need a JIRA tickets track this item?

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestTrashService.java
```diff
@@ -86,11 +86,10 @@ public void testRecoverTrash() throws IOException {
     String destinationBucket = "destBucket";
     createAndDeleteKey(keyName);
 
-    /* TODO:HDDS-2424. */
```
maobaolong: Yeah, what i means is like this TODO comment.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
cxorm: Thanks @maobaolong for the suggestion.

Yeah, there are HDDS-2425 and HDDS-2426 that tracked this item.
Updated.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
cxorm: Thanks @maobaolong for this advice.

As I know, we could use both ```@Nonnull``` and ```Preconditions.checkNotNull()```, and the former is to inform developer not use ```null``` in the parameter, and the later is to validate in runtime.

So, I think this part would not be fixed, and we could create a Jira for the annotation issue.

maobaolong: ok

bharatviswa504:
Recover trash is a write request command as this moves the keys from the delete table to the original key table. let me know if I am missing something here.

2 comments.
1. In code it is marked as read-only, should it be write request in OM?
2. If write request should follow the implementation of write requests (preExecute/ValidateAndUpdatecache)

As for write request commands, we should follow HA kind of request implementation. For reference this design link [write link](https://issues.apache.org/jira/secure/attachment/12973260/Handling%20Write%20Requests%20with%20OM%20HA.pdf)

bharatviswa504:
@cxorm Attached the Cache Design document to HDDS-505. Sorry for the trouble the links in the write link above are internal links. So, attached them to HDDS-505

cxorm:
> @cxorm Attached the Cache Design document to HDDS-505. Sorry for the trouble the links in the document of write link above are internal links. So, attached them to HDDS-505

Thank you @bharatviswa504 for the document.
I'm going to fix this.

cxorm:
Thanks @bharatviswa504 for the document.

The `recoverTrash` is type of write request, and with the document
I updated the  PR with `OMTrashRecoverRequest#preExecute` and `OMTrashRecoverRequest#validateAndUpdateCache` as well as the draft of `OMTrashRecoverResponse`.
The fully implementation would be completed by following PR.

elek:
@bharatviswa504 Can we commit this PR?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2237,13 +2237,22 @@ public void deleteKey(OmKeyArgs args) throws IOException {
...
   }
 
-  // TODO: HDDS-2424. recover-trash command server side handling.
   @Override
   public boolean recoverTrash(String volumeName, String bucketName,
       String keyName, String destinationBucket) throws IOException {
 
-    boolean recoverOperation = true;
-    return recoverOperation;
+    if (isAclEnabled) {
```
bharatviswa504: We don't need this anymore. As OMTrashRecoverRequest is handling recover request.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -871,6 +871,19 @@ private boolean isKeyEmpty(OmKeyInfo keyInfo) {
      startKeyName, keyPrefix, maxKeys);
   }
 
+  @Override
+  public boolean recoverTrash(String volumeName, String bucketName,
```
bharatviswa504: We don't need this any more, after adding OMTrashRecoverRequest.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
```diff
@@ -169,6 +169,18 @@ OmKeyLocationInfo allocateBlock(OmKeyArgs args, long clientID,
   List<RepeatedOmKeyInfo> listTrash(String volumeName, String bucketName,
       String startKeyName, String keyPrefix, int maxKeys) throws IOException;
 
+  /**
```
bharatviswa504: Same here, this can be removed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMTrashRecoverResponse.java
```diff
@@ -0,0 +1,64 @@
...
+      /* TODO: HDDS-2425. HDDS-2426. */
+    String trashKey = omMetadataManager
+        .getOzoneKey(omKeyInfo.getVolumeName(),
+            omKeyInfo.getBucketName(), omKeyInfo.getKeyName());
+    RepeatedOmKeyInfo repeatedOmKeyInfo = omMetadataManager
+        .getDeletedTable().get(trashKey);
+    omKeyInfo = OmUtils.prepareKeyForRecover(omKeyInfo, repeatedOmKeyInfo);
+    omMetadataManager.getDeletedTable()
+        .deleteWithBatch(batchOperation, omKeyInfo.getKeyName());
+    /* TODO: trashKey should be updated to destinationBucket. */
```
bharatviswa504: One question:
1. if the key is created, deleted, the key is created and the key is deleted. Now when recover, which omKeyInfo will be used from the delete table.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMTrashRecoverRequest.java
```diff
@@ -0,0 +1,138 @@
...
+        .setCmdType(Type.RecoverTrash).setStatus(Status.OK)
+        .setSuccess(true);
+
+    OMMetadataManager omMetadataManager = ozoneManager.getMetadataManager();
+    boolean acquireLock = false;
+    OMClientResponse omClientResponse = null;
+    try {
+
+      // check Acl
+      checkKeyAcls(ozoneManager, volumeName, destinationBucket, keyName,
```
bharatviswa504: For checkAcl, the key does not exist, checkKeyAcls will fail with KeyNotFound.
 So, I think we should check Bucket write Acls here?
cc @anuengineer

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMTrashRecoverResponse.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.Nonnull;
+
+/**
+ * Response for RecoverTrash request.
+ */
+public class OMTrashRecoverResponse extends OMClientResponse {
+  private OmKeyInfo omKeyInfo;
+
+  public OMTrashRecoverResponse(@Nullable OmKeyInfo omKeyInfo,
```
bharatviswa504: How we get omKeyInfo value, will be done in later jiras?

elek:
/pending Comments from @bharatviswa504 are not addressed, yet...

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
cxorm: Thanks for the review.
Updated.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
cxorm: Fixed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
cxorm: Fixed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMTrashRecoverResponse.java
cxorm: Yes, I will implement it in later patch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMTrashRecoverRequest.java
cxorm: Yes, thanks for this comment.
I think we should check bucket here. (For the deleted key not found.)
Fixed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
cxorm: Updated.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMTrashRecoverResponse.java
cxorm: Thank you @bharatviswa504 for the question.

Refer to processing of `DeletedTable` in [`OMKeyDeleteResponse#addToDBBatch()`](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeyDeleteResponse.java#L82#L88) and [`OmUtils#prepareKeyForDelete()`](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java#L489#L495), the latest deleted key is added in tail of [`RepeatedOmKeyInfo#omKeyInfoList`](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/RepeatedOmKeyInfo.java#L35).

So I think we could recover the latest deleted key from the `DeletedTable` in this `created-deleted-created-deleted` situation.

Would you please give me advice if I miss something ?

bharatviswa504: I am fine with recovering last delete key if that is the expected behavior.

>  (And when recovering the latest key, I think we should clear the old deleted key.)
We should not delete the other keys, as those keys will be picked by background trash service and the data for those keys need to be deleted.

And also doing this way, is also not correct from my understanding, let us say, we put those keys in delete table, and background delete key service will pick them up and send to SCM for deletion, at this point we got a recover trash command, so there is a chance that we recover the key which might have no data, as we submitted the request to SCM for deletion, and SCM, in turn, it will send to DN. How we shall handle this kind of scenarios?

Because deletion from delete table will happen when key purge request happens.

Code snippet link [#link]( https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyDeletingService.java#L167
)

elek:
@bharatviswa504 Can you please check it?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMTrashRecoverResponse.java
cxorm: /pending I'm tracing the background part. (Hope soon)

bharatviswa504:
@elek There are some pending comments which need to be resolved.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMTrashRecoverResponse.java
cxorm: Thank you @bharatviswa504 for taking time to review this.

Here is my thought,
We set [`modificationTime`](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java#L81L82) when deleting key.

So I think we can compare the `modificationTime` with `RECOVERY_WINDOW` to exclude keys(exist in trash-enabled buckets) from purging.

Code snippet would be added after this [line](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java#L923) might like 
```Java
if (trashEnable(info.getBucketName()) &&
    (Time.now() - info.getModificationTime()) < RECOVERY_WINDOW) {
    
  /* Would not delete key in this situation. */
}
```
Could you please give me your thoughts or ideas if I miss something, thank you.

And [here](https://issues.apache.org/jira/browse/HDDS-2416?focusedCommentId=17076151&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17076151) is discussion about trash-recovery.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
```diff
@@ -549,10 +549,12 @@ DBUpdates getDBUpdates(
...
-   * @return The recoverTrash
+   * @return The result of recovering operation is success or not.
    * @throws IOException
    */
-  boolean recoverTrash(String volumeName, String bucketName, String keyName,
-      String destinationBucket) throws IOException;
+  default boolean recoverTrash(String volumeName, String bucketName,
+      String keyName, String destinationBucket) throws IOException {
+    return false;
+  }
```
cxorm, line 558: @bharatviswa504
What do you think about this `cleanup write-request` ?
Could we set the write-operation with a `default` or 
we need a separated interface addressed the `write-operation` ?

bshashikant:
@cxorm , can you plz update the pr?

bshashikant:
@bharatviswa504 , can you please have a look?

cxorm:
Thanks @bshashikant for the reminder.
Rebase latest master-branch (#843).

cxorm:
Rebase latest master-branch (#839) to resolve conflict.

cxorm:
Rebase latest master-branch (#848) and trigger github-actions.

bshashikant:
Thanks @cxorm for working on this. I have committed this.

cxorm:
Thanks @bharatviswa504 for the review 
and thanks @bshashikant for the commit.

PR #137, opened 2019-11-09
HDDS-2455. Implement MiniOzoneHAClusterImpl#getOMLeader

smengcl:
## What changes were proposed in this pull request?

Implement MiniOzoneHAClusterImpl#getOMLeader

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-2455

## How was this patch tested?

1. Added a new test class
2. Ran the two previous unit tests that I have changed to use the new method, both passed: `TestOzoneFsHAURLs#testWithQualifiedDefaultFS` and `TestOzoneShellHA#testOzoneShCmdURIs`.

PR #400, opened 2019-12-29
HDDS-2819. Fix timeout threshold is too small to exit safemode when security enabled

runzhiwang:
#137  What changes were proposed in this pull request?

How to reproduce the problem:

1.  When I run ozone-0.5.0-SNAPSHOT/compose/ozonesecure/test.sh on computer with 8 core 1999.971 Mhz, 16G memory, it always fail and report "Safemode is still on".

2. Then I enlarge the timeout threshold in testlib.sh from 90 seconds to 360 seconds, and record the time to exist safemode 10 times,  and the average time to exit safemode is about 160 seconds, and the maxum time is about 240 seconds. 
image

image

How to fix the problem:

Enlarge the timeout threshold to 600 seconds when security enabled.

https://issues.apache.org/jira/browse/HDDS-2819

## How was this patch tested?

Run ozone-0.5.0-SNAPSHOT/compose/ozonesecure/test.sh on computer with 8 core 1999.971 Mhz, 16G memory.

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/testlib.sh
+++ b/hadoop-ozone/dist/src/main/compose/testlib.sh
@@ -40,8 +40,13 @@ wait_for_safemode_exit(){
   #Reset the timer
   SECONDS=0
 
-  #Don't give it up until 30 seconds
-  while [[ $SECONDS -lt 90 ]]; do
+  timeout_threshold=90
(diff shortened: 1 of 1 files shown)


runzhiwang:
@ChenSammi Could you help review this patch ? Thank you very much.

PR #403, opened 2020-01-01
HDDS-2820. OM Ratis dir creation may fail

adoroszlai:

It may happen that `exists()` returns `false` because the dir is does not exist yet, but `mkdirs()` also returns `false` because the dir is created in the meantime by the Ratis thread.  This PR proposes to swap the order of `exists` check and `mkdirs` call to avoid race condition.

https://issues.apache.org/jira/browse/HDDS-2820

## How was this patch tested?

Added unit test.

https://github.com/adoroszlai/hadoop-ozone/runs/366839202


adoroszlai:
Thanks @cxorm and @bshashikant for the review, and @bshashikant for committing it.

PR #404, opened 2020-01-02
HDDS-2552. Sonar: Save and reuse Random object in GenesisUtil

vivekratnavel:

Save and reuse Random object in GenesisUtil

https://issues.apache.org/jira/browse/HDDS-2552

## How was this patch tested?

Verified that the project compiles successfully.

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/genesis/GenesisUtil.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/genesis/GenesisUtil.java
@@ -68,6 +68,8 @@ public final class GenesisUtil {
 
   private static final int DB_FILE_LEN = 7;
   private static final String TMP_DIR = "java.io.tmpdir";
+  private static final Random RANDOM = new Random();
+  private static final String SCM_CLIENT_DEFAULT_ADDRESS = "127.0.0.1:0";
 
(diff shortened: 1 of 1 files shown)


vivekratnavel:
@adoroszlai @dineshchitlangia @nandakumar131 Please review

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/genesis/GenesisUtil.java
```diff
@@ -68,6 +68,8 @@ private GenesisUtil() {
 
   private static final int DB_FILE_LEN = 7;
   private static final String TMP_DIR = "java.io.tmpdir";
+  private static final Random RANDOM = new Random();
+  private static final String SCM_CLIENT_DEFAULT_ADDRESS = "127.0.0.1:0";
```
adoroszlai: Given that this is also used for non-client SCM and OM addresses, wouldn't something like `RANDOM_LOCAL_ADDRESS` be a better name?

vivekratnavel: Sure. Fixed it!

vivekratnavel:
@adoroszlai Thanks for the review and merge!

PR #405, opened 2020-01-02
HDDS-2763. Handle Datanode registration in Recon.

avijayanhwx:
- Datanode should use Recon address to register and heartbeat to it. 
- Recon will act as a "passive" SCM to the Datanode.  Here "passive" means that the Datanode will not get any commands from this Recon SCM. 
- Integrating SCM Node Manager into Recon.

https://issues.apache.org/jira/browse/HDDS-2763

## How was this patch tested?
Manually tested using docker.
Unit tests pending.


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
```diff
@@ -64,6 +66,8 @@
 import org.apache.hadoop.net.NetUtils;
 
 import com.google.common.net.HostAndPort;
+
+import static org.apache.hadoop.hdds.recon.ReconConfigKeys.*;
```
swagle: Advice to add explicit imports.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/recon/ReconConfigKeys.java
```diff
@@ -32,9 +32,11 @@ private ReconConfigKeys() {
 
   public static final String OZONE_RECON_DATANODE_ADDRESS_KEY =
       "ozone.recon.datanode.address";
+  public static final String OZONE_RECON_ADDRESS_KEY =
+      "ozone.recon.address";
```
swagle: Suggest rename: ozone.recon.server.address

swagle:
We should use something like serverId on commands as blanket assurance to make sure DN does not process any commands from passive servers.

avijayanhwx:
> We should use something like serverId on commands as blanket assurance to make sure DN does not process any commands from passive servers.

Yes, I will be doing that in a later PR, most likely in HDDS-2850.

avijayanhwx:
Thanks for the review @swagle.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/recon/ReconConfigKeys.java
avijayanhwx: I went with ozone.recon.address to keep it consistent with ozone.om.address and ozone.scm.names.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
avijayanhwx: Fixed

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/recon/ReconConfigKeys.java
```diff
@@ -32,9 +32,11 @@ private ReconConfigKeys() {
...
   public static final String OZONE_RECON_DATANODE_BIND_HOST_KEY =
       "ozone.recon.datanode.bind.host";
   public static final String OZONE_RECON_DATANODE_BIND_HOST_DEFAULT =
       "0.0.0.0";
-  public static final int OZONE_RECON_DATANODE_PORT_DEFAULT = 9865;
+  public static final int OZONE_RECON_DATANODE_PORT_DEFAULT = 9891;
```
anuengineer: This is already committed? Should we remove this change from here ?

avijayanhwx: Yes, it can be removed

PR #406, opened 2020-01-02
HDDS-2784. Read to ByteBuffer uses wrong offset

adoroszlai:

`OzoneFSInputStream#read(ByteBuffer)` uses the target buffer's position for offsetting into the temporary array, which may result in an `IndexOutOfBoundsException`.  Offset should always be 0, since the temporary array is sized exactly for the length to read, and it has no extra data at the start.

https://issues.apache.org/jira/browse/HDDS-2784

## How was this patch tested?

Added unit test.

```
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.143 s - in org.apache.hadoop.fs.ozone.contract.TestOzoneFSInputStreamUnit
```

https://github.com/adoroszlai/hadoop-ozone/runs/370513667

### DIFF
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneFSInputStream.java
@@ -104,11 +104,10 @@ public final class OzoneFSInputStream extends FSInputStream
       throw new ReadOnlyBufferException();
     }
 
-    int bufInitPos = buf.position();
     int readLen = Math.min(buf.remaining(), available());
 
     byte[] readData = new byte[readLen];
-    int bytesRead = read(readData, bufInitPos, readLen);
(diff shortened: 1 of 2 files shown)


cxorm:
Thanks @adoroszlai for the report and work.

Overll LGTM.
A little question, could we add new unit test in the existed file ? 
(```hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java```)

adoroszlai:
Thanks @cxorm for reviewing this.

> A little question, could we add new unit test in the existed file ?
> (`hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java`)

That file is an integration test (uses mini cluster) and I proposed moving it to `integration-test` in #383.  The new test cases are proper unit tests, and can be run without cluster.

adoroszlai:
Thanks @cxorm and @bshashikant for review, and @bshashikant for committing it.

PR #407, opened 2020-01-02
HDDS-2829. Deduplicate KDC docker image definition

adoroszlai:

Move definition of insecure Ozone KDC image (for security tests) to a common location for both `ozonesecure` and `ozonesecure-mr` to reuse.

 * Fix: `[Warning] One or more build-args [buildno] were not consumed`
 * Tag the image built for reuse in same acceptance test run
 * Squash `apk add` calls
 * No need to `rm` cache since using `--no-cache`

https://issues.apache.org/jira/browse/HDDS-2829

## How was this patch tested?

Built image locally, verified that:

 * warning is gone
 * image size did not increase from previous version

Verified that the image is created for `ozonesecure-mr` and then reused for `ozonesecure` in CI.

https://github.com/adoroszlai/hadoop-ozone/runs/370732273


xiaoyuyao:
Thanks @adoroszlai  for the improvement. The change LGTM, +1. I will merge it shortly.

adoroszlai:
Thanks @xiaoyuyao for reviewing and merging it.

PR #409, opened 2020-01-02
HDDS-2830. Remove unnecessary sleep in acceptance test

adoroszlai:

`start_docker_env` has a 10-second sleep after environment is up.  I think it was originally added to give SCM a chance to get out of safe mode after all datanodes are up.  Now that acceptance test has a proper `wait_for_safemode_exit` condition, I think the sleep is no longer necessary.

https://issues.apache.org/jira/browse/HDDS-2830

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/371306676

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/testlib.sh
+++ b/hadoop-ozone/dist/src/main/compose/testlib.sh
@@ -32,7 +32,7 @@ create_results_dir() {
 }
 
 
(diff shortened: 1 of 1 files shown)


adoroszlai:
Thanks @nandakumar131 for reviewing and merging this.

PR #413, opened 2020-01-07
HDDS-2834. Directly read into ByteBuffer if it has array

adoroszlai:

 * Use target `ByteBuffer`'s array, if it has one, to avoid buffer copy in `OzoneFSInputStream`.
 * Skip `ByteBuffer#put` on EOF (see similar logic in `CryptoInputStream`).
 * Rename `TestOzoneFSInputStreamUnit` to `TestOzoneFSInputStream` since integration test with same name was moved out to `integration-test`.  Also move it to the correct package (where `OzoneFSInputStream` resides).

https://issues.apache.org/jira/browse/HDDS-2834

## How was this patch tested?

Tweaked unit test to exercise the `read` method with both heap and direct buffers.

https://github.com/adoroszlai/hadoop-ozone/runs/377273986


adoroszlai:
@fapifta please review if you have some time

bharatviswa504:

adoroszlai:
Thanks @bharatviswa504 for reviewing and merging this.

PR #415, opened 2020-01-07
HDDS-2840. Implement ofs://: mkdir

smengcl:

Implement mkdir in ofs://

https://issues.apache.org/jira/browse/HDDS-2840

## How was this patch tested?

`TestRootedOzoneFilesystem` all passed. This test class is copied and modified from `TestOzoneFilesystem`.


smengcl:
Status update:
1. For now mkdir works in ozone fs shell. `mkdir -p` will create volume and bucket if they don't exist. See commit message ebd2bf8e28976f370e8fc17ecf42b1b3c4497a3e how I tested it. Note: Hasn't been tested throughly!
2. In order for the code to compile after rebase to latest dev branch, I have manually applied HDDS-2777 and HDDS-2785 to OFS classes (as they were copied from existing o3fs classes and some APIs has changed).

smengcl:
Status update:

1. This PR is ready for review.
2. This diff should help for the review, the Base is the commit after class copying (except RootedOzoneClientAdapterFactory): https://github.com/smengcl/hadoop-ozone/compare/0bee28acadfb2c358c6f008173e7eca6ed7fa23f...smengcl:HDDS-2840
3. I have also applied HDDS-2188 to affected OFS classes in order for code to compile.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,602 @@
...
+
+  /**
+   * Apply volumeStr and bucketStr stored in the object instance before
+   * executing a FileSystem operation.
+   *
+   * @param createIfNotExist Set this to true if the caller is a write operation
+   *                         in order to create the volume and bucket.
+   * @throws IOException
+   */
+  private void getVolumeAndBucket(String volumeStr, String bucketStr,
```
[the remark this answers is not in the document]
smengcl: done

smengcl:
> Thanks for working on this @smengcl. I have started reviewing the PR.
> 
> In BasicRootedOzoneClientAdapterImpl, setting volume and bucket globally could lead to discrepancies.
> Let's say my client creates one OzoneClient and multiple threads share this object.
> Thread1 - createFile /vol1/bucket1/key1
> Thread2 - createFile /vol1/bucket2/key2
> 
> ```
> t1 - Thread1 sets bucket = bucket1
> t2 - Thread2 sets bucket = bucket2
> t3 - Thread1 executes bucket.createFile -> This would create key1 in bucket2.
> ```

Good point. The current implementation is not thread-safe.
I believe I can just make `volume` and `bucket` local in each call to eliminate such race condition.

smengcl:
The race condition issue should've be resolved in d2a01d19f0cfa2cdce5e252be2dd33fcb0d4fb6a.
All `TestRootedOzoneFileSystem` tests passed. Pending pr-checks.

Quick note: We might want to add some robot tests in another jira for the new mkdir behavior. For example, `mkdir -p` inside a volume/bucket that doesn't exist, where user has or doesn't have permission.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,805 @@
...
+  }
+
+  /**
+   * Turn a path (relative or otherwise) into an Ozone key.
+   *
+   * @param path the path of the file.
+   * @return the key of the object that represents the file.
+   */
+  public String pathToKey(Path path) {
+    Objects.requireNonNull(path, "Path can't be null!");
```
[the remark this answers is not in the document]
hanishakoneru: This is a design decision which we can leave out in this Jira and probably add a TODO to keep track, if you agree.

smengcl:
> I think here we need to catch exception from createVolume, as by the time we create volume, if another client created the same volume, we get VOLUME_ALREADY_EXCEPTION, so this should be safe and ignored.
> 
> Same applicable for bucket creation also.

Nice catch. Addressed in the last commit.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,395 @@
...
+        .build();
+    cluster.waitForClusterToBeReady();
+    objectStore = cluster.getClient().getObjectStore();
+
+    // create a volume and a bucket to be used by OFileSystem
+    OzoneBucket bucket = TestDataUtil.createVolumeAndBucket(cluster);
+    volumeName = bucket.getVolumeName();
+    bucketName = bucket.getName();
+
+    // For now:
```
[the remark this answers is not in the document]
smengcl: I forgot to clean up this line of comment. the `rootPath` below should be good.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,395 @@
...
+    // TODO: Address this properly.
+    conf.set("fs.ofs.impl", "org.apache.hadoop.fs.ozone.RootedOzoneFileSystem");
+    assertEquals(
+        FileSystem.getFileSystemClass(OzoneConsts.OZONE_OFS_URI_SCHEME,
+            conf), RootedOzoneFileSystem.class);
+  }
+
+  @Test
+  public void testCreateDoesNotAddParentDirKeys() throws Exception {
+    Path rootBucket = new Path("/" + volumeName + "/" + bucketName);
```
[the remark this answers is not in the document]
smengcl: Yep this looks duplicated for those **existing** test cases modified from O3FS tests.

The point is there will definitely be more OFS-specific test cases (like `testMkdirOnNonExistentVolumeBucket()`) introduced into the this test class which won't use `rootBucket`.

I would make this class global and rename it.

Done in 7a3de51e6e23abba3e91971f80e8666c1fc6c108.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,604 @@
...
+  }
+
+  public List<FileStatusAdapter> listStatus(String pathStr, boolean recursive,
+      String startPath, long numEntries, URI uri,
+      Path workingDir, String username) throws IOException {
+
+    OFSPath ofsPath = new OFSPath(pathStr);
+    String keyName = ofsPath.getKeyName();
+    OFSPath ofsStartPath = new OFSPath(startPath);
+    String startKey = ofsStartPath.getKeyName();
```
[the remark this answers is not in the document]
smengcl: Since I made all the input path params (`pathStr`) for `BasicRootedOzoneClientAdapterImpl` methods full path (volume+bucket+key), I believe it's best to keep `startPath` consistent. - That's why I intentionally renamed the parameter name in API definition from `startKey` to `startPath`.
 
Totally agree that doc is required. This should've been done in o3fs `BasicOzoneClientAdapterImpl` as well.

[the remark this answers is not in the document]
smengcl: Done in b21005ebacf25f5a3a6bfdb97348a2e2e567423d.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,805 @@
...
+//  private boolean gIsolatedClassloader;
+
+  private RootedOzoneClientAdapter adapter;
+//  private String adapterPath;
+
+  private static final String URI_EXCEPTION_TEXT =
+      "URL should be one of the following formats: " +
+      "ofs://om-service-id/  OR " +
+      "ofs://om-host.example.com/  OR " +
+      "ofs://om-host.example.com:5678/";
```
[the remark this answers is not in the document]
smengcl: Yep. Done: da12d487fe5c5bb05f013c33cb2d02aba96b63c3

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,805 @@
...
+        "Invalid scheme provided in " + name);
+
+    String authority = name.getAuthority();
+    if (authority == null) {
+      // authority is null when fs.defaultFS is not a qualified ofs URI and
+      // ofs:/// is passed to the client. matcher will NPE if authority is null
+      throw new IllegalArgumentException(URI_EXCEPTION_TEXT);
+    }
+
+    String omHost;
```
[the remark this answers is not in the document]
smengcl: Done, `omHost` -> `omHostOrServiceId`: 8e34c5f6718680c35b683ca4b0969286e975e318

BTW this should be applied to o3fs as well.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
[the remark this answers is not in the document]
smengcl: This `pathToKey` is added in HDDS-177 by Anu. Honestly what I did here was fixing the typo.

Your understanding is right. It should convert a relative path into an absolute one (if not). Then remove leading `/`.

But I think defaulting volume to `user` and bucket name to `<USERNAME>` in this case is exactly the way we intend it to be? So when user run `ozone fs -put textfile.txt doc/` should automatically put the file under `doc/` in user's home directory. (`workingDir` should be renamed to `homeDir` IMO) CMIIW.

Yes we are assuming that admin would create a volume named `user` and created a bucket for each user beforehand. That is indeed the implication. Let's discuss it.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,805 @@
...
+    }
+
+    @Override
+    boolean processKeyPath(String keyPath) throws IOException {
+      if (keyPath.equals("")) {
+        LOG.trace("Skipping deleting root directory");
+        return true;
+      } else {
+        LOG.trace("deleting key path:" + keyPath);
+        boolean succeed = adapter.deleteObject(keyPath);
```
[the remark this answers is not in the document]
smengcl: Attempted to address it in commit dfc50cc8b0f9d35aa5d9dfcea3e4beedb4930341. See my comment above.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,805 @@
...
+      super(srcPath);
+      this.srcPath = pathToKey(srcPath);
+      this.dstPath = pathToKey(dstPath);
+      LOG.trace("rename from:{} to:{}", this.srcPath, this.dstPath);
+    }
+
+    @Override
+    boolean processKeyPath(String keyPath) throws IOException {
+      String newPath = dstPath.concat(keyPath.substring(srcPath.length()));
+      adapter.renamePath(keyPath, newPath);
```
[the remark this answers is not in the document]
smengcl: Yep this is a problem.

I attempted to address it in commit dfc50cc8b0f9d35aa5d9dfcea3e4beedb4930341.

I do realize my solution still have 2 minor issues:
1. code duplication, e.g. overloaded `renamePath`/`deleteObject` shares mostly the same code;
2. using key in `processKeyPath` would be more efficient -- but this would break my intention of using ONLY full path outside of `AdapterImpl`.

Those two issues above could be solved with some degree of refactoring. I'm doing it this way just to show one of the ways to solve the RPC call problem. Comments?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,805 @@
...
+     * be processed successfully.
+     *
+     * @return true if all keys are processed successfully, false otherwise.
+     * @throws IOException
+     */
+    boolean iterate() throws IOException {
+      LOG.trace("Iterating path: {}", path);
+      if (status.isDirectory()) {
+        LOG.trace("Iterating directory: {}", pathKey);
+        OFSPath ofsPath = new OFSPath(pathKey);
```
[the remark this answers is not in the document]
smengcl: Yes since I intend to convert key to full path before feeding it to `processKeyPath()`. I could change it back but again, my idea of "using only full path outside of AdapterImpl" will be broken. <- this could probably be solved with refactoring but I don't think that is a priority at the moment. 

Added comment and TODO in 10fe834c08cf72911b6d0409fb1138ff7d23294e.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,805 @@
...
+
+    // Process path.
+    URI newUri = fileStatusAdapter.getPath().toUri();
+    try {
+      newUri = new URIBuilder().setScheme(newUri.getScheme())
+          .setHost(newUri.getAuthority())
+          .setPath(newUri.getPath())
+          .build();
+    } catch (URISyntaxException e) {
+    }
```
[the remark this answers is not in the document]
smengcl: This chunk of code is useless now. It was previously added for stitching the listStatus result. Just removed it.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Utility class for Rooted Ozone Filesystem (OFS) path processing.
+ */
+@InterfaceAudience.Private
+@InterfaceStability.Unstable
+class OFSPath {
+  private String volumeName;
+  private String bucketName;
+  private String mountName;
+  private String keyName;
```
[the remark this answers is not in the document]
smengcl: I added a table for illustration: 4bb7ed19674dcb959703fe633e940ab4bf17873e

Note the table might change in the future. This just shows what it does at the moment.

hadoop-ozone/ozonefs/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystemWithMocks.java
```diff
@@ -0,0 +1,152 @@
...
+    PowerMockito.when(OzoneClientFactory.getRpcClient(eq("local.host"),
+        eq(omPort), eq(conf))).thenReturn(ozoneClient);
+
+    UserGroupInformation ugi = mock(UserGroupInformation.class);
+    PowerMockito.mockStatic(UserGroupInformation.class);
+    PowerMockito.when(UserGroupInformation.getCurrentUser()).thenReturn(ugi);
+    when(ugi.getShortUserName()).thenReturn("user1");
+
+    URI uri = new URI("o3fs://bucket1.volume1.local.host");
+
```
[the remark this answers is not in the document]
smengcl: Done in 5a20f9da30a25fd40fa1a73e6b0e742f40647a94.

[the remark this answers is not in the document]
smengcl: Done in 5a20f9da30a25fd40fa1a73e6b0e742f40647a94.

smengcl:
> Thanks @smengcl for working on this patiently. I know you had to rebase with master multiple times.
> 
> I have posted some suggestions inline.
> Apart from them, I have some general suggestions. Since this work is being done in a branch, we can work on them in later patches (before merging into master).
> 
> 1. It would be good to separate OFS and O3FS related classes into separate packages. We can have fs.ozone.o3fs for o3fs and fs.ozone.ofs for OFS. The common classes can stay under fs.ozone.
> 2. The following classes have a lot of code duplication.
>    a. OzoneClientAdapter and RootedOzoneClientAdapter
>    b. BasicOzoneClientAdapterImpl and BasicRootedOzoneClientAdapterImpl
>    c. BasicOzoneFileSystem and BasicRootedOzoneFileSystem
>    d. OzoneFileSystem and RootedOzoneFileSystem
>    e. OzoneClientAdapterFactory and RootedOzoneClientAdapterFactory
>    I think we can avoid most of these code duplications by creating common interfaces and extending them for O3FS and OFS.
> 3. In OzoneClientAdapter and RootedOzoneClientAdapter, since all the methods are same with just method name difference or parameter name difference, can we not have some common ground for the name convention. We could explain what the parameter denotes in the Javadoc.
> 4. We are sometimes referring to the new FS as RootedOzoneFileSystem and sometimes as O3FS. Let's follow one naming convention throughout.
> 5. We should add unit tests for testing OFSPath functionality.

Thanks so much for reviewing such a large diff. I appreciate it.

All the small issues you mentioned in the code should have a commit to address them now. For some I added TODO, others may require further discussion.

1. I agree. That is what I set out to do first (I tried to duplicate the entire `ozonefs` package). But then I realized the amount of refactoring required and maven pom editing is non-trivial. So I took @arp7 's advice to simply go out and copy those core classes for now.
2. Correct. This can be solved in the process of solving (1). This might require some intense refactoring and change on o3fs side? I think we should do this later (and eventually).
3. Do you mean we should change ALL the function names in case others misuse them when switching from o3fs world? -- I'm neutral about this (either works for me). We can discuss.
4. OFS = RootedOzoneFileSystem. Well.. I tried class name like `OFSOzoneClientAdapter` and found it unnecessarily annoying. Since `FS` in `OFS` is `FileSystem` and it doesn't mean anything in some class names. Let's just define that `OFS` is the scheme prefix, and `Rooted` prefix is for the class names (suggested by @arp7 in a call).
5. I'm on it.

smengcl:
When writing unit tests for `OFSPath`, I realized that there is one corner case for `mkdir` that I didn't handle properly: 'mkdir /volume1'. Will push new commit to fix this in a moment.

hanishakoneru:
Thanks @smengcl. We can address the refactoring in subsequent Jiras. 
I am +1 for merging this PR into branch HDDS-2665-ofs.

bharatviswa504:
> > I think here we need to catch exception from createVolume, as by the time we create volume, if another client created the same volume, we get VOLUME_ALREADY_EXCEPTION, so this should be safe and ignored.
> > Same applicable for bucket creation also.
> 
> Nice catch. Addressed in commit [ab64a98](https://github.com/apache/hadoop-ozone/commit/ab64a98fadd52b483136f99031d8ed9d401b36c9). @bharatviswa504 Please take a look.

Thanks for addressing it. This comment can be considered as resolved

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,477 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Ozone file system tests that are not covered by contract tests.
+ */
+public class TestRootedOzoneFileSystem {
```
xiaoyuyao: The test cases are very similar to TestOzoneFileSystem, most of the difference are in setup.
Can we refactor to have a base class that contains all the common test methods and leave only the setup in the subclass for different file system? (This is the approach taken in the contract test to avoid duplicate code)

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,694 @@
...
+    try {
+      bucket = proxy.getBucketDetails(volumeStr, bucketStr);
+    } catch (OMException ex) {
+      if (createIfNotExist) {
+        // Note: getBucketDetails always throws BUCKET_NOT_FOUND, even if
+        // the volume doesn't exist.
+        if (ex.getResult().equals(BUCKET_NOT_FOUND)) {
+          OzoneVolume volume;
+          try {
+            volume = proxy.getVolumeDetails(volumeStr);
```
xiaoyuyao: should we do getVolumeDetails/create first and if that fails, there is no need to create bucket.
As createVolume is an admin only operation, I would expect most of the client driven operation will fail here.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,694 @@
...
+          } catch (OMException getVolEx) {
+            if (getVolEx.getResult().equals(VOLUME_NOT_FOUND)) {
+              // Volume doesn't exist. Create it
+              try {
+                objectStore.createVolume(volumeStr);
+              } catch (OMException newVolEx) {
+                // Ignore the case where another client created the volume
+                if (!newVolEx.getResult().equals(VOLUME_ALREADY_EXISTS)) {
```
xiaoyuyao: Even the volume creation succeed, the bucket creation might fail if the ACLs are not set properly when security and acls are enabled. With o3fs, this is OK because the volume/bucket are provisioned and acls setup before it is mounted. With ofs, this may not work without ACL changes.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,694 @@
...
+            }
+            // Try get volume again
+            volume = proxy.getVolumeDetails(volumeStr);
+          }
+          // Create the bucket
+          try {
+            volume.createBucket(bucketStr);
+          } catch (OMException newBucEx) {
+            // Ignore the case where another client created the bucket
+            if (!newBucEx.getResult().equals(BUCKET_ALREADY_EXISTS)) {
```
xiaoyuyao: what is permission denied, do we recover the environment by delete the volume?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,694 @@
...
+  public boolean createDirectory(String pathStr) throws IOException {
+    LOG.trace("creating dir for path: {}", pathStr);
+    incrementCounter(Statistic.OBJECTS_CREATED);
+    OFSPath ofsPath = new OFSPath(pathStr);
+    // Volume name unspecified, return failure
+    if (ofsPath.getVolumeName().length() == 0) {
+      return false;
+    }
+    // Handle where only volume is specified in pathStr
+    if (ofsPath.getBucketName().length() == 0) {
```
xiaoyuyao: In the /tmp case, the bucketName length is also 0? do we want to create a /tmp volume here?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,694 @@
...
+      objectStore.createVolume(ofsPath.getVolumeName());
+      return true;
+    }
+    String keyStr = ofsPath.getKeyName();
```
xiaoyuyao: what is keyStr is empty(length 0)

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,694 @@
...
+   * Helper method to delete an object specified by key name in bucket.
+   *
+   * @param path path to a key to be deleted
+   * @return true if the key is deleted, false otherwise
+   */
+  @Override
+  public boolean deleteObject(String path) {
+    LOG.trace("issuing delete for path to key: {}", path);
+    OFSPath ofsPath = new OFSPath(path);
+    String keyName = ofsPath.getKeyName();
```
xiaoyuyao: check keyName is empty?

xiaoyuyao: Do we support delete volume/bucket via OFS?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,694 @@
...
+  }
+
+  public FileStatusAdapter getFileStatus(String path, URI uri,
+      Path qualifiedPath, String userName)
+      throws IOException {
+    OFSPath ofsPath = new OFSPath(path);
+    String key = ofsPath.getKeyName();
+    try {
+      OzoneBucket bucket = getBucket(ofsPath, false);
+      incrementCounter(Statistic.OBJECTS_QUERY);
```
xiaoyuyao: should we move the counter incr before getBucket so that we can accurately count the ops even though getBucket may fail?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: @xiaoyuyao Actually this part was coded previously that it checks if the volume exists first. Then @bharatviswa504 argued that this way we would always have at least two round trips for each `getBucket()` (first `getVolumeDetails`, then `getBucketDetails`) regardless of whether the bucket already exists. This change is made in d2a01d19f0cfa2cdce5e252be2dd33fcb0d4fb6a.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Utility class for Rooted Ozone Filesystem (OFS) path processing.
+ */
+@InterfaceAudience.Private
+@InterfaceStability.Unstable
+class OFSPath {
+  /**
+   * Here is a table illustrating what each name variable is given an input path
+   * Assuming /tmp is mounted to /tempVol/tempBucket
```
xiaoyuyao: How is the /tempVol/tempBucket created? Do we provide a CLI to admin?

smengcl: This is undecided. We could give admin a CLI command to do the initialization. But we might want to leave this for jira HDDS-2929.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
smengcl: Indeed most code are different, but not just in the setup. In quite a few existing test cases I changed some subtle variable to make it run on o3fs, meaning we likely would need to change o3fs's unit test a bit as well (potentially more conflicts when merging this branch with trunk). I would suggest adding a TODO for this class and do the refactoring at a later point. What do you say?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -0,0 +1,694 @@
...
+        throw new FileAlreadyExistsException(
+            ex.getResult().name() + ": " + ex.getMessage());
+      } else {
+        throw ex;
+      }
+    }
+  }
+
+  @Override
+  public void renamePath(String path, String newPath) throws IOException {
```
xiaoyuyao: Let's change renamePath to rename

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
smengcl: Added TODO.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: As we discussed, currently in Ozone `createVolume` will set owner to current user when `volArg` owner is null: https://github.com/smengcl/hadoop-ozone/blob/046a06f02783da516179ee8d8d1bed862d22f78d/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java#L268

Can we resolve this one?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: Done in e16a52ee7629ce89a185be00ab71742f56f1b7ce.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: Actually here we need to put the `keyStr.length()` check after `getBucket`, since `getBucket` need to create volume and bucket if they don't exist. If we check it before `try` clause it wouldn't create the volume and bucket then.

smengcl: Comment updated in c0e2ed7f4cf08490b151613fdb3415b097ee5ac0.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: keyName check done in ac0cd77f9c22c19c7442f7e248ab401bbb9ffb73.

smengcl: On supporting delete volume/bucket though, this is a good question. I'd argue it is weird to support delete volume/bucket via `rm` interface but from a flat filesystem point of view it should be a feature. What's you opinion? @arp7

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: Done in a9ef567103a13fd57d2c63197e7e356b447cafd7

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: Done in 5c1c08178a4d230ded059c8f11e82f19a375a862

smengcl:
Should be almost good to go. We can do one more round of checking using diff https://github.com/smengcl/hadoop-ozone/compare/0bee28acadfb2c358c6f008173e7eca6ed7fa23f...smengcl:HDDS-2840

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
xiaoyuyao: Let's follow up that on separate JIRA.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -0,0 +1,482 @@
...
+  }
+
+  /**
+   * Tests listStatus operation on root directory.
+   */
+  @Test
+  public void testListStatusOnLargeDirectory() throws Exception {
+    Path root = new Path("/" + volumeName + "/" + bucketName);
+    Set<String> paths = new TreeSet<>();
+    int numDirs = 5111;
```
xiaoyuyao: Consider reducing the numDirs as it will unnecessarily slow down the PR.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: Agree

xiaoyuyao:
Thanks @smengcl for the update. +1 on the latest change. The last comment can be addressed in followup JIRAs.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
smengcl: I'm not sure why o3fs test chose this random number. But basically `numDirs` only need to be larger than `LISTING_PAGE_SIZE` so it tests pagination. And BTW I believe integration tests are disabled in PR by default now. (They are not run in PR checks.)

Reduced in `0d7100604334f06fe304b143a3b906c9cbfc88dd`

xiaoyuyao:
+1, I will merge it shortly.

PR #418, opened 2020-01-07
HDDS-2716. ReadStateMachine failure should close down the ratis pipeline

swagle:
##  What changes were proposed in this pull request?
If readStateMachine call fails there is no way to recover for the follower and hence trigger pipeline close action. (Some sonar warning fixed along the way!)

https://issues.apache.org/jira/browse/HDDS-2716

## How was this patch tested?
Added a new end to end integration test to simulate failure and verify pipeline close.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineReportHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineReportHandler.java
@@ -110,8 +110,10 @@ public class PipelineReportHandler implements
     }
 
     if (pipeline.getPipelineState() == Pipeline.PipelineState.ALLOCATED) {
-      LOGGER.info("Pipeline {} {} reported by {}", pipeline.getFactor(),
-          pipeline.getId(), dn);
+      if (LOGGER.isDebugEnabled()) {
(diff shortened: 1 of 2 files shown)


bshashikant:
Thanks @swagle for working on this. The changes look good. As @swagle suggested, let's combine marking the stateMachine unhealthy as well as triggering pipeline close together for better clarity of code.

bshashikant:
@swagle , as per discussion with @mukul1987 , let's fix this in Ratis. ReadStateMachine failure should be propagated back to Ratis, and ratis should treat this a log failure and invoke Stamachine#notifyLogFailed() Api which will trigger pipeline close in Ozone.

swagle:
> @swagle , as per discussion with @mukul1987 , let's fix this in Ratis. ReadStateMachine failure should be propagated back to Ratis, and ratis should treat this a log failure and invoke Stamachine#notifyLogFailed() Api which will trigger pipeline close in Ozone.

Does that also apply to other places we mark the StateMachine UNHEALTHY? Namely: writeStateMachineData and applyTransaction?

elek:
Hi, what is the state of this PR? Do we need to merge it? @bshashikant ?

adoroszlai:
Sorry, I had to revert this, as it caused compilation error on `master`.  Some constants used by the test were recently removed in 4ce2796b1 and d497bf683.

As the patch only contains the test and some log level change, I assume the original issue is fixed in Ratis.  Please reopen the Jira issue if needed.

PR #419, opened 2020-01-08
HDDS-2853. NPE in OzoneContainer Start.

bharatviswa504:

Fix NPE in OzoneContainer Start in VersionEndPointTask.

https://issues.apache.org/jira/browse/HDDS-2853

## How was this patch tested?

Deployed the fix, and after this have not seen NPE and SCM is out of safe mode.


anuengineer:
> Thanks @bharatviswa504 for the fix. The problem is reproducible using TestBCSID integration test, too.

Any idea why this did not surface earlier ? or what is triggering this failure now ?

bharatviswa504:
> > Thanks @bharatviswa504 for the fix. The problem is reproducible using TestBCSID integration test, too.
> 
> Any idea why this did not surface earlier ? or what is triggering this failure now ?

I belive this is because previously in ContainerStateMachine initialization we don't need to call triggerHeartBeat to send pipeline report, a recent commit changed it. [code link](https://github.com/apache/hadoop-ozone/commit/41910d7eaab56dfc3ee55ac64722fd5584f920b1#diff-879b0be72f4f3af0570e050c406549cdR737)

PR #420, opened 2020-01-08
HDDS-2854. Add a config for the write chunk executor queue limit.

bshashikant:

Added a config for the write chunk executor blocking queue limit

https://issues.apache.org/jira/browse/HDDS-2854

## How was this patch tested?

Simple config. Already unit tests in place to test the config.


hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -203,6 +203,15 @@
       will use for writing chunks (60 by default).
     </description>
   </property>
+  <property>
+    <name>dfs.container.ratis.num.write.chunk.executor.queue.limit</name>
+    <value>60</value>
```
adoroszlai:
```suggestion
    <value>1024</value>
```

to be consistent with the code and the description.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
```diff
@@ -48,6 +48,12 @@
       = "dfs.container.ratis.num.write.chunk.threads";
   public static final int DFS_CONTAINER_RATIS_NUM_WRITE_CHUNK_THREADS_DEFAULT
       = 60;
+  public static final String
+          DFS_CONTAINER_RATIS_NUM_WRITE_CHUNK_EXECUTOR_QUEUE_LIMIT
+          = "dfs.container.ratis.num.write.chunk.executor.queue.limit";
```
adoroszlai: `num` stands for "number of ..." (eg. in `dfs.container.ratis.num.write.chunk.threads` it is "number of ... threads").  It seems to be unnecessary in this property name, "... queue limit" seems better than "number of ... queue limit".

lokeshj1703:
@bshashikant Thanks for working on this! The changes look good to me. I have no additional comments apart from attila's comments.

bshashikant:
Thanks @adoroszlai and @lokeshj1703 for the review. As per discussion with @mukul1987 , it was suggested that the data blocking queue size should be same as the number of max pending requests we can handle on the leader . So i removed the additional config and used the same value as that of max pending requests here. With proper tuning these 2 values, we can see batching happening during log sync.

PR #421, opened 2020-01-08
HDDS-2851. Some OM Ratis config properties missing from ozone-default.xml

bharatviswa504:

Removed this ozone.om.ratis.client.request.timeout.duration, as OM ratis client is removed as part of HDDS-1991

ozone.om.ratis.segment.preallocated.size is still needed, added it back. This is the OM Ratis Server configuration.

https://issues.apache.org/jira/browse/HDDS-2851

## How was this patch tested?

UT and IT should cover this change, as this caused TestOzoneConfigurationFields which is failing, now it is passed with this PR.

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1600,6 +1600,15 @@
   </property>
 
   <property>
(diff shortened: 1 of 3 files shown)


adoroszlai:
@hanishakoneru please take a look, can we merge this fix?

bharatviswa504:

PR #423, opened 2020-01-09
HDDS-2861. Support Freon progressbar in non-interactive environment

elek:

Freon tests use the `ProgressBar` to show the current progress of the Freon tests.

Progress bar use `\r` to print out a nice, interactive progress bar to the same line.

Unfortunately this doesn't work very well when the standard output is redirected to a file (or in a containerized environment). I would suggest to use a simplified method using log4j in non-interactive environments.

The easiest way to do is to use `System.console()` java method. If it returns with null we are in a non interactive environment. We can also increase the logging period from 1s to 10s in case of non-interactive environment (==long term testing).

https://issues.apache.org/jira/browse/HDDS-2861

## How was this patch tested?

```
export OZONE_REPLICATION_FACTOR=3
./run.sh -d
```

Interactive (expected output is the one line progress bar as earlier):

```
docker-compose exec scm bash
ozone freon ockg -n 10
```

Non-interactive (expected output is a log4j based testing):

```
export COMPOSE_FILE=docker-compose.yaml:freon-ockg.yaml
./run.sh -d
docker-compose logs freon
```

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ProgressBar.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ProgressBar.java
@@ -55,12 +55,24 @@ public class ProgressBar {
    */
   public ProgressBar(final PrintStream stream, final long maxValue,
                      final LongSupplier currentValue) {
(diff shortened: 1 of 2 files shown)


adoroszlai:
Thanks @elek for the update.

PR #428, opened 2020-01-10
HDDS-2868. Add ObjectID and UpdateID to OMKeyInfo.

hanishakoneru:

Similar to OMVolumeInfo and OMBucketInfo, we need objectID and updateID in OMKeyInfo as well to ensure that transactions are idempotent.
This Jira adds objectID and updateID to OMKeyInfo proto and sets these values when creating/ updating keys.

https://issues.apache.org/jira/browse/HDDS-2868

## How was this patch tested?

Integration and Acceptance tests. Follow-up Jiras for key operation idempotence will add more tests.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
@@ -52,6 +52,9 @@ public final class OmKeyInfo extends WithMetadata {
   private HddsProtos.ReplicationType type;
   private HddsProtos.ReplicationFactor factor;
   private FileEncryptionInfo encInfo;
+  private long objectID;
+  private long updateID;
+
   /**
    * ACL Information.
    */
@@ -64,7 +67,8 @@ public final class OmKeyInfo extends WithMetadata {
       HddsProtos.ReplicationType type,
       HddsProtos.ReplicationFactor factor,
       Map<String, String> metadata,
-      FileEncryptionInfo encInfo, List<OzoneAcl> acls) {
+      FileEncryptionInfo encInfo, List<OzoneAcl> acls,
+      long objectID, long updateID) {
     this.volumeName = volumeName;
     this.bucketName = bucketName;
     this.keyName = keyName;
@@ -88,6 +92,8 @@ public final class OmKeyInfo extends WithMetadata {
     this.metadata = metadata;
     this.encInfo = encInfo;
     this.acls = acls;
+    this.objectID = objectID;
+    this.updateID = updateID;
   }
 
   public String getVolumeName() {
@@ -133,6 +139,47 @@ public final class OmKeyInfo extends WithMetadata {
 
   public void updateModifcationTime() {
     this.modificationTime = Time.monotonicNow();
+  }
+
+  /**
+   * Set the Object ID. If this value is already set then this function throws.
+   * There is a reason why we cannot use the final here. The OMKeyInfo is
+   * deserialized from the protobuf in many places in code. We need to set
+   * this object ID, after it is deserialized.
+   *
+   * @param obId - long
+   */
+  public void setObjectID(long obId) {
+    if(this.objectID != 0) {
+      throw new UnsupportedOperationException("Attempt to modify object ID " +
+          "which is not zero. Current Object ID is " + this.objectID);
+    }
+    this.objectID = obId;
+  }
+
+  /**
+   * Sets the update ID. For each modification of this object, we will set
+   * this to a value greater than the current value.
+   * @param updateID  long
+   */
+  public void setUpdateID(long updateID) {
+    this.updateID = updateID;
+  }
+
+  /**
+   * Returns objectID.
+   * @return long
+   */
+  public long getObjectID() {
+    return objectID;
+  }
+
+  /**
+   * Returns updateID.
+   * @return long
+   */
+  public long getUpdateID() {
+    return updateID;
   }
 
   /**
@@ -265,6 +312,8 @@ public final class OmKeyInfo extends WithMetadata {
     private Map<String, String> metadata;
     private FileEncryptionInfo encInfo;
     private List<OzoneAcl> acls;
+    private long objectID;
+    private long updateID;
 
     public Builder() {
       this.metadata = new HashMap<>();
@@ -357,11 +406,21 @@ public final class OmKeyInfo extends WithMetadata {
       return this;
     }
 
+    public Builder setObjectID(long obId) {
+      this.objectID = obId;
+      return this;
+    }
+
+    public Builder setUpdateID(long id) {
+      this.updateID = id;
+      return this;
+    }
+
     public OmKeyInfo build() {
       return new OmKeyInfo(
           volumeName, bucketName, keyName, omKeyLocationInfoGroups,
           dataSize, creationTime, modificationTime, type, factor, metadata,
-          encInfo, acls);
+          encInfo, acls, objectID, updateID);
     }
   }
 
@@ -382,7 +441,9 @@ public final class OmKeyInfo extends WithMetadata {
         .setCreationTime(creationTime)
         .setModificationTime(modificationTime)
         .addAllMetadata(KeyValueUtil.toProtobuf(metadata))
-        .addAllAcls(OzoneAclUtil.toProtobuf(acls));
+        .addAllAcls(OzoneAclUtil.toProtobuf(acls))
+        .setObjectID(objectID)
+        .setUpdateID(updateID);
     if (encInfo != null) {
       kb.setFileEncryptionInfo(OMPBHelper.convert(encInfo));
     }
@@ -390,7 +451,7 @@ public final class OmKeyInfo extends WithMetadata {
   }
 
   public static OmKeyInfo getFromProtobuf(KeyInfo keyInfo) {
-    return new OmKeyInfo.Builder()
+    Builder builder = new Builder()
         .setVolumeName(keyInfo.getVolumeName())
         .setBucketName(keyInfo.getBucketName())
         .setKeyName(keyInfo.getKeyName())
@@ -404,9 +465,15 @@ public final class OmKeyInfo extends WithMetadata {
         .setReplicationFactor(keyInfo.getFactor())
         .addAllMetadata(KeyValueUtil.getFromProtobuf(keyInfo.getMetadataList()))
         .setFileEncryptionInfo(keyInfo.hasFileEncryptionInfo() ?
-            OMPBHelper.convert(keyInfo.getFileEncryptionInfo()): null)
-        .setAcls(OzoneAclUtil.fromProtobuf(keyInfo.getAclsList()))
-        .build();
+            OMPBHelper.convert(keyInfo.getFileEncryptionInfo()) : null)
+        .setAcls(OzoneAclUtil.fromProtobuf(keyInfo.getAclsList()));
+    if (keyInfo.hasObjectID()) {
+      builder.setObjectID(keyInfo.getObjectID());
+    }
+    if (keyInfo.hasUpdateID()) {
+      builder.setUpdateID(keyInfo.getUpdateID());
+    }
+    return builder.build();
   }
 
   @Override
@@ -429,7 +496,9 @@ public final class OmKeyInfo extends WithMetadata {
         type == omKeyInfo.type &&
         factor == omKeyInfo.factor &&
         Objects.equals(metadata, omKeyInfo.metadata) &&
-        Objects.equals(acls, omKeyInfo.acls);
+        Objects.equals(acls, omKeyInfo.acls) &&
+        objectID == omKeyInfo.objectID &&
+        updateID == omKeyInfo.updateID;
   }
 
   @Override
@@ -450,7 +519,9 @@ public final class OmKeyInfo extends WithMetadata {
         .setDataSize(dataSize)
         .setReplicationType(type)
         .setReplicationFactor(factor)
-        .setFileEncryptionInfo(encInfo);
+        .setFileEncryptionInfo(encInfo)
+        .setObjectID(objectID)
+        .setUpdateID(updateID);
 
     keyLocationVersions.forEach(keyLocationVersion -> {
       List<OmKeyLocationInfo> keyLocationInfos = new ArrayList<>();
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
@@ -142,28 +142,15 @@ public final class OmKeyInfo extends WithMetadata {
   }
 
   /**
-   * Set the Object ID. If this value is already set then this function throws.
-   * There is a reason why we cannot use the final here. The OMKeyInfo is
-   * deserialized from the protobuf in many places in code. We need to set
-   * this object ID, after it is deserialized.
-   *
-   * @param obId - long
-   */
-  public void setObjectID(long obId) {
-    if(this.objectID != 0) {
-      throw new UnsupportedOperationException("Attempt to modify object ID " +
-          "which is not zero. Current Object ID is " + this.objectID);
-    }
-    this.objectID = obId;
-  }
-
-  /**
    * Sets the update ID. For each modification of this object, we will set
    * this to a value greater than the current value.
-   * @param updateID  long
+   * @param updateId  long
    */
-  public void setUpdateID(long updateID) {
-    this.updateID = updateID;
+  public void setUpdateID(long updateId) {
+    Preconditions.checkArgument(updateId > this.updateID,
+        "Trying to set updateID to a value ({}) which is less than the " +
+            "current value ({}) for ()", updateId, this.updateID, this);
+    this.updateID = updateId;
   }
 
   /**
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
@@ -35,21 +35,66 @@ public class OmMultipartKeyInfo {
   private final ReplicationType replicationType;
   private final ReplicationFactor replicationFactor;
   private TreeMap<Integer, PartKeyInfo> partKeyInfoList;
+  private long objectID;
+  private long updateID;
 
   /**
    * Construct OmMultipartKeyInfo object which holds multipart upload
    * information for a key.
    */
   public OmMultipartKeyInfo(String id, long creationTime,
-                            ReplicationType replicationType,
-                            ReplicationFactor replicationFactor,
-                            Map<Integer, PartKeyInfo> list) {
+      ReplicationType replicationType, ReplicationFactor replicationFactor,
+      Map<Integer, PartKeyInfo> list, long objectID, long updateID) {
     this.uploadID = id;
     this.creationTime = creationTime;
     this.replicationType = replicationType;
     this.replicationFactor = replicationFactor;
     this.partKeyInfoList = new TreeMap<>(list);
+    this.objectID = objectID;
+    this.updateID = updateID;
   }
+
+  /**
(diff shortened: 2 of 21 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
```diff
@@ -349,7 +349,8 @@ protected OmKeyInfo createKeyInfo(@Nonnull KeyArgs keyArgs,
       @Nonnull HddsProtos.ReplicationType type, long size,
       @Nullable FileEncryptionInfo encInfo,
       @Nonnull PrefixManager prefixManager,
-      @Nullable OmBucketInfo omBucketInfo) {
+      @Nullable OmBucketInfo omBucketInfo,
+      @Nonnull long transactionLogIndex) {
```
bharatviswa504: NIT: primitive type, we dont need @NonNull

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
```diff
@@ -416,12 +419,14 @@ protected OmKeyInfo prepareKeyInfo(
       @Nonnull KeyArgs keyArgs, @Nonnull String dbKeyName, long size,
       @Nonnull List<OmKeyLocationInfo> locations,
       @Nullable FileEncryptionInfo encInfo,
-      @Nonnull PrefixManager prefixManager, @Nullable OmBucketInfo omBucketInfo)
+      @Nonnull PrefixManager prefixManager,
+      @Nullable OmBucketInfo omBucketInfo,
+      @Nonnull long transactionLogIndex)
```
bharatviswa504: NIT: primitive type, we dont need @nonNull

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3InitiateMultipartUploadRequest.java
```diff
@@ -158,9 +158,10 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           .setOmKeyLocationInfos(Collections.singletonList(
               new OmKeyLocationInfoGroup(0, new ArrayList<>())))
           .setAcls(OzoneAclUtil.fromProtobuf(keyArgs.getAclsList()))
+          .setObjectID(transactionLogIndex)
```
bharatviswa504: Missing to add objectID and updateID in multipartKeyInfo

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadAbortRequest.java
```diff
@@ -120,6 +124,21 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
         multipartKeyInfo = omMetadataManager
             .getMultipartInfoTable().get(multipartKey);
 
+        // Set updateID to current transactionLogIndex for all parts
+        TreeMap<Integer, PartKeyInfo> partKeyInfoMap = multipartKeyInfo
+            .getPartKeyInfoMap();
+        for (Map.Entry<Integer, PartKeyInfo> partKeyInfoEntry :
+            partKeyInfoMap.entrySet()) {
+          PartKeyInfo partKeyInfo = partKeyInfoEntry.getValue();
+          OmKeyInfo currentKeyPartInfo = OmKeyInfo.getFromProtobuf(
```
bharatviswa504: Not understood the reason of setting updateID here, as below we are setting Optional.absent(), so it  means entry is deleted from DB.

hanishakoneru: Yes, but this entry is added to DeletedTable. So I thought we should keep a record of the transaction which deleted the Key.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
hanishakoneru: done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
hanishakoneru: done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3InitiateMultipartUploadRequest.java
hanishakoneru: Thanks for catching this. Updated the patch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -943,9 +943,13 @@ private OmMultipartInfo createMultipartInfo(OmKeyArgs keyArgs,
 
       long currentTime = Time.now();
       Map<Integer, PartKeyInfo> partKeyInfoMap = new HashMap<>();
-      OmMultipartKeyInfo multipartKeyInfo = new OmMultipartKeyInfo(
-          multipartUploadID, currentTime, keyArgs.getType(),
-          keyArgs.getFactor(), partKeyInfoMap);
+      OmMultipartKeyInfo multipartKeyInfo = new OmMultipartKeyInfo.Builder()
```
arp7: Is this code still used? @bharatviswa504

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -210,7 +210,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
   private OmKeyInfo createDirectoryKeyInfo(OzoneManager ozoneManager,
       OmBucketInfo omBucketInfo, String volumeName, String bucketName,
-      String keyName, KeyArgs keyArgs)
+      String keyName, KeyArgs keyArgs, long transactionLogIndex)
```
arp7: Is the transaction log index reset when the Ratis term changes?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
```diff
@@ -52,6 +52,9 @@
   private HddsProtos.ReplicationType type;
   private HddsProtos.ReplicationFactor factor;
   private FileEncryptionInfo encInfo;
+  private long objectID;
+  private long updateID;
+
```
arp7: Also what happens for deployed clusters where the persisted keys have no object Id or update ID? What do we initialize it to?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
```diff
@@ -112,6 +112,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
     OMMetadataManager omMetadataManager = ozoneManager.getMetadataManager();
     boolean acquiredLock = false;
     OMClientResponse omClientResponse = null;
+    boolean success = true;
```
arp7: It may be safer to initialize success to false, and set it to true at the end of the try block when all operations are complete.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
```diff
@@ -83,6 +128,72 @@ public ReplicationFactor getReplicationFactor() {
     return replicationFactor;
   }
 
+  /**
+   * Builder of OmKeyInfo.
```
bharatviswa504: Minor NIT: OmKeyInfo -> OmMultipartKeyInfo

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
bharatviswa504: No @arp7. For write requests new OM HA code is used.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
```diff
@@ -126,6 +126,8 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
       if (omKeyInfo == null) {
         throw new OMException("Key not found", KEY_NOT_FOUND);
       }
+      // Set the UpdateID to current transactionLogIndex
+      omKeyInfo.setUpdateID(transactionLogIndex);
```
bharatviswa504: I think, we don't need to set updateID, as we delete this entry from KeyTable anyway. (As we set Optional.absent() below, which will be deleted from OM DB eventually when double buffer picks). Or is there any scenario by setting this helps.

Example: Create key Key1 - 1, Delete Key Key1 - 2 Create Key Key1 (again) - 3 Now replay happening from 1, as now in KeyTable it will have updateID 3, we think it as replay just by checking OmKeyInfo.getUpdateID, similar to Bucket/Volume delete.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadAbortRequest.java
```diff
@@ -120,6 +124,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
         multipartKeyInfo = omMetadataManager
             .getMultipartInfoTable().get(multipartKey);
 
+        multipartKeyInfo.setUpdateID(transactionLogIndex);
```
bharatviswa504: Same as above, I think we don't need to set updateID here, as anyway we are removing entry from multipartInfo table. Is there is any use case that will help in the detection of replay by setting here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
```diff
@@ -155,6 +156,9 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
       //Set modification time
       fromKeyValue.setModificationTime(renameKeyArgs.getModificationTime());
 
+      // Set the UpdateID to current transactionLogIndex
+      fromKeyValue.setUpdateID(transactionLogIndex);
```
bharatviswa504: As this is like new key is getting created, so do we need to set objectID also here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCompleteRequest.java
```diff
@@ -229,17 +229,33 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
                   Collections.singletonList(keyLocationInfoGroup))
-              .setAcls(OzoneAclUtil.fromProtobuf(keyArgs.getAclsList()))
-              .build();
+              .setAcls(OzoneAclUtil.fromProtobuf(keyArgs.getAclsList()));
+          // Check if db entry has ObjectID. This check is required because
+          // it is possible that between multipart key uploads and complete,
+          // we had an upgrade.
+          // TODO: If on upgrades we do not consider the non-completed multipart
+          //  uploads, then this check can be removed.
+          if (dbOpenKeyInfo.getObjectID() != 0) {
```
bharatviswa504: As discussed offline, for this we can set ObjectID, which we have in the dbOpenKeyInfo.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
```diff
@@ -35,22 +35,67 @@
...
 
+  /**
+   * Set the Object ID. If this value is already set then this function throws.
+   * There is a reason why we cannot use the final here. The
+   * OMMultipartKeyInfo is deserialized from the protobuf in many places in
+   * code. We need to set this object ID, after it is deserialized.
+   *
+   * @param obId - long
+   */
+  public void setObjectID(long obId) {
```
bharatviswa504: This method is not called any where, as we use builder methods.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
```diff
@@ -135,6 +141,47 @@ public void updateModifcationTime() {
...
+  /**
+   * Set the Object ID. If this value is already set then this function throws.
+   * There is a reason why we cannot use the final here. The OMKeyInfo is
+   * deserialized from the protobuf in many places in code. We need to set
+   * this object ID, after it is deserialized.
+   *
+   * @param obId - long
+   */
+  public void setObjectID(long obId) {
+    if(this.objectID != 0) {
```
bharatviswa504: This method is not called anywhere, as we use builder methods.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
```diff
@@ -52,6 +52,9 @@
   private HddsProtos.ReplicationType type;
   private HddsProtos.ReplicationFactor factor;
   private FileEncryptionInfo encInfo;
+  private long objectID;
```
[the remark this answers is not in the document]
hanishakoneru: If there are any old objects in the DB, then those will have objectId and updateId as 0. It is not invalid but if it is 0, then we cannot determine if a corresponding transaction is a replay.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
hanishakoneru: No Arpit. Transaction log index is the ratis log index. It is monotonically increasing always.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
hanishakoneru: It will default to 0 as we use long primitive type here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
hanishakoneru: Deleted Keys are put in DeletedKey Table and then purged when we get KeyPurgeRequest. As the purge request could also be a replay, we should only delete the keys which were deleted before the purge request. For this check, we need to set the updateID here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
hanishakoneru: I think either way should be ok. I thought since its the same key being renamed, we could have the same objectID. But I am ok changing it to current transactionLogIndex. Let me know what you think.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
hanishakoneru: Removed

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
hanishakoneru: Removed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCompleteRequest.java
hanishakoneru: Done.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
hanishakoneru: Done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
hanishakoneru: Done.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
```diff
@@ -135,6 +141,47 @@ public void updateModifcationTime() {
...
+    this.objectID = obId;
+  }
+
+  /**
+   * Sets the update ID. For each modification of this object, we will set
+   * this to a value greater than the current value.
+   * @param updateID  long
+   */
+  public void setUpdateID(long updateID) {
+    this.updateID = updateID;
```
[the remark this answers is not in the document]
hanishakoneru: Added the check.

hanishakoneru:
Thank you @bharatviswa504 and @arp7 for the reviews. I have addressed the comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
arp7: Should we file a separate jira to delete this code now?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
bharatviswa504: As discussed offline, we can change this when we use them in finding a transaction is replay or not.

bharatviswa504:
Thank You @hanishakoneru for the contribution and @arp7 for the review.

hanishakoneru:
Thank you @bharatviswa504 and @arp7 for the reviews and merging the PR.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
prashantpogde: Can transactionLogIndex be used as a unique objectId over time ?

PR #430, opened 2020-01-10
HDDS-2817. Fix listing buckets for setting --prefix equal to bucket name

cxorm:
This PR was created to let `--prefix` work without excluding bucket name that is equal to `prefix` when listing buckets.

- Fix.
In this [line](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java#L664), we exclude the bucket name that is equal to `--prefix` when listing buckets for now.
We updated the `>` to `>=`, and this would include the bucket name that is equal to `--prefix` when listing buckets.

https://issues.apache.org/jira/browse/HDDS-2817

## How was this patch tested?
UT updated.
[Github action passed](https://github.com/cxorm/hadoop-ozone/runs/376239895).

### DIFF
--- a/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestOmMetadataManager.java
+++ b/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestOmMetadataManager.java
@@ -149,7 +149,9 @@ public class TestOmMetadataManager {
 
     TreeSet<String> volumeABucketsPrefixWithOzoneOwner = new TreeSet<>();
     TreeSet<String> volumeABucketsPrefixWithHadoopOwner = new TreeSet<>();
-    for (int i=1; i<= 100; i++) {
+    volumeABucketsPrefixWithOzoneOwner.add(prefixBucketNameWithOzoneOwner);
+    addBucketsToCache(volumeName1, prefixBucketNameWithOzoneOwner);
+    for (int i = 1; i < 100; i++) {
       if (i % 2 == 0) {
         volumeABucketsPrefixWithOzoneOwner.add(
             prefixBucketNameWithOzoneOwner + i);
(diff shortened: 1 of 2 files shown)


cxorm:
Would you take a look on this PR if you have time, @xiaoyuyao

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestOmMetadataManager.java
```diff
@@ -83,7 +85,9 @@ public void testListBuckets() throws Exception {
     TreeSet<String> volumeBBucketsPrefixWithOzoneOwner = new TreeSet<>();
     TreeSet<String> volumeBBucketsPrefixWithHadoopOwner = new TreeSet<>();
     TestOMRequestUtils.addVolumeToDB(volumeName2, omMetadataManager);
-    for (int i=1; i<= 100; i++) {
+    volumeABucketsPrefixWithOzoneOwner.add(prefixBucketNameWithOzoneOwner);
```
smengcl: Should this be `volumeBBucketsPrefixWithOzoneOwner.add` instead? Otherwise this line is the same as line 70, adding the same element to a set won't do anything. :)

elek:
/pending Pending on the author to answer comment

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestOmMetadataManager.java
cxorm: Thank you @smengcl for address this line.
Fixed.

cxorm:
> The fix itself is straight-forward and lgtm.
> 
> The addition to the existing UT looks valid, but it's a bit obscure IMO. It'd be great to have some extra comments in the code.

Thanks @smengcl for the review, I add comments addressed for the test.

smengcl:
Merged to master. Thanks @cxorm for the contribution!

cxorm:
Thank you @smengcl for the review and commit.

PR #431, opened 2020-01-10
HDDS-2846. Handle Datanode Pipeline & Container Reports reports in Recon.

avijayanhwx:
Recon should be able to process Container and Pipeline reports from registered Datanodes. It still wont be able to understand new containers and pipelines though. Along with HDDS-2850, HDDS-2852 and HDDS-2869, Recon will be a fully functional passive SCM. 
In addition to the above, some refactoring work also has been done.

https://issues.apache.org/jira/browse/HDDS-2846

## How was this patch tested?
Manually tested on docker.
Created key (container) on ozone cluster, and verified Recon gets the container info in Datanode heartbeat.

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconStorageContainerManagerFacade.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconStorageContainerManagerFacade.java
@@ -18,6 +18,7 @@
 
 package org.apache.hadoop.ozone.recon.scm;
 
+import static org.apache.hadoop.hdds.recon.ReconConfigKeys.RECON_SCM_CONFIG_PREFIX;
 import static org.apache.hadoop.hdds.scm.server.StorageContainerManager.buildRpcServerStartMessage;
 
 import java.io.IOException;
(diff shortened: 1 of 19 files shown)


avijayanhwx:
@anuengineer  / @swagle Please review.

avijayanhwx:
@nandakumar131 Please review.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconStorageContainerManagerFacade.java
```diff
@@ -27,52 +28,99 @@
...
 import org.apache.hadoop.hdds.server.events.EventQueue;
+import org.apache.hadoop.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Recon's 'lite' version of SCM.
  */
-public class ReconStorageContainerManager
+public class ReconStorageContainerManagerFacade
```
swagle: Minor nit: Should drop the Facade from the class name, a facade merely provides access to underlying subsystems minus the management functions which this class actually does in addition to abstracting out access.

avijayanhwx: Thanks for the review @swagle. I will address this in the next patch (HDDS-2869).

PR #432, opened 2020-01-10
HDDS-2832. Fix listing buckets for setting --start with last bucket

cxorm:
Cause listing buckets with `--start` means that we would get buckets after the `--start`, we should get empty result when we set `--start` with last bucket.

But we get all buckets for now, this is caused by this [method](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java#L298#L305).

When we set `--start` with last bucket, [`getNextListOfBuckets(prevBucket)`](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java#L294) would be empty list, so the first call of [`hasNext()`](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java#L298#L305) would enter the `if-block` and `currentValue` is `null`.

This situation results in calling `getNextListOfBuckets(null)`, and [`currentIterator`](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java#L300#L302) would be iterator of all buckets with this length size, so we get all buckets for now.

#### Proposed fix.
The `currentValue` in [`if-block`](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java#L299#L303) would be `null` if and only if first call [`hasNext()`](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java#L298#L305) of empty list, so`currentValue` is tested ahead by this fix.

https://issues.apache.org/jira/browse/HDDS-2832

## How was this patch tested?
[Github action passed](https://github.com/cxorm/hadoop-ozone/runs/376330188).

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java
@@ -296,9 +296,8 @@ public class OzoneVolume extends WithMetadata {
 
     @Override
     public boolean hasNext() {
-      if(!currentIterator.hasNext()) {
-        currentIterator = getNextListOfBuckets(
-            currentValue != null ? currentValue.getName() : null)
+      if (!currentIterator.hasNext() && currentValue != null) {
+        currentIterator = getNextListOfBuckets(currentValue.getName())
             .iterator();
       }
       return currentIterator.hasNext();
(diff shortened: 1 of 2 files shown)


cxorm:
Test would be added in integration-test soon.

cxorm:
Thanks @elek for looking on it.

The [Error](https://github.com/apache/hadoop-ozone/pull/432/checks?check_run_id=495460038 ) here seems not caused by this patch.
I have tested it with other [branch](https://github.com/cxorm/hadoop-ozone/runs/495427388), and it passed.

elek:
Finally, full green build: Thanks the contribution, I am merging it to the master.

cxorm:
Thank you @elek for the review and commit.

PR #436, opened 2020-01-11
HDDS-2875. Add a config in ozone to tune max outstanding requests in …

bharatviswa504:
…raft client.

Add a config to tune the config value of "raft.client.async.outstanding-requests.max" config in raft client. 

There is a property scm.container.client.max.outstanding.requests but this is used to set outStandingAppendsMax, but the config of this is mentioned as the "Controls the maximum number of outstanding async requests that can be handled by the Standalone as well as Ratis client". So, I removed this from ScmClientConfig, and created a new class for this property with prefix dfs.ratis.client to be consistent with other ratis client properties. And used the same for outStandingAppendsMax as before, not sure this is the correct way, but to keep same with current code, used it.

https://issues.apache.org/jira/browse/HDDS-2875

## How was this patch tested?

UT and acceptance test run.

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientManager.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientManager.java
@@ -50,6 +50,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.apache.hadoop.hdds.conf.ConfigTag.CLIENT;
 import static org.apache.hadoop.hdds.conf.ConfigTag.OZONE;
 import static org.apache.hadoop.hdds.conf.ConfigTag.PERFORMANCE;
 
@@ -322,15 +323,6 @@ public class XceiverClientManager implements Closeable {
     )
     private long staleThreshold;
 
-    @Config(key = "max.outstanding.requests",
-        defaultValue = "100",
-        tags = {OZONE, PERFORMANCE},
-        description =
-            "Controls the maximum number of outstanding async requests that can"
-                + " be handled by the Standalone as well as Ratis client."
-    )
-    private int maxOutstandingRequests;
-
     public long getStaleThreshold(TimeUnit unit) {
       return unit.convert(staleThreshold, MILLISECONDS);
     }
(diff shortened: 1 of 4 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
```diff
@@ -206,6 +206,10 @@ static RaftClient newRaftClient(RpcType rpcType, RaftPeerId leader,
     GrpcConfigKeys.OutputStream.setOutstandingAppendsMax(properties,
```
lokeshj1703: I think we can remove this setter. This is used by GrpcOutputStream in ratis which is not used by ozone.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientManager.java
```diff
@@ -345,10 +337,30 @@ public void setMaxSize(int maxSize) {
...
 
+  }
+
+  /**
+   * Configuration for ratis client.
+   */
+  @ConfigGroup(prefix = "dfs.ratis.client")
+  public class DFSRatisClientConfig {
+
+    @Config(key = "async.max.outstanding.requests",
```
lokeshj1703: I dont think we have a corresponding config in ozone-default.xml.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
bharatviswa504: Done

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientManager.java
bharatviswa504: With new-style Java config, we don't need to add, it generates an ozone-default-generated.xml. In that file, I can see this new config.

```
[root@ozone-test-1 ex]# cat ozone-default-generated.xml | grep -A4 "dfs.ratis.client.async"
    <name>dfs.ratis.client.async.max.outstanding.requests</name>
    <value>64</value>
    <description>Controls the maximum number of outstanding async requests that can be handled by the Standalone as well as Ratis client.</description>
    <tag>OZONE, CLIENT, PERFORMANCE</tag>
  </property>
```

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
bharatviswa504: Done

bharatviswa504:
@lokeshj1703 Thank You for the review. Addressed review comments and acceptance test failures.

lokeshj1703:
@bharatviswa504 Thanks for updating the PR! The changes look good to me. I have committed it to master branch.

PR #441, opened 2020-01-15
HDDS-2889. Make DBStore and RDBStore more commons.

maobaolong:

As a user of RDBStore and TypedTable, i only use these class to maintain my rocksdb, i use hadoop-hdds-common as a maven dependency.

But during my progress running, i find a folder named `om.db.checkpoints`, it is terrible, i do not know om, i only want to use a common typedTable which help me maintains rocksdb.

So we should rename it, remove the concept of OM.

https://issues.apache.org/jira/browse/HDDS-2889

## How was this patch tested?

Only change the name of checkpoint and comments, no need to test.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/db/DBStore.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/db/DBStore.java
@@ -159,7 +159,7 @@ public interface DBStore extends AutoCloseable {
   void commitBatchOperation(BatchOperation operation) throws IOException;
 
   /**
-   * Get current snapshot of OM DB store as an artifact stored on
+   * Get current snapshot of DB store as an artifact stored on
    * the local filesystem.
    * @return An object that encapsulates the checkpoint information along with
    * location.
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
@@ -19,8 +19,6 @@
 
 package org.apache.hadoop.hdds.utils.db;
 
-import static org.apache.hadoop.ozone.OzoneConsts.OM_DB_CHECKPOINTS_DIR_NAME;
-
 import javax.management.ObjectName;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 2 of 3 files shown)


avijayanhwx:
@maobaolong Thanks for working on this. 
I am thinking if we can create the checkpoints directory only if a checkpoint is requested? This will make sure we don't create this directory for any other component than OM.

maobaolong:
@avijayanhwx  Yeah, an option configuration item is needed.

avijayanhwx:
@maobaolong I believe it can be done without adding a config as well. Can we move the creation of the checkpoint parent directory inside **RDBCheckpointManager#createCheckpoint** ? That way, unless someone requests a checkpoint, the directory will not be created. Also, this is not really a "user" configurable item, It is just a code improvement. Please let me know what you think.

maobaolong:
@avijayanhwx I think we should neither bring the OM concept into `RDBStore` nor ` RDBCheckpointManager`, and we should let the user of `RDBStore` determine if checkpoint is enabled. Yes, i do agree that it maybe not a strong requests for let checkpoint folder to be a configurable item, we just give a hard code name for it, for example "db.checkpoint".

Because OM is a concept of Ozone, and `RDBStore` and ` RDBCheckpointManager` is a common components which is dependency by our program now, as a user of `RDBStore`, but not a user of Ozone, in fact, i do not need checkpoint at all in our scenario, but i think there must someone need checkpoint, and nobody want to specify the location or name of checkpoint folder now.

So, i think we should create another ticket https://issues.apache.org/jira/browse/HDDS-2907 for track our discussion？For `HDDS-2889`, we just remove the OM concept.

maobaolong:
@elek  I have created another jira https://issues.apache.org/jira/browse/HDDS-2907.

If one RDBStore have only one RDBCheckpointManager, so there shouldn't have a name collision

If one RDBStore have more than one RDBCheckpointManager, we can give a index string as a suffix of the checkpoint folder name.

avijayanhwx:
@maobaolong Maybe there was a misunderstanding. I did not mention anything about OM in DB checkpoint class. My point was that we can create the checkpoint directory inside the RDBCheckpoint manager, if someone requests a checkpoint. That way, we don't create empty directories in all instances of RDBStore being used by your tool or by ozone. I am perfectly OK to handle this in HDDS-2907. Thanks for filing the issue. +1 for this change.

maobaolong:
@avijayanhwx Sorry for the misunderstanding and your clarification, now i understand you and catch you key point, and i'm agree to move the checkpoint directory configuration and creation into RDBCheckpoint.

PR #445, opened 2020-01-15
HDDS-2849. OM and SCM Web-server report HTTP 404 error when accessing…

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-2849

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/BaseHttpServer.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/BaseHttpServer.java
@@ -17,6 +17,7 @@
 
 package org.apache.hadoop.hdds.server;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hdds.HddsConfigKeys;
 import org.apache.hadoop.hdfs.DFSConfigKeys;
@@ -27,6 +28,7 @@ import org.apache.hadoop.http.HttpServer2;
 import org.apache.hadoop.metrics2.lib.DefaultMetricsSystem;
 import org.apache.hadoop.net.NetUtils;
 
+import org.apache.hadoop.ozone.OzoneConfigKeys;
 import org.eclipse.jetty.webapp.WebAppContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/BaseHttpServer.java
```diff
@@ -114,8 +118,13 @@ public BaseHttpServer(Configuration conf, String name) throws IOException {
                 + "production!");
         httpServer.addServlet("profile", "/prof", ProfileServlet.class);
       }
-    }
 
+      String baseDir = conf.get(OzoneConfigKeys.OZONE_HTTP_BASEDIR);
+      if (!StringUtils.isEmpty(baseDir)) {
+        httpServer.getWebAppContext().setAttribute(JETTY_BASETMPDIR, baseDir);
```
xiaoyuyao: Thanks @ChenSammi  for reporting the issue and working on the fix. 

According to jetty document [here](https://www.eclipse.org/jetty/documentation/current/ref-temporary-directories.html): 

"By default, Jetty will create this directory inside the directory named by the java.io.tmpdir System property. You can instruct Jetty to use a different parent directory by setting the context attribute org.eclipse.jetty.webapp.basetempdir to the name of the desired parent directory. The directory named by this attribute must exist and be writeable."

Given the behavior without specifying basetmpdir explicitly, jetty will automatically create the dir under java.io.tempdir, which should work. Can you elaborate on the proposed fix?

elek: From an offline discussion I learned the problem: In some environments the `/tmp` cleaned up regularly (For me it's just at the restart but in some environments it's cleaned up more frequently). In these environments the web UI is not available after a while (when jetty temporary files are deleted).
 
I am fine with this patch. We can remove this configuration setting long-term as we are planning to unify all the webui under the recon. Until that we can make it configurable...

elek: Let me know if you are happy with this approach @xiaoyuyao and I will commit the patch...

xiaoyuyao: Sounds good to me. Agree that could be a problem when /tmp is cleaned up more frequently.

PR #446, opened 2020-01-15
HDDS-2859. Hugo error should be propagated to build

adoroszlai:

Make failed `hugo` execution cause build failure, similar to compile error.  The goal is to be able to catch doc syntax errors in CI.

https://issues.apache.org/jira/browse/HDDS-2859

## How was this patch tested?

Verified that the build is failed with the current error:

```
$ mvn -pl :hadoop-hdds-docs clean package
...
[INFO] --- exec-maven-plugin:1.6.0:exec (default) @ hadoop-hdds-docs ---
Building sites … Total in 22 ms
Error: Error building site: "hadoop-hdds/docs/content/start/_index.zh.md:39:119": unterminated quoted string in shortcode parameter-argument: 'start
'
...
[ERROR] Command execution failed.
org.apache.commons.exec.ExecuteException: Process exited with an error: 255 (Exit value: 255)
...
[INFO] BUILD FAILURE
```

and that it is successful if syntax error is fixed:

```
...
[INFO] --- exec-maven-plugin:1.6.0:exec (default) @ hadoop-hdds-docs ---
Building sites …
                   | EN | ZH
+------------------+----+----+
  Pages            | 57 | 19
  Paginator pages  |  0 |  0
  Non-page files   | 10 |  8
  Static files     | 19 | 19
  Processed images |  0 |  0
  Aliases          |  1 |  0
  Sitemaps         |  2 |  1
  Cleaned          |  0 |  0

Total in 95 ms
...
[INFO] BUILD SUCCESS
```

CI run with updated Ozone build image with Hugo pre-installed:

https://github.com/adoroszlai/hadoop-ozone/runs/391572329


xiaoyuyao:
@adoroszlai thanks for fixing this. I took a look at the change so the main change is contained in the apache/ozone-build image itself? Can you elaborate?

adoroszlai:
Thanks @xiaoyuyao for taking a look at this. 

> so the main change is contained in the apache/ozone-build image itself?

No, the fix itself is really in 4c70b7ce0c1daa004a29ce6a84a92287767c2c6b.  `set -e` makes bash exit the script on any error, and thus propagate the result of `hugo` in case it fails.  Without this change the script ignores the error and exits with the result of the last command: `cd -` (which is usually success).

After testing the fix locally and pushing to my fork, I realized that the image used by CI checks does not even have Hugo installed.  (Which the script treats as OK, to avoid forcing developers to install it for Maven to work).  So the image change (apache/hadoop-docker-ozone#10) was only necessary to pre-install Hugo.

xiaoyuyao:
Thanks @adoroszlai  for the details, LGTM. +1.

adoroszlai:
Thanks @xiaoyuyao for reviewing and merging this.

PR #448, opened 2020-01-16
HDDS-2870. Handle replay of KeyCreate requests.

hanishakoneru:

To ensure that key creation operation is idempotent, compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

OMKeyCreateRequest and OMFileCreateRequest are made idempotent in this Jira.

https://issues.apache.org/jira/browse/HDDS-2870

## How was this patch tested?

Unit tests to verify replayed transactions return expected response.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
@@ -30,6 +30,7 @@ import javax.annotation.Nonnull;
 
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
+import org.apache.hadoop.ozone.om.exceptions.OMReplayException;
 import org.apache.hadoop.ozone.om.response.file.OMFileCreateResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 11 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -254,29 +276,23 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-      omKeyInfo = prepareKeyInfo(omMetadataManager, keyArgs,
-          omMetadataManager.getOzoneKey(volumeName, bucketName,
-              keyName), keyArgs.getDataSize(), locations,
-          encryptionInfo.orNull(), ozoneManager.getPrefixManager(),
-          bucketInfo, transactionLogIndex);
-
-      omClientResponse =  prepareCreateKeyResponse(keyArgs, omKeyInfo,
-          locations, encryptionInfo.orNull(), exception,
-          createFileRequest.getClientID(), transactionLogIndex, volumeName,
+      omKeyInfo = prepareKeyInfo(omMetadataManager, keyArgs, ozoneKey,
```
bharatviswa504: We can pass the dbKeyInfo to prepareKeyInfo, as for override keys we read from DB. As anyway to check whether it is replay or not we read from DB above we can use that.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
```diff
@@ -174,26 +182,40 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
       encryptionInfo = getFileEncryptionInfo(ozoneManager, bucketInfo);
 
-      omKeyInfo = prepareKeyInfo(omMetadataManager, keyArgs,
-          omMetadataManager.getOzoneKey(volumeName, bucketName, keyName),
+      omKeyInfo = prepareKeyInfo(omMetadataManager, keyArgs, dbKeyName,
           keyArgs.getDataSize(), locations, encryptionInfo.orNull(),
           ozoneManager.getPrefixManager(), bucketInfo, transactionLogIndex);
-      omClientResponse = prepareCreateKeyResponse(keyArgs, omKeyInfo,
-          locations, encryptionInfo.orNull(), exception,
+      omClientResponse = prepareCreateKeyResponse(omResponse, keyArgs,
```
bharatviswa504: We can pass the dbKeyInfo to prepareKeyInfo, as for override keys we read from DB. As anyway to check whether it is replay or not we read from DB above we can use that.

hanishakoneru:
Thank you @bharatviswa504 for the review. I addressed them in the new commit.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
hanishakoneru: Done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
hanishakoneru: Done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
```diff
@@ -250,91 +249,113 @@ public EncryptedKeyVersion run() throws IOException {
...
-        exception = ex;
-      }
+    long openVersion = omKeyInfo.getLatestVersionLocations().getVersion();
 
-      if (exception != null) {
-        LOG.error("{} failed for Key: {} in volume/bucket:{}/{}",
-            omAction.getAction(), keyName, bucketName, volumeName, exception);
-        omClientResponse = createKeyErrorResponse(ozoneManager.getMetrics(),
-            omAction, exception, omResponse);
+    // Append blocks
```
bharatviswa504: We are adding blocks twice for the case when keyInfo != null, as once in prepareKeyInfo and here again.
So, can we move the creation of KeyInfo logic to prepareKeyInfo and use prepareCreateKeyResponse just to generate OmClientResponse.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeyCreateResponse.java
```diff
@@ -38,26 +36,30 @@
...
-    // For OmResponse with failure, this should do nothing. This method is
-    // not called in failure scenario in OM code.
-    if (getOMResponse().getStatus() == OzoneManagerProtocolProtos.Status.OK) {
-      String openKey = omMetadataManager.getOpenKey(omKeyInfo.getVolumeName(),
-          omKeyInfo.getBucketName(), omKeyInfo.getKeyName(),
-          openKeySessionID);
-      omMetadataManager.getOpenKeyTable().putWithBatch(batchOperation,
-          openKey, omKeyInfo);
-    }
+    String openKey = omMetadataManager.getOpenKey(omKeyInfo.getVolumeName(),
```
bharatviswa504: NIT: public -> protected. (In Base class addToBatch is protected)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
bharatviswa504: I think we are adding blocks only once. But still, if we consolidate the logic it will be good. Okay with doing that in a new Jira also.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -194,6 +212,28 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
       }
 
+      // Check if Key already exists in KeyTable and this transaction is a
+      // replay.
+      String ozoneKey = omMetadataManager.getOzoneKey(volumeName, bucketName,
+          keyName);
+      OmKeyInfo dbKeyInfo = omMetadataManager.getKeyTable().get(ozoneKey);
+      if (dbKeyInfo != null) {
+        // Check if this transaction is a replay of ratis logs.
+        // Check if this transaction is a replay of ratis logs.
```
bharatviswa504: Minor: Duplicate  // Check if this transaction is a replay of ratis logs.
We can remove one.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -254,36 +294,88 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-          locations, encryptionInfo.orNull(), exception,
-          createFileRequest.getClientID(), transactionLogIndex, volumeName,
-          bucketName, keyName, ozoneManager,
-          OMAction.CREATE_FILE, ozoneManager.getPrefixManager(), bucketInfo);
+
+      omKeyInfo = prepareKeyInfo(omMetadataManager, keyArgs, dbKeyInfo,
+          keyArgs.getDataSize(), locations, encryptionInfo.orNull(),
+          ozoneManager.getPrefixManager(), bucketInfo, trxnLogIndex);
+
+      long openVersion = omKeyInfo.getLatestVersionLocations().getVersion();
```
bharatviswa504: L302-L318 is common for FileCreate and KeyCreate can we move into a method and use it.
Or we can move these to prepareKeyInfo from 302-310 and add new method to add to cache and response. In this way OmKeyInfo will be generated by prepareKeyInfo and response and adding to cache will be in a new method.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponse.java
```diff
@@ -18,24 +18,25 @@
...
-  public OMFileCreateResponse(@Nullable OmKeyInfo omKeyInfo,
-      long openKeySessionID, @Nonnull OMResponse omResponse) {
-    super(omKeyInfo, openKeySessionID, omResponse);
+  public OMFileCreateResponse(@Nonnull OMResponse omResponse,
+      @Nonnull OmKeyInfo omKeyInfo, long openKeySessionID) {
+    super(omResponse, omKeyInfo, openKeySessionID);
+  }
+
+  public OMFileCreateResponse(@Nonnull OMResponse omResponse) {
+    super(omResponse);
```
bharatviswa504: Minor: Missing checkStatusNotOK() and also can we add similar javadoc as OMKeyCreateResponse.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
hanishakoneru: The issue is we have to get the openVersion before appending new blocks. Hence, I kept these lines out of prepareKeyInfo.
I can add the appendBlocks and updateCache calls to another method. But since it was just 2 calls, i though it could be outside.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponse.java
hanishakoneru: Thanks for catching this. If you are ok with rest of the patch, I add this check in the next Jira HDDS-2944.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
bharatviswa504: @hanishakoneru We shall open a new Jira to address this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponse.java
bharatviswa504: Sure.

bharatviswa504:

hanishakoneru:
Thank you @bharatviswa504 . I have addressed the two pending comments in HDDS-2944.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
hanishakoneru: When i move the append blocks code inside prepareKeyInfo, it breaks multiplartUpload. TestOzoneSecure#testMultipartUploadOverride() fails because of this. This was done as part of HDDS-2944. 
So going to keep append blocks outside the prepareKeyInfo() function.

PR #449, opened 2020-01-16
HDDS-2891. Apache NiFi PutFile processor is failing with secure Ozone S3G

elek:

Multipart upload is not working in secure environment as the `OzoneClientProducer` might create the the OzoneClient after executing the `HeaderPreprocessor` filter. `HeaderPreprocessor` modifies the header and the AWS signature check is not possible after executing this filter. It should be done before.

Hard to detect the problem, but if you debug both the AWS SDK under NiFi and the signature check of Ozone (OM + S3G) you can see that the canonical message (the base of the signature) contains a wrong content type (`ozone/mpu`) instead of the original one. 

https://issues.apache.org/jira/browse/HDDS-2891

## How was this patch tested?

I started a local NiFi and created a flow with PutS3 processor. It worked

## What has been changed in this patch?

Over the time it become a bigger patch, so here is an overview:

```
deleted:    hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/header/AuthenticationHeaderParser.java
renamed:    hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSAuthParser.java -> hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/SignatureProcessor.java
renamed:    hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSV4AuthParser.java -> hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSV4SignatureProcessor.java
```

The main idea is to remove the duplicated code and use only one single class to process the signature. (Renamed to SignatureProcessor as we already have a header parser to parse the header entry).

The other big modification is to do a snapshot in the SignatureProcessor from the HTTP headers. Earlier it was a writable reference which made it very fragile as the HeaderPreprocessor modifies the header.

```
modified:   hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneClientProducer.java
```

Just use the one single header parser instead of the other, old one.

```
modified:   hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/HeaderPreprocessor.java
```

To make it more safe, I save the original header and the SignatureProcessor can restore the original Contant-Type.-

```
new file:   hadoop-ozone/s3gateway/src/test/java/org/apache/hadoop/ozone/s3/TestAWSV4SignatureProcessor.java
```

A new unit test to make sure the Content-Type restore is working

```
modified:   hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/util/OzoneS3Util.java

```

That's a tricky bug. We need to use `toLowerCase` as it's already used on the server side. Without that we can get very cryptic error. (For example during the createBucket the bucket is created but during the check at the end of the call the acl check is failing....

```
modified:   hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
modified:   hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/VirtualHostStyleFilter.java
modified:   hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/endpoint/BucketEndpoint.java
modified:   hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/endpoint/EndpointBase.java
modified:   hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/header/AuthorizationHeaderV4.java
modified:   hadoop-ozone/s3gateway/src/test/java/org/apache/hadoop/ozone/s3/TestOzoneClientProducer.java
modified:   hadoop-ozone/s3gateway/src/test/java/org/apache/hadoop/ozone/s3/TestVirtualHostStyleFilter.java
modified:   hadoop-ozone/s3gateway/src/test/java/org/apache/hadoop/ozone/s3/endpoint/TestBucketPut.java
modified:   hadoop-ozone/s3gateway/src/test/java/org/apache/hadoop/ozone/s3/endpoint/TestRootList.java
modified:   hadoop-ozone/s3gateway/src/test/java/org/apache/hadoop/ozone/s3/header/TestAuthorizationHeaderV4.java
```

The remaining modifications just test updated / clarifications.

```
modified:   hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/endpoint/RootEndpoint.java
```

This one thing what I couldn't save. The automatic redirect to the S3 browser is removed. (Earlier if you put the s3 bucket URL to the browser it worked. We need a bigger refactor to save it.
~
~


hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneClientProducer.java
```diff
@@ -75,36 +70,31 @@ private OzoneClient getClient(OzoneConfiguration config) throws IOException {
...
-            UserGroupInformation remoteUser =
-                UserGroupInformation.createRemoteUser(
-                    v4RequestParser.getAwsAccessId());
-            remoteUser.addToken(token);
-            UserGroupInformation.setLoginUser(remoteUser);
-          } catch (OS3Exception | URISyntaxException ex) {
-            LOG.error("S3 auth info creation failed.");
-            throw S3_AUTHINFO_CREATION_ERROR;
+        try {
+
```
bharatviswa504: Right now, we have only AWSV4Header authentication only right, so we need the header check.
Or is there any reason for removal of this check

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/util/OzoneS3Util.java
```diff
@@ -38,9 +38,9 @@
   private OzoneS3Util() {
   }
 
-  public static String getVolumeName(String userName) {
+  public static String getS3Username(String userName) {
```
bharatviswa504: Can we rename this getVolumeName, as it is converting to md5Hex, and we use that as volume name.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/RootPageDisplayFilter.java
```diff
@@ -0,0 +1,61 @@
...
+      ServletResponse servletResponse, FilterChain filterChain)
+      throws IOException, ServletException {
+    HttpServletRequest httpRequest = (HttpServletRequest) servletRequest;
+    String httpMethod = httpRequest.getMethod();
+    String uri = httpRequest.getRequestURI();
+    String authorizationHeader = httpRequest.getHeader("Authorization");
+    if (httpMethod.equalsIgnoreCase("GET") && authorizationHeader == null && uri
+        .equals("/")) {
+      ((HttpServletResponse) servletResponse).sendRedirect("/static/");
+    } else {
```
bharatviswa504: What does else part do, for other HTTP method requests?
Or we should fail the request as it is not a valid request.
Or I am not sure if my understanding is correct, could you provide some info here.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneClientProducer.java
elek: The idea was to avoid code duplication. Instead of using `AWSV4AuthParser` we use the common `SignatureProcessor`.
This particular check is included in the `AWS3SignatureProcessor->authorizationHeader->validateAlgorithm()` which will throw an exception if it's anything else than AWS4.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/RootPageDisplayFilter.java
elek: `filterChain.doFilter` (else branch) means: "do all the other stuff defined by other filters/servlets/jaxrs classes.

This condition has a short-circuit response for the "GET /" page (without Authorization header). In this case it immediately redirect to the internal documentation page.

For any other case it does the normal work. (else)

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/util/OzoneS3Util.java
elek: I am not sure what is the right name, I renamed after a long debug session when I was totally confused what is in a variable.

This method doesn't return with a volumeName it returns with and unique (and safe) identifier of the s3 user. It will be converted to an s3 volume name on the server side ("s3" + this unique id).

This string is used as a parameter of the `createS3Bucket(String userName, String bucketName)` which clearly shows that this is not a volumeName. This is a unique ID / username.

During the debug I was very confused as I expected to have the final volumeName from this method, but didn't get it.

But please let me know if you have any better suggestion for the name.

elek:
@bharatviswa504 @arp7 Any more comments? Can we commit it, please?

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/util/OzoneS3Util.java
bharatviswa504: Agreed. I am fine with current way.

elek:
Thanks a lot the review and commit @bharatviswa504 (I know it's very time consuming this kind of very low-level changes...)

PR #450, opened 2020-01-16
HDDS-2893. Handle replay of KeyPurge Request.

hanishakoneru:

If KeyPurgeRequest is replayed, we do not want to purge the keys which were created after the original purge request was received. This could happen if a key was deleted, purged and then created and deleted again. The the purge request was replayed, it would purge the key deleted after the original purge request was completed.
Hence, to maintain idempotence, we should only purge those keys from DeletedKeys table that have updateID < transactionLogIndex of the request.

https://issues.apache.org/jira/browse/HDDS-2893

## How was this patch tested?

Unit tests to verify replayed transactions are executed as expected.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyPurgeRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyPurgeRequest.java
@@ -18,20 +18,28 @@
 
 package org.apache.hadoop.ozone.om.request.key;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import org.apache.hadoop.ozone.om.OMMetadataManager;
 import org.apache.hadoop.ozone.om.OzoneManager;
+import org.apache.hadoop.ozone.om.helpers.OmKeyInfo;
+import org.apache.hadoop.ozone.om.helpers.RepeatedOmKeyInfo;
 import org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerDoubleBufferHelper;
 import org.apache.hadoop.ozone.om.response.OMClientResponse;
 import org.apache.hadoop.ozone.om.response.key.OMKeyPurgeResponse;
-import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.DeleteKeysInBucket;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMRequest;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMResponse;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.PurgeKeysRequest;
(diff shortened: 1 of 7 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyPurgeRequest.java
```diff
@@ -49,25 +57,76 @@ public OMKeyPurgeRequest(OMRequest omRequest) {
...
+        .setPurgeKeysResponse(PurgeKeysResponse.newBuilder().build())
         .setStatus(Status.OK)
-        .setSuccess(true)
-        .build();
+        .setSuccess(true);
+    OMClientResponse omClientResponse = null;
+    boolean success = true;
+    IOException exception = null;
+
+    // Filter the keys that objectID > transactionLogIndex. This is done so
```
bharatviswa504: objectID  -> updatedID

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyPurgeRequest.java
```diff
@@ -49,25 +57,76 @@ public OMKeyPurgeRequest(OMRequest omRequest) {
...
+            for (OmKeyInfo omKeyInfo : repeatedOmKeyInfo.getOmKeyInfoList()) {
+              // Discard those keys whose updateID is > transactionLogIndex.
+              // This could happen when the PurgeRequest is replayed.
+              if (isReplay(ozoneManager, omKeyInfo.getUpdateID(),
+                  transactionLogIndex)) {
+                purgeKey = false;
+                break;
+              }
+            }
+            if (purgeKey) {
```
bharatviswa504: As discussed offline.
Here there is a chance of deleting the old keyInfo from RepeatedKeyInfo when a new Key with transaction ID greater than the transactionLogIndex of purgeKeyRequest.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyPurgeRequest.java
```diff
@@ -115,18 +126,42 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
               bucketName);
         }
       }
+
+      if (result == Result.REPLAY) {
+        LOG.debug("Replayed Transaction {}. Request: {}", trxnLogIndex,
+            purgeKeysRequest);
+        if (!keysNotPurged.isEmpty()) {
```
bharatviswa504: NIT: We can use the Java8 String method String.join(", list)

bharatviswa504: And also as discussed offline for debug statements we can use isDebugEnabled, so that parameter conversion will not happen.

hanishakoneru:
Thanks for the review @bharatviswa504 . Addressed your comments and fixed CI issues. Can you please take a look.

bharatviswa504:
@hanishakoneru I don't see any latest commit, the last commit is from Yesterday.

hanishakoneru:
Sorry missed pushing the new commit. Added it now.

PR #451, opened 2020-01-16
HDDS-2876. Consolidate ObjectID and UpdateID from Info objects into one class.

hanishakoneru:

We use ObjectID and BucketID in OMVolumeArgs, OMBucketInfo, OMKeyInfo and OMMultipartKeyInfo. We can consolidate by having these Info objects extend a "WithObjectID" class which can host the common fields - objectID and updateID.

https://issues.apache.org/jira/browse/HDDS-2876

## How was this patch tested?

Code consolidation. Separate Tests not required.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
@@ -403,7 +403,13 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
 
   @Override
   public String getObjectInfo() {
-    return getProtobuf().toString();
+    return "OMVolumeArgs{" +
+        "volume='" + volumeName + '\'' +
+        ", bucket='" + bucketName + '\'' +
+        ", isVersionEnabled='" + isVersionEnabled + '\'' +
+        ", storageType='" + storageType + '\'' +
+        ", creationTime='" + creationTime + '\'' +
+        '}';
   }
 
   @Override
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
@@ -40,7 +40,7 @@ import com.google.common.base.Preconditions;
 /**
  * A class that encapsulates Bucket Info.
  */
-public final class OmBucketInfo extends WithMetadata implements Auditable {
+public final class OmBucketInfo extends WithObjectID implements Auditable {
   /**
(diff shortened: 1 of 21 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
```diff
@@ -463,6 +433,19 @@ public static OmKeyInfo getFromProtobuf(KeyInfo keyInfo) {
     return builder.build();
   }
 
+  @Override
+  public String getObjectInfo() {
+    return "OMVolumeArgs{" +
```
adoroszlai: Shouldn't this be `OmKeyInfo`?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
```diff
@@ -452,6 +401,17 @@ public static OmBucketInfo getFromProtobuf(BucketInfo bucketInfo) {
     return obib.build();
   }
 
+  @Override
+  public String getObjectInfo() {
+    return "OMVolumeArgs{" +
```
adoroszlai: Shouldn't this be `OmBucketInfo`?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
hanishakoneru: Fixed.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
hanishakoneru: Fixed.

hanishakoneru:
Thanks for the review @adoroszlai . Fixed the object names.

elek:
Fresh build is passed: 
https://github.com/apache/hadoop-ozone/commit/a7bc81ed113195a5c72871ad89cbb7308e307ac7

Committing now with a minor conflict resolution:

```
         OmKeyInfo dbKeyInfo = omMetadataManager.getKeyTable().get(ozoneKey);
         if (dbKeyInfo != null) {
-          if (isReplay(ozoneManager, dbKeyInfo.getUpdateID(), trxnLogIndex)) {
+          if (isReplay(ozoneManager, dbKeyInfo, trxnLogIndex)) {
             // This transaction is a replay. Send replay response.
             throw new OMReplayException();
```

hanishakoneru:
Thank you @elek and @adoroszlai for reviewing and merging the PR.

PR #452, opened 2020-01-16
HDDS-2894. Handle replay of KeyDelete and KeyRename Requests

hanishakoneru:

To ensure that key deletion and rename operations are idempotent, compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

OMKeyDeleteRequest and OMKeyRenameRequest are made idempotent in this Jira.

https://issues.apache.org/jira/browse/HDDS-2894

## How was this patch tested?

Unit tests to verify replayed transactions are executed as expected.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
@@ -175,7 +175,8 @@ public class OMKeyRenameRequest extends OMKeyRequest {
               // Add to cache. Only fromKey should be deleted. ToKey already
               // exists in DB as this transaction is a replay.
               result = Result.DELETE_FROM_KEY_ONLY;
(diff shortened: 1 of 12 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
```diff
@@ -133,79 +143,146 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+              keyTable.addCacheEntry(new CacheKey<>(fromKey),
+                  new CacheValue<>(Optional.absent(), trxnLogIndex));
+
+              omClientResponse = new OMKeyRenameResponse(omResponse
+                  .setRenameKeyResponse(RenameKeyResponse.newBuilder()).build(),
+                  fromKeyName, fromKeyValue);
+            }
+          }
+
+          if (result == null) {
```
bharatviswa504: Minor: We can write this also else part of fromKeyValue != null. It will be easy to follow the flow.

bharatviswa504:

PR #453, opened 2020-01-17
HDDS-2897. Ozone recon Start failed due to Kerberos principal not bei…

avijayanhwx:
…ng found.

The Ozone Configuration was not getting passed into the Http Server since there was a recent change in the ConfigurationProvider to provide an instance of 'OzoneConfiguration' instead of 'Configuration'. Hence, Guice created a new Configuration instance for the HttpServer. Since the server did not get the kerberos related configuration, it shut down in the secure environment.

https://issues.apache.org/jira/browse/HDDS-2897

## How was this patch tested?
Manually tested on a secure cluster and verified Recon starts up fine.


avijayanhwx:
cc @swagle / @shwetayakkali

mukul1987:
@avijayanhwx The changes look good to me. Can we also add a robot test to start recon in secure environment?
We already have robot test for secure cluster.

avijayanhwx:
> @avijayanhwx The changes look good to me. Can we also add a robot test to start recon in secure environment?
> We already have robot test for secure cluster.

Thanks for the review @mukul1987. I have added a basic robot test for testing recon REST API works in secure/unsecure environments. Created HDDS-2899 to track adding robot tests for Recon-OM integration.

hadoop-ozone/dist/src/main/smoketest/security/ozone-secure-recon.robot
```diff
@@ -0,0 +1,30 @@
...
+Resource            ../commonlib.robot
+
+*** Variables ***
+${ENDPOINT_URL}     http://recon:9888
+
+*** Test Cases ***
+Recon REST API
+    Run Keyword if      '${SECURITY_ENABLED}' == 'true'     Kinit HTTP user
+    ${result} =         Execute                             curl --negotiate -u : -v ${ENDPOINT_URL}/api/containers
+                        Should contain      ${result}       containers
```
adoroszlai: The robot tests look good, but I think we could get rid of the duplication.  `security/ozone-secure-recon.robot` works in both secure and unsecure env.  So I propose:

1. rename `security/ozone-secure-recon.robot` to `recon/recon-api.robot` (ie. get rid of the unsecure-only test, but keep its name and location)
2. add `execute_robot_test recon recon` in `ozonesecure/test.sh`, too

This way the same test would be executed in both secure and unsecure env, similar to eg. `s3` tests:

https://github.com/apache/hadoop-ozone/blob/1caf1e30865aa2b380a7fca6d87f5ae8034fee4e/hadoop-ozone/dist/src/main/compose/ozone/test.sh#L39

https://github.com/apache/hadoop-ozone/blob/1caf1e30865aa2b380a7fca6d87f5ae8034fee4e/hadoop-ozone/dist/src/main/compose/ozonesecure/test.sh#L36

avijayanhwx: @adoroszlai Thank you for the suggestion. I have addressed it.

adoroszlai:
Thanks @avijayanhwx for the fix, and @swagle and @mukul1987 for reviews.

PR #454, opened 2020-01-17
HDDS-2900. Avoid logging NPE when space usage check is not configured

adoroszlai:

If `hdds.datanode.du.factory.classname` is not configured, a harmless but annoying NPE appears in datanode log during startup.  This change avoids the NPE by explicitly checking for empty config.

https://issues.apache.org/jira/browse/HDDS-2900

## How was this patch tested?

Tweaked unit test.  Verified acceptance test docker output.

https://github.com/adoroszlai/hadoop-ozone/runs/394927391


adoroszlai:

PR #455, opened 2020-01-17
HDDS-2895. Generate only the required keytabs for docker based secure tests

elek:

We have acceptance tests with the help of docker/docker-compose where we generate the keytab files on the fly with the help of a lightweight (unsecure) REST endpoint.

But this generation can be very slow especially when the DNS is slow. When I start a VPN the secure cluster can't be started in the 90 sec period. (>100 keytabs are generated and one keytab generation is ~5 sec).

The solutions is to generate only the required keystabs in each of the containers.

Instead of request all the possible keytabs in the generic `docker-config`:

```
KERBEROS_KEYTABS=dn om scm HTTP testuser testuser2 s3g 
```

We can defined the required keytabs per service (in `docker-compose.yaml`)

```
environment:
  KERBEROS_KEYTABS=scm HTTP
```

With this approach ~20 keytab file will be generated instead of >100 and the secure tests will be significant faster.

https://issues.apache.org/jira/browse/HDDS-2895

## How was this patch tested?

This is an acceptance test improvements. If they are passed, it should be fine...

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
+++ b/hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
@@ -24,7 +24,6 @@ services:
     hostname: kdc
     volumes:
       - ../..:/opt/hadoop
-  
   kms:
       image: apache/hadoop:${HADOOP_VERSION}
       ports:
@@ -42,6 +41,8 @@ services:
     command: ["/opt/hadoop/bin/ozone","datanode"]
     env_file:
       - docker-config
+    environment:
+      KERBEROS_KEYTABS: dn
   om:
     image: apache/ozone-runner:${OZONE_RUNNER_VERSION}
     hostname: om
@@ -51,9 +52,11 @@ services:
       - 9874:9874
     environment:
       ENSURE_OM_INITIALIZED: /data/metadata/om/current/VERSION
+      KERBEROS_KEYTABS: om HTTP
     env_file:
       - docker-config
     command: ["/opt/hadoop/bin/ozone","om"]
+
   s3g:
     image: apache/ozone-runner:${OZONE_RUNNER_VERSION}
     hostname: s3g
@@ -64,6 +67,8 @@ services:
     env_file:
       - ./docker-config
     command: ["/opt/hadoop/bin/ozone","s3g"]
+    environment:
+      KERBEROS_KEYTABS: s3g HTTP
   recon:
     image: apache/ozone-runner:${OZONE_RUNNER_VERSION}
     hostname: recon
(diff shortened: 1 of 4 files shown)


hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
```diff
@@ -42,6 +41,8 @@ services:
     command: ["/opt/hadoop/bin/ozone","datanode"]
     env_file:
       - docker-config
+    environment:
+      KERBEROS_KEYTABS: dn
```
adoroszlai: Seems like acceptance test failure is caused by missing `HTTP` keytab on datanode.

```
datanode_1  | 2020-01-17 12:30:34,899 [main] ERROR ozone.HddsDatanodeService: HttpServer failed to start.
...
datanode_1  | Caused by: javax.servlet.ServletException: javax.servlet.ServletException: Keytab does not exist: /etc/security/keytabs/HTTP.keytab
```

```suggestion
      KERBEROS_KEYTABS: dn HTTP
```

hadoop-ozone/dist/src/main/compose/ozonesecure/docker-config
```diff
@@ -80,12 +80,11 @@ HDFS-SITE.XML_rpc.metrics.percentiles.intervals=60,300
...
 KERBEROS_KEYSTORES=hadoop
 KERBEROS_SERVER=kdc
 JAVA_HOME=/usr/lib/jvm/jre
 JSVC_HOME=/usr/bin
 SLEEP_SECONDS=5
 KERBEROS_ENABLED=true
 
-no_proxy=om,scm,s3g,kdc,localhost,127.0.0.1
\ No newline at end of file
+no_proxy=om,scm,s3g,kdc,localhost,127.0.0.1
```
xiaoyuyao: should we have recon in the no_proxy list?

elek: AFAIK it was added to support the the curl/wget executed by the smoketests. As we have no recon specific acceptance tests (yet, hopefully), it's not immediately required...

xiaoyuyao: There is one recon acceptance tests just added recently to catch a regression in HDDS-2897.
See #453 for details.

adoroszlai: This was added in another task, only newline at EOF is changed.  So I think it should be addressed in a new Jira.

adoroszlai:
Thanks @elek for the contribution and @xiaoyuyao for the review.

PR #457, opened 2020-01-17
HDDS-2896. Use regex to match with ratis properties when creating ratis client

bharatviswa504:

Use directly ratis client configuration, match them with regex, and set them to raft properties when creating ratis client.

https://issues.apache.org/jira/browse/HDDS-2896

## How was this patch tested?

Added UT to test the new methods.


bharatviswa504:

PR #458, opened 2020-01-17
HDDS-2903. Use regex to match with ratis properties when creating ratis server.

bharatviswa504:

Use directly ratis server configuration, match them with regex, and set them to raft properties when creating ratis server.

https://issues.apache.org/jira/browse/HDDS-2903

## How was this patch tested?

Added UT for the new method.

This PR combines both HDDS-2896 and HDDS-2903.


bharatviswa504:

PR #461, opened 2020-01-19
HDDS-2910. OzoneManager startup failure with throwing unhelpful excep…

linyiqun:
Add more helpful message in exception, so that it can quickly guide users how to resolve the issue in starting up OM service.

https://issues.apache.org/jira/browse/HDDS-2910

## How was this patch tested?
Just the log message update change, no need to do the test.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -335,7 +335,9 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
         OZONE_OM_USER_MAX_VOLUME + " value should be greater than zero");
 
     if (omStorage.getState() != StorageState.INITIALIZED) {
-      throw new OMException("OM not initialized.",
+      throw new OMException("OM not initialized, current OM storage state: " +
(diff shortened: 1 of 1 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -335,7 +335,9 @@ private OzoneManager(OzoneConfiguration conf) throws IOException,
         OZONE_OM_USER_MAX_VOLUME + " value should be greater than zero");
 
     if (omStorage.getState() != StorageState.INITIALIZED) {
-      throw new OMException("OM not initialized.",
+      throw new OMException("OM not initialized, current OM storage state: " +
+          omStorage.getState().name() + ". Please check if 'ozone om --init' "
```
xiaoyuyao: Can we rephrase "Please check if 'ozone om --init' command is executed before starting the OM service.",
Please ensure 'ozone om --init' command is executed once before starting the OM service.",

linyiqun: Thanks @xiaoyuyao  for the review. Have addressed the comment now.

linyiqun:
Thanks for the review and merge, @elek !

PR #466, opened 2020-01-21
HDDS-2869. Handle pipeline bootstrap from SCM and create pipeline use case in Recon.

avijayanhwx:

- Implement SCM pipeline metadata bootstrap in Recon.
- Handle CREATE pipeline use case in Recon (Recon sees a new pipeline ID in Datanode heartbeat)
- Stub out pipeline provider in Recon such that it can NEVER create a Ratis pipeline.
- Added SCM RPC endpoint to get a single Pipeline info.

https://issues.apache.org/jira/browse/HDDS-2869

## How was this patch tested?
Manually tested on a 3 node docker cluster.
Added unit tests for new Recon methods.


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineFactory.java
```diff
@@ -35,7 +35,7 @@
 /**
  * Creates pipeline based on replication type.
  */
-public final class PipelineFactory {
+public class PipelineFactory {
```
swagle: The access can be made package-private?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineReportHandler.java
```diff
@@ -120,4 +116,24 @@ private void processPipelineReport(PipelineReport report, DatanodeDetails dn,
       }
     }
   }
+
+
+  protected void reportDatanode(Pipeline pipeline, DatanodeDetails dn)
```
swagle: For cleaner semantics, this can be renamed to setReportedDatanode.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -200,6 +210,16 @@ public Pipeline getPipeline(PipelineID pipelineID)
     }
   }
 
+  @Override
+  public boolean containsPipeline(PipelineID pipelineID) {
```
swagle: Instead of contains, we could have made the get return an Optiona<Pipeline>

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -523,4 +543,21 @@ protected File getPipelineDBPath(Configuration configuration) {
     File metaDir = ServerUtils.getScmDbDir(configuration);
     return new File(metaDir, SCM_PIPELINE_DB);
   }
+
+  protected ReadWriteLock getLock() {
```
swagle: Instead of a get on a lock object, it would make sense to make the reference accessible.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
```diff
@@ -408,6 +408,15 @@ public Pipeline createReplicationPipeline(HddsProtos.ReplicationType type,
     return scm.getPipelineManager().getPipelines();
   }
 
+  @Override
+  public Pipeline getPipeline(HddsProtos.PipelineID pipelineID)
+      throws IOException {
+    AUDIT.logReadSuccess(
```
swagle: This might not be appropriate for Audit since it is not client access, we should log it in debug mode IMO.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconPipelineFactory.java
```diff
@@ -0,0 +1,72 @@
...
+        new ReconMockPipelineProvider();
+    setProviders(new DefaultedMap(reconMockPipelineProvider));
+  }
+
+  static class ReconMockPipelineProvider implements PipelineProvider {
+
+    @Override
+    public Pipeline create(HddsProtos.ReplicationFactor factor){
+      LOG.warn("Trying to create pipeline in Recon, which is prohibited!");
+      return null;
```
swagle: Should we consider throwing a RuntimeException instead?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconPipelineFactory.java
```diff
@@ -0,0 +1,72 @@
...
+  private static final Logger LOG =
+      LoggerFactory.getLogger(ReconPipelineFactory.class);
+
+  ReconPipelineFactory() {
+    ReconMockPipelineProvider reconMockPipelineProvider =
+        new ReconMockPipelineProvider();
+    setProviders(new DefaultedMap(reconMockPipelineProvider));
+  }
+
+  static class ReconMockPipelineProvider implements PipelineProvider {
```
swagle: I would avoid calling this a Mock since unit tests mocks have a different meaning as they return stubs data structures.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/StorageContainerServiceProviderImpl.java
```diff
@@ -0,0 +1,94 @@
...
+  private OzoneConfiguration configuration;
+  private StorageContainerLocationProtocol scmClient;
+  private boolean isInitialized = false;
+
+  @Inject
+  public StorageContainerServiceProviderImpl(OzoneConfiguration configuration,
+      StorageContainerLocationProtocol scmClient) {
+    this.configuration = configuration;
+    this.scmClient = scmClient;
+    if (this.scmClient != null) {
```
swagle: How about calling initialize from the constructor and make that idempotent using the flag.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/StorageContainerServiceProviderImpl.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+public class StorageContainerServiceProviderImpl
+    implements StorageContainerServiceProvider {
+
+  private static final Logger LOG =
+      LoggerFactory.getLogger(StorageContainerServiceProviderImpl.class);
+
+  private OzoneConfiguration configuration;
+  private StorageContainerLocationProtocol scmClient;
+  private boolean isInitialized = false;
```
swagle: This should be declared volatile.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/StorageContainerServiceProviderImpl.java
```diff
@@ -0,0 +1,94 @@
...
+      this.scmClient = newContainerRpcClient(configuration);
+    } catch (IOException ioEx) {
+      LOG.error("Exception encountered while creating SCM client.", ioEx);
+    }
+    isInitialized = true;
+  }
+
+  @Override
+  public List<Pipeline> getPipelines() throws IOException {
+    if (!isInitialized) {
```
swagle: Same here, and next check becomes redundant.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineFactory.java
avijayanhwx: This is extended in org.apache.hadoop.ozone.recon.scm.ReconPipelineFactory.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconPipelineFactory.java
avijayanhwx: We don't need to do this since this class protects us against any unintended actions from Recon.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
avijayanhwx: That will lead to findbug violation. Do you see any issues using a getter?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineReportHandler.java
avijayanhwx: Done

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/StorageContainerServiceProviderImpl.java
avijayanhwx: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
avijayanhwx: Since the API is "contains", I wanted to keep it true/false. Can I address this in the next patch?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
avijayanhwx: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
swagle: It's a minor nit, I am ok with it asis. More of a preference thing, returning Optional already sets the contract and reduces code dup.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
swagle: I see, ok let's ignore this comment.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/audit/SCMAction.java
```diff
@@ -32,6 +32,7 @@
   GET_CONTAINER_WITH_PIPELINE,
   LIST_CONTAINER,
   LIST_PIPELINE,
+  GET_PIPELINE,
```
arp7: Usually we add new fields at the end of enums.

hadoop-ozone/dist/src/main/compose/ozone/docker-compose.yaml
```diff
@@ -67,6 +67,5 @@ services:
     ports:
       - 9888:9888
     environment:
-      WAITFOR: om:9874
```
arp7: Why is this removed?

hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
```diff
@@ -79,7 +79,6 @@ services:
     env_file:
       - ./docker-config
     environment:
-      WAITFOR: om:9874
```
arp7: Same.

hadoop-ozone/dist/src/main/compose/ozone/docker-compose.yaml
avijayanhwx: I just stumbled upon this while testing. We no longer need this. Recon can start up anytime and can learn OM and SCM state.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/audit/SCMAction.java
avijayanhwx: Sure, I can move it to the end.

arp7: I think it is okay for now as we don't have to maintain compatibility.

avijayanhwx:
> Thanks @avijayanhwx for implementing this. I tried it locally on a docker-compose cluster and found that Recon does not get new pipeline information.
> 
> 1. Since now the recon container does not wait with startup, it received 0 pipelines initially:
>    ```
>    recon_1     | 2020-01-29 08:41:05,263 [main] INFO scm.ReconStorageContainerManagerFacade: Obtained 0 pipelines from SCM.
>    recon_1     | 2020-01-29 08:41:05,264 [main] INFO scm.ReconPipelineManager: Recon has 0 pipelines in house.
>    ```
>    
>    
>    That would be OK, but when the pipelines were created a bit later, they did not show up in Recon.  (I also checked with `WAITFOR` restored, 3+1 initial pipelines were received in that case, but I feel it's a bit timing-dependent.)
> 2. Closed the initial 3-node pipeline via `scmcli`, which eventually triggered creation of new one in SCM, but Recon never noticed.
> 
> Can you please check?

@adoroszlai I have not wired up the DN-> Recon path by default since that may cause some unintended test failures before finishing the whole "Recon as a Passive SCM" feature. In my local, I test with the following configs added. 

> OZONE-SITE.XML_ozone.recon.datanode.address=recon:9891
> OZONE-SITE.XML_ozone.recon.address=recon:9891
> OZONE-SITE.XML_ozone.recon.datanode.bind.host=recon 

That will lead to pipelines being picked up by Recon.

adoroszlai:
> I have not wired up the DN-> Recon path by default

Thanks, that explains it.

adoroszlai:
Thanks @avijayanhwx for the contribution, and @arp7 and @swagle for the review.

swagle:
Thanks @adoroszlai for actually trying it out :-) 👍

avijayanhwx:
> Thanks @adoroszlai for actually trying it out :-) 👍

+1. Thanks @adoroszlai

PR #467, opened 2020-01-21
HDDS-2917. Add a different retry policy for watch requests.

bshashikant:

Added a request dependent retry policy in ozone. For write type requests in ratis, it will retry with default retry policy of retrying upto max count with sleep. For watch requests, there will be no retries.

https://issues.apache.org/jira/browse/HDDS-2917

## How was this patch tested?
Ran client side write tests and verified that watch requests are not retried.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
@@ -43,6 +43,7 @@ import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.ratis.RaftConfigKeys;
 import org.apache.ratis.client.RaftClient;
 import org.apache.ratis.client.RaftClientConfigKeys;
+import org.apache.ratis.client.retry.RequestTypeDependentRetryPolicy;
 import org.apache.ratis.conf.RaftProperties;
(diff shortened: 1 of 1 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
```diff
@@ -277,9 +278,15 @@ static RetryPolicy createRetryPolicy(Configuration conf) {
...
-    RetryPolicy retryPolicy = RetryPolicies
+    RetryPolicy writeRetryPolicy = RetryPolicies
         .retryUpToMaximumCountWithFixedSleep(maxRetryCount, sleepDuration);
-    return retryPolicy;
+    final RequestTypeDependentRetryPolicy.Builder builder =
+            RequestTypeDependentRetryPolicy.newBuilder();
+    builder.set(RaftProtos.
+            RaftClientRequestProto.TypeCase.WRITE, writeRetryPolicy);
+    builder.set(RaftProtos.
+            RaftClientRequestProto.TypeCase.WATCH, RetryPolicies.noRetry());
```
lokeshj1703: I think we will need to change the watch timeout as well in the leader(default is 10sec, we dont have a config for it in ozone). Maybe we should keep 2-3 retries to include the case of leader changes.

PR #471, opened 2020-01-22
HDDS-2906. Added Unit Test Cases ofr CRLCodec

abhishekaypurohit:

Added more Unit Tests for CRLCodec

https://issues.apache.org/jira/browse/HDDS-2906

## How was this patch tested?

mvn build and unit tests. 

@anuengineer

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/utils/CRLCodec.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/utils/CRLCodec.java
@@ -78,7 +78,7 @@ public class CRLCodec {
    * Returns a X509 CRL from the CRL Holder.
    *
    * @param holder - Holder
-   * @return X509Certificate.
+   * @return X509CRL - X509 CRL.
    * @throws CRLException - on Error.
    */
   public static X509CRL get509CRL(X509CRLHolder holder)
@@ -121,7 +121,8 @@ public class CRLCodec {
    * Gets the X.509 CRL from PEM encoded String.
    *
    * @param pemEncodedString - PEM encoded String.
-   * @return X509Certificate  - Certificate.
+   * @return X509CRL  - Crl.
+   * @throws CRLException - Thrown on Failure.
    * @throws CertificateException - Thrown on Failure.
    * @throws IOException          - Thrown on Failure.
    */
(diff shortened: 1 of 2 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/utils/CRLCodec.java
```diff
@@ -78,7 +78,7 @@ public CRLCodec(SecurityConfig securityConfig) {
    * Returns a X509 CRL from the CRL Holder.
    *
    * @param holder - Holder
-   * @return X509Certificate.
+   * @return X509CRL - X509 CRL.
    * @throws CRLException - on Error.
    */
   public static X509CRL get509CRL(X509CRLHolder holder)
```
xiaoyuyao: NIT: should we rename the method from get509CRL to getX509CRL?

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/utils/TestCRLCodec.java
```diff
@@ -101,11 +122,124 @@ public void testWriteCRL() throws IOException, OperatorCreationException {
...
+
+    X500Name issuer = x509CertificateHolder.getIssuer();
+    Date now = new Date();
+    X509v2CRLBuilder builder = new X509v2CRLBuilder(issuer, now);
+    builder.addCRLEntry(x509CertificateHolder.getSerialNumber(), now,
+                        CRLReason.cACompromise);
+
+    InputStream inStream = null;
+    byte[] crlBytes = TMP_CRL_ENTRY.getBytes();
+    try {
```
xiaoyuyao: NIT: the InputStream can be wrapped with try-with-resource.

abhishekaypurohit: Done.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/utils/CRLCodec.java
abhishekaypurohit: Good point. Done.

xiaoyuyao:
+1 I will merge it shortly.

PR #472, opened 2020-01-22
HDDS-2920. Remove ozone ratis client specific config keys.

bharatviswa504:

Remove ozone client ratis config, and make use of HDDS-2896 which makes use of direct raft client config.

https://issues.apache.org/jira/browse/HDDS-2920

## How was this patch tested?

Existing UT and acceptance test should cover. In this PR, we use direct Ratis Client Config instead of ozone config. Also it requires RATIS-796, as removed watch client time out ozone client config.


bshashikant:
Thanks @bharatviswa504 for working on this. I think in the tests, we need to set the watch timeout to a smaller value to verify/test certain conditions. We should not use the default watch timeout as it may lead to test failures.

bharatviswa504:
I see that we need the value in one of the watcher tests. Few Other tests pass, without any config change. (Few of tests failing, I think which is not related to this PR)

PR #478, opened 2020-01-22
HDDS-2924. Fix Pipeline#nodeIdsHash collision issue.

xiaoyuyao:
Change to use the set of datanodes that form the pipeline to detect overlapped pipeline allocation. 
Remove the nodeIdHash that could collide. 

https://issues.apache.org/jira/browse/HDDS-2924

## How was this patch tested?
unit tests

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
@@ -21,11 +21,13 @@ package org.apache.hadoop.hdds.scm.pipeline;
 import java.io.IOException;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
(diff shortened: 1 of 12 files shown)


xiaoyuyao:
The checkstyle issue should be fixed in next commit. However, there is an issue when interop multi-raft with rackawareness, which is detected by the topology acceptance test. It is not related to this change but I will look into it.

timmylicheng:
> The checkstyle issue should be fixed in next commit. However, there is an issue when interop multi-raft with rackawareness, which is detected by the topology acceptance test. It is not related to this change but I will look into it.

I have a JIRA https://issues.apache.org/jira/browse/HDDS-2923 to fix rack awareness issue,. We may track there.

xiaoyuyao:
Thanks @timmylicheng. In this one, I include some minor fixes around PipelinePlacementPolicy to cover the non-fallback case where we have enough nodes on different racks but are not chosen properly. 

Agree, we should have the fall back fix in HDDS-2923.

anuengineer:
That is a pretty smart fix. You are absolutely right, we need no magic to find if a three node set is equal to another. +1, once @xiaoyuyao  is happy.

xiaoyuyao:
push the correct rebase.

xiaoyuyao:
Thanks @anuengineer and @timmylicheng for the reviews. I will merge it shortly.

PR #479, opened 2020-01-22
HDDS-2925. Remove hdfs-client dependency from hdds-common

elek:

As of now we have a `hdds-common` -> `hdfs-client` dependency but in reality we don't use any important thing from hdfs-client, only a few string utils.

I would propose to remove this dependency to have a safer / smaller and more independent ozone package.

https://issues.apache.org/jira/browse/HDDS-2925

## How was this patch tested?

Tested with build only (will see the acceptance test results) as there is no big modification. I put back the dependency to the SCM server in *test* scope (!) as some client utils are used to test the secure availability of the HTTP servers.


anuengineer:
One more push with Rat and Checkstyle and other oddities, please. I will commit as soon as we get a good run. Once more, thank you very much for these patches. I really like them.

elek:
Thank you very much to support this effort @anuengineer.

It turned out to be more messy than I thought: hadoop-hdfs (server) has no compile time dependency on hdfs-client but throws class not found exception at runtime if it's not there (I think it's a wrong usage of `provided` maven dependency scope).

I put hdfs-client back to our *server* side dependency. And will continue the cleanup in follow-up jiras.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/StingUtils.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+package org.apache.hadoop.hdds;
+
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Simple utility class to collection string conversion methods.
+ */
+public final class StingUtils {
```
adoroszlai:
```suggestion
public final class StringUtils {
```

hadoop-ozone/dev-support/checks/unit.sh
```diff
@@ -17,7 +17,7 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 cd "$DIR/../../.." || exit 1
 
 export MAVEN_OPTS="-Xmx4096m"
-mvn -B -DskipShade -Dskip.yarn -fae test -pl \!:hadoop-ozone-integration-test "$@"
+mvn -B -DskipShade -Dskip.yarn -fae install -pl \!:hadoop-ozone-integration-test "$@"
```
adoroszlai: Can you please explain the goal of this change?

elek: I tried to understand why the unit tests are failed and I was afraid that some changes are missing because the missing install phase. But I think it was because the tricky provided dependency between hdfs server and hdfs.

I reverted the change.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/StingUtils.java
elek: Not sure what is is your exact problem with [Sting](https://en.wikipedia.org/wiki/Sting_%28musician%29) but changed it to a more boring one ;-)

(=thanks, fixed it)

elek:
Thanks the review @adoroszlai and @anuengineer. I am merged it to the master.

PR #480, opened 2020-01-22
HDDS-2927. Cache EndPoint tasks instead of creating them all the time in RunningDatanodeState

linyiqun:
Cache EndPoint tasks instead of creating them all the time in the class RunningDatanodeState.

https://issues.apache.org/jira/browse/HDDS-2927

## How was this patch tested?
This is just a code refactoring change. Existed unit tests can cover this change.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
@@ -30,6 +30,7 @@ import org.apache.hadoop.util.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
@@ -54,7 +55,7 @@ public class RunningDatanodeState implements DatanodeState {
   private final StateContext context;
   private CompletionService<EndPointStates> ecs;
   /** Cache the end point task per end point per end point state. */
-  private Map<String, Map<EndPointStates,
+  private Map<EndpointStateMachine, Map<EndPointStates,
       Callable<EndPointStates>>> endpointTasks;
 
   public RunningDatanodeState(Configuration conf,
@@ -71,15 +72,15 @@ public class RunningDatanodeState implements DatanodeState {
    * each end point state.
    */
   private void initEndPointTask() {
-    endpointTasks = new HashMap<String, Map<EndPointStates,
+    endpointTasks = new HashMap<EndpointStateMachine, Map<EndPointStates,
         Callable<EndPointStates>>>();
     for (EndpointStateMachine endpoint : connectionManager.getValues()) {
-      Map<EndPointStates, Callable<EndPointStates>> endpointTaskForState
-          = new HashMap<EndPointStates, Callable<EndPointStates>>();
+      EnumMap<EndPointStates, Callable<EndPointStates>> endpointTaskForState =
+          new EnumMap<>(EndPointStates.class);
 
       for (EndPointStates state : EndPointStates.values()) {
         Callable<EndPointStates> endPointTask = null;
-        switch (endpoint.getState()) {
+        switch (state) {
         case GETVERSION:
           endPointTask = new VersionEndpointTask(endpoint, conf,
               context.getParent().getContainer());
@@ -101,12 +102,15 @@ public class RunningDatanodeState implements DatanodeState {
               .setContext(context)
               .build();
           break;
-        case SHUTDOWN:
+        default:
           break;
         }
-        endpointTaskForState.put(state, endPointTask);
+
+        if (endPointTask != null) {
+          endpointTaskForState.put(state, endPointTask);
+        }
       }
-      endpointTasks.put(endpoint.toString(), endpointTaskForState);
+      endpointTasks.put(endpoint, endpointTaskForState);
     }
   }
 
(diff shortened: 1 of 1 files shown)


linyiqun:
Thanks for the review @adoroszlai .
Actually I was thinking about there will be multiple endpoints here. If we just construct <State, EndPointTask> cache mapping, the same state of some other end point task will be override. But seems current logic is still not correct. I do a minor change to construct <EndPoint Adderess, <EndPointState, EndPointTask>> mapping. This mapping should be right now.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
```diff
@@ -49,14 +52,62 @@
...
+  private void initEndPointTask() {
+    endpointTasks = new HashMap<String, Map<EndPointStates,
+        Callable<EndPointStates>>>();
+    for (EndpointStateMachine endpoint : connectionManager.getValues()) {
+      Map<EndPointStates, Callable<EndPointStates>> endpointTaskForState
+          = new HashMap<EndPointStates, Callable<EndPointStates>>();
+
+      for (EndPointStates state : EndPointStates.values()) {
+        Callable<EndPointStates> endPointTask = null;
+        switch (endpoint.getState()) {
```
adoroszlai: Shouldn't this be `state` instead of `endpoint.getState()`?  Assuming the endpoint state is constant during initialization, wouldn't this create several instances of the same type (eg. `RegisterEndpointTask`) for each endpoint?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
```diff
@@ -49,14 +52,62 @@
...
+              .setConfig(conf)
+              .setEndpointStateMachine(endpoint)
+              .setDatanodeDetails(context.getParent().getDatanodeDetails())
+              .setContext(context)
+              .build();
+          break;
+        case SHUTDOWN:
+          break;
+        }
+        endpointTaskForState.put(state, endPointTask);
```
adoroszlai: Nit: no need to store `null` task for shutdown state.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
```diff
@@ -49,14 +52,62 @@
...
+  /**
+   * Initialize end point tasks corresponding to each end point,
+   * each end point state.
+   */
+  private void initEndPointTask() {
+    endpointTasks = new HashMap<String, Map<EndPointStates,
+        Callable<EndPointStates>>>();
+    for (EndpointStateMachine endpoint : connectionManager.getValues()) {
+      Map<EndPointStates, Callable<EndPointStates>> endpointTaskForState
+          = new HashMap<EndPointStates, Callable<EndPointStates>>();
```
adoroszlai: Nit: `EnumMap` is better suited for maps keyed by enums.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
```diff
@@ -98,35 +148,14 @@ public void execute(ExecutorService executor) {
...
-          .build();
-    case SHUTDOWN:
-      break;
-    default:
-      throw new IllegalArgumentException("Illegal Argument.");
+
+  private Callable<EndPointStates> getEndPointTask(
+      EndpointStateMachine endpoint) {
+    if (endpointTasks.containsKey(endpoint.toString())) {
+      return endpointTasks.get(endpoint.toString()).get(endpoint.getState());
```
adoroszlai: Can you please explain why a string-based map is better than using the endpoint itself as a key?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
linyiqun: Yes, it should be state, it's my typo. Here we will create endpoint task for each state type.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
linyiqun: Have removed SHUTDOWN switch case.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
linyiqun: Have addressed this comment

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
linyiqun: I just wan to use a more simple identity to indicate the different  Endpoint instance. So I use the endpoint address. But I have updated this to use Endpoint itself now.

adoroszlai:
Thanks @linyiqun for addressing review items.

@anuengineer @nandakumar131 This change is based on a TODO comment left in the code long time ago.  Can you please review if the approach is right?

nandakumar131:
Previously `getEndPointTask` would return `null` for `EndPointStates.SHUTDOWN` and the `execute` method handles this `null`.
With this change we will start throwing `IllegalArgumentException`, we should retain the old behavior of returning `null` for `EndPointStates.SHUTDOWN` .

nandakumar131:
Thanks @linyiqun for the contribution and thanks @adoroszlai for the review. I will merge this shortly.

linyiqun:
Thanks @nandakumar131  for review and merge.

PR #481, opened 2020-01-23
HDDS-2926. Intermittent failure in TestRecon due to thread timing.

swagle:
- Removed Thread.sleep and actually made deterministic calls.
- Refactored service start/stop code to make it test friendly.

https://issues.apache.org/jira/browse/HDDS-2926

## How was this patch tested?
The integration test passed.

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/ReconContainerDBProvider.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/ReconContainerDBProvider.java
@@ -69,7 +69,7 @@ public class ReconContainerDBProvider implements Provider<DBStore> {
     if (lastKnownContainerKeyDb != null) {
       LOG.info("Last known container-key DB : {}",
           lastKnownContainerKeyDb.getAbsolutePath());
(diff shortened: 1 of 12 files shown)


swagle:
Thanks @avijayanhwx for suggestions.

swagle:
UT failure is related will update patch with a fix.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/ReconContainerDBProvider.java
```diff
@@ -79,8 +81,8 @@ public DBStore get() {
     return dbStore;
   }
 
-  private static DBStore getDBStore(OzoneConfiguration configuration,
-                            ReconUtils reconUtils, String dbName) {
+  static DBStore initializeDBStore(OzoneConfiguration configuration,
```
avijayanhwx: This can be private.

adoroszlai:
Thanks for working on this @swagle.  The change LGTM, and the test passed locally multiple times.  But now there is some conflict in the branch.

avijayanhwx:
LGTM +1.

PR #487, opened 2020-01-24
HDDS-2883. Change the default client settings accordingly with change in default chunk size.

bharatviswa504:

This Jira proposes to change the following values

ozone.client.stream.buffer.flush.size - 64MB -> 16MB

ozone.client.stream.buffer.max.size - 128 MB -> 32MB

ozone.scm.chunk.size - 16MB -> 4MB

 

In this way, the client-side holding buffer size will be reduced, and data is transferred at smaller size intervals.

https://issues.apache.org/jira/browse/HDDS-2883

## How was this patch tested?

Multiple teragen tests are run on a 6 node cluster with these values and along changes from HDDS-2920 and HDDS-2921 we have seen teragen completed successfully.


bharatviswa504:
> Thanks @bharatviswa504 for working on this. Shouldn't the default values be changed in `OzoneConfigKeys`, too?

Thank You @adoroszlai for the review. Yes, we need to update. Updated in the latest commit.

bharatviswa504:
Thank You @adoroszlai and @arp7 for the review.

PR #490, opened 2020-01-24
HDDS-2916. OM HA cli getserviceroles not working.

bharatviswa504:

When om admin commands are run on a non-HA cluster, return error message to client.

**Reason for this change:**
1. ozone.om.ratis.enable is a temporary config, once OM starts using by default ratis this config will be gone. So, if ozone.om.service.ids is not defined then technically it is not an HA cluster, so returned an error message. If the user passed service id is matching with service ID in the config, the command will work as usual.

https://issues.apache.org/jira/browse/HDDS-2916

## How was this patch tested?

Added smoke test.


bharatviswa504:

PR #498, opened 2020-01-27
HDDS-2940. mkdir : create key table entries for intermediate directories in the path

supratimdeka:

for MKDIR operation - add separate directory entry in OM key table for each directory that appears in the path passed to mkdir.

https://issues.apache.org/jira/browse/HDDS-2940

## How was this patch tested?

unit tests.
TestOzoneFileInterfaces#testDirectory is modified to assert that prefix directories get created.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
@@ -44,6 +44,7 @@ import org.apache.hadoop.ozone.OzoneConfigKeys;
 import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.TestDataUtil;
 import org.apache.hadoop.ozone.client.OzoneBucket;
+import org.apache.hadoop.ozone.om.OMMetadataManager;
 import org.apache.hadoop.ozone.om.OMMetrics;
 import org.apache.hadoop.ozone.om.helpers.OmKeyArgs;
 import org.apache.hadoop.ozone.om.helpers.OzoneFileStatus;
@@ -224,37 +225,55 @@ public class TestOzoneFileInterfaces {
   @Test
   public void testDirectory() throws IOException {
     String leafName = RandomStringUtils.randomAlphanumeric(5);
+    OMMetadataManager metadataManager = cluster.getOzoneManager()
+        .getMetadataManager();
 
     String lev1dir = "abc";
     Path lev1path = createPath("/" + lev1dir);
+    String lev1key = metadataManager.getOzoneDirKey(volumeName, bucketName,
+        o3fs.pathToKey(lev1path));
     String lev2dir = "def";
     Path lev2path = createPath("/" + lev1dir + "/" + lev2dir);
-    FileStatus lev1status;
-    FileStatus lev2status;
+    String lev2key = metadataManager.getOzoneDirKey(volumeName, bucketName,
+        o3fs.pathToKey(lev2path));
+
+    FileStatus rootChild;
     FileStatus rootstatus;
     FileStatus leafstatus;
-    FileStatus rootChild;
 
     Path leaf = createPath("/" + lev1dir + "/" + lev2dir + "/" + leafName);
+    String leafKey = metadataManager.getOzoneDirKey(volumeName, bucketName,
+        o3fs.pathToKey(leaf));
 
     // verify prefix directories and the leaf, do not already exist
-    lev1status = verifyDirectory(lev1path);
-    lev2status = verifyDirectory(lev2path);
-    assertTrue((lev1status == null) && (lev2status == null));
-    leafstatus = verifyDirectory(leaf);
-    assertTrue(leafstatus == null);
+    assertTrue(metadataManager.getKeyTable().get(lev1key) == null);
+    assertTrue(metadataManager.getKeyTable().get(lev2key) == null);
+    assertTrue(metadataManager.getKeyTable().get(leafKey) == null);
 
     assertTrue("Makedirs returned with false for the path " + leaf,
         fs.mkdirs(leaf));
 
     // verify the leaf directory got created.
-    leafstatus = verifyDirectory(leaf);
+    leafstatus = getDirectoryStat(leaf);
     assertTrue(leafstatus != null);
 
     if (cluster.getOzoneManager().createPrefixEntries()) {
+      FileStatus lev1status;
+      FileStatus lev2status;
+
       // verify prefix directories got created when creating the leaf directory.
-      lev1status = verifyDirectory(lev1path);
-      lev2status = verifyDirectory(lev2path);
+      String kn = metadataManager.getKeyTable().get(lev2key).getKeyName();
+      assertTrue(kn.equals("abc/def/"));
+      assertTrue(metadataManager
+          .getKeyTable()
+          .get(lev1key)
+          .getKeyName().equals("abc/"));
+      assertTrue(metadataManager
+          .getKeyTable()
+          .get(lev2key)
+          .getKeyName().equals("abc/def/"));
+      lev1status = getDirectoryStat(lev1path);
+      lev2status = getDirectoryStat(lev2path);
       assertTrue((lev1status != null) && (lev2status != null));
       rootChild = lev1status;
     } else {
@@ -262,7 +281,7 @@ public class TestOzoneFileInterfaces {
     }
 
     // check the root directory
-    rootstatus = verifyDirectory(createPath("/"));
+    rootstatus = getDirectoryStat(createPath("/"));
     assertTrue(rootstatus != null);
 
     // root directory listing should contain the lev1 prefix directory
@@ -426,7 +445,7 @@ public class TestOzoneFileInterfaces {
    * @return null indicates FILE_NOT_FOUND, else the FileStatus
    * @throws IOException
    */
-  private FileStatus verifyDirectory(Path path) throws IOException {
+  private FileStatus getDirectoryStat(Path path) throws IOException {
 
     FileStatus status = null;
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
@@ -262,8 +262,6 @@ public class TestOzoneFileInterfaces {
       FileStatus lev2status;
 
       // verify prefix directories got created when creating the leaf directory.
-      String kn = metadataManager.getKeyTable().get(lev2key).getKeyName();
-      assertTrue(kn.equals("abc/def/"));
       assertTrue(metadataManager
           .getKeyTable()
           .get(lev1key)
@@ -456,7 +454,6 @@ public class TestOzoneFileInterfaces {
     }
     assertTrue("The created path is not directory.", status.isDirectory());
 
-    assertTrue(status.isDirectory());
     assertEquals(FsPermission.getDirDefault(), status.getPermission());
     verifyOwnerGroup(status);
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
@@ -18,6 +18,7 @@
 
 package org.apache.hadoop.fs.ozone;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
@@ -222,27 +223,52 @@ public class TestOzoneFileInterfaces {
 
   @Test
   public void testDirectory() throws IOException {
-    String dirPath = RandomStringUtils.randomAlphanumeric(5);
-    Path path = createPath("/" + dirPath);
-    assertTrue("Makedirs returned with false for the path " + path,
-        fs.mkdirs(path));
+    String leafName = RandomStringUtils.randomAlphanumeric(5);
 
-    FileStatus status = fs.getFileStatus(path);
-    assertTrue("The created path is not directory.", status.isDirectory());
+    String lev1dir = "abc";
+    Path lev1path = createPath("/" + lev1dir);
+    String lev2dir = "def";
+    Path lev2path = createPath("/" + lev1dir + "/" + lev2dir);
+    FileStatus lev1status;
+    FileStatus lev2status;
+    FileStatus rootstatus;
+    FileStatus leafstatus;
+    FileStatus rootChild;
 
-    assertTrue(status.isDirectory());
-    assertEquals(FsPermission.getDirDefault(), status.getPermission());
-    verifyOwnerGroup(status);
+    Path leaf = createPath("/" + lev1dir + "/" + lev2dir + "/" + leafName);
 
-    assertEquals(0, status.getLen());
+    // verify prefix directories and the leaf, do not already exist
+    lev1status = verifyDirectory(lev1path);
+    lev2status = verifyDirectory(lev2path);
+    assertTrue((lev1status == null) && (lev2status == null));
+    leafstatus = verifyDirectory(leaf);
+    assertTrue(leafstatus == null);
 
+    assertTrue("Makedirs returned with false for the path " + leaf,
+        fs.mkdirs(leaf));
+
+    // verify the leaf directory got created.
+    leafstatus = verifyDirectory(leaf);
+    assertTrue(leafstatus != null);
+
+    if (cluster.getOzoneManager().createPrefixEntries()) {
+      // verify prefix directories got created when creating the leaf directory.
+      lev1status = verifyDirectory(lev1path);
+      lev2status = verifyDirectory(lev2path);
+      assertTrue((lev1status != null) && (lev2status != null));
+      rootChild = lev1status;
+    } else {
+      rootChild = leafstatus;
+    }
+
+    // check the root directory
+    rootstatus = verifyDirectory(createPath("/"));
+    assertTrue(rootstatus != null);
+
+    // root directory listing should contain the lev1 prefix directory
     FileStatus[] statusList = fs.listStatus(createPath("/"));
     assertEquals(1, statusList.length);
-    assertEquals(status, statusList[0]);
-
-    fs.getFileStatus(createPath("/"));
-    assertTrue("Root dir (/) is not a directory.", status.isDirectory());
-    assertEquals(0, status.getLen());
+    assertEquals(rootChild, statusList[0]);
   }
 
   @Test
@@ -392,5 +418,31 @@ public class TestOzoneFileInterfaces {
     } else {
       return new Path(relativePath);
     }
+  }
+
+  /**
+   * verify that a directory exists and is initialized correctly.
+   * @param path of the directory
+   * @return null indicates FILE_NOT_FOUND, else the FileStatus
+   * @throws IOException
+   */
+  private FileStatus verifyDirectory(Path path) throws IOException {
+
+    FileStatus status = null;
+
+    try {
+      status = fs.getFileStatus(path);
+    } catch (FileNotFoundException e) {
+      return null;
+    }
+    assertTrue("The created path is not directory.", status.isDirectory());
+
+    assertTrue(status.isDirectory());
+    assertEquals(FsPermission.getDirDefault(), status.getPermission());
+    verifyOwnerGroup(status);
+
+    assertEquals(0, status.getLen());
+
+    return status;
   }
 }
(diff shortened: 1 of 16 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -84,6 +92,70 @@ public static OMDirectoryResult verifyFilesInPath(
...
+      } else {
+        missing.add(pathName);
+      }
+      keyPath = keyPath.getParent();
+    }
+
+    return missing;
+  }
+
+  private static OmKeyInfo getKeyInfo(
```
avijayanhwx: There seems to be no usage for this method.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -84,6 +92,70 @@ public static OMDirectoryResult verifyFilesInPath(
...
+      String dbDirKeyName = omMetadataManager.getOzoneDirKey(volumeName,
+          bucketName, pathName);
+
+      if (omMetadataManager.getKeyTable().isExist(dbKeyName)) {
+        // Found a file in the given path.
+        String errorMsg = "File " + dbKeyName + " exists with same name as " +
+            " directory in path : " + pathName;
+        throw new IOException(errorMsg);
+      } else if (omMetadataManager.getKeyTable().isExist(dbDirKeyName)) {
+        // Found a directory in the given path. Higher parents must exist.
```
avijayanhwx: Do we need to worry about this assumption not being true in clusters already deployed?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -126,31 +129,41 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+      List<String> missingParents = OMFileRequest.getMissingParents(
+          omMetadataManager, volumeName, bucketName, keyPath.getParent());
 
       OmBucketInfo omBucketInfo = omMetadataManager.getBucketTable().get(
           omMetadataManager.getBucketKey(volumeName, bucketName));
+      long baseObjId = OMFileRequest.getObjIdFromTxId(transactionLogIndex);
+      long objectCount = 1;
+
       OmKeyInfo dirKeyInfo = null;
       if (omDirectoryResult == FILE_EXISTS ||
```
avijayanhwx: I believe this can be moved up to after the "verifyFilesInPath" call. Also, should we repeat the presence of intermediate directory path in keyTable in getMissingParents() again?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
supratimdeka: because clusters are not already deployed, if we introduce this change now, we do not have that problem. Of course we are referring to customer deployments here and not any internal test setups, right?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
supratimdeka: thanks for pointing it out, will remove. a relic from an earlier version of my patch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
supratimdeka: will move the code snippet as suggested.
would there be any benefit checking for the presence of intermediate directory in keyTable again? the objective of the function, as the name suggests is focussed on getting parents in the path which are missing.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
avijayanhwx: Yes, agreed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
hanishakoneru: I think we can combine the two functions to avoid doing DB reads twice. We could probably change OMDirectoryResult structure to include the result + missing dirs.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -159,8 +172,24 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-            volumeName, bucketName, keyName, keyArgs, transactionLogIndex);
+        dirKeyInfo = createDirectoryKeyInfo(ozoneManager, keyName, keyArgs,
+            baseObjId, transactionLogIndex);
+
+        for (String missingKey : missingParents) {
+          LOG.debug("missing parent {}", missingKey);
+          // what about keyArgs for parent directories? TODO
+          OmKeyInfo parentKeyInfo = createDirectoryKeyInfoNoACL(ozoneManager,
+              missingKey, keyArgs, baseObjId + objectCount,
+              transactionLogIndex);
```
hanishakoneru: How are we calculating the objectID here? Is this to ensure that the objectIDs are unique?
I am not sure what the effects would be as we set the transactionID as the objectID everywhere else.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -228,10 +289,8 @@ private OmKeyInfo createDirectoryKeyInfo(OzoneManager ozoneManager,
         .setReplicationType(HddsProtos.ReplicationType.RATIS)
         .setReplicationFactor(HddsProtos.ReplicationFactor.ONE)
         .setFileEncryptionInfo(encryptionInfo.orNull())
-        .setAcls(OzoneAclUtil.fromProtobuf(keyArgs.getAclsList()))
-        .setObjectID(transactionLogIndex)
-        .setUpdateID(transactionLogIndex)
-        .build();
+        .setObjectID(objectId)
+        .setUpdateID(0);
```
hanishakoneru: Let's not set the updateID here as it will be set later.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMDirectoryCreateResponse.java
```diff
@@ -39,11 +40,17 @@
...
       LoggerFactory.getLogger(OMDirectoryCreateResponse.class);
   private OmKeyInfo dirKeyInfo;
+  private List<OmKeyInfo> parentKeyInfos;
+  private boolean createPrefix;
 
   public OMDirectoryCreateResponse(@Nullable OmKeyInfo dirKeyInfo,
-      @Nonnull OMResponse omResponse) {
+      @Nullable List<OmKeyInfo> parentKeyInfos,
+      @Nonnull OMResponse omResponse,
+      boolean createPrefix) {
```
hanishakoneru: We can avoid passing the createPrefix param and use parentKeyInfos != null check only instead.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
avijayanhwx: Yes, @hanishakoneru's suggestion makes sense if we want to avoid doing the same reads twice.

supratimdeka: Thanks, makes sense. Will make the suggested change.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
supratimdeka: the base object ID for every create is the transaction log index left shifted by 8 bits (tx log index << 8). OMFileRequest.getObjIdFromTxId() implements this logic. I have replaced all 3 places in the code where a object id is generated, with this call. intermediate directories are created iteratively by doing a +1 starting from the base object ID. that will guarantee each object id is unique - given that the base object id starts with the unique transaction log index.

I will put in asserts to ensure there is no overflow.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
supratimdeka: sure, will make the change. any specific reason why it might not be a good idea?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMDirectoryCreateResponse.java
supratimdeka: the createPrefix represents a configuration parameter. whereas parentKeyInfos can actually be null.

yes, things will work as you suggested. I can remove the createPrefix argument. But given that it represents a config control, I would wish to retain it.

If you insist, then I will go ahead and remove it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
hanishakoneru: We could still end up with the same objectID for keys and buckets/ volumes.

For example, Lets say Trxn 1 creates dir1/dir2/dir3. So objectID for dir1 would be 1 << 8 + 1 i.e. 257.
Now let's say Trxn 257 comes in and creates bucket1. Then, dir1 and bucket1 would have the same objectID.

Let's change the objectID for volumes, buckets and prefixInfo's too.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMDirectoryCreateResponse.java
hanishakoneru: Sure. I am ok with keeping the createPrefix param.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
hanishakoneru: No strong reason. Just to avoid setting it twice.

But this leads me to another thought. Instead of having a separate setObjectID and setUpdateID methods, we could have 1 method which updates both. This would ensure that we never miss setting the updateID when setting objectID. Just a thought.. unrelated to this Jira.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
supratimdeka: sure, will use the same object id scheme for volume and bucket.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -229,12 +243,69 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+    for (String missingKey : missingParents) {
+      long nextObjId = baseObjId + objectCount;
+      if (nextObjId > maxObjId) {
+        throw new OMException("Too many directories in path. Exceeds limit of "
+            + maxLevels + ". Unable to create directory: " + keyName
+            + " in volume/bucket: " + volumeName + "/" + bucketName,
+            INVALID_KEY_NAME);
+      }
+
+      LOG.debug("missing parent {}", missingKey);
```
hanishakoneru: Can we update this debug log message to indicate that the missing parent directory is being created/ added to KeyTable?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -64,24 +72,79 @@ public static OMDirectoryResult verifyFilesInPath(
...
+  }
+
+  /**
+   * Get the valid base object id given the transaction id.
+   * @param id of the transaction
+   * @return base object id allocated against the transaction
+   */
+  public static long getObjIDFromTxId(long id) {
+    return getObjIdRangeFromTxId(id).getLeft();
+  }
```
hanishakoneru: We use this function multiple times (more than getObjIdRangeFromTxId()). Wouldn't it be better to directly execute shift operation here instead of calling getObjIdRangeFromTxId()?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
```diff
@@ -340,4 +385,30 @@ private Path createPath(String relativePath) {
...
+    FileStatus status = null;
+
+    try {
+      status = fs.getFileStatus(path);
+    } catch (FileNotFoundException e) {
+      return null;
+    }
+    assertTrue("The created path is not directory.", status.isDirectory());
+
+    assertTrue(status.isDirectory());
```
hanishakoneru: Double check of status.isDirectory()

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
```diff
@@ -217,27 +219,70 @@ private void verifyOwnerGroup(FileStatus fileStatus) {
...
+          .getKeyName().equals("abc/"));
+      assertTrue(metadataManager
+          .getKeyTable()
+          .get(lev2key)
+          .getKeyName().equals("abc/def/"));
+      lev1status = getDirectoryStat(lev1path);
+      lev2status = getDirectoryStat(lev2path);
+      assertTrue((lev1status != null) && (lev2status != null));
+      rootChild = lev1status;
+    } else {
```
hanishakoneru: This case would never be reached in this test suit as createPrefix config is true by default and we are not overriding it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -229,12 +243,69 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+  }
+
+  private List<OmKeyInfo> getAllParentInfo(OzoneManager ozoneManager,
+      KeyArgs keyArgs, List<String> missingParents, long trxnLogIndex)
+      throws OMException, IOException {
+    OMMetadataManager omMetadataManager = ozoneManager.getMetadataManager();
+    List<OmKeyInfo> missingParentInfos = new ArrayList<>();
+
+    ImmutablePair<Long, Long> objIdRange = OMFileRequest
+        .getObjIdRangeFromTxId(trxnLogIndex);
```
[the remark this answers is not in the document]
supratimdeka: the code around the previous call doesn't need to bother about the object range, only the base id. the code here requires the object range.
worry about an extra call during directory create might be premature in the overall cpu profile?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -161,14 +167,18 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
       // cannot create a directory with the given key.
+      OMFileRequest.OMPathInfo omPathInfo =
+          OMFileRequest.verifyFilesInPath(omMetadataManager, volumeName,
+              bucketName, keyName, keyPath);
       OMFileRequest.OMDirectoryResult omDirectoryResult =
-          OMFileRequest.verifyFilesInPath(omMetadataManager,
-          volumeName, bucketName, keyName, Paths.get(keyName));
+          omPathInfo.getDirectoryResult();
+      List<String> missingParents = omPathInfo.getMissingParents();
+      long baseObjId = OMFileRequest.getObjIDFromTxId(trxnLogIndex);
```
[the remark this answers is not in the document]
supratimdeka: ok

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
supratimdeka: ok

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
supratimdeka: ok

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
supratimdeka: will remove

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
supratimdeka: I'll fix that in a subsequent patch. this is a comment from @mukul1987 as well. we will include both cases in the parametrised test.

supratimdeka:
Regarding fixing object ID logic in 
OMKeyRenameRequest#validateAndUpdateCache

Not sure why object id should get updated during a rename. that would not be correct for an inode number in a filesystem.  Ozone FS should not be any different either. What about object semantics? In the code, I do not see object id manipulation. 

The addAcl() and setAcl() in PrefixManagerImpl, I could not find references to object id getting set/updated here. Can you please clarify what needs to be done there.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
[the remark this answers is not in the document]
hanishakoneru: No I was just thinking from code cleaning perspective.

hanishakoneru:
> Not sure why object id should get updated during a rename. that would not be correct for an inode number in a filesystem. Ozone FS should not be any different either. What about object semantics? In the code, I do not see object id manipulation.
Sure. Will file a Jira to fix this separately.

> The addAcl() and setAcl() in PrefixManagerImpl, I could not find references to object id getting set/updated here. Can you please clarify what needs to be done there.
In PrefixManagerImpl lines 318 and 368 we set the objectID. Are you rebased with the latest master? These changes went in as part of HDDS-2962.

supratimdeka:
Setting the object id in addAcl/setAcl inside PrefixManagerImpl, requires a separate discussion. Discussing with @arp7, it looked like we would end up with separate object ids for the same prefix - one in the prefix table and another in the key table. That would be a problem.

Is this understanding correct? If so, I would want to defer the fix from the current patch. Pl give your opinion. Any other blocking comments?

hanishakoneru:
@supratimdeka , sure we can fix the Prefix ACL ObjectID in a separate Jira.
I am +1 for merge pending CI.

supratimdeka:
CI is successful with the latest patch. Request to merge, this will unblock HDDS-2941.

PR #499, opened 2020-01-27
HDDS-2855. Recon getContainers API should return a maximum of 1000 co…

vivekratnavel:
…ntainers by default.

Change the default limit of Recon API responses to 1000 items for any API request.

https://issues.apache.org/jira/browse/HDDS-2855

## How was this patch tested?

I tested the patch by hitting the API - `http://localhost:9888/api/containers/1/keys` and `http://localhost:9888/api/containers` to check if the response defaults to 1000 items and it did.

Screen Shot 2020-01-27 at 1 21 46 PM


avijayanhwx:
LGTM +1.

vivekratnavel:
Hi @elek 

Thanks for committing this change. To answer your question, this patch just limits the number of items returned by default when there is no limit query param in the API request. Previously, when there is no limit specified, the API would return all the results. Now, the API returns just 1000 results by default. Once we support proper pagination, the UI will have page numbers listed at the bottom.

PR #501, opened 2020-01-29
HDDS-2944. Handle replay of KeyCommitRequest and DirectoryCreateRequest

hanishakoneru:

To ensure that key commit operations are idempotent, compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

OMKeyCommitRequest and OMDirectoryCreateRequest are made idempotent in this Jira.

https://issues.apache.org/jira/browse/HDDS-2944

## How was this patch tested?

Unit tests are added.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
@@ -143,7 +143,7 @@ public class OMDirectoryCreateRequest extends OMKeyRequest {
     boolean acquiredLock = false;
     IOException exception = null;
     OMClientResponse omClientResponse = null;
-    Result result = null;
+    Result result = Result.FAILURE;
     try {
       // check Acl
       checkKeyAcls(ozoneManager, volumeName, bucketName, keyName,
@@ -210,9 +210,6 @@ public class OMDirectoryCreateRequest extends OMKeyRequest {
         omClientResponse = new OMDirectoryCreateResponse(
             createReplayOMResponse(omResponse));
       } else {
-        if (result == null) {
-          result = Result.FAILURE;
-        }
         exception = ex;
         omClientResponse = new OMDirectoryCreateResponse(
             createErrorOMResponse(omResponse, exception));
@@ -237,18 +234,23 @@ public class OMDirectoryCreateRequest extends OMKeyRequest {
     switch (result) {
     case SUCCESS:
       omMetrics.incNumKeys();
-      LOG.debug("Directory created. Volume:{}, Bucket:{}, Key:{}", volumeName,
-          bucketName, keyName);
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Directory created. Volume:{}, Bucket:{}, Key:{}", volumeName,
+            bucketName, keyName);
+      }
       break;
     case REPLAY:
-      LOG.debug("Replayed Transaction {} ignored. Request: {}", trxnLogIndex,
-          createDirectoryRequest);
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Replayed Transaction {} ignored. Request: {}", trxnLogIndex,
+            createDirectoryRequest);
+      }
       break;
     case DIRECTORY_ALREADY_EXISTS:
       LOG.error("Directory already exists. Volume:{}, Bucket:{}, Key{}",
           volumeName, bucketName, keyName, exception);
       break;
     case FAILURE:
+      omMetrics.incNumCreateDirectoryFails();
       LOG.error("Directory creation failed. Volume:{}, Bucket:{}, Key{}. " +
           "Exception:{}", volumeName, bucketName, keyName, exception);
       break;
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
@@ -194,7 +194,7 @@ public class OMDirectoryCreateRequest extends OMKeyRequest {
         String dirKey = omMetadataManager.getOzoneDirKey(volumeName,
             bucketName, keyName);
         OmKeyInfo dbKeyInfo = omMetadataManager.getKeyTable().get(dirKey);
-        if (isReplay(ozoneManager, dbKeyInfo.getUpdateID(), trxnLogIndex)) {
+        if (isReplay(ozoneManager, dbKeyInfo, trxnLogIndex)) {
           throw new OMReplayException();
         } else {
           result = Result.DIRECTORY_ALREADY_EXISTS;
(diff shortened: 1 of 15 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -126,16 +143,17 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
       // check Acl
       checkKeyAcls(ozoneManager, volumeName, bucketName, keyName,
           IAccessAuthorizer.ACLType.CREATE, OzoneObj.ResourceType.KEY);
 
       // Check if this is the root of the filesystem.
       if (keyName.length() == 0) {
-        return new OMDirectoryCreateResponse(null,
-            omResponse.setCreateDirectoryResponse(
-                CreateDirectoryResponse.newBuilder()).build());
+        throw new OMException("Directory create failed. Cannot create " +
```
bharatviswa504: Is this the expected behavior, as the original code has returned normal OMResponse.

hanishakoneru: I thought the previous response was wrong. We should return the correct error code, right?

arp7: Yeah the change seems correct to me. @bharatviswa504 what do you think?

bharatviswa504: Looks like hdfs does the same

Tried on a HDFS cluster.
[root@ozone-test-bh-2 ~]# hdfs dfs -mkdir hdfs://ozone-test-bh-6.vpc.cloudera.com:8020
mkdir: Call From ozone-test-bh-2.vpc.cloudera.com/10.65.51.138 to ozone-test-bh-6.vpc.cloudera.com:8020 failed on connection exception: java.net.ConnectException: Connection refused; For more details see:  http://wiki.apache.org/hadoop/ConnectionRefused
[root@ozone-test-bh-2 ~]# hdfs dfs -mkdir hdfs://ozone-test-bh-2.vpc.cloudera.com:8020
mkdir: `Item: hdfs://ozone-test-bh-2.vpc.cloudera.com:8020 parent's path is null. This can happen if mkdir is called on root, so there's no parent.': No such file or directory
[root@ozone-test-bh-2 ~]# hdfs dfs -mkdir hdfs://ozone-test-bh-2.vpc.cloudera.com:8020/
mkdir: `hdfs://ozone-test-bh-2.vpc.cloudera.com:8020/': File exists

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -160,52 +178,86 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
     } catch (IOException ex) {
-      exception = ex;
-      omClientResponse = new OMDirectoryCreateResponse(null,
-          createErrorOMResponse(omResponse, exception));
+      if (ex instanceof OMReplayException) {
+        result = Result.REPLAY;
+        omClientResponse = new OMDirectoryCreateResponse(
+            createReplayOMResponse(omResponse));
+      } else {
+        if (result == null) {
```
bharatviswa504: Minor NIT:
we can set result = Result.Failure in the beginning, in that way we don't need if check.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -160,52 +178,86 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-    } else {
-      LOG.error("CreateDirectory failed for Key: {} in volume/bucket:{}/{}",
-          keyName, volumeName, bucketName, exception);
-      omMetrics.incNumCreateDirectoryFails();
-      return omClientResponse;
+      LOG.debug("Directory created. Volume:{}, Bucket:{}, Key:{}", volumeName,
+          bucketName, keyName);
+      break;
+    case REPLAY:
+      LOG.debug("Replayed Transaction {} ignored. Request: {}", trxnLogIndex,
```
bharatviswa504: Minor:
isDebugEnabled check.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -160,52 +178,86 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
                 dirKeyInfo.getKeyName())),
-            new CacheValue<>(Optional.of(dirKeyInfo), transactionLogIndex));
+            new CacheValue<>(Optional.of(dirKeyInfo), trxnLogIndex));
+
+        omClientResponse = new OMDirectoryCreateResponse(omResponse.build(),
+            dirKeyInfo);
+      } else {
+        // omDirectoryResult == DIRECTORY_EXITS
+        // Check if this is a replay of ratis logs
+        String dirKey = omMetadataManager.getOzoneDirKey(volumeName,
```
bharatviswa504: Instead of having DB read here, can we modify VerifyFilesInpath also have key info and check that here.
In this way, we don't do twice DB reads. 

```

else if (omMetadataManager.getKeyTable().isExist(dbDirKeyName)) {
        // Found a directory in the given path.
        // Check if this is actual directory or a directory in the given path
        if (dbDirKeyName.equals(dirNameFromDetails)) {
          return new Result[OMDirectoryResult.DIRECTORY_EXISTS, omMetadataManager.getKeyTable(dirNameFromDetails);
        }
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -160,52 +178,86 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+    case REPLAY:
+      LOG.debug("Replayed Transaction {} ignored. Request: {}", trxnLogIndex,
+          createDirectoryRequest);
+      break;
+    case DIRECTORY_ALREADY_EXISTS:
+      LOG.error("Directory already exists. Volume:{}, Bucket:{}, Key{}",
+          volumeName, bucketName, keyName, exception);
+      break;
+    case FAILURE:
+      LOG.error("Directory creation failed. Volume:{}, Bucket:{}, Key{}. " +
```
bharatviswa504: omMetrics.incNumCreateDirectoryFails(); this is missing.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
hanishakoneru: This change breaks Contract tests too.
So instead of throwing an exception, I am returning a Result code. Currently, the createDirectory client API returns a void. And added a TODO to return a boolean instead and then we can pass false when directory already exists.

bharatviswa504:

PR #504, opened 2020-01-29
HDDS-2953. Handle replay of S3 requests

hanishakoneru:

To ensure that S3 operations is idempotent, compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

In this Jira, the following requests are made idempotent:

S3InitiateMultipartUploadRequest
S3MultipartUploadCommitPartRequest
S3MultipartUploadCompleteRequest
S3MultipartUploadAbortRequest

https://issues.apache.org/jira/browse/HDDS-2953

## How was this patch tested?

Will add unit tests in next commit

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
@@ -453,6 +453,11 @@ public final class OmUtils {
     return dirFile;
   }
 
+  public static RepeatedOmKeyInfo prepareKeyForDelete(OmKeyInfo keyInfo,
+      RepeatedOmKeyInfo repeatedOmKeyInfo) throws IOException {
+    return prepareKeyForDelete(keyInfo, repeatedOmKeyInfo, 0L);
+  }
+
   /**
    * Prepares key info to be moved to deletedTable.
    * 1. It strips GDPR metadata from key info
@@ -460,13 +465,20 @@ public final class OmUtils {
    * implies that no entry for the object key exists in deletedTable so we
    * create a new instance to include this key, else we update the existing
    * repeatedOmKeyInfo instance.
+   * 3. Set the updateID to the transactionLogIndex
    * @param keyInfo args supplied by client
    * @param repeatedOmKeyInfo key details from deletedTable
+   * @param trxnLogIndex For Multipart keys, this is the transactionLogIndex
+   *                     of the MultipartUploadAbort request which needs to
+   *                     be set as the updateID of the partKeyInfos.
+   *                     For regular Key deletes, this value should be passed
+   *                     as 0 as the updateID is already set.
    * @return {@link RepeatedOmKeyInfo}
    * @throws IOException if I/O Errors when checking for key
    */
   public static RepeatedOmKeyInfo prepareKeyForDelete(OmKeyInfo keyInfo,
-      RepeatedOmKeyInfo repeatedOmKeyInfo) throws IOException{
+      RepeatedOmKeyInfo repeatedOmKeyInfo, long trxnLogIndex)
+      throws IOException {
     // If this key is in a GDPR enforced bucket, then before moving
     // KeyInfo to deletedTable, remove the GDPR related metadata and
     // FileEncryptionInfo from KeyInfo.
@@ -475,6 +487,11 @@ public final class OmUtils {
       keyInfo.getMetadata().remove(OzoneConsts.GDPR_ALGORITHM);
       keyInfo.getMetadata().remove(OzoneConsts.GDPR_SECRET);
       keyInfo.clearFileEncryptionInfo();
+    }
+
+    // If trxnLogIndex is 0, then the updateID has already been set.
+    if (trxnLogIndex != 0) {
+      keyInfo.setUpdateID(trxnLogIndex);
     }
 
     if(repeatedOmKeyInfo == null) {
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
@@ -453,11 +453,6 @@ public final class OmUtils {
     return dirFile;
   }
 
-  public static RepeatedOmKeyInfo prepareKeyForDelete(OmKeyInfo keyInfo,
-      RepeatedOmKeyInfo repeatedOmKeyInfo) throws IOException {
-    return prepareKeyForDelete(keyInfo, repeatedOmKeyInfo, 0L);
-  }
-
   /**
    * Prepares key info to be moved to deletedTable.
    * 1. It strips GDPR metadata from key info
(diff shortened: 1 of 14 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3InitiateMultipartUploadRequest.java
```diff
@@ -118,6 +127,28 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
 
+      // Check if this transaction is a replay.
+      // We check only the KeyTable here and not the OpenKeyTable. In case
+      // this transaction is a replay but the transaction was not committed
+      // to the KeyTable, then we recreate the key in OpenKey table. This is
+      // okay as all the subsequent transactions would also be replayed and
+      // the openKey table would eventually reach the same state.
+      // The reason we do not check the OpenKey table is to avoid a DB read
+      // in regular non-replay scenario.
+      String dbKeyName = omMetadataManager.getOzoneKey(volumeName, bucketName,
```
bharatviswa504: Looks like checking keyTable is not required here. I feel we can ignore it here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadAbortRequest.java
```diff
@@ -116,36 +122,51 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-            new CacheValue<>(Optional.absent(), transactionLogIndex));
-        omMetadataManager.getMultipartInfoTable().addCacheEntry(
-            new CacheKey<>(multipartKey),
-            new CacheValue<>(Optional.absent(), transactionLogIndex));
       }
 
-      omClientResponse = new S3MultipartUploadAbortResponse(multipartKey,
-          multipartKeyInfo,
+      // Check the OpenKeyTable if this transaction is a replay of ratis logs.
+      if (isReplay(ozoneManager, omKeyInfo.getUpdateID(), trxnLogIndex)) {
```
bharatviswa504: How this check will help here, as we don't set updateID for open key OmKeyInfo.

Example:
 T1 -> MPU Initiate
T2 -> MPU abort

Now replay, T2 will not be considered as a replay. As OmKeyInfo will have updateID as 1.

hanishakoneru: If the openKey was created after this transaction, then it is safe to assume that this is a replay.

hanishakoneru: I see what you are saying now. MultipartUploadID would be unique even for the same keys for different initiates. Is that correct? If yes, ya we can remove this check here.

hanishakoneru:
Thanks for the review @bharatviswa504 . I removed the replay checks from Initiate and Abort requests.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -453,20 +453,32 @@ public static File createOMDir(String dirPath) {
...
+   *                     of the MultipartUploadAbort request which needs to
+   *                     be set as the updateID of the partKeyInfos.
+   *                     For regular Key deletes, this value should be passed
+   *                     as 0 as the updateID is already set.
    * @return {@link RepeatedOmKeyInfo}
    * @throws IOException if I/O Errors when checking for key
    */
   public static RepeatedOmKeyInfo prepareKeyForDelete(OmKeyInfo keyInfo,
-      RepeatedOmKeyInfo repeatedOmKeyInfo) throws IOException{
+      RepeatedOmKeyInfo repeatedOmKeyInfo, long trxnLogIndex)
```
bharatviswa504: This looks brittle, if someone mistakenly sent for normal key deletes trxnLogIndex we will update it. (I think no harm)
So, make this logic generic for all requests, instead of special cases, and remove the above method.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3InitiateMultipartUploadRequest.java
```diff
@@ -118,6 +126,11 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       validateBucketAndVolume(omMetadataManager, volumeName, bucketName);
 
+      // We do not check if this transaction is a replay here to avoid extra
+      // DB reads. Even if this transaction is replayed, in
+      // S3MultipartUploadComplete request, we would delete this entry from
```
bharatviswa504: Can we mention this as TODO? (As still it is not addressed)

bharatviswa504:
T1 _ Initiate MPU K1
T2 - P1 Create
T3 - P1 Commit
T4 - P2 Create
T5 - P2 Create
T6 - K1 Complete

Replay
T1 will create entry in openKey and MPU Key
T2 - replay as we have entry in Key table
T3 - replay as we have entry in Key table
T4 - replay as we have entry in Key table
T5 - replay as we have entry in Key table
T6 - replay (We should delete entry created from T1, will be handled in a new Jira)

Only Replay T6 We shall consider as replay by checking key table

If replay from T3
T3- replay as we have entry in Key table
T4 - replay as we have entry in Key table
T5 - replay as we have entry in Key table
T6- Replay


T1 _ Initiate MPU K1
T2 - P1 Create
T3 - P1 Commit
T4 - P2 Create
T5 - P2 Create
T6 - MPU Abort K1

Replay All
T1 will create entry in openKey and MPU Key
T2- will create entry in openKey
T3- will delete entry in openKey and move to MPU P1
T4- will create entry in openKey
T5- will delete entry in openKey and move to MPU P1,P2
T6- We do as normal op and move to delete key table and delete entries

Only Replay T6 (This is fine, as we cannot find it as replay)
We throw NO SUCH MPU as no entries will be there in openKeyTable.

If replay from T3
T3- KEY_NOT_FOUND, as we don’t have entry in openKey and MPU
T4 - replay as we have no entry in Key table
T5 - NO_SUCH_MPU error as no entry in MPU table
T6- We throw No_SUCH_MULTIPART_UPLOAD as no entry for iniitiateMPU in openKey(In this case also we should delete entries if any)

Just came up with few scenarios and checked how we have. We should handle these cases. (Mentioning here so that it will help in looking further)

And one more thing observed is with replay our metrics might get completely messed up. (As in few cases for optimization we have not to find the transaction is replay and in few cases there is no chance in finding a transaction as a replay.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3InitiateMultipartUploadRequest.java
hanishakoneru: Added a TODO in MultipartComplete request.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
hanishakoneru: Done.

bharatviswa504:

PR #505, opened 2020-01-30
HDDS-2956. Handle Replay of AllocateBlock request

hanishakoneru:

To ensure that allocate block operations are idempotent, compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

OMAllocateBlockRequest is made idempotent in this Jira.

https://issues.apache.org/jira/browse/HDDS-2956

## How was this patch tested?

Unit test added.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
@@ -197,7 +197,8 @@ public class OMAllocateBlockRequest extends OMKeyRequest {
             throw new OMReplayException();
           }
         }
-        throw new OMException("Open Key not found " + openKeyName, KEY_NOT_FOUND);
+        throw new OMException("Open Key not found " + openKeyName,
+            KEY_NOT_FOUND);
       }
 
       // Check if this transaction is a replay of ratis logs.
@@ -220,8 +221,8 @@ public class OMAllocateBlockRequest extends OMKeyRequest {
 
       // Add to cache.
       omMetadataManager.getOpenKeyTable().addCacheEntry(
-          new CacheKey<>(openKeyName), new CacheValue<>(Optional.of(openKeyInfo),
-              trxnLogIndex));
+          new CacheKey<>(openKeyName),
+          new CacheValue<>(Optional.of(openKeyInfo), trxnLogIndex));
 
       omResponse.setAllocateBlockResponse(AllocateBlockResponse.newBuilder()
           .setKeyLocation(blockLocation).build());
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
@@ -25,6 +25,7 @@ import java.util.Map;
 
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
+import org.apache.hadoop.ozone.om.exceptions.OMReplayException;
 import org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerDoubleBufferHelper;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer;
 import org.apache.hadoop.util.Time;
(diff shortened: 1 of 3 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -160,71 +160,98 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
 
-      omKeyInfo = omMetadataManager.getOpenKeyTable().get(openKey);
-      if (omKeyInfo == null) {
-        throw new OMException("Open Key not found " + openKey, KEY_NOT_FOUND);
+      openKeyInfo = omMetadataManager.getOpenKeyTable().get(openKeyName);
+      if (openKeyInfo == null) {
+        // Check if this transaction is a replay of ratis logs.
+        // If the Key was already committed and this transaction is being
+        // replayed, we should ignore this transaction.
+        String ozoneKey = omMetadataManager.getOzoneKey(volumeName,
```
bharatviswa504: I think, this check will help only when Key is created and Committed. But it will not help in the scenario:
T1 - Key Create
T2 - allocate block
T3 - Key Commit
T4- Key Delete
When replay, we will still throw an error, as it will not be in the openKeyTable. So, I think it is okay not to get from key table and do a check. As we cannot 100% find replay transactions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -134,8 +135,7 @@ public OMRequest preExecute(OzoneManager ozoneManager) throws IOException {
 
   @Override
   public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
-      long transactionLogIndex,
-      OzoneManagerDoubleBufferHelper ozoneManagerDoubleBufferHelper) {
+      long trxnLogIndex, OzoneManagerDoubleBufferHelper omDoubleBufferHelper) {
```
bharatviswa504: Just one minor comment, can we eliminate renaming variable changes, as this is increasing jira scope and code line changes. As with focussed change, it will be simple.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -160,71 +160,98 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+        omMetrics.incNumBlockAllocateCallFails();
+        exception = ex;
+        omClientResponse = new OMAllocateBlockResponse(createErrorOMResponse(
+            omResponse, exception));
+        LOG.error("Allocate Block failed. Volume:{}, Bucket:{}, OpenKey:{}. " +
+            "Exception:{}", volumeName, bucketName, openKeyName, exception);
+      }
     }
 
     auditLog(auditLogger, buildAuditMessage(OMAction.ALLOCATE_BLOCK, auditMap,
```
bharatviswa504: We want to do auditLog when it is not a replay(best try), like how we did in other key patches.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
hanishakoneru: My intention was to capture the maximum number of cases. If Key is deleted, we cannot help it. And this does not affect the regular case when the openKey exists. So it is not a performance concern.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
hanishakoneru: Sure. I will avoid it in future. The variable names were so big that it was causing unnecessary increase in the lines of code. I was trying to club code cleanup along with these changes. But I will reserve it for separate Jira in the future.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
hanishakoneru: done.

bharatviswa504:

PR #507, opened 2020-01-30
HDDS-2936. Hive queries fail at readFully

bshashikant:

It fixes in the retry path of ozone client where length of data written was getting updated incorrectly while write in KeyouputStream.

https://issues.apache.org/jira/browse/HDDS-2936

## How was this patch tested?
The existing test "TestCloseContainerHandlingByClient" was failing because the issue. All these tests are executing successfully with the fix. The patch also tested in a real deployment where HIve workload was run and all hive queries succeed now.

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
@@ -75,7 +75,13 @@ public class KeyOutputStream extends OutputStream {
   private FileEncryptionInfo feInfo;
   private final Map<Class<? extends Throwable>, RetryPolicy> retryPolicyMap;
   private int retryCount;
+  // how much of data is actually written yet to underlying stream
   private long offset;
+  // how much data has been ingested into the stream
+  private long writeOffset;
+  // whether an exception is encountered while write and whole write could
+  // not succeed
+  private boolean isException;
   private final BlockOutputStreamEntryPool blockOutputStreamEntryPool;
 
   /**
@@ -134,6 +140,8 @@ public class KeyOutputStream extends OutputStream {
     this.retryPolicyMap = HddsClientUtils.getRetryPolicyByException(
         maxRetryCount, retryInterval);
     this.retryCount = 0;
+    this.isException = false;
+    this.writeOffset = 0;
   }
 
   /**
@@ -188,6 +196,7 @@ public class KeyOutputStream extends OutputStream {
       return;
     }
     handleWrite(b, off, len, false);
+    writeOffset += len;
   }
 
   private void handleWrite(byte[] b, int off, long len, boolean retry)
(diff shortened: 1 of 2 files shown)


hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
```diff
@@ -201,7 +209,11 @@ private void handleWrite(byte[] b, int off, long len, boolean retry)
         // comes via Exception path.
         int writeLen = Math.min((int) len, (int) current.getRemaining());
         long currentPos = current.getWrittenDataLength();
-        writeToOutputStream(current, retry, len, b, writeLen, off, currentPos);
+        // writeLen will be updated based on whether the write was succeeded
+        // or if it sees an exception, how much the actual write was
+        // acknowledged.
+        writeLen = writeToOutputStream(current, retry, len, b, writeLen,
```
mukul1987: writeLen, goes in both as an agrument as well as a return value to the function. Can we change this to something like expectedWritelen, and writtenLen

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
```diff
@@ -356,6 +370,7 @@ private void handleRetry(IOException exception, long len) throws IOException {
         msg = "Retry request failed. " + action.reason;
         LOG.error(msg, exception);
       }
+      isException = true;
       throw new IOException(msg, exception);
```
mukul1987: Lets put 
isException &  throw new IOException(msg, exception); into one function. So that both of these are always updated together.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
```diff
@@ -484,6 +499,9 @@ public void close() throws IOException {
     closed = true;
     try {
       handleFlushOrClose(StreamAction.CLOSE);
+      if (!isException) {
+        Preconditions.checkArgument(writeOffset == offset);
```
fapifta: At this point in the code I am not the fan of throwing an IllegalArgumentException, wouldn't it be better to keep this as an IOException with a specific explaining message?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
fapifta: Actually I would say this whole duplication is because RetryPolicy.shouldRetry has two fail states, one when it fails with an exception, and one when it fails with a fail status without an exception by contract.
The handling of this underlying speciality at once should go into some method, and perhaps we should use that method to throw an exception in all the cases, and handle the exception only in this code, and the utility method over RetryPolicy.shouldRetry can do the logging and tinkering with the exception message.
What do you think?

bshashikant:
Thanks @fapifta and @mukul1987 for the review comments. I have moved all the exception throw logic in a common function where the exception state is set.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
bshashikant: I would prefer to have the precondition check intact, as its just in place to validate the logic. If its not met, the code is buggy. Also, throwing Runtime exception is okay as this exception does not need be caught and handled anywhere in general.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
bshashikant: Addressed in the latest patch..

bshashikant:
> @bshashikant , which unit test can reproduce this issue?

The issue can be reproduced with TestCloseContainerHandlingByClient#testBlockWriteViaRatis.

adoroszlai:
> The issue can be reproduced with TestCloseContainerHandlingByClient#testBlockWriteViaRatis.

Thanks @bshashikant for pointing this out.  I can confirm that the test is fixed by this patch.

Now that integration tests are enabled, can you please go ahead and remove `@Ignore` from `TestCloseContainerHandlingByClient`?

adoroszlai:
> I don't see the test is ignored in my branch

@bshashikant It's ignored in `master`, so a merge from `master` is needed, then it can be unignored.

bharatviswa504:
Thank You @bshashikant for the contribution, @adoroszlai and @fapifta for the reviews.

PR #508, opened 2020-01-30
HDDS-2950. Upgrade jetty to the latest 9.4 release

elek:

The jetty which is used by web interfaces of Ozone is from the September of 2018.

Since then many bug and security fixes added to the Jetty project. I would suggest to use the latest jetty (from January of 2020).

As `HttpServer2` (hadoop 3.2 class) has a strong dependency on the older version of Jetty (it uses SessionManager which is removed), it seems to be easier to clone HttpServer2 and move it to the Ozone project.

It provides us the flexibility:

  * To upgrade jetty independent from the Hadoop releases
  * To remove unused features and make our HTTP stack more reliable (eg. Yarn, WebHDFS part of the code)
  * To customize our HTTP server usage (eg. SPNEGO should be ignore for S3 REST)
  * To support ozone style configuration and maintain all our configuration est
  * To add additional insights/metrics to our own HttpServer
  * To simplify the current server initialization (current BaseHttpServer class of hdds/ozone is a wrapper around the Hadoop class, but we can combine the two functionalities)

## What is included in this patch

 * `HttpServer2` and lightweight dependencies are cloned from Hadoop **3.2**
 * HADOOP-16152 (Jetty upgrade since 3.2 release)
 * Checkstyle / findbugs (to follow our coding standards)
 * As we have >5 http related classes I moved them to a dedicated package in the *framework* project.
 * **The main logic has not been modified by this patch ** This is mainly code organization and code import.

https://issues.apache.org/jira/browse/HDDS-2950

## How was this patch tested?

There are related unit tests, but also checked with docker-compose based cluster + checking the ui pages from browser.


arp7:
@smengcl can you take a look at this pull request? This is forking the `HttpWebServer` from Hadoop and applying your changes to port to 9.4.26.

I think long term it makes sense to fork the Hadoop HttpWebServer.

smengcl:
> @smengcl can you take a look at this pull request? This is forking the `HttpWebServer` from Hadoop and applying your changes to port to 9.4.26.
> 
> I think long term it makes sense to fork the Hadoop HttpWebServer.

Sure. Looking at this.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/HttpServer2.java
```diff
@@ -0,0 +1,1701 @@
...
+ * This class is a fork of the old HttpServer. HttpServer exists for
+ * compatibility reasons. See HBASE-10336 for more details.
+ */
+@InterfaceAudience.Private
+@InterfaceStability.Evolving
+public final class HttpServer2 implements FilterContainer {
+  public static final Logger LOG = LoggerFactory.getLogger(HttpServer2.class);
+
+  private static final String HTTP_SCHEME = "http";
+  public static final String HTTPS_SCHEME = "https";
```
smengcl: This can be changed to `private` as well?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/HttpServer2.java
```diff
@@ -0,0 +1,1701 @@
...
+  public static final int HTTP_SELECTOR_COUNT_DEFAULT = -1;
+  // idle timeout in milliseconds
+  public static final String HTTP_IDLE_TIMEOUT_MS_KEY =
+      "hadoop.http.idle_timeout.ms";
+  public static final int HTTP_IDLE_TIMEOUT_MS_DEFAULT = 10000;
+  public static final String HTTP_TEMP_DIR_KEY = "hadoop.http.temp.dir";
+
+  public static final String FILTER_INITIALIZER_PROPERTY
+      = "ozone.http.filter.initializers";
+
```
smengcl: `hadoop.http.sni.host.check.enabled` is added in [HADOOP-16718](https://issues.apache.org/jira/browse/HADOOP-16718) in trunk. We might want it here as well?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/HttpServer2.java
```diff
@@ -0,0 +1,1701 @@
...
+     * Specify the SSL configuration to load. This API provides an alternative
+     * to keyStore/keyPassword/trustStore.
+     */
+    public Builder setSSLConf(Configuration sslCnf) {
+      this.sslConf = sslCnf;
+      return this;
+    }
+
+    public Builder setPathSpec(String[] pathSpec) {
+      this.pathSpecs = pathSpec.clone();
```
smengcl: I see `.clone()` is added here. For safe practice?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/HttpServer2.java
elek: sure, done.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/HttpServer2.java
elek: Yes, it was a findbugs violation (internal representation is leaked...)

elek:
Thank you very much @smengcl 

As a background: as of now we use hadoop 3.2. The most safest import is to use exactly the same `HttpServer2.java` what we have in the latest hadoop 3.2 release.

I tried to use trunk HttpServer2 from the Hadoop **trunk** but found multiple changes related the authorization broke S3 gateway. 

After that I decided to keep the same `HttpServer2` which is used now (with Jetty patch included).

Therefore (as a rule of thumb) I prefer to import here just minimal changes and import newer patches in separated jira (where we can carefully test s3g).

I checked the suggested patches (thanks for the suggestions):

 * HADOOP-16727 It's a small null check, can be good to add it as soon as possible. I added it to the patch.
 * HADOOP-16398 Prometheus support is backported from Ozone to Hadoop. We don't need the patch as we already have the original solution. (Later we can simplify our initialization to make it more similar to the other default servlets but it's a bigger refactor.) 
 * HADOOP-16718 I tried to understand why this is required and didn't find any information. If you have something, let me know. (Might be required for webhdfs which shouldn't be supported in our side.) For me it seems to be optional, and s3g and other ozone components work well based on the tests. Unless we have a strong reason I would prefer to keep the default Jetty behavior (and use one less config).

smengcl:
> Thank you very much @smengcl
> 
> As a background: as of now we use hadoop 3.2. The most safest import is to use exactly the same `HttpServer2.java` what we have in the latest hadoop 3.2 release.
> 
> I tried to use trunk HttpServer2 from the Hadoop **trunk** but found multiple changes related the authorization broke S3 gateway.
> 
> After that I decided to keep the same `HttpServer2` which is used now (with Jetty patch included).
> 
> Therefore (as a rule of thumb) I prefer to import here just minimal changes and import newer patches in separated jira (where we can carefully test s3g).
> 
> I checked the suggested patches (thanks for the suggestions):
> 
> * HADOOP-16727 It's a small null check, can be good to add it as soon as possible. I added it to the patch.
> * HADOOP-16398 Prometheus support is backported from Ozone to Hadoop. We don't need the patch as we already have the original solution. (Later we can simplify our initialization to make it more similar to the other default servlets but it's a bigger refactor.)
> * HADOOP-16718 I tried to understand why this is required and didn't find any information. If you have something, let me know. (Might be required for webhdfs which shouldn't be supported in our side.) For me it seems to be optional, and s3g and other ozone components work well based on the tests. Unless we have a strong reason I would prefer to keep the default Jetty behavior (and use one less config).

Thanks for digging into this. I'm good with using same `HttpServer2.java` as Hadoop 3.2.

The two new commits lgtm.

As for HADOOP-16718, I believe the scope is larger than WebHDFS. Web UIs with HTTPS using jetty may all be impacted under certain cases. In short, this can cause connection failure if:

> ... the server's JKS file has a private/public key/cert pairing that is valid but it also has another trustedCertEntry certificate that has the hostname in subjectAltName extension, the trusted cert gets picked.
> This triggers an internal failure to determine a common cipher to use and the server will return the following exception to the client:
> fatal error: 40: no cipher suites in common

For details, please take a look at the link I sent over to you on Slack.

smengcl:
I am good with backporting HADOOP-16718 in another jira. So far lgtm+1

Thanks @elek @arp7

xiaoyuyao:
Thanks @elek for the update. The latest change LGTM, +1.

PR #509, opened 2020-01-30
HDDS-2960. Collect output of crashed tests

adoroszlai:

Save output of crashed unit/integration tests to the diagnostic bundle, similar to how it's done for failed tests.  Result files (`TEST-*.xml` and `*.txt`) may or may not be available, so we need to search by test name.  This also lets us avoid including files for successful tests of which failed tests are prefixes (eg. `TestContainerStateMachine` and `TestContainerStateMachineFailures`).

https://issues.apache.org/jira/browse/HDDS-2960

## How was this patch tested?

Temporarily introduced fake crashes (`System.exit`) in some tests, verified that `unit.zip` includes output for these:

https://github.com/adoroszlai/hadoop-ozone/runs/417135911


adoroszlai:
@elek may I ask you to review?  This might help getting info on why a test crashes, eg. TestSCMNodeManager in [this run](https://github.com/apache/hadoop-ozone/pull/538/checks?check_run_id=435309306).

adoroszlai:

PR #510, opened 2020-01-30
HDDS-2958. Handle replay of OM Volume ACL requests

hanishakoneru:

To ensure that volume acl operations are idempotent, compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

OMVolumeAclRequests (Add, Remove and Set ACL requests) are made idempotent in this Jira.

https://issues.apache.org/jira/browse/HDDS-2958

## How was this patch tested?

Unit test added.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/acl/OMVolumeAclRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/acl/OMVolumeAclRequest.java
@@ -25,10 +25,12 @@ import org.apache.hadoop.ozone.om.OMMetadataManager;
 import org.apache.hadoop.ozone.om.OMMetrics;
 import org.apache.hadoop.ozone.om.OzoneManager;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
+import org.apache.hadoop.ozone.om.exceptions.OMReplayException;
 import org.apache.hadoop.ozone.om.helpers.OmVolumeArgs;
 import org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerDoubleBufferHelper;
 import org.apache.hadoop.ozone.om.request.OMClientRequest;
 import org.apache.hadoop.ozone.om.response.OMClientResponse;
+import org.apache.hadoop.ozone.om.response.volume.OMVolumeAclOpResponse;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMResponse;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer;
@@ -57,8 +59,7 @@ public abstract class OMVolumeAclRequest extends OMClientRequest {
 
   @Override
   public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
-      long transactionLogIndex,
-      OzoneManagerDoubleBufferHelper ozoneManagerDoubleBufferHelper) {
+      long trxnLogIndex, OzoneManagerDoubleBufferHelper omDoubleBufferHelper) {
     // protobuf guarantees volume and acls are non-null.
     String volume = getVolumeName();
     List<OzoneAcl> ozoneAcls = getAcls();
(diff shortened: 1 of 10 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/acl/OMVolumeSetAclRequest.java
```diff
@@ -80,29 +80,36 @@ public String getVolumeName() {
...
-  void onComplete(IOException ex) {
-    if (ex == null) {
-      if (LOG.isDebugEnabled()) {
-        LOG.debug("Set acls: {} to volume: {} success!",
-            getAcls(), getVolumeName());
-      }
-    } else {
-      LOG.error("Set acls {} to volume {} failed!",
-          getAcls(), getVolumeName(), ex);
+  void onComplete(Result result, IOException ex, long trxnLogIndex) {
```
bharatviswa504: Now no use of exception here. Unused param

bharatviswa504: Same applies for all requests.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/acl/OMVolumeAclRequest.java
```diff
@@ -73,53 +74,71 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
       try {
         omVolumeAclOp.apply(ozoneAcls, omVolumeArgs);
       } catch (OMException ex) {
-        result = false;
+        applyAcl = false;
       }
 
-      if (result) {
+      if (applyAcl) {
+        omVolumeArgs.setUpdateID(trxnLogIndex);
```
bharatviswa504: We need to set updateID even in the case of applyAcl false.
Scenario:
1. Add Acl T1
2. Add existing Acl T2

Now replay T2 it will not be considered as replay as updateID will be still 1.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/acl/OMVolumeSetAclRequest.java
```diff
@@ -80,29 +80,36 @@ public String getVolumeName() {
...
+    switch (result) {
+    case SUCCESS:
+      LOG.debug("Set acls: {} to volume: {} success!", getAcls(),
```
bharatviswa504: Removed isDebugEnabled.
Need same in other requests also. (I see it is missing in other requests)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/acl/OMVolumeSetAclRequest.java
hanishakoneru: We are using it to log the exception in case of FAILURE.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/acl/OMVolumeAclRequest.java
hanishakoneru: We do not update the DB in case applyAcl is false. So this would add an unnecessary DB op. Tradeoff is replay transaction would also check and reach the same decision - applyAcl false. So I think it should be ok here. Thoughts?

bharatviswa504: No when replay it will not be considered as replay as UpdateID will be less than transaction ID. As we are doing best effort to find all the replay cases, I think we can set updateID when applyAcl false and it will help in replay.

> So this would add an unnecessary DB op.

We are using DoubleBuffer and Batch, so we should be fine here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/acl/OMVolumeSetAclRequest.java
bharatviswa504: Yes, missed it.

hanishakoneru:
Thanks @bharatviswa504 for the review. Update the patch to address your comments.

arp7:
Kicked off retest.

hanishakoneru:
Thank you @bharatviswa504 and @arp7 for the reviews. 
CI run is green. I will merge this PR.

PR #512, opened 2020-01-30
HDDS-2959. Handle replay of OM Key ACL requests

hanishakoneru:

To ensure that Key acl operations are idempotent, compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

OMKeyAclRequests (Add, Remove and Set ACL requests) are made idempotent in this Jira.

https://issues.apache.org/jira/browse/HDDS-2959

## How was this patch tested?

Unit tests are added

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/acl/OMKeyAclResponse.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/acl/OMKeyAclResponse.java
@@ -21,12 +21,10 @@ package org.apache.hadoop.ozone.om.response.key.acl;
 import java.io.IOException;
 
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 
 import org.apache.hadoop.ozone.om.OMMetadataManager;
 import org.apache.hadoop.ozone.om.helpers.OmKeyInfo;
 import org.apache.hadoop.ozone.om.response.OMClientResponse;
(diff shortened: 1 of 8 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/acl/OMKeyAclResponse.java
```diff
@@ -36,27 +34,30 @@
...
+  public OMKeyAclResponse(@Nonnull OMResponse omResponse) {
+    super(omResponse);
+  }
+
   @Override
   public void addToDBBatch(OMMetadataManager omMetadataManager,
       BatchOperation batchOperation) throws IOException {
 
-    // If response status is OK and success is true, add to DB batch.
-    if (getOMResponse().getStatus() == OzoneManagerProtocolProtos.Status.OK &&
```
arp7: Was this check moved somewhere else?

hanishakoneru: Yes Arpit. OmClientResponse#checkAndUpdateDB() performs this check before calling addToDBBatch(). OMDoubleBuffer now calls checkAndUpdateDB() instead of addToDBBatch() directly.

hanishakoneru:
Thank you @bharatviswa504 and @arp7 for the reviews. 
CI run is green. I will merge this PR.

PR #513, opened 2020-01-31
HDDS-2962. Handle replay of OM Prefix ACL requests

hanishakoneru:

To ensure that Prefix acl operations are idempotent, compare the transactionID with the objectID and updateID to make sure that the transaction is not a replay. If the transactionID <= updateID, then it implies that the transaction is a replay and hence it should be skipped.

OMPrefixAclRequests (Add, Remove and Set ACL requests) are made idempotent in this Jira.

https://issues.apache.org/jira/browse/HDDS-2962

## How was this patch tested?

Unit tests added.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/PrefixManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/PrefixManagerImpl.java
@@ -22,6 +22,7 @@ import org.apache.hadoop.ozone.om.exceptions.OMException;
 import org.apache.hadoop.ozone.om.helpers.OmBucketInfo;
 import org.apache.hadoop.ozone.om.helpers.OmPrefixInfo;
 import org.apache.hadoop.ozone.om.helpers.OzoneAclUtil;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
 import org.apache.hadoop.ozone.security.acl.OzoneObj;
 import org.apache.hadoop.ozone.security.acl.RequestContext;
 import org.apache.hadoop.ozone.util.RadixNode;
@@ -106,7 +107,8 @@ public class PrefixManagerImpl implements PrefixManager {
       OmPrefixInfo prefixInfo =
           metadataManager.getPrefixTable().get(prefixPath);
 
-      OMPrefixAclOpResult omPrefixAclOpResult = addAcl(obj, acl, prefixInfo);
+      OMPrefixAclOpResult omPrefixAclOpResult = addAcl(obj, acl, prefixInfo,
+          0L);
 
       return omPrefixAclOpResult.isSuccess();
     } catch (IOException ex) {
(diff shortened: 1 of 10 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/PrefixManagerImpl.java
```diff
@@ -176,9 +177,10 @@ public boolean setAcl(OzoneObj obj, List<OzoneAcl> acls) throws IOException {
       OmPrefixInfo prefixInfo =
           metadataManager.getPrefixTable().get(prefixPath);
 
-      OMPrefixAclOpResult omPrefixAclOpResult = setAcl(obj, acls, prefixInfo);
+      OMPrefixAclOpResult omPrefixAclOpResult = setAcl(obj, acls, prefixInfo,
+          0L);
```
arp7: Dumb q. - why pass 0 here?

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMPrefixAclRequest.java
```diff
@@ -0,0 +1,101 @@
...
+
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests Prefix ACL requests.
+ */
+public class TestOMPrefixAclRequest extends TestOMKeyRequest {
+
+  @Test
+  public void testReplayRequest() throws Exception {
```
arp7: Is one unit test case sufficient? Are we covering all the new code paths?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/PrefixManagerImpl.java
hanishakoneru: This code path was used for the non-ratis case before the two code paths were merged. This should go away after code cleanup. 
@bharatviswa504 please correct me if I am wrong.

arp7: You're right.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMPrefixAclRequest.java
hanishakoneru: I think it should be. This test case does not cover the change in logging on completion. But all the PrefixAcl requests go through a common code path for replay check which is covered here.
Do we want to add a test for proto changes?

arp7: Yes we must have tests for proto [de]serialization. However I am not sure we have been rigorous about adding them in the past.

Let's file a separate newbie jira to add missing protobuf [de]serialization unit tests.

PR #514, opened 2020-01-31
HDDS-2957. listBuckets result should include the exact match of bucketPrefix

smengcl:

I noticed that `OzoneVolume.listBuckets(String bucketPrefix)` behaves differently than `ObjectStore.listVolumes(String volumePrefix)` with a given prefix: `listBuckets` ignores the `bucketPrefix` in the result if it is an exact match; `listVolumes` doesn't.

e.g. If we have a bucket named `bucket-12345`, currently `OzoneVolume.listBuckets("bucket-12345")` will NOT return bucket-12345 in its result.

https://issues.apache.org/jira/browse/HDDS-2957

## How was this patch tested?

Added unit test `testListBucketWithPrefixAsExactBucketName` and passed.

Undoing the fix in `OmMetadataManagerImpl` will cause the last step in the unit test to fail.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -1432,6 +1432,32 @@ public abstract class TestOzoneRpcClientAbstract {
   }
 
   @Test
+  public void testListBucketWithPrefixAsExactBucketName()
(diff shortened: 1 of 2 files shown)


cxorm:
Thanks @smengcl for the work.
Is this issue duplicated with [the PR](https://github.com/apache/hadoop-ozone/pull/430) ?

smengcl:
> Thanks @smengcl for the work.
> Is this issue duplicated with [the PR](https://github.com/apache/hadoop-ozone/pull/430) ?
> If so and this patch is committed, I will close my PR.

Yep they looks like the same fix. Closing this one.

PR #516, opened 2020-02-01
HDDS-2923 Add fall-back protection for rack awareness in pipeline creation.

timmylicheng:
#HDDS-2923 Add fall-back protection for rack awareness in pipeline creation.

https://issues.apache.org/jira/browse/HDDS-2923

## How was this patch tested?
UT. Test in actual env.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
@@ -248,15 +248,18 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
           nodesRequired, SCMException.ResultCodes.INVALID_CAPACITY);
     }
 
-    // Assume rack awareness is enabled.
-    boolean rackAwareness = true;
+    // Assume rack awareness is not enabled.
+    boolean rackAwareness = false;
     List <DatanodeDetails> results = new ArrayList<>(nodesRequired);
     // Since nodes are widely distributed, the results should be selected
     // base on distance in topology, rack awareness and load balancing.
     List<DatanodeDetails> exclude = new ArrayList<>();
     // First choose an anchor nodes randomly
     DatanodeDetails anchor = chooseNode(healthyNodes);
-    if (anchor == null) {
+    if (anchor != null) {
+      results.add(anchor);
+      exclude.add(anchor);
+    } else {
       LOG.warn("Unable to find healthy node for anchor(first) node.");
       throw new SCMException("Unable to find anchor node.",
           SCMException.ResultCodes.FAILED_TO_FIND_SUITABLE_NODE);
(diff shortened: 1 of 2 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -241,29 +269,30 @@ boolean meetCriteria(DatanodeDetails datanodeDetails, int nodesRequired) {
...
+    DatanodeDetails nextNode = chooseNodeBasedOnRackAwareness(
         healthyNodes, exclude,
         nodeManager.getClusterNetworkTopologyMap(), anchor);
-    if (nodeOnDifferentRack == null) {
-      LOG.warn("Pipeline Placement: Unable to find 2nd node on different " +
-          "racks that meets the criteria. Required nodes: {}, Found nodes:" +
-          " {}", nodesRequired, results.size());
-      throw new SCMException("Unable to find required number of nodes.",
-          SCMException.ResultCodes.FAILED_TO_FIND_SUITABLE_NODE);
+    if (nextNode == null) {
```
xiaoyuyao: NIT: I think we can change to check positive instead and let the for loop handle the remaining node picking naturally. 
```
boolean rackAwareness = false;
if (nextNode != null) {
  rackAwareness = true;
  results.add(nextNode);
  exclude.add(nextNode);
}
```

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelinePlacementPolicy.java
```diff
@@ -83,10 +83,36 @@ public void testChooseNodeBasedOnRackAwareness() {
     DatanodeDetails nextNode = placementPolicy.chooseNodeBasedOnRackAwareness(
         healthyNodes, new ArrayList<>(PIPELINE_PLACEMENT_MAX_NODES_COUNT),
         topologyWithDifRacks, anchor);
+    Assert.assertNotNull(nextNode);
     Assert.assertFalse(anchor.getNetworkLocation().equals(
         nextNode.getNetworkLocation()));
   }
 
+  @Test
```
xiaoyuyao: can we add a test simple test case with topology that all nodes on the same rack with topology aware and fall back enabled?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Updated.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelinePlacementPolicy.java
timmylicheng: I add a test. Please check the latest commit.

timmylicheng:
Acceptance test failed due to:
ERROR: Test execution of /home/runner/work/hadoop-ozone/hadoop-ozone/hadoop-ozone/dist/target/ozone-0.5.0-SNAPSHtOT/compose/ozonesecure is FAILED!!!!

Doesn't seem to be related to this change. @xiaoyuyao Do you have any ideas on the ozonesecure test failure? Thanks.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -220,50 +243,57 @@ boolean meetCriteria(DatanodeDetails datanodeDetails, int nodesRequired) {
...
+    if (nextNode != null) {
+      // Rack awareness is detected.
+      rackAwareness = true;
+      results.add(nextNode);
+      exclude.add(nextNode);
+    } else {
+      LOG.debug("Pipeline Placement: Unable to find 2nd node on different " +
+          "rack based on rack awareness.");
     }
     if (LOG.isDebugEnabled()) {
```
ChenSammi: Better to move this if statement into if (nextNode != null).

timmylicheng: Sure. Updated.

ChenSammi:
+1

PR #521, opened 2020-02-03
HDDS-2592 Add Datanode command to allow the datanode to persist its admin state

sodonnel:

When the datanode state is change in SCMNodeManager a Datanode Command is created to transmit the new datanode state to the datanode. The datanode will then persist this state in the "datanodeID.yaml" file.

The current datanode.yaml file simply persists a DatanodeDetails object in YAML format. Therefore two new fields have been added to DatanodeDetails to support this:

persistedOpState,
persistedOpStateExpiryEpochSec

The first is the operational state and the second is the seconds from the epoch the setting should expire (for maintenance mode).

When a datanode starts up, it will read these values and supply them in the heartbeat to SCM.

If SCM notices the DN was registered before, SCM is seen as the source of truth and will update the DN state to that set in SCM if they differ.

If the DN was never registered with SCM (eg if SCM was restarted), then the DN state is seen as the source of truth and SCM will be updated accordingly.

This change also moves the state expiry time into the NodeStatus object, so it now carries operationalState, healthState and State expiry.

Note that the DN does not do anything with the persisted operational state. They are simply persisted and reported in the heartbeat. In this respect the DNs are being used as a form of distributed storage for the operational state so the decommission process can survive a restart.

https://issues.apache.org/jira/browse/HDDS-2592

## How was this patch tested?

Tested manually using a docker build. This change still needs unit tests added if the current approach is agreed upon.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/DatanodeIdYaml.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/DatanodeIdYaml.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.MapUtils;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 
@@ -82,6 +83,12 @@ public final class DatanodeIdYaml {
           .setIpAddress(datanodeDetailsYaml.getIpAddress())
           .setHostName(datanodeDetailsYaml.getHostName())
           .setCertSerialId(datanodeDetailsYaml.getCertSerialId());
+      if (datanodeDetailsYaml.getPersistedOpState() != null) {
+        builder.setPersistedOpState(HddsProtos.NodeOperationalState.valueOf(
+            datanodeDetailsYaml.getPersistedOpState()));
+      }
+      builder.setPersistedOpStateExpiry(
+          datanodeDetailsYaml.getPersistedOpStateExpiryEpochSec());
 
       if (!MapUtils.isEmpty(datanodeDetailsYaml.getPortDetails())) {
         for (Map.Entry<String, Integer> portEntry :
@@ -105,6 +112,8 @@ public final class DatanodeIdYaml {
     private String ipAddress;
     private String hostName;
     private String certSerialId;
+    private String persistedOpState;
+    private long persistedOpStateExpiryEpochSec = 0;
     private Map<String, Integer> portDetails;
 
     public DatanodeDetailsYaml() {
@@ -113,11 +122,15 @@ public final class DatanodeIdYaml {
 
     private DatanodeDetailsYaml(String uuid, String ipAddress,
                                 String hostName, String certSerialId,
+                                String persistedOpState,
+                                long persistedOpStateExpiryEpochSec,
                                 Map<String, Integer> portDetails) {
       this.uuid = uuid;
       this.ipAddress = ipAddress;
       this.hostName = hostName;
       this.certSerialId = certSerialId;
+      this.persistedOpState = persistedOpState;
+      this.persistedOpStateExpiryEpochSec = persistedOpStateExpiryEpochSec;
       this.portDetails = portDetails;
     }
 
@@ -135,6 +148,14 @@ public final class DatanodeIdYaml {
 
     public String getCertSerialId() {
       return certSerialId;
+    }
+
+    public String getPersistedOpState() {
+      return persistedOpState;
+    }
+
+    public long getPersistedOpStateExpiryEpochSec() {
+      return persistedOpStateExpiryEpochSec;
     }
 
     public Map<String, Integer> getPortDetails() {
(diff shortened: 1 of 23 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/DatanodeIdYaml.java
```diff
@@ -82,6 +83,12 @@ public static DatanodeDetails readDatanodeIdFile(File path)
           .setIpAddress(datanodeDetailsYaml.getIpAddress())
           .setHostName(datanodeDetailsYaml.getHostName())
           .setCertSerialId(datanodeDetailsYaml.getCertSerialId());
+      if (datanodeDetailsYaml.getPersistedOpState() != null) {
+        builder.setPersistedOpState(HddsProtos.NodeOperationalState.valueOf(
+            datanodeDetailsYaml.getPersistedOpState()));
+      }
+      builder.setPersistedOpStateExpiry(
```
anuengineer: Shouldn't this be also inside the if != null ? , I am going to presume that even if we have no data, datanodeDetailsYaml.getPersistedOpStateExpiryEpochSec is going to return zero, but since the expiry is a long, JVM would initialize it to zero by default, would it not?. So curious why this check is outside the if.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/commandhandler/SetNodeOperationalStateCommandHandler.java
```diff
@@ -0,0 +1,157 @@
...
+    setNodeCmdProto = setNodeCmd.getProto();
+    DatanodeDetails dni = context.getParent().getDatanodeDetails();
+    dni.setPersistedOpState(setNodeCmdProto.getNodeOperationalState());
+    dni.setPersistedOpStateExpiryEpochSec(
+        setNodeCmd.getStateExpiryEpochSeconds());
+    try {
+      persistDatanodeDetails(dni);
+    } catch (IOException ioe) {
+      LOG.error("Failed to persist the datanode state", ioe);
+      // TODO - this should probably be raised, but it will break the command
```
anuengineer: We can add this to the Dn reports if you want to propagate this back into SCM.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/DatanodeIdYaml.java
```diff
@@ -105,6 +112,8 @@ public static DatanodeDetails readDatanodeIdFile(File path)
     private String ipAddress;
     private String hostName;
     private String certSerialId;
+    private String persistedOpState;
+    private long persistedOpStateExpiryEpochSec = 0;
```
anuengineer: Since zero is the default value for the long, would it make sense to enforce time; that is for each set command we need an explicit time. In that case, we can define a constant called INFINITY and set it to something like -1; but in the source, it will be very explicit when someone reads code, and in the command handler we can enforce the requirement (or perhaps better on the SCM side). So commands like decommission will have an infinite value explicitly set by SCM and it would be more obvious and explicit.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/DatanodeIdYaml.java
sodonnel: If I recall (I wrote this code a month or so ago and forgot to create the PR) the reason the OpState is wrapped in a IF block is because I need take the string read from the YAML file and turn it into the enum. The enum didn't like me calling valueOf(null), so I only make that call if its non-null. All the other parameters in the builder are strings except for the two new ones (enum and long), so they never really had to handle the null case before.

You are correct, in that if there is no data, the "getPersistedOpStateExpiryEpochSec" will end up as zero. That is OK, as SCM treats zero as "no expiry" right now.

I've got similar logic at the end of that class too to handle the turning the enum into a string:

```
    String persistedOpString = null;
    if (datanodeDetails.getPersistedOpState() != null) {
      persistedOpString = datanodeDetails.getPersistedOpState().name();
    }
    return new DatanodeDetailsYaml(
        datanodeDetails.getUuid().toString(),
        datanodeDetails.getIpAddress(),
        datanodeDetails.getHostName(),
        datanodeDetails.getCertSerialId(),
        persistedOpString,
        datanodeDetails.getPersistedOpStateExpiryEpochSec(),
        portDetails);
  }
```

I think the shorter summary is that the enums needs special handling around the nulls while the long and Strings behave OK without wrapping them in IF blocks.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/DatanodeIdYaml.java
sodonnel: This might be a good idea. The current "infinity" value in SCM is zero, as it checks for greater than zero expiry right now, but it could be changed to a constant easily enough.

It would be good for the command to always get a time so its consistent. I need to go back and refresh my memory on the command etc here.

sodonnel:
@anuengineer Thanks for the review on this. I probably need to add some tests around these changes. I did not do that initially, as I wanted to be sure we were happy with the approach (storing the state in the datanode details, allowing it to be included in the heartbeat and the DN not making further use of the value). Now you have reviewed and are happy with this direction, I will get some tests added and check your comments in more detail.

elek:
/pending "I will get some tests added and check your comments in more detail."

sodonnel:
We have parked the decommission work at the moment. Anu approved this change, but I feel we need some unit tests around the changes.

As this is only going onto a branch which we will need to review in more detail later, I have raised HDDS-3303 to log the fact we need some tests and then I propose we commit this to the branch as is to get it out of the PR queue.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/helpers/DatanodeIdYaml.java
sodonnel: I think this is a good suggestion so I have logged it in HDDS-3304 and we can consider it when the decommission work is started up again.

PR #527, opened 2020-02-04
HDDS-2977. ozonesecure acceptance test fails due to unexpected error message

adoroszlai:

Fix expected error message in `Create volume with non-admin user` acceptance test.  Volume creation is rejected because `testuser2` is not an admin, not due to authentication failure.

https://issues.apache.org/jira/browse/HDDS-2977

## How was this patch tested?

Ran acceptance test (specific test locally, all tests in CI):

https://github.com/adoroszlai/hadoop-ozone/runs/425107849


adoroszlai:
Thanks @xiaoyuyao for reviewing and merging this.

PR #528, opened 2020-02-04
HDDS-2979. Implement ofs://: Fix getFileStatus for mkdir volume

smengcl:

Xiaoyu Yao discovered that when running ozone fs -mkdir -p ofs://om/vol1/ (only volume name is given, no bucket name or key path), the command would fail in getFileStatus (before reaching createDirectory()) in Hadoop common code:
```
bash-4.2$ ozone fs -mkdir -p ofs://om/vol1/
-mkdir: Bucket or Volume length is illegal, valid length is 3-63 characters

# Same w/o -p
bash-4.2$ ozone fs -mkdir ofs://om/vol1/
-mkdir: Bucket or Volume length is illegal, valid length is 3-63 characters
And we discovered with debugger attached that the root cause is that getFileStatus() is not behaving as expected.
```
Solution: Patch existing OFS code, throw proper exception in getBucket() code to make Hadoop common happy.

https://issues.apache.org/jira/browse/HDDS-2979

## How was this patch tested?

Manually ran `ozone fs -mkdir -p ofs://om/vol1/` in `docker-compose` cluster:

```
bash-4.2$ ozone fs -mkdir -p ofs://om/vol1/
2020-02-04 20:58:25,828 [main] INFO rpc.RpcClient: Creating Volume: vol1, with hadoop as owner.
```

Will add a new unit test for this case.

### DIFF
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
@@ -439,6 +439,12 @@ public class BasicRootedOzoneClientAdapterImpl
     incrementCounter(Statistic.OBJECTS_QUERY);
     OFSPath ofsPath = new OFSPath(path);
     String key = ofsPath.getKeyName();
+    // getFileStatus is called for root
+    if (ofsPath.getVolumeName().isEmpty() &&
+        ofsPath.getBucketName().isEmpty()) {
+      // Generate a FileStatusAdapter for root
+      return rootFileStatusAdapter();
+    }
     try {
       OzoneBucket bucket = getBucket(ofsPath, false);
       OzoneFileStatus status = bucket.getFileStatus(key);
@@ -655,6 +661,22 @@ public class BasicRootedOzoneClientAdapterImpl
         status.getOwner(),
         status.getGroup(),
         status.getPath()
+    );
+  }
+
+  private FileStatusAdapter rootFileStatusAdapter() {
+    return new FileStatusAdapter(
+        0L,
+        null,
+        true,
+        (short)0,
+        0L,
+        0L,
+        0L,
+        (short)0,
+        null,
+        null,
+        null
     );
   }
 }
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
@@ -81,8 +81,8 @@ public class BasicRootedOzoneClientAdapterImpl
       LoggerFactory.getLogger(BasicRootedOzoneClientAdapterImpl.class);
 
   private OzoneClient ozoneClient;
-  private ClientProtocol proxy;
   private ObjectStore objectStore;
+  private ClientProtocol proxy;
   private ReplicationType replicationType;
   private ReplicationFactor replicationFactor;
   private boolean securityEnabled;
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
@@ -199,11 +199,11 @@ public class BasicRootedOzoneClientAdapterImpl
       boolean createIfNotExist) throws IOException {
     Preconditions.checkNotNull(volumeStr);
     Preconditions.checkNotNull(bucketStr);
-
-    if (!createIfNotExist && bucketStr.isEmpty()) {
-      // Make Hadoop common happy by throwing FileNotFoundException in this case
-      throw new FileNotFoundException("getBucket: Invalid arguments: "
-          + "bucket is empty string while create flag is not set.");
+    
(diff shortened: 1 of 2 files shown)


smengcl:
Note the current commit still doesn't work when `-p` isn't given in param. Should be further fixed:

```
bash-4.2$ ozone fs -mkdir ofs://om/vol2/
mkdir: `ofs://om/': No such file or directory
```

I will rebase my local work branch in a minute.

smengcl:
I found I already have a test case for create volume only. But it couldn't catch it because the problem originates in Hadoop common.

```java
  /**
   * OFS: Tests mkdir on a volume that doesn't exist.
   */
  @Test
  public void testMkdirNonExistentVolume() throws Exception {
    String volumeNameLocal = getRandomNonExistVolumeName();
    Path newVolume = new Path("/" + volumeNameLocal);
    fs.mkdirs(newVolume);
```

smengcl:
I figured this out. We simply need to handle the case in `getFileStatus` where Hadoop is asking for root(`/`). After the next commit the behavior should become:

```
bash-4.2$ ozone fs -mkdir ofs://om/vol1/
2020-02-05 00:59:59,518 [main] INFO rpc.RpcClient: Creating Volume: vol1, with hadoop as owner.

bash-4.2$ ozone fs -mkdir ofs://om/vol1/
2020-02-05 01:00:34,016 [main] INFO rpc.RpcClient: Creating Volume: vol1, with hadoop as owner.
mkdir: Volume already exists

bash-4.2$ ozone fs -mkdir ofs://om/vol2/
2020-02-05 01:00:39,262 [main] INFO rpc.RpcClient: Creating Volume: vol2, with hadoop as owner.
```

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -648,4 +663,20 @@ private FileStatusAdapter toFileStatusAdapter(OzoneFileStatus status) {
         status.getPath()
     );
   }
+
+  private FileStatusAdapter rootFileStatusAdapter() {
+    return new FileStatusAdapter(
```
xiaoyuyao: should we make this static final?

smengcl:
> LGTM overall. Can we add a unit or acceptance test for the volume/bucket only creation case with ofs?

As I mentioned above the exact case (where Hadoop common is involved) is hard to be covered by unit test inside `ozonefs` package. But I'm adding another test to ensure `getListStatus` won't return `null` or throw for root.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: Good point. We should.

Made it static. As it's private we won't need final.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -648,4 +663,27 @@ private FileStatusAdapter toFileStatusAdapter(OzoneFileStatus status) {
...
+    // TODO: Revisit the return value.
+    return new FileStatusAdapter(
+        0L,
+        null,
+        true,
+        (short)0,
+        0L,
+        0L,
+        0L,
+        (short)0,
```
smengcl: Try 755

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -648,4 +663,27 @@ private FileStatusAdapter toFileStatusAdapter(OzoneFileStatus status) {
...
+        null,
```
smengcl: use ozone admin

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -648,4 +663,27 @@ private FileStatusAdapter toFileStatusAdapter(OzoneFileStatus status) {
...
+        null,
```
smengcl: ozone admin group?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: But for now we are leaving this field as null since we didn't have an API to retrieve such info from OM yet.

smengcl:
Not sure why Github actions failed the acceptance. I ran one locally and it passed:
```bash
ozone-basic.Ozone-Shell :: Test ozone shell CLI usage
RpcClient with port                                                   | PASS |
RpcClient volume acls                                                 | PASS |
```

while the latest acceptance shows failure for the same `RpcClient with port` test:
Screen Shot 2020-02-06 at 10 08 52 AM

My command to run the test manually:
```bash
mvn clean install -Pdist -DskipTests -e -Dmaven.javadoc.skip=true -DskipShade -DskipRecon
./test.sh
```

smengcl:
I just rebased this PR on to the feature branch. No conflicts what so ever.

PR #529, opened 2020-02-05
HDDS-2982. Recon server start failed with CNF exception in a cluster …

avijayanhwx:
…with Auto TLS enabled.

Recon server crashes on a cluster with Auto TLS enabled due to CNF error while starting up its HTTP server. Recon's Httpserver is inherited from BaseHttpServer that is used by all the other components in Ozone. The underlying problem is being fixed in HDDS-2041. Since the other components like the SCM, OM start successfully even if their HTTP Server is not up, this patch relaxes the start for Recon as well. 

https://issues.apache.org/jira/browse/HDDS-2982

## How was this patch tested?
Manually tested.

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
@@ -82,7 +82,13 @@ public class ReconServer extends GenericCli {
       LOG.info("Creating Recon Schema.");
       reconSchemaManager.createReconSchema();
 
-      httpServer = injector.getInstance(ReconHttpServer.class);
+      // TODO Remove try-catch block after HDDS-2041 is fixed.
+      // Allowing Recon server start even if HTTP server start failed
+      try {
+        httpServer = injector.getInstance(ReconHttpServer.class);
+      } catch (Exception ex) {
+        LOG.error("Error trying to start HTTP Server. ", ex);
+      }
       this.ozoneManagerServiceProvider =
           injector.getInstance(OzoneManagerServiceProvider.class);
       this.reconStorageContainerManager =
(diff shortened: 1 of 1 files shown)


avijayanhwx:
cc @swagle

mukul1987:
@avijayanhwx there are some integration test failures. they do not seem related however can you please push an empty patch to re-trigger the tests.

adoroszlai:
> there are some integration test failures.

Acceptance test failures are tracked in HDDS-2977 and HDDS-2983.  Please do not re-trigger until those are merged.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
```diff
@@ -82,7 +82,13 @@ protected void configureServlets() {
       LOG.info("Creating Recon Schema.");
       reconSchemaManager.createReconSchema();
 
-      httpServer = injector.getInstance(ReconHttpServer.class);
+      // TODO Remove try-catch block after HDDS-2041 is fixed.
+      // Allowing Recon server start even if HTTP server start failed
+      try {
```
bharatviswa504: One question if HttpServer start failed will any functionality of recon works, like getContainers etc.,

But whereas in OM SCM, core functionality still works even if HttpServer does not start. Not sure if I am missing something here.

avijayanhwx: @bharatviswa504 Yes that is a valid question. On further investigation, I discovered that this may be due to jetty upgrade in hadoop in the distribution. This affects all Ozone components. Also, in the case of SCM, even though it withstands a HttpServer crash, since the initialization is stopped midway due to exception, none of the event handlers are registered thereby making the component useless.

I plan to hold on to this PR, wont commit it. If jetty version is indeed the problem, I will close this PR without merging the changes.

PR #536, opened 2020-02-08
HDDS-2988. Use getPropertiesByPrefix instead of regex in matching ratis client and server properties.

bharatviswa504:

1. Remove getValByRegex usage and regex usage.
2. Use getByPrefix.
3. Use prefix for Ratis Client, GRPC and Serve Conf.

https://issues.apache.org/jira/browse/HDDS-2988

## How was this patch tested?

UT's are modified to test the behavior.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
@@ -30,6 +30,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hdds.conf.ConfigurationException;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.pipeline.Pipeline;
@@ -42,7 +43,9 @@ import org.apache.hadoop.ozone.OzoneConfigKeys;
 
 import org.apache.ratis.RaftConfigKeys;
 import org.apache.ratis.client.RaftClient;
+import org.apache.ratis.client.RaftClientConfigKeys;
 import org.apache.ratis.conf.RaftProperties;
+import org.apache.ratis.grpc.GrpcConfigKeys;
 import org.apache.ratis.grpc.GrpcFactory;
 import org.apache.ratis.grpc.GrpcTlsConfig;
 import org.apache.ratis.proto.RaftProtos;
@@ -54,12 +57,11 @@ import org.apache.ratis.retry.RetryPolicies;
 import org.apache.ratis.retry.RetryPolicy;
 import org.apache.ratis.rpc.RpcType;
 import org.apache.ratis.rpc.SupportedRpcType;
+import org.apache.ratis.server.RaftServerConfigKeys;
 import org.apache.ratis.thirdparty.com.google.protobuf.ByteString;
 import org.apache.ratis.util.TimeDuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import static org.apache.hadoop.ozone.conf.DatanodeRatisServerConfig.DATANODE_RATIS_SERVER_CONFIG_PREFIX;
 
 /**
  * Ratis helper methods.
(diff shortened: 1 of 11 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
```diff
@@ -54,29 +57,34 @@
...
  */
 public interface RatisHelper {
   Logger LOG = LoggerFactory.getLogger(RatisHelper.class);
 
-  // Ratis Client and Grpc header regex filters.
-  String RATIS_CLIENT_HEADER_REGEX = "raft\\.client\\.([a-z\\.]+)";
-  String RATIS_GRPC_CLIENT_HEADER_REGEX = "raft\\.grpc\\.(?!server|tls)" +
-      "([a-z\\.]+)";
+  // Prefix for Ratis Server GRPC and Ratis client conf.
+  String HDDS_DATANODE_RATIS_PREFIX_KEY = "datanode.ratis.";
```
elek: Nit: `HDDS_DATANODE_RATIS_PREFIX_KEY` can be reused.

```
  String HDDS_DATANODE_RATIS_PREFIX_KEY = "datanode.ratis.";
  String HDDS_DATANODE_RATIS_SERVER_PREFIX_KEY =
      HDDS_DATANODE_RATIS_PREFIX_KEY + ".server";
  String HDDS_DATANODE_RATIS_CLIENT_PREFIX_KEY =
      HDDS_DATANODE_RATIS_PREFIX_KEY + ".client";
  String HDDS_DATANODE_RATIS_GRPC_PREFIX_KEY =
      HDDS_DATANODE_RATIS_PREFIX_KEY + ".grpc";
```

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
```diff
@@ -54,29 +57,34 @@
...
+  String HDDS_DATANODE_RATIS_CLIENT_PREFIX_KEY = "datanode.ratis.client";
+  String HDDS_DATANODE_RATIS_GRPC_PREFIX_KEY = "datanode.ratis.grpc";
+
 
-  // Ratis Server header regex filter.
-  String RATIS_SERVER_HEADER_REGEX = "datanode\\.ratis\\.raft\\.server\\" +
-      ".([a-z\\.]+)";
-  String RATIS_SERVER_GRPC_HEADER_REGEX = "datanode\\.ratis\\.raft\\.grpc\\" +
-      ".([a-z\\.]+)";
+  String RATIS_SERVER_PREFIX_KEY =
```
elek: Do we really need the `server` and `client` part here? Why don't we use `datanode.ratis.raft.client.*` and `datanode.ratis.raft.server.*` instead of `datanode.ratis.server.raft.server`? Is there any use case where we need to configure the client configs on the server side? (`datanode.ratis.server.raft.client.*`)

BTW, this is just a matter of taste but I would prefer to avoid substring. This seems to be more readable for me:

```
  String HDDS_DATANODE_RATIS_SERVER_TAG = "server";
  
String HDDS_DATANODE_RATIS_PREFIX_KEY = "datanode.ratis.";

  String HDDS_DATANODE_RATIS_SERVER_PREFIX_KEY =
      HDDS_DATANODE_RATIS_PREFIX_KEY + "." HDDS_DATANODE_RATIS_SERVER_TAG;


String RATIS_SERVER_PREFIX_KEY =  RATIS_SERVER_PREFIX_KEY +"." + RaftServerConfigKeys.PREFIX;
```

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
```diff
@@ -218,58 +223,74 @@ static RaftClient newRaftClient(RpcType rpcType, RaftPeerId leader,
...
+      if (!key.startsWith(RATIS_CLIENT_PREFIX_KEY)) {
+        raftProperties.set(removeDatanodePrefix(key), val);
+      }
+    });
   }
 
   static String removeDatanodePrefix(String key) {
-    return key.replaceFirst(DATANODE_RATIS_SERVER_CONFIG_PREFIX, "");
+
+    if (key.startsWith("server.")) {
```
elek: Is there any specific reason to limit this to `server`, `client` and `grpc`? (BTW, can we use constants to here? It's not clear if they are ratis segments or ozone segments...)

bharatviswa504:
> Thanks the patch @bharatviswa504
> 
> Overall I like the idea to use string manipulation instead of regexp but it's a little hard to read the structure of the configuration. Maybe because the duplication of the "server/grpc" words in `datanode.ratis.grpc.raft.grpc.flow.control` (As I asked in an other comment why do we need the first grpc?).

The reason for server, client and grpc prefix is to group the configurations by the group. So, that it will be easy to add these configurations in different classes(like have a class for Ratis Server, One for Ratis Client ..) instead of just having all the configuration in a single class.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
bharatviswa504: Because first, we call getByPrefix is called with "datanode.ratis." and which it returns a map by removing 'datanode.ratis." from the key, so here we will be left with 3 prefixes server/client and grpc (As ratis currently has 3 groups one for Ratis Server, Ratis Client and other for Ratis Grpc). This is the reason for the check and then based on the prefix removing the prefix. Yes, I will use constants here.

elek:
> The reason for server, client and grpc prefix is to group the configurations by the group

Thanks the answer. I understand that grouping is useful. I don't understand why do we need double grouping. 

> Do we really need the server and client part here? Why don't we use `datanode.ratis.raft.client.*` and `datanode.ratis.raft.server.*` instead of `datanode.ratis.server.raft.server`? Is there any use case where we need to configure the client configs on the server side? (`datanode.ratis.server.raft.client.*` for example?)

If we already have a grouping on the ratis side why do we introduce an other one on the ozone side?

bharatviswa504:
> > The reason for server, client and grpc prefix is to group the configurations by the group
> 
> Thanks the answer. I understand that grouping is useful. I don't understand why do we need double grouping.
> 
> > Do we really need the server and client part here? Why don't we use `datanode.ratis.raft.client.*` and `datanode.ratis.raft.server.*` instead of `datanode.ratis.server.raft.server`? Is there any use case where we need to configure the client configs on the server side? (`datanode.ratis.server.raft.client.*` for example?)
> 
> If we already have a grouping on the ratis side why do we introduce an other one on the ozone side?

The reasons for a prefix "datanode" is in future if OM also uses this similar approach, we need a way to distinguish which config is for which component. So, that is the reason for prefix. So, created 3 config classes, one for ratis server with datanode.ratis.server, similarly for others. I understand the ratis.server is duplicated again with this approach. If any other thoughts how to handle this?

elek:
>  we can use "datanode.raft.server" for RatisServer config and for client "datanode.raft.client" and for grpc "datanode.raft.grpc". In this way we can use directly ratis grouping. Is this is what you are suggesting here?

Exactly. Sorry if was not clear. It's a very small suggestion, and not a big deal. 

My impression was that it simplifies the whole code as you don't need to do all the magic with the middle `[server|grpc]` tag in the `datanode.ratis.server.raft.server`. You can use the configuration based on the ratis groups.

bharatviswa504:
> > we can use "datanode.raft.server" for RatisServer config and for client "datanode.raft.client" and for grpc "datanode.raft.grpc". In this way we can use directly ratis grouping. Is this is what you are suggesting here?
> 
> Exactly. Sorry if was not clear. It's a very small suggestion, and not a big deal.
> 
> My impression was that it simplifies the whole code as you don't need to do all the magic with the middle `[server|grpc]` tag in the `datanode.ratis.server.raft.server`. You can use the configuration based on the ratis groups.

Thank You @elek for review. Addressed review comments.

(With this approach, not able to use direct Ratis variables. (But I think that is okay)

PR #537, opened 2020-02-08
HDDS-2990. Set the default value of grpc flow control window for ratis client and ratis server

bharatviswa504:

The grpc flow control window by default is set to 1Mb by default. During performance tests, it was observed that the flow control window has to be greater than the chunk size for optimum performance.

This PR is dependent on #536. For the ease of review refer only 2nd commit, which are the changes for this PR.

https://issues.apache.org/jira/browse/HDDS-2990

## How was this patch tested?

Deployed on physical cluster and ran multiple teragen tests.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/conf/DatanodeRatisGrpcConfig.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/conf/DatanodeRatisGrpcConfig.java
@@ -18,6 +18,8 @@
 
 package org.apache.hadoop.hdds.conf;
 
+import org.apache.ratis.grpc.GrpcConfigKeys;
+
 import static org.apache.hadoop.hdds.conf.ConfigTag.CLIENT;
 import static org.apache.hadoop.hdds.conf.ConfigTag.OZONE;
 import static org.apache.hadoop.hdds.conf.ConfigTag.PERFORMANCE;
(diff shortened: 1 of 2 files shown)


hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/conf/RatisClientConfig.java
```diff
@@ -26,9 +29,9 @@
  * RaftClient creation.
  *
  */
-@ConfigGroup(prefix = "raft.client")
+@ConfigGroup(prefix = RatisHelper.HDDS_DATANODE_RATIS_CLIENT_PREFIX_KEY)
 public class RatisClientConfig {
-  @Config(key = "async.outstanding-requests.max",
+  @Config(key = RaftClientConfigKeys.Rpc.REQUEST_TIMEOUT_KEY,
```
bshashikant: RaftClientConfigKeys.Rpc.REQUEST_TIMEOUT_KEY,  the config key is wrong. Can you please fix it as it denotes the max async requests count on client?

bshashikant:

The ratis client config and ratis server config keys defined are specific to ozone client / datanode or will be used by OM as well?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/conf/RatisClientConfig.java
bharatviswa504: This is taken care of in HDDS-2988.

bharatviswa504:
> Thanks @bharatviswa504 for working on this.
> 
> The ratis client config and ratis server config keys defined are specific to ozone client / datanode or will be used by OM as well?

This config class if for Datanode. So, that is the reason for prefix "hdds.ratis"

PR #539, opened 2020-02-10
HDDS-2992. Handle existing volume/bucket in contract tests

adoroszlai:

Make at most 5 attempts to create randomly named volume/bucket in integration tests if creation fails due to conflict (volume/bucket already exists).

https://issues.apache.org/jira/browse/HDDS-2992

## How was this patch tested?

Faked random "volume already exists" and "bucket already exists" events, verified that creation is attempted at most 5 times.

https://github.com/adoroszlai/hadoop-ozone/runs/436426070


elek:
```
if (i == attempts) {
   LOG.info("$$$$$ JACKPOT $$$$$");
}
```

;-)

adoroszlai:

PR #542, opened 2020-02-11
HDDS-2974. Create Freon test to test isolated Ratis Follower

elek:

We have a specific Freon (=load generator) test to test one specific working Ratis pipeline (3 datanodes): `DatanodeChunkGenerator`

It sends Ozone GRPC requests to the datanode with `WriteChunk` requests.

But we have no easy way to test just one the running Ratis component.

In this Jira I propose to create test for isolated ratis components. There two possible components: follower and leader. In this patch I introduce a follower load tester

FollowerAppendLogEntryGenerator

When Datanode is started without registration it contains an empty Ratis server instance.

Freon test can initialize a new group and request vote with behaving like a real LEADER. It can send real appendLogEntry requests (the Ratis HB messages) with GRPC

With this approach we can force the datanode to be a FOLLOWER and analyze and profile the behavior.

https://issues.apache.org/jira/browse/HDDS-2974

## How was this patch tested?

Tested in Kubernetes with [multiple, separated load tests](https://hackmd.io/fzbXUQ-xTTmnGDL7fufZrA), but it also can be tested locally:

**1.** Modify the configuration of the simple ozone cluster definition at `compose/ozone`. Add the following line to the `docker-config`:

```
OZONE_DATANODE_STANDALONE_TEST=follower
OZONE-SITE.XML_dfs.ratis.leader.election.minimum.timeout.duration=2d
```

**2.** Start one single datanode:

```
docker-compose down
docker-compose up -d datanode
```
**3.** Find the UUID generated for the datanode:

```
docker-compose logs datanode | grep "GrpcService started"
datanode_1  | 2020-02-11 10:32:53,549 [main] INFO server.GrpcService: 5ddf8cba-618f-4af8-ba3e-8c2c48b01ae7: GrpcService started, listening on 0.0.0.0/0.0.0.0:9858
```

In this case the UUID is: `5ddf8cba-618f-4af8-ba3e-8c2c48b01ae7`

**4.** Start freon test with the UUID

```
docker-compose exec datanode ozone freon falg --thread=1 --batching=1 --size=1024 --number-of-tests=1000  -r 5ddf8cba-618f-4af8-ba3e-8c2c48b01ae7
```


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/FollowerAppendLogEntryGenerator.java
```diff
@@ -0,0 +1,470 @@
...
+  /**
+   * Seend a new HB and record the call id to handle response.
+   */
+  private void sendAppendLogEntryRequest(long sequence) {
+    timer.time(() -> {
+      AppendEntriesReplyProto replyProto = null;
+      try {
+        long callId = callIdRandom.nextLong();
+        inFlightMessages.put(callId);
+        sender.onNext(createAppendLogEntry(sequence, callId));
```
adoroszlai: Result of `createAppendLogEntry` should be assigned to `replyProto` if it is to be used in the log below.

elek: Earlier I used sync call and I stored the *reply*. Result of `createAppendLogEntry` is the request and we have no *reply* here as the same observer (`sender`) is used for all the requests.

I removed the old, unused code (logging just the exception)

elek:
Thanks the review @adoroszlai I am merging it now (the rebased branch is also passed...)

PR #543, opened 2020-02-11
HDDS-2997. Support github comment based commands

elek:

Before we started to use github actions we had the opportunity to use some "commands" in github comments. For example when a `/label xxx` comment has been added to a PR, a bot added the label (by default just the committers can use labels, but with this approach it was possible for everyone).

Since the move to use github actions I got multiple question about re-triggering the test. Even it it's possible to do with pushing an empty commit (only by the owner or committer) I think it would be better to restore the support of comment commands.

This patch follows a very simple approach. The available commands are store in a separated subdirectory as shell scripts and they are called by a lightweight wrapper.

https://issues.apache.org/jira/browse/HDDS-2997

## How was this patch tested?

As it works only if merged to the master I tested it in my own fork:

Feel free to use this example PR to test it:

https://github.com/elek/hadoop-ozone/pull/9

### DIFF
--- a/.github/comment-commands/pending.sh
+++ b/.github/comment-commands/pending.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#doc: Add a REQUESTED_CHANGE type review to mark issue non-mergeable: `/pending <reason>`
+# shellcheck disable=SC2124
+MESSAGE="Marking this issue as un-mergeable as requested.
+
+Please use \`/ready\` comment when it's resolved.
+
+> $@"
+
+URL="$(jq -r '.issue.pull_request.url' "$GITHUB_EVENT_PATH")/reviews"
+set +x #GITHUB_TOKEN
+curl -s -o /dev/null \
+  --data "$(jq --arg body "$MESSAGE" -n '{event: "REQUEST_CHANGES", body: $body}')" \
+  --header "authorization: Bearer $GITHUB_TOKEN" \
+  --header 'content-type: application/json' \
+  "$URL"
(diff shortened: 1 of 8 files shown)


.github/comment-commands/retest.sh
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#doc: add new empty commit to trigger new CI build
+
+PR_URL=$(jq -r '.issue.pull_request.url' "$GITHUB_EVENT_PATH")
+read -r REPO_URL BRANCH <<<"$(curl "$PR_URL" | jq -r '.head.repo.ssh_url + " " + .head.ref')"
```
adoroszlai: I tried `/retest` in the PR in your repo, but it [failed](https://github.com/elek/hadoop-ozone/pull/9#issuecomment-584615406).  Just guessing: maybe `-s` parameter for `curl` is missing and the extra output trips up `jq`?

.github/comment-commands/pending.sh
```diff
@@ -0,0 +1,31 @@
...
+#doc: Add a REQUESTED_CHANGE type review to mark issue non-mergeable: `/pending <reason>`
+# shellcheck disable=SC2124
+MESSAGE="Marking this issue as un-mergeable as requested.
+
+Please use \`/ready\` comment when it's resolved.
+
+> $@"
+
+URL="$(jq -r '.issue.pull_request.url' "$GITHUB_EVENT_PATH")/reviews"
+set +x #GITHUB_TOKEN
```
adoroszlai: Is `set +x` needed for safety in the other scripts, or is this one special?

elek: I am not sure if github expose the token to the console or it's masked. If not masked we should either use `set +x` or never use `set -x`. Both works. Do you prefer to put it to everywhere or remove it from everwhere?

adoroszlai: I wouldn't have thought about it, but since you added it here, I guess I'd prefer adding it everywhere.

.github/comment-commands/retest.sh
elek: Seems to be a permission problem. For some reason github couldn't push. I can try to use https remote and the token as the password...

elek: OK, it's fixed now. Tested on the same PR

.github/comment-commands/pending.sh
elek: Ok. I added `set +x` to everywhere where GITHUB_TOKEN is used.

.github/comment-commands/retest.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+!/usr/bin/env bash
```
adoroszlai: Is removal of `#` intentional?

adoroszlai:
```
2020-02-13T09:37:16.1836554Z /home/runner/work/hadoop-ozone/hadoop-ozone/.github/comment-commands/retest.sh: line 1: !/usr/bin/env: No such file or directory
```

elek: Not intentional. Just disappeared in the fog. Thanks to catch it.

elek:
Thanks the review and the test @adoroszlai I am merging it to the master right now...

PR #545, opened 2020-02-11
HDDS-3000. Update guava version to 28.2-jre.

jojochuang:
Update Ozone's guava dependency to the latest.

https://issues.apache.org/jira/browse/HDDS-3000

## How was this patch tested?


jojochuang:
Looks like the newer Guava versions updated the annotation for the API 
`void onSuccess(@Nullable V result);`

(it didn't have the `@Nullable` annotation in Gauva 11.) Updated the PR to add the nullity check.

See also: https://github.com/spotbugs/spotbugs/issues/633

PR #551, opened 2020-02-13
HDDS-2717. Handle chunk increments in datanode

adoroszlai:

Introduce a new chunk layout to write one file per block, instead of one per chunk.

Changes since the previous PR:

 * use block ID for server-side filename instead of client-supplied chunk name
 * fixed retry path
 * fixed integration tests
 * implemented file close on last putBlock (uses new optional flag in putBlock request)
 * open block files are kept in a Guava cache with 10-minute expiry after last access

https://issues.apache.org/jira/browse/HDDS-2717

## How was this patch tested?

Tested with buffer sizes in the range 64KB-16MB with key sizes 1MB-512MB, eg.

```
ozone freon ockg -p warmup -n 1 -t 1
ozone freon ockg -p test -t 8 -n 64 -s 67108864 
ozone freon ockv -p test -t 8 -n 64
```

Found [RATIS-815](https://issues.apache.org/jira/browse/RATIS-815) with small buffer size and large key size combinations.

Tested upgrade locally:

1. created some keys running `master` via `freon ockg`
2. restarted datanode with new code
3. created more keys: written in open containers using per-chunk layout
4. closed existing containers
5. create more keys: written in new containers using per-block layout
6. verified both old and new keys via `freon ockv`


hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -274,13 +271,16 @@ public void writeOnRetry(long len) throws IOException {
...
     Preconditions.checkArgument(len <= streamBufferMaxSize);
+    int count = 0;
     while (len > 0) {
-      long writeLen;
-      writeLen = Math.min(chunkSize, len);
-      if (writeLen == chunkSize) {
-        writeChunk(bufferPool.getBuffer(count));
+      ChunkBuffer buffer = bufferPool.getBuffer(count);
+      long writeLen = Math.min(buffer.position(), len);
+      if (!buffer.hasRemaining()) {
```
bshashikant: Is it really required to initiate writeChunk only at the buffer boundary?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -360,20 +360,33 @@ private void watchForCommit(boolean bufferFull) throws IOException {
...
         ContainerCommandResponseProto> flushFuture;
     try {
+      BlockData blockData = containerBlockData.build();
+      if (close) {
+        blockData = BlockData.newBuilder(blockData)
+            .addMetadata(KeyValue.newBuilder()
+                .setKey(OzoneConsts.LAST_PUT_FOR_BLOCK)
+                .setValue(Boolean.TRUE.toString()))
+            .build();
+      }
```
bshashikant: I think it will bring in compatibility changes with the older client with no explicit close option set and latest server jar with new containers which will expect an explicit close option set? Are we ok with this?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -360,20 +360,33 @@ private void watchForCommit(boolean bufferFull) throws IOException {
   }
 
   private CompletableFuture<ContainerProtos.
```
bshashikant: What is the use of force flag here? can you update the comments here as well?

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
```diff
@@ -123,6 +123,12 @@
    * */
   public static final String OZONE_ADMINISTRATORS_WILDCARD = "*";
 
+  public static final String OZONE_CLIENT_STREAM_BUFFER_SIZE =
```
bshashikant: Any specific reason to define a new config rather than using the chunk config itself?

bshashikant:
Thanks @adoroszlai . I am started going through the change in detail and not completed yet. Will continue looking at this.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
adoroszlai: The `force` flag is needed to handle the case when the stream is closed right after a putBlock, without further new data.  Since we don't know in advance which putBlock will be the last, we need to "force" another putBlock in this case.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
adoroszlai: I think so.  By sending each buffer only once, we can simply send the complete buffer.  If we sent it incrementally, we would need to keep track of the last write end position separately.

https://github.com/apache/hadoop-ozone/blob/9eb8e0ef28db64e573feec5bb652228e4add522d/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java#L452

This is not new behavior, previously it checked `writeLen == chunkSize`, which resulted in the same:

https://github.com/apache/hadoop-ozone/blob/9eb8e0ef28db64e573feec5bb652228e4add522d/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java#L282-L283

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
adoroszlai: Ideally the file-per-chunk layout could be improved to use the existing chunk size config, eg. even if client sends 64KB chunks, server writes 4MB chunk files.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
adoroszlai: Good point, but that's only part of the problem.  Old client is anyway incompatible with new layout because it always sends 0 offset.  If backward compatibility is required, I think we would need to force old layout for old client.  For this, even block allocation needs to account for client version.

arp7: I don't think we need to preserve protocol compatibility with old clients right now since we don't have rolling upgrades. As long as the old chunks are still readable after upgrade on DN we should be fine.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
bshashikant: > I think so. By sending each buffer only once, we can simply send the complete buffer. If we sent it incrementally, we would need to keep track of the last write end position separately.
> 
> https://github.com/apache/hadoop-ozone/blob/9eb8e0ef28db64e573feec5bb652228e4add522d/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java#L452
> 
> This is not new behavior, previously it checked `writeLen == chunkSize`, which resulted in the same:
> 
> https://github.com/apache/hadoop-ozone/blob/9eb8e0ef28db64e573feec5bb652228e4add522d/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java#L282-L283

The previous approach was to ensure we only call writeChunk at the chunk boundary. With the current code, its driven by the buffer size which may or may not be equal to the chunk size chosen. 

Are we ensuring buffer size will always be equal to chunk size?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
bshashikant: > The `force` flag is needed to handle the case when the stream is closed right after a putBlock, without further new data. Since we don't know in advance which putBlock will be the last, we need to "force" another putBlock in this case.

If my understanding , if a stream gets closed, as a part of close itself it should send the last putBlock. Also, we always know which putBlock will be last, with this putBlock as cumulative length of data written will be higher and same putBlock future is tracked in futureMap in commitWatcher.

adoroszlai: > if a stream gets closed, as a part of close itself it should send the last putBlock

That's happening here.

> we always know which putBlock will be last

This is a stream, caller can write as much data as wanted (within block size limit).  How do we know which will be last before `close()` is called?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
adoroszlai: > Are we ensuring buffer size will always be equal to chunk size?

No, the goal was to make these independent.  However, we can get rid of the client-side config for now, and introduce it if/when server-side aggregation for per-chunk layout is implemented.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
bshashikant: > > if a stream gets closed, as a part of close itself it should send the last putBlock
> 
> That's happening here.
> 
> > we always know which putBlock will be last
> 
> This is a stream, caller can write as much data as wanted (within block size limit). How do we know which will be last before `close()` is called?

What I meant here was a part of close, whatever putBlock request sent will always be the last. But i get why force flag is being used here. Its the case where, for example, let's we did a flush first and then did a close without writing anything in between. In such case, even though no data has been written in between, a putBlock needs to be enforced. Am i correct?

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
bshashikant: > Ideally the file-per-chunk layout could be improved to use the existing chunk size config, eg. even if client sends 64KB chunks, server writes 4MB chunk files.

I like the idea of a new config , but with different config of chunk size and buffer size can lead to a case where ratis client itself can not send any request to server if buffer size exceeds chunk size. We can enforce client buffer size <= chunk size in the code itself.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
adoroszlai: > Am i correct?

Yes.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
adoroszlai: > We can enforce client buffer size <= chunk size in the code itself.

I'll update the checks in `BlockOutputStreamEntryPool` with this.  Does this need to consider actual `ozone.scm.chunk.size` config or the max. chunk size (32MB) allowed, which is used for setting max. GRPC message size?

bshashikant: I think its ok to restrict the buffer size check to chunk size itself. Currently it is set to MAX_CHUNK_SIZE = 32 MB, but it can also be changed in DatanodeGrpcConfig.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/ImmutableVolumeSet.java
```diff
@@ -0,0 +1,59 @@
...
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
```
bshashikant: Any other case or situation other than the test wich will use ImmutableVolumeSet?

bshashikant:
@adoroszlai , I think with the new filePerBlockStragey we can sequentially read multiple chunks for block limited by protobuf max limit(128 MB) which should be faster than the previous approach in a single rpc based on how much of length we want to read.

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -386,13 +386,19 @@
...
+  </property>
   <property>
     <name>ozone.client.stream.buffer.flush.size</name>
     <value>16MB</value>
     <tag>OZONE, CLIENT</tag>
     <description>Size which determines at what buffer position , a partial
-      flush will be initiated during write. It should be ideally a multiple
-      of chunkSize.
+      flush will be initiated during write. It should be a multiple
+      of buffer size.
```
lokeshj1703: Can we mention the exact config here instead of buffer size?

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
```diff
@@ -369,34 +352,27 @@ public void testContainerThrottle() throws Exception {
...
-        try {
-          timesToProcess.incrementAndGet();
-          deleteAndWait(service, timesToProcess.get());
-          if (getNumberOfChunksInContainers(containerSet) == 0) {
-            return true;
-          }
-        } catch (Exception ignored) {}
-        return false;
-      }, 100, 100000);
-      Assert.assertEquals(0, getNumberOfChunksInContainers(containerSet));
```
lokeshj1703: For tests in this class, I think it would be a good idea to still make an assertion on the number of chunks in the containers at the end of call.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -81,18 +82,17 @@
   private final BlockData.Builder containerBlockData;
   private XceiverClientManager xceiverClientManager;
   private XceiverClientSpi xceiverClient;
-  private final ContainerProtos.ChecksumType checksumType;
   private final int bytesPerChecksum;
   private int chunkIndex;
-  private int chunkSize;
+  private final AtomicLong chunkOffset = new AtomicLong();
```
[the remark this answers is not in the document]
adoroszlai: That was my initial attempt, but it turned out that for retries a new `BlockOutputStream` with 0 offset is created, so I needed to change the retry logic: f8429f8b946318e73bff6497ab619a0e53d78562

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -360,20 +359,26 @@ private void watchForCommit(boolean bufferFull) throws IOException {
   }
 
   private CompletableFuture<ContainerProtos.
-      ContainerCommandResponseProto> executePutBlock()
-      throws IOException {
+      ContainerCommandResponseProto> executePutBlock(boolean close,
+      boolean force) throws IOException {
```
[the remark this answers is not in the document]
adoroszlai: I guess I'll need to document those. ;)

`close` flag is simply for distinguishing between regular `flush()` and `close()`.

Description of `force` flag from @bshashikant:

> Its the case where, for example, let's we did a flush first and then did a close without writing anything in between. In such case, even though no data has been written in between, a putBlock needs to be enforced.

hadoop-hdds/common/src/main/resources/ozone-default.xml
adoroszlai: Sure, will do.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
adoroszlai: The problem with this test (and the service it tests) was that it used internal information about the layout.  Will check whether such information is still available without depending on filenames etc.

adoroszlai:
Thanks @bshashikant and @lokeshj1703 for the reviews.

> For some unit tests it would be great if we could make them Parameterized so that they run using both the chunk layouts. I think it's already in your plan?

Yes, [HDDS-2943](https://issues.apache.org/jira/browse/HDDS-2943).

bshashikant:
The changes look good to me. @adoroszlai , i think we can make the open file expire interval configurable but it can be done as a separate patch.

elek:
I tested with this path with teragen (100GB) and compared the numbers with the current master.

 1. I see high variance in the test (same test can provide +/- 10% performance at a different time)
 2. During my first test (used an old version of master / old version of this branch) the patch was slower: https://hackmd.io/HbYNWTE2SXyOgZQ6CeXaXg
 3. Yesterday I tested the final version and couldn't reproduce any slowness: https://hackmd.io/P3Hkf09fTLS78Ho4W82YbQ
   * This patch always provided better performance than the master
 4. Using load tests (freon) instead of teragen I have seen significant better performance. The current implementation uses a lot of unnecessary file.exists check which makes impossible to get better performance than 1.2 k / sec chunk write. This patch (even with the old layout) can easily provide > 4k without any adjustment.

elek:
Thanks the review @bshashikant and the patch @adoroszlai 

I will merge it if no objections (@lokeshj1703? @arp7)

elek:
> Thanks the review @bshashikant and the patch @adoroszlai

And thanks the review @lokeshj17 and @arp7. 

I am pushing it now.

adoroszlai:
Thanks @bshashikant, @lokeshj1703, @elek and @arp7 for the reviews.

PR #553, opened 2020-02-14
HDDS-3021. Ozone S3 CLI path command not working on HA cluster.

bharatviswa504:

ozone s3 path <<bucketname>>

ozone s3 path are not working on OM HA cluster

Fix these commands to work on OM HA cluster.

https://issues.apache.org/jira/browse/HDDS-3021

## How was this patch tested?

Added UT and also tested on the docker-compose om-ha-s3 cluster.
```
bash-4.2$ ozone s3 path b12345 --om-service-id=id1
Volume name for S3Bucket is : s34124bc0a9335c27f086f24ba207a4912
Ozone FileSystem Uri is : o3fs://b12345.s34124bc0a9335c27f086f24ba207a4912
```

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/ozShell/TestOzoneShellHA.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/ozShell/TestOzoneShellHA.java
@@ -27,6 +27,9 @@ import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.om.OMConfigKeys;
 import org.apache.hadoop.ozone.om.OzoneManager;
 import org.apache.hadoop.ozone.web.ozShell.OzoneShell;
+import org.apache.hadoop.ozone.web.ozShell.Shell;
+import org.apache.hadoop.ozone.web.ozShell.s3.S3Shell;
+import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.test.GenericTestUtils;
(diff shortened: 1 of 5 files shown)


bharatviswa504:
If conf has a service ID, but it is of the remote cluster. (I think this can happen when conf has remote HA cluster details and the local non-HA cluster) . So, in HA for all commands, we have mandated to pass the serviceID, instead of best effort.

And also one more scenario can be in the future, when we have federated OM, with multiple serviceIDs like HDFS, even in that case these commands can still work, as passing explicitly serviceID is mandatory.

adoroszlai:
Thanks @bharatviswa504 for the explanation.

> So, in HA for all commands, we have mandated to pass the serviceID, instead of best effort.

But this also means the same command does not work in both non-HA and HA case.

My current concern is that smoke tests need to be aware of the environment they are run in.  Eg. I wanted to add a test to verify behavior of `ozone s3 getsecret` command, but it needs to distinguish between `ozone` and `ozone-om-ha-s3` environments (as well as `ozonesecure`, but we already have an env. variable for that).

elek:
> If conf has a service ID, but it is of the remote cluster. (I think this can happen when conf has remote HA cluster details and the local non-HA cluster) . So, in HA for all commands, we have mandated to pass the serviceID, instead of best effort.

Thanks to explain it. Just for the usability I propose to use the serviceID from the config file if it's set:

If serviceId is not specified but there is exactly one in the configuration: use that one.

I think it provides better usability (and it can also solve the testing problem of @adoroszlai )

PR #559, opened 2020-02-17
HDDS-3024. README is missing from the source release tar

elek:

When we do a dist build with `-Psrc` the `README.md` of the root project is not packaged to the tar file which makes it impossible to do a build from the source package as the `README.md` is required by the dist script.

https://issues.apache.org/jira/browse/HDDS-3024

## How was this patch tested?

```
mvn clean install -DskipTests -Psrc -Pdist
cd hadoop-ozone/dist/target
tar zvxf hadoop-ozone-0.5.0-SNAPSHOT-src.tar.gz
cd hadoop-ozone-0.5.0-SNAPSHOT-src
#The build should work from here as this is the source release (=the release)
mvn clean install -DskipTests
```

### DIFF
--- a/hadoop-ozone/dist/src/main/assemblies/ozone-src.xml
+++ b/hadoop-ozone/dist/src/main/assemblies/ozone-src.xml
@@ -33,7 +33,7 @@
       <outputDirectory>/</outputDirectory>
     </file>
     <file>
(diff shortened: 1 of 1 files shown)


elek:
Thanks the review @adoroszlai, I am pushing it to the master right now...

PR #560, opened 2020-02-18
HDDS-3025. Fix fail to connect to s3g by http://ip:port

runzhiwang:

How to reproduce the problem:

If use command `aws s3api --endpoint http://0.0.0.0:9878 create-bucket --bucket test-bucket`, then log error message `Host value mentioned in signed header is not valid. Host: 0.0.0.0:9878` by https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSV4SignatureProcessor.java#L226.

Root cause:
1. when use http://0.0.0.0:9878, `context.getHeaders()` contains <"Host", "0.0.0.0:9878">
2. `new URI("0.0.0.0:9878")` throw `URISyntaxException` in https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSV4SignatureProcessor.java#L222.
3. It is also explained in the follow image from https://stackoverflow.com/questions/28114352/java-lang-illegalargumentexception-illegal-character-in-scheme-at-index-0-loca.

image

https://issues.apache.org/jira/browse/HDDS-3025

## How was this patch tested?

change UT to cover this case.

### DIFF
--- a/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSV4SignatureProcessor.java
+++ b/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSV4SignatureProcessor.java
@@ -219,7 +219,8 @@ public class AWSV4SignatureProcessor implements SignatureProcessor {
     switch (header) {
     case HOST:
       try {
-        URI hostUri = new URI(headerValue);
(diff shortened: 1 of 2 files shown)


runzhiwang:
@elek Could you help review this PR ? Thank you very much.

PR #564, opened 2020-02-19
HDDS-3031. OM HA- Client requests get LeaderNotReadyException after OM's restart.

bharatviswa504:

Fix LeaderNotReady exception which is thrown for write requests after OM restarts.

https://issues.apache.org/jira/browse/HDDS-3031

## How was this patch tested?

Deployed on the cluster and tried the scenario. With the PR fix, now able to execute client requests after OM restart.

I will see if I can add UT for this.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3InitiateMultipartUploadRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3InitiateMultipartUploadRequest.java
@@ -203,7 +203,8 @@ public class S3InitiateMultipartUploadRequest extends OMKeyRequest {
       omClientResponse = new S3InitiateMultipartUploadResponse(
           createErrorOMResponse(omResponse, exception));
     } finally {
-
+      addResponseToDoubleBuffer(transactionLogIndex, omClientResponse,
+          ozoneManagerDoubleBufferHelper);
       if (acquiredBucketLock) {
         omMetadataManager.getLock().releaseWriteLock(BUCKET_LOCK, volumeName,
             bucketName);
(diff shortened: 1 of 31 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
```diff
@@ -164,7 +164,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           // Replay implies the response has already been returned to
           // the client. So take no further action and return a dummy
           // OMClientResponse.
-          LOG.debug("Replayed Transaction {} ignored. Request: {}",
+          LOG.info("Replayed Transaction {} ignored. Request: {}",
```
hanishakoneru: Keeping this log message at info level will flood the log file. It prints the Request body also.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3InitiateMultipartUploadRequest.java
```diff
@@ -203,11 +203,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
       omClientResponse = new S3InitiateMultipartUploadResponse(
           createErrorOMResponse(omResponse, exception));
     } finally {
-      if (omClientResponse != null) {
-        omClientResponse.setFlushFuture(
-            ozoneManagerDoubleBufferHelper.add(omClientResponse,
-                transactionLogIndex));
-      }
+
```
hanishakoneru: Missed adding addResponseToDoubleBuffer() here.

bharatviswa504:
Addressed review comments.

hanishakoneru:
LGTM. +1.
Thanks for fixing this. Will merge this PR.

PR #565, opened 2020-02-19
HDDS-3030. Key Rename should preserve the ObjectID.

bharatviswa504:

On Key Renames, objectID should be preserved from the original Key. 
Currently it is being set to the new transactionLogIndex of the rename request.

Use original ObjectID from actual Key when renaming.

https://issues.apache.org/jira/browse/HDDS-3030

## How was this patch tested?

Existing UT should cover this.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
@@ -210,15 +210,11 @@ public class OMKeyRenameRequest extends OMKeyRequest {
           throw new OMException("Key not found " + fromKey, KEY_NOT_FOUND);
         }
 
-        // Copy fromKeyValue into toKeyValue and set objectID and updateID to
-        // current transactionLogIndex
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java
```diff
@@ -210,15 +210,11 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           throw new OMException("Key not found " + fromKey, KEY_NOT_FOUND);
         }
 
-        // Copy fromKeyValue into toKeyValue and set objectID and updateID to
-        // current transactionLogIndex
-        toKeyValue = fromKeyValue.copyObject(false);
-        toKeyValue.setObjectID(trxnLogIndex);
-        toKeyValue.setUpdateID(trxnLogIndex);
```
hanishakoneru: We should still set the UpdateID here.

bharatviswa504: Line no 213 does that.

hanishakoneru:
LGTM. +1

PR #566, opened 2020-02-19
HDDS-3034. Broken return code check in unit/integration

adoroszlai:

Make result of pipeline commands reflect return code of failed command, if any, in `unit.sh` and `integration.sh`.

https://issues.apache.org/jira/browse/HDDS-3034

## How was this patch tested?

Introduced non-test failure (compilation error) locally.  Verified exit code of `unit.sh`.


adoroszlai:

PR #570, opened 2020-02-20
HDDS-3035. Add ability to enable Ratis metrics in OzoneManager.

swagle:
Enabled Dropwizard metrics registry to publish all OM Ratis metrics to JMX and configured metrics system endpoint.

https://issues.apache.org/jira/browse/HDDS-3035

## How was this patch tested?
1. Added an integration test 
2. Verified using docker build


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1068,14 +1074,20 @@ public OMMetrics getMetrics() {
    * Start service.
    */
   public void start() throws IOException {
-
     omClientProtocolMetrics.register();
+    HddsServerUtil.initializeMetrics(configuration, "OzoneManager");
```
avijayanhwx: One comment. We can probably add ratis reporters only if Ratis is enabled.

avijayanhwx:
Thank you for the fix @swagle. I am OK with merging it.

PR #571, opened 2020-02-20
HDDS-3042. Support running full Ratis pipeline from IDE (IntelliJ)

elek:

[HDDS-1522](https://issues.apache.org/jira/browse/HDDS-1522) introduced a method to run full cluster in IntelliJ. The runner configurations can be copied with a shell script and a basic `ozone-site.xml` and log configuration to make it easy to run ozone from IDE.

Unfortunately this setup supports only one Datanode and it's hard to debug full Ozone pipeline (3 datanodes) from IDE.

This patch provides 3 different configuration for 3 datanodes with different ports to make it possible to run them on the same host from the IDE.

https://issues.apache.org/jira/browse/HDDS-3042

## How was this patch tested?

Copy the scripts to the IntelliJ:

```
./hadoop-ozone/dev-support/intellij/install-runconfigs.sh
```

And run the following run Configuration (in this order)
 * StorageContainerManagerInit
 * StorageContainerManager
 * OzoneManagerInit
 * OzoneManager
 * Datanode1
 * Datanode2
 * Datanode3

You can test the cluster with the OzoneShell runner (it creates a volume, but it can be adjusted to create a bucket and put a key)


hadoop-ozone/dev-support/intellij/runConfigurations/Datanode2.xml
```diff
@@ -0,0 +1,35 @@
...
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Datanode2" type="Application" factoryName="Application">
+    <option name="ALTERNATIVE_JRE_PATH" value="13" />
```
xiaoyuyao: Some setup may not have JRE 13. Should we keep JDK8 as a minimum supported Java version that is common on Mac Dev box?

elek: Good point. Those lines seem to be optional. I removed them...

PR #575, opened 2020-02-20
HDDS-3046. Fix Retry handling in Rpc Client.

bharatviswa504:

Right now for all other exceptions other than serviceException we use FailOverOnNetworkException.

This Exception policy is created with 15 max fail overs and 15 retries. 

 

retryPolicyOnNetworkException.shouldRetry(
 exception, retries, failovers, isIdempotentOrAtMostOnce);
2 issues with this:

When shouldRetry returns action FAILOVER_AND_RETRY, it will stuck with same OM, and does not perform failover to next OM.  As OMFailoverProxyProvider#performFailover() is a dummy call does not perform any failover.
When ozone.client.failover.max.attempts is set to 15, now with 2 policies with each set to 15, we will retry 15*2 times in worst scenario. 

https://issues.apache.org/jira/browse/HDDS-3046

## How was this patch tested?

Deployed fix on OM HA cluster and did failover and it worked fine. And existing tests should cover this.


hanishakoneru:
LGTM. +1. Thanks Bharat for fixing this.

PR #576, opened 2020-02-20
HDDS-2716. Add integration test to verify pipeline closed on read statemachine failure

swagle:
If readStateMachine call fails there is no way to recover for the follower and RATIS-795 results in calling notifyLogFailed() which closes the pipeline.

https://issues.apache.org/jira/browse/HDDS-2716

## How was this patch tested?
Only integration test added.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineReportHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineReportHandler.java
@@ -114,8 +114,10 @@ public class PipelineReportHandler implements
     }
 
     if (pipeline.getPipelineState() == Pipeline.PipelineState.ALLOCATED) {
-      LOGGER.info("Pipeline {} {} reported by {}", pipeline.getFactor(),
-          pipeline.getId(), dn);
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("Pipeline {} {} reported by {}", pipeline.getFactor(),
+            pipeline.getId(), dn);
+      }
       if (pipeline.isHealthy()) {
         pipelineManager.openPipeline(pipelineID);
         if (pipelineAvailabilityCheck && scmSafeModeManager.getInSafeMode()) {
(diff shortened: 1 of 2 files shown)


swagle:
> https://github.com/adoroszlai/hadoop-ozone/runs/459688379

Hi Attila, I rebase this with master and ran 20 times with minor changes similar to TestDeletedWithSlowFollower and I get it to work locally, could you please relaunch the 20 test run you did?

adoroszlai:
Thank @swagle for updating the patch.  Looks more stable now, it improved to 1 failure out of 20 runs:

https://github.com/adoroszlai/hadoop-ozone/runs/469165226

swagle:
Hi @adoroszlai, I am not able to repro this problem and the test exception trace is not sufficient to identify what is wrong:
java.io.IOException: INTERNAL_ERROR org.apache.hadoop.ozone.om.exceptions.OMException: Allocated 0 blocks. Requested 1 blocks
@bharatviswa504 any idea if a recent regression was introduced?

bshashikant:
> Hi @adoroszlai, I am not able to repro this problem and the test exception trace is not sufficient to identify what is wrong:
> java.io.IOException: INTERNAL_ERROR org.apache.hadoop.ozone.om.exceptions.OMException: Allocated 0 blocks. Requested 1 blocks
> @bharatviswa504 any idea if a recent regression was introduced?

Thanks @swagle for continuing with this and @adoroszlai for testing it out. @swagle , Let's add a container creation call to the after creating the ozone key and before shutting down the datanode and use ContainerTestHelper.isRatisFollower/isRatisLeader Api to selectively shhutdown dn in the test instead of relying on the leader info from the pipeline object.. Creating the container before writing data will make sure the pipeline is healthy and way before writing and using dn to determine the ratis role will always ensure you are shutting down a follower always rather than relying on scm to tell which may be outdated?

swagle:
> Thanks @swagle for the latest update. I think it is OK: 19/20 runs passed, the remaining one crashed, filed [HDDS-3104](https://issues.apache.org/jira/browse/HDDS-3104) for that, I observed it elsewhere, too.
> 
> https://github.com/adoroszlai/hadoop-ozone/runs/474218807

Thanks @adoroszlai can this be merged?

PR #582, opened 2020-02-22
HDDS-3047. ObjectStore#listVolumesByUser and CreateVolumeHandler#call should get principal name by default

smengcl:

`ObjectStore#listVolumesByUser` and `CreateVolumeHandler#call` should get user's full principal name instead of login name by default. When Kerberos is enabled, `getUserName()` returns full principal name e.g. `om/om@EXAMPLE.COM`, but `getShortUserName()` will return login name e.g. `hadoop`.

When the user creates a volume with ozone shell (uses `getUserName()` internally) then try to list it with `ObjectStore#listVolumesByUser(null, ...)` (uses `getShortUserName()` by default), the user won't see any volumes listed because of the mismatch.

For more information, see the Apache JIRA description.

https://issues.apache.org/jira/browse/HDDS-3047

## How was this patch tested?

- Manually tested in `ozonesecure` docker-compose cluster.
- ~Might add a test case in secure integration test, like in `TestSecureOzoneRpcClient` or `TestSecureOzoneCluster`~ After discussing with Xiaoyu, it seems it is not viable to do the test via integration test because in unit test/integration test scm, om, etc. all run in one JVM, and one JVM can't have multiple security context (?). Will add an acceptance (robot) test instead.

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/ObjectStore.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/ObjectStore.java
@@ -250,7 +250,7 @@ public class ObjectStore {
       String volumePrefix, String prevVolume)
       throws IOException {
     if(Strings.isNullOrEmpty(user)) {
-      user = UserGroupInformation.getCurrentUser().getShortUserName();
+      user = UserGroupInformation.getCurrentUser().getUserName();
     }
     return new VolumeIterator(user, volumePrefix, prevVolume);
   }
--- a/hadoop-ozone/dist/src/main/smoketest/ozonefs/ozonefs.robot
+++ b/hadoop-ozone/dist/src/main/smoketest/ozonefs/ozonefs.robot
@@ -33,7 +33,7 @@ Check volume from ozonefs
     ${result} =         Execute               ozone sh volume list
                         Should contain    ${result}         fstest
                         Should contain    ${result}         fstest2
-                        Should contain    ${result}         "admin" : "testuser/scm@EXAMPLE.COM"
+                        Should Match Regexp  ${result}      "admin" : "(hadoop|testuser\/scm@EXAMPLE\.COM)"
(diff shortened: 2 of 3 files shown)


smengcl:
Added robot test to verify the fix.

So without the change in `CreateVolumeHandler`, `Check volume from ozonefs` step will fail since the "admin" field is short user name:
```
Check volume from ozonefs                                             | FAIL |
'{
  "metadata" : { },
  "name" : "fstest",
  "admin" : "root",
  "owner" : "testuser/scm@EXAMPLE.COM",
  "creationTime" : "2020-03-24T13:24:58.505Z",
  "acls" : [ {
    "type" : "USER",
    "name" : "testuser/scm@EXAMPLE.COM",
    "aclScope" : "ACCESS",
    "aclList" : [ "ALL" ]
  }, {
    "type" : "GROUP",
    "name" : "root",
    "aclScope" : "ACCESS",
    "aclList" : [ "ALL" ]
  } ],
  "quota" : 109951162777600
}
{
  "metadata" : { },
  "name" : "fstest2",
  "admin" : "root",
  "owner" : "testuser/scm@EXAMPLE.COM",
  "creationTime" : "2020-03-24T13:25:02.525Z",
  "acls" : [ {
    "type" : "USER",
    "name" : "testuser/scm@EXAMPLE.COM",
    "aclScope" : "ACCESS",
    "aclList" : [ "ALL" ]
  }, {
    "type" : "GROUP",
    "name" : "root",
    "aclScope" : "ACCESS",
    "aclList" : [ "ALL" ]
  } ],
  "quota" : 109951162777600
}' does not contain '"admin" : "testuser/scm@EXAMPLE.COM"'
```

smengcl:
The test failure is due to hard-coding the check value for ozonesecure.
Need to skip it for non-secure.

smengcl:
Rebased to latest master to include test failure fix HDDS-3284.

smengcl:
Thanks @xiaoyuyao for the review!

PR #586, opened 2020-02-23
HDDS-3058. OzoneFileSystem should override unsupported set type FileS…

linyiqun:
OzoneFileSystem overrides unsupported set type FileSystem API.

https://issues.apache.org/jira/browse/HDDS-3058

## How was this patch tested?
Just the unsupported API overriding, seems don't need to have a test.


ayushtkn:
Seems there is a test failure, Can you give a check once

PR #588, opened 2020-02-23
HDDS-2886. parse and dump datanode segment file to pritable text

mukul1987:

This change adds a tool to parse datanode's ratis log files, this tool will be useful to debug issues.

https://issues.apache.org/jira/browse/HDDS-2886

## How was this patch tested?

Unit tests, added the log files into the jira.

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/segmentparser/DatanodeParser.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/segmentparser/DatanodeParser.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.ozone.segmentparser;
+
+import org.apache.hadoop.hdds.cli.HddsVersionProvider;
+import org.apache.hadoop.ozone.container.common.transport.server
+    .ratis.ContainerStateMachine;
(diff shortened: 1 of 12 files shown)


hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/datanode/cli/ParseDnRatisLogSegment.java
```diff
@@ -0,0 +1,64 @@
...
+    name = "dnRatisparser", mixinStandardHelpOptions = true)
+public class ParseDnRatisLogSegment implements Runnable {
+  @CommandLine.Option(names = {"-s", "--segmentPath"}, required = true,
+      description = "Path of the segment file")
+  private static File segmentFile;
+
+  private static String smToContainerLogString(
+      StateMachineLogEntryProto logEntryProto) {
+    return ContainerStateMachine.
+        smProtoToString(RaftGroupId.randomId(), null, logEntryProto);
```
bharatviswa504: Passing randomGroupID and setting this in pipelineID for ContainerCommandRequestProto. So, when it is printing pipelineID will be dummy value right?

mukul1987: yes, I added a dummy pipeline ID for this.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/segmentparser/DatanodeParser.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Command line utility to parse and dump a datanode ratis segment file.
+ */
+@CommandLine.Command(
+    name = "datanode",
+    description = "dump datanode segment file",
+    mixinStandardHelpOptions = true,
+    versionProvider = HddsVersionProvider.class)
+public class DatanodeParser extends BaseLogParser implements Callable<Void> {
```
bharatviswa504: Can we name this class as DatanodeRatisLogParser

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/segmentparser/GenericParser.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Command line utility to parse and dump any generic ratis segment file.
+ */
+@CommandLine.Command(
+    name = "generic",
+    description = "dump generic ratis segment file",
+    mixinStandardHelpOptions = true,
+    versionProvider = HddsVersionProvider.class)
+public class GenericParser extends BaseLogParser implements Callable<Void> {
```
bharatviswa504: Can we rename this as GenericRatisLogParser?

mukul1987:
Thanks for the review @bharatviswa504 , addressed comments in the next path.

PR #590, opened 2020-02-24
HDDS-3049. Fix Replication factor passed in create API doesn't take effect

captainzmc:

Replication did not take effect when we created the file through o3fs. This PR fix that.

https://issues.apache.org/jira/browse/HDDS-3049

## How was this patch tested?

Ut already exists

### DIFF
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
@@ -210,13 +210,21 @@ public class BasicOzoneClientAdapterImpl implements OzoneClientAdapter {
   }
 
   @Override
-  public OzoneFSOutputStream createFile(String key, boolean overWrite,
-      boolean recursive) throws IOException {
+  public OzoneFSOutputStream createFile(String key, short replication,
+      boolean overWrite, boolean recursive) throws IOException {
     incrementCounter(Statistic.OBJECTS_CREATED);
     try {
-      OzoneOutputStream ozoneOutputStream = bucket
-          .createFile(key, 0, replicationType, replicationFactor, overWrite,
-              recursive);
+      OzoneOutputStream ozoneOutputStream = null;
+      if (replication == ReplicationFactor.ONE.getValue()
+          || replication == ReplicationFactor.THREE.getValue()) {
+        ReplicationFactor clientReplication = ReplicationFactor
+            .valueOf(replication);
+        ozoneOutputStream = bucket.createFile(key, 0, replicationType,
+            clientReplication, overWrite, recursive);
+      } else {
+        ozoneOutputStream = bucket.createFile(key, 0, replicationType,
+            replicationFactor, overWrite, recursive);
+      }
       return new OzoneFSOutputStream(ozoneOutputStream.getOutputStream());
     } catch (OMException ex) {
       if (ex.getResult() == OMException.ResultCodes.FILE_ALREADY_EXISTS
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
@@ -210,6 +210,25 @@ public class TestOzoneFileInterfaces {
     assertEquals(statistics.getLong("objects_read").longValue(), 1);
   }
 
+  @Test
+  public void testReplication() throws IOException {
+    int stringLen = 20;
+    String data = RandomStringUtils.randomAlphanumeric(stringLen);
+    String filePath = RandomStringUtils.randomAlphanumeric(5);
+
+    Path pathIllegal = createPath("/" + filePath + "illegal");
+    try (FSDataOutputStream streamIllegal = fs.create(pathIllegal, (short)2)) {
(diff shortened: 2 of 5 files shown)


hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -210,10 +210,14 @@ protected void incrementCounter(Statistic objectsRead) {
...
 
   @Override
-  public OzoneFSOutputStream createFile(String key, boolean overWrite,
+  public OzoneFSOutputStream createFile(String key, short replication, boolean overWrite,
       boolean recursive) throws IOException {
     incrementCounter(Statistic.OBJECTS_CREATED);
     try {
+      if (replication == ReplicationFactor.ONE.getValue()
+          || replication == ReplicationFactor.THREE.getValue()) {
+        replicationFactor = ReplicationFactor.valueOf(replication);
```
adoroszlai: `replicationFactor` can now be a local variable.

captainzmc:
> Thanks @captainzmc for working on this fix.
> 
> Maybe I'm missing something, but with this change files are created with factor=1 by default on a cluster configured with default factor=3. Shouldn't `BasicOzoneFileSystem` override `getDefaultReplication()` to return the configured replication factor instead of 1?
> 
> Here's how I tested:
> 
> ```
> $ ozone fs -put /etc/passwd o3fs://bucket1.vol1/passwd
> $ ozone sh key info /vol1/bucket1/passwd
> { ... "replicationFactor" : 1, ... }
> ```
Hi @adoroszlai Thanks for you reviewing. Yes, use ozone fs or ozone sh is ok. But we test o3fs using org.apache.hadoop.fs.FileSystem API and there were a lot of problems. For more details you can see [here](https://issues.apache.org/jira/browse/HDDS-3048).

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
captainzmc: The replicationFactor is initialized in the constructor. When user-set Replication is invalid (neither 1 nor 3), the previous default is used.

adoroszlai: Please correct me, but it looks like requests with valid `replication` will override the default value for later invalid requests.  Let's say that the cluster has `replicationFactor=3` configured and create requests with the following `replication` values are issued:

1. `replication=2`
2. `replication=1`
3. `replication=2`

The requests in step 1 and 3 will behave differently, since `replicationFactor` was changed by the request in step 2.

(It could be argued that requests with `replication=2` or other invalid value should be rejected, but that's another topic.)

captainzmc: Yes, you're right. This is a place where it is possible to override the default replication. I'm going to add a local variable here to avoid this problem.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -210,13 +210,33 @@ protected void incrementCounter(Statistic objectsRead) {
   }
 
   @Override
-  public OzoneFSOutputStream createFile(String key, boolean overWrite,
+  public OzoneFSOutputStream createFile(String key,
+      short replication,
```
ChenSammi: Please try to put parameters in the same line. One line one parameter is not recommended. 
Same comment for all the following piece of code.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -210,13 +210,33 @@ protected void incrementCounter(Statistic objectsRead) {
...
+            replicationType,
+            clientReplication,
+            overWrite,
+            recursive);
+      } else {
+        ozoneOutputStream = bucket.createFile(
+            key,
+            0,
+            replicationType,
+            replicationFactor,
```
ChenSammi: If the replicationFactor is 2, will the bucket create the key using system default replication here?

ChenSammi:
@captainzmc , the commit seems not clean, there is file like hadoop-ozone/recon/src/main/resources/webapps/recon/ozone-recon-web/yarn.lock, I'm not sure it's related with this fix or not. Can you double check it? 

Also please add new unit tests for replication factor, verify that if the replication factor is supported, new created key will has the specified replication factor, otherwise, it matches the system default replication factor.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
captainzmc: Yes, if the replicationFactor is not 1 or 3, the default value is used.

adoroszlai:
Thanks @captainzmc for updating the patch.

ChenSammi:
Thanks @captainzmc for the contribution and @adoroszlai for review the code.

PR #594, opened 2020-02-24
HDDS-2995. Add integration test for Recon's Passive SCM state.

avijayanhwx:

- Verify Recon gets pipeline, node and container report from Datanode.
- Verify SCM metadata state == Recon metadata state (Create pipeline , Close pipeline, create container)
- Fix issue in DN StateContext.

https://issues.apache.org/jira/browse/HDDS-2995

## How was this patch tested?
Manually tested.


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestReconAsPassiveScm.java
```diff
@@ -0,0 +1,213 @@
...
+public class TestReconAsPassiveScm {
+
+  private MiniOzoneCluster cluster = null;
+  private OzoneConfiguration conf;
+
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Before
+  public void init() throws Exception {
```
swagle: Do we need to start a new cluster for every test? This could be @BeforeClass, can it not?

avijayanhwx: The tests are not mutually exclusive. Hence, it may lead to side effects if we move it to @BeforeClass, and the tests run in random order or in parallel.

avijayanhwx:
The IT failures are related. Working on fixing them.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestReconAsPassiveScm.java
swagle: Ok, sounds good.

avijayanhwx:
I am not able to reproduce this issue locally. Trying to root cause this based of PR runs.

avijayanhwx:
The test failures seem to be unrelated. I will try one more time.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneCluster.java
```diff
@@ -273,6 +283,10 @@ void restartHddsDatanode(DatanodeDetails dn, boolean waitForDatanode)
     protected Optional<Long> streamBufferMaxSize = Optional.empty();
     protected Optional<Long> blockSize = Optional.empty();
     protected Optional<StorageUnit> streamBufferSizeUnit = Optional.empty();
+    protected boolean includeRecon = false;
+    protected Optional<Integer> reconHttpPort = Optional.empty();
+    protected Optional<Integer> reconDatanodePort = Optional.empty();
```
adoroszlai: Nit: please use `OptionalInt`.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneClusterImpl.java
```diff
@@ -293,14 +296,9 @@ public void restartOzoneManager() throws IOException {
...
-      reconServer.stop();
-      reconServer.join();
-      reconServer = new ReconServer();
-      reconServer.execute(new String[]{});
-    } catch (Exception e) {
-      LOG.info("Exception while restarting Recon", e);
-    }
+    stopRecon(reconServer);
+    reconServer = new ReconServer();
+    reconServer.execute(new String[]{});
```
adoroszlai: Can call `startRecon()` instead to avoid duplicate code?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerReportHandler.java
```diff
@@ -63,19 +63,21 @@ public void onMessage(final ContainerReportFromDatanode reportFromDatanode,
...
-          ContainerWithPipeline containerWithPipeline =
-              scmClient.getContainerWithPipeline(id.getId());
-          LOG.debug("Verified new container from SCM {} ",
-              containerWithPipeline.getContainerInfo().containerID());
-          containerManager.addNewContainer(id.getId(), containerWithPipeline);
-        } catch (IOException ioEx) {
-          LOG.error("Exception while getting new container info from SCM",
-              ioEx);
-          return;
+      synchronized (getContainerManager()) {
```
adoroszlai: Shouldn't it sync on `containerManager` variable?

avijayanhwx:
Thanks for the review @saintstack  & @swagle. Closing this PR since it is taking longer than expected to fix the flakiness of Recon's integration tests. This patch contains other important refactoring that need not wait for the integration tests to be added. I will be creating separate JIRAs for that.

PR #595, opened 2020-02-24
HDDS-3065. Ozone Filesystem should return real default replication

adoroszlai:

Make `BasicOzoneFileSystem#getDefaultReplication()` return the configured replication factor instead of 1.

https://issues.apache.org/jira/browse/HDDS-3065

## How was this patch tested?

Added unit test for default and custom config cases.

https://github.com/adoroszlai/hadoop-ozone/runs/465304016


adoroszlai:
Thanks @xiaoyuyao for reviewing and committing it.

PR #60, opened 2019-10-18
HDDS-2325. BenchMarkDatanodeDispatcher genesis test is failing with NPE

elek:
## What changes were proposed in this pull request?

TLDR; Fix BenchmarkDatanodeDispatcher.writeChunk test.

Genesis is a microbenchmark tool for Ozone based on JMH (https://openjdk.java.net/projects/code-tools/jmh/).

Due to the recent Datanode changes the BenchMarkDatanodeDispatcher is failing with NPE:

```
java.lang.NullPointerException
	at org.apache.hadoop.ozone.container.common.interfaces.Handler.<init>(Handler.java:69)
	at org.apache.hadoop.ozone.container.keyvalue.KeyValueHandler.<init>(KeyValueHandler.java:114)
	at org.apache.hadoop.ozone.container.common.interfaces.Handler.getHandlerForContainerType(Handler.java:78)
	at org.apache.hadoop.ozone.genesis.BenchMarkDatanodeDispatcher.initialize(BenchMarkDatanodeDispatcher.java:115)
	at org.apache.hadoop.ozone.genesis.generated.BenchMarkDatanodeDispatcher_createContainer_jmhTest._jmh_tryInit_f_benchmarkdatanodedispatcher0_G(BenchMarkDatanodeDispatcher_createContainer_jmhTest.java:438)
	at org.apache.hadoop.ozone.genesis.generated.BenchMarkDatanodeDispatcher_createContainer_jmhTest.createContainer_Throughput(BenchMarkDatanodeDispatcher_createContainer_jmhTest.java:71)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.openjdk.jmh.runner.BenchmarkHandler$BenchmarkTask.call(BenchmarkHandler.java:453)
	at org.openjdk.jmh.runner.BenchmarkHandler$BenchmarkTask.call(BenchmarkHandler.java:437)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #596, opened 2020-02-24
HDDS-3066. SCM crash during loading containers to DB.

bharatviswa504:

 This is happening because pipeline scrubber came and removed pipeline, and it closed pipeline and removed from DB and triggered close containers to set them to CLOSING. When SCM is restarted before close container command is handled and change the state to CLOSING, the below issue can happen.

 

This can happen in other scenarios like when safeModeHandler calls finalizeAndDestroyPipeline and do SCM restart. 

 

The root cause for this is Pipeline removed from DB and the container is in open state in this scenario, and when trying to get pipeline we will crash SCM due to the PipelineNotFoundException error.

 `2020-02-21 13:57:34,888 [main] ERROR org.apache.hadoop.hdds.scm.server.StorageContainerManagerStarter: SCM start failed with exception org.apache.hadoop.hdds.scm.pipeline.PipelineNotFoundException: PipelineID=35dff62d-9bfa-449b-b6e8-6f00cc8c1b6e not found at org.apache.hadoop.hdds.scm.pipeline.PipelineStateMap.getPipeline(PipelineStateMap.java:133) at org.apache.hadoop.hdds.scm.pipeline.PipelineStateMap.addContainerToPipeline(PipelineStateMap.java:110) at org.apache.hadoop.hdds.scm.pipeline.PipelineStateManager.addContainerToPipeline(PipelineStateManager.java:59) at org.apache.hadoop.hdds.scm.pipeline.SCMPipelineManager.addContainerToPipeline(SCMPipelineManager.java:309) at org.apache.hadoop.hdds.scm.container.SCMContainerManager.loadExistingContainers(SCMContainerManager.java:121) at org.apache.hadoop.hdds.scm.container.SCMContainerManager.<init>(SCMContainerManager.java:107) at org.apache.hadoop.hdds.scm.server.StorageContainerManager.initializeSystemManagers(StorageContainerManager.java:412) at org.apache.hadoop.hdds.scm.server.StorageContainerManager.<init>(StorageContainerManager.java:283) at org.apache.hadoop.hdds.scm.server.StorageContainerManager.<init>(StorageContainerManager.java:215) at org.apache.hadoop.hdds.scm.server.StorageContainerManager.createSCM(StorageContainerManager.java:612) at org.apache.hadoop.hdds.scm.server.StorageContainerManagerStarter$SCMStarterHelper.start(StorageContainerManagerStarter.java:142) at org.apache.hadoop.hdds.scm.server.StorageContainerManagerStarter.startScm(StorageContainerManagerStarter.java:117) at org.apache.hadoop.hdds.scm.server.StorageContainerManagerStarter.call(StorageContainerManagerStarter.java:66) at org.apache.hadoop.hdds.scm.server.StorageContainerManagerStarter.call(StorageContainerManagerStarter.java:42) at picocli.CommandLine.execute(CommandLine.java:1173) at picocli.CommandLine.access$800(CommandLine.java:141) at picocli.CommandLine$RunLast.handle(CommandLine.java:1367) at picocli.CommandLine$RunLast.handle(CommandLine.java:1335) at picocli.CommandLine$AbstractParseResultHandler.handleParseResult(CommandLine.java:1243) at picocli.CommandLine.parseWithHandlers(CommandLine.java:1526) at picocli.CommandLine.parseWithHandler(CommandLine.java:1465) at org.apache.hadoop.hdds.cli.GenericCli.execute(GenericCli.java:65) at org.apache.hadoop.hdds.cli.GenericCli.run(GenericCli.java:56) at org.apache.hadoop.hdds.scm.server.StorageContainerManagerStarter.main(StorageContainerManagerStarter.java:55) 2020-02-21 13:57:34,892 [shutdown-hook-0] INFO org.apache.hadoop.hdds.scm.server.StorageContainerManagerStarter: SHUTDOWN_MSG: /************************************************************ SHUTDOWN_MSG: Shutting down StorageContainerManager at om-ha-1.vpc.cloudera.com/10.65.51.49 ************************************************************/
`

https://issues.apache.org/jira/browse/HDDS-3066

## How was this patch tested?

Thank You @nandakumar131 for the offline discussion.
Existing tests. Deployed the fix on the cluster, and SCM able to boot up.

`2020-02-24 12:02:12,531 [main] WARN org.apache.hadoop.hdds.scm.container.SCMContainerManager: Found a Container ContainerInfo{id=3, state=OPEN, pipelineID=PipelineID=afb60e8a-0a69-410a-8699-d2a75e053225, stateEnterTime=1159646, owner=om2} which is in OPEN state with out a pipeline PipelineID=afb60e8a-0a69-410a-8699-d2a75e053225. Triggering Close Container.`

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
@@ -360,7 +360,7 @@ public class SCMContainerManager implements ContainerManager {
       containerStateManager.updateContainerState(containerID, event);
       final LifeCycleState newState = container.getState();
 
-      if (skipPipelineToContainerRemove) {
+      if (!skipPipelineToContainerRemove) {
         if (oldState == LifeCycleState.OPEN &&
             newState != LifeCycleState.OPEN) {
           pipelineManager
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
@@ -24,13 +24,17 @@ import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ContainerInfoProto;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.LifeCycleState;
 import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.container.metrics.SCMContainerManagerMetrics;
+import org.apache.hadoop.hdds.scm.events.SCMEvents;
 import org.apache.hadoop.hdds.scm.exceptions.SCMException;
 import org.apache.hadoop.hdds.scm.pipeline.Pipeline;
 import org.apache.hadoop.hdds.scm.pipeline.PipelineManager;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationFactor;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationType;
+import org.apache.hadoop.hdds.scm.pipeline.PipelineNotFoundException;
 import org.apache.hadoop.hdds.server.ServerUtils;
+import org.apache.hadoop.hdds.server.events.EventPublisher;
+import org.apache.hadoop.hdds.server.events.EventQueue;
 import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.hdds.utils.BatchOperation;
 import org.apache.hadoop.hdds.utils.MetadataStore;
@@ -73,6 +77,8 @@ public class SCMContainerManager implements ContainerManager {
 
   private final SCMContainerManagerMetrics scmContainerManagerMetrics;
 
+  private final EventPublisher eventPublisher;
+
   /**
    * Constructs a mapping class that creates mapping between container names
    * and pipelines.
@@ -82,10 +88,12 @@ public class SCMContainerManager implements ContainerManager {
    * in MB.
    * @param conf - {@link Configuration}
    * @param pipelineManager - {@link PipelineManager}
+   * @param eventQueue
    * @throws IOException on Failure.
    */
   public SCMContainerManager(final Configuration conf,
-      PipelineManager pipelineManager) throws IOException {
+      PipelineManager pipelineManager, EventQueue eventQueue)
+      throws IOException {
 
     final File containerDBPath = getContainerDBPath(conf);
     final int cacheSize = conf.getInt(OZONE_SCM_DB_CACHE_SIZE_MB,
@@ -103,6 +111,7 @@ public class SCMContainerManager implements ContainerManager {
     this.numContainerPerOwnerInPipeline = conf
         .getInt(ScmConfigKeys.OZONE_SCM_PIPELINE_OWNER_CONTAINER_COUNT,
             ScmConfigKeys.OZONE_SCM_PIPELINE_OWNER_CONTAINER_COUNT_DEFAULT);
+    this.eventPublisher = eventQueue;
 
     loadExistingContainers();
 
@@ -117,9 +126,17 @@ public class SCMContainerManager implements ContainerManager {
           ContainerInfoProto.PARSER.parseFrom(entry.getValue()));
       Preconditions.checkNotNull(container);
       containerStateManager.loadContainer(container);
-      if (container.getState() == LifeCycleState.OPEN) {
-        pipelineManager.addContainerToPipeline(container.getPipelineID(),
-            ContainerID.valueof(container.getContainerID()));
+      try {
+        if (container.getState() == LifeCycleState.OPEN) {
+          pipelineManager.addContainerToPipeline(container.getPipelineID(),
+              ContainerID.valueof(container.getContainerID()));
+        }
+      } catch (PipelineNotFoundException ex) {
+        LOG.warn("Found a Container {} which is in {} state with out a " +
+            "pipeline {}. Triggering Close Container.", container,
+            container.getState(), container.getPipelineID());
+        eventPublisher.fireEvent(SCMEvents.CLOSE_CONTAINER,
+            container.containerID());
       }
     }
   }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
@@ -131,9 +131,10 @@ public class SCMContainerManager implements ContainerManager {
 
         // Not firing CLOSE_CONTAINER event because CloseContainer Event
         // event queue handler is not registered by the time when we come
-        // here. So, we are calling update Container state to CLOSING, and
-        // replication manager takes care of send close commands to datanode
-        // to close containers on the datanode.
+        // here. So, we are calling update Container state to set
+        // container state to CLOSING, and later replication manager takes care
+        // of send close commands to datanode to close containers on the
+        // datanode.
 
         // Skipping pipeline to container removal because, we got a
         // pipelineNotFoundException when adding container to
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
@@ -128,17 +128,6 @@ public class SCMContainerManager implements ContainerManager {
         LOG.warn("Found a Container {} which is in {} state with pipeline {} " +
                 "that does not exist. Closing Container.", container,
             container.getState(), container.getPipelineID());
-
-        // Not firing CLOSE_CONTAINER event because CloseContainer event
-        // handler is not registered by the time when we come
-        // here. So, we are calling update Container state to set
-        // container state to CLOSING, and later replication manager takes care
-        // of send close commands to datanode to close containers on the
-        // datanode.
-
-        // Skipping pipeline to container removal because, we got a
-        // pipelineNotFoundException when adding container to
-        // pipeline. So, we can only update container state.
         updateContainerState(container.containerID(),
             HddsProtos.LifeCycleEvent.FINALIZE, true);
       }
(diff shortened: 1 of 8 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -117,9 +126,17 @@ private void loadExistingContainers() throws IOException {
           ContainerInfoProto.PARSER.parseFrom(entry.getValue()));
       Preconditions.checkNotNull(container);
       containerStateManager.loadContainer(container);
-      if (container.getState() == LifeCycleState.OPEN) {
-        pipelineManager.addContainerToPipeline(container.getPipelineID(),
-            ContainerID.valueof(container.getContainerID()));
+      try {
+        if (container.getState() == LifeCycleState.OPEN) {
```
xiaoyuyao: Pipeline scrubber only destroy pipeline that has been in ALLOCATED for too long. And the container loader only call addContainerToPipeline if the container is in OPEN state.

If I understand correctly, we should not have OPEN container assigned on an ALLOCATED pipeline.

bharatviswa504: When SCM is restarted, all pipelines will be in allocated state, they will be moved to open state. And when pipeline reports are received from DN, they will be moved to Open State.
So, here scrubber removed those pipelines and container is in open State.

2020-02-20 12:42:18,947 [RatisPipelineUtilsThread] INFO org.apache.hadoop.hdds.scm.pipeline.SCMPipelineManager: Destroying pipeline:Pipeline[ Id: 35dff62d-9bfa-449b-b6e8-6f00cc8c1b6e, Nodes: 53fc2e1a-73da-4ae7-8725-9cc23ac6c393{ip: 10.65.54.245, host: om-ha-3.vpc.cloudera.com, networkLocation: /default-rack, certSerialId: null}2346b987-3126-48b8-b2d2-e8244cb2e0ae{ip: 10.65.51.168, host: om-ha-2.vpc.cloudera.com, networkLocation: /default-rack, certSerialId: null}45987d8b-4bfd-4ccc-bf2f-224bcf5b0dcd{ip: 10.65.51.49, host: om-ha-1.vpc.cloudera.com, networkLocation: /default-rack, certSerialId: null}, Type:RATIS, Factor:THREE, State:ALLOCATED, leaderId:null, CreationTimestamp2020-02-20T03:59:02.043Z]
2020-02-20 12:42:18,947 [RatisPipelineUtilsThread] INFO org.apache.hadoop.hdds.scm.pipeline.PipelineStateManager: Pipeline Pipeline[ Id: 35dff62d-9bfa-449b-b6e8-6f00cc8c1b6e, Nodes: 53fc2e1a-73da-4ae7-8725-9cc23ac6c393{ip: 10.65.54.245, host: om-ha-3.vpc.cloudera.com, networkLocation: /default-rack, certSerialId: null}2346b987-3126-48b8-b2d2-e8244cb2e0ae{ip: 10.65.51.168, host: om-ha-2.vpc.cloudera.com, networkLocation: /default-rack, certSerialId: null}45987d8b-4bfd-4ccc-bf2f-224bcf5b0dcd{ip: 10.65.51.49, host: om-ha-1.vpc.cloudera.com, networkLocation: /default-rack, certSerialId: null}, Type:RATIS, Factor:THREE, State:CLOSED, leaderId:null, CreationTimestamp2020-02-20T03:59:02.043Z] moved to CLOSED state

And SCM log in first restart.

Now after 2nd restart the pipeline is not in DB, but the container is in open state.

bharatviswa504: And also I am thinking scrubber should not come and delete pipelines until we are out of SafeMode, will try to do that in a new jira.

bharatviswa504: As mentioned in Jira, another scenario where this can happen

This can happen in other scenarios like when safeModeHandler calls finalizeAndDestroyPipeline and do SCM restart.

As we remove the pipeline from DB and container can be in an open state. (This can happen because close container command is triggered, but not yet processed.) When SCM restart, we can be in this scenario.

xiaoyuyao: bq. And also I am thinking scrubber should not come and delete pipelines until we are out of SafeMode, will try to do that in a new jira.
Agree, we should have a special state to indicate that state of the pipeline. We can fix it in a separate JIRA.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -117,9 +119,28 @@ private void loadExistingContainers() throws IOException {
...
+          pipelineManager.addContainerToPipeline(container.getPipelineID(),
+              ContainerID.valueof(container.getContainerID()));
+        }
+      } catch (PipelineNotFoundException ex) {
+        LOG.warn("Found a Container {} which is in {} state with out a " +
```
xiaoyuyao: NIT: can we reword "with out a pipeline {}" to "with pipeline {} that does not exist"

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
bharatviswa504: Scrubbing to start after safe mode, I have handled in this PR #605 
I see that we can do without introducing a new state.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -117,9 +119,28 @@ private void loadExistingContainers() throws IOException {
...
+            "pipeline {}. Close Container.", container, container.getState(),
+            container.getPipelineID());
+
+        // Not firing CLOSE_CONTAINER event because CloseContainer Event
```
xiaoyuyao: This comment does not match with the code any more. Should we remove it? 
"// Not firing CLOSE_CONTAINER event ..."

bharatviswa504: This was added to explain the reason why we are not calling CLOSE_CONTAINER event to close container. In future, if someone changes this part of the code, they will know the reason in doing this way.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
bharatviswa504: Fixed it.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
xiaoyuyao: I think there are values to keep the scurbber running in safemode. Without it, any pipeline created/restored during safemode will hold there forever if any issue hit during pipeline creation. This prevent new pipeline from being created to exit safemode. 


E.g., when datanodes restart during safemode pipeline creation, before the pipeline report changed the SCM pipeline state from ALLOCATE to OPEN.

bharatviswa504: If old pipelines are reported, and they are only accounted for SafeMode rule calculation.
Once, we are out of safe mode, we are triggering pipeline creation.

bharatviswa504: The fix is in HDDS-3072, we can discuss more on that over #60 
If this is fine, we can get this in.

xiaoyuyao: bq. The main purpose not to run scrub in safe mode is if it is closing the pipelines where datanodes have still not reported, we shall never come out of safe mode.

Multi-raft allows additional pipeline being created on top if existing one if they are not functional. To new OM client to write, there is no difference between pipeline loaded/reported or created/reported. If those loaded but not reported pipeline is not working, we should use scrubber to allow recreate/report. Agree, we can discuss this on #605.

xiaoyuyao:
+1, Let's discuss the scrubber issue in #605.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -117,9 +119,28 @@ private void loadExistingContainers() throws IOException {
...
+        // Not firing CLOSE_CONTAINER event because CloseContainer event
+        // handler is not registered by the time when we come
+        // here. So, we are calling update Container state to set
+        // container state to CLOSING, and later replication manager takes care
+        // of send close commands to datanode to close containers on the
+        // datanode.
+
+        // Skipping pipeline to container removal because, we got a
+        // pipelineNotFoundException when adding container to
+        // pipeline. So, we can only update container state.
```
nandakumar131: We don't need this comment. As `SCMContainerManager` is the one which is taking this decision, it should not file an event to close the container. It is logical to directly update the container state.

It is very explicit that we are performing this in case of `PipelineNotFoundException`, so there is no need to repeat it again.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -117,9 +119,28 @@ private void loadExistingContainers() throws IOException {
...
+        updateContainerState(container.containerID(),
+            HddsProtos.LifeCycleEvent.FINALIZE, true);
```
nandakumar131: We don't have to introduce a new method to `updateContainerState`. Since we are doing this inside a constructor, we don't need any lock.

We can directly call `containerStateManager.updateContainerState(container.containerID(), HddsProtos.LifeCycleEvent.FINALIZE)`

bharatviswa504: This call is being done so that it is reflected in container DB also.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
bharatviswa504: Done.

bharatviswa504:
Thank You @nandakumar131 and @xiaoyuyao for the review.

PR #597, opened 2020-02-24
HDDS-2996. Create REST API to serve Node information and integrate with UI in Recon.

vivekratnavel:

- Added a REST API to serve Node information in Recon
- Added unit tests to test node endpoint response
- Added Robot test to test the node endpoint response
- Disabled Recon tests in Ozone Secure Robot test setup to avoid test failures. This will be enabled in the future after testing and stabilizing Recon in Secure environments.
- Enabled Hash Router in React to fix this issue https://issues.apache.org/jira/browse/HDDS-3011

https://issues.apache.org/jira/browse/HDDS-2996

## How was this patch tested?

This patch was tested with unit tests, robot tests and manual tests.

One can manually test the patch by compiling locally and bringing up a cluster via docker by
```
docker-compose up -d
```
Then pointing the browser to localhost:9888 

This will bring up datanodes UI in Recon as the default. Scale up the number of datanodes or create keys in ozone via freon and see the change in Recon UI.

```
docker-compose scale datanode=6
ozone freon rk --replicationType=RATIS --numOfVolumes=10 --numOfBuckets=10 --numOfKeys=10 --factor=THREE --numOfThreads=20
```


hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/types/DatanodesResponse.java
```diff
@@ -0,0 +1,69 @@
...
+    this(0, new ArrayList<>());
+  }
+
+  public DatanodesResponse(long totalCount,
+                           Collection<DatanodeMetadata> datanodes) {
+    this.totalCount = totalCount;
+    this.datanodes = datanodes;
+  }
+
+  public String toJsonString() {
```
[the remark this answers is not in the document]
vivekratnavel: Removed.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/types/DatanodesResponse.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Class that represents the API Response structure of Datanodes.
+ */
+public class DatanodesResponse {
+  /**
+   * Total count of the datanodes.
+   */
+  @JsonProperty("totalCount")
+  private long totalCount;
```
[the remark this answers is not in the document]
vivekratnavel: We would need this "totalCount" as part of DatanodeMetadata response to support server side pagination in the future.

hadoop-ozone/dist/src/main/compose/ozonesecure/test.sh
```diff
@@ -37,7 +37,7 @@ execute_robot_test s3g s3
 
 execute_robot_test scm scmcli
 
-execute_robot_test scm recon
+# execute_robot_test scm recon
```
elek: Is it intentional. As I see you updated the robot file as well. Does it work. If not (yet) some comments can be added what is missing. Maybe with a JIRA link.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/NodeEndpoint.java
```diff
@@ -0,0 +1,113 @@
...
+
+  private static final Logger LOG =
+      LoggerFactory.getLogger(NodeEndpoint.class);
+
+  private ReconNodeManager nodeManager;
+
+  @Inject
+  NodeEndpoint(OzoneStorageContainerManager reconSCM) {
+    this.nodeManager =
+        (ReconNodeManager) reconSCM.getScmNodeManager();
```
elek: Is there any reason to inject `OzoneStorageContainerManager` instead of `ReconNodeManager` (just curious).

hadoop-ozone/dist/src/main/compose/ozonesecure/test.sh
vivekratnavel: @elek I updated the robot test and docker configs to enable recon as a passive SCM and test the datanodes endpoint. There is still some work needs to be done to support this setup in a secure environment since Datanodes fail to connect to recon and recon fails to connect to SCM with kerberos auth failure. I am planning to handle this in a separate JIRA - https://issues.apache.org/jira/browse/HDDS-3071

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/NodeEndpoint.java
avijayanhwx: @elek Thanks for reviewing this. We cannot inject 'ReconNodeManager' since it is not part of the injected classes in the Controller Module. If we bring in the node manager through injection, we will have to move all the other sub modules within Recon SCM also into injection. Currently, they are instantiated very similar to the way in actual SCM, staying true to the "Recon as Passive SCM" story.

elek: Thanks the answer. I had similar suspicion. IMHO even without dependency injection we can do better separation between creating and wiring the objects on the SCM side. For example we can do the wiring on the SCM side in a separated utility and in this case you can do better wiring on recon.

Or even we can introduce dependency injection on the SCM side as well.

(Not in this patch. Just thinking loudly ;-))

hadoop-ozone/dist/src/main/compose/ozonesecure/test.sh
elek: Thanks a lot. I will add this link to the comment during the commit to make it easier to follow from the code. (Hope you don't mind...)

```
# TODO: https://issues.apache.org/jira/browse/HDDS-3071
# execute_robot_test scm recon
```

vivekratnavel:
Thanks a lot for the review @avijayanhwx and @elek.

 @elek I am going to wait sometime to see how multi-raft works out and then decide how we could improve the display of pipelines in Datanodes page. Potentially there could be too many active pipelines per datanode with multi-raft and I am not sure if it would help the user to list all of them in Datanodes page. But, there is definitely some scope for improvement here. Thanks for your idea!

PR #598, opened 2020-02-25
HDDS-3067. Fix Bug in Scrub Pipeline causing destory pipelines after SCM restart.

bharatviswa504:

Fix destroy of pipelines after SCM restart, due to a bug in loading pipeline setting timestamp with actual pipeline time. In the case of a restart, we should set time to when it is loading from pipeline DB.
 

https://issues.apache.org/jira/browse/HDDS-3067

## How was this patch tested?

Tested it on the cluster and now SCM is not destroying pipelines after the restart. And also fixed RatisPipelineProvider test.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
@@ -153,6 +153,8 @@ public class SCMPipelineManager implements PipelineManager {
           .newBuilder(HddsProtos.Pipeline.PARSER.parseFrom(entry.getValue()));
       Pipeline pipeline = Pipeline.getFromProtobuf(pipelineBuilder.setState(
           HddsProtos.PipelineState.PIPELINE_ALLOCATED).build());
+      // When SCM is restarted, set Creation time with current time.
+      pipeline.setCreationTimestamp(Instant.now());
       Preconditions.checkNotNull(pipeline);
       stateManager.addPipeline(pipeline);
       nodeManager.addPipeline(pipeline);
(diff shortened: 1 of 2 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -153,6 +153,8 @@ protected void initializePipelineState() throws IOException {
           .newBuilder(HddsProtos.Pipeline.PARSER.parseFrom(entry.getValue()));
       Pipeline pipeline = Pipeline.getFromProtobuf(pipelineBuilder.setState(
           HddsProtos.PipelineState.PIPELINE_ALLOCATED).build());
+      // When SCM is restarted, set Creation time with current time.
+      pipeline.setCreationTimestamp(Instant.now());
```
adoroszlai: Nit: this should go after `Preconditions.checkNotNull` in the next line, otherwise the check becomes unnecessary (will never fail).  The result is the same anyway (NPE).

bharatviswa504:
> I'm not sure about the other part: maybe it's just me, but it seems to be a workaround for the behavior of the scrubber, and we are losing information (maybe not important information, I don't know). I think other possible fixes should be considered (eg. adding a new state for the restored pipelines, or introducing a "grace period" in the scrubber after startup).


From my understanding, creation time stamp is set when pipeline is created. This is what we do when SCM is setup fresh and started. 
Now SCM is restarted, all pipelines will be in allocated state, until pipeline reports are received from datanode. So, I assume we can set the creation time with SCM start time and use this to scrub the pipeline. As scrub main purpose is to detect a pipeline is allocated state for more than configured ozone.scm.pipeline.allocated.timeout. But one thing is in DB we don't change the timestamp it will be always with creation timestamp if the pipeline is not destroyed.

Let me know if you have any other approach, from above comment I have not understood the proposal clearly.


It seems that the two changes are not strictly related: the fix for the test is correct by itself, the test passes with or without the call to pipeline.setCreationTimestamp.

Yes, test fix is not related to this, as I am looking for any tests testing Ratis pipeline I have found this and fixed it as part of this PR.

bharatviswa504:
Thank You @adoroszlai for the review and offline discussion.

PR #602, opened 2020-02-25
HDDS-2984. Allocate Block failing with NPE

adoroszlai:

NPE may happen in `SCMPipelineMetrics#incNumBlocksAllocated` if pipeline is not found while incrementing counter.  Given that `Optional#ifPresent` is already checked, I think the intention was to allow such cases.  The problem is that `Optional#of` does not allow `null`.  This change simply replaces it with `Optional#ofNullable`.

https://issues.apache.org/jira/browse/HDDS-2984

## How was this patch tested?

Passes existing tests, although I don't think any specific test covers this case.

https://github.com/adoroszlai/hadoop-ozone/runs/465744823


adoroszlai:
Thanks @elek and @lokeshj1703 for the review, and @elek for committing it.

PR #604, opened 2020-02-25
HDDS-3038. TestRatisPipelineLeader fails since we no longer wait for leader in the HealthyPipelineSafeModeExitRule.

swagle:
Fixed the HealthyPipelineSafeModeExitRule to wait on pipeline healthy state implying, the pipeline leader is available upon OPEN, fixed related tests.

https://issues.apache.org/jira/browse/HDDS-3038

## How was this patch tested?
Ran related integration tests.


ChenSammi:
+1.  I don't see there is CI checks.  But the compile and related UTs all passed locally.

swagle:
Thanks @ChenSammi for review and @adoroszlai for re-running the test.
Failure 1: Seems miniOzone cluster flakiness, hard to determine why that failed
Failure 2: Leader seems to not be available when the test expected it to be, made a change to test to a) wait for a multiple of leader election timeout and then changed the exit criteria to wait for 20 seconds until the leader is verified.

PR #608, opened 2020-02-26
HDDS-3084. Extend network topology acceptance test to read data when datanodes are stopped.

sodonnel:

This adds a new robot test to the network-topology environment, which:

1. Creates a volume, bucket and key.
2. Stops 1 rack and ensures the data is still readable
3. Restart the rack and stop the other rack and again check the data is readable

That way we can have some confidence the data is being written to both racks OK.

One issue with a test like this on a small cluster, is that there is a high chance the data will end up on 2 racks naturally, even if no network topology is configured. If that was the case, we would expect intermittent test failures.

However, if network topology is working fine, then we would not expect any failures.

https://issues.apache.org/jira/browse/HDDS-3084

## How was this patch tested?

Ran the new smoke test locally.

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozone-topology/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/ozone-topology/test.sh
@@ -43,6 +43,10 @@ execute_robot_test scm topology/readdata.robot
 
 start_containers datanode_1 datanode_2 datanode_3
 
+wait_for_port datanode_1 9858 60
+wait_for_port datanode_2 9858 60
+wait_for_port datanode_3 9858 60
+
 stop_containers datanode_4 datanode_5 datanode_6
 
 execute_robot_test scm topology/readdata.robot
--- a/hadoop-ozone/dist/src/main/compose/ozone-topology/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/ozone-topology/test.sh
@@ -33,6 +33,20 @@ execute_robot_test scm basic/basic.robot
 
 execute_robot_test scm topology/scmcli.robot
 
+# Ensure data can be read even when a full rack
+# is stopped.
+execute_robot_test scm topology/loaddata.robot
+
+stop_containers datanode_1 datanode_2 datanode_3
+
+execute_robot_test scm topology/readdata.robot
+
+start_containers datanode_1 datanode_2 datanode_3
+
+stop_containers datanode_4 datanode_5 datanode_6
+
+execute_robot_test scm topology/readdata.robot
+
 stop_docker_env
 
 generate_report
--- a/hadoop-ozone/dist/src/main/compose/testlib.sh
+++ b/hadoop-ozone/dist/src/main/compose/testlib.sh
@@ -128,7 +128,7 @@ execute_command_in_container(){
 ## @param       List of container names, eg datanode_1 datanode_2
 stop_containers() {
   set -e
-  docker-compose -f "$COMPOSE_FILE" stop $@
+  docker-compose -f "$COMPOSE_FILE" --no-ansi stop $@
   set +e
 }
 
@@ -137,7 +137,7 @@ stop_containers() {
 ## @param       List of container names, eg datanode_1 datanode_2
 start_containers() {
   set -e
-  docker-compose -f "$COMPOSE_FILE" start $@
+  docker-compose -f "$COMPOSE_FILE" --no-ansi start $@
   set +e
 }
 
--- a/hadoop-ozone/dist/src/main/compose/testlib.sh
+++ b/hadoop-ozone/dist/src/main/compose/testlib.sh
@@ -124,6 +124,23 @@ execute_command_in_container(){
   set +e
 }
 
+## @description Stop a list of named containers
+## @param       List of container names, eg datanode_1 datanode_2
+stop_containers() {
+  set -e
+  docker-compose -f "$COMPOSE_FILE" stop $@
+  set +e
+}
+
+
+## @description Start a list of named containers
+## @param       List of container names, eg datanode_1 datanode_2
+start_containers() {
+  set -e
+  docker-compose -f "$COMPOSE_FILE" start $@
+  set +e
+}
+
 
 ## @description  Stops a docker-compose based test environment (with saving the logs)
 stop_docker_env(){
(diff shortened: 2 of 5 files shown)


hadoop-ozone/dist/src/main/smoketest/topology/readdata.robot
```diff
@@ -0,0 +1,28 @@
...
+Library             OperatingSystem
+Library             BuiltIn
+Resource            ../commonlib.robot
+
+*** Variables ***
+
+
+*** Test Cases ***
+Read data from previously created key
+    ${output} =         Execute          ozone sh key get /topvol1/bucket1/key1
```
adoroszlai:
```suggestion
    ${output} =         Execute          ozone sh key get /topvol1/bucket1/key1 /tmp/key1
```

The `key get` command requires local filename as second argument, so currently it produces this output (trimmed) and does not really verify ability to read data:

```
Missing required parameter: <fileName>
```

Note that on the second run (for the other rack) it will refuse to overwrite `/tmp/key1`, so

1. delete local file after `key get` OR
2. run test case in different container (eg. `om`) OR
3. use random local filename ([example](https://github.com/apache/hadoop-ozone/blob/912dcf0dcbcfb48952443e54f56b33695c413eee/hadoop-ozone/dist/src/main/smoketest/ozonefs/hadoopo3fs.robot#L29) for random string generation)

hadoop-ozone/dist/src/main/compose/testlib.sh
```diff
@@ -124,6 +124,23 @@ execute_command_in_container(){
   set +e
 }
 
+## @description Stop a list of named containers
+## @param       List of container names, eg datanode_1 datanode_2
+stop_containers() {
+  set -e
+  docker-compose -f "$COMPOSE_FILE" stop $@
```
adoroszlai:
```suggestion
  docker-compose -f "$COMPOSE_FILE" --no-ansi stop $@
```

Please add `--no-ansi` to avoid funky characters in the output:

```
2020-02-26T17:01:00.9125703Z [3BStarting datanode_1 ... 
2020-02-26T17:01:00.9127089Z Starting datanode_2 ... 
2020-02-26T17:01:00.9128127Z Starting datanode_3 ... 
2020-02-26T17:01:01.4869851Z [1A[2K
2020-02-26T17:01:01.4870604Z Starting datanode_3 ... [32mdone[0m
2020-02-26T17:01:01.5357903Z [1B[3A[2K
2020-02-26T17:01:01.5359166Z Starting datanode_1 ... [32mdone[0m
2020-02-26T17:01:01.6387370Z [3B[2A[2K
```
```

File: hadoop-ozone/dist/src/main/compose/testlib.sh [hunk 3]
```diff
@@ -124,6 +124,23 @@ execute_command_in_container(){
...
+  set +e
+}
+
+
+## @description Start a list of named containers
+## @param       List of container names, eg datanode_1 datanode_2
+start_containers() {
+  set -e
+  docker-compose -f "$COMPOSE_FILE" start $@
```
adoroszlai:
```suggestion
  docker-compose -f "$COMPOSE_FILE" --no-ansi start $@
```

hadoop-ozone/dist/src/main/smoketest/topology/readdata.robot
sodonnel: Well spotted. I will add a random filename using the example you gave.

hadoop-ozone/dist/src/main/compose/testlib.sh
sodonnel: Of course, I had noticed this in other commands but forgot to add the switch to my new ones. I will add it in.

sodonnel:
I have pushed a change to address the review comments, however this test is failing intermittently locally, so do not commit even if it comes back as all green. 

I think the failures are highlighting issues with the network topology implementation, rather than a problem with the test.

hadoop-ozone/dist/src/main/compose/ozone-topology/test.sh
```diff
@@ -33,6 +33,20 @@ execute_robot_test scm basic/basic.robot
...
+execute_robot_test scm topology/loaddata.robot
+
+stop_containers datanode_1 datanode_2 datanode_3
+
+execute_robot_test scm topology/readdata.robot
+
+start_containers datanode_1 datanode_2 datanode_3
+
+stop_containers datanode_4 datanode_5 datanode_6
+
```
elek: what about if datanode_1, datanode_2 and datnaode_3 are not yet started / registered but 4-6 are stopped. Do we need some safety check here?

In the `testlib.sh/start_docker_env` we have a `wait_for_safemode_exit` part. We might need it here.

sodonnel: Looking at some of the intermittent failures, I think we do need a safety check, but the safemode check will not do it. I believe we need to wait for the xceiver  port on the DNs to come up. I have a plan for this which I will post shortly.

However we will still need HDDS-3107 as that is another reason this test can fail.

sodonnel:
All checks have passed on this PR, however I still believe this test will fail intermittently due to HDDS-3116 and HDDS-3107. Therefore we should not merge this until those issues have been resolved.

adoroszlai:
/pending "I still believe this test will fail intermittently due to HDDS-3116 and HDDS-3107. Therefore we should not merge this until those issues have been resolved."

elek:
We can rename `test.sh` to something else and commit this patch. In this case we will have a test but won't be picked app by the acceptance tests.

What do you think?

sodonnel:
> We can rename `test.sh` to something else and commit this patch. In this case we will have a test but won't be picked app by the acceptance tests.
> 
> What do you think?

The test.sh here is not completely new, as there was some existing logic in it already. We could raise a new Jira and comment out the lines for the new tests in test.sh (with that Jira number in the comment) and then I could revisit it when we get the blockers fixed. If you are happy with that, let me know and I will push a commit to do that.

sodonnel:
@adoroszlai has approved this one, so I will go ahead and commit it, and we can get the new test enabled via HDDS-3135 when the blockers are resolved.

PR #610, opened 2020-02-27
HDDS-2929. Implement ofs://: temp directory mount

smengcl:

Support client-side `/tmp/` mount for convention / legacy application support. Each user will have their own temp bucket in this implementation.

We plan to have globally-shared temp bucket once sticky-bit is implemented in Ozone.

https://issues.apache.org/jira/browse/HDDS-2929

## How was this patch tested?

Added new test `testTempMount`.

## NOTES

1. An admin must create volume named `tmp` (hard-coded at the moment) with world access ACL before any user can use `/tmp/` correctly. i.e. Run the following for cluster initial setup:
```bash
ozone sh volume create tmp
ozone sh volume setacl tmp -al world::a
```
2. When a user is trying to read or write under `/tmp/`, a bucket with that user's name will be automatically created under volume `tmp` during the first operation:
```bash

bash-4.2$ HADOOP_USER_NAME=user2 ozone fs -ls ofs://om/tmp/
2020-02-20 12:02:08,931 [main] INFO rpc.RpcClient: Creating Bucket: tmp/user2

bash-4.2$ HADOOP_USER_NAME=user1 ozone fs -put README.md ofs://om/tmp/
2020-02-20 12:02:26,945 [main] INFO rpc.RpcClient: Creating Bucket: tmp/user1

bash-4.2$ HADOOP_USER_NAME=user1 ozone fs -ls ofs://om/tmp/
Found 1 items
-rw-rw-rw-   1 user1 user1       3813 2020-02-20 12:02 ofs://om/tmp/README.md

bash-4.2$ HADOOP_USER_NAME=user2 ozone fs -ls ofs://om/tmp/
```

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -42,7 +42,6 @@ import org.apache.hadoop.ozone.om.exceptions.OMException;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer.ACLIdentityType;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer.ACLType;
 import org.apache.hadoop.ozone.security.acl.OzoneAclConfig;
-import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.test.GenericTestUtils;
 import org.junit.After;
 import org.junit.Assert;
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -731,7 +731,7 @@ public class TestRootedOzoneFileSystem {
     Assert.assertNotNull(vol);
 
     // Begin test
-    String hashedUsername = OFSPath.getTempMountBucketName(null);
+    String hashedUsername = OFSPath.getTempMountBucketNameOfCurrentUser();
 
     // Expect failure since temp bucket for current user is not created yet
     try {
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -128,12 +128,13 @@ public class TestRootedOzoneFileSystem {
 
   @Test
   public void testOzoneFsServiceLoader() throws IOException {
-    OzoneConfiguration cfg = new OzoneConfiguration();
+    OzoneConfiguration confTestLoader = new OzoneConfiguration();
     // Note: FileSystem#loadFileSystems won't load OFS class due to META-INF
     //  hence this workaround.
-    cfg.set("fs.ofs.impl", "org.apache.hadoop.fs.ozone.RootedOzoneFileSystem");
-    Assert.assertEquals(
-        FileSystem.getFileSystemClass(OzoneConsts.OZONE_OFS_URI_SCHEME, cfg),
+    confTestLoader.set("fs.ofs.impl",
+        "org.apache.hadoop.fs.ozone.RootedOzoneFileSystem");
+    Assert.assertEquals(FileSystem.getFileSystemClass(
+        OzoneConsts.OZONE_OFS_URI_SCHEME, confTestLoader),
         RootedOzoneFileSystem.class);
   }
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -745,12 +745,9 @@ public class TestRootedOzoneFileSystem {
     // Write under /tmp/, OFS will create the temp bucket if not exist
     fs.mkdirs(new Path("/tmp/dir1"));
 
-//    FSDataOutputStream stream = fs.create(new Path("/tmp/dir1/file1"));
-//    stream.write(1);
-
-//    try (FSDataOutputStream stream = fs.create(new Path("/tmp/dir1/file1"))) {
-//      stream.write(1);
-//    }
+    try (FSDataOutputStream stream = ofs.create(new Path("/tmp/dir1/file1"))) {
+      stream.write(1);
+    }
 
     // Verify temp bucket creation
     OzoneBucket bucket = vol.getBucket(hashedUsername);
@@ -760,6 +757,12 @@ public class TestRootedOzoneFileSystem {
     Assert.assertEquals(1, fileStatuses.length);
     Assert.assertEquals(
         "/tmp/dir1", fileStatuses[0].getPath().toUri().getPath());
+    // Verify file1 creation
+    FileStatus[] fileStatusesInDir1 =
+        fs.listStatus(new Path("/tmp/dir1"));
+    Assert.assertEquals(1, fileStatusesInDir1.length);
+    Assert.assertEquals("/tmp/dir1/file1",
+        fileStatusesInDir1[0].getPath().toUri().getPath());
   }
 
 }
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -29,12 +29,20 @@ import org.apache.hadoop.fs.contract.ContractTestUtils;
 import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.ozone.MiniOzoneCluster;
+import org.apache.hadoop.ozone.OzoneAcl;
 import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.TestDataUtil;
 import org.apache.hadoop.ozone.client.ObjectStore;
 import org.apache.hadoop.ozone.client.OzoneBucket;
 import org.apache.hadoop.ozone.client.OzoneKeyDetails;
 import org.apache.hadoop.ozone.client.OzoneVolume;
+import org.apache.hadoop.ozone.client.VolumeArgs;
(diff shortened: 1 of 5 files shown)


hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -67,10 +75,9 @@ private void initOFSPath(String pathStr) {
       // TODO: Compare a keyword list instead for future expansion.
       if (firstToken.equals(OFS_MOUNT_NAME_TMP)) {
         mountName = firstToken;
-        // TODO: Retrieve volume and bucket of the mount from user protobuf.
-        //  Leave them hard-coded just for now. Will be addressed in HDDS-2929
-        volumeName = "tempVolume";
-        bucketName = "tempBucket";
+        // Future: retrieve volume and bucket from UserVolumeInfo
+        volumeName = OFS_MOUNT_TMP_VOLUMENAME;
+        bucketName = getTempMountBucketName(null);
```
timmylicheng: Is it better to have a 'DEFAULT_USER' string to replace a null input?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -67,10 +75,9 @@ private void initOFSPath(String pathStr) {
       // TODO: Compare a keyword list instead for future expansion.
       if (firstToken.equals(OFS_MOUNT_NAME_TMP)) {
         mountName = firstToken;
-        // TODO: Retrieve volume and bucket of the mount from user protobuf.
-        //  Leave them hard-coded just for now. Will be addressed in HDDS-2929
-        volumeName = "tempVolume";
-        bucketName = "tempBucket";
+        // Future: retrieve volume and bucket from UserVolumeInfo
```
timmylicheng: Do you mean to comment a TODO here?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -113,12 +125,12 @@ public void teardown() {
 
   @Test
   public void testOzoneFsServiceLoader() throws IOException {
-    OzoneConfiguration conf = new OzoneConfiguration();
+    OzoneConfiguration cfg = new OzoneConfiguration();
```
timmylicheng: NIT. 'conf' may be a more consistent choice of naming in Ozone repo.

smengcl: This is due to I put another `conf` in global. So IntelliJ is complaining this `conf` inside shadows the global `conf` in this function.

Anyway, will rename it.

smengcl: Done in 28cedc4db0820b84e8d2af151525d66ec2af944b

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: Thanks for the review and comment.

Yeah good point. `null` here doesn't have have good readability.

Added new function `getTempMountBucketNameOfCurrentUser()` in 5ed5cfa523cb6a0ff1d2c5179ef3f0f04d8c7ae9

Initially I thought of overriding it but ended up adding a new function with a different name for even better code readability.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: Done in 5ed5cfa523cb6a0ff1d2c5179ef3f0f04d8c7ae9

smengcl:
Note this might conflict with https://github.com/apache/hadoop-ozone/pull/606 so I'd so a rebase after that one get in.

smengcl:
Just rebased on to latest OFS branch.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -213,7 +213,9 @@ private OzoneBucket getBucket(String volumeStr, String bucketStr,
     try {
       bucket = proxy.getBucketDetails(volumeStr, bucketStr);
     } catch (OMException ex) {
-      if (createIfNotExist) {
+      // Note: always create bucket if volumeStr matches "tmp" so -put works
+      if (createIfNotExist ||
+          volumeStr.equals(OFSPath.OFS_MOUNT_TMP_VOLUMENAME)) {
```
xiaoyuyao: Do we create the tmp bucket even for read operation?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -44,7 +48,7 @@
    * /vol1/buc2/dir3/key4  vol1           buc2           (empty)      dir3/key4
    * /vol1/buc2            vol1           buc2           (empty)      (empty)
    * /vol1                 vol1           (empty)        (empty)      (empty)
-   * /tmp/dir3/key4        tempVolume     tempBucket     tmp          dir3/key4
+   * /tmp/dir3/key4        tmp            <username>     tmp          dir3/key4
```
xiaoyuyao: How do we allow userA to access userB's tmp bucket? This may be needed for the ETL workload where an injest process run as user A and the process process run as user B.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -174,4 +181,32 @@ public boolean isRoot() {
...
+  /**
+   * Get the bucket name of temp for the current user from UserGroupInformation.
+   * @return Username MD5 hash in hex digits.
+   */
+  @VisibleForTesting
+  static String getTempMountBucketNameOfCurrentUser() {
+    String username;
+    try {
+      username = UserGroupInformation.getCurrentUser().getUserName();
+    } catch (IOException ex) {
```
xiaoyuyao: why do we need to catch the exception here? This should not fail, if it does, the exception can bubble up directly.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: That's correct. This is my implementation decision.

I initially only create temp bucket for write, as you would expect. Then I found Hadoop FS file write operations calls `getFileStatus` beforehand.

smengcl: After some discussion, we decide to let user create their temp bucket manually upon first use. i.e. `ozone fs -mkdir ofs://om/tmp/`

smengcl: Addressed in 1d4164b153386c5e476f27df6420512b60894d5d

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: Done in 1d4164b153386c5e476f27df6420512b60894d5d

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: As of now userA can't access userB's temp with ofs:// . Because OFSPath parses `ofs://tmp/dir1/dir2` into `volume=tmp bucketname=<USER_TEMP_BUCKET_NAME>, key(in bucket)=dir1/dir2`

smengcl: I'm thinking of adding some `.reserved` like HDFS did to allow user to bypass the mount or add some options, if there is a need.

Anyway this may not fit in this jira.

xiaoyuyao: agree, can you add the proposal to the design doc in the umbrella JIRA?

xiaoyuyao: NIT: should we update the comment to match with the code?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -174,4 +185,27 @@ public boolean isRoot() {
...
+   * Get the bucket name of temp for given username.
+   * @param username Input user name String. Mustn't be null.
+   * @return Username MD5 hash in hex digits.
+   */
+  @VisibleForTesting
+  static String getTempMountBucketName(String username) {
+    Preconditions.checkNotNull(username);
+    // TODO: Improve this to "slugify(username)-md5(username)" for better
+    //  readability?
+    return DigestUtils.md5Hex(username);
```
xiaoyuyao: Can we use the name directly to make it more user friendly?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: Will resolve in a follow up jira.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: As discussed, since other users' `tmp` buckets aren't visible to current user via `ofs://`. This could be addressed when the requirement of accessing other users' temp bucket happen.
And there are several related future improvements. Like:
1. allowing users to access all volumes (that they have access to) via a special reserved path (e.g. `ofs://om/.volumes/<volume_name>/..` or `ofs://om/.reserved/..`). Also, admins should be able to list all volumes, not just the ones they own.
2. Provide a utility in Ozone Shell to access the users' temp buckets conveniently for admins.

PR #617, opened 2020-02-28
HDDS-3109. Refactor 'Recon' in MiniOzoneCluster to use ephemeral port.

avijayanhwx:

Currently, Recon uses an ephemeral port only in the integration test for Recon. In all other integration tests, we end up using the default (9888) that causes failures in other integration tests that start up a Mini ozone cluster. In addition, we want to start up Recon in MiniOzoneCluster by explicitly requesting it rather than by default.

https://issues.apache.org/jira/browse/HDDS-3109

## How was this patch tested?
Unit, integration and docker based sanity testing.


avijayanhwx:
@vivekratnavel / @swagle Please review.

vivekratnavel:
LGTM +1, pending CI.

PR #618, opened 2020-03-01
HDDS-2911. Fix lastUsed and stateEnterTime value in container info is not human friendly

captainzmc:

When we use:  ozone scmcli container list 
We can get lastUsed and stateEnterTime like this. It's not easy to know the exact time.
  "lastUsed" : 5388521335, 
  "stateEnterTime" : 808947405,
By this PR we use currentTimeMillis  instead of  nanoTime for easier viewing. Like This:
  "lastUsed" : 1583057429534,
  "stateEnterTime" : 1583057388710,

https://issues.apache.org/jira/browse/HDDS-2911

## How was this patch tested?
No need UT

Tested with `ozone admin container list`


captainzmc:
@ChenSammi Can you help to review this PR?

ChenSammi:
With this PR, will the time like this "2020-02-26T02:32:36.016Z"?  @captainzmc

captainzmc:
> With this PR, will the time like this "2020-02-26T02:32:36.016Z"? @captainzmc
Hi @ChenSammi 
Timestamp is more general in hadoop. It feels better to use Timestamp here. The type of this  was Long, but to change it to "2020-02-26T02:32:36.016Z" need to change the type to String.This may involve some other modifications. Are we sure we want to change this?

elek:
> In this case, can we display it as string in CLI ouput, and keep the long value internally.

+1 It seems to be to more flexible option. Keep the long value for protobuf (we can keep even the nanosec) but print out in a human readable form...

We can also change the Java type (and not the protobuf type) to a proper java8 Time object (like `Instant`). That is more meaningful and might be printed out properly by the current default JSON serializer...

captainzmc:
When output, the long type will be changed to date, and the format is as follows:
{"lastUsed":"2020-03-16 14:18:10.208","owner":"029b69de-91b5-4fa0-bee1-ef6e139b7273","replicationFactor":"ONE","replicationType":"RATIS","deleteTransactionId":0,"state":"OPEN","numberOfKeys":1,"usedBytes":800,"stateEnterTime":"2020-03-16 11:01:09.565","containerID":1,"sequenceId":0,"open":true}

ChenSammi:
@captainzmc  suggest add the time zone information.  You can take a look at how key info is displayed. 

ozone sh key list instagram/images
{
  "volumeName" : "instagram",
  "bucketName" : "images",
  "name" : "1a8ea5a7-952b-4ebc-8fa5-5017e90bce5f-100000083",
  "dataSize" : 268435456,
  "creationTime" : "2020-03-10T04:51:55.133Z",
  "modificationTime" : "2020-03-10T04:51:55.149Z",
  "replicationType" : "RATIS",
  "replicationFactor" : 3
}

captainzmc:
hi @ChenSammi The time format has been modified and shown as follows:
"lastUsed":"2020-03-18T13:33:39.915Z",
"stateEnterTime":"2020-03-18T12:56:26.015Z"

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerInfo.java
```diff
@@ -98,7 +101,7 @@
     this.pipelineID = pipelineID;
     this.usedBytes = usedBytes;
     this.numberOfKeys = numberOfKeys;
-    this.lastUsed = Time.monotonicNow();
```
xiaoyuyao: If you change the lastUsed and stateEnterTime to lnstant type, the code will be much simpler and clean as @elek  suggested.

captainzmc:
As @xiaoyuyao and @elek suggested. Have changed lastUsed and stateEnterTime to lnstant type. 
And output with the previous ObjectWriter. Now the format is:
{
  "state" : "OPEN",
  "replicationFactor" : "ONE",
  "replicationType" : "RATIS",
  "usedBytes" : 441,
  "numberOfKeys" : 1,
  "lastUsed" : "2020-03-29T11:47:50.347Z",
  "stateEnterTime" : "2020-03-29T10:38:26.681Z",
  "owner" : "20a01508-2173-4e80-923c-5b2b6d3353dd",
  "containerID" : 1,
  "deleteTransactionId" : 0,
  "sequenceId" : 0,
  "open" : true
}

PR #621, opened 2020-03-02
HDDS-3106. Intermittent timeout in TestOzoneManagerDoubleBufferWithOMResponse#testDoubleBuffer

adoroszlai:

Increase max. concurrency to 2x the number of volumes to be created, because `OMVolumeCreateRequest` requires 2 write locks: volume- and user-level.

https://issues.apache.org/jira/browse/HDDS-3106

## How was this patch tested?

Regular CI with all checks:

https://github.com/adoroszlai/hadoop-ozone/runs/479812239

Ran `TestOzoneManagerDoubleBufferWithOMResponse` 50 times:

https://github.com/adoroszlai/hadoop-ozone/runs/479814646

### DIFF
--- a/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/ratis/TestOzoneManagerDoubleBufferWithOMResponse.java
+++ b/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/ratis/TestOzoneManagerDoubleBufferWithOMResponse.java
@@ -71,6 +71,8 @@ import static org.mockito.Mockito.when;
  */
 public class TestOzoneManagerDoubleBufferWithOMResponse {
 
+  private static final int MAX_VOLUMES = 1000;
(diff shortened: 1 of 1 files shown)


adoroszlai:
@bharatviswa504 I think there may be a real possibility of deadlock here, this patch just fixes the intermittent test.

bharatviswa504:
> @bharatviswa504 I think there may be a real possibility of deadlock here, this patch just fixes the intermittent test.

Thank You @adoroszlai. Deadlock in OM? Not sure what is referred here.

bharatviswa504:
> Increase max. concurrency to 2x the number of volumes to be created, because OMVolumeCreateRequest requires 2 write locks: volume- and user-level.
But one question even if we have max. concurrency set to 1000, the thread will be blocked and when other thread releases the lock, it will acquire the lock. Not sure if I am missing something here.

adoroszlai:

PR #624, opened 2020-03-02
HDDS-3110. Fix race condition in Recon's container and pipeline handling.

avijayanhwx:

- Both the Incremental container report handler and the regular container report handler add new containers from SCM whenever they see a new container. This test and add step must be synchronized between the 2 handlers to avoid any inconsistent metadata state.
- NodeStateMap in allow does not addition of a single container to the Map of Node -> Set of Containers since it instantiates with a Collections.emptySet(), and then relies on a map.put() to update the value. Changing this to a "new HashSet" allows addition of a container one by one which is possible in Recon.
- Improve logging in Recon Container Manager when it receives a container report from a node before receiving the pipeline report for a newly created pipeline.

https://issues.apache.org/jira/browse/HDDS-3110

## How was this patch tested?
Unit, integration and docker based testing.

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerReportHandler.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerReportHandler.java
@@ -63,19 +63,21 @@ public class ReconContainerReportHandler extends ContainerReportHandler {
     for (ContainerReplicaProto containerReplicaProto : reportsList) {
       final ContainerID id = ContainerID.valueof(
           containerReplicaProto.getContainerID());
-      if (!getContainerManager().exists(id)) {
-        LOG.info("New container {} got from {}.", id,
-            reportFromDatanode.getDatanodeDetails());
-        try {
-          ContainerWithPipeline containerWithPipeline =
-              scmClient.getContainerWithPipeline(id.getId());
-          LOG.debug("Verified new container from SCM {} ",
-              containerWithPipeline.getContainerInfo().containerID());
-          containerManager.addNewContainer(id.getId(), containerWithPipeline);
-        } catch (IOException ioEx) {
-          LOG.error("Exception while getting new container info from SCM",
-              ioEx);
-          return;
+      synchronized (containerManager) {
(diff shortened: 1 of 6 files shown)


hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconIncrementalContainerReportHandler.java
```diff
@@ -70,19 +70,22 @@ public void onMessage(final IncrementalContainerReportFromDatanode report,
...
+                  scmClient.getContainerWithPipeline(id.getId());
+              LOG.info("Verified new container from SCM {} ",
+                  containerWithPipeline.getContainerInfo().containerID());
+              containerManager
+                  .addNewContainer(id.getId(), containerWithPipeline);
+            } catch (IOException ioEx) {
+              LOG.error("Exception while getting new container info from SCM",
+                  ioEx);
+              return;
+            }
```
adoroszlai: Looks like this block could be shared between the two container report handlers.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerReportHandler.java
```diff
@@ -63,19 +63,21 @@ public void onMessage(final ContainerReportFromDatanode reportFromDatanode,
...
-          LOG.error("Exception while getting new container info from SCM",
-              ioEx);
-          return;
+      synchronized (containerManager) {
+        if (!containerManager.exists(id)) {
+          LOG.info("New container {} got from {}.", id,
+              reportFromDatanode.getDatanodeDetails());
+          try {
+            ContainerWithPipeline containerWithPipeline =
+                scmClient.getContainerWithPipeline(id.getId());
```
adoroszlai: Might this be improved in a followup to avoid RPC while holding the lock?

adoroszlai:
Thanks @avijayanhwx for the fix and @nandakumar131 for reviewing it.

avijayanhwx:
Thanks for the reviews @nandakumar131 and @adoroszlai. I will address the pending review comments in a subsequent JIRA.

PR #625, opened 2020-03-02
HDDS-2980. Delete replayed entry from OpenKeyTable during commit

hanishakoneru:

During KeyCreate (and S3InitiateMultipartUpload), we do not check the OpenKeyTable if the key already exists. If it does exist and the transaction is replayed, we just override the key in OpenKeyTable. This is done to avoid extra DB reads.

During KeyCommit (or S3MultipartUploadCommit), if the key was already committed, then we do not replay the transaction. This would result in the OpenKeyTable entry to remain in the DB till it is garbage collected. 

To avoid storing stale entries in OpenKeyTable, during commit replays, we should check the openKeyTable and delete the entry if it exists.

https://issues.apache.org/jira/browse/HDDS-2980

## How was this patch tested?

Added unit test.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
@@ -198,7 +198,7 @@ public class OMKeyCommitRequest extends OMKeyRequest {
           new CacheValue<>(Optional.of(omKeyInfo), trxnLogIndex));
 
       omClientResponse = new OMKeyCommitResponse(omResponse.build(),
-          omKeyInfo, dbOpenKey);
+          omKeyInfo, dbOzoneKey, dbOpenKey);
 
       result = Result.SUCCESS;
     } catch (IOException ex) {
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
@@ -68,6 +68,7 @@ public class OMKeyCommitRequest extends OMKeyRequest {
   private enum Result {
     SUCCESS,
     REPLAY,
+    DELETE_OPEN_KEY_ONLY,
     FAILURE
   }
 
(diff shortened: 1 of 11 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -183,21 +196,26 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           new CacheKey<>(dbOzoneKey),
           new CacheValue<>(Optional.of(omKeyInfo), trxnLogIndex));
 
-      omResponse.setCommitKeyResponse(CommitKeyResponse.newBuilder().build());
       omClientResponse = new OMKeyCommitResponse(omResponse.build(),
-          omKeyInfo, commitKeyRequest.getClientID());
+          omKeyInfo, dbOpenKey);
```
bharatviswa504: We can still pass here  clientID and generate openKey. Like how we used to generate open key before.

hanishakoneru: We are anyway computing the openKey using the clientID here. So instead of passing the clientID and computing that again in OMKeyCommitResponse, we can directly pass the openKey.

bharatviswa504: I see that we are doing similar thing for ozone key, so that is why my question. Then we can do similar thing for ozone key also then?

hanishakoneru: We could. But since we have to pass the omKeyInfo irrespective of whether we pass ozoneKey or not, I didnt change that. 
I am ok to have it either way. Will post an update.

hanishakoneru:
Integration test failure (timeout) is unrelated and passes locally. Re-triggering the checks.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCommitPartRequest.java
```diff
@@ -147,11 +146,6 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
         throw new OMException("Failed to commit Multipart Upload key, as " +
             openKey + "entry is not found in the openKey table",
             KEY_NOT_FOUND);
-      } else {
-        // Check the OpenKeyTable if this transaction is a replay of ratis logs.
```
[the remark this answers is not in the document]
hanishakoneru: This check was redundant. Irrespective of whether KeyCreate Request was replayed or not, if key+clientID exits in the openKey table, then the CommitPart request should also be executed (same as we do for KeyCommit Request). 
If the same Key part was created again, the clientID would be different. Hence the openKey would also be different.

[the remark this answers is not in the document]
bharatviswa504: Got it. thanks.

hanishakoneru:
Thanks for the review @bharatviswa504 .

PR #627, opened 2020-03-03
HDDS-3118. Possible deadlock in LockManager.

bharatviswa504:

Possible deadlock in lock manager.
This is because of lockPool.borrowObject waits indefinitely and activeLocks.compute is an atomic operation that holds a map lock.

https://issues.apache.org/jira/browse/HDDS-3118

## How was this patch tested?

The reproed patch is now passing.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
@@ -19,7 +19,6 @@ package org.apache.hadoop.ozone.lock;
 
 import org.apache.commons.pool2.impl.GenericObjectPool;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.hdds.HddsConfigKeys;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -59,6 +58,7 @@ public class LockManager<R> {
   public LockManager(final Configuration conf, boolean fair) {
     lockPool =
         new GenericObjectPool<>(new PooledLockFactory(fair));
+    lockPool.setMaxTotal(-1);
   }
 
   /**
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
@@ -186,21 +186,42 @@ public class LockManager<R> {
      * be removed from the activeLocks map and returned to
      * the object pool.
      */
-    return activeLocks.compute(resource, (k, v) -> {
+    ActiveLock computedLock = activeLocks.compute(resource, (k, v) -> {
       final ActiveLock lock;
+      if (v == null) {
+        return null;
+      } else {
+        lock = v;
+      }
+      lock.incrementActiveCount();
+      return lock;
+    });
+
+
+    if (computedLock == null) {
       try {
-        if (v == null) {
-          lock = lockPool.borrowObject();
-        } else {
-          lock = v;
-        }
-        lock.incrementActiveCount();
+        final ActiveLock lock2 = lockPool.borrowObject();
+        computedLock = activeLocks.compute(resource, (k, v) -> {
+          final ActiveLock lock;
+          try {
+            if (v == null) {
+              lock = lock2;
+            } else {
+              lockPool.returnObject(lock2);
+              lock = v;
+            }
+            lock.incrementActiveCount();
+          } catch (Exception ex) {
+            throw new RuntimeException(ex);
+          }
+          return lock;
+        });
       } catch (Exception ex) {
         LOG.error("Unable to obtain lock.", ex);
         throw new RuntimeException(ex);
       }
-      return lock;
-    });
+    }
+    return computedLock;
   }
 
   /**
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
@@ -183,19 +183,19 @@ public class LockManager<R> {
      * the object pool.
      */
     return activeLocks.compute(resource, (k, v) -> {
+      final ActiveLock lock;
       try {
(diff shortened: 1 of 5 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
```diff
@@ -186,21 +186,42 @@ private ActiveLock getLockForLocking(final R resource) {
...
+          try {
+            if (v == null) {
+              lock = lock2;
+            } else {
+              lockPool.returnObject(lock2);
+              lock = v;
+            }
+            lock.incrementActiveCount();
+          } catch (Exception ex) {
+            throw new RuntimeException(ex);
```
adoroszlai: Nit: Is this necessary?  I don't see any checked exception being thrown in the `try` block, and unchecked exceptions don't need to be wrapped.

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/lock/TestLockManager.java
```diff
@@ -170,4 +175,32 @@ public void testMultiReadWriteLockWithSameResource() throws Exception {
     Assert.assertTrue(gotLock.get());
   }
 
+  @Test
+  public void testConcurrentWriteLockWithDifferentResource() throws Exception {
+    OzoneConfiguration conf = new OzoneConfiguration();
+    final int count = 100;
+    conf.setInt(HDDS_LOCK_MAX_CONCURRENCY, count/4);
```
xiaoyuyao: Thanks @bharatviswa504 for the patch. The PR LGTM. I understand we are trying to avoid wait forever while holding the lock of the pool map. However, it does not completely solve the lock resource exhausting problem. I have a cluster with this patch but create key still can live lock by the limited number of locks (100). 

My question is why we are restricting # of locks here via hdds.lock.max.concurrency.  In my opinion, if we change the default from 100 to -1, this deadlock will not happen. This also match the default max pool size (-1) of GenericObjectPool, especially lock is a cheaper resource compared with threads or tcp connections. To allow handling 4K+ request per second, we should consider removing this key that can easily cause deadlock, especially when handler counter is several times larger than the lock pool size.

bharatviswa504: Yes, if we set max pool size to -1, the max total objects created by GenericcObjectPool used in Integer.MAX. I agree we can change to -1, and we shall not see this issue. And we can remove this property, and set it to -1 so that we shall not see this kind of issue when someone mistakenly sets this to a low value and operates on the cluster with very large number of volumes/buckets in parallel. With this way, we shall not see lock contention during acquiring lock.


But few questions, right now we take a lock on resource name, so if there is a lock Object which is already in lockPool Object and we put that in the lock map, and we reuse that object. So,  when we operate on more than 100 buckets simultaneously we should see this problem when creating keys or when more than 100 volumes operations are going in parallel we can see some contention in acquiring locks. So, it is not exactly related to handler count in OM. Just trying to understand is the issue is observed on the test cluster?

bharatviswa504: Thank You @xiaoyuyao for the review. Removed the config in the latest commit.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
bharatviswa504: Based on Xiaoyu's comment reverted this change.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
```diff
@@ -57,12 +56,9 @@ public LockManager(final Configuration conf) {
    * @param fair - true to use fair lock ordering, else non-fair lock ordering.
    */
   public LockManager(final Configuration conf, boolean fair) {
-    final int maxPoolSize = conf.getInt(
-        HddsConfigKeys.HDDS_LOCK_MAX_CONCURRENCY,
-        HddsConfigKeys.HDDS_LOCK_MAX_CONCURRENCY_DEFAULT);
     lockPool =
         new GenericObjectPool<>(new PooledLockFactory(fair));
-    lockPool.setMaxTotal(maxPoolSize);
+    lockPool.setMaxTotal(-1);
```
xiaoyuyao: NIT: we don't need to explicitly set -1 as the default maxTotal from GenericObjectPool is -1.

bharatviswa504: yes, I did this expicility, if in future released they changed the defaults it might break.

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/lock/TestLockManager.java
xiaoyuyao: bq. Just trying to understand is the issue is observed on the test cluster, where some test is working in parallel across more than 100 volumes/100 buckets and a lock contention is observed?

I hit the problem last week and had been reviewing related code along with the OM stack dumps this week. Found the fix here yesterday, tried it and found that here may not completely solve the problem. Here is the root cause:

OMKeyCreateRequest.validateAndUpdateCache requires two locks from the lock pool to finish a key creation. One write lock for the BUCKET at the beginning and one lock for longest prefix path close to the end during prepareKeyInfo call. 

If you have 200 threads handlers but only 100 lock limit from the lock manager. When the first 100 key creation threads all grab the bucket lock, OM will stuck there forever as we don't have any locks from the pool for the second call. My initial thought is to add a timeout but later found removing the limit seems to be a better solution.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lock/LockManager.java
xiaoyuyao: Sounds good. +1 pending CI.

adoroszlai:
Thanks @bharatviswa504 for the fix and @xiaoyuyao for the review.

PR #628, opened 2020-03-03
HDDS-1008. Invalidate closed container replicas on a failed volume.

lokeshj1703:

The hdds volume check will now check for the existence of its root directory. Further the containers pertaining to the volume will be marked unhealthy.

https://issues.apache.org/jira/browse/HDDS-1008

## How was this patch tested?

Unit test

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ContainerSet.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ContainerSet.java
@@ -122,6 +122,18 @@ public class ContainerSet {
     return containerMap.size();
   }
 
+  public void handleVolumeFailures(Set<HddsVolume> failedVolumes) {
+    containerMap.values().forEach(c -> {
+      if (failedVolumes.contains(c.getContainerData().getVolume())) {
+        try {
+          c.markContainerUnhealthy();
+        } catch (StorageContainerException e) {
+          LOG.error("Failed to move container to UNHEALTHY state", e);
+        }
+      }
+    });
+  }
+
   /**
    * Return an container Iterator over {@link ContainerSet#containerMap}.
    * @return {@literal Iterator<Container<?>>}
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ContainerSet.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ContainerSet.java
@@ -122,9 +122,9 @@ public class ContainerSet {
     return containerMap.size();
   }
 
-  public void handleVolumeFailures(Set<HddsVolume> failedVolumes) {
+  public void handleVolumeFailures() {
     containerMap.values().forEach(c -> {
-      if (failedVolumes.contains(c.getContainerData().getVolume())) {
+      if (c.getContainerData().getVolume().isFailed()) {
         try {
           c.markContainerUnhealthy();
         } catch (StorageContainerException e) {
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ContainerSet.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ContainerSet.java
@@ -128,7 +128,9 @@ public class ContainerSet {
         try {
           c.markContainerUnhealthy();
         } catch (StorageContainerException e) {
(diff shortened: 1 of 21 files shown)


lokeshj1703:
@supratimdeka Thanks for reviewing the PR! I have added a commit which addresses your comments.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeSet.java
```diff
@@ -239,23 +249,24 @@ private void checkAllVolumes() throws IOException {
...
-        this.writeUnlock();
+        failVolume(v.getHddsRootDir().getPath());
       }
-
-      // TODO:
-      // 1. Mark all closed containers on the volume as unhealthy.
-      // 2. Consider stopping IO on open containers and tearing down
-      //    active pipelines.
-      // 3. Handle Ratis log disk failure.
+      if (failedVolumeListener != null) {
```
supratimdeka: This can probably be done outside the writeLock - should be safe.
also, we can invoke the listener outside the for loop - perhaps like:
if (!failedVolumes.isEmpty()) {failedVolumeListener.run();}

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
```diff
@@ -298,6 +298,8 @@ public void markContainerForClose() throws StorageContainerException {
   public void markContainerUnhealthy() throws StorageContainerException {
     writeLock();
     try {
+      // TODO: container should be marked UNHEALTHY in memory even if container
```
supratimdeka: Good point. But, I think this is how it is already implemented. updateContainerData() on encountering an exception, does not reset the state if the state is UNHEALTHY. Or did I get that wrong?

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/volume/TestVolumeSet.java
```diff
@@ -64,7 +64,7 @@
   private static final String DUMMY_IP_ADDR = "0.0.0.0";
 
   private void initializeVolumeSet() throws Exception {
-    volumeSet = new VolumeSet(UUID.randomUUID().toString(), conf);
+    volumeSet = new VolumeSet(UUID.randomUUID().toString(), null, conf);
```
supratimdeka: any reason why we are migrating to this constructor with the additional argument?
same question applies to all the other places where this change has been made.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ContainerSet.java
```diff
@@ -122,6 +122,18 @@ public int containerCount() {
     return containerMap.size();
   }
 
+  public void handleVolumeFailures() {
+    containerMap.values().forEach(c -> {
+      if (c.getContainerData().getVolume().isFailed()) {
+        try {
+          c.markContainerUnhealthy();
+        } catch (StorageContainerException e) {
+          LOG.error("Failed to move container to UNHEALTHY state", e);
```
supratimdeka: might make sense to add the container id and the volume in this error log.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
lokeshj1703: You are right. I missed the state check.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeSet.java
lokeshj1703: Yes we can move it outside. But I think there is a race condition in marking a volume as failed. A container in process of creation can still get created on the failed volume. We can handle it as part of separate jira though.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/volume/TestVolumeSet.java
lokeshj1703: This is change related to old commit. Will fix it.

lokeshj1703:
@supratimdeka Thanks for the review! I have addressed your review comments in latest commit.

PR #631, opened 2020-03-03
HDDS-3119. When ratis is enabled in OM, double Buffer metrics not get…

avijayanhwx:
…ting updated.

Ozone Manager creates 2 instances of OzoneManagerDoubleMetrics, and the way Hadoop metrics2 works, the second one overwrites the first. In our case, it is the first instance that is on the Ratis write path. Fixed this issue by making sure there is only instance of OzoneManagerDoubleMetrics.

https://issues.apache.org/jira/browse/HDDS-3119

## How was this patch tested?
Manually tested.


bharatviswa504:
Can we avoid double instantiation of doublebuffer which is not required in OM, that will resolve this issue.

avijayanhwx:
Unit test failure TestSCMPipelineBytesWrittenMetrics is not related to this patch.

bharatviswa504:
There is already a Jira I have opened for the same test.

adoroszlai:
> There is already a Jira I have opened for the same test.

Me too. :)

PR #632, opened 2020-03-04
HDDS-3124. Fix time interval calculate error

runzhiwang:
Just as the image shows, the time interval in log "Unable to communicate to SCM server at scm-0.scm:9861 for past " is 0, 3000 seconds, but actually it is 0, 300 seconds.

image

https://issues.apache.org/jira/projects/HDDS/issues/HDDS-3124

## How was this patch tested?

No need to test.


runzhiwang:
@ChenSammi Could you help review this patch ? Thank you very much.

PR #634, opened 2020-03-05
HDDS-3129. Skip KeyTable check in OMKeyCommit.

bharatviswa504:

Perform the additional KeyTable check only when ratis is enabled. This check is only required for ratis scenario.

https://issues.apache.org/jira/browse/HDDS-3129

## How was this patch tested?

Deployed this on a cluster and checked the perf.
Tested on mac, with 50 client threads

2295 Keys/sec with Additional Key Table check
2824 Keys/sec with removing that check


arp7:
+1 pending CI. Thanks for investigating and fixing this @bharatviswa504.

bharatviswa504:

PR #645, opened 2020-03-06
HDDS-3130. Add jaeger trace span in s3gateway

runzhiwang:

Add jaeger trace span in s3gateway.

https://jira.apache.org/jira/browse/HDDS-3130

## How was this patch tested?

The jaeger UI shows the right trace information.
image

### DIFF
--- a/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/Gateway.java
+++ b/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/Gateway.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import org.apache.hadoop.hdds.cli.GenericCli;
 import org.apache.hadoop.hdds.cli.HddsVersionProvider;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
+import org.apache.hadoop.hdds.tracing.TracingUtil;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


runzhiwang:

elek:
Thank you very much the idea and the patch @runzhiwang I 100% support to introduce tracing for s3g. It's a very good idea.

About the implementation: It's definitely a step forward but I started to think how is it possible to do in a transparent way without duplication. I check if the resource name is available from some generic filter and it seems to be possible.

My proposal is to do it from a central place, something like this:

```
@Provider
public class TracingFilter implements ContainerRequestFilter,
    ContainerResponseFilter {

  public static final String TRACING_SCOPE = "TRACING_SCOPE";

  @Context
  private ResourceInfo resourceInfo;

  @Override
  public void filter(ContainerRequestContext requestContext)
      throws IOException {
    Scope scope = GlobalTracer.get().buildSpan(
        resourceInfo.getResourceClass().getSimpleName() + "." + resourceInfo
            .getResourceMethod().getName()).startActive(true);

    requestContext.setProperty(TRACING_SCOPE, scope);
  }

  @Override
  public void filter(ContainerRequestContext requestContext,
      ContainerResponseContext responseContext) throws IOException {

    Object scope = requestContext.getProperty(TRACING_SCOPE);
    if (scope != null) {
      ((Scope) scope).span().finish();
    }
  }
}
``` 

What do you think? I think it would simplify this patch (you don't need to modify all the methods...)

runzhiwang:
@elek Very reasonable suggestion. I will try it, thank you very much.

runzhiwang:
@elek Hi, I have updated PR according to you suggestion, thank you very much. Additionally, if throw exception when execute request, the method `public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext)` will not be executed, and span can not be finished, then the following jaeger span will have a error parent, just as the image shows. So I finish the old span before build a new span. Besides, `scope.close` also can not be ignored.
image

runzhiwang:
The CI failure has nothing to do with this PR.

runzhiwang:
@elek CI passed.

PR #649, opened 2020-03-06
HDDS-3120. Freon work with OM HA.

bharatviswa504:

Make Freon work with OM HA.
And also refactor OzoneClientFactory.

https://issues.apache.org/jira/browse/HDDS-3120

## How was this patch tested?

Tested it on ozone-om-ha-s3 docker-compose cluster.

Freon tests are added to test with OM HA.

```
bash-4.2$ ozone freon rk --om-service-id=id1 --numOfVolumes=1 --numOfBuckets=1 --numOfKeys=1 --numOfThreads=1
2020-03-06 20:46:21,267 [main] INFO impl.MetricsConfig: Loaded properties from hadoop-metrics2.properties
2020-03-06 20:46:21,529 [main] INFO impl.MetricsSystemImpl: Scheduled Metric snapshot period at 10 second(s).
2020-03-06 20:46:21,529 [main] INFO impl.MetricsSystemImpl: ozone-freon metrics system started
2020-03-06 20:46:23,391 [main] INFO freon.RandomKeyGenerator: Number of Threads: 1
2020-03-06 20:46:23,392 [main] INFO freon.RandomKeyGenerator: Number of Volumes: 1.
2020-03-06 20:46:23,392 [main] INFO freon.RandomKeyGenerator: Number of Buckets per Volume: 1.
2020-03-06 20:46:23,393 [main] INFO freon.RandomKeyGenerator: Number of Keys per Bucket: 1.
2020-03-06 20:46:23,393 [main] INFO freon.RandomKeyGenerator: Key size: 10240 bytes
2020-03-06 20:46:23,393 [main] INFO freon.RandomKeyGenerator: Buffer size: 4096 bytes
2020-03-06 20:46:23,394 [main] INFO freon.RandomKeyGenerator: validateWrites : false
2020-03-06 20:46:23,398 [main] INFO freon.RandomKeyGenerator: Starting progress bar Thread.

 0.00% |?                                                                                                    |  0/1 Time: 0:00:002020-03-06 20:46:23,466 [pool-2-thread-1] INFO rpc.RpcClient: Creating Volume: vol-0-45597, with hadoop as owner.
2020-03-06 20:46:23,533 [pool-2-thread-1] INFO rpc.RpcClient: Creating Bucket: vol-0-45597/bucket-0-56744, with Versioning false and Storage Type set to DISK and Encryption set to false 
2020-03-06 20:46:24,103 [pool-2-thread-1] WARN impl.MetricsSystemImpl: ozone-freon metrics system already initialized!
 100.00% |?????????????????????????????????????????????????????????????????????????????????????????????????????|  1/1 Time: 0:00:01

***************************************************
Status: Success
Git Base Revision: e97acb3bd8f3befd27418996fa5d4b50bf2e17bf
Number of Volumes created: 1
Number of Buckets created: 1
Number of Keys added: 1
Ratis replication factor: ONE
Ratis replication type: STAND_ALONE
Average Time spent in volume creation: 00:00:00,104
Average Time spent in bucket creation: 00:00:00,026
Average Time spent in key creation: 00:00:00,079
Average Time spent in key write: 00:00:00,735
Total bytes written: 10240
Total Execution time: 00:00:06,431
***************************************************
```

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
@@ -372,17 +372,9 @@ public class BaseFreonGenerator {
   public void ensureVolumeAndBucketExist(OzoneClient rpcClient,
       String volumeName, String bucketName) throws IOException {
 
-    OzoneVolume volume = null;
-    try {
-      volume = rpcClient.getObjectStore().getVolume(volumeName);
-    } catch (OMException ex) {
-      if (ex.getResult() == ResultCodes.VOLUME_NOT_FOUND) {
-        rpcClient.getObjectStore().createVolume(volumeName);
-        volume = rpcClient.getObjectStore().getVolume(volumeName);
-      } else {
-        throw ex;
-      }
-    }
+    OzoneVolume volume;
+    ensureVolumeExists(rpcClient, volumeName);
+    volume = rpcClient.getObjectStore().getVolume(volumeName);
 
     try {
       volume.getBucket(bucketName);
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
@@ -399,6 +399,8 @@ public class BaseFreonGenerator {
     } catch (OMException ex) {
       if (ex.getResult() == ResultCodes.VOLUME_NOT_FOUND) {
         rpcClient.getObjectStore().createVolume(volumeName);
+      } else {
+        throw ex;
       }
     }
   }
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
@@ -400,19 +400,14 @@ public class BaseFreonGenerator {
    * Create missing target volume.
    */
   public void ensureVolumeExists(
-      OzoneConfiguration ozoneConfiguration,
+      OzoneClient rpcClient,
       String volumeName) throws IOException {
-    try (OzoneClient rpcClient = OzoneClientFactory
-        .getRpcClient(ozoneConfiguration)) {
-
-      try {
-        rpcClient.getObjectStore().getVolume(volumeName);
-      } catch (OMException ex) {
-        if (ex.getResult() == ResultCodes.VOLUME_NOT_FOUND) {
-          rpcClient.getObjectStore().createVolume(volumeName);
-        }
+    try {
+      rpcClient.getObjectStore().getVolume(volumeName);
+    } catch (OMException ex) {
+      if (ex.getResult() == ResultCodes.VOLUME_NOT_FOUND) {
+        rpcClient.getObjectStore().createVolume(volumeName);
       }
-
     }
   }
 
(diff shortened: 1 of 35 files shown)


elek:
/pending "WIP to add tests will add tests in next commit."

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
```diff
@@ -469,4 +456,13 @@ public AtomicLong getAttemptCounter() {
   public int getThreadNo() {
     return threadNo;
   }
+
+  protected OzoneClient createOzoneClient(String omServiceID,
+      OzoneConfiguration conf) throws Exception {
+    if (omServiceID != null) {
+      return OzoneClientFactory.getRpcClient(omServiceID, conf);
+    } else {
+      return OzoneClientFactory.getRpcClient(conf);
```
adoroszlai: I think here `getClient` should be used to validate presence of `--om-service-id` param.

```suggestion
      return OzoneClientFactory.getClient(conf);
```

There are two methods in `OzoneClientFactory` that take `Configuration` and return `OzoneClient`:

 * `getClient(Configuration)` throws exception if configuration has OM service ID, but
 * `getRpcClient(Configuration)` does not

I'm getting timeout on OM HA cluster when `--om-service-id` is omitted.

```
$ ozone freon ockg -n 1 -t 1
...
[main] INFO ipc.Client: Retrying connect to server: 0.0.0.0/0.0.0.0:9862. Already tried 0 time(s); retry policy is RetryUpToMaximumCountWithFixedSleep(maxRetries=10, sleepTime=1000 MILLISECONDS)
...
[main] ERROR ha.OMFailoverProxyProvider: Failed to connect to OMs: [nodeId=null,nodeAddress=0.0.0.0:9862]. Attempted 15 failovers.
```

After making the suggested change locally, I get the error message as expected:

```
$ ozone freon ockg -n 1 -t 1
...
Following ServiceID's [id1] are defined in the configuration. Use the method getClient which takes serviceID and configuration as param
```

bharatviswa504:
Addressed review comments.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
bharatviswa504: Removed getClient, and used getRpcClient, to avoid these kind of mistakes instead of duplicating the same code in multiple functions.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/OzoneClientKeyGenerator.java
```diff
@@ -86,22 +91,24 @@ public Void call() throws Exception {
...
 
-    try (OzoneClient rpcClient = OzoneClientFactory
-        .getRpcClient(ozoneConfiguration)) {
-
-      bucket =
-          rpcClient.getObjectStore().getVolume(volumeName)
-              .getBucket(bucketName);
+    OzoneClient rpcClient = null;
+    try {
+      rpcClient = createOzoneClient(omServiceID, ozoneConfiguration);
```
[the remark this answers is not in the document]
bharatviswa504: Initially I was using if,else to create OzoneClient, later moved the code to a method. Used try with resource.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
```diff
@@ -374,53 +369,37 @@ public String generateObjectName(long counter) {
...
-        rpcClient.getObjectStore().getVolume(volumeName);
-      } catch (OMException ex) {
-        if (ex.getResult() == ResultCodes.VOLUME_NOT_FOUND) {
-          rpcClient.getObjectStore().createVolume(volumeName);
-        }
+    try {
+      rpcClient.getObjectStore().getVolume(volumeName);
+    } catch (OMException ex) {
+      if (ex.getResult() == ResultCodes.VOLUME_NOT_FOUND) {
+        rpcClient.getObjectStore().createVolume(volumeName);
```
adoroszlai: Should `throw ex` in `else` branch, otherwise volume creation fails silently and will run into NPE elsewhere.

bharatviswa504: Nice catch. But that should fail in the next steps in our case like during getVolume/CreateBucket. But it is better to throw an exception from here. Addressed in the latest commit.

bharatviswa504:

PR #650, opened 2020-03-07
HDDS-3089. TestSCMNodeManager intermittent crash

adoroszlai:

Based on the logs the test crashes during teardown when trying to delete pipeline in `EventQueue-DeadNodeForDeadNodeHandler` thread, as the RocksDB for pipelines was already closed during SCM stop.

1. Clear `db` reference in `RocksDBStore` after closing it, to avoid calls on closed RocksDB.  This might result in NPE, which would be handled in Ozone normally, while call on closed DB might result in seg.fault in native code.
2. Protect `pipelineStore.close()` call with write lock in `SCMPipelineManager` to avoid concurrent `pipelineStore.delete()` (or other similar operations).

https://issues.apache.org/jira/browse/HDDS-3089

## How was this patch tested?

Executed TestSCMNodeManager 2x20 times successfully:
https://github.com/adoroszlai/hadoop-ozone/runs/491868681
https://github.com/adoroszlai/hadoop-ozone/runs/491888786

Regular CI with all checks (only it-freon failed):
https://github.com/adoroszlai/hadoop-ozone/runs/491950482


adoroszlai:

PR #653, opened 2020-03-09
HDDS-3143. Rename silently ignored tests

adoroszlai:

1. `HddsServerUtilTest` and `LogSubcommandTest` are currently silently ignored, as their name does not match the `Test*` pattern Ozone includes in Surefire execution.  Rename these to `TestHddsServerUtil` and `TestLogSubcommand`, respectively.
2. Ignore `TestLogSubcommand#filterLog` as it is broken.  Filed [HDDS-3144](https://issues.apache.org/jira/browse/HDDS-3144) for the fix.
3. Move `testGetSCMAddresses` to `TestHddsUtils` since the method it tests (`getSCMAddresses`) is in `HddsUtils`, not `HddsServerUtil`

https://issues.apache.org/jira/browse/HDDS-3143

## How was this patch tested?

Ran affected unit tests.

https://github.com/adoroszlai/hadoop-ozone/runs/495296117

### DIFF
--- a/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/TestHddsUtils.java
+++ b/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/TestHddsUtils.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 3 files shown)


adoroszlai:
Thanks @avijayanhwx for the review.

PR #656, opened 2020-03-10
HDDS-3142. Create isolated enviornment for OM to test it without SCM.

elek:

`OmKeyGenerator` class from Freon can generate keys (open key + commit key). But this test tests both OM and SCM performance. It seems to be useful to have a method to test only the OM performance with faking the response from SCM.

Can be done easily with the same approach what we have in HDDS-3023: A simple utility class can be implemented and with byteman we can replace the client calls with the fake method.

https://issues.apache.org/jira/browse/HDDS-3142

## How was this patch tested?

 1. Download [byteman](https://byteman.jboss.org/)
 2. Start a pure OM (`ozone om --init` + `ozone om`) with the following JVM parameters: (change the path) 

```
-javaagent:/home/elek/prog/byteman/lib/byteman.jar=script:/home/elek/projects/ozone/dev-support/byteman/mock-scm.btm,boot:/home/elek/prog/byteman/lib/byteman.jar
-Dorg.jboss.byteman.transform.all
``` 

 3. Start a simple freon test: `ozone freon omkg`

Expected result: It should be possible to init and start OM without SCM and test it with the key generator)

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/freon/FakeClusterTopology.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/freon/FakeClusterTopology.java
@@ -1,0 +1,92 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hdds.freon;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.UUID;
+
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.DatanodeDetailsProto;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.Pipeline;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.Port;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationFactor;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationType;
+import org.apache.hadoop.hdds.scm.pipeline.PipelineID;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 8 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/freon/FakeScmBlockLocationProtocolClient.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Fake SCM client to return a simulated block location.
+ */
+public final class FakeScmBlockLocationProtocolClient {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(FakeScmBlockLocationProtocolClient.class);
+
+  public static final int BLOCK_PER_COUNTER = 1000;
```
adoroszlai: Intended to be `BLOCK_PER_CONTAINER`?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/freon/FakeScmContainerLocationProtocolClient.java
```diff
@@ -0,0 +1,80 @@
...
+  public static ScmContainerLocationResponse submitRequest(
+      ScmContainerLocationRequest req)
+      throws IOException {
+    try {
+      if (req.getCmdType() == Type.QueryNode) {
+        Builder builder = NodeQueryResponseProto.newBuilder()
+            .addDatanodes(Node.newBuilder()
+                .setNodeID(FakeClusterTopology.INSTANCE.getDatanode(1))
+                .addNodeStates(NodeState.HEALTHY)
+                .build());
```
adoroszlai: Isn't node 1 added in the loop, too?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/freon/FakeClusterTopology.java
```diff
@@ -0,0 +1,92 @@
...
+public class FakeClusterTopology {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(FakeClusterTopology.class);
+
+  public static final FakeClusterTopology INSTANCE = new FakeClusterTopology();
+
+  private List<DatanodeDetailsProto> datanodes = new ArrayList<>();
+
+  private List<Pipeline> piplines = new ArrayList<>();
```
adoroszlai: typo: `piplines` -> `pipelines`

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/freon/FakeScmContainerLocationProtocolClient.java
elek: Yeah, it came from an old code, thanks.

elek:
Thanks the review @adoroszlai, I fixed the typo bugs.

> Would be nice to include mock-scm.btm script in hadoop-ozone/dist/target for testing.

100% agree. I created an issue to track this: https://issues.apache.org/jira/browse/HDDS-3232

adoroszlai:

elek:
HDDS-3160 is accidentally added to this pr (sorry for that)
 1. I will revert the original merge
 2. Remove HDDS-3160
 3. Create a new PR without that
 4. And merge it after a green build

PR #660, opened 2020-03-11
HDDS-3104. Integration test crashes due to critical error in datanode.

swagle:
Created a flag to tell StateContext that shutDown was called and not to overreact!
(More details in Jira comments)

https://issues.apache.org/jira/browse/HDDS-3104

## How was this patch tested?
Verified by running TestContainerStateMachineFailureOnRead.


swagle:
> Thanks @swagle for working on this. I managed to reproduce the problem you describe by adding some sleep here and there. I think your patch would fix that.
> 
> In case shutdown state _is_ set due to some error I still don't think datanode should call `System.exit` when running tests. We might need a flag to disable exit in "mini cluster mode".

Thanks for review @adoroszlai

avijayanhwx:
Thanks for the patch @swagle and the review @adoroszlai.

PR #662, opened 2020-03-11
HDDS-3157. Fixed docker startup command.

cku328:

Fixed some docker startup commands.

https://issues.apache.org/jira/browse/HDDS-3157

## How was this patch tested?

No need.


cku328:
Thanks @adoroszlai for the review.

PR #663, opened 2020-03-11
HDDS-3156 update allocateContainer to remove additional createPipeline step.

timmylicheng:
Upate allocateContainer not to try creating pipeline every time.

https://issues.apache.org/jira/browse/HDDS-3156

## How was this patch tested?
Not needed.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManager.java
@@ -247,23 +247,29 @@ public class ContainerStateManager {
       final HddsProtos.ReplicationType type,
       final HddsProtos.ReplicationFactor replicationFactor, final String owner)
       throws IOException {
-
+    final List<Pipeline> pipelines = pipelineManager
+        .getPipelines(type, replicationFactor, Pipeline.PipelineState.OPEN);
     Pipeline pipeline;
-    try {
-      // TODO: #CLUTIL remove creation logic when all replication types and
-      // factors are handled by pipeline creator job.
-      pipeline = pipelineManager.createPipeline(type, replicationFactor);
-      pipelineManager.waitPipelineReady(pipeline.getId(), 0);
-    } catch (IOException e) {
-      final List<Pipeline> pipelines = pipelineManager
-          .getPipelines(type, replicationFactor, Pipeline.PipelineState.OPEN);
-      if (pipelines.isEmpty()) {
-        throw new IOException("Could not allocate container. Cannot get any" +
-            " matching pipeline for Type:" + type +
-            ", Factor:" + replicationFactor + ", State:PipelineState.OPEN");
-      }
+
+    if (!pipelines.isEmpty() && replicationFactor == ReplicationFactor.THREE
+        && type == ReplicationType.RATIS) {
+      // Let background create Ratis THREE pipelines.
       pipeline = pipelines.get((int) containerCount.get() % pipelines.size());
+    } else {
+      try {
+        pipeline = pipelineManager.createPipeline(type, replicationFactor);
+        pipelineManager.waitPipelineReady(pipeline.getId(), 0);
+      } catch (IOException e) {
+
+        if (pipelines.isEmpty()) {
+          throw new IOException("Could not allocate container. Cannot get any" +
+              " matching pipeline for Type:" + type +
+              ", Factor:" + replicationFactor + ", State:PipelineState.OPEN");
+        }
+        pipeline = pipelines.get((int) containerCount.get() % pipelines.size());
+      }
     }
+
     synchronized (pipeline) {
       return allocateContainer(pipelineManager, owner, pipeline);
     }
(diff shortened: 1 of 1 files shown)


timmylicheng:
This is going to let Ratis Three pipeline creation be handled only in BackgroundPipelineCreator. For all types or factor of pipelines, it still stay the same behavior.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManager.java
```diff
@@ -247,23 +247,29 @@ ContainerInfo allocateContainer(final PipelineManager pipelineManager,
...
-          .getPipelines(type, replicationFactor, Pipeline.PipelineState.OPEN);
-      if (pipelines.isEmpty()) {
-        throw new IOException("Could not allocate container. Cannot get any" +
-            " matching pipeline for Type:" + type +
-            ", Factor:" + replicationFactor + ", State:PipelineState.OPEN");
-      }
+
+    if (!pipelines.isEmpty() && replicationFactor == ReplicationFactor.THREE
+        && type == ReplicationType.RATIS) {
+      // Let background create Ratis THREE pipelines.
```
bharatviswa504: If ozone.scm.pipeline.creation.auto.factor.one is not disabled even ratis with factor one pipeline creation is taken care by BackGroundPipelineCreator.

bharatviswa504: So, shall we need to check that config, if enabled and let factor 1 also be skipped here?

timmylicheng: Yea, that's a good point. I update it.

nandakumar131:
Thanks @timmylicheng for working on this.

We don't need any complex conditions here. The below logic should be enough to handle

```
try {
  final List<Pipeline> pipelines = pipelineManager.getPipelines(type, replicationFactor, Pipeline.PipelineState.OPEN);
  Pipeline pipeline;
  if (pipelines.isEmpty()) {
    pipeline = pipelineManager.createPipeline(type, replicationFactor);
    pipelineManager.waitPipelineReady(pipeline.getId(), 0);
  } else {
    pipeline = pipelines.get((int) containerCount.get() % pipelines.size());
  }
} catch (IOException e) {
  throw new IOException("Could not allocate container. Cannot get any" +
              " matching pipeline for Type:" + type +
              ", Factor:" + replicationFactor + ", State:PipelineState.OPEN");
}
```

timmylicheng:
> Thanks @timmylicheng for working on this.
> 
> We don't need any complex conditions here. The below logic should be enough to handle
> 
> ```
> try {
>   final List<Pipeline> pipelines = pipelineManager.getPipelines(type, replicationFactor, Pipeline.PipelineState.OPEN);
>   Pipeline pipeline;
>   if (pipelines.isEmpty()) {
>     pipeline = pipelineManager.createPipeline(type, replicationFactor);
>     pipelineManager.waitPipelineReady(pipeline.getId(), 0);
>   } else {
>     pipeline = pipelines.get((int) containerCount.get() % pipelines.size());
>   }
> } catch (IOException e) {
>   throw new IOException("Could not allocate container. Cannot get any" +
>               " matching pipeline for Type:" + type +
>               ", Factor:" + replicationFactor + ", State:PipelineState.OPEN");
> }
> ```

@nandakumar131 I actually tried this and some tests fail. Basically some uts are expecting to have a new Ratis ONE pipeline when a new container is created. In your posted code, it will use an existing pipeline by default. 
My idea is to keep Ratis One behavior as it is: try to create pipeline every time a container is created. After pipeline is used up for creation, it will use the existing ones. Your idea seems to be using existing pipelines for both Ratis one and Ratis three. But like Bharat posts, Ratis one doesn't have to be auto created by background thread. We do have conditions when Ratis one is not auto created by a config. Also for StandAlone pipelines, background thread is also not working for them.
I was thinking of changing the least UTs. Would you suggest to modify UTs on the other hand? That could be a lot tho.

nandakumar131:
@timmylicheng  thanks for the clarification.

As we create the pipeline via heartbeat now it's not good to have it in the critical path; unless there are no pipelines available, we should try to avoid the creation of pipelines in allocate container calls and leave the job to background pipeline creator.

I can see that changing all the related tests seems hectic and we should also have a discussion before making one node pipeline creation default in background pipeline creator.

I'm good with the current changes for now. We can revisit this later.

timmylicheng:
> @timmylicheng thanks for the clarification.
> 
> As we create the pipeline via heartbeat now it's not good to have it in the critical path; unless there are no pipelines available, we should try to avoid the creation of pipelines in allocate container calls and leave the job to background pipeline creator.
> 
> I can see that changing all the related tests seems hectic and we should also have a discussion before making one node pipeline creation default in background pipeline creator.
> 
> I'm good with the current changes for now. We can revisit this later.

@nandakumar131 Sounds fair to me. Let's revisit the Ratis One pipeline later since it's not blocking anyone's use cases in real cluster.

nandakumar131:
Thanks @timmylicheng for the patch and thanks @bharatviswa504  for the review. I will merge this shortly.

PR #664, opened 2020-03-11
HDDS-3159. Bump RocksDB version to the latest one.

elek:

6.0.1 – our current version from RocksDB – released one year ago. Since than many new versions are released with important bug fixes.

I propose to update to the latest one...

https://issues.apache.org/jira/browse/HDDS-3159

## How was this patch tested?

Tested with `freon`:  I wrote 200M keys (allocate + close) to the OM with simulated SCM (which means that I had very high tp, ~10k key allocation / sec). It worked well.

For more details about the test please see #656


avijayanhwx:
Thank you @elek for the patch, @mukul1987 and @HorizonNet for the reviews.

PR #668, opened 2020-03-12
HDDS-3139. Pipeline placement should max out pipeline usage

timmylicheng:

Pipeline placement should select lowest load datanode as anchor.

https://issues.apache.org/jira/browse/HDDS-3139

## How was this patch tested?

UT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
@@ -25,7 +25,6 @@ import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.SCMCommonPlacementPolicy;
-import org.apache.hadoop.hdds.scm.container.placement.metrics.SCMNodeMetric;
 import org.apache.hadoop.hdds.scm.exceptions.SCMException;
 import org.apache.hadoop.hdds.scm.net.NetworkTopology;
 import org.apache.hadoop.hdds.scm.net.Node;
@@ -320,6 +319,50 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
     return results;
   }
 
+  private DatanodeDetails randomPick(List<DatanodeDetails> healthyNodes) {
+    DatanodeDetails datanodeDetails;
+    int firstNodeNdx = getRand().nextInt(healthyNodes.size());
+    int secondNodeNdx = getRand().nextInt(healthyNodes.size());
+
+    // There is a possibility that both numbers will be same.
+    // if that is so, we just return the node.
+    if (firstNodeNdx == secondNodeNdx) {
+      datanodeDetails = healthyNodes.get(firstNodeNdx);
+    } else {
+      DatanodeDetails firstNodeDetails = healthyNodes.get(firstNodeNdx);
+      DatanodeDetails secondNodeDetails = healthyNodes.get(secondNodeNdx);
+      datanodeDetails = nodeManager.getPipelinesCount(firstNodeDetails)
+          >= nodeManager.getPipelinesCount(secondNodeDetails)
+          ? secondNodeDetails : firstNodeDetails;
+    }
+    return datanodeDetails;
+  }
+
+  private List<DatanodeDetails> getLowerLoadNodes(
+      List<DatanodeDetails> nodes, int num) {
+    int maxPipelineUsage = nodes.size() * heavyNodeCriteria /
+        HddsProtos.ReplicationFactor.THREE.getNumber();
+    return nodes.stream()
+        // Skip the nodes which exceeds the load limit.
+        .filter(p -> nodeManager.getPipelinesCount(p) < num - maxPipelineUsage)
+        .collect(Collectors.toList());
+  }
+
+  private DatanodeDetails lowerLoadPick(List<DatanodeDetails> healthyNodes) {
+    int curPipelineCounts =  stateManager
+        .getPipelines(HddsProtos.ReplicationType.RATIS).size();
+    DatanodeDetails datanodeDetails;
+    List<DatanodeDetails> nodes = getLowerLoadNodes(
+        healthyNodes, curPipelineCounts);
+    if (nodes.isEmpty()) {
+      // random pick node if nodes load is at same level.
+      datanodeDetails = randomPick(healthyNodes);
+    } else {
+      datanodeDetails = nodes.stream().findFirst().get();
+    }
+    return datanodeDetails;
+  }
+
   /**
    * Find a node from the healthy list and return it after removing it from the
    * list that we are operating on.
@@ -333,24 +376,7 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
     if (healthyNodes == null || healthyNodes.isEmpty()) {
       return null;
     }
-    int firstNodeNdx = getRand().nextInt(healthyNodes.size());
-    int secondNodeNdx = getRand().nextInt(healthyNodes.size());
-
-    DatanodeDetails datanodeDetails;
-    // There is a possibility that both numbers will be same.
-    // if that is so, we just return the node.
-    if (firstNodeNdx == secondNodeNdx) {
-      datanodeDetails = healthyNodes.get(firstNodeNdx);
-    } else {
-      DatanodeDetails firstNodeDetails = healthyNodes.get(firstNodeNdx);
-      DatanodeDetails secondNodeDetails = healthyNodes.get(secondNodeNdx);
-      SCMNodeMetric firstNodeMetric =
-          nodeManager.getNodeStat(firstNodeDetails);
-      SCMNodeMetric secondNodeMetric =
-          nodeManager.getNodeStat(secondNodeDetails);
-      datanodeDetails = firstNodeMetric.isGreater(secondNodeMetric.get())
-          ? firstNodeDetails : secondNodeDetails;
-    }
+    DatanodeDetails datanodeDetails = lowerLoadPick(healthyNodes);
     healthyNodes.remove(datanodeDetails);
     return datanodeDetails;
   }
@@ -373,14 +399,14 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
       return null;
     }
 
-    for (DatanodeDetails node : healthyNodes) {
-      if (excludedNodes.contains(node) ||
-          anchor.getNetworkLocation().equals(node.getNetworkLocation())) {
-        continue;
-      } else {
-        return node;
-      }
+    List<DatanodeDetails> nodesOnOtherRack = healthyNodes.stream().filter(
+        p -> !excludedNodes.contains(p)
+            && !anchor.getNetworkLocation().equals(p.getNetworkLocation()))
+        .collect(Collectors.toList());
+    if (!nodesOnOtherRack.isEmpty()) {
+      return lowerLoadPick(nodesOnOtherRack);
     }
+
     return null;
   }
 
@@ -414,6 +440,16 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
     Collection<Node> excluded = new ArrayList<>();
     if (excludedNodes != null && excludedNodes.size() != 0) {
       excluded.addAll(excludedNodes);
+    }
+
+    List<DatanodeDetails> healthyNodes = nodeManager.getNodes(
+        HddsProtos.NodeState.HEALTHY);
+    List<DatanodeDetails> lowerLoadNodes = getLowerLoadNodes(healthyNodes,
+        stateManager.getPipelines(HddsProtos.ReplicationType.RATIS).size());
+    if (!lowerLoadNodes.isEmpty()) {
+      // Consider nodes with lower load first.
+      lowerLoadNodes.stream().forEach(p->healthyNodes.remove(p));
+      excludedNodes.addAll(healthyNodes);
     }
 
     Node pick = networkTopology.chooseRandom(
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
@@ -348,12 +348,23 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
    * Get a list of nodes with lower load than max pipeline number.
    */
   private List<DatanodeDetails> getLowerLoadNodes(
-      List<DatanodeDetails> nodes, int num) {
-    int maxPipelineUsage = nodes.size() * heavyNodeCriteria /
-        HddsProtos.ReplicationFactor.THREE.getNumber();
+      List<DatanodeDetails> nodes, int mark) {
+    int limit = nodes.size() * heavyNodeCriteria
+        / HddsProtos.ReplicationFactor.THREE.getNumber();
     return nodes.stream()
         // Skip the nodes which exceeds the load limit.
-        .filter(p -> nodeManager.getPipelinesCount(p) < maxPipelineUsage - num)
+        .filter(p -> nodeManager.getPipelinesCount(p) < limit - mark)
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Get a list of nodes with higher load than max pipeline number.
+   */
+  private List<DatanodeDetails> getHigherLoadNodes(
+      List<DatanodeDetails> nodes, int mark) {
+    return nodes.stream()
+        // Skip the nodes with lower load than mark.
+        .filter(p -> nodeManager.getPipelinesCount(p) > mark)
         .collect(Collectors.toList());
   }
 
@@ -370,7 +381,7 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
       // random pick node if nodes load is at same level.
       datanodeDetails = randomPick(healthyNodes);
     } else {
-      datanodeDetails = nodes.stream().findFirst().get();
+      datanodeDetails = nodes.stream().findAny().get();
     }
     return datanodeDetails;
   }
@@ -453,24 +464,38 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
     if (excludedNodes != null && excludedNodes.size() != 0) {
       excluded.addAll(excludedNodes);
     }
-
-    List<DatanodeDetails> healthyNodes = nodeManager.getNodes(
+    List<DatanodeDetails> candidates = nodeManager.getNodes(
         HddsProtos.NodeState.HEALTHY);
-    List<DatanodeDetails> lowerLoadNodes = getLowerLoadNodes(healthyNodes,
-        stateManager.getPipelines(HddsProtos.ReplicationType.RATIS).size());
-    if (!lowerLoadNodes.isEmpty()) {
+    int currentPipelineNum  = stateManager
+        .getPipelines(HddsProtos.ReplicationType.RATIS).size();
+    List<DatanodeDetails> higherLoadNodes = getHigherLoadNodes(candidates,
+        currentPipelineNum);
+    if (!higherLoadNodes.isEmpty()) {
       // Consider nodes with lower load first.
-      lowerLoadNodes.stream().forEach(p->healthyNodes.remove(p));
-      excludedNodes.addAll(healthyNodes);
+      excluded.addAll(higherLoadNodes);
     }
 
     Node pick = networkTopology.chooseRandom(
         anchor.getNetworkLocation(), excluded);
     DatanodeDetails pickedNode = (DatanodeDetails) pick;
+
     if (pickedNode == null) {
       LOG.debug("Pick node is null, excluded nodes {}, anchor {}.",
           excluded, anchor);
+      return null;
     }
+
+    if (nodeManager.getPipelinesCount(pickedNode)
+        >= heavyNodeCriteria - currentPipelineNum
+        - HddsProtos.ReplicationFactor.THREE.getNumber()) {
+      LOG.debug("Picked node violates pipeline load balance," +
+              " pickedNode {}, pipeline load {}.",
+          pickedNode, nodeManager.getPipelinesCount(pickedNode));
+      // When topology picked a node that violates the pipeline load balance,
+      // we ignore the result.
+      pickedNode = null;
+    }
+
     return pickedNode;
   }
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
@@ -27,13 +27,12 @@ import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.SCMCommonPlacementPolicy;
 import org.apache.hadoop.hdds.scm.exceptions.SCMException;
 import org.apache.hadoop.hdds.scm.net.NetworkTopology;
-import org.apache.hadoop.hdds.scm.net.Node;
 import org.apache.hadoop.hdds.scm.node.NodeManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
-import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -75,19 +74,8 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
         ScmConfigKeys.OZONE_DATANODE_PIPELINE_LIMIT_DEFAULT);
   }
 
-  /**
-   * Returns true if this node meets the criteria.
-   *
-   * @param datanodeDetails DatanodeDetails
-   * @param nodesRequired nodes required count
-   * @return true if we have enough space.
-   */
-  @VisibleForTesting
-  boolean meetCriteria(DatanodeDetails datanodeDetails, int nodesRequired) {
-    if (heavyNodeCriteria == 0) {
-      // no limit applied.
-      return true;
-    }
+  int currentPipelineCount(DatanodeDetails datanodeDetails, int nodesRequired) {
+
     // Datanodes from pipeline in some states can also be considered available
     // for pipeline allocation. Thus the number of these pipeline shall be
     // deducted from total heaviness calculation.
@@ -109,16 +97,10 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
         pipelineNumDeductable++;
       }
     }
-    boolean meet = (nodeManager.getPipelinesCount(datanodeDetails)
-        - pipelineNumDeductable) < heavyNodeCriteria;
-    if (!meet && LOG.isDebugEnabled()) {
-      LOG.debug("Pipeline Placement: can't place more pipeline on heavy " +
-          "datanode： " + datanodeDetails.getUuid().toString() +
-          " Heaviness: " + nodeManager.getPipelinesCount(datanodeDetails) +
-          " limit: " + heavyNodeCriteria);
-    }
-    return meet;
+    return pipelines.size() - pipelineNumDeductable;
   }
+
+
 
   /**
    * Filter out viable nodes based on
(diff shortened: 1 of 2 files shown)


timmylicheng:
When there are 5 nodes and each is allowed to have 5 pipelines:
Former pipeline placement track:
Pipeline count on node: 2
Pipeline count on node: 5
Pipeline count on node: 4
Pipeline count on node: 5
Pipeline count on node: 5

pipeline max: 8, pipeline actual: 7

after this change:
Pipeline count on node: 4
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5

pipeline max: 8, pipeline actual: 8

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -315,6 +314,50 @@ DatanodeDetails fallBackPickNodes(
...
+    int curPipelineCounts =  stateManager
+        .getPipelines(HddsProtos.ReplicationType.RATIS).size();
+    DatanodeDetails datanodeDetails;
+    List<DatanodeDetails> nodes = getLowerLoadNodes(
+        healthyNodes, curPipelineCounts);
+    if (nodes.isEmpty()) {
+      // random pick node if nodes load is at same level.
+      datanodeDetails = randomPick(healthyNodes);
+    } else {
+      datanodeDetails = nodes.stream().findFirst().get();
```
sodonnel: Could this be simplified to just `nodes.get(0)`? In the else branch we already know the list is not empty, so I think we can just pick the first node in the list safely.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -315,6 +314,50 @@ DatanodeDetails fallBackPickNodes(
     return results;
   }
 
+  private DatanodeDetails randomPick(List<DatanodeDetails> healthyNodes) {
```
sodonnel: I think a Java doc would be useful for this method to explain how it works. It seems to pick two random nodes and then return the one with the less load - is that correct?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
sodonnel: Actually, thinking about this more - should this pick a random node from LowerLoadNodes, rather than the first node? We call `getLowerLoadNodes` which returns a list of nodes where the overloaded nodes are removed, but I don't think the list is sorted in anyway. The healthyNodes will will probably be naturally in the same order each time it is generated from the NodeManager. This means this lowerLoadPick method might return the same node on each call until it is overloaded. Then it would be excluded and the next node would be picked and so on. It would probably be better if we picked a random node from the less loaded nodes. Or, sort the list by load and return the first one so we are always picked the node with the least node.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -315,6 +314,50 @@ DatanodeDetails fallBackPickNodes(
...
+    return datanodeDetails;
+  }
+
+  private List<DatanodeDetails> getLowerLoadNodes(
+      List<DatanodeDetails> nodes, int num) {
+    int maxPipelineUsage = nodes.size() * heavyNodeCriteria /
+        HddsProtos.ReplicationFactor.THREE.getNumber();
+    return nodes.stream()
+        // Skip the nodes which exceeds the load limit.
+        .filter(p -> nodeManager.getPipelinesCount(p) < num - maxPipelineUsage)
```
sodonnel: I think there is a bug here. I put some debug in and ran the test you added as part of this change, and this method always returned an empty list.

maxPipelineUsage starts at 13, so we have:

```
"0 < 3 - 13" -> "0 < -10 "-> false and all the nodes are filtered out.

Should this be:

```
 .filter(p -> nodeManager.getPipelinesCount(p) < maxPipelineUsage - num)
```

File: hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java (hunk 1 above)
Comment (sodonnel) [reply hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java]: When i fixed the suspected bug I mentioned above, and then ran the test. The nodes do appear to fill their piplines on a node by node basis and then the test failed as each node did not have at least the average pipelines.

Making this change got it to pass again:

```
datanodeDetails = nodes.get(getRand().nextInt(nodes.size()) );//stream().findFirst().get();
```

But it might be ever better if we sorted the list by pipeline count ascending and then took the first one, but it would be more expensive.

Comment (timmylicheng) [reply hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java]: Sorting the node list would be expensive for large cluster. That's the reason why I choose to do this 'water mark' filter for selecting nodes with lower load. 

I can def do a findAny() kinda thing for random pick.

File: hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java (hunk 2 above)
Comment (timmylicheng) [reply hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java]: Yea. Your understanding is correct. I will add doc as description.

File: hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java (hunk 1 above)
Comment (sodonnel) [reply hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java]: You could probably do a "top N" type of sort where you only keep the lowest loaded node, but it would require a bit of a refactor of `getLowerLoadNodes`, probably changing it to `getLowestLodeNode`. That would avoid a full sort and would not be much more expensive that the current code, eg:
```
DatanodeDetails lowest = null;
int lowestPipelineCount;
for (DatanodeDetails dn : nodes) {
  int nodePipelines = node.getPipelineCount();
  if (nodePipelines > limit) {
    continue;
  }
  if ((lowest == null) || lowestPipelineCount > nodePipelines {
    lowest = node;
    lowestPipelineCount = nodePipelines;
  }
}
return lowest;
```

**Comment by timmylicheng** [comment]
Before
pipeline max: 13, pipeline actual: 12
Pipeline count on node: 1
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5

After
pipeline max: 13, pipeline actual: 13
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 4
Pipeline count on node: 5
Pipeline count on node: 5

When the node number gets large, they perform identically.
Before
pipeline max: 63, pipeline actual: 63
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 4
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5

After
pipeline max: 63, pipeline actual: 63
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 5
Pipeline count on node: 4
Pipeline count on node: 5

File: hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java (hunk 1 above)
Comment (timmylicheng) [reply hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java]: I feel like the outcome will be similar tho. The current implementation should work just fine.

File: hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java (hunk 3 above)
Comment (timmylicheng) [reply hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java]: Good catch. I update it. Thanks!

**Comment by sodonnel** [comment]
Hi @timmylicheng this patch is giving some conflicts now, probably as we merged the other pipeline related change. Could you rebase against master and push it again please?

File: hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java (hunk 1 above)
Comment (sodonnel) [reply hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java]: I am fine with doing a "random pick" from the `getLowerLoadNodes` list or using the idea I had above to get the node with the lowest load each time. I think there are advantages to each of them. I believe we do need to go with one of those ideas, as picking the first one will not work well.

The random pick is simple, but it may not spread the load evenly every time.

Picking the lowest one each time is slightly more complicated, but it does guarantee to always use the lowest load node first and will spread the load evenly for sure. However it is less random - eg if a new node joins the cluster, then it will be used for the next N pipelines until it reaches the same load as some others.

File: hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java [hunk 4]
```diff
@@ -316,36 +315,75 @@ DatanodeDetails fallBackPickNodes(
...
     }
+    return datanodeDetails;
+  }
+
+  /**
+   * Get a list of nodes with lower load than max pipeline number.
+   */
+  private List<DatanodeDetails> getLowerLoadNodes(
+      List<DatanodeDetails> nodes, int num) {
+    int maxPipelineUsage = nodes.size() * heavyNodeCriteria /
```
sodonnel: I am not sure if this calculation is correct. The reason is that the healthy node list is already filtered to include only nodes with fewer than `heavyNodeCritera` pipelines in `filterViableNodes()`. Therefore the size of the list passed into this method gets smaller as the nodes are used up and eventually it stops returning any nodes, even though there are nodes valid to return. From some debug messages I added:

```
2020-03-30 14:18:31,682 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:chooseDatanodes(202)) - There is no topology
2020-03-30 14:18:31,682 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:chooseNode(391)) - In chooseNode
2020-03-30 14:18:31,682 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:getLowerLoadNodes(351)) - Max pipeline usage is: 8
2020-03-30 14:18:31,682 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:lowerLoadPick(368)) - getLowerLoadNodes() returned empty list
2020-03-30 14:18:31,682 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:chooseNode(391)) - In chooseNode
2020-03-30 14:18:31,682 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:getLowerLoadNodes(351)) - Max pipeline usage is: 6
2020-03-30 14:18:31,682 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:lowerLoadPick(368)) - getLowerLoadNodes() returned empty list
2020-03-30 14:18:31,682 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:chooseNode(391)) - In chooseNode
2020-03-30 14:18:31,683 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:getLowerLoadNodes(351)) - Max pipeline usage is: 5
2020-03-30 14:18:31,683 INFO  pipeline.PipelinePlacementPolicy (PipelinePlacementPolicy.java:lowerLoadPick(368)) - getLowerLoadNodes() returned empty list
2020-03-30 14:18:31,683 INFO  pipeline.PipelineStateManager (PipelineStateManager.java:addPipeline(54)) - Created pipeline Pipeline[ Id: 3da2deda-7839-4e5c-88e0-a613729b99fd, Nodes: 0cbe69da-7ef6-43ff-a11a-7ba716ec7c9c{ip: 242.96.90.116, host: localhost-242.96.90.116, networkLocation: /default-rack, certSerialId: null}b4d70058-fb84-410a-8ca8-cc4e8e944fae{ip: 28.158.147.87, host: localhost-28.158.147.87, networkLocation: /default-rack, certSerialId: null}03846f48-f6e5-4dba-97b2-5ab8a7be1564{ip: 128.200.5.118, host: localhost-128.200.5.118, networkLocation: /default-rack, certSerialId: null}, Type:RATIS, Factor:THREE, State:ALLOCATED, leaderId:null, CreationTimestamp2020-03-30T13:18:31.683Z]
2020-03-30 14:18:31,683 INFO  pipeline.TestPipelinePlacementPolicy (TestPipelinePlacementPolicy.java:testPickLowestLoadAnchor(112)) - Pipeline count for this node is 5
2020-03-30 14:18:31,683 INFO  pipeline.TestPipelinePlacementPolicy (TestPipelinePlacementPolicy.java:testPickLowestLoadAnchor(112)) - Pipeline count for this node is 5
2020-03-30 14:18:31,683 INFO  pipeline.TestPipelinePlacementPolicy (TestPipelinePlacementPolicy.java:testPickLowestLoadAnchor(112)) - Pipeline count for this node is 5
2020-03-30 14:18:31,683 INFO  pipeline.TestPipelinePlacementPolicy (TestPipelinePlacementPolicy.java:testPickLowestLoadAnchor(112)) - Pipeline count for this node is 3
2020-03-30 14:18:31,683 INFO  pipeline.TestPipelinePlacementPolicy (TestPipelinePlacementPolicy.java:testPickLowestLoadAnchor(112)) - Pipeline count for this node is 2
2020-03-30 14:18:31,683 INFO  pipeline.TestPipelinePlacementPolicy (TestPipelinePlacementPolicy.java:testPickLowestLoadAnchor(112)) - Pipeline count for this node is 2
2020-03-30 14:18:31,683 INFO  pipeline.TestPipelinePlacementPolicy (TestPipelinePlacementPolicy.java:testPickLowestLoadAnchor(112)) - Pipeline count for this node is 3
2020-03-30 14:18:31,684 INFO  pipeline.TestPipelinePlacementPolicy (TestPipelinePlacementPolicy.java:testPickLowestLoadAnchor(112)) - Pipeline count for this node is 2
```

If the nodes are filtered by load count in filterViableNodes, do we actually need this getLowerLoadNodes method?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: @sodonnel I'm navigating thru another problem with chooseNodeFromNetworkTopology. Because topology only gives interface like networkTopology.chooseRandom(anchor.getNetworkLocation(), excluded), I have to move all higher load nodes into excluded and allow topology to pick one. So the original water mark cannot guarantee full load balance due to fall back logic I added. 

I could see your proposed sort will help picking the lowEST node. How could it help leverage to have higher load with topology? I'm still experimenting...

timmylicheng:
> Hi @timmylicheng this patch is giving some conflicts now, probably as we merged the other pipeline related change. Could you rebase against master and push it again please?

@sodonnel I rebase the code with minor conflicts in test class, but the test won't pass. I took a close look and made some change. But I realize the issue that I mention in the last comment about how to leverage with chooseNodeFromTopology. Wanna learn your thoughts.

sodonnel:
> @sodonnel I rebase the code with minor conflicts in test class, but the test won't pass. I took a close look and made some change. But I realize the issue that I mention in the last comment about how to leverage with chooseNodeFromTopology. Wanna learn your thoughts.

I think one problem is this line:

```
datanodeDetails = nodes.stream().findAny().get();
```
The findAny method does not seem to return a random entry - so the same node is returned until it uses up its pipeline allocation.

I am also not sure about the limit calculation in getLowerLoadNodes:

```
 int limit = nodes.size() * heavyNodeCriteria
        / HddsProtos.ReplicationFactor.THREE.getNumber();
```

Adding debug, I find this method starts to return an empty list when there are still available nodes to handle the pipeline.

Also in `filterViableNodes()` via the `meetCriteria()` method, nodes with more than the heavy load limit are already filtered out, so you are guaranteed your healthy node list container only nodes with the capacity to take another pipeline. So I wonder why we need to filter the nodes further.

> But I realize the issue that I mention in the last comment about how to leverage with chooseNodeFromTopology.

There seems to be some inconsistency in how we pick the nodes (not just in this PR, but in the wider code). Eg in `chooseNodeBasedOnRackAwareness()` we don't call into NetworkTopology(), but instead we use the `getNetworkLocation()` method on the `DatanodeDetails` object to find nodes that do not match the anchor's location. 

Then later in `chooseNodeFromNetworkTopology()` we try to find a node where location is equal to the anchor and that is where we call into `networkTopology.chooseRandom()`. Could we not avoid that call, and avoid generating a new list of nodes and do something similar to `chooseNodeBasedOnRackAwareness()`, using the `getNetworkLocation()` method to find matching nodes. That would probably be more efficient that the current implementation.

As we are also then able to re-use the same list of healthy nodes everywhere without more filtering, maybe we could sort that list once by pipeline count in filterViableNodes or meetCriteria and then later always pick the node with the lowest load, filling the nodes up that way.

I hope this comment makes sense as it is very long.

sodonnel:
@timmylicheng I did some more experimenting and I have attached a patch file that should apply on your latest PR. In this, I quickly implemented what I outlined above. It passes all the existing tests, but this code is probably not production ready. There may be some cases I have not considered yet. It is something we can consider to change things here. Let me know your thoughts please.

sodonnel:
[example.txt](https://github.com/apache/hadoop-ozone/files/4416772/example.txt)

timmylicheng:
> > @sodonnel I rebase the code with minor conflicts in test class, but the test won't pass. I took a close look and made some change. But I realize the issue that I mention in the last comment about how to leverage with chooseNodeFromTopology. Wanna learn your thoughts.
> 
> I think one problem is this line:
> 
> ```
> datanodeDetails = nodes.stream().findAny().get();
> ```
> 
> The findAny method does not seem to return a random entry - so the same node is returned until it uses up its pipeline allocation.
> 
> I am also not sure about the limit calculation in getLowerLoadNodes:
> 
> ```
>  int limit = nodes.size() * heavyNodeCriteria
>         / HddsProtos.ReplicationFactor.THREE.getNumber();
> ```
> 
> Adding debug, I find this method starts to return an empty list when there are still available nodes to handle the pipeline.
> 
> Also in `filterViableNodes()` via the `meetCriteria()` method, nodes with more than the heavy load limit are already filtered out, so you are guaranteed your healthy node list container only nodes with the capacity to take another pipeline. So I wonder why we need to filter the nodes further.
> 
> > But I realize the issue that I mention in the last comment about how to leverage with chooseNodeFromTopology.
> 
> There seems to be some inconsistency in how we pick the nodes (not just in this PR, but in the wider code). Eg in `chooseNodeBasedOnRackAwareness()` we don't call into NetworkTopology(), but instead we use the `getNetworkLocation()` method on the `DatanodeDetails` object to find nodes that do not match the anchor's location.
> 
> Then later in `chooseNodeFromNetworkTopology()` we try to find a node where location is equal to the anchor and that is where we call into `networkTopology.chooseRandom()`. Could we not avoid that call, and avoid generating a new list of nodes and do something similar to `chooseNodeBasedOnRackAwareness()`, using the `getNetworkLocation()` method to find matching nodes. That would probably be more efficient that the current implementation.
> 
> As we are also then able to re-use the same list of healthy nodes everywhere without more filtering, maybe we could sort that list once by pipeline count in filterViableNodes or meetCriteria and then later always pick the node with the lowest load, filling the nodes up that way.
> 
> I hope this comment makes sense as it is very long.

@sodonnel Thanks for the consideration. I've updated the patch according to your example.

sodonnel:
Thanks for the update. Its a holiday weekend here in Europe, so it will probably be Tuesday before I get to look at this.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -373,13 +354,31 @@ protected DatanodeDetails chooseNodeBasedOnRackAwareness(
...
+  @VisibleForTesting
+  protected DatanodeDetails chooseNodeBasedOnSameRack(
+      List<DatanodeDetails> healthyNodes,  List<DatanodeDetails> excludedNodes,
+      NetworkTopology networkTopology, DatanodeDetails anchor) {
+    Preconditions.checkArgument(networkTopology != null);
+    if (checkAllNodesAreEqual(networkTopology)) {
+      return null;
+    }
+
+    List<DatanodeDetails> nodesOnOtherRack = healthyNodes.stream().filter(
```
sodonnel, line 376: Can you rename nodeOnOtherRack to nodesOnSameRack here please?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -110,17 +97,11 @@ boolean meetCriteria(DatanodeDetails datanodeDetails, int nodesRequired) {
...
-    }
-    return meet;
+    return pipelines.size() - pipelineNumDeductable;
   }
 
+
+
   /**
    * Filter out viable nodes based on
    * 1. nodes that are healthy
```
sodonnel, line 107: We should add a note to this Java doc that the returned list is sorted in order of pipeline count, starting with the lowest to highest.

sodonnel:
I think this looks better and I have just a few minor comments. I added a few inline plus these two:

At line 243:

```
// First choose an anchor nodes randomly
DatanodeDetails anchor = chooseNode(healthyNodes);
```

This no longer picks a random node - it just picks the lowest loaded node. I wonder if this first anchor node should be a random node?

Method getHigherLoadNodes() does not seem to be used any longer, so we can remove it.

This code has changed a lot with this patch, and this is really my first time looking at it. It would be great if @ChenSammi could give this a review too, after addressing the minor points I mentioned here?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Updated

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Updated.

timmylicheng:
> I think this looks better and I have just a few minor comments. I added a few inline plus these two:
> 
> At line 243:
> 
> ```
> // First choose an anchor nodes randomly
> DatanodeDetails anchor = chooseNode(healthyNodes);
> ```
> 
> This no longer picks a random node - it just picks the lowest loaded node. I wonder if this first anchor node should be a random node?
> 
> Method getHigherLoadNodes() does not seem to be used any longer, so we can remove it.
> 
> This code has changed a lot with this patch, and this is really my first time looking at it. It would be great if @ChenSammi could give this a review too, after addressing the minor points I mentioned here?

I've addressed these comments. Thanks.
I will check if @ChenSammi had bandwidth to do a review.

sodonnel:
I am a little concerned about the expense of forming the list of healthy nodes on large clusters. We have to do quite a lot of work to form a list and then only use 3 nodes from the list. Even the method `currentPipelineCount()` needs to do a few map lookups per node to get the current pipeline count. This is the case even before this change. Creating a pipeline on a large cluster would be expensive already, but this change probably makes it worse, due to the sort needed. I know it was me who suggested the sort.

I think the code as it is will work OK upto about 1000 nodes, and then the performance will drop off as the number of nodes goes toward 10k.

Eg here are some benchmarks I created using this test code, which is similar to what we are doing in `filterViableNodes()`:

```
  public List<Object> sortingWithMap(BenchmarkState state) {
    return state.otherList.stream()
        .map(o -> new Mock(o, state.rand.nextInt(20)))
        .filter(o -> o.getSize() <= 20)
        .sorted(Comparator.comparingInt(Mock::getSize))
        .map(o -> o.getObject())
  }
```

The OPs per second for various list sizes are:

```
Benchmark               (listSize)   Mode  Cnt       Score     Error  Units
Sorting.sortingWithMap         100  thrpt    3  113948.345 ± 446.426  ops/s
Sorting.sortingWithMap        1000  thrpt    3    9468.507 ± 894.138  ops/s
Sorting.sortingWithMap        5000  thrpt    3    1931.612 ± 263.919  ops/s
Sorting.sortingWithMap       10000  thrpt    3     970.745 ±  25.823  ops/s
Sorting.sortingWithMap      100000  thrpt    3      87.684 ±  35.438  ops/s
```

For a 1000 node cluster, with 10 pipelines per node, we would be looking at about 1 second to form all the piplines.

For a 5k node cluster, it would be about 25 seconds

For a 10k node cluster it would be 103 seconds, but even here, that would be at close to 1000 pipelines per second.

Those approx times are just for the filterViableNodes step, and not any other logic.

Removing the sort step from the above code makes it go about 5 times faster and removing the two map steps and the sort makes it overall 10x faster.

It we want to allocate the pipelines evenly, where the lowest loaded nodes always get the new pipeline, then I don't think we can avoid the sort. However on a larger cluster, picking nodes that have not exceeded the limit at random would probably result in quite an even spread, and we could do that more efficiently I think.

What do you think? 

If Sammi does not have bandwidth to give this a further review, I will see if I can get someone else to take a look too.

fapifta:
Hi @timmylicheng, as we discussed with @sodonnel I have taken a look at the PR, and we discussed about that sorting in the algorithm which feels to limit the performance of this algorithm.

At this stage, as we do not have really large cluster implementations, we do not have a real measurable impact with the current solution, in the long run we might need to optimise the code, if we change the pipeline lifecycle, or if we have some really large clusters running into trouble, because of the performance of this code, but in that case we should certainly measure which part is the bottleneck, as at this point we can not be sure that the sort in this piece is the problem. (I feel that the way how we get the initial healthy nodes list is a similarly heavy part for example, though as I said at this point I am unsure.)

So all in all I am +1 (non-binding) to commit this change, but we should note somewhere that the pipeline creation logic in the rack aware setup can cause startup performance problems as the algorithm is complex.

sodonnel:
With the second look @fapifta gave this, I am +1 to commit. We just need to do two more things:

1. Create a Jira with my benchmark results and some of the details from my comment earlier today in it, and add a comment to `filterViableNodes()` mentioning this new Jira as a "todo for performance on large clusters".

2. Push a new commit to get a green CI run. We have one failing integration test, which is probably due to it being flaky. However I cannot commit the change until we have all tests green.

timmylicheng:
@sodonnel Thanks for the efforts for benchmarking. That was a valid concern for large cluster in the future. I created https://issues.apache.org/jira/browse/HDDS-3466 to track this potential bottleneck.
Say we allow 5 pipelines per DN, 10K pipelines require 10K*3/5 = 6K DNs.
My team is testing how many DNs every OM and SCM can hold up. My takeaway is we are going to resolve quite a lot of issues on GRPC and Ratis before we move Ozone up to 6K DNs per SCM. So we may not be able to see this bottleneck in short term in prod cluster. However, we def wanna track this or do some testing for this to prevent it from being a blocker. 

The major impact for this once being a bottleneck is createPipeline process would be long which could cause a slow cold start for SCM and DNs. On the other hand, 10K pipeline leads to many pipeline reports and container reports for single SCM. We may wanna limit overall pipeline counts per SCM ultimately.

@fapifta Yea I agree with you. Let's use HDDS-3466 to track this and we are going to learn more about SCM's ability and best setup as we gradually move up to large cluster. My team is testing 200 DNs with single OM and SCM at this point (but in k8s env). Hopefully we can move further to have a clearer picture.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Removed and updated with new method.

sodonnel:
@timmylicheng Thanks for the update and the Jira. Unfortunately Github is now reporting conflicts, possibly due to HDDS-3101 which was merged. Could you resolve the conflicts and push again and hopefully we will get a green build to commit this. Thanks!

PR #669, opened 2020-03-12
HDDS-3144. Fixed failed test case.

cku328:

I use the debugger to get a log string that calls the [processLogLine(...)](https://github.com/apache/hadoop-ozone/blob/c1997218a4e1a6695a275c73cf85360cd046329c/hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/LogSubcommand.java#L147) method, which is:
```
2020-03-12 20:57:54,001 [TRACE|org.apache.hadoop.hdds.scm.node.SCMNodeManager|SCMNodeManager] HB is received from [datanode=localhost]: <json>storageReport {\n  storageUuid: "DS-f65eb957-fc2d-4b77-b4a3-e96ae2bd2ca6"\n  storageLocation: "/tmp/hadoop-neo/dfs/data"\n  capacity: 250438021120\n  scmUsed: 16384\n  remaining: 212041244672\n  storageType: DISK\n  failed: false\n}\n</json>
```
debug1

When I call this method with this string as a parameter, it outputs:
```
2020-03-12 20:57:54,001 [TRACE|org.apache.hadoop.hdds.scm.node.SCMNodeManager|SCMNodeManager] HB is received from [datanode=localhost]: \n
storageReport {\n
  storageUuid: "DS-f65eb957-fc2d-4b77-b4a3-e96ae2bd2ca6"\n
  storageLocation: "/tmp/hadoop-neo/dfs/data"\n
  capacity: 250438021120\n
  scmUsed: 16384\n
  remaining: 212041244672\n
  storageType: DISK\n
  failed: false\n
}\n
```
debug2

Finally, I modified the test case to apply the above test strings and results.

https://issues.apache.org/jira/browse/HDDS-3144

## How was this patch tested?

Ran UTs & checkstyle.sh

### DIFF
--- a/hadoop-ozone/insight/src/test/java/org/apache/hadoop/ozone/insight/TestLogSubcommand.java
+++ b/hadoop-ozone/insight/src/test/java/org/apache/hadoop/ozone/insight/TestLogSubcommand.java
@@ -18,7 +18,6 @@
 package org.apache.hadoop.ozone.insight;
 
 import org.junit.Assert;
-import org.junit.Ignore;
 import org.junit.Test;
 
 /**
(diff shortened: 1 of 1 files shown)


cku328:
Thanks @elek for the review.

PR #670, opened 2020-03-12
HDDS-3169. Revert Recon config key change for HTTP Server keytab.

avijayanhwx:
There was a recent change to the config key used by Recon's HTTP server for its keytab. This needs to be reverted since it is breaking compatibility with existing cluster deployments. 

https://issues.apache.org/jira/browse/HDDS-3169

## How was this patch tested?
Manually tested.


avijayanhwx:
cc @vivekratnavel

avijayanhwx:
Thank you for the reviews @arp7 & @mukul1987.

PR #671, opened 2020-03-13
HDDS-3086. Failure running integration test it-freon.

bshashikant:
Added configs for rpc and watch timeouts for ratis clent and server.
Also, added a method to wait for atleast one ratis pipeline to get into open state before client IO starts.

https://issues.apache.org/jira/browse/HDDS-3086

## How was this patch tested?
Ran he unit tests and verified.


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/freon/TestDataValidate.java
```diff
@@ -33,9 +38,28 @@
...
+                            RATIS_SERVER_WATCH_REQUEST_TIMEOUT_KEY,
+            10, TimeUnit.SECONDS);
+    conf.setTimeDuration(
+            RatisHelper.HDDS_DATANODE_RATIS_CLIENT_PREFIX_KEY+ "." +
+                    "rpc.request.timeout",
+            3, TimeUnit.SECONDS);
+    conf.setTimeDuration(
+            RatisHelper.HDDS_DATANODE_RATIS_CLIENT_PREFIX_KEY+ "." +
+                    "watch.request.timeout",
+            10, TimeUnit.SECONDS);
```
adoroszlai: Can we make these configs to be the default, somewhere around `MiniOzoneCluster`?  Eg. let `Builder.build()` add them if they are not explicitly set.  I hope that would let us avoid repeating them for all integration tests.

bshashikant: yes, i think we can do that..

adoroszlai:
Thanks @bshashikant for the update.

PR #673, opened 2020-03-13
HDDS-3064. Get Key is hung when READ delay is injected in chunk file path.

bshashikant:
Added a deadline for rpc stub in the read path. It will ensure if a rpc response does not complete
within the deadline, it will fail.

https://issues.apache.org/jira/browse/HDDS-3064

## How was this patch tested?
It was verified in the actual test up where the problem originated.

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
@@ -189,7 +189,14 @@ public class XceiverClientGrpc extends XceiverClientSpi {
     ManagedChannel channel = channelBuilder.build();
     XceiverClientProtocolServiceStub asyncStub =
         XceiverClientProtocolServiceGrpc.newStub(channel);
-    asyncStubs.put(dn.getUuid(), asyncStub);
+    long duration = config.getTimeDuration(OzoneConfigKeys.
+            OZONE_CLIENT_READ_TIMEOUT, OzoneConfigKeys
+            .OZONE_CLIENT_READ_TIMEOUT_DEFAULT, TimeUnit.SECONDS);
+
+    // set the grpc dealine here so as if the response is not received
+    // in the configured time, the rpc will fail with DEADLINE_EXCEEDED here
+    asyncStubs.put(dn.getUuid(), asyncStub.withDeadlineAfter(duration,
+            TimeUnit.SECONDS));
     channels.put(dn.getUuid(), channel);
   }
 
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -359,6 +359,9 @@ public final class OzoneConfigKeys {
   public static final String OZONE_CLIENT_VERIFY_CHECKSUM =
       "ozone.client.verify.checksum";
   public static final boolean OZONE_CLIENT_VERIFY_CHECKSUM_DEFAULT = true;
+  public static final String OZONE_CLIENT_READ_TIMEOUT
+          = "ozone.client.read.timeout";
+  public static final String OZONE_CLIENT_READ_TIMEOUT_DEFAULT = "30s";
   public static final String OZONE_ACL_AUTHORIZER_CLASS =
       "ozone.acl.authorizer.class";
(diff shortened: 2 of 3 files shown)


adoroszlai:
> After 30s, the stub will error out even if we have other chunks to read. We rather need a deadline per chunk read call.

@lokeshj1703 I'm just curious: what's the use of reading other chunks of the block if one of them has error?

bshashikant:
> The changes look good to me. But I think setting a deadline on stub sets an absolute deadline. Since we are using the stub for reading the entire block, I think setting an absolute deadline will not work for us. After 30s, the stub will error out even if we have other chunks to read. We rather need a deadline per chunk read call.

@lokeshj1703 , even if a single read chunk call times out, we will move over to different datanode to read. I don't see the reason why we have to wait for entire block to be read

lokeshj1703:
@adoroszlai @bshashikant I am not saying we wait for the entire block to be read. The client will be reused for all the chunk reads. Lets take a scenario.
Chunk1 fails on dn1. Succeded on dn2.
Chunk2 -> we should try reading from either dn2 or dn3. We will need to mark dn1 timed out in the code. Currently we can still retry dn1.
Chunk3 -> dn2, dn3 fail. We should retry dn1 before failing the read.
For retrying dn1 we will need to reestablish the connection. We will need to add handling for that as well.

elek:
> It was verified in the actual test up where the problem originated.

Can you please share the details how was this patch tested and how the problem can be reproduced?

bshashikant:
> @adoroszlai @bshashikant I am not saying we wait for the entire block to be read. The client will be reused for all the chunk reads. Lets take a scenario.
> Chunk1 fails on dn1. Succeded on dn2.
> Chunk2 -> we should try reading from either dn2 or dn3. We will need to mark dn1 timed out in the code. Currently we can still retry dn1.
> Chunk3 -> dn2, dn3 fail. We should retry dn1 before failing the read.
> For retrying dn1 we will need to reestablish the connection. I think we are handling this.
> 
> Regarding the deadline, if we set a deadline at t=0, the stub will fail at t=30 even if the client is able to read chunks from the stub. Please verify if this is how the deadline works. I think we will need to differentiate between this scenario and scenario where chunk read is timing out.

@lokeshj1703 , in the read path, once a read op fails on a datanode, the same op never gets retried on the same datanode. 
bb --- Regarding the deadline, if we set a deadline at t=0, the stub will fail at t=30 even if the client is able to read chunks from the stub. Please verify if this is how the deadline works. I think we will need to differentiate between this scenario and scenario where chunk read is timing out.

The deadline which is set is for on the rpc call. If the rpc response doesn't complete within teh deadline , it marks it deadline exceeded. The same deadline is being used in Ratis as well. The client itself doesn't maintain a timer as such.

bshashikant:
> > It was verified in the actual test up where the problem originated.
> 
> Can you please share the details how was this patch tested and how the problem can be reproduced?

It was reproduced in fault injection testing environment. @nilotpalnandi , can you please add some details.?

lokeshj1703:
Had an offline discussion with @bshashikant. The grpc client maintains a cache for datanode so that the datanode which gives a successful reply is always used for subsequent calls.
Second, the grpc client creates a new stream for every request so the deadline in the stub works.
I have no other comments. +1.

elek:
> It was reproduced in fault injection testing environment. @nilotpalnandi , can you please add some details.?

I would appreciate it. I would like to test the patch and the behavior but couldn't do it without having the steps to reproduce the problems.

nilotpalnandi:
Tested on a cluster with a build having this patch :

steps taken :

---------------

Mounted noise injection FUSE on all datanodes.
Write a key.
In one of the container replicas of the key, Inject READ delay of 5 seconds on chunk file directory path and hdds.datanode.dir path
Run get Key operation.
Ran 100+ iteration
Operations were successful. No hung was seen.

elek:
Thanks @nilotpalnandi the information. I am trying to follow the steps but stuck at the first step:

> Mounted noise injection FUSE on all datanodes.

Is there any information about this one? It is open source?  What does it do?

nilotpalnandi:
Noise injection framework can inject slowness as well as failure on read/write path with the help of FUSE layer. The source is available at https://github.com/prashantpogde/NoiseInjector and a more detailed post will be published soon about the usage.

elek:
> more detailed post will be published soon about the usage.

Thanks the information. Looking forward to it and will check the project, too... 

If others are also interested, would be a good idea to have a demo from @prashantpogde during the Community Sync (At least I am interested...)

PR #674, opened 2020-03-13
HDDS-3175. Healthy datanodes are marked as stale

lokeshj1703:

The datanode state machine stops as the state moves to SHUTDOWN. As a result datanode stops sending heartbeats to SCM.

https://issues.apache.org/jira/browse/HDDS-3175

## How was this patch tested?

On the cluster.


avijayanhwx:
@lokeshj1703 Can you fix the checkstyle issues in this PR?

PR #676, opened 2020-03-13
HDDS-1995. Generate renewTime on OMLeader for GetDelegationToken

hanishakoneru:

Use renewTimeInterval of the OM leader across quorum of OM's.

Right now each OM generates renew time when updating token in-memory and DB using its own configured HDDS_BLOCK_TOKEN_EXPIRY_TIME.

If different OM's have different token renew interval set, for the same token we will have different renewal time across a quorum of OM's.
This Jira is to fix this issue.

https://issues.apache.org/jira/browse/HDDS-1995

## How was this patch tested?

Currently no unit tests for GetDelegationTokenRequest. Will add in the next commit.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
@@ -162,9 +162,8 @@ public class OzoneDelegationTokenSecretManager
    * @return renewTime - If updated successfully, return renewTime.
    */
   public long updateToken(Token<OzoneTokenIdentifier> token,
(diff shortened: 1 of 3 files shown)


bharatviswa504:
Test failure is unrelated to this PR.

bharatviswa504:
Thank you @hanishakoneru for the contribution.

bharatviswa504:
One more issue is for a token generated, it will use leader token renewal time. But if 3 OM's have different renewal time, so token generated by OM's, whenever leader changes will have a different expiry time.

hanishakoneru:
> One more issue is for a token generated, it will use leader token renewal time. But if 3 OM's have different renewal time, so token generated by OM's, whenever leader changes will have a different expiry time for different tokens.
Yes, but how do you propose to fix this? If the OMs have different renewal times, we cannot pick one over the other.

PR #677, opened 2020-03-13
HDDS-3174. Rebalance integration tests

adoroszlai:

1. Group `hdds` integration tests with `om` instead of `client`, because the latter takes the most time.
2. Reduce time required for `TestBlockOutputStreamWithFailures` by lowering stale/dead node interval.
3. Apply integration test profiles to `mini-chaos-tests`, too.  Previously all `it-*` runs triggered chaos tests.  Although the test itself was skipped, `@BeforeClass` still took 1 minute.

These save about 8 minutes overall.

https://issues.apache.org/jira/browse/HDDS-3174

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/506630819


adoroszlai:
Thanks @bshashikant for reviewing and merging it.

PR #678, opened 2020-03-14
HDDS-3179 Pipeline placement based on Topology does not have fallback

timmylicheng:
When rack awareness and topology is enabled, pipeline placement can fail when there is only one node on the rack.
Should add fall back logic to search for nodes from other racks.

https://issues.apache.org/jira/browse/HDDS-3179

## How was this patch tested?
UT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
@@ -99,9 +99,11 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
       try {
         pipeline = stateManager.getPipeline(pid);
       } catch (PipelineNotFoundException e) {
-        LOG.error("Pipeline not found in pipeline state manager during" +
-            " pipeline creation. PipelineID: " + pid +
-            " exception: " + e.getMessage());
+        if (LOG.isDebugEnabled()) {
+          LOG.debug("Pipeline not found in pipeline state manager during" +
+              " pipeline creation. PipelineID: " + pid +
+              " exception: " + e.getMessage());
+        }
         continue;
       }
       if (pipeline != null &&
@@ -292,10 +294,15 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
     int nodesToFind = nodesRequired - results.size();
     for (int x = 0; x < nodesToFind; x++) {
       // Pick remaining nodes based on the existence of rack awareness.
-      DatanodeDetails pick = rackAwareness
-          ? chooseNodeFromNetworkTopology(
-              nodeManager.getClusterNetworkTopologyMap(), anchor, exclude)
-          : fallBackPickNodes(healthyNodes, exclude);
+      DatanodeDetails pick;
+      try {
+        pick = rackAwareness
+            ? chooseNodeFromNetworkTopology(
+            nodeManager.getClusterNetworkTopologyMap(), anchor, exclude)
+            : fallBackPickNodes(healthyNodes, exclude);
+      } catch (SCMException e) {
+        pick = fallBackPickNodes(healthyNodes, exclude);
+      }
       if (pick != null) {
         results.add(pick);
         exclude.add(pick);
@@ -403,7 +410,7 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
   @VisibleForTesting
   protected DatanodeDetails chooseNodeFromNetworkTopology(
       NetworkTopology networkTopology, DatanodeDetails anchor,
-      List<DatanodeDetails> excludedNodes) {
+      List<DatanodeDetails> excludedNodes) throws SCMException {
     Preconditions.checkArgument(networkTopology != null);
 
     Collection<Node> excluded = new ArrayList<>();
@@ -414,6 +421,10 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
     Node pick = networkTopology.chooseRandom(
         anchor.getNetworkLocation(), excluded);
     DatanodeDetails pickedNode = (DatanodeDetails) pick;
+    if (pickedNode == null) {
+      throw new SCMException("Unable to find node based on Topology.",
+          SCMException.ResultCodes.FAILED_TO_FIND_SUITABLE_NODE);
+    }
     return pickedNode;
   }
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
@@ -408,7 +408,7 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
   @VisibleForTesting
   protected DatanodeDetails chooseNodeFromNetworkTopology(
       NetworkTopology networkTopology, DatanodeDetails anchor,
-      List<DatanodeDetails> excludedNodes) throws SCMException {
+      List<DatanodeDetails> excludedNodes) {
     Preconditions.checkArgument(networkTopology != null);
 
     Collection<Node> excluded = new ArrayList<>();
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
@@ -99,11 +99,8 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
       try {
         pipeline = stateManager.getPipeline(pid);
       } catch (PipelineNotFoundException e) {
-        if (LOG.isDebugEnabled()) {
-          LOG.debug("Pipeline not found in pipeline state manager during" +
-              " pipeline creation. PipelineID: " + pid +
-              " exception: " + e.getMessage());
-        }
+        LOG.debug("Pipeline not found in pipeline state manager during" +
+            " pipeline creation. PipelineID: {}", pid, e);
         continue;
       }
       if (pipeline != null &&
@@ -284,31 +281,32 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
         LOG.debug("Second node chosen: {}", nextNode);
       }
     } else {
-      if (LOG.isDebugEnabled()) {
-        LOG.debug("Pipeline Placement: Unable to find 2nd node on different " +
-            "rack based on rack awareness.");
-      }
+      LOG.debug("Pipeline Placement: Unable to find 2nd node on different " +
+          "rack based on rack awareness. anchor: {}", anchor);
     }
 
     // Then choose nodes close to anchor based on network topology
     int nodesToFind = nodesRequired - results.size();
     for (int x = 0; x < nodesToFind; x++) {
       // Pick remaining nodes based on the existence of rack awareness.
-      DatanodeDetails pick;
-      try {
-        pick = rackAwareness
-            ? chooseNodeFromNetworkTopology(
-            nodeManager.getClusterNetworkTopologyMap(), anchor, exclude)
-            : fallBackPickNodes(healthyNodes, exclude);
-      } catch (SCMException e) {
-        pick = fallBackPickNodes(healthyNodes, exclude);
+      DatanodeDetails pick = null;
+      if (rackAwareness) {
+        pick = chooseNodeFromNetworkTopology(
+            nodeManager.getClusterNetworkTopologyMap(), anchor, exclude);
       }
+      // fall back protection
+      if (pick == null) {
+        pick = fallBackPickNodes(healthyNodes, exclude);
+        if (rackAwareness) {
+          LOG.debug("Failed to choose node based on topology. Fallback " +
+              "picks node as: {}", pick);
+        }
+      }
+
       if (pick != null) {
         results.add(pick);
         exclude.add(pick);
-        if (LOG.isDebugEnabled()) {
-          LOG.debug("Remaining node chosen: {}", pick);
-        }
+        LOG.debug("Remaining node chosen: {}", pick);
       }
     }
 
@@ -422,8 +420,8 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
         anchor.getNetworkLocation(), excluded);
     DatanodeDetails pickedNode = (DatanodeDetails) pick;
     if (pickedNode == null) {
-      throw new SCMException("Unable to find node based on Topology.",
-          SCMException.ResultCodes.FAILED_TO_FIND_SUITABLE_NODE);
+      LOG.debug("Pick node is null, excluded nodes {}, anchor {}.",
+          excluded, anchor);
     }
     return pickedNode;
   }
--- a/hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelinePlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelinePlacementPolicy.java
@@ -21,6 +21,7 @@ package org.apache.hadoop.hdds.scm.pipeline;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
+import org.apache.hadoop.hdds.protocol.MockDatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.container.MockNodeManager;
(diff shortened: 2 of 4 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -292,10 +294,15 @@ DatanodeDetails fallBackPickNodes(
...
-          ? chooseNodeFromNetworkTopology(
-              nodeManager.getClusterNetworkTopologyMap(), anchor, exclude)
-          : fallBackPickNodes(healthyNodes, exclude);
+      DatanodeDetails pick;
+      try {
+        pick = rackAwareness
+            ? chooseNodeFromNetworkTopology(
+            nodeManager.getClusterNetworkTopologyMap(), anchor, exclude)
+            : fallBackPickNodes(healthyNodes, exclude);
+      } catch (SCMException e) {
```
linyiqun: I would like to log the exception in debug level log here. Then we could know the fallback happened.

timmylicheng: Debug log makes sense. 
My thought was: fall back will happen every time when there is only one node on the rack. Under all other scenarios, this fallback logic shouldn't take effects. As a known case, I was reluctant to add logs for it regardless of its log level. But I guess it can be valuable if fallback logic takes effect in other cases.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -99,9 +99,11 @@ boolean meetCriteria(DatanodeDetails datanodeDetails, int nodesRequired) {
       try {
         pipeline = stateManager.getPipeline(pid);
       } catch (PipelineNotFoundException e) {
-        LOG.error("Pipeline not found in pipeline state manager during" +
-            " pipeline creation. PipelineID: " + pid +
-            " exception: " + e.getMessage());
+        if (LOG.isDebugEnabled()) {
```
sodonnel: I don't think we need to wrap the debug logs in `if (LOG.isDebugEnabled())` if we use the SLF4J style logs with {} placeholders rather than string interpolation. The log message will only be evaluated if the debug level is set.

For this log, I would suggest using the {} placeholders and let the full stack trace be logged rather than just `e.getMessage()`, eg:

```
LOG.debug("Pipeline not found in pipeline state manager during" +
              " pipeline creation. PipelineID: {}", pid, e);
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -292,10 +294,19 @@ DatanodeDetails fallBackPickNodes(
...
-          : fallBackPickNodes(healthyNodes, exclude);
+      DatanodeDetails pick;
+      try {
+        pick = rackAwareness
+            ? chooseNodeFromNetworkTopology(
+            nodeManager.getClusterNetworkTopologyMap(), anchor, exclude)
+            : fallBackPickNodes(healthyNodes, exclude);
+      } catch (SCMException e) {
+        pick = fallBackPickNodes(healthyNodes, exclude);
+        if (LOG.isDebugEnabled()) {
```
sodonnel: I think we can remove this if statement safely.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -414,6 +425,14 @@ protected DatanodeDetails chooseNodeFromNetworkTopology(
     Node pick = networkTopology.chooseRandom(
         anchor.getNetworkLocation(), excluded);
     DatanodeDetails pickedNode = (DatanodeDetails) pick;
+    if (pickedNode == null) {
+      if (LOG.isDebugEnabled()) {
```
sodonnel: Again, I don't think we need the `if (LOG.isDebugEnabled()) {` wrapping the debug log.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Thanks for the info. How about the logs without "{}" ? Shall we check if debug is enabled?

sodonnel: If you don't use the {} syntax, then it makes sense to wrap the log in the if statement to avoid needing to construct the string. Ideally you should use the {} syntax for the logs, as they are easier to write, generally easier to read and more efficient.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Updated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Updated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: I added {} and removed isDebugEnabled check

sodonnel: The other logs have been changed, but this one is still using interpolation and is wrapped in the if statement. Could you change it too please?

With SLF4J, if you pass an exception object as the last parameter it will print the full stack into the log automatically, eg:
```
LOG.debug("Pipeline not found in pipeline state manager during" +
              " pipeline creation. PipelineID: {}", pid, e);
```
I don't think the full stack will get printed using string concatenation as that will call toString on the exception which will just return the message part and not the stack trace.

timmylicheng: Yea just updated. Thanks.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -99,9 +99,8 @@ boolean meetCriteria(DatanodeDetails datanodeDetails, int nodesRequired) {
       try {
         pipeline = stateManager.getPipeline(pid);
       } catch (PipelineNotFoundException e) {
-        LOG.error("Pipeline not found in pipeline state manager during" +
-            " pipeline creation. PipelineID: " + pid +
-            " exception: " + e.getMessage());
+        LOG.debug("Pipeline not found in pipeline state manager during" +
+            " pipeline creation. PipelineID: {}" + pid + e);
```
sodonnel: This line still is not correct. The way SLF4J works, is it substitutes the {} for each parameter in turn. If the last parameter is a Throwable, it prints the stack trace after the log message. So I think it should be:
```
LOG.debug("Pipeline not found in pipeline state manager during" +
            " pipeline creation. PipelineID: {}", pid, e);
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
sodonnel: Could we avoid the exception block here, especially if we think the exception block will be the "normal code path" if the rack only has one node?

Eg, if we change `chooseNodeFromNetworkTopology()` to return null or the picked node, and then refactor this code block to look like this:

```
      DatanodeDetails pick = null;
      if (rackAwareness) {
        pick = chooseNodeFromNetworkTopology(
            nodeManager.getClusterNetworkTopologyMap(), anchor, exclude);
      }
      if (pick == null) {
        pick = fallBackPickNodes(healthyNodes, exclude);
        if (rackAwareness) {
          LOG.debug("Failed to choose node based on topology. Fallback " +
              "picks node as: {}", pick);
        }
      }
```

I think that is easier to read and avoid the expensive exception handler. What do you think?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -414,6 +416,12 @@ protected DatanodeDetails chooseNodeFromNetworkTopology(
     Node pick = networkTopology.chooseRandom(
         anchor.getNetworkLocation(), excluded);
     DatanodeDetails pickedNode = (DatanodeDetails) pick;
+    if (pickedNode == null) {
+      LOG.debug("Pick node is null, excluded nodes {}, anchor {}.",
+          excluded, anchor);
+      throw new SCMException("Unable to find node based on Topology.",
+          SCMException.ResultCodes.FAILED_TO_FIND_SUITABLE_NODE);
+    }
```
sodonnel: As I mentioned above, I feel it would be simpler if this just returns null rather than throwing an exception.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelinePlacementPolicy.java
```diff
@@ -21,6 +21,7 @@
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
```
sodonnel: There are a lot of changes in this test file, but do we have a test that reproduces the problem reported? The problem is that if the anchor node is on a rack with only a single node, then only 2 nodes would be allocated rather than 3? It should be possible to reproduce this easily with a 3 node and 3 rack cluster, one rack per node.

I think we should add a test that reproduces the problem so we know the changes correct it and we have no regressions in the future. We should also test this via the public API of PipelinePlacementPolicy if possible. Right now, most of the tests in TestPipelinePolicy seem to call "@VisibleForTesting" methods rather than chooseDatanodes or getResultSet, which are the public methods.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: This is clean. I will make updates. Thanks for the suggestion.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Updated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Updated.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelinePlacementPolicy.java
timmylicheng: I just added a test for this case. Please see the updates.

sodonnel:
Acceptance is failing with:

```
Execute PI calculation                                                | FAIL |
1 != 0
Execute WordCount                                                     | FAIL |
1 != 0
ozonesecure-mr-mapreduce :: Execute MR jobs                           | FAIL |
```

This is known issue and will be fixed soon - HDDS-3284.

Integration tests are flaky. it-Freon:

```
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 395.168 s <<< FAILURE! - in org.apache.hadoop.ozone.freon.TestRandomKeyGenerator
[ERROR] bigFileThan2GB(org.apache.hadoop.ozone.freon.TestRandomKeyGenerator)  Time elapsed: 326.297 s  <<< FAILURE!
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
```

One of these issues covers it:  https://issues.apache.org/jira/browse/HDDS-3266 and it-freon: https://issues.apache.org/jira/browse/HDDS-3257

it-client I am not sure

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
```diff
@@ -403,7 +408,7 @@ private boolean checkAllNodesAreEqual(NetworkTopology topology) {
   @VisibleForTesting
   protected DatanodeDetails chooseNodeFromNetworkTopology(
       NetworkTopology networkTopology, DatanodeDetails anchor,
-      List<DatanodeDetails> excludedNodes) {
+      List<DatanodeDetails> excludedNodes) throws SCMException {
```
sodonnel: You can remove the `throws SCMException` in the method definition now, as it is no longer used.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelinePlacementPolicy.java
```diff
@@ -43,36 +47,98 @@
...
+  @Test
+  public void testChooseNodeWithSingleNodeRack() throws SCMException {
+    // There is only one node on 3 racks altogether.
+    List<DatanodeDetails> datanodes = new ArrayList<>();
+    for (Node node : SINGLE_NODE_RACK) {
+      DatanodeDetails datanode = overwriteLocationInNode(
+          MockDatanodeDetails.randomDatanodeDetails(), node);
+      datanodes.add(datanode);
+    }
+    MockNodeManager localNodeManager = new MockNodeManager(null, datanodes,
```
sodonnel: This test doesn't reproduce the error if the fix in this PR is removed. I commented out the fix in PipelinePlacement policy and added back in the old logic and ran this and it still passed. The reason, is that PipelinePlacementPolicy did not believe networkTopology was present. Changing this line as follows makes the test reproduce the problem:

```
    MockNodeManager localNodeManager = new MockNodeManager(initTopology(), datanodes,
        false, datanodes.size());
```
Note I added `initTopology()` when constructing the NodeManager.

With that, the test fails without the fix, and passes with the fix in this PR, so that is good.

sodonnel:
Thanks for the updates here. I think the code looks much cleaner now with the debug statements and refactored block in getResultSet().

There are just a couple of minor changes needed to finish this one off.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelinePlacementPolicy.java
timmylicheng: You are right. I updated this part. Thanks!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
timmylicheng: Updated.

timmylicheng:
> Thanks for the updates here. I think the code looks much cleaner now with the debug statements and refactored block in getResultSet().
> 
> There are just a couple of minor changes needed to finish this one off.

Thanks for the detailed review. It really helps. @sodonnel

sodonnel:
Thanks for quickly addressing the final issues. 

I am +1 on this now. I will commit it later pending the CI checks looking good.

sodonnel:
All tests are green except this one, which I have seen fail in several other PRs, so it is flaky:

```
 [INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 52.235 s - in org.apache.hadoop.ozone.freon.TestDataValidateWithUnsafeByteOperations
[INFO] Running org.apache.hadoop.ozone.freon.TestRandomKeyGenerator
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 338.009 s <<< FAILURE! - in org.apache.hadoop.ozone.freon.TestRandomKeyGenerator
[ERROR] bigFileThan2GB(org.apache.hadoop.ozone.freon.TestRandomKeyGenerator)  Time elapsed: 276.915 s  <<< FAILURE!
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
```

elek:
If you see a flaky test, please disable it (without and issue) + create a new open issue and repeat the build. There is some risk that one flakiness hides a other. Merging patches without green build makes harder to debug flaky tests.

PR #680, opened 2020-03-14
HDDS-3177. Periodic dependency update (Java)

adoroszlai:

1. Upgrade:
   * jackson-databind2.9.9 --> 2.10.3
   * netty-all 4.0.52 --> 4.1.47
   * nimbus-jose-jwt 4.41.1 --> 7.9
2. Use existing `bouncycastle.version` property (= 1.60)
3. Remove unused dependencies:
   * `org.apache.htrace:*`
   * `org.apache.hbase:*`
4. Remove unused/useless properties:
   * `hadoop.assemblies.version`
   * `hadoop.common.build.dir` (hadoop-common source no longer present in the same repo)
   * `hbase.*.version`
5. Exclude `ratis-examples` (accidental dependency of `ratis-tools`)

https://issues.apache.org/jira/browse/HDDS-3177
https://issues.apache.org/jira/browse/HDDS-3176

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/507994482

### DIFF
--- a/hadoop-hdds/tools/pom.xml
+++ b/hadoop-hdds/tools/pom.xml
@@ -54,6 +54,12 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <artifactId>ratis-tools</artifactId>
       <groupId>org.apache.ratis</groupId>
       <version>${ratis.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.ratis</groupId>
+          <artifactId>ratis-examples</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>commons-cli</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -1186,16 +1186,6 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
         <version>0.1.54</version>
       </dependency>
       <dependency>
-        <groupId>org.apache.htrace</groupId>
-        <artifactId>htrace-core</artifactId>
-        <version>3.1.0-incubating</version>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.htrace</groupId>
-        <artifactId>htrace-core4</artifactId>
-        <version>4.1.0-incubating</version>
-      </dependency>
-      <dependency>
         <groupId>org.jdom</groupId>
         <artifactId>jdom</artifactId>
         <version>1.1</version>
--- a/pom.xml
+++ b/pom.xml
@@ -111,15 +111,11 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <!-- Version number for xerces used by JDiff -->
     <xerces.jdiff.version>2.11.0</xerces.jdiff.version>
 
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -1791,11 +1708,6 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
           <!-- @argLine is filled by jacoco maven plugin. @{} means late evaluation -->
           <argLine>${maven-surefire-plugin.argLine} @{argLine}</argLine>
           <environmentVariables>
-            <HADOOP_COMMON_HOME>${hadoop.common.build.dir}</HADOOP_COMMON_HOME>
-            <!-- HADOOP_HOME required for tests on Windows to find winutils -->
-            <HADOOP_HOME>${hadoop.common.build.dir}</HADOOP_HOME>
-            <LD_LIBRARY_PATH>${env.LD_LIBRARY_PATH}:${project.build.directory}/native/target/usr/local/lib:${hadoop.common.build.dir}/native/target/usr/local/lib</LD_LIBRARY_PATH>
```
jojochuang: I don't know if we need to remove the native library in the tests. It's probably okay.
Other than this one I am +1.

adoroszlai: Thanks for the review @jojochuang.  This seems to have been copied from `hadoop-project/pom.xml`.  I don't think Ozone has any native library.

jojochuang: yeah. I think the assumption here is the behavior is the same with the native Hadoop library or not, which makes sense to me.

hadoop-hdds/tools/pom.xml
```diff
@@ -54,6 +54,12 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <artifactId>ratis-tools</artifactId>
       <groupId>org.apache.ratis</groupId>
       <version>${ratis.version}</version>
+      <exclusions>
```
elek: I would fix it from the Ratis side: RATIS-827

adoroszlai: I had seen that issue.  The reason I included it here as a workaround is there were warnings related to it during shading.

xiaoyuyao: Make sense to me. I've +1 on RATIS-827. We can backoff this change when updating the ratis-snapshot version.

pom.xml
xiaoyuyao: The only native library I remembered for ozone is via ratis-thirdparty jar. So I'm OK with removing it.

hadoop-hdds/tools/pom.xml
adoroszlai: We can remove it later, but I don't think it's strictly necessary.  Ozone builds fine using locally built Ratis snapshot with RATIS-827 applied.

adoroszlai:
Thanks @elek, @jojochuang and @xiaoyuyao for the review.

hadoop-hdds/tools/pom.xml
elek: Sure, I agree. We can remove the exclusion in the commit where we start to use a Ratis version which includes RATIS-827. It was just a quick reminder that we can remove it after RATIS-827.

PR #683, opened 2020-03-15
HDDS-3149. Make o3fs support set and get acl

captainzmc:

BasicOzoneFileSystem currently lacks an interface to set and get acls through HDFS. Supplement with this PR.

https://issues.apache.org/jira/browse/HDDS-3149

## How was this patch tested?

Ut has been added

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OzoneAcl.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OzoneAcl.java
@@ -21,11 +21,13 @@ package org.apache.hadoop.ozone;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.google.protobuf.ByteString;
+import org.apache.hadoop.fs.permission.AclEntry;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OzoneAclInfo;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OzoneAclInfo.OzoneAclScope;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OzoneAclInfo.OzoneAclType;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer.ACLIdentityType;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer.ACLType;
+import org.apache.hadoop.util.StringUtils;
 
 import java.util.ArrayList;
 import java.util.BitSet;
@@ -339,5 +341,38 @@ public class OzoneAcl {
   public enum AclScope {
     ACCESS,
     DEFAULT;
+  }
+
+  public String toAclString(AclEntry aclEntry) {
+    StringBuilder sb = new StringBuilder();
+
+    if (aclEntry.getType() != null) {
+      sb.append(StringUtils.toLowerCase(aclEntry.getType().toString()));
+    }
+    sb.append(':');
+    if (aclEntry.getName() != null) {
+      sb.append(aclEntry.getName());
+    }
+    sb.append(':');
+    // For example, change "r-x" to "rx"
+    if (aclEntry.getPermission() != null) {
+      sb.append(aclEntry.getPermission().SYMBOL.replace("-", ""));
+    }
+    return sb.toString();
+  }
+
+  /**
+   * Convert a List of AclEntries into a string.
+   *
+   * @param aclSpec List of AclEntries to convert
+   * @return String representation of aclSpec
+   */
+  public String aclSpecToString(List<AclEntry> aclSpec) {
+    StringBuilder buf = new StringBuilder();
+    for (AclEntry e : aclSpec) {
+      buf.append(toAclString(e));
+      buf.append(",");
+    }
+    return buf.substring(0, buf.length()-1);
   }
 }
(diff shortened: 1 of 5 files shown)


hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -322,6 +330,37 @@ public void makeQualified(FileStatus status, URI uri, Path path,
     return new IteratorAdapter(bucket.listKeys(pathKey));
   }
 
+  @Override
+  public Map<String, List<OzoneAcl>> getAcl(String keyName) throws IOException {
```
lokeshj1703, line 334: We can rename the api to getAcls and setAcls since we are dealing with multiple acls.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -703,6 +710,78 @@ public String toString() {
...
+   * @throws IOException if an ACL could not be modified
+   * @throws UnsupportedOperationException if the operation is unsupported
+   *         (default outcome).
+   */
+  @Override
+  public void setAcl(Path path, List<AclEntry> aclSpec) throws IOException {
+    adapter.setAcl(pathToKey(path), aclSpecToString(aclSpec));
+  }
+
+  public String toAclString(AclEntry aclEntry) {
```
lokeshj1703, line 752: I think we are converting AclEntry to form which can be parsed by OzoneAcl. Would it make sense to move this function to OzoneAcl class?

captainzmc:
> @captainzmc Thanks for working on this! The changes look good to me. Please find my comments below.
> 
> 1. There are few test failures. Can you please look at that?
> 2. Can you also add a unit test for the added apis in TestOzoneFileSystem?
> 3. There are other acl based apis in FileSystem class. Can you also create jiras for those?

Hi @lokeshj1703 Very thanks for your review and you suggestion is good.
1. By reopen this PR, run the UT, the second run was successful. Look at the previous error, should be caused by other places. Maybe there need another JIRA to fix ut.
2. I will consider increase ut in TestOzoneFileSystem. 
3. We've tested the BaseOzoneFile API before and found a lot of unsupported content([HDDS-3048](https://issues.apache.org/jira/browse/HDDS-3048)). Currently, we have plans to improve this part of the content(such as removeAcl and so on).

captainzmc:
Hi @lokeshj1703 I had fixed review issue and added ut. Could you please review this pr again?
    I made some changes according to your suggestion. About increase UT TestOzoneFileSystem, I found that the current o3fs doesn't support to do that. Current access to o3fs is based on volume and bucket, they need to be created in advance. And when we enable acl, the volume creation permission is denied, because o3fs API cannot currently manipulate volumes and buckets, and  permissions also cannot be set by o3fs API.  I added some UT to TestReadWriteStatistics to test this PR.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OzoneAcl.java
```diff
@@ -340,4 +342,37 @@ public OzoneAcl setAclScope(AclScope scope) {
     ACCESS,
     DEFAULT;
   }
+
+  public String toAclString(AclEntry aclEntry) {
```
lokeshj1703, line 346: Can we add documentation for these functions that String returned can be parsed using OzoneAcl#parseAcl?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -322,6 +330,38 @@ public void makeQualified(FileStatus status, URI uri, Path path,
...
+        .setBucketName(bucket.getName()).setVolumeName(volume.getName())
+        .setKeyName(keyName).setResType(OzoneObj.ResourceType.KEY)
+        .setStoreType(OZONE).build();
+    List<OzoneAcl> userResult = new ArrayList<OzoneAcl>();
+    List<OzoneAcl> groupResult = new ArrayList<OzoneAcl>();
+
+    ozoneClient.getObjectStore().getAcl(obj).forEach(ozoneAcl -> {
+      if (ozoneAcl.getType().equals(ACLIdentityType.USER)) {
+        userResult.add(ozoneAcl);
+      } else if (ozoneAcl.getType().equals(ACLIdentityType.GROUP)) {
```
lokeshj1703, line 346: Sorry for identifying this late, do we also need to check the other ACLIdentityType here like ACLIdentityType.WORLD etc.?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -703,6 +708,43 @@ public String toString() {
...
+   *
+   * @param path Path to modify
+   * @param aclSpec List describing modifications, which must include entries
+   *   for user, group for compatibility with permission bits.
+   * @throws IOException if an ACL could not be modified
+   * @throws UnsupportedOperationException if the operation is unsupported
+   *         (default outcome).
+   */
+  @Override
+  public void setAcl(Path path, List<AclEntry> aclSpec) throws IOException {
```
lokeshj1703, line 744: The AclEntry can be of type MASK, OTHER. I think we should map these into OzoneAcl as well, but dont know the correct mapping here. cc @xiaoyuyao

captainzmc:
Hi @lokeshj1703, I see this part of the implementation in ofs (hdds-2991) as well, the implementations of the two acls should be consistent, and there should be common methods on both sides. To avoid any overlap, I will temporarily close the pr until hdds-2991 is merged.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
xiaoyuyao: OzoneAcl currently does not support MASK acl. But I don't think we should put the Ozone user/group acl into owner and group string here. Instead, they should be mapped to AclStatus as List<AclEntry>.

PR #685, opened 2020-03-15
HDDS-3184. Fix MiniOzoneChaosTest to set the correct defaults.

mukul1987:

TestMiniOzoneChaosCluster is failing because of incorrect default values.

https://issues.apache.org/jira/browse/HDDS-3184

## How was this patch tested?
Fixed by running the unit tests.


adoroszlai:
> /retest

Unfortunately `/retest` doesn't really work, empty commit is created but CI not triggered.  We have to push an empty commit manually.

PR #687, opened 2020-03-16
HDDS-2184. Rename ozone scmcli to ozone admin

elek:

Originally ozone scmcli designed to be used only by the developers. A very cryptic name is chosen intentionally to frighten away the beginner users.

As we realized recently we started to use `ozone scmcli` as a generic admin tool. More and more tools has been added which are useful not only for the developers but for the administrators.

Therefore I suggest to rename `ozone scmcli` to something more meaningful.

For example to `ozone admin`

This idea has already been discussed during a community meeting but was not implemented.

https://issues.apache.org/jira/browse/HDDS-2184

## How was this patch tested?

Manually with a docker-compose based cluster and with acceptance tests.


hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/WithScmClient.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+package org.apache.hadoop.hdds.scm.cli.container;
+
+import org.apache.hadoop.hdds.scm.client.ScmClient;
+
+/**
+ * Command which provides and SCM client based on the current config.
+ */
+public interface WithScmClient {
```
dineshchitlangia: Since you renamed `scmcli` to `admin`, I feel that OzoneAdmin or Admin would be a more appropriate name for the interface. Thoughts?

hadoop-hdds/docs/content/tools/SCMCLI.md
```diff
@@ -1,27 +0,0 @@
----
```
dineshchitlangia: Could you please add a doc page for the new command?

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/WithScmClient.java
elek: Based on the function, this interface nothing more just providing an SCM client. Yes, I agree that it's used by admin commands but I am not sure if it's important here. (It can be used for any other command, eg. with Freon)

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/SafeModeCommands.java
```diff
@@ -46,15 +48,15 @@
...
-  public SCMCLI getParent() {
+  public WithScmClient getParent() {
     return parent;
   }
 
   @Override
   public Void call() throws Exception {
     throw new MissingSubcommandException(
-        this.parent.getCmd().getSubcommands().get("safemode"));
+        new CommandLine(new SafeModeCommands()));
```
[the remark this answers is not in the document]
elek: I agree with you about the usability benefit, but I deleted it intentionally. I don't like that we have a strong dependency on the parent command (getCmd() should be available) and it was not easy to keep it as from now, we have subcommands in multiple subprojects (eg. ozone-tools, hdds-tools)

Fortunately I found an other way to get the desired output: It runed out that it's enough to have a `CommandSpec` reference which can be injected with `@Spec`. I updated the subcommand, and it works again:

```
Incomplete command
Usage: ozone admin safemode [-hV] [COMMAND]
Safe mode specific operations
  -h, --help      Show this help message and exit.
  -V, --version   Print version information and exit.
```

hadoop-hdds/docs/content/tools/SCMCLI.md
elek: I deleted the old file because it didn't have any useful information. But I agree, it can be a good idea to create new documentation.

I created a short, initial version : I documented all the subcommand groups to make it clear what functionalities are available, but didn't specify all the details/flags as they can be get with `--help` easily.

elek:
Base branch is changed. Please merge it manually instead of using github UI button.

elek:
@dineshchitlangia Are you happy with the changes? Can I merge it? Your "change requested" flag is till on....

PR #689, opened 2020-03-16
HDDS-3183. Deprecate old Recon HTTP Server Keytab config key.

vivekratnavel:

- Deprecate old config: `ozone.recon.keytab.file` 
- Introduce new config: `ozone.recon.http.kerberos.keytab.file`

https://issues.apache.org/jira/browse/HDDS-3183

## How was this patch tested?

This patch was tested in a real cluster with both old config and new config.


vivekratnavel:
@avijayanhwx Please review

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServer.java
```diff
@@ -66,6 +67,9 @@ public static void main(String[] args) {
   @Override
   public Void call() throws Exception {
     OzoneConfiguration ozoneConfiguration = createOzoneConfiguration();
+    Configuration.addDeprecation(
+        ReconServerConfigKeys.OZONE_RECON_HTTP_KEYTAB_FILE_OLD,
```
avijayanhwx: Maybe we can add a method 'addConfigurationOverrides' in the ConfigurationProvider provider and move this logic there? When we have more deprecated keys in the future, it may be easier to maintain them there.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ConfigurationProvider.java
```diff
@@ -32,10 +33,17 @@
...
+    Configuration.addDeprecation(
+        ReconServerConfigKeys.OZONE_RECON_HTTP_KEYTAB_FILE_OLD,
+        ReconServerConfigKeys.OZONE_RECON_HTTP_KEYTAB_FILE);
+  }
+
   @VisibleForTesting
   public static void setConfiguration(OzoneConfiguration conf) {
     if (configuration == null) {
       ConfigurationProvider.configuration = conf;
+      ConfigurationProvider.addDeprecations();
```
xiaoyuyao: NIT: should we move this to a static block of the class to avoid issues like caller forget to invoke setConfiguration explicitly?

  static {
    addDeprecations();  
  }

vivekratnavel: Good suggestion! I have fixed it in the latest commit. Thanks!

avijayanhwx:
Thank you for the patch @vivekratnavel & the review @xiaoyuyao.

avijayanhwx:
The unit test failure is unrelated.

vivekratnavel:
Thank you @avijayanhwx and @xiaoyuyao for the reviews!

PR #691, opened 2020-03-17
HDDS-3226. Tracing Ozone Manager DB write batch operations.

xiaoyuyao:

Currently the OM DB write is asynchronously handled in OzoneManagerDoubleBuffer. But the OM response does not have traceID properly populated. As a result, we can't get insight for the DB part of the OM request handling.

This ticket is opened to add traceID properly so that the addBatch and commitBatch cost of the request can be shown up in properly in Opentracing/Jaeger.

https://issues.apache.org/jira/browse/HDDS-3226

## How was this patch tested?
docker-compose with jaeger

ozone freon ockg -n100
ozone freon rk --factor=THREE --replicationType=RATIS --numOfVolumes=10 --numOfBuckets=10 --numOfKeys=5 --keySize=512

Screen Shot 2020-03-17 at 4 40 34 PM

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
@@ -29,10 +29,10 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
-import io.opentracing.Scope;
 import io.opentracing.Span;
-import io.opentracing.util.GlobalTracer;
+import org.apache.hadoop.hdds.function.SupplierWithIOException;
 import org.apache.hadoop.hdds.tracing.TracingUtil;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMResponse;
 import org.apache.hadoop.util.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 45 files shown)


xiaoyuyao:
bq. But what do you think about tracing all the methods of DBStore instances with proxy? Would it help? (Not as a replacement of this patch but in addition of this patch).

That's a good idea. I will file a separate ticket for that.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
```diff
@@ -135,10 +138,20 @@ private void flushTransactions() {
           try(BatchOperation batchOperation = omMetadataManager.getStore()
               .initBatchOperation()) {
 
+            AtomicReference<String> lastTraceId = new AtomicReference<>();
```
bharatviswa504, line 141: AtomicReference Do we need atomic here? As this is just used in a single thread.

bharatviswa504:
I have one question in general, will this add some overhead and is there a way to disable it?
The reason is because, in non-HA, we wait for flush to DB future to be completed, so want to understand if it adds overhead, is there a way I can disable this?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
```diff
@@ -147,7 +160,11 @@ private void flushTransactions() {
             });
 
             long startTime = Time.monotonicNowNanos();
-            omMetadataManager.getStore().commitBatchOperation(batchOperation);
+            try (Scope s2 = TracingUtil.importAndCreateScope(
+                "DB-commitWriteBatch", lastTraceId.get())) {
```
bharatviswa504, line 164: One more question, here for DB commit Batch we are using last trace id, so DB Batch commit span will be linked with the last operation in the queue?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
xiaoyuyao: This is a workaround to update the lastTraceId from the lambda.  Java lambdas require the captured local variable must be final or effectively final.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
xiaoyuyao: Yes. We are effectively piggy back commitWrite to the last Op that triggered the flush. The other choice is to do another loop of the ready buffers to add it to each of the individual Ops, which will have higher cost.

bharatviswa504: Got it.

bharatviswa504:
> Overall LGTM. One minor comment and a general question.
> I have one question in general, will this add some overhead and is there a way to disable it?
> The reason is because, in non-HA, we wait for flush to DB future to be completed, so want to understand if it adds overhead, is there a way I can disable this?

How about this, is there a way to disable?

xiaoyuyao:
bq. Overall LGTM. One minor comment and a general question.
I have one question in general, will this add some overhead and is there a way to disable it?
The reason is because, in non-HA, we wait for flush to DB future to be completed, so want to understand if it adds overhead, is there a way I can disable this?

How about this, is there a way to disable?

Good point. We can make this a no-op if hdds.tracing.enabled = false.

bharatviswa504:
> > Overall LGTM. One minor comment and a general question.
> > I have one question in general, will this add some overhead and is there a way to disable it?
> > The reason is because, in non-HA, we wait for flush to DB future to be completed, so want to understand if it adds overhead, is there a way I can disable this?
> 
> > How about this, is there a way to disable?
> 
> Good point. We can make this a no-op if hdds.tracing.enabled = false.

+1 for making it a no-op if hdds.tracing.enabled = false.

xiaoyuyao:
@bharatviswa504 all the tracing logic added to doublebuffer now honor the hdds.tracing.enabled configuration, let me know if you have any additional comment.

bharatviswa504:
+1 LGTM. Thank You @xiaoyuyao for the contribution.

PR #692, opened 2020-03-18
HDDS-3168. Improve read efficiency by merging a lot of RPC call getContainerWithPipeline into one

runzhiwang:

What's the problem ?

1. If download a big file, there will be a lot of RPC call `getContainerWithPipeline` at https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L703. 
2. Use jaeger trace the download operation. Download 187MB will be splited into 24 GET requests, each request will send  24 RPC call  `getContainerWithPipeline`, the average cost of 24 GET requests is about 57 ms.
image
image

How to fix ?
1. I merge the 24 RPC call  `getContainerWithPipeline` into one by `getContainerWithPipelineBatch`
2. Use jaeger trace the download operation. The average cost of 24 GET requests is about 47 ms, improve about 17%
image
image

https://issues.apache.org/jira/browse/HDDS-3168

## How was this patch tested?

Change existed UT to test it.


runzhiwang:
@ChenSammi @xiaoyuyao Could you help review it ? Thank you very much.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -687,36 +687,49 @@ public OmKeyInfo lookupKey(OmKeyArgs args, String clientAddress)
...
+    if (scmClient.getContainerClient() == null) {
+      return;
+    }
+
+    if (value == null
+            || CollectionUtils.isEmpty(value.getKeyLocationVersions())) {
+      return;
+    }
+
+    List<Long> containerIDs = new ArrayList<>();
```
xiaoyuyao: Can we use a set instead of a list to avoid the overhead of contains call for large Key? You can convert to a list at the end if that is easier to pass over RPC.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
```diff
@@ -275,6 +275,66 @@ public ContainerWithPipeline getContainerWithPipeline(long containerID)
...
+
+        if (pipeline == null) {
+          pipeline = scm.getPipelineManager().createPipeline(
+                  HddsProtos.ReplicationType.STAND_ALONE,
+                  container.getReplicationFactor(),
+                  scm.getContainerManager()
+                          .getContainerReplicas(cid).stream()
+                          .map(ContainerReplica::getDatanodeDetails)
+                          .collect(Collectors.toList()));
+        }
```
nandakumar131: This logic can be extracted into a separate method and can be reused by `getContainerWithPipeline` and `getContainerWithPipelineBatch`

runzhiwang:
@xiaoyuyao @nandakumar131 Could you help review it again ?

xiaoyuyao:
Thanks @runzhiwang  for the update. The latest change LGTM, +1.

PR #693, opened 2020-03-18
HDDS-3220. Filesystem client should not retry on AccessControlException.

swagle:
Change retry policy on the client if the exception is of type AccessControlException.

https://issues.apache.org/jira/browse/HDDS-3220

## How was this patch tested?
Verified fix running ozone secure using docker. I will add an integration test later.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
@@ -146,6 +146,7 @@ import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.VolumeI
 import org.apache.hadoop.ozone.protocolPB.OMPBHelper;
 import org.apache.hadoop.ozone.security.OzoneTokenIdentifier;
 import org.apache.hadoop.ozone.security.acl.OzoneObj;
+import org.apache.hadoop.security.AccessControlException;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.proto.SecurityProtos.CancelDelegationTokenRequestProto;
 import org.apache.hadoop.security.proto.SecurityProtos.GetDelegationTokenRequestProto;
@@ -241,6 +242,9 @@ public final class OzoneManagerProtocolClientSideTranslatorPB
       public RetryAction shouldRetry(Exception exception, int retries,
           int failovers, boolean isIdempotentOrAtMostOnce)
           throws Exception {
+        if (isAccessControlException(exception)) {
+          return RetryAction.FAIL; // do not retry
+        }
         if (exception instanceof ServiceException) {
           OMNotLeaderException notLeaderException =
               getNotLeaderException(exception);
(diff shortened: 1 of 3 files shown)


swagle:
cc: @arp7 and @mukul1987

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -292,13 +295,30 @@ private RetryAction getRetryAction(RetryAction fallbackAction,
...
 
+  /**
+   * Unwrap exception to check if it is a {@link AccessControlException}.
+   */
+  private boolean isAccessControlException(Exception ex) {
+    if (ex instanceof ServiceException) {
+      Throwable t = ex.getCause();
+      if (t instanceof IOException) {
+        t = t.getCause();
+        if (t instanceof IOException) {
```
arp7: Wow is it wrapped that deeply? Can we make this check a bit more robust, if any of the nested exceptions is an instanceOf `AccessControlException` and also check in a loop while `getCause` returns non-nill?

swagle: null check is covered by instanceof operator which returns false if ex is null.
Making a change for checking ACE on each level.

arp7: I meant - can we keep checking for ACE until we finally get null? I am not sure if any server side changes could impact how the exception is wrapped e.g. introduce more nesting.

swagle: Oh yes, doing the same, thanks!

swagle:
> LGTM, +1.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -292,13 +295,29 @@ private RetryAction getRetryAction(RetryAction fallbackAction,
...
+  }
+  
   /**
    * Check if exception is a OMNotLeaderException.
    * @return OMNotLeaderException.
    */
   private OMNotLeaderException getNotLeaderException(Exception exception) {
     Throwable cause = exception.getCause();
-    if (cause != null && cause instanceof RemoteException) {
+    if (cause instanceof RemoteException) {
```
dineshchitlangia: @swagle Why are we skipping the null check for `cause` here and in line 338 ?

Aside from this, rest of the code change LGTM.

swagle: @dineshchitlangia  Because the null check is redundant, instanceOf operator is null safe and returns false if obj reference is null.

swagle:
Thank you @dineshchitlangia for review and merge.

PR #696, opened 2020-03-18
HDDS-3056. Allow users to list volumes they have access to, and optionally allow all users to list all volumes

smengcl:

Optionally allow non-admin users to list all volumes.

https://issues.apache.org/jira/browse/HDDS-3056

## How was this patch tested?

NOTE: The chunk below is out-dated. Will update once the patch is complete.

Tested manually in ozonesecure docker-compose.
`testuser` is an admin user (listed in `ozone.administrators`), `testuser2` is a non-admin user.

|                              | Before                          | After (ACL disabled)                                         | After (ACL enabled)                          |
| ---------------------------- | ------------------------------- | ------------------------------------------------------------ | -------------------------------------------- |
| `ozone sh volume list`       | Lists volumes **owned** by user | Lists volumes **owned** by user                              | Lists volumes that user has ACL **LIST** to. |
| `ozone sh volume list --all` | n/a                             | Lists all volumes if `ozone.om.volume.listall.allowed` is `true`; if `false`, only admins can list all volumes, non-admin will get `OmException` with `PERMISSION_DENIED`. | Same as ACL disabled.                        |

### After the patch (non-admin can list all volumes)

```bash
bash-4.2$ kinit -kt /etc/security/keytabs/testuser.keytab testuser/scm@EXAMPLE.COM
bash-4.2$ klist
Ticket cache: FILE:/tmp/krb5cc_1000
Default principal: testuser/scm@EXAMPLE.COM

Valid starting     Expires            Service principal
03/18/20 21:20:25  03/19/20 21:20:25  krbtgt/EXAMPLE.COM@EXAMPLE.COM
	renew until 03/25/20 21:20:25
bash-4.2$ ozone sh volume create vol1
bash-4.2$ ozone sh volume list
{
  "metadata" : { },
  "name" : "vol1",
  "admin" : "root",
  "owner" : "testuser/scm@EXAMPLE.COM",
...
bash-4.2$ kdestroy
bash-4.2$ kinit -kt /etc/security/keytabs/testuser2.keytab testuser2/scm@EXAMPLE.COM
bash-4.2$ ozone sh volume list --all
{
  "metadata" : { },
  "name" : "vol1",
  "admin" : "root",
  "owner" : "testuser/scm@EXAMPLE.COM",
  "creationTime" : "2020-03-18T21:20:35.370Z",
...
```

### For comparison, before the patch (non-admin can't list all volumes)

```bash
# kinit'ed as testuser2 (non-admin)
bash-4.2$ ozone sh volume list --all
PERMISSION_DENIED org.apache.hadoop.ozone.om.exceptions.OMException: Only admin users are authorized to create or list Ozone volumes.
bash-4.2$ klist
Ticket cache: FILE:/tmp/krb5cc_1000
Default principal: testuser2/scm@EXAMPLE.COM

Valid starting     Expires            Service principal
03/18/20 21:15:34  03/19/20 21:15:34  krbtgt/EXAMPLE.COM@EXAMPLE.COM
	renew until 03/25/20 21:15:34
```

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -584,7 +584,7 @@
     </description>
   </property>
   <property>
-    <name>ozone.om.user.volume.listall.allowed</name>
+    <name>ozone.om.volume.listall.allowed</name>
     <value>true</value>
     <tag>OM, MANAGEMENT</tag>
     <description>
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -584,6 +584,16 @@
     </description>
   </property>
   <property>
+    <name>ozone.om.user.volume.listall.allowed</name>
+    <value>true</value>
+    <tag>OM, MANAGEMENT</tag>
+    <description>
+      Allows everyone to list all volumes when set to true. Defaults to true.
+      When set to false, non-admin users can only list the volumes they have
+      access to. Admins can always list all volumes.
+    </description>
+  </property>
+  <property>
     <name>ozone.om.user.max.volume</name>
     <value>1024</value>
     <tag>OM, MANAGEMENT</tag>
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -76,7 +76,7 @@ public final class OMConfigKeys {
       "ozone.om.db.cache.size.mb";
   public static final int OZONE_OM_DB_CACHE_SIZE_DEFAULT = 128;
 
-  public static final String OZONE_OM_USER_VOLUME_LISTALL_ALLOWED =
+  public static final String OZONE_OM_VOLUME_LISTALL_ALLOWED =
       "ozone.om.volume.listall.allowed";
   public static final boolean OZONE_OM_USER_VOLUME_LISTALL_ALLOWED_DEFAULT =
       true;
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -76,6 +76,10 @@ public final class OMConfigKeys {
       "ozone.om.db.cache.size.mb";
   public static final int OZONE_OM_DB_CACHE_SIZE_DEFAULT = 128;
 
+  public static final String OZONE_OM_USER_VOLUME_LISTALL_ALLOWED =
+      "ozone.om.user.volume.listall.allowed";
+  public static final boolean OZONE_OM_USER_VOLUME_LISTALL_ALLOWED_DEFAULT =
+      true;
   public static final String OZONE_OM_USER_MAX_VOLUME =
       "ozone.om.user.max.volume";
   public static final int OZONE_OM_USER_MAX_VOLUME_DEFAULT = 1024;
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -78,8 +78,7 @@ public final class OMConfigKeys {
 
   public static final String OZONE_OM_VOLUME_LISTALL_ALLOWED =
       "ozone.om.volume.listall.allowed";
-  public static final boolean OZONE_OM_USER_VOLUME_LISTALL_ALLOWED_DEFAULT =
-      true;
+  public static final boolean OZONE_OM_VOLUME_LISTALL_ALLOWED_DEFAULT = true;
   public static final String OZONE_OM_USER_MAX_VOLUME =
       "ozone.om.user.max.volume";
   public static final int OZONE_OM_USER_MAX_VOLUME_DEFAULT = 1024;
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -77,7 +77,7 @@ public final class OMConfigKeys {
   public static final int OZONE_OM_DB_CACHE_SIZE_DEFAULT = 128;
 
   public static final String OZONE_OM_USER_VOLUME_LISTALL_ALLOWED =
-      "ozone.om.user.volume.listall.allowed";
+      "ozone.om.volume.listall.allowed";
   public static final boolean OZONE_OM_USER_VOLUME_LISTALL_ALLOWED_DEFAULT =
       true;
   public static final String OZONE_OM_USER_MAX_VOLUME =
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
@@ -130,10 +130,6 @@ public class TestOzoneManagerListVolumes {
     ClientProtocol proxy = objectStore.getClientProxy();
     objectStore.createVolume(volumeName);
     proxy.setVolumeOwner(volumeName, ownerName);
-//    OzoneAcl acl1 = new OzoneAcl(IAccessAuthorizer.ACLIdentityType.USER,
-//        "johndoe", IAccessAuthorizer.ACLType.ALL, ACCESS);
-//    List<OzoneAcl> aclList = new ArrayList<>();
-//    aclList.add(acl1);
     setVolumeAcl(objectStore, volumeName, aclString);
   }
 
@@ -183,14 +179,10 @@ public class TestOzoneManagerListVolumes {
       try {
         objectStore.listVolumes("volume");
         Assert.fail("listAllVolumes should fail for " + user.getUserName());
-      } catch (OMException ex) {
-        // Expect PERMISSION_DENIED if user is not admin and listall disallowed
-        if (ex.getResult() != OMException.ResultCodes.PERMISSION_DENIED) {
-          throw ex;
-        }
       } catch (RuntimeException ex) {
         // Current listAllVolumes throws RuntimeException
         if (ex.getCause() instanceof OMException) {
+          // Expect PERMISSION_DENIED
           if (((OMException) ex.getCause()).getResult() !=
               OMException.ResultCodes.PERMISSION_DENIED) {
             throw ex;
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
@@ -1,0 +1,215 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.ozone.om;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.UUID;
+
+import org.apache.hadoop.hdds.client.ReplicationFactor;
+import org.apache.hadoop.hdds.client.ReplicationType;
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
+import org.apache.hadoop.ozone.MiniOzoneCluster;
+import org.apache.hadoop.ozone.client.ObjectStore;
+import org.apache.hadoop.ozone.client.OzoneBucket;
+import org.apache.hadoop.ozone.client.OzoneClient;
+import org.apache.hadoop.ozone.client.OzoneKey;
+import org.apache.hadoop.ozone.client.OzoneVolume;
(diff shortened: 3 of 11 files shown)


smengcl:
@elek @xiaoyuyao Would you take a look? Thanks.

Essentially what this does is removing the admin check in `listAllVolumes`.

I believe this shouldn't impose security risks other than intentionally "leaking" all volume names (for now). What do you security experts think?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1870,7 +1870,6 @@ public void deleteVolume(String volume) throws IOException {
     auditMap.put(OzoneConsts.USERNAME, null);
     try {
       metrics.incNumVolumeLists();
-      checkAdmin();
```
nandakumar131: Is this change intentional?

smengcl:
> Thanks @smengcl for working on this.
> The code change looks good.
> 
> I do not have details about the conversation with Arpit/Sanjay you mentioned thus I am not merging this PR yet. As far as security is concerned, listing all volumes is a functionality suited for system admins. If I am not supposed to have access to read the contents of some volumes what benefit will I achieve by being able to list those volumes? Thus I am not able to understand what value this change brings to ozone or the user experience.

Thanks for the comment Dinesh. You concern is totally valid.

One motivation/background of this change is that in `ofs://`, all users should be able to list "root", which is essentially a `listVolumesByUser()` client call at this moment that returns only the list of volumes current user creates.

So there is also a `tmp` mount (HDDS-2929) implemented for `ofs://`, the current implementation of which is asking an admin to create a volume named `tmp` (with world `ALL` ACL) during cluster setup, so every other user will be able to use `ofs://om/tmp/` as a temp folder. While the `tmp` mount transparently points to bucket `md5(current_username)` in volume `tmp`.

Now there is a visual glitch when listing. Since the volume `tmp` is created by an admin, if another user tries `ozone fs -ls ofs://om/`, he won't see volume/mount `tmp` at all. -- I admit a better way to fix this might be to hack listing results and inject mount `tmp`. The point is this is one example of the motivation.

Yes there is this security implication behind this. I discussed with @xiaoyuyao a bit about this today.
We plan to discuss with @arp7 about this tomorrow and see if we actually want to loosen the admin check of `listAllVolumes()`.

smengcl:
Discussed with @arp7 @xiaoyuyao @dineshchitlangia . Our conclusion is that we want to allow users who have access to some volumes to be able to list them. We will add a new config param in `ozone-site.xml` to allow OM admin to control whether they want to (1) shows all volumes to all users or (2) only allow users to list volumes they have access to. This config param should default to (1).

I'll repurposing this jira. TODOs:
1. Modify the existing API(s) `listVolumesByUser`(?) to allow users who have access to some volumes to be able to list them;
2. Add new config param as mentioned above;
3. Add unit/integration tests.

smengcl:
https://github.com/apache/hadoop-ozone/pull/696/commits/4bdeafd675765712967be9bb23b297e5ee700f29 should've accomplished the "optionally allow all users to list all volumes" part. I'll add test for that later.

As for the "Allow users to list volumes they have access to" part, first I was trying to write the code in `OmMetadataManagerImpl#listAllVolumes`, but then realized it wasn't a good spot to check for ACLs since that class shouldn't really import `IAccessAuthorizer` (as used in `OzoneManager#checkAcls`).

So eventually I decide to implement the logic in `OzoneManager#listVolumeByUser`. Since the deeper level calls into `VolumeManager` and `OmMetadataManager` doesn't have access to ACL (for the reason described in the previous paragraph), the implementation lists all volumes internally. And then checks ACL for each volume with current UGI when ACL is enabled.

-- This might lead to higher memory consumption when there are a lot of volumes existing in OM. But it requires a bit more change to mitigate this. I might try to put some of the listing iterator logic in `OzoneManager` to solve this, if this is a problem.

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -583,6 +583,16 @@
       metadata to be cached in memory. This makes OM operations faster.
     </description>
   </property>
+  <property>
+    <name>ozone.om.user.volume.listall.allowed</name>
```
smengcl: Remove `.user`

smengcl: 3c9ba6bdee456217549a5b65ce1250b1394ccd96

smengcl:
Rebased onto latest master to include the acceptance test failure fix HDDS-3284.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
```diff
@@ -76,6 +76,10 @@ private OMConfigKeys() {
       "ozone.om.db.cache.size.mb";
   public static final int OZONE_OM_DB_CACHE_SIZE_DEFAULT = 128;
 
+  public static final String OZONE_OM_USER_VOLUME_LISTALL_ALLOWED =
```
smengcl: Remove `_USER`.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1819,7 +1823,30 @@ public void deleteVolume(String volume) throws IOException {
     auditMap.put(OzoneConsts.USERNAME, userName);
     try {
       metrics.incNumVolumeLists();
-      return volumeManager.listVolumes(userName, prefix, prevKey, maxKeys);
+      if (isAclEnabled) {
+        if (LOG.isDebugEnabled()) {
```
smengcl: Remove log

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1819,7 +1823,30 @@ public void deleteVolume(String volume) throws IOException {
...
+          LOG.debug("ACL is enabled. Listing volumes accessible by user. "
+                  + "Principal: {}, keytab: {}",
+              configuration.get(OZONE_OM_KERBEROS_PRINCIPAL_KEY),
+              configuration.get(OZONE_OM_KERBEROS_KEYTAB_FILE_KEY));
+        }
+        // List all volumes first
+        List<OmVolumeArgs> listOfAllVolumes = volumeManager.listVolumes(
+            null, prefix, prevKey, maxKeys);
+        // Filter all volumes by ACL LIST permission of UGI
+        return listOfAllVolumes.stream().filter(v -> v.getAclMap()
```
smengcl: Use `accessAuthorizer.checkAccess`

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1819,7 +1823,30 @@ public void deleteVolume(String volume) throws IOException {
...
+            .hasAccess(IAccessAuthorizer.ACLType.LIST, remoteUserUgi))
+            .collect(Collectors.toList());
+      } else {
+        if (LOG.isDebugEnabled()) {
```
smengcl: Remove log

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/volume/ListVolumeHandler.java
```diff
@@ -63,6 +63,10 @@
       description = "Owner of the volumes to list.")
```
smengcl, line 63: Improve the description.

smengcl:
Will add the integration test soon.
Will resolve minor conflict with the Ozone Shell refactoring.

xiaoyuyao:
Looks like this need a rebase after HDDS-2691.

smengcl:
Just added integration tests. But it seems only one cluster can be launched in one test class.
Four tests can succeed separately. But when I run them at once in IntelliJ, the second would get stuck in a loop. Any help?

smengcl:
> Just added integration tests. But it seems only one cluster can be launched in one test class.
> Four tests can succeed separately. But when I run them at once in IntelliJ, the second would get stuck in a loop. Any help?

Rebased.

smengcl:
Rebased again because of HDDS-3340 (class `ListVolumeHandler` moved).

smengcl:
I'm able to dig a bit into the root cause of the timeout.

Turns out, when a mini ozone cluster launches for a second time in the **same** test class. In `setOwner()` call the OM side would [add the same volume to owner list](https://github.com/apache/hadoop-ozone/blob/80e9f0a7238953e41b06d22f0419f04ab31d4212/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeSetOwnerRequest.java#L156-L158) for a second time and **succeed**, which is very weird. This causes `testAclDisabledListAllDisallowed` to get stuck in the `it.hasNext()` infinite loop and eventually timeout.

I am able to confirm my discovery by setting a breakpoint inside `addVolumeToOwnerList()`.

If I only run `testAclDisabledListAllDisallowed` this one test directly in IntelliJ, the test case would just pass. This makes the problem weirder. Because I do call the shutdown function in `MiniOzoneClusterImpl` to do the cleanup. And it did [delete the temp directory for the entire cluster](https://github.com/apache/hadoop-ozone/blob/e2ebbf874d5e33565b27a24a02cfb4cee6330ea1/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneClusterImpl.java#L392). This in theory should have performed the clean up work.

My questions:

1. Unless there are some other in-memory cache (`TableCache`) that is accidentally persisted across mini cluster (i.e. not fully cleaned up in `MiniOzoneClusterImpl`)? If this is the case we just need to somehow fix the test utility.

2. Or could it be the case that the [`userTable`](https://github.com/apache/hadoop-ozone/blob/876bec0130094b24472a7017fdb1fd81a65023bc/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java#L140) is flushed by mistake? In this case this would be a major bug (outside the scope of this jira) that should be fixed.

Pinging for some help @bharatviswa504 @elek

smengcl:
I just posted https://github.com/apache/hadoop-ozone/pull/806, might be related to the issue.

smengcl:
#806 is merged. will rebase this PR.

smengcl:
just rebased.

somehow all the integration tests are timing out on my machine so I can't verify locally if the problem is fixed by #806. we'll see.

smengcl:
Blocked on #821

smengcl:
Rebased one more time as #821 is merged.

smengcl:
@xiaoyuyao Pls give this PR one more look. Then we can close it if there aren't more issues since the tests are green now, finally. This blocks OFS list FS root functionality.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
```diff
@@ -0,0 +1,238 @@
...
+
+  @Test
+  public void testAclEnabledListAllAllowed() throws Exception {
+    // ozone.acl.enabled = true, ozone.om.volume.listall.allowed = true
+    MiniOzoneCluster cluster = startCluster(true, true);
+    checkUser(cluster, user1, Arrays.asList("volume1", "volume4", "volume5"),
+        true);
+    checkUser(cluster, user2, Arrays.asList("volume2", "volume3", "volume5"),
+        true);
+    stopCluster(cluster);
```
xiaoyuyao, line 203: can we wrap the stopCluster inside the try/final block in case the check failure the cluster will be stopped properly.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/VolumeManagerImpl.java
```diff
@@ -453,23 +449,11 @@ public boolean checkVolumeAccess(String volume, OzoneAclInfo userAcl)
   @Override
   public List<OmVolumeArgs> listVolumes(String userName,
       String prefix, String startKey, int maxKeys) throws IOException {
-    metadataManager.getLock().acquireLock(USER_LOCK, userName);
+    metadataManager.getLock().acquireWriteLock(USER_LOCK, userName);
```
xiaoyuyao, line 452: Do we really need a write lock for list volume operation?

smengcl: This is merely cleaning up the deprecated call.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
smengcl: Sure will do

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/VolumeManagerImpl.java
smengcl: I could either leave it alone, or change it to `acquireReadLock`. The latter should be fine though.

xiaoyuyao: I think we should change it to acquireReadLock.

smengcl: done: c9a92918ef0363ee711222688bf8d851e6321418

smengcl:
Even after reverting the lock change the acceptance test still fails.
Previously there is a passing run: https://github.com/apache/hadoop-ozone/runs/590172040
The details shows there are some s3 tests failures which are seemingly unrelated.
Rebased to latest master to hopefully resolve this test failure.

smengcl:
Acceptance failure is related to [HDDS-3456](https://issues.apache.org/jira/browse/HDDS-3456).
This PR is blocked by the above jira then.

Or we can commit since the failures are unrelated. What do you think? @xiaoyuyao

adoroszlai:
> Or we can commit since the failures are unrelated. What do you think? @xiaoyuyao

Please don't commit with unrelated failures.  These may hide other, related failures, as we've seen in the past.

Please review #844 instead. ;)

smengcl:
> > Or we can commit since the failures are unrelated. What do you think? @xiaoyuyao
> 
> Please don't commit with unrelated failures. These may hide other, related failures, as we've seen in the past.
> 
> Please review #844 instead. ;)

Sure thing. Let's wait for #844 merge. Then I'll do a rebase. :)

smengcl:
Rebased as HDDS-3456 is committed.

smengcl:
Got a green run after the rebase. Will commit shortly.

Thanks @xiaoyuyao and @dineshchitlangia for the review.

PR #700, opened 2020-03-19
HDDS-3172. Use DBStore instead of MetadataStore in SCM

elek:

Description

The MetadataStore interface provides a generic view to any key / value store with a LevelDB and RocksDB implementation.

Since the early version of MetadataStore we also go the DBStore interface which is more andvanced (it supports DB profiles and ColumnFamilies).

To simplify the introduction of new features (like versioning or rocksdb tuning) we should use the new interface everywhere instead of the old interface.

We should update SCM and Datanode to use the DBStore instead of MetadataStore.

This patch is the first part of the cleanup it starts to use DBStore in SCM.

https://issues.apache.org/jira/browse/HDDS-3172

## How was this patch tested?

Execution full integration test suite.


hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconNodeManager.java
```diff
@@ -44,6 +40,10 @@
 import org.apache.hadoop.ozone.protocol.commands.SCMCommand;
 import org.apache.hadoop.ozone.recon.ReconUtils;
 import org.apache.hadoop.util.Time;
+
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_DB_CACHE_SIZE_DEFAULT;
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_DB_CACHE_SIZE_MB;
+import static org.apache.hadoop.ozone.recon.ReconConstants.RECON_SCM_NODE_DB;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
```
avijayanhwx: ReconNodeManager also uses the MetadataStore to persist Node information. Any reason why we don't want to change this also like the rest?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconStorageContainerManagerFacade.java
```diff
@@ -92,14 +95,22 @@ public ReconStorageContainerManagerFacade(OzoneConfiguration conf,
     this.ozoneConfiguration = getReconScmConfiguration(conf);
     this.scmStorageConfig = new ReconStorageConfig(conf);
     this.clusterMap = new NetworkTopologyImpl(conf);
+    DBStore dbStore = new SCMDBDefinition().createDBStore(conf);
```
avijayanhwx: Should this not be "new ReconDBDefinition()"?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconNodeManager.java
elek: Not particular reason, just reducing the size the patch. I renamed my original Jira to fix SCM only. Recon and Datanode can be fixed in next two jira. 

(But thanks to point it to me, I was not aware of this. But it also can be moved to SCM. AFAK @Nanda had a plan to persist it on the SCM side, too...)

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconStorageContainerManagerFacade.java
elek: Good catch, thanks. Fixed.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconNodeManager.java
avijayanhwx: Sounds good, thanks!

adoroszlai:
@elek I think the latest commit (4a8721c) was intended for #578

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
```diff
@@ -0,0 +1,97 @@
...
+          PipelineID.class,
+          new PipelineIDCodec(),
+          Pipeline.class,
+          new PipelineCodec());
+
+  public static final DBColumnFamilyDefinition<Long, ContainerInfo>
+      CONTAINERS =
+      new DBColumnFamilyDefinition<>(
+          "containers",
+          Long.class,
```
nandakumar131, line 77: Can we use `ContainerID` instead of `Long` here?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/block/TestBlockManager.java
```diff
@@ -105,16 +107,25 @@ public void setUp() throws Exception {
     // Override the default Node Manager in SCM with this Mock Node Manager.
     nodeManager = new MockNodeManager(true, 10);
     eventQueue = new EventQueue();
+
+    scmMetadataStore = new SCMMetadataStoreRDBImpl(conf);
+    scmMetadataStore.start(conf);
     pipelineManager =
-        new SCMPipelineManager(conf, nodeManager, eventQueue);
+        new SCMPipelineManager(conf, nodeManager,
+            SCMDBDefinition.PIPELINES.getTable(scmMetadataStore.getStore()),
```
nandakumar131, line 115:
```suggestion
            scmMetadataStore.getPipelineTable(),
```

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/block/TestBlockManager.java
```diff
@@ -105,16 +107,25 @@ public void setUp() throws Exception {
...
+
     PipelineProvider mockRatisProvider =
         new MockRatisPipelineProvider(nodeManager,
             pipelineManager.getStateManager(), conf, eventQueue);
     pipelineManager.setPipelineProvider(HddsProtos.ReplicationType.RATIS,
         mockRatisProvider);
     SCMContainerManager containerManager =
-        new SCMContainerManager(conf, pipelineManager);
+        new SCMContainerManager(conf,
+            SCMDBDefinition.CONTAINERS.getTable(scmMetadataStore.getStore()),
```
nandakumar131, line 126:
```suggestion
            scmMetadataStore.getContainerTable(),
```

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestSCMPipelineManager.java
```diff
@@ -78,17 +80,26 @@ public void setUp() throws Exception {
...
+  public void cleanup() throws Exception {
+    store.close();
     FileUtil.fullyDelete(testDir);
   }
 
   @Test
   public void testPipelineReload() throws IOException {
     SCMPipelineManager pipelineManager =
-        new SCMPipelineManager(conf, nodeManager, new EventQueue());
+
```
nandakumar131, line 98: empty line can be removed.

elek:
Thanks @nandakumar131 the review:

> `DBDefinition` is a generic interface which should not have anything specific to particular db. It's using `DBStoreBuilder` which is specific to rocksdb.

Good point, I didn't notice it. And definition of the format and the creation logic were mixed in the new `DBDefinition` interface.

I pushed a new commit (d74528210) where all the creation logic is moved to the `DBStoreBuilder`, the `DBDefinition` is a simple DB independent interface. (In the future we can also create an interface for `DBStoreBuilder` and create a `RDBStoreBuilder` if required.)

> `SCMMetadataStoreRDBImpl` is no more a rocksdb specific implementation. We can do the rename/removal in follow-up jira.

Yes. Agree. Do you prefer rename or removal? I am not sure if it's required or not.

> I know there is a plan to refactor datanode part, is there a plan to re-write OMDBStore to follow this new model of DBDefinition?

Yes. I have two motivation:

 * Use only the new DBStore interface everywhere (which would make it easy to add upgrade/versioning support) and fully deprecate the old `Metadata` interface.

 * Re-create the debug tool (`ozone sql` If I remember well) which can list/read keys in any DBStore. It requires a unified way to create DB (same DBStore should be created from tools and OM/DN/SCM) and maintain the list of codecs for each table.

> Can we use ContainerID instead of Long here?

Not sure. Some of the code uses `long` instead of `CotainerId` (eg. `updateDeleteTransactionId` is based on long, and `ContainerInfo` is based no long). I can use `ContainerId` but it requires some temporary `ContainerId` object creation.

I created a commit in this PR to show how can it work (26c8a58ba) check the usage of `new ContainerId`. I am fine with both keeping or reverting it. It's more typesafe with `ContainerID` and the performance penalty can be very small.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/block/TestBlockManager.java
elek: Thanks, fixed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
elek: Done with notice. See my comments, below.

nandakumar131:
Thanks @elek  for updating the patch.

> Do you prefer rename or removal? I am not sure if it's required or not.

I'm leaning more towards renaming and using it, as it will give us easy to use API.
If we are going with renaming and using it, it's better to use `SCMMetadataStoreRDBImpl` in all the places rather than `DBStoreBuilder.createDBStore(conf, new SCMDBDefinition()`

Instead of this:
```
DBStore dbStore = DBStoreBuilder.createDBStore(conf, new SCMDBDefinition());
Table<PipelineID, Pipeline> pipelineTable = SCMDBDefinition.PIPELINES.getTable(dbStore);
dbStore.close();
```

we can use this:
```
SCMMetadataStore scmMetadataStore = new SCMMetadataStoreRDBImpl(conf);
Table<PipelineID, Pipeline> pipelineTable = scmMetadataStore.getPipelineTable();
scmMetadataStore.close();
```

> Re-create the debug tool (ozone sql If I remember well) which can list/read keys in any DBStore.

+1 on fixing ozone sql tool

elek:
> If we are going with renaming and using it, it's better to use SCMMetadataStoreRDBImpl in all the places rather than DBStoreBuilder.createDBStore(conf, new SCMDBDefinition()

I like the idea. Created an issue for this: https://issues.apache.org/jira/browse/HDDS-3479

Marked as newbie but if it won't be picked up by anybody in the next weeks I will do it quickly.

Thanks the review again, will merge it soon.

PR #701, opened 2020-03-20
HDDS-2691. Ozone Shell code cleanup

adoroszlai:

Reduce code duplication in Ozone Shell by:

1. moving the [common structure of subcommand logic](https://github.com/adoroszlai/hadoop-ozone/blob/d797bc1b74dbea0a0f2c1779b91cc7f648d66024/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/Handler.java#L67-L79) to `Handler#call()`
2. extracting verbose mode printing of volume/bucket/key name into [`OzoneAddress`](https://github.com/adoroszlai/hadoop-ozone/blob/d797bc1b74dbea0a0f2c1779b91cc7f648d66024/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/OzoneAddress.java#L327-L337)
3. encapsulating common options or groups of options in classes (eg. [`ListOptions`](https://github.com/adoroszlai/hadoop-ozone/blob/d797bc1b74dbea0a0f2c1779b91cc7f648d66024/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/ListOptions.java#L22-L57)) that can be mixed into [multiple](https://github.com/adoroszlai/hadoop-ozone/blob/d797bc1b74dbea0a0f2c1779b91cc7f648d66024/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/bucket/ListBucketHandler.java#L42-L43) [handlers](https://github.com/adoroszlai/hadoop-ozone/blob/d797bc1b74dbea0a0f2c1779b91cc7f648d66024/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/volume/ListVolumeHandler.java#L51-L52) in just 2 lines

Re-using `AclOption` for all ACL specifications also implies that `addacl` and `removeacl` now also accept list of ACLs (previously this was true only for `setacl`).

https://issues.apache.org/jira/browse/HDDS-2691

## How was this patch tested?

Regular CI:
https://github.com/adoroszlai/hadoop-ozone/runs/521601707

Also tested various `ozone sh` commands manually in docker-compose cluster, eg.:

```
# setup
$ ozone freon ockg -n 1 -t 1 -p warmup # easy volume/bucket creation
$ for i in {2..10}; do ozone sh bucket create vol1/bucket${i}; done

$ ozone sh --verbose key put vol1/bucket1/passwd /etc/passwd
Volume Name : vol1
Bucket Name : bucket1
Key Name : passwd
File Hash : b01f053617ddfeb782a3e757d9c08912

$ ozone sh --verbose bucket list vol1 -l 2
Volume Name : vol1
{
  "metadata" : { },
  "volumeName" : "vol1",
  "name" : "bucket1",
  "storageType" : "DISK",
  "versioning" : false,
  "creationTime" : "2020-03-20T10:04:52.049Z",
  "encryptionKeyName" : null
}
{
  "metadata" : { },
  "volumeName" : "vol1",
  "name" : "bucket10",
  "storageType" : "DISK",
  "versioning" : false,
  "creationTime" : "2020-03-20T10:06:46.591Z",
  "encryptionKeyName" : null
}
Found : 2 buckets for volume : vol1

$ ozone sh vol addacl -a user:user1:r,user:user2:l,group:ozone:a vol1
ACL user:user1:r[ACCESS] added successfully.
ACL user:user2:l[ACCESS] added successfully.
ACL group:ozone:a[ACCESS] added successfully.
```


hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/web/utils/TestJsonUtils.java
```diff
@@ -15,38 +15,33 @@
...
-            new ArrayList<>());
-
-    String result = ObjectPrinter.getObjectAsJson(volume);
-    Assert.assertTrue("Result is not a proper json",
-        result.contains("\"owner\""));
-    Assert.assertTrue("Result is not a proper json",
-        result.contains("\"1970-01-01T00:00:00Z\""));
+  private static void assertContains(String str, String part) {
+    assertTrue(str.contains(part),
+        () -> "Expected JSON to contain '" + part + "', but didn't: " + str);
```
fapifta: I am not sure what is the benefit here of using the assertTrue(boolean, Supplier<String>) method, why not simply use the good old org.junit.Assert.assertTrue("Expected JSON to contain '" + part + "', but didn't: " + str, str.contains()) approach? Not a big deal, just I would like to understand if there is a benefit, as for me the usage of the lamda is not 100% helpful to read the code.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/Handler.java
```diff
@@ -54,4 +58,55 @@ public OzoneConfiguration createOzoneConfiguration() {
...
+
+  @Override
+  public Void call() throws Exception {
+    conf = createOzoneConfiguration();
+
+    OzoneAddress address = getAddress();
+    try (OzoneClient client = createClient(address)) {
+      if (isVerbose()) {
+        address.print(out());
+      }
```
fapifta: Just an idea, but aren't this few lines better placed inside createClient() to have this code clean from logging stuff?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/acl/AclHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.apache.hadoop.ozone.web.ozShell.StoreTypeOption;
+import picocli.CommandLine;
+
+import java.io.IOException;
+
+/**
+ * Base class for ACL-related commands.
+ */
+public abstract class AclHandler extends Handler {
```
fapifta: Based on child classes this class might implement the getAddress() method, as all of the descendants as I see implements that method exactly the same way. (returning address.getValue())

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/acl/AclOption.java
```diff
@@ -0,0 +1,91 @@
...
+          "d = DELETE, " +
+          "l = LIST, " +
+          "a = ALL, " +
+          "n = NONE, " +
+          "x = READ_ACL, " +
+          "y = WRITE_ACL.")
+  private OzoneAcl[] values;
+
+  private List<OzoneAcl> getAclList() {
+    return ImmutableList.copyOf(values);
```
fapifta: Do we need to wrap the array into an ImutableList here? I understand the intent I think, but as this is an internal variable visible only for the descendants, and is once used, I don't feel the need to make it immutable, especially as the usages are just iterating through the value. Why do you think we need to make this immutable?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/token/PrintTokenHandler.java
```diff
@@ -18,54 +18,31 @@
...
 
 /**
- * Executes getDelegationToken api.
+ * Prints token decoded from file.
  */
 @Command(name = "print",
     description = "print a delegation token.")
-public class PrintTokenHandler extends Handler {
+@SuppressWarnings("squid:S106") // CLI
+public class PrintTokenHandler implements Callable<Void> {
```
fapifta: Am I understand well that this one is on purpose not a Handler implementation, as it is dealing with a local client side thing, and there is no need for the facilities that are there to connect to Ozone server side?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/token/TokenCommands.java
```diff
@@ -61,4 +61,5 @@ public boolean isVerbose() {
   public OzoneConfiguration createOzoneConfiguration() {
     return shell.createOzoneConfiguration();
   }
+
```
fapifta: We probably do not need this newline here.

fapifta:
Hi Attila, thank you for the great work and simplification done here.

The changes looks great, in general I do not see any issues with it, though I left some comments inline to discuss a few questions, and suggest some changes. All of them are minor stuff, except the one about the getAddress method in ACLHandler implementations which should be pulled up to the base abstract ACLHandler class as all of them are the same effectively duplicating code.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/acl/AclHandler.java
adoroszlai: They differ in the type of `address`, which can be `VolumeUri`, `BucketUri`, or `KeyUri`.

`AclHandler` subclasses mix two aspects: operation (get/set/etc.) and target (volume/bucket/key).  I kept the hierarchy shallow and extracted logic into `*Uri` and `AclOption`, so that duplicated implementations are reduced to one-liners.

We could introduce additional intermediate level classes into the hierarchy either by operation or target, but unfortunately the other aspect would still be duplicated.  I didn't find it worth introducing a new class to extract a one-liner method out of 4 classes.  (`extends ...` would still be duplicated ;) )

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/Handler.java
adoroszlai: `createClient()` is overridden in `S3Handler`, so that would introduce some duplication.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/acl/AclOption.java
adoroszlai: `List` is needed because it's also passed to `ObjectStore.setAcl`.  We could use `Arrays.asList`, but I prefer immutables.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/token/PrintTokenHandler.java
adoroszlai: Correct.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/token/TokenCommands.java
adoroszlai: Sure.  I'll remove it if I need to change the patch for some other reason.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/acl/AclOption.java
fapifta: Ah, I haven't noticed that part, thank you for the clarification.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/token/PrintTokenHandler.java
fapifta: great, thank you for clarifying

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/token/TokenCommands.java
fapifta: +1 thanks

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/Handler.java
fapifta: fair point

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/acl/AclHandler.java
fapifta: Hmm, again you caught me with the different types... :)
Ok, so VolumeUri, BucketUri, and KeyUri implements "CommandLine.ITypeConverter<OzoneAddress>" this ensures the convert method from string to address which as I know is handled by PicoCLI.
If we introduce a common interface or any type of parent to deal with these URI implementations in the handlers then we can get around the multiple type problem in the ACLHandler, though I agree that does not worth it and probably does not help with understandability either. +1 to leave it this way, thank you for helping to understand it.

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/web/utils/TestJsonUtils.java
adoroszlai: The benefit is very tiny: avoid eager string concatenation.  It's still good JUnit, only they reversed the order of parameters in 5.  I can remove the `() -> ` if you wish.

fapifta: Not necessary at all :) Thank you for clarifying though. I was just curious as I do not see any real benefit, but yeah, if it is only the evaluation time of the string concatenation, then that really is not much in a JUnit test.
Also if we remove the `() ->` part we loose even this advantage ;)

fapifta:
After discussing the getAddress method in detail, the patch looks good to me, I am curious what changes are you planning, but I will check back when ready.

adoroszlai:
> I am especially grateful for this part:
> 
> ```
> -package org.apache.hadoop.ozone.web.ozShell;
> +package org.apache.hadoop.ozone.web.utils;
> ```

If you mean the non-conformant `ozShell` package name: only one class is moved for now, to avoid unnecessarily large diffs.  Full package rename can be done separately.

PR #704, opened 2020-03-23
HDDS-3248. shutdown defaultMetricsSystem before tests

esahekmat:
clean up DefaultMetricSystem after TestHandler and TestHddsDispatcher

https://issues.apache.org/jira/browse/HDDS-3248

### DIFF
--- a/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/impl/TestHddsDispatcher.java
+++ b/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/impl/TestHddsDispatcher.java
@@ -37,7 +37,6 @@ import org.apache.hadoop.hdds.protocol.datanode.proto.ContainerProtos
 import org.apache.hadoop.hdds.protocol.proto
     .StorageContainerDatanodeProtocolProtos.ContainerAction;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerDatanodeProtocolProtos.ContainerReplicaProto;
-import org.apache.hadoop.metrics2.lib.DefaultMetricsSystem;
 import org.apache.hadoop.ozone.common.Checksum;
 import org.apache.hadoop.ozone.container.common.helpers.ContainerMetrics;
 import org.apache.hadoop.ozone.container.common.interfaces.Container;
@@ -54,7 +53,6 @@ import org.apache.hadoop.test.GenericTestUtils;
 
 import org.apache.ratis.thirdparty.com.google.protobuf.ByteString;
 import org.junit.Assert;
-import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
 
@@ -79,11 +77,6 @@ public class TestHddsDispatcher {
 
   public static final Consumer<ContainerReplicaProto> NO_OP_ICR_SENDER =
       c -> {};
-
-  @Before
-  public void setup() {
-    DefaultMetricsSystem.instance().shutdown();
-  }
 
   @Test
   public void testContainerCloseActionWhenFull() throws IOException {
@@ -139,6 +132,7 @@ public class TestHddsDispatcher {
 
     } finally {
       volumeSet.shutdown();
+      ContainerMetrics.remove();
       FileUtils.deleteDirectory(new File(testDir));
     }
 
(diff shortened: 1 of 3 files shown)


fapifta:
Thank you for figuring this out, and posting a JIRA about it.

As in #705 I disagree with how the solution is implemented.
The main problem with the approach is that we are fixing a code which is not broken because some other code does not do the necessary cleanup...

(also the PR template is missing here as well, would be good to include it)

esahekmat:
Hi, you're right. we should clean up after tests, not before them. and I change this PR to use this approach. (I also edit the PR description to follow PR template)

esahekmat:
I don't understand why it-freon failed, I ran this command in my environment and it finished successfully
```
./hadoop-ozone/dev-support/checks/integration.sh -Pfreon
...
[INFO] Reactor Summary for Apache Hadoop Ozone Integration Tests 0.5.0-SNAPSHOT:
[INFO] 
[INFO] Apache Hadoop Ozone Integration Tests .............. SUCCESS [07:50 min]
[INFO] Apache Hadoop Ozone Mini Ozone Chaos Tests ......... SUCCESS [  1.363 s]
[INFO] BUILD SUCCESS

```

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/impl/TestHddsDispatcher.java
```diff
@@ -132,6 +132,7 @@ public void testContainerCloseActionWhenFull() throws IOException {
 
     } finally {
       volumeSet.shutdown();
+      ContainerMetrics.remove();
```
fapifta: Should we call shutdownDispatcher() here as well instead of just ContainerMetrics.remove()?

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/impl/TestHddsDispatcher.java
```diff
@@ -276,6 +283,13 @@ private HddsDispatcher createDispatcher(DatanodeDetails dd, UUID scmId,
     return hddsDispatcher;
   }
 
+  private void shutdownDispatcher(HddsDispatcher dispatcher){
+    if(dispatcher != null) {
+      dispatcher.shutdown();
+    }
+    ContainerMetrics.remove();
```
fapifta: I am not convinced whether this step to remove the ContainerMetrics belongs to a method called shutdownDispatcher, at least from the name it is not clear it will do so besides shutting down the dispatcher. Shouldn't we separate this two?

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/interfaces/TestHandler.java
```diff
@@ -58,6 +60,7 @@
 
   @Before
   public void setup() throws Exception {
+    DefaultMetricsSystem.instance().shutdown();
```
fapifta: Do we still need this one after the added cleanup in TestHddsDispatcher?

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/volume/TestHddsVolumeChecker.java
```diff
@@ -91,6 +92,7 @@
 
   @Before
   public void setup() throws IOException {
+    DefaultMetricsSystem.instance().shutdown();
```
fapifta: Do we still need this one after the added cleanup in TestHddsDispatcher?

fapifta:
Thank you for the follow up, and addressing my concern, I have added a few inline comments, can you please check to the questions there?

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/interfaces/TestHandler.java
esahekmat: My mistake, I will remove this

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/volume/TestHddsVolumeChecker.java
esahekmat: My mistake, I will remove this

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/impl/TestHddsDispatcher.java
esahekmat: actually dispatcher.shutdown() method do nothing, and I put it here to follow the principle of shutdown used objects. 
I will edit this part to just call ContainerMetrics.remove()

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/impl/TestHddsDispatcher.java
esahekmat: I removed that method and unify all the calls to ContainerMetrics.remove()

fapifta:
The changeset this way looks good to me, thank you for addressing the concerns, if it runs fine I am +1 on it (non-binding)

esahekmat:
Thank you for guiding me through this. :)

PR #708, opened 2020-03-23
HDDS-3239. Provide message-level metrics from the generic protocol dispatch.

elek:

Ozone RPC protocols (due to the limitation of protobuf) are very simple: There is only one message per service and the message is routed to the appropriate method based on the type.

It makes very easy to add tracing / metrics as we have one generic dispatcher (OzoneProtocolMessageDispatcher.java) where we implemented opentracing support and basic metric collection.

In this patch I propose to improve this metric to add the message type to the metrics as a tag to make it easier to follow what's going on... (eg. see the distribution of the incoming message types)

We can also measure the total amount of time spent to serve one specific type of requests which can be used to calculate the moving average for latency.

https://issues.apache.org/jira/browse/HDDS-3239

## How was this patch tested?

Deployed to Kubernetes and checked the prometheus metrics.


elek:
Base branch is changed. Please don't merge it from the github web ui, only manually.

PR #709, opened 2020-03-23
HDDS-3244. Improve write efficiency by opening RocksDB only once

runzhiwang:

**What's the problem ?**

1. when datanode create a new container, a new RocksDB instance will be 
[created](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/helpers/KeyValueContainerUtil.java#L76) in `HddsDispatcher.WriteChunk` , but then the created RocksDB was  [closed](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/helpers/KeyValueContainerUtil.java#L83), until `HddsDispatcher.PutBlock` the RocsDB will be [opend](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java#L123) again and then put it into cache, so the RocksDB was open twice in each datanode.  Becase RocksDB.open cost about 200ms, so open it twice is a waste.
 

**How to fix it ?**

1. Put the RocksDB handler into cache when open it the first time , to avoid open it again in `HddsDispatcher.PutBlock`.

https://issues.apache.org/jira/projects/HDDS/issues/HDDS-3244

## How was this patch tested?

Existed UT and IT.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/helpers/KeyValueContainerUtil.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/helpers/KeyValueContainerUtil.java
@@ -32,6 +32,8 @@ import org.apache.hadoop.ozone.container.common.helpers.ContainerUtils;
 import org.apache.hadoop.ozone.container.common.helpers.BlockData;
 import org.apache.hadoop.ozone.container.keyvalue.KeyValueContainerData;
 import org.apache.hadoop.hdds.utils.MetadataKeyFilters;
+import org.apache.hadoop.hdds.utils.MetadataStore;
+import org.apache.hadoop.hdds.utils.MetadataStoreBuilder;
 
 import com.google.common.base.Preconditions;
 import org.apache.commons.io.FileUtils;
@@ -71,6 +73,13 @@ public final class KeyValueContainerUtil {
       throw new IOException("Unable to create directory for metadata storage." +
           " Path: " + containerMetaDataPath);
     }
+
+    MetadataStore store = MetadataStoreBuilder.newBuilder().setConf(conf)
+        .setCreateIfMissing(true).setDbFile(dbFile).build();
+    ReferenceCountedDB db =
+        new ReferenceCountedDB(store, dbFile.getAbsolutePath());
+    //add db handler into cache
+    BlockUtils.addDB(db, dbFile.getAbsolutePath(), conf);
 
     if (!chunksPath.mkdirs()) {
       LOG.error("Unable to create chunks directory Container {}",
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/helpers/KeyValueContainerUtil.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/helpers/KeyValueContainerUtil.java
@@ -32,8 +32,6 @@ import org.apache.hadoop.ozone.container.common.helpers.ContainerUtils;
 import org.apache.hadoop.ozone.container.common.helpers.BlockData;
 import org.apache.hadoop.ozone.container.keyvalue.KeyValueContainerData;
 import org.apache.hadoop.hdds.utils.MetadataKeyFilters;
-import org.apache.hadoop.hdds.utils.MetadataStore;
-import org.apache.hadoop.hdds.utils.MetadataStoreBuilder;
 
 import com.google.common.base.Preconditions;
 import org.apache.commons.io.FileUtils;
@@ -73,14 +71,6 @@ public final class KeyValueContainerUtil {
       throw new IOException("Unable to create directory for metadata storage." +
           " Path: " + containerMetaDataPath);
     }
-    MetadataStore store = MetadataStoreBuilder.newBuilder().setConf(conf)
-        .setCreateIfMissing(true).setDbFile(dbFile).build();
-
-    // we close since the SCM pre-creates containers.
-    // we will open and put Db handle into a cache when keys are being created
-    // in a container.
-
-    store.close();
 
     if (!chunksPath.mkdirs()) {
       LOG.error("Unable to create chunks directory Container {}",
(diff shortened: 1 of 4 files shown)


runzhiwang:
@bharatviswa504 Could you help review this patch ?  If you agree with the change, I will fix the CI failure. Thank you very much.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/helpers/KeyValueContainerUtil.java
```diff
@@ -73,14 +71,6 @@ public static void createContainerMetaData(File containerMetaDataPath, File
       throw new IOException("Unable to create directory for metadata storage." +
           " Path: " + containerMetaDataPath);
     }
-    MetadataStore store = MetadataStoreBuilder.newBuilder().setConf(conf)
-        .setCreateIfMissing(true).setDbFile(dbFile).build();
```
mukul1987: Thanks for working on this @runzhiwang. Can we add to the cache here ?
Also the entry should be added to cache here and not in the apply transaction phase.

runzhiwang: @mukul1987 This is in HddsDispatcher.WriteData, if add to the cache here, 2 follower datanodes will not execute RocksDB.open in their HddsDispatcher.WriteData until the leader datanode finish HddsDispatcher.WriteData including RocksDB.open. So RocksDB.open in sequential.  If in apply transaction phase, RocksDB.open in parallel, it save 200ms. I think RocksDB was not used until HddsDispatcher.PutBlock, so add to cache in apply transaction phase does not matter. Please fix me if I am wrong.

bshashikant: @runzhiwang , createContainer happens as a part of writeStateMachineData which gets executed on all nodes(leader and follower) independent of each other. I would prefer to add it to the container cache as soon as it gets created on the datanode on disk.

runzhiwang: @bshashikant Hi, but it looks like they are not independent of each other, follower must wait leader finish createContainer.

bshashikant: > @bshashikant Hi, but it looks like they are not independent of each other, follower must wait leader finish createContainer.

Can we point me to the code where you see this happening or we can have a call to discuss this further offline?

runzhiwang: @bshashikant @mukul1987 
1. From the code,  the [future thread](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L672) do `getCachedStateMachineData ` in `readStateMachineData `and the [future thread](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L459) do `createContainer` in `writeStateMachineData `are the same [thread](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L505). Because `writeStateMachineData  `called before `readStateMachineData`. So leader must wait `createContainer `finish then `getCachedStateMachineData `and append logs to the follower, so leader and follower are not independent, follower must wait leader finish `createContainer`.  I think this order can be improved by distinguishing the thread used by `getCachedStateMachineData `  and `createContainer `. What do you think ?
2. From the jaeger UI, you can see follower create container after leader finishing it.
image

bshashikant: Thanks @runzhiwang . readStateMachine getting invoked after writeStateMachineData completes, is happening bcoz of recent change. i agree with ur proposal here.

runzhiwang: @bshashikant Thanks, I will open another PR to fix the dependent of leader and follower.
And do you think where we should put the RocksDB into cache ? Here or CommitData i.e. apply transaction phase. If place it here, once cache miss, follower must wait leader, the wait time including RocksDB.open which cost 200ms. If place it in CommitData, once cache miss, follower need not wait so long.

runzhiwang:
@bharatviswa504 Could you help review it again.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/helpers/KeyValueContainerUtil.java
```diff
@@ -73,14 +73,13 @@ public static void createContainerMetaData(File containerMetaDataPath, File
...
-
-    // we close since the SCM pre-creates containers.
-    // we will open and put Db handle into a cache when keys are being created
-    // in a container.
-
-    store.close();
+    ReferenceCountedDB db =
+        new ReferenceCountedDB(store, dbFile.getAbsolutePath());
+    //add db handler into cache
+    BlockUtils.addDB(db, dbFile.getAbsolutePath(), conf);
```
bshashikant, line 82: I think its better to update the cache only at the end of function "createContainerMetaData" so that even if the chunk dir creation fails, we it won't be added to the container cache.

PR #711, opened 2020-03-23
HDDS-3250. Create a separate log file for Warnings and Errors in MiniOzoneChaosCluster.

mukul1987:
Adding a new log file for just the error while running MiniOzoneChaosCluster

https://issues.apache.org/jira/browse/HDDS-3250

## How was this patch tested?
Ran MiniOzoneChaosCluster

### DIFF
--- a/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/resources/log4j.properties
+++ b/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/resources/log4j.properties
@@ -11,7 +11,7 @@
 #   limitations under the License.
 # log4j configuration used during build and unit tests
 
-log4j.rootLogger=INFO,stdout,TEST
+log4j.rootLogger=INFO,stdout,PROBLEM
 log4j.threshold=ALL
 log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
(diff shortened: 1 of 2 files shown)


fapifta:
The changes looks good to me, one thing that we may consider is naming as the errorfilename covers something which contains warns errors and fatals which was surprising for the first sight, but thinking it through it is reasonable, and I don't have a better name in the current context :)

hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/resources/log4j.properties
```diff
@@ -28,4 +28,11 @@ log4j.appender.CHAOS.File=${chaoslogfilename}
 log4j.appender.CHAOS=org.apache.log4j.FileAppender
 log4j.appender.CHAOS.layout=org.apache.log4j.PatternLayout
 log4j.appender.CHAOS.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} (%F:%M(%L)) - %m%n
+
+log4j.appender.TEST.File=${errorlogfilename}
```
adoroszlai: Nit: I think we could find a better name to describe this appender, eg. `errors`, `problems` or something along that line.

mukul1987: great suggestion, thanks

PR #712, opened 2020-03-23
HDDS-3243. Recon should not have the ability to send Create/Close Con…

avijayanhwx:
…tainer commands to Datanode.

Recon should not be able to send Create/Close containers commands to the Datanode.

https://issues.apache.org/jira/browse/HDDS-3243

## How was this patch tested?
Added integration test case.


avijayanhwx:
Thanks for the review @vivekratnavel & @bharatviswa504.

PR #265, opened 2019-11-22
HDDS-2621. Enable OM HA acceptance tests.

hanishakoneru:
## What changes were proposed in this pull request?

OM HA robot tests were disable in HDDS-2533 as they were failing intermittently. HDDS-2454 fixes some issues in the HA tests. Creating this Jira so as to run re-enable the HA acceptance tests.

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-2621

## How was this patch tested?

CI acceptance test suit.

PR #718, opened 2020-03-24
HDDS-3224. Enforce volume and bucket name rule at create time.

swagle:
Validations performed by RpcClient are not done when using OzoneManager client directly. Added resource name checks for volume and bucket.

https://issues.apache.org/jira/browse/HDDS-3224

## How was this patch tested?
Added an integration test.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -57,6 +57,7 @@ import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.protocol.proto.SCMSecurityProtocolProtos.SCMGetCertResponseProto;
 import org.apache.hadoop.hdds.protocolPB.SCMSecurityProtocolClientSideTranslatorPB;
+import org.apache.hadoop.hdds.scm.client.HddsClientUtils;
 import org.apache.hadoop.hdds.server.http.RatisDropwizardExports;
 import org.apache.hadoop.hdds.utils.HddsServerUtil;
 import org.apache.hadoop.hdds.scm.ScmInfo;
@@ -1560,8 +1561,9 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
   @Override
   public void createVolume(OmVolumeArgs args) throws IOException {
     try {
-      metrics.incNumVolumeCreates();
       checkAdmin();
+      HddsClientUtils.verifyResourceName(args.getVolume());
+      metrics.incNumVolumeCreates();
       volumeManager.createVolume(args);
       AUDIT.logWriteSuccess(buildAuditMessageForSuccess(OMAction.CREATE_VOLUME,
           (args == null) ? null : args.toAuditMap()));
@@ -1883,7 +1885,7 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
   }
 
   private void checkAdmin() throws OMException {
-    if(isAclEnabled) {
+    if (isAclEnabled) {
       if (!ozAdmins.contains(OZONE_ADMINISTRATORS_WILDCARD) &&
           !ozAdmins.contains(ProtobufRpcEngine.Server.getRemoteUser()
               .getUserName())) {
@@ -1905,10 +1907,12 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
   @Override
   public void createBucket(OmBucketInfo bucketInfo) throws IOException {
     try {
-      if(isAclEnabled) {
+      if (isAclEnabled) {
         checkAcls(ResourceType.VOLUME, StoreType.OZONE, ACLType.CREATE,
             bucketInfo.getVolumeName(), bucketInfo.getBucketName(), null);
       }
+      HddsClientUtils.verifyResourceName(bucketInfo.getVolumeName(),
+          bucketInfo.getBucketName());
       metrics.incNumBucketCreates();
       bucketManager.createBucket(bucketInfo);
       AUDIT.logWriteSuccess(buildAuditMessageForSuccess(OMAction.CREATE_BUCKET,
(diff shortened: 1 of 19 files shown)


swagle:
cc: @ChenSammi / @bharatviswa504 for review.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1894,10 +1896,12 @@ private void checkAdmin() throws OMException {
   @Override
   public void createBucket(OmBucketInfo bucketInfo) throws IOException {
     try {
-      if(isAclEnabled) {
+      if (isAclEnabled) {
         checkAcls(ResourceType.VOLUME, StoreType.OZONE, ACLType.CREATE,
             bucketInfo.getVolumeName(), bucketInfo.getBucketName(), null);
       }
+      HddsClientUtils.verifyResourceName(bucketInfo.getVolumeName(),
```
bharatviswa504: For write requests, this code is no more used.
These changes need to bbe done in OMBucketcreateRequest#preExecute and OMVolumeCreateRequest#preExecute and the same thing needs to be done in S3BucketCreateRequest

bharatviswa504: And if illegal, verifyResourceName throws IllegalArgumentException, here we need to catch this, and return proper error code to the client here.

swagle:
> I have a few comments.

Thanks, Bharat for your comments. I will make those changes requested but one point, like Sammi pointed out in the Jira, this code path is still possible.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
swagle: Added the checks at suggested sites.

bharatviswa504:
> > I have a few comments.
> 
> Thanks, Bharat for your comments. I will make those changes requested but one point, like Sammi pointed out in the Jira, this code path is still possible.

When using clientsideTranslatorPB, it will reach to OzoneManagerProtocolServerSideTranslator and then it will use new request code. I see only way, old code can be used is by using direct OzoneManager instance, but I see no real use case in doing that.

swagle:
> > > I have a few comments.
> > 
> > 
> > Thanks, Bharat for your comments. I will make those changes requested but one point, like Sammi pointed out in the Jira, this code path is still possible.
> 
> When using clientsideTranslatorPB, it will reach to OzoneManagerProtocolServerSideTranslator and then it will use new request code. I see only way, old code can be used is by using direct OzoneManager instance, but I see no real use case in doing that.

I reverted the changes.

swagle:
S3BucketCreateRequest checked for length but the verifyResourceName is more comprehensive.
We are still doing length check in S3BucketDeleteRequest, technically we should not need validation in the delete path but kept those changes asis.

ChenSammi:
@swagle, thanks for working on the fix.  I tried the patch locally with non-HA OM. A volume with name "11" still can be created using Nanda's tool. The volume and bucket name check is in RpcClient.java while Nanda's tool use OzoneManagerProtocolClientSideTranslatorPB directly which bypass the check.  Can we move these checks to server side as HDFS does? so that we just need one place to verify volume and bucket name integrity.

swagle:
> @swagle, thanks for working on the fix. I tried the patch locally with non-HA OM. A volume with name "11" still can be created using Nanda's tool. The volume and bucket name check is in RpcClient.java while Nanda's tool use OzoneManagerProtocolClientSideTranslatorPB directly which bypass the check. Can we move these checks to server-side as HDFS does? so that we just need one place to verify volume and bucket name integrity.

Hi Sammy, that is correct, it would be possible since I modified the patch based on @bharatviswa504  comments. Do you know where is the source for the tool?

swagle:
> > @swagle, thanks for working on the fix. I tried the patch locally with non-HA OM. A volume with name "11" still can be created using Nanda's tool. The volume and bucket name check is in RpcClient.java while Nanda's tool use OzoneManagerProtocolClientSideTranslatorPB directly which bypass the check. Can we move these checks to server-side as HDFS does? so that we just need one place to verify volume and bucket name integrity.
> 
> Hi Sammy, that is correct, it would be possible since I modified the patch based on @bharatviswa504 comments. Do you know where is the source for the tool?

Actually, I am not able to verify by writing simple test, it fails as expected:

`@Test
  public void testInvalidVolumeBucketNamesThrowException() throws Exception {
    OmVolumeArgs volumeArgs = OmVolumeArgs.newBuilder()
        .setVolume("v1")
        .setOwnerName("owner1")
        .setAdminName("admin1")
        .build();
    RPC.setProtocolEngine(conf, OzoneManagerProtocolPB.class,
        ProtobufRpcEngine.class);
    OzoneManagerProtocol client =
        new OzoneManagerProtocolClientSideTranslatorPB(
        RPC.getProxy(OzoneManagerProtocolPB.class,
            RPC.getProtocolVersion(OzoneManagerProtocolPB.class),
            OmUtils.getOmAddressForClients(conf),
            UserGroupInformation.getCurrentUser(), conf,
            NetUtils.getDefaultSocketFactory(conf),
            Client.getRpcTimeout(conf)),
        "Ozone Manager Perf Test");
    client.createVolume(volumeArgs);
  }`

Above failed with exception:
`INVALID_VOLUME_NAME org.apache.hadoop.ozone.om.exceptions.OMException: Invalid volume name: v1`

Could it be something not updated I your test setup?

swagle:
@ChenSammi wanted to get your input on above, thanks.

swagle:
> Changes look good @swagle, however, there are multiple acceptance test failures for OM HA and Freon. Could you please review that?

@dineshchitlangia The same test seems to fail every time: ozone-om-ha-s3, I see HDDS-3313 is opened to address this.

adoroszlai:
> The same test seems to fail every time: ozone-om-ha-s3, I see HDDS-3313 is opened to address this.

HDDS-3313 is only about `ozone-om-ha` (enabled in #265 and now disabled in ce6ad30a3939e3b211c9eede197d240cf26e2391), not `ozone-om-ha-s3`.  The latter has an intermittent, but very rare failure, reported in HDDS-2934.

adoroszlai:
OM bucket generator fails even in single OM environment:

```
Freon OM Bucket Generator                                             | FAIL |
255 != 0
```

due to:

```
OMException: Invalid bucket name: ELAtCj6bFe7
```

Can fix it by converting random Freon prefix to lowercase:

https://github.com/apache/hadoop-ozone/blob/61b6be001b78d1f3902db4a3a40734c19a90459c/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java#L240

swagle:
> HDDS-2934

I am actually getting this failure on

swagle:
> OM bucket generator fails even in single OM environment:
> 
> ```
> Freon OM Bucket Generator                                             | FAIL |
> 255 != 0
> ```
> 
> due to:
> 
> ```
> OMException: Invalid bucket name: ELAtCj6bFe7
> ```
> 
> Can fix it by converting random Freon prefix to lowercase:
> 
> https://github.com/apache/hadoop-ozone/blob/61b6be001b78d1f3902db4a3a40734c19a90459c/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java#L240


Thanks @adoroszlai got the acceptance to pass finally!

PR #719, opened 2020-03-24
HDDS-3270. Allow safemode listeners to be notified when some precheck rules pass

sodonnel:

As part of the SCM safemode process, there are some rules which must pass before safemode can be exited.

One of these rules is the number of registered datanodes and another is that at least one pipeline must be created and open.

Currently, pipeline creation is attempted each time a node registers. As soon as the 3rd node registers, pipelines will be created.

There are two issue with this:

1. With network topology, if the first 3 nodes are all from the same rack, a non-rackaware pipeline will get created.

2. With multi-raft, it would be better if more nodes are registered to allow the multiple pipelines per node to be spread across all the available nodes.

The proposal here, is to introduce a new sub-state into the safemode process, call "preCheckComplete". When adding rules to the Safemode Manager, some rules can be tagged as "preCheck" (eg number of datanodes registered). When all all the pre-check rules have passed a notification will be sent to all safemode listeners:
```
  safeModeIsOn -> true
  preCheckComplete -> true
```
That will allow the listener to take action on this first stage completing. In the case of PipelineManager, it will then allow pipelines to be created.

After the remaining rules have been passed, safemode will exit as normal, by sending a second event:
```
  safeModeIsOn -> false
  preCheckComplete -> true
```

https://issues.apache.org/jira/browse/HDDS-3270

## How was this patch tested?

New unit tests added and existing tests modified.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
@@ -221,7 +221,7 @@ public class SCMPipelineManager implements PipelineManager {
   @Override
   public synchronized Pipeline createPipeline(ReplicationType type,
       ReplicationFactor factor) throws IOException {
-    if (!allowPipelineCreation.get()) {
+    if (!allowPipelineCreation.get() && factor != ReplicationFactor.ONE) {
       LOG.debug("Pipeline creation is not allowed until safe mode prechecks " +
           "complete");
       throw new IOException("Pipeline creation is not allowed as safe mode " +
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
@@ -88,6 +88,9 @@ public class SCMPipelineManager implements PipelineManager {
   private ObjectName pmInfoBean;
 
   private final AtomicBoolean isInSafeMode;
+  // Used to track if the safemode pre-checks have completed. This is designed
+  // to prevent pipelines being created until sufficient nodes have registered.
+  private final AtomicBoolean allowPipelineCreation;
 
   public SCMPipelineManager(Configuration conf, NodeManager nodeManager,
       EventPublisher eventPublisher)
@@ -134,6 +137,9 @@ public class SCMPipelineManager implements PipelineManager {
     this.isInSafeMode = new AtomicBoolean(conf.getBoolean(
         HddsConfigKeys.HDDS_SCM_SAFEMODE_ENABLED,
         HddsConfigKeys.HDDS_SCM_SAFEMODE_ENABLED_DEFAULT));
+    // Pipeline creation is only allowed after the safemode prechecks have
+    // passed, eg sufficient nodes have registered.
+    this.allowPipelineCreation = new AtomicBoolean(!this.isInSafeMode.get());
   }
 
   public PipelineStateManager getStateManager() {
@@ -144,6 +150,16 @@ public class SCMPipelineManager implements PipelineManager {
   public void setPipelineProvider(ReplicationType replicationType,
                                   PipelineProvider provider) {
     pipelineFactory.setProvider(replicationType, provider);
+  }
+
+  @VisibleForTesting
+  public void setAllowPipelineCreation(boolean newState) {
+    this.allowPipelineCreation.set(newState);
+  }
+
+  @VisibleForTesting
+  public boolean getAllowPipelineCreation() {
+    return allowPipelineCreation.get();
   }
 
   protected void initializePipelineState() throws IOException {
@@ -205,6 +221,12 @@ public class SCMPipelineManager implements PipelineManager {
   @Override
   public synchronized Pipeline createPipeline(ReplicationType type,
       ReplicationFactor factor) throws IOException {
+    if (!allowPipelineCreation.get()) {
+      LOG.debug("Pipeline creation is not allowed until safe mode prechecks " +
+          "complete");
+      throw new IOException("Pipeline creation is not allowed as safe mode " +
+          "prechecks have not yet passed");
+    }
     lock.writeLock().lock();
     try {
       Pipeline pipeline = pipelineFactory.create(type, factor);
@@ -641,13 +663,26 @@ public class SCMPipelineManager implements PipelineManager {
   }
 
   @Override
-  public void handleSafeModeTransition(
+  public synchronized void handleSafeModeTransition(
       SCMSafeModeManager.SafeModeStatus status) {
-    this.isInSafeMode.set(status.getSafeModeStatus());
-    if (!status.getSafeModeStatus()) {
-      // TODO: #CLUTIL if we reenter safe mode the fixed interval pipeline
-      // creation job needs to stop
+    // TODO: #CLUTIL - handle safemode getting re-enabled
+    boolean currentAllowPipelines =
+        allowPipelineCreation.getAndSet(status.isPreCheckComplete());
+    boolean currentlyInSafeMode =
+        isInSafeMode.getAndSet(status.isInSafeMode());
+    boolean triggerPipelines = false;
+
+    // Trigger pipeline creation only if the preCheck status has changed to
+    // complete.
+    if (allowPipelineCreation.get() && !currentAllowPipelines) {
+      triggerPipelines = true;
+    }
+    // Start the pipeline creation thread only when safemode switches off
+    if (!getSafeModeStatus() && currentlyInSafeMode) {
       startPipelineCreator();
+      triggerPipelines = true;
+    }
+    if (triggerPipelines) {
       triggerPipelineCreation();
     }
   }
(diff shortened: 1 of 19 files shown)


sodonnel:
Getting intermitted unit test failure on:

```
[ERROR] Errors: 
[ERROR]   TestDeadNodeHandler.testOnMessage:156 ? Timeout timeout: after 120000 millis
```

However it passes locally and passes sometimes on github.

Acceptance failure is:

```
Output:  /tmp/smoketest/ozonesecure-mr/result/robot-ozonesecure-mr-ozonesecure-mr-kinit-hadoop-rm.xml
ozonesecure-mr-mapreduce :: Execute MR jobs                                   
Execute PI calculation                                                | FAIL |
1 != 0
Execute WordCount                                                     | FAIL |
1 != 0
ozonesecure-mr-mapreduce :: Execute MR jobs                           | FAIL |
```

This passes locally. The actual seems to be some sort of space issue:

```
2020-03-25 16:36:06 INFO  Job:1574 - The url to track the job: http://rm:8088/proxy/application_1585154084804_0002/
2020-03-25 16:36:06 INFO  Job:1619 - Running job: job_1585154084804_0002
2020-03-25 16:36:08 INFO  Job:1640 - Job job_1585154084804_0002 running in uber mode : false
2020-03-25 16:36:08 INFO  Job:1647 -  map 0% reduce 0%
2020-03-25 16:36:09 INFO  Job:1660 - Job job_1585154084804_0002 failed with state FAILED due to: Application application_1585154084804_0002 failed 2 times due to AM Container for appattempt_1585154084804_0002_000002 exited with  exitCode: -1000
Failing this attempt.Diagnostics: [2020-03-25 16:36:08.025]No space available in any of the local directories.
For more detailed output, check the application tracking page: http://rm:8088/cluster/app/application_1585154084804_0002 Then click on links to logs of each attempt.
. Failing the application.
```

IT-Client failure seems intermittent too:

```
[ERROR] Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 174.231 s <<< FAILURE! - in org.apache.hadoop.ozone.client.rpc.TestCommitWatcher
[ERROR] testReleaseBuffers(org.apache.hadoop.ozone.client.rpc.TestCommitWatcher)  Time elapsed: 106.835 s  <<< ERROR!
java.util.concurrent.ExecutionException: org.apache.ratis.protocol.AlreadyClosedException: SlidingWindow$Client:client-17BDABCAD80C->RAFT is closed.
	at java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:357)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1908)
	at org.apache.hadoop.ozone.client.rpc.TestCommitWatcher.testReleaseBuffers(TestCommitWatcher.java:208)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -146,6 +152,16 @@ public void setPipelineProvider(ReplicationType replicationType,
...
+  @VisibleForTesting
+  public void setAllowPipelineCreation(boolean newState) {
+    this.allowPipelineCreation.set(newState);
+  }
+
+  @VisibleForTesting
+  public boolean getAllowPipelineCreation() {
+    return allowPipelineCreation.get();
+  }
+
```
fapifta: I would suggest to use pipelineCreationAllowed as the internal state name, it makes the code easier to read out as I I see.

I know this is a regular getter and and setter pair, but...
In the tests, the set method is always called with true as the parameter, so I suggest to call this method simply allowPipelineCreation() without a parameter, the getter part of the pair in this case can be called as isPipelineCreationAllowed().

This suggestion is simply about expressions from the usage point of view, and I don't have a strong opinion on this, however if we want to keep this name and want to name it as a regular getter setter as it is suggested by the books then we should use isAllowPipelineCreation as the getter name.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -641,13 +663,26 @@ public boolean getSafeModeStatus() {
...
+    // complete.
+    if (allowPipelineCreation.get() && !currentAllowPipelines) {
+      triggerPipelines = true;
+    }
+    // Start the pipeline creation thread only when safemode switches off
+    if (!getSafeModeStatus() && currentlyInSafeMode) {
       startPipelineCreator();
+      triggerPipelines = true;
+    }
+    if (triggerPipelines) {
```
fapifta: Let me understand a bit this part of the code:
we trigger pipelines in two cases:
1: the pipeline creation was not allowed before, but now it is allowed, effectively when precheck is complete
2: safemode is transitioning to off, and currently we are in safe mode

Looking at the code of triggerPipelineCreation, it is calling the same method from the BackgroundPipelineCreator, which is doing this only after the creator itself is started.
Now this means that we are effectively doing the same as the old code, and trigger pipeline creation after leaving safe mode, am I understand this well?

sodonnel: In the logic prior to this patch, `BackgroundPipelineCreator` only gets started when safemode exits. However, the creator does not need to be started for pipelines to get created. Infact, in order to exit safemode, there must be a replication factor 3 pipeline created, and that must happen without the creator running.

There are actually two ways a pipeline can be created:

1. When the backgroundPipelineCreator is running, it starts every 2 minutes and tries to create pipelines.

2. When a node registers with SCM, it triggers an event and calls `backgroundPipelineCreator.triggerPipelineCreation();` which tries to create pipelines immediately - even if the BackgroundPipelineCreator has already been started.

Therefore in safemode, before this patch, pipelines are initially created by (2) above. However now we want to stop that happening. Therefore whenever the precheck completes (enough nodes registered) we trigger the pipeline creation, but hold off on starting the background creator thread. Then it is started when safemode exits.

Now that you have highlighted this, it is possibly valid to just start the BackgroundPipeLineCreator in both cases. However there is one further detail - in the existing implementation, there is a delay for safemode exiting. Once the rules have passed some services are not started for a delay of 5 minutes and the BackgroundPipelineCreator thread is one of them. I am not sure on the history of that so it probably makes sense to keep that logic for now.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
sodonnel: I think your naming is better. I will push a new commit for that change.

sodonnel:
@fapifta Thanks for the review. I have taken on your suggested changes and also added some info related to your other question. Please let me know if you have more questions or suggestions.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
fapifta: Oh, I missed one small detail in the shouldSchedulePipelineCreator method... and misinterpreted it sorry about that... So triggerPipelineCreation will create pipelines if the safemode precheck is finished and allowPipelineCreation is already true in SCMPipelineManager.

With that it makes perfect sense, that we want to trigger the pipeline creation once when precheck is complete, and start the scheduled runs after the safe mode is exited.

But in this case, it would be enough to call triggerPipelineCreation when we arrived to the end of the precheck stage, and start the background pipeline creator schedule when we left safe mode.

Now what I understand happens at a startup when this is interesting is the following:
for (0..nodecount) {
  triggerPipelineCreation() // noop as allowPipelineCreation is false
}
precheck finished
triggerPipelineCreation // runs and creates pipelines
safe mode exit
startFixedIntervalPipelineCreator // calls createPipelines right away once at start
triggerPipelineCreation // tries to call createPipelines if it is not running from the prev. call

I think we don't need this last call, which makes the triggerPipelines boolean unnecessary and the code a bit simpler. Is it correct, or I am still missing something? :)

sodonnel: Good observation - you are correct. I have simplified this as you suggested.

fapifta:
Hi @sodonnel thank you for addressing the concerns, and thank you for working on the changes.
The patch in its current form looks good to me, +1 (non-binding) to commit it ;)

elek:
Thanks the patch @sodonnel . I created a patch to show my suggestion from the previous review (#690) 

See here #745

I would prefer to get a consensus first....

elek:
@sodonnel What is your opinion? Do you agree with the approach of #745. Can you merge both of these? (The order doesn't matter for me, I would be happy to rebase any of them).

sodonnel:
@elek Yes I agree we should go ahead with #745, but we should merge this one first as it is already reviewed and ready. We can rework #745 after this one goes in.

sodonnel:
Note this still needs a binding +1 before it can be merged.

elek:
> @elek Yes I agree we should go ahead with #745,

Thank you very much the feedback. It's good to have a consensus. 

As I wrote earlier I am fine with any order between this and that. (But I didn't have time to check this patch fully (yet))

sodonnel:
@elek I believe @avijayanhwx is going to give this one a review later today. Once he is happy I believe we can go ahead and merge this one, and then I can sort out #745.

avijayanhwx:
LGTM +1

PR #724, opened 2020-03-26
HDDS-3282. ozone.http.filter.initializers can't be set properly for S…

xiaoyuyao:
Change to use the Initializer class from hadoop-common while keeping the ozone initializer configuration key as-is. 

More specifically, to enable SPNEGO auth for ozone web console, you need to properly set the auth filter initializer like below.

**ozone.http.filter.initializers = org.apache.hadoop.security.AuthenticationFilterInitializer**

Standardize the SPNEGO related configurations to follow the SPNEGO filter prefix rules as follows:

ozone.[component].http.auth.type = [simple, kerberos]
ozone.[component].http.auth.kerberos.principal
ozone.[component].http.auth.kerberos.keytab

component should be [om,scm,datanode,s3g,recon]

https://issues.apache.org/jira/browse/HDDS-3282

## How the fix is tested:

Test with ozonesecure docker-compose locally and verified that the change ensure unauthenticated user will get 401 error like below.

```
URI: | /
-- | --
401
Authentication required
org.eclipse.jetty.servlet.DefaultServlet-1e1b061
```


avijayanhwx:
@xiaoyuyao Is it possible to add a negative test case for this in acceptance tests?

elek:
Thanks the patch @xiaoyuyao 

Not clear what is the exact problem with forking? I would like to remove the dependency on Hadoop classes (to be compatible with multiple hadoop version). And interested what was the problem with the forks...

xiaoyuyao:
bq. And interested what was the problem with the forks...
The problem is with the partially forked Filter implementation from hadoop-common, e.g. the SPNEGO implemented by org.apache.hadoop.security.AuthenticationFilterInitializer has assumption on the SPNEGO configuration prefix and key names.  When they are not matched in ozone configurations, the AuthenticationFilter will not work as expected. This need larger change then the current PR, I will try to isolate them into small PR with documents.

vivekratnavel:
@xiaoyuyao Thanks for working on this! Can you please fix the acceptance test failure as @adoroszlai pointed out?

xiaoyuyao:
Still need some additional changes to handle the deprecation of the inconsistent key names related to HTTP auth.

xiaoyuyao:
These two tests are not very reliable they both passed in my local setup but failed on CI.
org.apache.hadoop.ozone.client.rpc.Test2WayCommitInRatis
org.apache.hadoop.fs.ozone.contract.ITestOzoneContractSeek

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/HttpServer2.java
```diff
@@ -57,6 +57,9 @@
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.utils.LegacyHadoopConfigurationSource;
+import org.apache.hadoop.http.FilterContainer;
+import org.apache.hadoop.http.FilterInitializer;
+import org.apache.hadoop.http.lib.StaticUserWebFilter;
```
adoroszlai, line 62: Are these classes copied from Hadoop being removed and dependency on Hadoop being re-introduced intentionally? CC @elek

xiaoyuyao: @adoroszlai Yes, these are brought back to get SPNEGO work with Httpserver2. SPNEGO is implemented by KerberosAuthenticationHandler/KerberosAuthenticator in hadoop-common.

On server side, KerberosAuthenticationHandler is initialized by AuthenticationFilterInitializer that implements org.apache.hadoop.http.FilterInitializer. Unless, we decide to fork more code from hadoop-common, I think we should stick with the FilterContainer/FilterInitilaizer/StaticUserWebFilter from hadoop-common given the fact that there are barely major change in SPNEGO from hadoop 2 to hadoop 3.

elek:
Still +1 

As I wrote earlier, I am fine to remove the (half-made) fork temporary to have a working state and do proper (working) fork later.

> Still need some additional changes to handle the deprecation of the inconsistent key names related to HTTP auth.

I assume it also can be done in a follow-up jira if required.

I will merge this issue if no more objections.

xiaoyuyao:
TestBlockOutputStreamWithFailures seems to take long time to finish (10m) with pending flaky issues as query below. Should we consider disable it?

https://issues.apache.org/jira/issues/?jql=text%20~%20%22TestBlockOutputStreamWithFailures%22%20and%20resolution%20%3D%20%22Unresolved%22

elek:
> I will merge this issue if no more objections.

Green build, no objections --> I am merging it.

xiaoyuyao:
Thanks all for the reviews and discussions. I've merged it to master.

elek:
> I've merged it to master.

Sorry, I merged it locally as I promised but forget to push it. Thanks to finish it.

PR #726, opened 2020-03-26
HDDS-3267. replace ContainerCache in BlockUtils by LoadingCache

esahekmat:
ContainerCache removed and instead of it, LoadingCache replaced in BlockUtils,
The count in ReferenceCountedDB removed and this class renamed to ReferenceDB

https://issues.apache.org/jira/browse/HDDS-3267

## How was this patch tested?

mvn clean package
acceptance.sh
checkstyle.sh


elek:
Ups, I tried to retrigger the 762 not the 726. Sorry for the noise. Anyway a new build always can be useful... ;)

elek:
Just realized that there is no review here. @fapifta might have better context, but will try to do on Monday if nobody else...

elek:
Thanks the patch @esahekmat, I am trying to understand it, but to be honest I am not very familiar with this code path. I found this comment from @fapifta which was referenced by the Jira:

>  LoadingCache is a guava utility, it provides and LRU cache, and provides facilities to create the instances that are cached, and also guava provides facilities to configure how the cache works.
What we need to consider before replacing is that why the ContainerCache is implemented in a way where you can not evict a DB instance from the cache if there is a reference for it.
Looking at all the usages of BlockUtils getDb and removeDB methods, all the cached instances are accessed for just short term, and even though creating an instance might be costly, with the default 1024 cache size if the LoadingCache evicts the least recently used item when the cache is full, we do not suffer a huge perf hit, while in the current implementation if the cache is full, then we get an exception, so I would definitely trade this off.

> With LoadingCache which just have a size restriction, we can have the same cache functionality, if we get rid of reference counting which I think we do not need, as we use the instances just for short periods of time, so most of the time they are in the cache with a zero reference count, and can be evicted. If we still want to ensure that in extreme cases when the cache is pressured, and all items in the cache are referenced, we can use weakValues() in the guava cache to prevent eviction in this case, but this does not seem to be relevant based on the usage of the cached items.

Based on my understanding (but fix me if I am wrong) the reference count is used to check if the DB is used by anybody at the moment. The reference count is increased when the db is retrieved from the cache and decreased at the auto closable block.

If I understand well, this is removed in this patch. Let's say we have one code block which starts to use the DB but after a very short time the DB is evicted from the cache (which also means that it will be closed). The code block still tries to use the db reference and an exception will be thrown.

With LRU cache there is a very low chance to this, unless we have more open DBs at the same time than the limit. Which means that DB can be evicted after a get.

It is a possible problem? What do you think?

elek:
One more comment: In this patch an Apache commons based cache is replaced by a Guava based cache.

Guava can be better from technical point of view and it might be a good step, but I have to mention here the unlimited pain from Guava, especially the pain related to the backward compatibility of Guava. Check the "guava shading" thread on Apache Hadoop dev mailing lists.

Hadoop uses guava but downstream projects may also use guava but with different version. It's very frequent that there are non-compatible changes between versions. of Guava which makes where hard to use Hadoop (with unshaded guava). Which is understandable because guava development is driven by the needs of Google. And inside Google it's not a big problem due to the usage of one mono-repo.

But in external, open source world, a project which has open governance and more strict backward compatible rules, can be better. 

It doesn't mean that I think Guava never can be used (especially if there are no alternatives). But this is huge, read, blinking warning.

esahekmat:
@elek Thank you for your review. you understand this PR very well.
about the potential problem of evicting a reference while it is in use you're completely right, however we can address it by using weakValue eviction strategy which evict a value from the cache if there is no other (strong or soft) references to the value(ReferenceDB here).
I will edit this PR to use weakValue eviction instead of size-based eviction.
About the pain of using guava, you're right however, Ozone uses guava already. I think the ultimate solution to this is to shade it like the way Hadoop is trying to address it although I don't know the difficulties of this approach(shading).

elek:
Thanks the @esahekmat the update. I am not convinced if the the proposed patch (using `weakValues()`) solves the problem. It's about the GC, not about the eviction AFAIK.

Setting the ` .maximumSize()` using `OZONE_CONTAINER_CACHE_SIZE` is also missing from the new code.

Let's assume that the `maximumSize` is set to `2`:

```java
    cache = CacheBuilder.newBuilder()
          .weakValues()
          .maximumSize(2)
          .removalListener((notification) ->
              ((ReferenceDB) notification.getValue()).cleanup())
          .concurrencyLevel(1)
          ....
``` 

In this case the DB reference can be evicted even if an obtained reference is used:

```java
    conf.setInt(OzoneConfigKeys.OZONE_CONTAINER_CACHE_SIZE, 2);

    KeyValueContainerData container1instance1 = createContainer(1L);
    KeyValueContainerData container1instance2 = createContainer(1L);

    ReferenceDB dba = BlockUtils.getDB(container1instance1, conf);
    ReferenceDB dbb = BlockUtils.getDB(container1instance2, conf);

    //should be the same, if cache is used
    Assert.assertEquals(dba, dbb);

    //closing first reference, second is still used
    dba.close();

    //works well as the db itself is not closed, yet
    dbb.getStore().get("asd".getBytes());

    //force eviction with adding two more containers.
    BlockUtils.getDB(createContainer(3L), conf);
    BlockUtils.getDB(createContainer(4L), conf);

    //doesn't work, even if the dbb is not yet closed.
    dbb.getStore().get("asd".getBytes());
```

The last call throws a `NullPointerException`. The existing code (based on my understanding) works well, because it closes the DB only it there are no more ("unclosed") references to the DB.

fapifta:
I can not add much on the guava/no-guava question besides the fact that we already use it, and this code seems to be much more simpler with LoadingCache.

@elek let me address a few things:
1. in the current implementation if you try to run the test you get an NPE with, then it should throw an exception when the reference for container 4L is obtained, but it does not only because in the ContainerCache code there is the potential to easily violate the cache max size constraint...
The ContainerCache code in line 131 uses this.put() to put the value to the LRUMap, but in LRUMap put is not overridden, it provides addMapping() which checks properly for boundaries, and since the ContainerCache code does not check for boundaries, basically in the current implementation we have an unlimited cache which lies about itself and says it is limited by maxSize.
2. If the current implementation works correctly within boundaries, then in those cases when eviction poses a possible problem that needs proper handling though, we should see issues because the cache full and no items can be evicted.


I agree, weak values are not solving the problem when the cache is full, but in that case we either need to do not load a new ReferenceDB, or we should not close the db connection of the evicted ReferenceDB.

The problem to solve as I understand now is the following:
- we want to cache db connections and keep them live
- we want to keep the size restriction on the cache
- we want to close the db connection in case an unused cached connection is evicted from the cache
- we want to close the db connection in case it is used and got evicted from the cache only at the end of the current use, or - god forbid us - uses on multiple threads.

I don't have a good solution for all of these at the moment, and I am seeing this for a while now, let me get back to this later this week, in the meantime I am open to ideas and further discussion, but for now it seem for me that we can not spare the reference counting somewhere due to the nature of what we are caching and how we are using the cached stuff now.

elek:
> The problem to solve as I understand now is the following: ...

Agree with your 4 points, they are very good description of the problems.

elek:
/pending  I don't have a good solution for all of these at the moment, and I am seeing this for a while now, let me get back to this later this week...

elek:
For the records, there is a new proposal by @runzhiwang which would use only 1-2 (shared) rocksdb instances which would make this caching unnecessary.

https://issues.apache.org/jira/browse/HDDS-3630

I think this issue can be closed until the discussion (and reopened if required).

@fapifta / @esahekmat, any thoughts?

fapifta:
I think we should close this one, I sorry but I could not get back to it, and back then I haven't been able to come up with anything better than reference counting...
Two things:
- if we leave the caching in, then we should take care of the fact that the cache max size is not respected in the current implementation
- if we remove this caching completely, then we don't need to take care of the possible overrun for now

I am adding a comment to HDDS-3630 about this.

@esahekmat sorry for the unnecessary work on this, but thank you for taking it on.

esahekmat:
no problem @fapifta, This issue helps me to understand the underlying of the Ozon better

PR #727, opened 2020-03-27
HDDS-3273. getConf does not return all OM addresses.

swagle:
Added a new utility method to return a list of addresses keyed by serviceId. 

https://issues.apache.org/jira/browse/HDDS-3273

## How was this patch tested?
Added a unit test and verified via docker:

`bash-4.2$ /opt/hadoop/bin/ozone getconf -ozonemanagers`
`{omservice=[om1/172.22.0.6:9862, om2/172.22.0.4:9862, om3/172.22.0.2:9862]}`

`bash-4.2$ /opt/hadoop/bin/ozone getconf -ozonemanagers`
`om`

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
@@ -112,6 +112,9 @@ public final class OmUtils {
             addKeySuffixes(OZONE_OM_ADDRESS_KEY, serviceId, nodeId));
         if (rpcAddr != null) {
           result.get(serviceId).add(NetUtils.createSocketAddr(rpcAddr));
+        } else {
+          LOG.warn("Address undefined for nodeId: {} for service {}", nodeId,
+              serviceId);
         }
       }
     }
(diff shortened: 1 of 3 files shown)


swagle:
cc: @xiaoyuyao / @bharatviswa504 for review.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -89,6 +93,31 @@ public static InetSocketAddress getOmAddress(Configuration conf) {
...
+    Map<String, List<InetSocketAddress>> result = new HashMap<>();
+    for (String serviceId : conf.getTrimmedStringCollection(
+        OZONE_OM_SERVICE_IDS_KEY)) {
+      if (!result.containsKey(serviceId)) {
+        result.put(serviceId, new ArrayList<>());
+      }
+      for (String nodeId : getOMNodeIds(conf, serviceId)) {
+        String rpcAddr = getOmRpcAddress(conf,
+            addKeySuffixes(OZONE_OM_ADDRESS_KEY, serviceId, nodeId));
+        if (rpcAddr != null) {
```
bharatviswa504: One minor comment. when for one of nodeId is undefined, do we want to print "unknown address" instead of silently ignoring?

swagle: Thanks, Bharat for review.

Actually I modeled the method based on getOmAddress() and return InetSocketAddress so cannot put random string in the map, I added a log message instead.

bharatviswa504:
Thank You @swagle for the contribution.

PR #733, opened 2020-03-27
HDDS-3291. Write operation when both OM followers are shutdown.

bharatviswa504:

Add IPC client time out, so that the client will fail with socket time out exception in cases of 2 OM node failures.

https://issues.apache.org/jira/browse/HDDS-3291

## How was this patch tested?

Tested it in docker compose cluster with 1 minute, and see that it failed finally, instead of hanging.

To repro this test, we need to change leader.election.time.out value also to large value, as we need this request to be submitted to ratis, and as ratis server keeps on retry then only we will see this issue.

2020-03-27 16:25:27,625 [main] INFO  RetryInvocationHandler:411 - com.google.protobuf.ServiceException: java.net.SocketTimeoutException: Call From c5263a1df1ad/172.22.0.3 to om2:9862 failed on socket timeout exception: java.net.SocketTimeoutException: 60000 millis timeout while waiting for channel to be ready for read. ch : java.nio.channels.SocketChannel[connected local=/172.22.0.3:56460 remote=om2/172.22.0.4:9862]; For more details see:  http://wiki.apache.org/hadoop/SocketTimeout, while invoking $Proxy19.submitRequest over nodeId=om2,nodeAddress=om2:9862 after 15 failover attempts. Trying to failover immediately.
2020-03-27 16:25:27,626 [main] ERROR OMFailoverProxyProvider:285 - Failed to connect to OMs: [nodeId=om1,nodeAddress=om1:9862, nodeId=om3,nodeAddress=om3:9862, nodeId=om2,nodeAddress=om2:9862]. Attempted 15 failovers.

For now, the changed config affects all rpc clients in the system, in the future if we want to add a new RPC client time out, we can add a new config. For now, using the same config of ipc client.

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1686,6 +1686,17 @@
   </property>
 
   <property>
+    <name>ipc.client.rpc-timeout.ms</name>
+    <value>900000</value>
+    <description>
+      RpcClient timeout on waiting response from server. The default value is
+      set to 15 minutes. If ipc.client.ping is set to true and this rpc-timeout
+      is greater than the value of ipc.ping.interval, the effective value of
+      the rpc-timeout is rounded up to multiple of ipc.ping.interval.
+    </description>
+  </property>
+
+  <property>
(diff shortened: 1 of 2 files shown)


bharatviswa504:

PR #734, opened 2020-03-29
HDDS-3240. Improve write efficiency by creating container in parallel

runzhiwang:

**What's the problem ?**

Now follower cannot create container until leader finish creating container. But follower and leader can create container in parallel when leader can  getCachedStateMachineData from cache, rather than in sequential.

**Why leader and follower create container in sequential now ?**

1.  From the code,  the [future thread](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L672) do `getCachedStateMachineData ` in `readStateMachineData `and the [future thread](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L459) do `createContainer` in `writeStateMachineData `are the same [thread](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L505). Because `writeStateMachineData  `called before `readStateMachineData`. So leader must wait `createContainer `finish then `getCachedStateMachineData `and append logs to the follower, so follower must wait leader finish `createContainer`.  

2. From the jaeger UI, you can also see follower create container after leader finishing it currently.
image

**How to improve it ?**

1. This order can be improved by distinguishing the thread used by `getCachedStateMachineData `  and `createContainer `, and  [data = readStateMachineData(requestProto, term, logIndex)](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L619)  use same thread with `createContainer `. If [stateMachineDataCache.get(logIndex)](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L617) does not return null,  leader can get stateMachineData from cache and need not wait `createContainer` finish, thus leader and follower can be independent. But if it return null, leader must finish `createContainer `and then apennd logs to the follower, so I think [data = readStateMachineData(requestProto, term, logIndex)](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L619) should use the same thread with `createContainer` rather than the whole [getCachedStateMachineData](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java#L614)

2. From the jaeger UI, you can see leader and follower create container in parallel.
image

https://issues.apache.org/jira/browse/HDDS-3240

## How was this patch tested?

Existed UT and IT


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java
```diff
@@ -664,12 +689,12 @@ private ByteString getCachedStateMachineData(Long logIndex, long term,
...
                 getCachedStateMachineData(entry.getIndex(), entry.getTerm(),
                     requestProto));
-          } catch (IOException e) {
+          } catch (Exception e) {
             metrics.incNumReadStateMachineFails();
             future.completeExceptionally(e);
           }
           return future;
-        }, getChunkExecutor(requestProto.getWriteChunk()));
+        }, getReadChunkExecutor(requestProto.getWriteChunk()));
```
adoroszlai: Instead of introducing another set of executors for read chunk operation, can the cache lookup be executed in the per-container executor returned by `getCommandExecutor`?

runzhiwang: @adoroszlai  Thank you very much. I will try it.

bshashikant: I think, lookUp from the cache can be done in the same thread where readStateMachineData is called. I don't think there is a need here to delegate the cache lookup to a different thread. Reading and writing from the disk can be done using the same set of chunk executor threads.

runzhiwang: @adoroszlai  How do you think of @bshashikant 's advice ? I think it's also reasonable, because lookup from cache is so fast, there is no need to do it in another thread.

adoroszlai: Sounds good.

runzhiwang:
@adoroszlai @bshashikant Could you help review it again.

PR #740, opened 2020-03-31
HDDS-3307. Improve write efficiency by avoiding reverse DNS lookup

runzhiwang:

**What's the problem ?**
1. When create key, [getHostName()](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java#L130) by IP need reverse DNS lookup, cost about 15ms. It's also reported in [JDK-6450279](https://bugs.openjdk.java.net/browse/JDK-6450279) and [ZOOKEEPER-1666](https://issues.apache.org/jira/browse/ZOOKEEPER-1666). 

2. Besides there is addressCache  in JDK for getting IP from Hostname, but there is no cache for getting Hostname from IP. So getting hostname from ip is slow, and maybe should avoid it.

**How to fix ?**
client getLocalHost().getHostName() and add it in request, server get hostname from request.

https://issues.apache.org/jira/browse/HDDS-3307

## How was this patch tested?

Existed UT and IT.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
@@ -130,7 +130,7 @@ public abstract class OMClientRequest implements RequestAuditor {
       OzoneManagerProtocolProtos.UserInfo userInfoInProtocol = getOmRequest()
           .toBuilder().getUserInfo();
       if (userInfoInProtocol.getHostName().isEmpty()) {
-        userInfo.setHostName(remoteAddress.getHostAddress());
+        userInfo.setHostName(remoteAddress.getHostName());
       } else {
         userInfo.setHostName(userInfoInProtocol.getHostName());
       }
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
@@ -128,7 +128,7 @@ public abstract class OMClientRequest implements RequestAuditor {
 
     if (remoteAddress != null) {
       OzoneManagerProtocolProtos.UserInfo userInfoInProtocol = getOmRequest()
-          .toBuilder().getUserInfo();
+          .getUserInfo();
       if (userInfoInProtocol.getHostName().isEmpty()) {
         userInfo.setHostName(remoteAddress.getHostName());
       } else {
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
@@ -127,7 +127,14 @@ public abstract class OMClientRequest implements RequestAuditor {
     }
 
     if (remoteAddress != null) {
-      userInfo.setHostName(remoteAddress.getHostName());
+      OzoneManagerProtocolProtos.UserInfo userInfoInProtocol = getOmRequest()
+          .toBuilder().getUserInfo();
+      if (userInfoInProtocol.getHostName().isEmpty()) {
+        userInfo.setHostName(remoteAddress.getHostAddress());
+      } else {
+        userInfo.setHostName(userInfoInProtocol.getHostName());
+      }
+
       userInfo.setRemoteAddress(remoteAddress.getHostAddress()).build();
     }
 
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
```diff
@@ -127,7 +127,14 @@ public OMRequest getOmRequest() {
     }
 
     if (remoteAddress != null) {
-      userInfo.setHostName(remoteAddress.getHostName());
+      OzoneManagerProtocolProtos.UserInfo userInfoInProtocol = getOmRequest()
+          .toBuilder().getUserInfo();
+      if (userInfoInProtocol.getHostName().isEmpty()) {
```
bharatviswa504, line 132: If userInfoInProtocol.getHostName(), we need to set it as before right?
Like userInfo.setHostName(remoteAddress.getHostName());

As when someone uses RpcClient, this if check will not be true. So if HostName is empty from UserInfo, I think we still need to set hostname as before.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
```diff
@@ -127,7 +127,14 @@ public OMRequest getOmRequest() {
     }
 
     if (remoteAddress != null) {
-      userInfo.setHostName(remoteAddress.getHostName());
+      OzoneManagerProtocolProtos.UserInfo userInfoInProtocol = getOmRequest()
```
bharatviswa504, line 130: We don't need toBuilder here, we can get UserInfo by calling getUserInfo()

Like `getOmRequest().getUserInfo()`

bharatviswa504:
And also test failure looks related to the code change.

runzhiwang:
> @runzhiwang Thanks for the PR and improvement.
> 
> I have few comments.
> 
> I have one general question, now in native authorizer, the hostname is not used during checkAcls, in the future if we use, allow requests from this host like settings, then with this approach, someone can set the hostName, even though the request is not from that host. This can happen with this code change, I am not sure we will support this kind of scenario or not. Just flagging this here.

@bharatviswa504 Thanks for your review. Very reasonable question. And I also doubt whether we need hostName in authorizer, because getHostName() from IP is a block call, maybe block a very long time, it's dangerous. If we do not need hostName in authorizer, we can delete `getHostName `and also no need to `setHostName `in client.

bharatviswa504:
> > @runzhiwang Thanks for the PR and improvement.
> > I have few comments.
> > I have one general question, now in native authorizer, the hostname is not used during checkAcls, in the future if we use, allow requests from this host like settings, then with this approach, someone can set the hostName, even though the request is not from that host. This can happen with this code change, I am not sure we will support this kind of scenario or not. Just flagging this here.
> 
> @bharatviswa504 Thanks for your review. Very reasonable question. And I also doubt whether we need hostName in authorizer, because `getHostName `from IP is a block call, maybe block very long time. If we do not need hostName in authorizer, we can delete `getHostName `and also no need to `setHostName `in client.

HostName is added based on the comments from @eyang during the review of HDDS-1600. Have a look into it. From reading the current code, in NativeAuthorizer, we have not used hostName during ACL validation. MIght be some external providers like ranger might need this, not sure ranger is currently using it or not.

elek:
What is the plan this PR?

I tried to understand but didn't get.

The hostname is either used for authorization (in this case I am not sure if it's safe to use any value from the client) or not used at all (in this case we can just ignore it).

Fix me If am wrong but hostname seems to be set on the client side, which is unreliable. (IMHO)

runzhiwang:
@elek This PR try to avoid finding DNS from IP address, it is a block call, and sometimes cost a long time, and also has no cache in JDK. It's true getting hostname from client side is unreliable. So maybe we should not use hostname at all.

runzhiwang:
@elek  I close it temporarily, because I‘m not sure whether hostname will be used or not in future.

elek:
> @elek I close it temporarily, because I‘m not sure whether hostname will be used or not in future.

We can't be prepared for all the possible challenges in the future. If it's not used now and we don't now any current plan, I would remove it. (which would solve your performance problem as I understand).

runzhiwang:
@elek It's ok to remove it. I get hostname from client because now hostname only for log, unreliable does not matter.  If you have no time, I am also available to remove the hostname.

elek:
> Might be some external providers like ranger might need this, not sure ranger is currently using it or not.

I checked the ranger plugin and I couldn't see any usage:

https://github.com/apache/ranger/tree/master/plugin-ozone/

> If you have no time, I am also available to remove the hostname.

@runzhiwang Please go ahead and create the patch. I can offer my help to review it....

PR #723, opened 2020-03-25
HDDS-3281. Add timeouts to all robot tests

hanishakoneru:
## What changes were proposed in this pull request?

We have seen in some CI runs that the acceptance test suit is getting cancelled as it runs for more than 6 hours. Because of this, the test results and logs are also not saved. 

This Jira aims to add a 5 minute timeout to all robot tests. In case some tests require more time, we can update the timeout. This would help to isolate the test which could be causing the whole acceptance test suit to time out.

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-3281

## How was this patch tested?

CI acceptance test suit can test this change as it only adds a timeout to robot tests.

PR #743, opened 2020-04-01
HDDS-3281. (Addendum) add missing timeouts to the topology robot files.

elek:

Just an addendum to the already reviewed #723 

https://issues.apache.org/jira/browse/HDDS-3281


elek:
Thanks the review @adoroszlai 

As it's a small change based on the approach where we already agreed, I am committing without waiting for other time-zones (especially as it is required to have better feedback from builds).

PR #749, opened 2020-04-01
HDDS-3322. StandAlone Pipelines are created in an infinite loop

hanishakoneru:

BackgroundPipelineCreator keeps creating pipelines of configured Replication type and all available Replication factors until some exception occurs while creating the pipeline such as no more available nodes.

When Replication Type is set to STAND_ALONE, we do not check if a DN has already been used to create a pipeline of same factor or not and keep reusing the same DNs to create new pipelines. This causes the pipeline creation to happen in an infinite loop.

https://issues.apache.org/jira/browse/HDDS-3322

## How was this patch tested?

Tested on a real cluster.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SimplePipelineProvider.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SimplePipelineProvider.java
@@ -21,7 +21,6 @@ package org.apache.hadoop.hdds.scm.pipeline;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationType;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationFactor;
-import org.apache.hadoop.hdds.protocol.proto.HddsProtos.NodeState;
 import org.apache.hadoop.hdds.scm.node.NodeManager;
 import org.apache.hadoop.hdds.scm.pipeline.Pipeline.PipelineState;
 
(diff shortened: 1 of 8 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SimplePipelineProvider.java
```diff
@@ -32,18 +31,18 @@
...
+  public SimplePipelineProvider(NodeManager nodeManager,
+      PipelineStateManager stateManager) {
+    super(nodeManager, stateManager);
   }
 
   @Override
   public Pipeline create(ReplicationFactor factor) throws IOException {
-    List<DatanodeDetails> dns =
-        nodeManager.getNodes(NodeState.HEALTHY);
+    List<DatanodeDetails> dns = pickNodesNeverUsed(ReplicationType.STAND_ALONE,
```
bharatviswa504, line 43: And also for Simple Pipeline Provider, we should not create pipeline with factor 3.
This is an existing issue, but when looking into this patch found it.

vivekratnavel:
@hanishakoneru Thanks for working on this! Can you please fix checkstyle issues and address review comments?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SimplePipelineProvider.java
hanishakoneru: Updated to create only factor ONE pipelines for StandAlone type.

hanishakoneru:
Thanks @bharatviswa504 and @vivekratnavel for the reviews.

> One question, so you have tested the scenario by changing Replication Type (ozone.replication.type) to STAND_ALONE.

Yup.


> And also now we have support for Ratis with 1 and 3 factor. I think now we can remove SimplePipelineProvider, Any thoughts?

Agree that we can remove it eventually.

bharatviswa504:
Thank You @hanishakoneru for the contribution and @vivekratnavel for the review.

PR #750, opened 2020-04-01
HDDS-3309. Add timeout to all integration tests

smengcl:

Add timeout to all integration tests that doesn't have one yet.

This should help getting the logs for long-running flaky tests (e.g. `TestBlockOutputStreamWithFailures`) in Github actions. It seems we are not getting useful logs as a result of the timeout.

https://issues.apache.org/jira/browse/HDDS-3309

## How was this patch tested?

Using existing tests.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSInputStream.java
@@ -38,12 +38,20 @@ import org.apache.commons.lang3.RandomStringUtils;
 import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.BeforeClass;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 
 /**
  * Test OzoneFSInputStream by reading through multiple interfaces.
  */
 public class TestOzoneFSInputStream {
+
+  /**
+    * Set a timeout for each test.
+    */
+  @Rule
+  public Timeout timeout = new Timeout(300000);
   private static MiniOzoneCluster cluster = null;
   private static FileSystem fs;
   private static Path filePath = null;
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
@@ -62,7 +62,9 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
@@ -75,6 +77,12 @@ import org.junit.runners.Parameterized.Parameters;
  */
 @RunWith(Parameterized.class)
 public class TestOzoneFileInterfaces {
+
+  /**
+    * Set a timeout for each test.
+    */
+  @Rule
+  public Timeout timeout = new Timeout(300000);
 
   private String rootPath;
   private String userName;
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -45,7 +45,9 @@ import static org.junit.Assert.assertTrue;
 
 import org.apache.hadoop.test.LambdaTestUtils;
 import org.junit.After;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -53,6 +55,12 @@ import org.slf4j.LoggerFactory;
  * Ozone file system tests that are not covered by contract tests.
  */
 public class TestOzoneFileSystem {
+
+  /**
+    * Set a timeout for each test.
+    */
+  @Rule
+  public Timeout timeout = new Timeout(300000);
 
   private static final Logger LOG =
       LoggerFactory.getLogger(TestOzoneFileSystem.class);
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFsHAURLs.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFsHAURLs.java
@@ -51,6 +51,8 @@ import java.util.OptionalInt;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
+import org.junit.Rule;
+import org.junit.rules.Timeout;
 import static org.apache.hadoop.hdds.HddsUtils.getHostName;
 import static org.apache.hadoop.hdds.HddsUtils.getHostPort;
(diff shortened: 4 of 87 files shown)


elek:
Thanks the patch @smengcl This timeouts (and especially the root cause) the biggest problem with the integration tests right now... 

> This helps ruling out flaky long-running tests (e.g. TestRandomKeyGenerator.bigFileThan2GB) that are taking a very long time to run in Github actions.

Did you see any long running integration tests?

It seems that we have a global timeout:

In the main `pom.xml`:
```
  <surefire.fork.timeout>900</surefire.fork.timeout>
...
  <forkedProcessTimeoutInSeconds>${surefire.fork.timeout}</forkedProcessTimeoutInSeconds>
```

Isn't it easier to decrease this number? (If I understand well it does the same as we fork the JVM)

Did you check the current execution time of all the integration tests? Is the proposed value is significant bigger than the expected time of the slowest test?

smengcl:
> Thanks the patch @smengcl This timeouts (and especially the root cause) the biggest problem with the integration tests right now...
> 
> > This helps ruling out flaky long-running tests (e.g. TestRandomKeyGenerator.bigFileThan2GB) that are taking a very long time to run in Github actions.
> 
> Did you see any long running integration tests?
> 
> It seems that we have a global timeout:
> 
> In the main `pom.xml`:
> 
> ```
>   <surefire.fork.timeout>900</surefire.fork.timeout>
> ...
>   <forkedProcessTimeoutInSeconds>${surefire.fork.timeout}</forkedProcessTimeoutInSeconds>
> ```
> 
> Isn't it easier to decrease this number? (If I understand well it does the same as we fork the JVM)
> 
> Did you check the current execution time of all the integration tests? Is the proposed value is significant bigger than the expected time of the slowest test?

Hey Marton, thanks for the comment.

`TestRandomKeyGenerator.bigFileThan2GB` failed in [this](https://github.com/apache/hadoop-ozone/runs/540098578) run. This isn't really a timeout, just flaky. I should change the description into another one.
```
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 333.028 s <<< FAILURE! - in org.apache.hadoop.ozone.freon.TestRandomKeyGenerator
[ERROR] bigFileThan2GB(org.apache.hadoop.ozone.freon.TestRandomKeyGenerator)  Time elapsed: 267.469 s  <<< FAILURE!
	at org.junit.Assert.fail(Assert.java:88)
```

But `TestOzoneRpcClient` seems like a problem [here](https://github.com/apache/hadoop-ozone/runs/540098466). When it fails, it doesn't seem to give any useful information in the log. And it is running for way too long:
```
[INFO] Apache Hadoop Ozone Integration Tests .............. FAILURE [53:03 min]
[INFO] Apache Hadoop Ozone Mini Ozone Chaos Tests ......... SKIPPED
[INFO] BUILD FAILURE
[INFO] Total time:  53:04 min
[INFO] Finished at: 2020-03-27T18:32:50Z
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M1:test (default-test) on project hadoop-ozone-integration-test: There was a timeout or other error in the fork -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
org.apache.hadoop.ozone.client.rpc.TestOzoneRpcClient
```

I'm not sure if those issues are already fixed or not. Anyway the idea of this jira is to add a class-global timeout to EACH of the test functions in those classes. Does `forkedProcessTimeoutInSeconds` achieve the same thing?
If it does, from the above `TestOzoneRpcClient` example it doesn't seem to work well either, as the whole it-client tests were running for ~53min = 3180sec.
It seems that as a result of the timeout, we are not getting useful logs to diagnose the flakiness. <- This is the main reason @arp7 asks me to add the timeout to all the tests.

We might want to further lower the timeout of some known flaky (timing out intermittently) tests later.
What do you think?

elek:
> I'm not sure if those issues are already fixed or not. Anyway the idea of this jira is to add a class-global timeout that applies to EACH test functions in those test classes.

Thanks to explain it. So with this approach we will have the name of the problematic method instead of having only the name of the test class. Very nice.

smengcl:
> @smengcl `it-freon` is currently disabled globally, so any change to those tests is not being validated, ie. whether they work OK with the new timeouts or not. I think we should avoid changing them, to avoid unexpected failures later when enabling them.
> 
> `forkedProcessTimeoutInSeconds` applies to each test class. Test output is available in the artifact (eg. it-client.zip), although test results are not.
> 
> I think it would be more useful to change tests that time out with the current limit to use `GenericTestUtils.waitFor`, which prints a thread dump in case of timeout.

Hmm I see `it-freon` is running in master branch? e.g. https://github.com/apache/hadoop-ozone/runs/564223039

I can remove the modifications to the ignored freon tests. i.e. `TestOzoneClientKeyGenerator` and `TestRandomKeyGenerator`.

I agree `GenericTestUtils.waitFor` is useful, but on a lower level. For example when we pinpointed the timeout to exact lines like the last assertion in `TestRandomKeyGenerator.bigFileThan2GB`. The goal of this jira however is just to add a class-global timeout. This should achieve the goal of collecting logs and displaying more friendly output.

adoroszlai:
> Hmm I see `it-freon` is running in master branch? e.g. https://github.com/apache/hadoop-ozone/runs/564223039

Yes, it seems a66aae8613c20171fa78cfa434282bcaf28691ed only disabled it for PRs, not for post-commit build.  Thanks for the info.

> I can remove the modifications to the ignored freon tests. i.e. TestOzoneClientKeyGenerator and TestRandomKeyGenerator.

Thanks.  What about `TestDataValidate*` and `TestFreonWith*`?

smengcl:
> > Hmm I see `it-freon` is running in master branch? e.g. https://github.com/apache/hadoop-ozone/runs/564223039
> 
> Yes, it seems [a66aae8](https://github.com/apache/hadoop-ozone/commit/a66aae8613c20171fa78cfa434282bcaf28691ed) only disabled it for PRs, not for post-commit build. Thanks for the info.
> 
> > I can remove the modifications to the ignored freon tests. i.e. TestOzoneClientKeyGenerator and TestRandomKeyGenerator.
> 
> Thanks. What about `TestDataValidate*` and `TestFreonWith*`?

Cool.

Let's keep the modification to `TestDataValidate*` and `TestFreonWith*` then? As they are still being run in post-commit.

elek:
Sorry for late comment, but today I learned that Hadoop has a [TimedOutTestsListener](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/TimedOutTestsListener.java).

It's configured in the `pom.xml files`, for example here: https://github.com/apache/hadoop/blob/1189af4746919774035f5d64ccb4d2ce21905aaa/hadoop-hdfs-project/hadoop-hdfs/pom.xml#L236

Wouldn't it be more effective to use a similar listener? (If yes, I would prefer to fork it instead of adding one more Hadoop dependency, especially after HDDS-3353 and HDDS-3312).

smengcl:
> Sorry for late comment, but today I learned that Hadoop has a [TimedOutTestsListener](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/TimedOutTestsListener.java).
> 
> It's configured in the `pom.xml files`, for example here: https://github.com/apache/hadoop/blob/1189af4746919774035f5d64ccb4d2ce21905aaa/hadoop-hdfs-project/hadoop-hdfs/pom.xml#L236
> 
> Wouldn't it be more effective to use a similar listener? (If yes, I would prefer to fork it instead of adding one more Hadoop dependency, especially after HDDS-3353 and HDDS-3312).

Good find. I'll take a look into that.

smengcl:
I will open a new PR for `TimedOutTestsListener` for the sake of clarity.

PR #752, opened 2020-04-02
HDDS-3319. Handle HA for BasicOzoneClientAdapterImpl#renew/cancel().

xiaoyuyao:
Fix the token renew and cancel logic in the Renewer class. By adding omserviceId into OzoneTokenIdentifier. This way, the renewer can renew and cancel the token without referring any additional ozone uri.
 
Fix the ozone token tool to support get/renew/cancel token in secure om ha setup.

https://issues.apache.org/jira/browse/HDDS-3319

## How was this patch tested?
Test with the private build in real cluster using ozone token tool and the distcp. 
Before the fix, the YARN RM will stuck due to the omservice id is not being used to connect to the om proxy up job submit to renew the ozone token. After the fix, the distcp will be able to finish successfully.

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneClientFactory.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneClientFactory.java
@@ -18,6 +18,8 @@
 
 package org.apache.hadoop.ozone.client;
 
+import java.io.ByteArrayInputStream;
+import java.io.DataInputStream;
 import java.io.IOException;
 import java.lang.reflect.Proxy;
 
@@ -173,7 +175,12 @@ public final class OzoneClientFactory {
   public static OzoneClient getOzoneClient(Configuration conf,
       Token<OzoneTokenIdentifier> token) throws IOException {
     Preconditions.checkNotNull(token, "Null token is not allowed");
-    String omServiceId = token.decodeIdentifier().getOmServiceId();
+    OzoneTokenIdentifier tokenId = new OzoneTokenIdentifier();
+    ByteArrayInputStream buf = new ByteArrayInputStream(
+        token.getIdentifier());
+    DataInputStream in = new DataInputStream(buf);
+    tokenId.readFields(in);
+    String omServiceId = tokenId.getOmServiceId();
     if (StringUtils.isNotEmpty(omServiceId)) {
       // new OM should always issue token with omServiceId
       if (!OmUtils.isServiceIdsDefined(conf)
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneClientFactory.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneClientFactory.java
@@ -184,10 +184,10 @@ public final class OzoneClientFactory {
         return OzoneClientFactory.getRpcClient(omServiceId, conf);
       } else {
         // HA with mismatched service id
-        throw new IOException("Service ID specified does not match" +
-            " with " + OZONE_OM_SERVICE_IDS_KEY + " defined in the " +
-            "configuration. Configured " + OZONE_OM_SERVICE_IDS_KEY +
-            " are" + conf.getTrimmedStringCollection(
+        throw new IOException("Service ID specified " + omServiceId +
+            " does not match" + " with " + OZONE_OM_SERVICE_IDS_KEY +
+            " defined in the " + "configuration. Configured " +
+            OZONE_OM_SERVICE_IDS_KEY + " are" + conf.getTrimmedStringCollection(
             OZONE_OM_SERVICE_IDS_KEY));
       }
     } else {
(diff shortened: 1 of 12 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/token/GetTokenHandler.java
```diff
@@ -39,7 +40,9 @@
     description = "get a delegation token.")
 public class GetTokenHandler extends Handler {
 
-
+  @CommandLine.Parameters(arity = "1..1",
+      description = Shell.OZONE_BUCKET_URI_DESCRIPTION)
```
bharatviswa504, line 44: OZONE_BUCKET_URI_DESCRIPTION -> need to be updated as o3://om or o3://<<serviceid>>

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneClientFactory.java
```diff
@@ -159,6 +163,47 @@ private static OzoneClient getRpcClient(ClientProtocol clientProtocol,
...
+   * @return
+   * @throws IOException
+   */
+  public static OzoneClient getOzoneClient(Configuration conf,
+      Token<OzoneTokenIdentifier> token) throws IOException {
+    Preconditions.checkNotNull(token, "Null token is not allowed");
+    String omServiceId = token.decodeIdentifier().getOmServiceId();
+    if (StringUtils.isNotEmpty(omServiceId)) {
+      // new OM should always issue token with omServiceId
+      if (omServiceId.equals(OzoneConsts.OM_SERVICE_ID_DEFAULT)) {
```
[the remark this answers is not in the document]
bharatviswa504: Do we need to update it to something like this  if (omServiceId.equals(OzoneConsts.OM_SERVICE_ID_DEFAULT) && !OmUtils.isOmHAServiceId(conf, omServiceId))

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneClientFactory.java
```diff
@@ -159,6 +163,47 @@ private static OzoneClient getRpcClient(ClientProtocol clientProtocol,
...
+        // Non-HA
+        return OzoneClientFactory.getRpcClient(conf);
+      } else if (OmUtils.isOmHAServiceId(conf, omServiceId)) {
+        // HA with matching service id
+        return OzoneClientFactory.getRpcClient(omServiceId, conf);
+      } else {
+        // HA with mismatched service id
+        throw new IOException("Service ID specified does not match" +
```
bharatviswa504, line 187: Minor: Can we also found service ID which helps during debug

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneClientFactory.java
```diff
@@ -159,6 +163,47 @@ private static OzoneClient getRpcClient(ClientProtocol clientProtocol,
...
+            "configuration. Configured " + OZONE_OM_SERVICE_IDS_KEY +
+            " are" + conf.getTrimmedStringCollection(
+            OZONE_OM_SERVICE_IDS_KEY));
+      }
+    } else {
+      // Old OM may issue token without omServiceId that should work
+      // with non-HA case
+      if (!OmUtils.isServiceIdsDefined(conf)) {
+        return OzoneClientFactory.getRpcClient(conf);
+      } else {
```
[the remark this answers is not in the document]
xiaoyuyao: The serviceID must be unique to support distcp cross clusters. In the case when token does not provide omServiceId or only the default serviceID is provided. Token renew/cancel may not be able to figure out the right OM clusters to talk to. That's the main reason that we fail here.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneClientFactory.java
xiaoyuyao: good point, will fix in the next commit.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneClientFactory.java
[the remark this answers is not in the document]
xiaoyuyao: Do we use the OM_SERVICE_ID_DEFAULT for non-HA case? My first version does not have the following logic and I thought this is purely for non-HA where we don't need this serviceID to create RPC client. 

if (omServiceId.equals(OzoneConsts.OM_SERVICE_ID_DEFAULT)) {
        // Non-HA
        return OzoneClientFactory.getRpcClient(conf);
      }

If all the HA and HA are unified, we can skip those lines and only use the second if when the omServiceId match with the local configure. Still, the default om service id is risky in cross cluster operations like distcp as I mentioned below.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/web/ozShell/token/GetTokenHandler.java
xiaoyuyao: Fixed.

PR #757, opened 2020-04-02
HDDS-3332. Make Robot test invocation independent from Python version

adoroszlai:

Invoke `robot` directly, without referencing `python` command.  This allows using Python 2 or 3 depending on which version Robot Framework was installed with.  After this, we can upgrade the `ozone-runner` docker image to Python 3.

https://issues.apache.org/jira/browse/HDDS-3332

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/555065119

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/testlib.sh
+++ b/hadoop-ozone/dist/src/main/compose/testlib.sh
@@ -99,7 +99,7 @@ execute_robot_test(){
   OUTPUT_PATH="$RESULT_DIR_INSIDE/robot-$OUTPUT_NAME.xml"
   # shellcheck disable=SC2068
   docker-compose -f "$COMPOSE_FILE" exec -T "$CONTAINER" mkdir -p "$RESULT_DIR_INSIDE" \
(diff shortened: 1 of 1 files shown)


elek:
We need to update the version of the `ozone-runner` for the full fix, doesnt' we?

If yes, would to do it here in an other PR?

adoroszlai:
> I would expect that `robot` uses the same python which is used to install it (eg `pip2` --> `python`). And it seems to be strange if `robot` which is installed with `pip` can be visible for a different python version.

I expect the same.  I didn't mean that the same Robot installation can be used regardless of Python version (I don't know if that would work or not).  This change only makes it possible to install via `pip3` without breaking the test.



> We need to update the version of the `ozone-runner` for the full fix, doesnt' we?
> If yes, would to do it here in an other PR?

Yes, we need that, too.  I'd prefer a separate PR (but the same issue).

adoroszlai:
Thanks @elek for the review.  I'll submit another PR for updating the runner after a clean run in my fork.

PR #758, opened 2020-04-02
HDDS-3280. Ozone BaseHTTPServer should honor ozone.security.enabled c…

vivekratnavel:
…onfig

 - Base HTTP server should start with SPNEGO username and keytab only if "ozone.security.enabled" and "ozone.security.kerberos.http.enabled" configs are set to true. 

https://issues.apache.org/jira/browse/HDDS-3280

## How was this patch tested?
This patch was tested in a cluster with Kerberos and auto-TLS enabled under different configurations:
 - Both "ozone.security.enabled" and "ozone.security.kerberos.http.enabled" set to false
 - "ozone.security.enabled" set to true
 - Both "ozone.security.enabled" and "ozone.security.kerberos.http.enabled" set to true


vivekratnavel:
@xiaoyuyao Please review

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
```diff
@@ -347,6 +347,10 @@
       "ozone.security.enabled";
   public static final boolean OZONE_SECURITY_ENABLED_DEFAULT = false;
 
+  public static final String OZONE_HTTP_SECURITY_ENABLED_KEY =
+      "ozone.security.kerberos.http.enabled";
```
xiaoyuyao, line 351: This should be better named like ozone.security.http.kerberos.enabled to have a common prefix for http related settings as there will be ozone.security.http.simple.* as well.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/BaseHttpServer.java
```diff
@@ -145,16 +146,19 @@ public BaseHttpServer(Configuration conf, String name) throws IOException {
       final InetSocketAddress httpsAddr, String name, String spnegoUserNameKey,
       String spnegoKeytabFileKey) throws IOException {
     HttpConfig.Policy policy = getHttpPolicy(conf);
+    boolean isSecurityEnabled = OzoneSecurityUtil.isSecurityEnabled(conf) &&
```
xiaoyuyao, line 149: NIT: can we wrap this into OzoneSecurityUtil as isHttpSecurityEnabled?

PR #760, opened 2020-04-02
HDDS-3334. OM Client failover to next OM on NotLeaderException

hanishakoneru:

On receiving NotLeaderException from OM server, we should failover to the next OM in the list instead of the suggested Leader. This will ensure that all the OMs are contacted in a round robin way. Failing always to the suggested leader is not robust.

https://issues.apache.org/jira/browse/HDDS-3334

## How was this patch tested?

TestOzoneManagerHA#testOMProxyProviderFailoverToCurrentLeader already covers this test.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
@@ -250,11 +250,13 @@ public final class OzoneManagerProtocolClientSideTranslatorPB
               getNotLeaderException(exception);
           if (notLeaderException != null &&
               notLeaderException.getSuggestedLeaderNodeId() != null) {
-            // We need to failover manually to the suggested Leader OM Node.
+            // TODO: NotLeaderException should include the host
+            //  address of the suggested leader along with the nodeID.
+            //  Failing over just based on nodeID is not very robust.
+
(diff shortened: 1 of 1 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHA.java
```diff
@@ -726,7 +726,7 @@ public void testOMProxyProviderFailoverToCurrentLeader() throws Exception {
     // information must be returned via the response and a failover must
     // happen to the leader proxy node.
     createVolumeTest(true);
-    Thread.sleep(2000);
+    Thread.sleep(5000);
```
vivekratnavel, line 729: Can we use lambdaTestUtils instead of sleep?
```suggestion
    LambdaTestUtils.await(6000, 2000, () -> (omFailoverProxyProvider.getCurrentProxyOMNodeId().equals(leaderOMNodeId)));
```

hanishakoneru: Didn't realize till Bharat pointed out that this test is disabled. I ll revert the change to test and we can fix this when enabling TestOzoneManagerHA. That test was disabled as it was flaky and would need more work before it can be enabled.

vivekratnavel: Sounds good!

vivekratnavel:
+1 LGTM

PR #761, opened 2020-04-03
HDDS-3336. Metrics for Recon OzoneManager DB sync.

avijayanhwx:
Added some useful metrics for tracking how Recon's OM DB requests are going on. 

https://issues.apache.org/jira/browse/HDDS-3336

## How was this patch tested?
Manually tested on top of https://github.com/apache/hadoop-ozone/pull/715.
Added unit tests.


avijayanhwx:
cc @vivekratnavel / @swagle

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/metrics/OzoneManagerSyncMetrics.java
```diff
@@ -0,0 +1,151 @@
...
+
+  @Metric(about = "OM snapshot request latency")
+  private MutableRate snapshotRequestLatency;
+
+  @Metric(about = "Number of OM delta requests made by Recon.")
+  private MutableCounterLong numDeltaRequests;
+
+  @Metric(about = "Number of OM delta requests made by Recon that had " +
+      "at least 1 update in the response.")
+  private MutableCounterLong numNonZeroDeltaRequests;
```
swagle, line 70: Why the need to track nonZero? Zero and total should be enough, no?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/metrics/OzoneManagerSyncMetrics.java
```diff
@@ -0,0 +1,151 @@
...
+  public void incrNumDeltaRequests() {
+    this.numDeltaRequests.incr();
+  }
+
+  public void incrNumDeltaRequestsFailed() {
+    this.numSnapshotRequestsFailed.incr();
+  }
+
+  public void incrNumUpdatesInDeltaTotal(long n) {
+    this.numUpdatesInDeltaTotal.incr(n);
```
swagle, line 105: Can n == 0?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/metrics/OzoneManagerSyncMetrics.java
avijayanhwx: We don't track zero. We track num requests in total and num requests where there were actually updates. Only the latter should contribute to the Average number of updates per response.

swagle: So why the relevance of nonZero when I already know how many delta updates came in last request.

avijayanhwx: We need a counter of non zero updates to maintain an average right? For example, if the number of updates we get from OM are like 2,2,0,0,0,1,3 
NumDeltaRequests = 7
NumNonZeroDeltaUpdateRequests = 4
TotalDeltaUpdates=8
AverageUpdates=2 (8/4)

swagle: For the math, yes, maybe we do not need to report it though.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/metrics/OzoneManagerSyncMetrics.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  public MutableGaugeFloat getAverageNumUpdatesInDeltaRequest() {
+    return averageNumUpdatesInDeltaRequest;
+  }
+
+  public MutableCounterLong getNumNonZeroDeltaRequests() {
+    return numNonZeroDeltaRequests;
+  }
+}
```
swagle, line 155: Newline

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/metrics/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This package contains Recon metrics related classes.
+ */
+package org.apache.hadoop.ozone.recon.metrics;
```
swagle, line 22: Newline

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/metrics/OzoneManagerSyncMetrics.java
swagle: This is ok, just resolve conv, I thought zero would be interesting but maybe not.

elek:
Thanks the patch @avijayanhwx 

FYI: HDDS-3239 introduced message level metrics for all the RPC endpoints. I think some of the metrics can be removed from this patch (eg. generic counter / latency)

avijayanhwx:
> Thanks the patch @avijayanhwx
> 
> FYI: HDDS-3239 introduced message level metrics for all the RPC endpoints. I think some of the metrics can be removed from this patch (eg. generic counter / latency)

Thank you for the information Marton! I am wondering if I should still keep the counter and time to keep it specific per recon client instance. In the future, more clients may use the same request with different input arguments, and the metrics will be averaged out across all clients.

elek:
> Thank you for the information Marton! I am wondering if I should still keep the counter and time to keep it specific per recon client instance. In the future, more clients may use the same request with different input arguments, and the metrics will be averaged out across all clients.

I am not sure if it will ever happen related to the OM db download RPC, but how knows....

But if you need a client side metrics: what about adding a `ProtocolMessageMetrics` to the `OzoneManagerProtocolClientSideTranslatorPB` (with some client specific tag, if needed). In that case you can have well-defined, unified metrics for all the OM calls on the recon side (without increasing the complexity anywhere else in the Recon code).

avijayanhwx:
> > Thank you for the information Marton! I am wondering if I should still keep the counter and time to keep it specific per recon client instance. In the future, more clients may use the same request with different input arguments, and the metrics will be averaged out across all clients.
> 
> I am not sure if it will ever happen related to the OM db download RPC, but how knows....
> 
> But if you need a client side metrics: what about adding a `ProtocolMessageMetrics` to the `OzoneManagerProtocolClientSideTranslatorPB` (with some client specific tag, if needed). In that case you can have well-defined, unified metrics for all the OM calls on the recon side (without increasing the complexity anywhere else in the Recon code).

@elek I think that is a good idea. I have removed the generic counter and timer metric for the RPC call, and I will do the client specific protocol message metrics in a future JIRA (since there is no other client who is going to use the getDBUpdates API in the near future).

vivekratnavel:
@elek Please review

PR #762, opened 2020-04-03
HDDS-3337. Export the IPC port of OM in docker-compose

maobaolong:

Export the IPC port of OM in docker-compose, so that we can use the ozone service inside the docker container while execute command at docker host.

HDDS-3337

## How was this patch tested?

See the `docker ps` result, you can see the port 9862 of OM container is exported after execute ```bash
hadoop-ozone/hadoop-ozone/dist/target/ozone-0.6.0-SNAPSHOT/compose/ozone/run.sh
```

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozone-hdfs/docker-compose.yaml
+++ b/hadoop-ozone/dist/src/main/compose/ozone-hdfs/docker-compose.yaml
@@ -42,6 +42,7 @@ services:
          - ../..:/opt/hadoop
       ports:
          - 9874:9874
+         - 9862:9862
       environment:
          ENSURE_OM_INITIALIZED: /data/metadata/om/current/VERSION
       env_file:
--- a/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-compose.yaml
+++ b/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-compose.yaml
@@ -33,6 +33,7 @@ services:
       - ../../..:/opt/hadoop
     ports:
       - 9874:9874
+      - 9862:9862
     environment:
       WAITFOR: scm:9876
       ENSURE_OM_INITIALIZED: /data/metadata/om/current/VERSION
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by maobaolong** [comment]
@elek Thank you for your approve, I found the error of it-client  pr-check is 

```
[ERROR] (node:162) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[INFO] info fsevents@1.2.11: The platform "linux" is incompatible with this module.
[INFO] info "fsevents@1.2.11" is an optional dependency and failed compatibility check. Excluding it from installation.
[INFO] info fsevents@2.0.6: The platform "linux" is incompatible with this module.
[INFO] info "fsevents@2.0.6" is an optional dependency and failed compatibility check. Excluding it from installation.
[INFO] [3/4] Linking dependencies...
[ERROR] warning " > less-loader@5.0.0" has unmet peer dependency "webpack@^2.0.0 || ^3.0.0 || ^4.0.0".
```

And i met this error during my local build too, i think it is a warning?

**Comment by adoroszlai** [comment]
@maobaolong the error in it-client is in `Test2WayCommitInRatis`.  Can you trigger a new build with an empty commit?

```
2020-04-03T08:10:19.2261022Z [ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 84.892 s <<< FAILURE! - in org.apache.hadoop.ozone.client.rpc.Test2WayCommitInRatis
2020-04-03T08:10:19.2342304Z [ERROR] test2WayCommitForRetryfailure(org.apache.hadoop.ozone.client.rpc.Test2WayCommitInRatis)  Time elapsed: 84.752 s  <<< ERROR!
```

elek:
OM protocol is easy, but datanode protocol is more tricky. 

 1. You can't use the same port for all the containers. If you define a fixed port, you can't scale-up datanode containers
 2. We need to have a notion of "external" host names. For example OM should return with a pipeline information which contains the external ip/host (localhost in this case) and the external port of the datanodes. We can add it but that would be more tricky.  (IMHO)

PR #763, opened 2020-04-03
HDDS-3302. Add ability in MiniOzoneCluster to create multiple storage directories.

mukul1987:
Add functionality in MiniOzoneCluster to create multiple datanode data directory per datanode.

https://issues.apache.org/jira/browse/HDDS-3302

## How was this patch tested?
Added new unit test in TestMiniOzoneCluster.

### DIFF
--- a/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/MiniOzoneChaosCluster.java
+++ b/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/MiniOzoneChaosCluster.java
@@ -360,6 +360,19 @@ public abstract class MiniOzoneChaosCluster extends MiniOzoneHAClusterImpl {
       conf.setInt(OzoneConfigKeys.DFS_CONTAINER_RATIS_LOG_PURGE_GAP, 100);
     }
 
+    /**
+     * Sets the number of HddsDatanodes data dirs to be started as part of
+     * MiniOzoneCluster.
+     *
+     * @param val number of datanodes dirs
+     *
+     * @return MiniOzoneCluster.Builder
+     */
+    public Builder setNumDatanodesDirs(int val) {
+      numOfDatanodesDirs = val;
+      return this;
+    }
+
     @Override
     public MiniOzoneChaosCluster build() throws IOException {
 
--- a/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/MiniOzoneChaosCluster.java
+++ b/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/MiniOzoneChaosCluster.java
@@ -361,15 +361,14 @@ public abstract class MiniOzoneChaosCluster extends MiniOzoneHAClusterImpl {
     }
 
     /**
-     * Sets the number of HddsDatanodes data dirs to be started as part of
-     * MiniOzoneCluster.
+     * Sets the number of data volumes per datanode.
      *
-     * @param val number of datanodes dirs
+     * @param val number of volumes per datanode.
      *
      * @return MiniOzoneCluster.Builder
      */
-    public Builder setNumDatanodesDirs(int val) {
-      numOfDatanodesDirs = val;
+    public Builder setNumDataVolumes(int val) {
+      numDataVolumes = val;
       return this;
     }
 
--- a/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/TestMiniChaosOzoneCluster.java
+++ b/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/TestMiniChaosOzoneCluster.java
@@ -22,6 +22,7 @@ import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.ozone.client.ObjectStore;
 import org.apache.hadoop.ozone.client.OzoneVolume;
 import org.apache.hadoop.ozone.MiniOzoneChaosCluster.FailureService;
+import org.apache.hadoop.util.ExitUtil;
 import org.junit.BeforeClass;
 import org.junit.AfterClass;
 import org.junit.Ignore;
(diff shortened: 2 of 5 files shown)


hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/TestMiniChaosOzoneCluster.java
```diff
@@ -123,6 +133,8 @@ public void run() {
       cluster.startChaos(failureInterval, failureInterval, TimeUnit.SECONDS);
       loadGenerator.startIO(numMinutes, TimeUnit.MINUTES);
     } catch (Exception e) {
+      LOG.error("terminating with exception", e);
+      ExitUtil.terminate(-2, e);
```
elek, line 137: Can you please help to understand the motivation behind this? I thought that if an exception is thrown the exit code will be non-zero anyway (If something is wrong with the error handling, GenericCli might already have the solution. As I remember it has some generic error handling logic).

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestMiniOzoneCluster.java
```diff
@@ -58,14 +58,12 @@
...
 import org.junit.Assert;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.yaml.snakeyaml.Yaml;
 
 /**
  * Test cases for mini ozone cluster.
  */
-@Ignore
```
elek, line 68: Why was it ignored and how is the problem fixed by this patch? Do we need to run multiple checks to prove that there is no intermittency any more?

adoroszlai: [Disabled on Dec 27](https://github.com/apache/hadoop-ozone/commit/9000323e3741a39cc8b645a3cb2acad7f13d2a88) on the `integration-test-cleanup` branch due to [a failed run](https://github.com/apache/hadoop-ozone/runs/365409169).  Unfortunately logs are no longer available to check what was the problem.

elek: Thanks to check it. It can be because a flakiness. In this case it would be great to have at least 3-5 green builds before the merge .

adoroszlai: > It can be because a flakiness. In this case it would be great to have at least 3-5 green builds before the merge .

Running 20x, passed 5x so far:
https://github.com/adoroszlai/hadoop-ozone/runs/570025575

adoroszlai: 20/20 passed

hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/MiniOzoneChaosCluster.java
```diff
@@ -360,6 +360,19 @@ protected void initializeConfiguration() throws IOException {
       conf.setInt(OzoneConfigKeys.DFS_CONTAINER_RATIS_LOG_PURGE_GAP, 100);
     }
 
+    /**
+     * Sets the number of HddsDatanodes data dirs to be started as part of
+     * MiniOzoneCluster.
```
adoroszlai, line 365: This looks like partially leftover comment -- are data dirs really "started"?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneCluster.java
```diff
@@ -394,6 +395,19 @@ public Builder setNumDatanodes(int val) {
       return this;
     }
 
+    /**
+     * Sets the number of HddsDatanodes data dirs to be started as part of
+     * MiniOzoneCluster.
```
adoroszlai, line 400: Same here.

hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/TestMiniChaosOzoneCluster.java
```diff
@@ -90,6 +99,7 @@ public static void init() throws Exception {
         .setNumOzoneManagers(numOzoneManagers)
         .setFailureService(failureService)
         .setOMServiceID(omServiceID)
+        .setNumDatanodesDirs(numDataVolumes)
```
adoroszlai, line 102: Can we make the naming consistent?

 * `numOfDatanodesDirs`
 * `setNumDatanodesDirs` (implies `numDatanodesDirs` is being set, note no `Of`)
 * `numDataVolumes`

hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/TestMiniChaosOzoneCluster.java
```diff
@@ -69,6 +74,10 @@
       description = "no of clients writing to OM")
   private static int numClients = 3;
 
+  @Option(names = {"-v", "--numDataVolume"},
+      description = "no of datanode volumes")
```
adoroszlai, line 78:
```suggestion
      description = "number of datanode volumes to create")
```

hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/TestMiniChaosOzoneCluster.java
mukul1987: done, moved to GenericCli.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestMiniOzoneCluster.java
mukul1987: This works, so leaving this.

PR #767, opened 2020-04-03
HDDS-3064. Get Key is hung when READ delay is injected in chunk file path

bshashikant:
.

Added a deadline while sending a request over grpc stream

https://issues.apache.org/jira/browse/HDDS-3064

## How was this patch tested?

Idea here is to run all existing tests which failed intermittently.


adoroszlai:
@bshashikant please don't force-push, we lose checks history

bshashikant:
The test failure is caused by testRatisLogParsing(org.apache.hadoop.ozone.dn.ratis.TestDnRatisLogParser)  Time elapsed: 19.342 s  <<< FAILURE!
 and is tracked by https://issues.apache.org/jira/browse/HDDS-3404. It's not related to the patch as its failing in quite a few tests.

bshashikant:
@elek , can we commit this now?

elek:
> @elek , can we commit this now?

This test caused intermittent test failures earlier. From my point of view, it's safe to merge after 5 greed build. I couldn't see 5 green build on this PR and couldn't understand the root cause of the failures in the past. It may or may not be related based on my very limited understanding.

I couldn't take the responsibility to merge it. If somebody can take, feel free to do it, but please don't introduce more intermittency. (I am on 0 not -1).

bshashikant:
> > @elek , can we commit this now?
> 
> This PR caused intermittent test failures earlier. From my point of view, it's safe to merge after 5 greed build. I couldn't see 5 green build on this PR and couldn't understand the root cause of the failures in the past. It may or may not be related based on my very limited understanding.
> 
> I couldn't take the responsibility to merge it. If somebody can take, feel free to do it, but please don't introduce more intermittency. (I am on 0 not -1).

Thanks @elek . I have filed https://issues.apache.org/jira/browse/HDDS-3550 which is happening because of expected state of container in SCM does not seem to match and https://issues.apache.org/jira/browse/HDDS-3551 which is happening bcoz of incremental OM leader not ready as its still applying transactions for time longer than expected.

The change proposed here is totally on client to datanode path while doing read over grpc which is not related to any of the failures.

We can still have more iterations of test with this.

elek:
Sure, I trust in your judge. I am just not smart enough to fully understand it ;-) I hoped that somebody else here is it and merge it. (I am not against it just not sure enough). @lokeshj1703 for example?

But if nobody else, I can do it based purely on trust ;-)

bshashikant:
https://issues.apache.org/jira/browse/HDDS-3628 and https://issues.apache.org/jira/browse/HDDS-3629 track the unit test failures. Both of them are not related to the patch here.

The acceptance test failures are also not related.
@lokeshj1703 , as discussed, its too complex to add a unit test case to repro the issue here. It was detected with a fault injection framework by adding delays during read from the disk.

bshashikant:
https://issues.apache.org/jira/browse/HDDS-3629 is fixed now. Reenabled the test back.

lokeshj1703:
@bshashikant Thanks for the contribution! @elek @adoroszlai Thanks for the reviews! I have committed the PR to master branch.

PR #769, opened 2020-04-03
HDDS-3344. Remove hardcoded ozone version in Jenkins file

dineshchitlangia:

Ozone version is hardcoded in Jenkins file.
So acceptance stage runner refers to non-existent path.
This PR removes the hard coded version from Jenkins file and replaces it with wild character *
At any given time, the path hadoop-ozone/dist/target/ will have only one ozone SNAPSHOT directory, so the wild character * use will solve this problem.

https://issues.apache.org/jira/browse/HDDS-3344

## How was this patch tested?

Based on inputs from @adoroszlai , tested by executing hadoop-ozone/dist/src/main/compose/test-all.sh. The results were available in the expected path as mentioned in Jenkins file.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -148,7 +148,6 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <protobuf.version>2.5.0</protobuf.version>
     <protoc.path>${env.HADOOP_PROTOC_PATH}</protoc.path>
 
-    <curator.version>2.12.0</curator.version>
     <findbugs.version>3.0.0</findbugs.version>
     <spotbugs.version>3.1.12</spotbugs.version>
     <dnsjava.version>2.1.7</dnsjava.version>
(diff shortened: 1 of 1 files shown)


hadoop-ozone/Jenkinsfile
```diff
@@ -48,7 +48,7 @@ node("ubuntu") {
         }, 'checkstyleResult')
 
         stageRunner('Acceptance', "acceptance", {
-             archiveArtifacts 'hadoop-ozone/dist/target/ozone-0.4.0-SNAPSHOT/smoketest/result/**'
+             archiveArtifacts 'hadoop-ozone/dist/target/ozone-*/smoketest/result/**'
```
adoroszlai, line 51:
```suggestion
             archiveArtifacts 'hadoop-ozone/dist/target/ozone-*/compose/result/**'
```

`acceptance.sh` no longer stores results in `smoketest/result`.  You can confirm this by running the script (though first you may want to tweak `hadoop-ozone/dist/src/main/compose/test-all.sh` to only run a single specific test, eg. `ozone-topology`).

dineshchitlangia: @adoroszlai Thank you, that was helpful. I tried your idea to test it.

elek:
I don't think this Jenkins file is up-to-date or can be used. It also can be deleted. But I am also fine with fixing it.

dineshchitlangia:
@elek Alright, I will update the PR to delete this file since it is not in use.

PR #773, opened 2020-04-05
HDDS-3347. PipelineActionHandler should handle unknown pipeline.

nandakumar131:
PipelineActionHandler is not able to handle PipelineActions for pipelines which is not present in SCM (deleted pipelines)

After SCM sends out the close pipeline command to datanodes, it removes the pipeline information from the pipeline db. If a datanode sends PipelineAction for the deleted pipeline, PipelineActionHandler cannot handle it.
In such a situation SCM should fire ClosePipelineCommand.

https://issues.apache.org/jira/browse/HDDS-3347

## How was this patch tested?
Added a new unit test to cover the scenario.


avijayanhwx:
LGTM +1

PR #775, opened 2020-04-06
HDDS-3340. Move Ozone Shell from ozone-manager to tools

adoroszlai:

1. Move Ozone Shell from `ozone-manager` module to `tools`, since it is not part of OM server-side
2. Rename the package `org.apache.hadoop.ozone.web.ozShell` to `org.apache.hadoop.ozone.shell`
   a. it has nothing to do with any web
   b. package name should be lowercase
   c. `oz` is unnecessary

https://issues.apache.org/jira/browse/HDDS-3340

## How was this patch tested?

CI: https://github.com/adoroszlai/hadoop-ozone/runs/562411892


adoroszlai:
Thanks @elek and @xiaoyuyao for the review.

PR #776, opened 2020-04-06
HDDS-3348. scmcli container info command shows the wrong container state.

nandakumar131:
Container State from SCM is displayed in output of `scmcli container info` command

https://issues.apache.org/jira/browse/HDDS-3348

## How was this patch tested?
The output was manually checked.

### DIFF
--- a/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
+++ b/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
@@ -22,8 +22,6 @@ import java.util.stream.Collectors;
 
 import org.apache.hadoop.hdds.cli.HddsVersionProvider;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
-import org.apache.hadoop.hdds.protocol.datanode.proto.ContainerProtos
-    .ContainerDataProto;
 import org.apache.hadoop.hdds.scm.client.ScmClient;
 import org.apache.hadoop.hdds.scm.container.common.helpers
     .ContainerWithPipeline;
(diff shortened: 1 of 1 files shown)


hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
```diff
@@ -57,26 +55,13 @@
...
-
       // Print container report info.
       LOG.info("Container id: {}", containerID);
-      String openStatus =
-          containerData.getState() == ContainerDataProto.State.OPEN ? "OPEN" :
-              "CLOSED";
-      LOG.info("Container State: {}", openStatus);
-      LOG.info("Container Path: {}", containerData.getContainerPath());
-
-      // Output meta data.
```
avijayanhwx, line 75: We don't want this extra information  about 'ContainerPath' & 'Datanodes' anymore?

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
```diff
@@ -57,26 +55,13 @@
   @Override
   public Void call() throws Exception {
     try (ScmClient scmClient = parent.getParent().createScmClient()) {
-      ContainerWithPipeline container = scmClient.
+      final ContainerWithPipeline container = scmClient.
```
bharatviswa504, line 58: Print pipeline information of the container will be useful, can we add that info.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
nandakumar131: We were printing the container path from one of the datanode, this could be misleading if datanodes are configured with a different path for the data directory. It's better to not print the container path in this context.

Printing Datanodes are not removed, it's still printed.

PR #744, opened 2020-04-01
HDDS-3312. Remove server-side dependencies from hdds/ozone-common.

elek:
## What changes were proposed in this pull request?

`hadoop-ozone/common` and `hadoop-hdds/common` projects are common between client and server. Therefore we should remove any server side utilities / dependencies from them as they would be added to the client classpath which makes harder the Ozone client / ozonefs adoption.

1. hadoop-hdds should depend on a minimal set of hadoop dependencies (can be managed a separated technical project / pom.xml)
2. code shared between server side projects (and not with the client) should be moved to the framework
3. OM related code should be moved to the ozone-manager instead of ozone/common
4. For technical reason some test utilities are imported to the Ozone code base (from Hadoop)

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-3312

## How was this patch tested?

Full integration test. This is a project organization patch without real change of the logic. If the clients can work without class-not-found exception. It should be fine.

PR #777, opened 2020-04-06
HDDS-3351. Remove unnecessary dependency Curator.

fapifta:

Remove Curator dependency which is not used, and remained there from the time when we migrated the pom.xml to ozone.pom.xml in the Hadoop project.

https://issues.apache.org/jira/browse/HDDS-3351

## How was this patch tested?

Locally I ran 'mvn clean install -DskipTests', it was running fine.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -148,7 +148,6 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <protobuf.version>2.5.0</protobuf.version>
     <protoc.path>${env.HADOOP_PROTOC_PATH}</protoc.path>
 
-    <curator.version>2.12.0</curator.version>
     <findbugs.version>3.0.0</findbugs.version>
     <spotbugs.version>3.1.12</spotbugs.version>
     <dnsjava.version>2.1.7</dnsjava.version>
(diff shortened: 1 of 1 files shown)


bharatviswa504:
Thank You @fapifta for the contribution and @swagle for the review.

elek:
Thanks the patch @fapifta (And the quick review @bharatviswa504 and @swagle) 

For the record: this patch *doesn't* remove the curator dependency. It modifies a `<dependencyManagement>` not a `<dependency>` section. `<depdendencyManagement>` is reponsible for defining the **default versions** not dependencies. 

> Locally I ran 'mvn clean install -DskipTests', it was running fine.

I did the same after this patch (using 9872253e8). And I checked the jar files:

```
mvn clean install -DskipTests

cd hadoop-ozone/dist/target/ozone-0.6.0-SNAPSHOT
find -name "*cura*"

./share/ozone/lib/curator-client-2.12.0.jar
./share/ozone/lib/curator-framework-2.12.0.jar
./share/ozone/lib/curator-recipes-2.12.0.jar
```

Curator is added as a dependency via the transitive dependencies of Hadoop artifacts. To fully remove it you should `<exclude>` it from the dependency hierarchy.

But don't worry. It's a good patch anyway (I am +1, even it if's too late ;-) ).

 1. We are not interested in curator, so it can be removed from the `<dependencyManagement>`, too
 2. But to remove curator fully we need some exclude rules (the good news is that is should be done by #744, at least on the client side. We should double check the server-side)

PR #784, opened 2020-04-08
HDDS-3355. Add wait time between client retries to OM

hanishakoneru:

Currently, client keeps retrying to connect to leader OM in a tight loop and fails after configured number of retires/ failovers.
If the Leader OM is not ready, the client can timeout quickly. So, we should instead try each OM once, and then wait before retrying again. 

https://issues.apache.org/jira/browse/HDDS-3355

## How was this patch tested?

Will add unit test in next commit.


hanishakoneru:
Thank you @arp7 for the review. I will merge this shortly.

PR #785, opened 2020-04-08
HDDS-3234. Fix retry interval default in Ozone client

elek:
This PR is about re-commit a reverted commit, already reviewed at  #698 

The original commit is reverted due to the intermittent failures, this PR is an attempt to re-commit it. Should have multiple green builds before merging.


elek:
Closing it for now. We need to investigate the problems as it seems to be related to the original patch.

PR #787, opened 2020-04-08
HDDS-3359. Remove Yetus helper code from Ozone.

mukul1987:

Yetus related helper code is not currently being used. It can be removed from the repo.

https://issues.apache.org/jira/browse/HDDS-3359

## How was this patch tested?

A build after this change works correctly.


mukul1987:
Thanks for the review @elek . I have changed the lines as suggested.

PR #275, opened 2019-11-27
HDDS-2633. Let findbugs.sh skip frontend plugin for Recon

adoroszlai:
## What changes were proposed in this pull request?

Findbugs/Spotbugs only checks Java code.  We can skip frontend plugin execution for Recon to save ~2 minutes.  Makes a difference mostly when running it locally.

https://issues.apache.org/jira/browse/HDDS-2633

## How was this patch tested?

Ran `findbugs.sh`, verified Ozone Recon is checked, but its `frontend-maven-plugin` executions are skipped.

```
[INFO] --- frontend-maven-plugin:1.6:install-node-and-yarn (Install node and yarn locally to the project) @ hadoop-ozone-recon ---
[INFO] Skipping execution.
[INFO] 
[INFO] --- frontend-maven-plugin:1.6:yarn (yarn install) @ hadoop-ozone-recon ---
[INFO] Skipping execution.
[INFO] 
[INFO] --- frontend-maven-plugin:1.6:yarn (Build frontend) @ hadoop-ozone-recon ---
[INFO] Skipping execution.
...
[INFO] --- spotbugs-maven-plugin:3.1.12:spotbugs (default-cli) @ hadoop-ozone-recon ---
[INFO] Fork Value is true
[INFO] Done SpotBugs Analysis....
...
[INFO] Total time:  05:12 min
```

vs. earlier:

```
[INFO] Total time:  07:56 min
```

PR #788, opened 2020-04-08
HDDS-3357. Add check for import from shaded package

adoroszlai:

Add a checkstyle rule to reject import from shaded packages.  This helps avoid accidentally using shaded transitive dependencies instead of direct, unshaded dependencies, eg. `org.apache.curator.shaded.com.google.common.collect.ImmutableSet` instead of `com.google.common.collect.ImmutableSet`.

Also skip frontend install/compile for Recon (as checkstyle only applies to Java code) similarly to ac93176a0.

https://issues.apache.org/jira/browse/HDDS-3357

## How was this patch tested?

Temporarily added an import from a `sun.*` package, verified that it is flagged (along with the only current import from shaded package):

```
$ hadoop-ozone/dev-support/checks/checkstyle.sh
...
hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
 21: Illegal import - sun.security.ec.ECKeyFactory.
hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/volume/TestVolumeSetDiskChecks.java
 38: Illegal import - org.apache.curator.shaded.com.google.common.collect.ImmutableSet.
$ echo $?
1
```

Failed check due to import from shaded package:
https://github.com/adoroszlai/hadoop-ozone/runs/570614153

Successful check after removing offender:
https://github.com/adoroszlai/hadoop-ozone/runs/570825882

### DIFF
--- a/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/volume/TestVolumeSetDiskChecks.java
+++ b/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/volume/TestVolumeSetDiskChecks.java
@@ -33,9 +33,9 @@ import org.apache.hadoop.test.GenericTestUtils;
 import org.apache.hadoop.util.DiskChecker.DiskErrorException;
 import org.apache.hadoop.util.Timer;
 
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import org.apache.commons.io.FileUtils;
-import org.apache.curator.shaded.com.google.common.collect.ImmutableSet;
 import static org.hamcrest.CoreMatchers.is;
 import org.junit.After;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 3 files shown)


adoroszlai:
@fapifta 70df72dc58d5d3067f6f45a8233ff6683dc3a1f5 is your change, it's included here only to make the check pass.  If you create a PR with it, we can merge that first.

adoroszlai:
Thanks @lokeshj1703 for the review.

> Why do we need -Dskip.yarn and -Dskip.installyarn in MAVEN_OPTIONS?

`yarn` in this case refers to the [Javascript package manager](https://yarnpkg.com/) by this name.  These flags save some time during the build, since Checkstyle does not care about Javascript code.  #275 added similar feature in the Findbugs script, which is also Java-specific.

fapifta:
Hmm, great improvement, the changes seem to be good, and I support committing them (non-binding +1), I have added the mentioned changeset a bit late... it is under PR #818 currently being checked by automated checks.

lokeshj1703:
@adoroszlai Thanks for clarifying! I am +1 on the PR.

vivekratnavel:
@adoroszlai Thanks for working on this! Merged it.

adoroszlai:
Thanks @fapifta and @lokeshj1703 for the review and @vivekratnavel for review and merge.

PR #789, opened 2020-04-08
HDDS-3361. Remove ContainerReplica states representing decomission and maintenance

sodonnel:

In an earlier change, we decided that the Datanode would change the state of decommissioning / maintenance replicas and report the new state in its container report. Using that assumption, the logic in replication manager was changed to expect the ContainerReplica to contain the decom / maintenance state.

However in HDDS-2592 it was decided that the DN would simply store the operational state and report only the operation state in the heartbeat. This state is reported back to SCM and stored in the datanodeDetails instances.

This means that earlier assumption (ContainerReplica will contain the decom state) is no longer valid.

This Jira is to change the logic so replication decisions are based on the "persistedOpState" in the datanode details object instead.

https://issues.apache.org/jira/browse/HDDS-3361

## How was this patch tested?

Existing and modified unit tests

### DIFF
--- a/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/protocol/MockDatanodeDetails.java
+++ b/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/protocol/MockDatanodeDetails.java
@@ -17,6 +17,8 @@
  */
 package org.apache.hadoop.hdds.protocol;
 
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
+
 import java.io.IOException;
 import java.net.ServerSocket;
 import java.util.Random;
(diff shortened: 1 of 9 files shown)


sodonnel:
This change is going to branch to remove an incorrect earlier assumption. The branch is currently a work in progress toward an MVP decommission feature, hence merging this without any review.

elek:
> The branch is currently a work in progress toward an MVP decommission feature, hence merging this without any review

I don't think it's a good practice. Pull requests to feature branches are also reviewed (until now). If you would like to experiment with something, you can create a branch your own fork.

I would prefer to get review for all the commits in `apache/hadoop-ozone` repositories.

sodonnel:
@elek Thanks for the feedback. In that case I will not do that again.

Quoting from the email I was sent from Anu when setting up this branch:

> You are free to develop on these branches as you choose, but by
> continuing to open JIRAs, solicit reviews, and post high-level designs, you increase the chances > that others can support a branch when it is ready to merge. 

This gave me the impression that unreviewed commits are fine on branch, but I agree it will make a later merge easier if things are reviewed as we go. 

The current state of the branch is that the end to end flow does not work (it never worked as we never got that far with it) so it would be good to get it to the minimum working feature state and then we can refactor toward the event driven model you suggested some months back.

I will be sure to post any further changes for review and we can also change anything done here if needed.

PR #790, opened 2020-04-08
HDDS-3135. Enable topology acceptance test added in HDDS-3084 to read data when racks stopped

sodonnel:

With HDDS-3270 now committed, pipelines are not created until a defined number of DNs have registered with SCM. This should prevent non-rack-aware pipelines being created.

We can now enabled the test added in HDDS-3084 (but left disabled) as it should no longer fail intermittently.

As part of this change I have also adjusted `hdds.scm.safemode.min.datanode` to be 6 rather than 3 for this test to ensure both racks have registered.

https://issues.apache.org/jira/browse/HDDS-3135

## How was this patch tested?

Ran the acceptance test locally 10 times and got no failures.

I plan to run it several times here on github too, so we can be sure it does not fail intermittently.


sodonnel:
Thanks for the review. I will go ahead and merge this one now.

PR #799, opened 2020-04-09
HDDS-3360. Duplicate Ozone Client in renew/cancel token handlers

adoroszlai:

1. Avoid creating duplicate Ozone Client for token operations.
2. Make URI optional for `token get`

This PR is based on top of #775, so that one should be merged first.  The only change here is 480b467302ea1b6175b27c673c8f3ee6c8fbe2e5.

https://issues.apache.org/jira/browse/HDDS-3360

## How was this patch tested?

Added smoke test for token operations.


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/Handler.java
```diff
@@ -58,15 +58,25 @@ public OzoneConfiguration createOzoneConfiguration() {
...
 
-  protected abstract OzoneAddress getAddress() throws OzoneClientException;
+  protected OzoneAddress getAddress() throws OzoneClientException {
+    return new OzoneAddress();
+  }
 
   protected abstract void execute(OzoneClient client, OzoneAddress address)
       throws IOException, OzoneClientException;
 
+  protected boolean isApplicable() {
```
xiaoyuyao, line 68: NIT: isApplicable is very generic, can this be more specific? when it return null (e.g., the user input miss some arguments), do we ensure context related help returned from the console output? Maybe we can add a negative case in the new acceptance test.

adoroszlai: Thanks @xiaoyuyao for the review.

> NIT: isApplicable is very generic, can this be more specific?

`Handler` itself is very generic, at that level I don't think the method can be more specific.  Added javadoc for `isApplicable` to explain its intended usage.  Also changed few more classes to override it and got rid of their own explicit `if` check in `execute`.

> when it return null (e.g., the user input miss some arguments),

Note that `Void call()` always returns `null`, it's not a sign of success/failure.

> do we ensure context related help returned from the console output? Maybe we can add a negative case in the new acceptance test.

I have added such negative test case for token operations.

xiaoyuyao: LGTM, +1. Thanks for the update!

adoroszlai:

PR #800, opened 2020-04-09
HDDS-3093. Allow forced overwrite of local file

adoroszlai:

Add `-f` (`--force`) flag for `ozone sh key get`, to allow force overwrite of local file.  This is useful for testing:

 * download the same key multiple times without having to manually delete the target
 * download to `/dev/null` (which already exists), only to verify existence and checksum of the key

Hash cannot be calculated for `/dev/null`.

This PR is based on top of #775, so that one should be merged first.  The only change here is 2ca1fe23cf72eb813154c17260868b6ecd030801.

https://issues.apache.org/jira/browse/HDDS-3093

## How was this patch tested?

```
$ ozone freon ockg -n 1 -p test
...

$ ozone sh key get /vol1/bucket1/test/0 /tmp/asdf

$ ozone sh key get /vol1/bucket1/test/0 /tmp/asdf
/tmp/asdf exists. Download would overwrite an existing file. Aborting.

$ ozone sh key get --force /vol1/bucket1/test/0 /tmp/asdf

$ ozone sh --verbose key get --force /vol1/bucket1/test/0 /tmp/asdf
Volume Name : vol1
Bucket Name : bucket1
Key Name : test/0
Downloaded file hash : 528e4f96633a453d18ee01fdb5d7c536

$ ozone sh --verbose key get --force /vol1/bucket1/test/0 /dev/null
Volume Name : vol1
Bucket Name : bucket1
Key Name : test/0
```

Added smoke test for overwrite case.


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/GetKeyHandler.java
```diff
@@ -80,7 +88,7 @@ protected void execute(OzoneClient client, OzoneAddress address)
       IOUtils.copyBytes(input, output, chunkSize);
     }
 
-    if (isVerbose()) {
+    if (isVerbose() && !"/dev/null".equals(dataFile.getAbsolutePath())) {
```
xiaoyuyao, line 91: Hardcoded path like "/dev/null" may not work in other platforms like Windows.

adoroszlai: I think that's OK.    This check does not affect functionality and we're not relying on any specific property of `/dev/null` here.  The worst thing that can happen is that if someone issues `ozone sh --verbose key get ... /dev/null` on Windows, hash is not printed.  But I doubt that's a valid file on Windows anyway.

adoroszlai:

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/GetKeyHandler.java
xiaoyuyao: Sounds good to me. +1

adoroszlai:
Thanks @dineshchitlangia for reviewing and merging it.

PR #803, opened 2020-04-09
Revert "HDDS-3291. Write operation when both OM followers are shutdown."

mukul1987:
Reverts apache/hadoop-ozone#733

Reverting this as this causes other Hadoop application in the cluster to use the new IPC timeout values.


bharatviswa504:
Thank You @mukul1987 for taking care of this issue.

bharatviswa504:
This is a revert and test failures are unrelated. So I will go ahead and commit this.

PR #806, opened 2020-04-10
HDDS-3374. OMVolumeSetOwnerRequest doesn't check if user is already the owner

smengcl:

`OMVolumeSetOwnerRequest` doesn't check if the user is already the owner.

If the user is already the owner, it shouldn't proceed to the update logic, otherwise the resulting volume list for that user in `UserVolumeInfo` would have duplicate volume entry. As demonstrated in the test case.

https://issues.apache.org/jira/browse/HDDS-3374

## How was this patch tested?

The test case (`setOwner` twice on the same volume with the same user) should pass.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeRequest.java
```diff
@@ -106,6 +106,13 @@ protected UserVolumeInfo addVolumeToOwnerList(UserVolumeInfo volumeList,
...
+    // Sanity check, a user should not own same volume twice
+    //  TODO: May want to remove this due to perf if user owns a lot of volumes.
+    if (prevVolList.contains(volume)) {
+      throw new IOException("Invalid operation: User " + owner +
+          " is about to own a same volume " + volume + " twice!" +
+          " Check for DB consistency error.");
+    }
 
     // Add the new volume to the list
     prevVolList.add(volume);
```
dineshchitlangia, line 118: Instead of throwing exception here, I was wondering if we can instead perform the subsequent "add new volume to list" if !prevVolList.contains(volume) and complement with WARN log.

```suggestion
    // Avoid adding a user to the same volume twice
    if (!prevVolList.contains(volume)) {
      // Add the new volume to the list
      prevVolList.add(volume);
      UserVolumeInfo newVolList = UserVolumeInfo.newBuilder()
          .setObjectID(objectID)
          .setUpdateID(txID)
          .addAllVolumeNames(prevVolList).build();
    }
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeSetOwnerRequest.java
```diff
@@ -143,6 +143,11 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       oldOwner = omVolumeArgs.getOwnerName();
 
+      if (oldOwner.equals(newOwner)) {
+        throw new OMException("Owner of volume " + volume + " is already " +
+            newOwner, OMException.ResultCodes.ACCESS_DENIED);
+      }
+
```
dineshchitlangia, line 150: Like previous comment, we can make similar change here too.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeRequest.java
```diff
@@ -106,6 +106,13 @@ protected UserVolumeInfo addVolumeToOwnerList(UserVolumeInfo volumeList,
       objectID = volumeList.getObjectID();
     }
 
+    // Sanity check, a user should not own same volume twice
+    //  TODO: May want to remove this due to perf if user owns a lot of volumes.
+    if (prevVolList.contains(volume)) {
```
bharatviswa504, line 111: I am thinking instead of returning an error like AccessDenied, can we return a code that this owner is already owner for this volume.
Because access Denied looks not proper here.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerSetOwner.java
```diff
@@ -0,0 +1,123 @@
...
+    String omId = UUID.randomUUID().toString();
+    conf.setInt(OZONE_OPEN_KEY_EXPIRE_THRESHOLD_SECONDS, 2);
+    conf.set(OZONE_ADMINISTRATORS, "user1");
+    conf.setInt(OZONE_SCM_RATIS_PIPELINE_LIMIT, 10);
+
+    // Use native impl here, default impl doesn't do actual checks
+    conf.set(OZONE_ACL_AUTHORIZER_CLASS, OZONE_ACL_AUTHORIZER_CLASS_NATIVE);
+    // Note: OM doesn't support live config reloading
+    conf.setBoolean(OZONE_ACL_ENABLED, aclEnabled);
+
```
bharatviswa504, line 75: We don't need an IT test to test this behavior. We have a UT TestOMVolumeSetOwnerRequest which we can use this to cover this test.

smengcl: Yes I acknowledge that. This integration test is a quick hack from the https://github.com/apache/hadoop-ozone/pull/696 because I discovered this issue when debugging that PR. Just a POC for now. Will figure out a way to write this in `TestOMVolumeSetOwnerRequest`.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeRequest.java
smengcl: I was also thinking of adding a new type of exception. But now I think @dineshchitlangia 's suggestion might be better -- handle it silently and maybe throw a warning.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeRequest.java
smengcl: Will do. Thanks :)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeRequest.java
bharatviswa504: Can we document this behavior, so that the end-user knows about this behavior?

smengcl: IMO the client shouldn't notice this. The warnings should only show up on the server (if any).

smengcl:
Test looks good. Should be good to go. Please take a look at the latest changes.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/volume/OMVolumeSetOwnerResponse.java
```diff
@@ -55,11 +56,27 @@ public OMVolumeSetOwnerResponse(@Nonnull OMResponse omResponse,
 
   /**
    * For when the request is not successful or it is a replay transaction.
```
xiaoyuyao, line 58: should we remove this comment as checkStatusNotOK is now conditioned with success=true? Also, do we redirect existing callers for success=false case appropriately?

xiaoyuyao: I think the reply response is taken care by createReplayOMResponse now. So we can safely clean up the comments above.

smengcl: I believe the comment is still accurate. There are four places that calls this constructor at the moment:

- INVALID_REQUEST
```java
    // In production this will never happen, this request will be called only
    // when we have ownerName in setVolumePropertyRequest.
    if (!setVolumePropertyRequest.hasOwnerName()) {
      omResponse.setStatus(OzoneManagerProtocolProtos.Status.INVALID_REQUEST)
          .setSuccess(false);
      return new OMVolumeSetOwnerResponse(omResponse.build());
    }
```
- Replay
```java
      // Check if this transaction is a replay of ratis logs.
      // If this is a replay, then the response has already been returned to
      // the client. So take no further action and return a dummy
      // OMClientResponse.
      if (isReplay(ozoneManager, omVolumeArgs, transactionLogIndex)) {
        LOG.debug("Replayed Transaction {} ignored. Request: {}",
            transactionLogIndex, setVolumePropertyRequest);
        return new OMVolumeSetOwnerResponse(createReplayOMResponse(omResponse));
      }
```
- newOwner is the same as oldOwner (added in this PR)
```java
      // Return OK immediately if newOwner is the same as oldOwner.
      if (oldOwner.equals(newOwner)) {
        LOG.warn("Volume '{}' owner is already user '{}'.", volume, oldOwner);
        omResponse.setStatus(OzoneManagerProtocolProtos.Status.OK)
          .setMessage(
            "Volume '" + volume + "' owner is already '" + newOwner + "'.")
          .setSuccess(false);
        return new OMVolumeSetOwnerResponse(omResponse.build());
      }
```
- IOException
```java
    } catch (IOException ex) {
      exception = ex;
      omClientResponse = new OMVolumeSetOwnerResponse(
          createErrorOMResponse(omResponse, exception));
    } finally {
```

smengcl: Though in all four cases the success flag is set to false, rendering the condition around `checkStatusNotOK()` pointless.

smengcl: Just BTW in a discussion with @bharatviswa504 we concluded that the reason of setting success to `false` for "newOwner equals to oldOwner" case is that in [`addAcl`](https://github.com/apache/hadoop-ozone/blob/ebc616061da6f337626ee85e4282230aedfd748a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/acl/AclOption.java#L58-L62) we have some logic to print different results on the client depending on the success flag. Though the success flag of `setOwner` isn't utilized at the moment, we might be able to use it later.

smengcl: > Though in all four cases the success flag is set to false, rendering the condition around `checkStatusNotOK()` pointless.

Maybe we should just remove the call to `checkStatusNotOK()` in this constructor now that we established this new use case. What do you think?

bharatviswa504: Just to add more info, we set the success flag and also there is a response which we return that is used by the end client which has a response flag which is set based on the result of the operation. 

```
  OMClientResponse onSuccess(OMResponse.Builder omResponse,
      OmBucketInfo omBucketInfo, boolean operationResult) {
    omResponse.setSuccess(operationResult);
    omResponse.setAddAclResponse(AddAclResponse.newBuilder()
         .setResponse(operationResult));
```

But in OmVolumeSetOwner case, there is no such flag. If we can introduce a flag in SetVolumePropertyResponse, we can use the success flag.

      omResponse.setSetVolumePropertyResponse(
          SetVolumePropertyResponse.newBuilder().build());

smengcl: > Just to add more info, we set the success flag and also there is a response which we return that is used by the end client which has a response flag which is set based on the result of the operation.
> 
> ```
>   OMClientResponse onSuccess(OMResponse.Builder omResponse,
>       OmBucketInfo omBucketInfo, boolean operationResult) {
>     omResponse.setSuccess(operationResult);
>     omResponse.setAddAclResponse(AddAclResponse.newBuilder()
>          .setResponse(operationResult));
> ```
> 
> But in OmVolumeSetOwner case, there is no such flag. If we can introduce a flag in SetVolumePropertyResponse, we can use the success flag.
> 
> ```
>   omResponse.setSetVolumePropertyResponse(
>       SetVolumePropertyResponse.newBuilder().build());
> ```

Good idea we can add `optional bool response = 1;` in [here](https://github.com/apache/hadoop-ozone/blob/15db251f16236228c6596253dab6946494fc8f5b/hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto#L425-L427) in another jira.

smengcl: Discussed with @xiaoyuyao and we agree the comment don't need to be changed at the moment.

Resolving this conversation. Will commit shortly.

smengcl:
Thanks @dineshchitlangia @bharatviswa504 @xiaoyuyao for the review and comments!

PR #809, opened 2020-04-11
HDDS-2576. Handle InterruptedException in OzoneManagerDoubleBuffer

danidelvalle:

Handle InterruptedException in ratis related files:
* OzoneManagerDoubleBuffer: https://sonarcloud.io/project/issues?id=hadoop-ozone&issues=AW5md-VxKcVY8lQ4Zrtu&open=AW5md-VxKcVY8lQ4Zrtu
* ~~OzoneManagerRatisClient~~: removed in [HDDS-1991](https://issues.apache.org/jira/browse/HDDS-1991)

https://issues.apache.org/jira/browse/HDDS-2576

## How was this patch tested?

Compiles successfully without any errors in local.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
@@ -278,7 +278,7 @@ public class OzoneManagerDoubleBuffer {
         // Wait for daemon thread to exit
         daemon.join();
       } catch (InterruptedException e) {
-        LOG.error("Interrupted while waiting for daemon to exit.", e);
+        LOG.debug("Interrupted while waiting for daemon to exit.", e);
       }
 
       // stop metrics.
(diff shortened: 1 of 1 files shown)


elek:
Thanks to post this patch @danidelvalle 

I am trying to understand this part of the code, and it seems to be false positive. As the daemon thread is interrupted right before, I think it's safe to ignore the `InterruptedException`.

Especially as with the proposed approach the metrics (after these lines) won't be stopped. 

I am not sure what is the good approach here, I think it's more safe to keep it as is, but I can be convinced...

elek:
My understanding was wrong. The metric code will be executed all the time, this code just flags the current thread as interrupted and disable the usage of any sleep/wait.

But it's still an open question: it seems to be the normal flow of a shutdown, I don't think that we need to mark the thread itself as interrupted.

danidelvalle:
@elek I agree this is probably a false possitive and marking the thread as interrupted is quite unnecessary. However, I guess it would worth it just for homogeneity and to get rid of the sonar's bug.

elek:
> However, I guess it would worth it just for homogeneity and to get rid of the sonar's bug.

I am fine to get it for homogeneity, unless it cause problems in the behavior of the application. Marking a thread as interrupted during the normal operation can cause unexpected results (during unit test, or run).

For me, it's hard to estimate the risk, and it seems to be unnecessary. Wouldn't be more safe to mark the violation in Sonar as false positive?

danidelvalle:
> For me, it's hard to estimate the risk, and it seems to be unnecessary. Wouldn't be more safe to mark the violation in Sonar as false positive?

It would, you've convinced me. Thanks @elek, I'm declining the PR.

adoroszlai:
Instead of marking the violation as false positive in SonarCloud, I think it's more transparent to do it by adding `@SuppressWarnings("squid:S2142")` to the method (with some explanation).  It also lets @danidelvalle get credit for the contribution.  Please reopen and update the PR if you agree.

PR #812, opened 2020-04-11
HDDS-3161. Block illegal characters when creating keys.

cku328:

Added illegal character check (using regular expressions) for key name when creating key.

For the definition of illegal characters, I refer to [Amazon S3's object key naming guide](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html), which specifies the characters to avoid.

https://issues.apache.org/jira/browse/HDDS-3161

## How was this patch tested?

- Ran UTs.
- Using Ozone Shell to create a key with illegal characters in the name, errors can be detected, like this: 
test


elek:
Thanks @cku328 to create this patch.  I am fine with following the S3 naming convention but not sure how will it work when buckets will be accessed from NFS/ozone fs. 

What do you think @arp7 , can we do this restriction?

From technical point of view: Didn't check the patch very closely, but it seems to be a client side check. I think it should be checked on the server side, too (but it might be included)...

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
```diff
@@ -196,6 +196,27 @@ public static void verifyResourceName(String... resourceNames) {
...
+   *
+   * @param keyName key name to be validated
+   *
+   * @throws IllegalArgumentException
+   */
+  public static void verifyKeyName(String keyName) {
+    if (keyName == null) {
+      throw new IllegalArgumentException("Key name is null");
+    }
+    String regex = "^[^^{}<>^?%~#`\\[\\]\\|\\\\(\\x80-\\xff)]$";
```
mukul1987, line 210: Please move this regex to OzoneConsts and please add a simplified example on what characters are not allowed.

cku328:
Thanks to @elek for the comments and @mukul1987 for the review.

I referenced the commit and review comments for this pull request, implemented the key name check at `OMKeyCreateRequest#preExecute`, and applied it the same way to other requests (e.g.: CreateFile, RenameKey). 

In the newly submitted patch, I fixed some of the requests for missing check names, and I tested the put file through OzoneFS and it worked.

test

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
cku328: Thanks, I've moved the regex to OzoneConsts and explained it.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
```diff
@@ -328,5 +328,19 @@ private OzoneConsts() {
...
+   * This regular expression is used to check if key name
+   * contains illegal characters when creating/renaming key.
+   *
+   * Avoid the following characters in a key name:
+   * "\", "{", "}", "^", "<", ">", "#", "|", "%", "`", "[", "]", "~", "?"
+   * and Non-printable ASCII characters (128–255 decimal characters).
+   * https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
+   */
+  public static final String KEYNAME_AVOID_CHARACTERS_REGEX =
+          "^[^^{}<>^?%~#`\\[\\]\\|\\\\(\\x80-\\xff)]$";
```
adoroszlai, line 343: Constants for regex patterns should be pre-compiled (`Pattern.compile(...)`).

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
```diff
@@ -196,6 +196,27 @@ public static void verifyResourceName(String... resourceNames) {
...
+   * @throws IllegalArgumentException
+   */
+  public static void verifyKeyName(String keyName) {
+    if (keyName == null) {
+      throw new IllegalArgumentException("Key name is null");
+    }
+    for (int index = 0; index < keyName.length(); index++) {
+      char currChar = keyName.charAt(index);
+      if (!(Character.toString(currChar)
+              .matches(OzoneConsts.KEYNAME_AVOID_CHARACTERS_REGEX))){
```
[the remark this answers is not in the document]
cku328: Thanks @adoroszlai for the comment.
Checking each character independently with regular expressions to know exactly which one is not allowed. But I feel the same way, independently check each character is expensive, so I modified it to use full-string to check.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
cku328: Thanks! I fixed it.

cku328:
Thanks @adoroszlai for the review.
In the new patch, this keyname check is optional and it can be enabled by Ozone global config (default is false).

In `ozone-site.xml`, add the following snippet to enable checking:
```
<property>
  <name>ozone.om.keyname.character.check.enabled</name>
  <value>true</value>
</property>
```

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
```diff
@@ -196,6 +196,24 @@ public static void verifyResourceName(String... resourceNames) {
...
+   * @throws IllegalArgumentException
+   */
+  public static void verifyKeyName(String keyName) {
+    if (keyName == null) {
+      throw new IllegalArgumentException("Key name is null");
+    }
+    if(!OzoneConsts.KEYNAME_ILLEGAL_CHARACTER_CHECK_REGEX
+            .matcher(keyName).matches()){
+      throw new IllegalArgumentException("key name contains " +
+              "illegal characters.");
```
adoroszlai, line 213: I would like to propose making this message consistent with the one in `OmUtils`, so that `ozone fs -put` and `ozone sh key put` both print the same error.  I think it can be most simply done by moving the message from `OmUtils` here and letting the `OMException` propagate the underlying exception's message.

```suggestion
      throw new IllegalArgumentException("Invalid key name: " + keyName);
```

If you agree, please feel free to wait for others' reviews before making this change, to reduce the number of rounds.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -506,4 +506,17 @@ public static long getOMClientRpcTimeOut(Configuration configuration) {
...
+
+  /**
+   * Verify key name is a valid name.
+   */
+  public static void validateKeyName(String keyName)
+          throws OMException {
+    try {
+      HddsClientUtils.verifyKeyName(keyName);
+    } catch (IllegalArgumentException e) {
+      throw new OMException("Invalid key name: " + keyName,
```
adoroszlai, line 518: Corresponding to the suggestion for `HddsClientUtils`:

```suggestion
      throw new OMException(e.getMessage(),
```

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
cku328: Thanks for the suggestion, I agree with you. 
I'll wait for others' reviews and suggestions before I make this change.

cku328:
Update message of exception and resolve conflict. (latest master-branch #399)

elek:
This patch seems to be reviewed and all the comments are addressed. Will merge it if no objections.

I resolved the conflicts and started a new build to get a green build with the latest master.

elek:
:musical_note:  Green build, green build, tada-tada-tada... :musical_note: Merging it right now, tada-tada-tada :musical_note: 

Thanks the patch @cku328 (and sorry for the long review time) and the review @adoroszlai

PR #814, opened 2020-04-13
HDDS-3286. BasicOzoneFileSystem  support batchDelete.

captainzmc:

Currently delete file is to get all the keys in the directory, and then delete one by one. This makes for poor performance.
By tested the deletion directory with 100,000 files, which took 3718.70 sec. And rename it took 7327.936 sec.
Using this PR, when batch-size is set to 100, the time of delete and rename's directory of 100,000  files is 62.498 sec and 46.002 sec. Performance improved nearly 100 times。

Rename has the same problem, and I'll add rename to the other PR.

##What is the link to the Apache JIRA
https://issues.apache.org/jira/browse/HDDS-3286

## How was this patch tested?
Added UT
Existing acceptance tests.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -393,6 +393,10 @@ public final class OzoneConfigKeys {
       "ozone.s3.token.max.lifetime";
   public static final String OZONE_S3_AUTHINFO_MAX_LIFETIME_KEY_DEFAULT = "3m";
 
+  public static final String OZONE_FS_ITERATE_BATCH_SIZE =
+      "ozone.fs.iterate.batch-size";
+  public static final int OZONE_FS_ITERATE_BATCH_SIZE_DEFAULT = 100;
+
   // Ozone Client Retry and Failover configurations
   public static final String OZONE_CLIENT_FAILOVER_MAX_ATTEMPTS_KEY =
       "ozone.client.failover.max.attempts";
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1937,6 +1937,14 @@
   </property>
   
   <property>
+    <name>ozone.fs.iterate.batch-size</name>
+    <value>100</value>
+    <tag>OZONE, OZONEFS</tag>
+    <description>
+      Iterate batch size of delete when use BasicOzoneFileSystem.
+    </description>
+  </property>
+  <property>
     <name>ozone.manager.db.checkpoint.transfer.bandwidthPerSec</name>
     <value>0</value>
     <tag>OZONE</tag>
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
@@ -382,6 +382,21 @@ public class OzoneBucket extends WithMetadata {
     proxy.deleteKey(volumeName, name, key);
   }
 
+  /**
+   * Deletes the given list of keys from the bucket.
+   * @param keyList List of the key name to be deleted.
+   * @throws IOException
+   */
+  public void deleteKeys(List<String> keyList) throws IOException {
+    proxy.deleteKeys(volumeName, name, keyList);
+  }
+
+  /**
+   * Rename the keyname from fromKeyName to toKeyName.
+   * @param fromKeyName The original key name.
+   * @param toKeyName New key name.
+   * @throws IOException
+   */
   public void renameKey(String fromKeyName, String toKeyName)
       throws IOException {
     proxy.renameKey(volumeName, name, fromKeyName, toKeyName);
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -45,6 +45,7 @@ import org.apache.hadoop.test.GenericTestUtils;
 
 import org.apache.commons.io.IOUtils;
 
+import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_FS_ITERATE_BATCH_SIZE;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
@@ -171,6 +172,7 @@ public class TestOzoneFileSystem {
 
     testCreateDoesNotAddParentDirKeys();
     testDeleteCreatesFakeParentDir();
+    testFileDelete();
     testNonExplicitlyCreatedPathExistsAfterItsLeafsWereRemoved();
 
     testRenameDir();
@@ -204,6 +206,8 @@ public class TestOzoneFileSystem {
 
     // Set the fs.defaultFS and start the filesystem
     conf.set(CommonConfigurationKeysPublic.FS_DEFAULT_NAME_KEY, rootPath);
+    // Set the number of keys to be processed during batch operate.
+    conf.setInt(OZONE_FS_ITERATE_BATCH_SIZE, 5);
     fs = FileSystem.get(conf);
   }
 
@@ -260,6 +264,44 @@ public class TestOzoneFileSystem {
     String parentKey = o3fs.pathToKey(parent) + "/";
     OzoneKeyDetails parentKeyInfo = getKey(parent, true);
     assertEquals(parentKey, parentKeyInfo.getName());
+  }
+
+  private void testFileDelete() throws Exception {
+    Path grandparent = new Path("/testBatchDelete");
+    Path parent = new Path(grandparent, "parent");
+    Path childFolder = new Path(parent, "childFolder");
+    // BatchSize is 5, so we're going to set a number that's not a
+    // multiple of 5. In order to test the final number of keys less than
+    // batchSize can also be deleted.
+    for (int i = 0; i < 8; i++) {
+      Path childFile = new Path(parent, "child" + i);
+      Path childFolderFile = new Path(childFolder, "child" + i);
+      ContractTestUtils.touch(fs, childFile);
+      ContractTestUtils.touch(fs, childFolderFile);
+    }
+
+    assertTrue(fs.listStatus(grandparent).length == 1);
+    assertTrue(fs.listStatus(parent).length == 9);
+    assertTrue(fs.listStatus(childFolder).length == 8);
+
+    Boolean successResult = fs.delete(grandparent, true);
+    assertTrue(successResult);
+    assertTrue(!o3fs.exists(grandparent));
+    for (int i = 0; i < 8; i++) {
+      Path childFile = new Path(parent, "child" + i);
+      // Make sure all keys under testBatchDelete/parent should be deleted
+      assertTrue(!o3fs.exists(childFile));
+
+      // Test to recursively delete child folder, make sure all keys under
+      // testBatchDelete/parent/childFolder should be deleted.
+      Path childFolderFile = new Path(childFolder, "child" + i);
+      assertTrue(!o3fs.exists(childFolderFile));
+    }
+    // Will get: WARN  ozone.BasicOzoneFileSystem delete: Path does not exist.
+    // This will return false.
+    Boolean falseResult = fs.delete(parent, true);
+    assertFalse(falseResult);
+
   }
 
   private void testListStatus() throws Exception {
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
@@ -43,7 +43,7 @@ import org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerRatisUtils;
 import org.apache.hadoop.ozone.om.response.OMClientResponse;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos
-    .UnDeletedKeysResponse;
+    .DeleteKeysResponse;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos
     .OMRequest;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos
@@ -237,13 +237,10 @@ public abstract class OMClientRequest implements RequestAuditor {
       @Nonnull IOException ex, @Nonnull Set<OmKeyInfo> unDeletedKeys) {
     omResponse.setSuccess(false);
     StringBuffer errorMsg = new StringBuffer();
-    errorMsg.append(exceptionErrorMessage(ex) + "\n The Keys not deleted: ");
-    UnDeletedKeysResponse.Builder resp =
-        UnDeletedKeysResponse.newBuilder();
+    DeleteKeysResponse.Builder resp = DeleteKeysResponse.newBuilder();
     for (OmKeyInfo key : unDeletedKeys) {
       if(key != null) {
-        resp.addKeyInfo(key.getProtobuf());
-        errorMsg.append(key.getObjectInfo() + "\n");
+        resp.addUnDeletedKeys(key.getProtobuf());
       }
     }
     if (errorMsg != null) {
@@ -252,7 +249,7 @@ public abstract class OMClientRequest implements RequestAuditor {
     // TODO: Currently all delete operations in OzoneBucket.java are void. Here
     //  we put the List of unDeletedKeys into Response. These KeyInfo can be
     //  used to continue deletion if client support delete retry.
-    omResponse.setUnDeletedKeysResponse(resp.build());
+    omResponse.setDeleteKeysResponse(resp.build());
     omResponse.setStatus(OzoneManagerRatisUtils.exceptionToResponseStatus(ex));
     return omResponse.build();
   }
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Set;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
@@ -35,11 +36,14 @@ import org.apache.hadoop.ozone.audit.AuditLogger;
 import org.apache.hadoop.ozone.audit.AuditMessage;
 import org.apache.hadoop.ozone.om.OzoneManager;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
+import org.apache.hadoop.ozone.om.helpers.OmKeyInfo;
 import org.apache.hadoop.ozone.om.helpers.WithObjectID;
 import org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerDoubleBufferHelper;
 import org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerRatisUtils;
 import org.apache.hadoop.ozone.om.response.OMClientResponse;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos
+    .UnDeletedKeysResponse;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos
     .OMRequest;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos
@@ -216,6 +220,39 @@ public abstract class OMClientRequest implements RequestAuditor {
     if (errorMsg != null) {
       omResponse.setMessage(errorMsg);
     }
+    omResponse.setStatus(OzoneManagerRatisUtils.exceptionToResponseStatus(ex));
+    return omResponse.build();
+  }
+
+  /**
+   * Set parameters needed for return error response to client.
+   *
+   * @param omResponse
+   * @param ex         - IOException
+   * @param unDeletedKeys    - Set<OmKeyInfo>
+   * @return error response need to be returned to client - OMResponse.
+   */
+  protected OMResponse createOperationKeysErrorOMResponse(
+      @Nonnull OMResponse.Builder omResponse,
+      @Nonnull IOException ex, @Nonnull Set<OmKeyInfo> unDeletedKeys) {
+    omResponse.setSuccess(false);
+    StringBuffer errorMsg = new StringBuffer();
+    errorMsg.append(exceptionErrorMessage(ex) + "\n The Keys not deleted: ");
+    UnDeletedKeysResponse.Builder resp =
+        UnDeletedKeysResponse.newBuilder();
+    for (OmKeyInfo key : unDeletedKeys) {
+      if(key != null) {
+        resp.addKeyInfo(key.getProtobuf());
+        errorMsg.append(key.getObjectInfo() + "\n");
+      }
+    }
+    if (errorMsg != null) {
+      omResponse.setMessage(errorMsg.toString());
+    }
+    // TODO: Currently all delete operations in OzoneBucket.java are void. Here
+    //  we put the List of unDeletedKeys into Response. These KeyInfo can be
+    //  used to continue deletion if client support delete retry.
+    omResponse.setUnDeletedKeysResponse(resp.build());
     omResponse.setStatus(OzoneManagerRatisUtils.exceptionToResponseStatus(ex));
     return omResponse.build();
   }
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/protocol/ClientProtocol.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/protocol/ClientProtocol.java
@@ -295,6 +295,17 @@ public interface ClientProtocol {
       throws IOException;
 
   /**
+   * Deletes keys through the list.
+   * @param volumeName Name of the Volume
+   * @param bucketName Name of the Bucket
+   * @param keyNameList List of the Key
+   * @throws IOException
+   */
+  void deleteKeys(String volumeName, String bucketName,
+                  List<String> keyNameList)
+      throws IOException;
+
+  /**
    * Renames an existing key within a bucket.
    * @param volumeName Name of the Volume
    * @param bucketName Name of the Bucket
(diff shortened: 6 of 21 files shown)


mukul1987:
Please have a look at https://issues.apache.org/jira/browse/HDDS-2939. The problem is being addressed there.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/RenameInfo.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.ozone.om.helpers;
+
+/**
+ * The data interface needed to the rename operation.
+ */
+public class RenameInfo {
+  private String  fromKey;
```
rakeshadr, line 24: Should we need `fromKey` member variable? Is this duplicate info and we can get it from `fromKeyValue.getKeyName();` , right?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/RenameInfo.java
```diff
@@ -0,0 +1,45 @@
...
+    this.fromKeyValue = fromKeyValue;
+    this.toKey = toKey;
+  }
+
+  public String getFromKey() {
+    return fromKey;
+  }
+
+  public OmKeyInfo getFromKeyValue() {
+    return fromKeyValue;
```
rakeshadr, line 39: Can we do` fromKeyValue.getKeyName(); `?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -748,19 +751,34 @@ public String toString() {
...
       if (status.isDirectory()) {
         LOG.trace("Iterating directory:{}", pathKey);
         while (keyIterator.hasNext()) {
           BasicKeyInfo key = keyIterator.next();
           LOG.trace("iterating key:{}", key.getName());
-          if (!processKey(key.getName())) {
+          if (!key.getName().equals("")) {
+            keyList.add(key.getName());
+          }
+          int batchSize = getConf().getInt("ozone.fs.iterate.batch-size", 1);
```
rakeshadr, line 763: Instead of hard coding, can you do `OzoneConfigKeys.OZONE_FS_ITERATE_BATCH_SIZE` ?

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyRenameRequest.java
```diff
@@ -327,11 +331,13 @@ private OMRequest doPreExecute(OMRequest originalOmRequest) throws Exception {
    * @return OMRequest
    */
   private OMRequest createRenameKeyRequest(String toKeyName) {
-    KeyArgs keyArgs = KeyArgs.newBuilder().setKeyName(keyName)
+    Map<String, String> renameKeyMap = new HashMap<>();
```
rakeshadr, line 334: General comment for both `delete and rename `batch apis.

Please add some test cases with `all, partial, no success` behaviors. Thanks!

captainzmc:
> Please have a look at https://issues.apache.org/jira/browse/HDDS-2939. The problem is being addressed there.

@mukul1987 Thanks for the tip.  After looking at the design documentation, I found that is a better implementation. I'm going to colse this PR.

captainzmc:
> Thanks for the good work.
> Added few suggestions and comments. Please go through it. Thanks!

Hi @rakeshadr Thank for you review. By looking at the design documentation of https://issues.apache.org/jira/browse/HDDS-2939. I found that the HDDS-2939 works better, and I'm going to colse this PR.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
```diff
@@ -116,51 +119,53 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-        throw new OMException("Key not found", KEY_NOT_FOUND);
+      Table<String, OmKeyInfo> keyTable = omMetadataManager.getKeyTable();
+      for (String keyName : keyNameList) {
+        // check Acl
+        checkKeyAcls(ozoneManager, volumeName, bucketName, keyName,
+                IAccessAuthorizer.ACLType.DELETE, OzoneObj.ResourceType.KEY);
+        String objectKey = omMetadataManager.getOzoneKey(
+                volumeName, bucketName, keyName);
+        OmKeyInfo omKeyInfo = keyTable.get(objectKey);
+        if (omKeyInfo == null) {
```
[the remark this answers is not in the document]
captainzmc: > What if one of the key not found in the list of keys and assume there are 10keys and the 5th key not found. What is the contract batch api provides `executes all or executes till first failure or none of them` ?
> 
> Please consider `Check if this transaction is a replay of ratis logs.` this validation check also or any other exception. As there is no strict recommendations, I'm keeping this open for discussions:-).

When delete multiple keys, such as a directory, FS will get all the keys in this directory first, so at this time the key exists. But if there are two people delete a directory at the same time, this can cause problems, which can also occur when a single key is deleted before.
So here we're executing till first failure. Let the user know that the key he deleted has just been deleted by someone else. If any other error occurs, it will break and throw the exception to the user.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyArgs.java
```diff
@@ -70,6 +72,8 @@ private OmKeyArgs(String volumeName, String bucketName, String keyName,
     this.refreshPipeline = refreshPipeline;
     this.acls = acls;
     this.sortDatanodesInPipeline = sortDatanode;
+    this.keyNameList = keyNameList;
```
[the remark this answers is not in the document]
captainzmc: > Do we need `String keyName` argument ? Can you please incorporate `String keyName` argument into the `keyNameList` argument, something similar you have very well refactored for `keyNameList.add(keyName);` in deleteKey api.

The keyName in OmKeyArgs is Shared with other interfaces. Operations like create and list still need to use it.

captainzmc:
hi @rakeshadr Thank for you review. 
After negotiation, I re-opened this PR, and I modified it according to your suggestion. In addition, for batch optimization of rename, I plan to submit a PR hdds-3620 separately.

captainzmc:
Hi @ChenSammi @xiaoyuyao 
As we discussed before, because the progress of HDDS-2939 was slow,  and it's going to take a long time to merge into master. So we optimized the batch operation based on the current interface. We were able to get better performance with this PR before the HDDS-2939 merged. 
In addition, for batch rename operation, I plan to submit a PR HDDS-3620 separately based on this.

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
```diff
@@ -830,7 +832,7 @@ message LookupKeyResponse {
 
 message RenameKeyRequest{
     required KeyArgs keyArgs = 1;
```
timmylicheng, line 834: Can we have a list of KeyArgs here?

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
```diff
@@ -695,6 +695,8 @@ message KeyArgs {
 
     // This will be set by leader OM in HA and update the original request.
     optional FileEncryptionInfoProto fileEncryptionInfo = 15;
+    // This is a key list to support the batch operation of keys.
+    repeated string keyNames = 16;
```
timmylicheng, line 699: Why not remain KeyArgs as for single key?

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
```diff
@@ -848,6 +850,7 @@ message DeleteKeyResponse {
     // (similar to a cookie).
     optional uint64 ID = 3;
     optional uint64 openVersion = 4;
+    repeated KeyInfo keyInfoList = 5;
```
timmylicheng, line 853: Can we use repeated KeyArgs here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
```diff
@@ -111,51 +114,53 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+          throw new OMException("Key not found", KEY_NOT_FOUND);
+        }
+
+        // Check if this transaction is a replay of ratis logs.
+        if (isReplay(ozoneManager, omKeyInfo, trxnLogIndex)) {
```
timmylicheng, line 139: We need to iterate all omKeyInfo in the list as well here

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -157,6 +159,7 @@ public void testFileSystem() throws Exception {
```
timmylicheng, line 159: Not related to this class. But you wanna visit TestOzoneManagerHAWithData and see if HA needs a test case for deleting a list of keys.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyArgs.java
```diff
@@ -256,11 +263,16 @@ public Builder setSortDatanodesInPipeline(boolean sort) {
       return this;
     }
 
+    public Builder setKeyNameList(List<String> keyList) {
+      this.keyNameList = keyList;
+      return this;
+    }
+
     public OmKeyArgs build() {
       return new OmKeyArgs(volumeName, bucketName, keyName, dataSize, type,
```
timmylicheng, line 272: Why do we keep keyName and keyNameList at the same time?

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
ChenSammi: Agree. Use list of KeyArgs instead of add a new keyNameList field.

ChenSammi: Agree. Use list of KeyArgs instead of add a new keyNameList field.

ChenSammi: Agree. Use list of KeyArgs instead of add a new keyNameList field.

ChenSammi: Agree. Use list of KeyArgs instead of add a new keyNameList field.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -248,6 +253,34 @@ private void testDeleteCreatesFakeParentDir() throws Exception {
     assertEquals(parentKey, parentKeyInfo.getName());
   }
```
timmylicheng, line 255: It would be great if we add a failure case here like there is an unknown key in a list of known keys. We would like to also test exceptions.

captainzmc:
Hi @bharatviswa504, Could you please help me review this PR? Let's see if there is any effect on OM HA.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyArgs.java
xiaoyuyao: I think other operations such as createKey still assume a single KeyName.

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
```diff
@@ -677,7 +677,7 @@ message ListBucketsResponse {
 message KeyArgs {
     required string volumeName = 1;
     required string bucketName = 2;
-    required string keyName = 3;
```
xiaoyuyao, line 680: Change required to option will be an incompatible change.

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
```diff
@@ -830,7 +832,7 @@ message LookupKeyResponse {
 
 message RenameKeyRequest{
     required KeyArgs keyArgs = 1;
-    required string toKeyName = 2;
+    optional string toKeyName = 2;
```
xiaoyuyao, line 835: why do we need to change toKeyName from required to optional?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
```diff
@@ -382,6 +382,21 @@ public void deleteKey(String key) throws IOException {
     proxy.deleteKey(volumeName, name, key);
   }
 
+  /**
+   * Deletes the given list of keys from the bucket.
+   * @param keyList List of the key name to be deleted.
+   * @throws IOException
+   */
+  public void deleteKeys(List<String> keyList) throws IOException {
```
xiaoyuyao, line 390: When we delete a list of keys, upon failure in the middle, can we return a list of deleted keys and undeleted keys? This may not be an issue when you delete a single key but when batch deleting, it is hard to recover from the failures without that information.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
```diff
@@ -111,51 +114,53 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-      checkKeyAcls(ozoneManager, volumeName, bucketName, keyName,
-          IAccessAuthorizer.ACLType.DELETE, OzoneObj.ResourceType.KEY);
-
-      String objectKey = omMetadataManager.getOzoneKey(
-          volumeName, bucketName, keyName);
-
+      if (keyNameList.size() == 0) {
+        throw new OMException("Key not found", KEY_NOT_FOUND);
+      }
       acquiredLock = omMetadataManager.getLock().acquireWriteLock(BUCKET_LOCK,
```
xiaoyuyao, line 122: This is OK with o3fs as we will mount a single bucket. In the context of ofs where you can have multiple volume buckets under the root. This lock can't guarantee atomic across all the delete keyname list.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
```diff
@@ -187,7 +192,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
     case SUCCESS:
       omMetrics.decNumKeys();
       LOG.debug("Key deleted. Volume:{}, Bucket:{}, Key:{}", volumeName,
-          bucketName, keyName);
+          bucketName, keyNameList);
```
xiaoyuyao, line 195: This will only print the address of the keNameList. You may want to expand the list and also protect the LOG.debug with a if LOG.isDebugEnabled().

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
```diff
@@ -196,7 +201,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
     case FAILURE:
       omMetrics.incNumKeyDeleteFails();
       LOG.error("Key delete failed. Volume:{}, Bucket:{}, Key{}. Exception:{}",
-          volumeName, bucketName, keyName, exception);
+          volumeName, bucketName, keyNameList, exception);
```
xiaoyuyao, line 204: same as above. we might only want to print the key that failed in the deletion instead of the whole list upon failures. print the whole list can be a debug or trace level log.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -272,15 +272,15 @@ public boolean createDirectory(String keyName) throws IOException {
...
    * Helper method to delete an object specified by key name in bucket.
    *
-   * @param keyName key name to be deleted
+   * @param keyNameList key name list to be deleted
    * @return true if the key is deleted, false otherwise
    */
   @Override
-  public boolean deleteObject(String keyName) {
-    LOG.trace("issuing delete for key {}", keyName);
+  public boolean deleteObject(List<String> keyNameList) {
```
xiaoyuyao, line 279: Should we define a new method called deleteObjects to be backward compatible?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OzoneClientAdapter.java
```diff
@@ -48,7 +48,7 @@ OzoneFSOutputStream createFile(String key, short replication,
 
   boolean createDirectory(String keyName) throws IOException;
 
-  boolean deleteObject(String keyName);
```
xiaoyuyao, line 51: Should we define a new method called deleteObjects to be backward compatible?

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -1966,6 +1966,14 @@
       jar and false for the ozone-filesystem-lib-current.jar
     </description>
   </property>
+  <property>
+    <name>ozone.fs.iterate.batch-size</name>
+    <value>1</value>
+    <tag>OZONE, OZONEFS</tag>
+    <description>
+      iterate batch size of delete and rename when use BasicOzoneFileSystem.
```
xiaoyuyao, line 1974: Can we update the title of the JIRA to reflect batch rename?

captainzmc:
Thanks all for the reviews and suggestions. I will revise it as soon as possible.

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
captainzmc: Very good advice. Instead of modifying KeyArgs, I'll try to implement delete with a List that contains KeyArgs.

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
captainzmc: Yes, We can avoid changing it here，by use list of KeyArgs.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyArgs.java
captainzmc: I will delete keyNameList in KeyArgs, by use list of KeyArgs. To avoid changes to KeyArgs.

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
captainzmc: This place is useless for modification, I will restore this place.

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
captainzmc: I will change this place back to required and not change the KeyArgs,  by use list of KeyArgs to delete.

hadoop-hdds/common/src/main/resources/ozone-default.xml
captainzmc: I can modify the description here until rename's PR commits.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
captainzmc: Yes, because ofs and o3fs implementations are different, the compatibility issues above are caused. Later we can open a jira to track and solve these.

captainzmc:
A new commit was committed to fix the review problem:
1. Use multiple KeyArgs to delete in batches
2. Increase the batch operation interface as a whole to ensure compatibility
3, increase the UT, in TestOzoneFileSystem and TestOzoneManagerHAWithData increase test, to test the FileSystem and HA, also test the delete key does not exist
4. Fix other review issues.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
captainzmc: Here I will make sure that all buckets I used are locked to ensure compatibility with ofs.

captainzmc:
A new commit was committed to fix the review problem:
1. Use multiple KeyArgs to delete in batches.
2. Increase the batch operation interface as a whole to ensure compatibility.  
3. Increase the UT, in TestOzoneFileSystem and TestOzoneManagerHAWithData increase test, to test the FileSystem and HA, also test the delete key does not exist
4. Fix other review issues.
This PR can continue to review.

captainzmc:
This PR can continue to be reviewed. cc @timmylicheng @ChenSammi @xiaoyuyao Thanks.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMBatchKeyDeleteRequest.java
```diff
@@ -0,0 +1,226 @@
...
+        getOmRequest());
+    OMMetadataManager omMetadataManager = ozoneManager.getMetadataManager();
+    try {
+      for (KeyArgs deleteKeyArgs : deleteKeyArgsList) {
+        acquiredLockSet.add(deleteKeyArgs.getVolumeName() + "_" +
+            deleteKeyArgs.getBucketName());
+      }
+
+      for (String volumeAndVolume : acquiredLockSet) {
+        omMetadataManager.getLock().acquireWriteLock(
```
xiaoyuyao, line 117: Note when the batch size is large, this could hold the lock for a longer period of time compare with the individual lock and release.

captainzmc: I had fixed this part by getting the lock separately when the key is deleting.

captainzmc:
Fixed issues with large lock granularity. PR can continue to be reviewed.

captainzmc:
Had rebased master and fixed conflict. PR can continue to be reviewed. cc @timmylicheng @ChenSammi @xiaoyuyao Thanks.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMBatchKeyDeleteRequest.java
```diff
@@ -0,0 +1,229 @@
...
+      }
+      omClientResponse = new OMBatchKeyDeleteResponse(omResponse
+          .setDeleteBatchKeyResponse(DeleteBatchKeyResponse.newBuilder())
+          .build(),
+          omKeyInfoList, ozoneManager.isRatisEnabled());
+      result = Result.SUCCESS;
+    } catch (IOException ex) {
+      if (ex instanceof OMReplayException) {
+        result = Result.REPLAY;
+        omClientResponse = new OMKeyDeleteResponse(createReplayOMResponse(
```
ChenSammi, line 181: Can we return the list of deleted keys together with the exception to client so that client knows which keyes are deleted and which key trigger the exception?

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
```diff
@@ -401,6 +401,10 @@
       "ozone.s3.token.max.lifetime";
   public static final String OZONE_S3_AUTHINFO_MAX_LIFETIME_KEY_DEFAULT = "3m";
 
+  public static final String OZONE_FS_ITERATE_BATCH_SIZE =
```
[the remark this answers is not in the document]
captainzmc: Yes, batch Rename will also use this property.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
captainzmc: I will add unDeletedKeys and deletedKeys in exception when an exception occurs when delete keys.

captainzmc:
Status updates:
This PR can continue to be reviewed. In addition, batch Rename will also implement in the same way as this PR.

captainzmc:
Status updates:
Some changes have been made based on the new comments. This PR can continue to be reviewed.

hadoop-ozone/common/src/main/proto/OzoneManagerProtocol.proto
```diff
@@ -856,6 +864,10 @@ message DeletedKeys {
     repeated string keys = 3;
 }
 
+message DeleteKeysResponse {
```
xiaoyuyao, line 867: Usually we have a XXXRequest match with XXXResponse. Here we have DeleteKeysRequest->DeleteKeysResponse and UnDeletedKeysResponse. 

NIT: can we keep this consistent to avoid confusion for UnDeleteKeysRequest if any in the future?

```suggestion
message DeleteKeysResponse {
    repeated KeyInfo deletedKeys = 1;
    repeated KeyInfo unDeletedKeys = 2;
}```

File: hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java [hunk 27]
```diff
@@ -220,6 +224,39 @@ protected OMResponse createErrorOMResponse(
...
+        errorMsg.append(key.getObjectInfo() + "\n");
+      }
+    }
+    if (errorMsg != null) {
+      omResponse.setMessage(errorMsg.toString());
+    }
+    // TODO: Currently all delete operations in OzoneBucket.java are void. Here
+    //  we put the List of unDeletedKeys into Response. These KeyInfo can be
+    //  used to continue deletion if client support delete retry.
+    omResponse.setUnDeletedKeysResponse(resp.build());
```
xiaoyuyao, line 255: I think put the undeleted keys in OMResponse is good enough. Not sure if we want to append all the key names into the string message, which can hurt the performance when the list is huge.

captainzmc:
Thanks  @xiaoyuyao  for starting to review this PR again. I have made some modifications according to the suggestions above. In addition, Due to the conflict problem, I rebase the previous commit.
So, this PR can continue to be reviewed.

xiaoyuyao:
Thanks @captainzmc for the update. The latest change LGTM, +1.

PR #816, opened 2020-04-14
HDDS-3381. OzoneManager starts 2 OzoneManagerDoubleBuffer for HA cluster.

bharatviswa504:

Removed initialization of doubleBuffer when ratis is enabled in OzoneManagerProtocolServerSideTranslatorPB.java. As doubleBuffer is only needed in StateMachine when ratis is enabled.

https://issues.apache.org/jira/browse/HDDS-3381

## How was this patch tested?

Existing tests should cover this.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
@@ -106,6 +106,12 @@ public class OzoneManagerRequestHandler implements RequestHandler {
   private final OzoneManager impl;
   private OzoneManagerDoubleBuffer ozoneManagerDoubleBuffer;
 
+
+  public OzoneManagerRequestHandler(OzoneManager om) {
+    this.impl = om;
+    this.ozoneManagerDoubleBuffer = null;
+  }
+
   public OzoneManagerRequestHandler(OzoneManager om,
       OzoneManagerDoubleBuffer ozoneManagerDoubleBuffer) {
     this.impl = om;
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
```diff
@@ -106,6 +106,12 @@
   private final OzoneManager impl;
   private OzoneManagerDoubleBuffer ozoneManagerDoubleBuffer;
 
+
+  public OzoneManagerRequestHandler(OzoneManager om) {
```
mukul1987, line 110: Should the isRatisEnabled flag passed here and then we should assign the ozoneManagerDoubleBuffer using a getter in the OzoneManagerProtocolServerSideTranslatorPB.java ?

bharatviswa504: I think end result will be the same. If you want that way I can change it? Any advantage with that?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
```diff
@@ -106,6 +106,12 @@
...
+    this.impl = om;
+    this.ozoneManagerDoubleBuffer = null;
```
adoroszlai, line 112: Do we really need this new constructor?  `null` can be passed to the existing one.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
adoroszlai: @mukul1987 can you please answer @bharatviswa504's question?

mukul1987: Sorry I missed this earlier. the suggestion was wrt code cleanliness, yes the results will still be the same.

bharatviswa504:
This is taken care as part of HDDS-3475

PR #818, opened 2020-04-14
HDDS-3386. Remove unnecessary transitive hadoop-common dependencies on server side (addendum).

fapifta:
In HDDS-3353 we have created two new modules to manage the hadoop-common dependency and excludes related to it in a common place.
hadoop-hdds-dependency-test, and hadoop-hdds-dependency-server, similarly we had hadoop-hdds-dependency-client before.

The following modules still depend on hadoop-common for tests instead of the new test dependency:
hadoop-hdds/client
hadoop-hdds/container-service
hadoop-ozone/common
hadoop-ozone/fault-injection-test/mini-chaos-tests
hadoop-ozone/insight
hadoop-ozone/integration-test
hadoop-ozone/tools

In hadoop-dependency-client, we exclude named curator packages, similar to the new modules we should instead exclude all curator packages.

In TestVolumeSetDiskChecks.java we still have an accidental shaded import from curator: import org.apache.curator.shaded.com.google.common.collect.ImmutableSet;

The PR is proposing to fix these.

https://issues.apache.org/jira/browse/HDDS-3386

## How was this patch tested?

As a result of mvn clean install -DskipTests the build was successful.

### DIFF
--- a/hadoop-hdds/client/pom.xml
+++ b/hadoop-hdds/client/pom.xml
@@ -42,10 +42,8 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
 
     <dependency>
       <groupId>org.apache.hadoop</groupId>
-      <artifactId>hadoop-common</artifactId>
-      <version>${hadoop.version}</version>
+      <artifactId>hadoop-hdds-hadoop-dependency-test</artifactId>
       <scope>test</scope>
-      <type>test-jar</type>
     </dependency>
 
   </dependencies>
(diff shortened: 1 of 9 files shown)


adoroszlai:
Thanks @fapifta for the contribution.

adoroszlai:
Sorry, I should have triggered CI here, since unfortunately my PR was merged first, introducing a duplicate import (checkstyle violation) without merge conflict...  Fixed in follow-up commit 17456bc75bc56cea4f02a7570c0a2a2f535aada0.

PR #820, opened 2020-04-14
HDDS-3388. Add bucket encryption key info to bucket create audit log.

xiaoyuyao:

Adding bucket encryption key info to bucket create audit log.

https://issues.apache.org/jira/browse/HDDS-3388

## How was this patch tested?

Manual testing with the real cluster upon bucket creation.


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
```diff
@@ -215,6 +215,8 @@ public static Builder newBuilder() {
     auditMap.put(OzoneConsts.STORAGE_TYPE,
         (this.storageType != null) ? this.storageType.name() : null);
     auditMap.put(OzoneConsts.CREATION_TIME, String.valueOf(this.creationTime));
+    auditMap.put(OzoneConsts.BUCKET_ENCRYPTION_KEY,
+        (bekInfo != null) ? bekInfo.getKeyName() :  null);
```
vivekratnavel, line 219: Nit: extra space
```suggestion
        (bekInfo != null) ? bekInfo.getKeyName() : null);
```

PR #821, opened 2020-04-14
HDDS-3374. Addendum: OMVolumeSetOwnerRequest doesn't check if user is already the owner

smengcl:

The previous PR #806 's unit test didn't expose the potential NPE issue that the removed integration test would have exposed: the flushFuture isn't set so [this line](https://github.com/apache/hadoop-ozone/blob/1ceb51477e0bb7f44204ba8ea98d7e409277a748/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerProtocolServerSideTranslatorPB.java#L218) would raise NPE.

https://issues.apache.org/jira/browse/HDDS-3374

## How was this patch tested?

Would add back the integration test if needed.


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerSetOwner.java
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Test OzoneManager setOwner command.
+ */
+public class TestOzoneManagerSetOwner {
+
+  @Rule
+  public Timeout timeout = new Timeout(120_000);
+
+  /**
```
bharatviswa504, line 50: Can we add this test in TestOzoneRpcClient?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeSetOwnerRequest.java
```diff
@@ -147,18 +145,20 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           .setMessage(
             "Volume '" + volume + "' owner is already '" + newOwner + "'.")
           .setSuccess(false);
-        return new OMVolumeSetOwnerResponse(omResponse.build());
+        omResponse.setSetVolumePropertyResponse(
+            SetVolumePropertyResponse.newBuilder().build());
+        omClientResponse = new OMVolumeSetOwnerResponse(omResponse.build());
+        addResponseToDoubleBuffer(transactionLogIndex, omClientResponse,
+            ozoneManagerDoubleBufferHelper);
+        return omClientResponse;
```
bharatviswa504, line 153: Instead of return response here, set it to omClientResponse, and let finally block take care of it.

smengcl: done

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerSetOwner.java
smengcl: done

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -241,6 +242,19 @@ public void testOMClientProxyProvider() {
...
+  @Test
+  public void testVolumeSetOwner() throws IOException {
+    String volumeName = UUID.randomUUID().toString();
+    store.createVolume(volumeName);
+
+    String ownerName = "someRandomUser1";
+
+    ClientProtocol proxy = store.getClientProxy();
+    proxy.setVolumeOwner(volumeName, ownerName);
+    // Set owner again. Expect no NPEs nor timeouts
```
bharatviswa504, line 254: Minor NIT: We don't need this comment, as this PR fixes the bug in the code :)

smengcl: yep my intention is to remind someone in the future in case this breaks again by some new code.

smengcl: Revised comment.

PR #824, opened 2020-04-15
HDDS-2682: OM File create request does not check for existing directory with the same name

rakeshadr:
Verified the behavior in the latest code. This issue has been already taken care as part of some of the Ozone FS namespace optimisation changes.

https://issues.apache.org/jira/browse/HDDS-2682

## How was this patch tested?
Added unit test case to validate this.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -21,6 +21,7 @@ package org.apache.hadoop.fs.ozone;
 import java.io.IOException;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.concurrent.TimeoutException;
 
 import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
 import org.apache.hadoop.fs.FSDataInputStream;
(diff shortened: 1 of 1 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -65,6 +68,110 @@
...
+    OzoneBucket bucket = TestDataUtil.createVolumeAndBucket(cluster);
+    volumeName = bucket.getVolumeName();
+    bucketName = bucket.getName();
+
+    String rootPath = String.format("%s://%s.%s/",
+            OzoneConsts.OZONE_URI_SCHEME, bucket.getName(), bucket.getVolumeName());
+
+    // Set the fs.defaultFS and start the filesystem
+    conf.set(CommonConfigurationKeysPublic.FS_DEFAULT_NAME_KEY, rootPath);
+    fs = FileSystem.get(conf);
```
adoroszlai, line 97: Nit: can you please extract this logic into a helper method, to avoid duplication?

rakeshadr: Thanks for the comments. Uploaded the changes.

PR #825, opened 2020-04-15
HDDS-3395. Move the protobuf convert code to the OMHelper

maobaolong:

Clear the source file above the proto layer. 

https://issues.apache.org/jira/browse/HDDS-3395

## How was this patch tested?

Passed the ci tests.


maobaolong:
@elek Please take a look, thank you.

bharatviswa504:
I believe instead of moving this in to OMPBHelper, I believe current approach of Each Class having the logic of conversion to/from protobuf looks cleaner. With the PR approach, there will be one big huge class that does all this conversion. I feel the current way looks easy, wherein a single place has an entire code for it.

Thoughts?? If there is any strong reason in doing this way, I am happy to hear the reasons.

maobaolong:
@bharatviswa504 Thank you for your review and your reply. There are the reasons why i did this moving.

- If I don't build my repo locally, it means that the local source repo lack of the java source files generated by protobuf, for the current approach, many error appear in the classes contains reference of protobuf generated class while you view source in an IDE, whatever intellij idea and Eclipse. 
- Put the serialized and deserialized logic into the high level class, i think it is harmful for reading high level code logic, actually, some people do not care about how to serialized but care more about the data structure. Now each class have two method for serialized and deserialized purpose, and it is red color by IDE before i successful compile my repo.
- I think we can separate the class into two layer, and the serialized and deserialized logic should be in the low-level layer.
- In fact, i'm effected by HDFS and Alluxio, so, i'm very happy to discuss with you further more.

elek:
Thanks to start this discussion @maobaolong. I agree with @bharatviswa504 , I like when the de-serialization and the serialization logic is close to the data object itself and I don't like the big utility classes.

But I also understand your arguments (except the first, one, I think if the code is read in Intellij, it's a hard situation anyway).

What do you think about introducing a companion implementation / interface similar to `org.apache.hadoop.hdds.utils.db.Codec`. With this approach we can keep the serialization logic close to the data objects but they can be in a separated class.

bharatviswa504:
> Thanks to start this discussion @maobaolong. I agree with @bharatviswa504 , I like when the de-serialization and the serialization logic is close to the data object itself and I don't like the big utility classes.
> 
> But I also understand your arguments (except the first, one, I think if the code is read in Intellij, it's a hard situation anyway).
> 
> What do you think about introducing a companion implementation / interface similar to `org.apache.hadoop.hdds.utils.db.Codec`. With this approach we can keep the serialization logic close to the data objects but they can be in a separated class.

I like the proposal of the interface, and move the logic of protobuf serialization to that class. In this way, it is closer to objects.

The main reason for my comment is the same as @elek if we have the logic of protobuf in a single class, it will become one huge class and to follow it will be hard.

maobaolong:
@elek Yeah, i think your approach is better, it is not only clear the logic but also avoid a big util class.

elek:
/pending "Yeah, i think your approach is better, it is not only clear the logic but also avoid a big util class."

PR #827, opened 2020-04-15
HDDS-3397. Make jmh jar dependencies optional

elek:

jmh dependencies used by `ozone genesis` are licensed by GPL + classpath exception

It's better to make it optional and download it on demand (and exclude it from the release package).

https://issues.apache.org/jira/browse/HDDS-3397

## How was this patch tested?

 1. `find -name "*jmh*` should return without any match (**before** testing ozone genesis)
 2. `ozone genesis` should work (and download jmh first time)


hadoop-ozone/dist/src/shell/ozone/ozone
```diff
@@ -130,6 +130,18 @@ function ozonecmd_case
       OZONE_RUN_ARTIFACT_NAME="hadoop-ozone-tools"
     ;;
     genesis)
+      ARTIFACT_LIB_DIR="${HADOOP_HDFS_HOME}/share/ozone/lib/hadoop-ozone-tools"
+      mkdir -p "$ARTIFACT_LIB_DIR"
+      if [[ ! -f "$ARTIFACT_LIB_DIR/jmh-core-1.23.jar" ]]; then
+        echo "jmh-core jar is missing from $ARTIFACT_LIB_DIR, trying to download from maven central (License: GPL + classpath execption)"
+        wget -O "$ARTIFACT_LIB_DIR/jmh-core-1.23.jar"  https://repo1.maven.org/maven2/org/openjdk/jmh/jmh-core/1.23/jmh-core-1.23.jar
```
adoroszlai, line 137: How do we choose between `wget` and `curl`?

elek: > How do we choose between wget and curl?

`if random.getNextInt() % 2 == 0 --> wget` 

;-)

Will replace it with `curl` if that one is your preference. Is it more common?

adoroszlai: > Is it more common?

I don't have numbers for that.

https://daniel.haxx.se/docs/curl-vs-wget.html says:

> More portable. curl builds and runs on lots of more platforms than wget. For example: OS/400, TPF and other more "exotic" platforms that aren't straight-forward unix clones.

elek:
cc @jojochuang

elek:
> Should we also remove jmh from LICENSE.txt?

Good catch, thanks. Removed.

elek:
Thanks the review @arp7, @adoroszlai, @jojochuang 

I am merging it.

PR #828, opened 2020-04-15
HDDS-3002. NFS mountd support for Ozone

prashantpogde:

-Support for NFS v3 mountd support.
  These set of changes are enough to get the mountd started using
      '# ozone nfs3'
      and it should show up in
     '# showmount -e'

https://issues.apache.org/jira/browse/HDDS-3002

## How was this patch tested?

It is tested manually by running a local nfs3 mount server for ozone.


mukul1987:
@prashantpogde , there are multiple failure in the checks. can you please take a look into it.

prashantpogde:
yup @mukul1987, looking at failures and fixing them. I will upload another patch.

elek:
Just one unrelated question: do we need a separated branch for the NFS work? If I understood  well the only big change required on the OM side is adding the unique identifiers. Using the existing metadata capabilities it would be just a few lines and safe.

I am totally fine with using separated branch, but couldn't see any problem with avoiding the overhead of the separated feature branch and adding the new patches directly to the master.

prashantpogde:
@elek  @mukul1987  I will be ok either way with a separate branch or directly on the master. Mostly changes would be isolated in NFS directly with few hooks occasionally going in the ozone code base. But I guess now that we have a branch,We can continue with this branch.

mukul1987:
Hi @prashantpogde , I can see close to 144 commits. Can you please rebase the branch against master and then post the change.

Also there are still checkstyle and other issues which needs to be taken care of.

mukul1987:
Also lets add an acceptance test to check that Mountd server is coming up

prashantpogde:
I will write mountd acceptance test  in a separate pull request to avoid clubbing too many changes here.

prashantpogde:
I created another pull request "https://github.com/apache/hadoop-ozone/pull/917" which is much cleaner to look at the Mountd changes isolated from sync with master.

mukul1987:
Closing this pull request as we have another PR for this.

PR #829, opened 2020-04-15
HDDS-3384. Update SpringFramework to 5.1.14.

swagle:
Upgraded spring-jdbc due to CVE.

https://issues.apache.org/jira/browse/HDDS-3384

## How was this patch tested?
Ran the relevant test with success.

### DIFF
--- a/hadoop-ozone/pom.xml
+++ b/hadoop-ozone/pom.xml
@@ -26,6 +26,8 @@
 
   <properties>
     <docker.image>apache/ozone:${project.version}</docker.image>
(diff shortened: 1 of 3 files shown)


swagle:
@avijayanhwx / @vivekratnavel for review.

swagle:
> We also use the same dependency in hadoop-ozone/recon/pom.xml
> 
> Can we move spring.version to the root pom file and use it for both?

Ah, sorry missed that. we do not use spring anywhere outside of recon but yes, let me do that

PR #830, opened 2020-04-16
HDDS-3392.OM create key/file should not generate different data encryption key during validateAndUpdateCache.

bharatviswa504:

Generate EncryptionInfo only by leader OM in HA.

https://issues.apache.org/jira/browse/HDDS-3392

## How was this patch tested?

As no secure docker-compose cluster for HA, no tests added for this scenario.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
@@ -26,11 +26,9 @@ import java.util.stream.Collectors;
 
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
-import org.apache.hadoop.ozone.protocolPB.OMPBHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.apache.hadoop.fs.FileEncryptionInfo;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.scm.container.common.helpers.ExcludeList;
 import org.apache.hadoop.hdds.utils.db.cache.CacheKey;
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
@@ -143,6 +143,7 @@ public class OMKeyCreateRequest extends OMKeyRequest {
       newKeyArgs = keyArgs.toBuilder().setModificationTime(Time.now());
     }
 
+    generateRequiredEncryptionInfo(keyArgs, newKeyArgs, ozoneManager);
     newCreateKeyRequest =
         createKeyRequest.toBuilder().setKeyArgs(newKeyArgs)
             .setClientID(UniqueId.next());
@@ -213,7 +214,6 @@ public class OMKeyCreateRequest extends OMKeyRequest {
 
       OmBucketInfo bucketInfo = omMetadataManager.getBucketTable().get(
           omMetadataManager.getBucketKey(volumeName, bucketName));
-      encryptionInfo = getFileEncryptionInfo(ozoneManager, bucketInfo);
 
       omKeyInfo = prepareKeyInfo(omMetadataManager, keyArgs, dbKeyInfo,
           keyArgs.getDataSize(), locations, encryptionInfo.orNull(),
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
@@ -26,6 +26,7 @@ import java.util.stream.Collectors;
 
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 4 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
```diff
@@ -451,4 +446,62 @@ protected void checkKeyAclsInOpenKeyTable(OzoneManager ozoneManager,
...
+        OmBucketInfo bucketInfo = omMetadataManager.getBucketTable().get(
+            omMetadataManager.getBucketKey(volumeName, bucketName));
+
+
+        // Don't throw exception of bucket not found when bucketinfo is not
+        // null. If bucketinfo is null, later when request
+        // is submitted and if bucket does not really exist it will fail in
+        // applyTransaction step. Why we are doing this is if OM thinks it is
+        // the leader, but it is not, we don't want to fail request in this
+        // case. As anyway when it submits request to ratis it will fail with
```
xiaoyuyao, line 488: If we could fail the client earlier without the request adding to request log, that can avoid adding the request to ratis log as it will fail application any way.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
```diff
@@ -213,7 +214,6 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       OmBucketInfo bucketInfo = omMetadataManager.getBucketTable().get(
           omMetadataManager.getBucketKey(volumeName, bucketName));
-      encryptionInfo = getFileEncryptionInfo(ozoneManager, bucketInfo);
 
       omKeyInfo = prepareKeyInfo(omMetadataManager, keyArgs, dbKeyInfo,
           keyArgs.getDataSize(), locations, encryptionInfo.orNull(),
```
xiaoyuyao, line 219: encryptionInfo.orNull() should be removed as the keyArgs now have the data encryption key info.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
bharatviswa504: We are doing this because of our current isLeader semantics.
The same is explained in the comments, once after strong leader semantics, we can revisit this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
bharatviswa504: Done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
```diff
@@ -213,10 +213,9 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       OmBucketInfo bucketInfo = omMetadataManager.getBucketTable().get(
           omMetadataManager.getBucketKey(volumeName, bucketName));
-      encryptionInfo = getFileEncryptionInfo(ozoneManager, bucketInfo);
 
       omKeyInfo = prepareKeyInfo(omMetadataManager, keyArgs, dbKeyInfo,
-          keyArgs.getDataSize(), locations, encryptionInfo.orNull(),
+          keyArgs.getDataSize(), locations,  getFileEncryptionInfo(keyArgs),
```
xiaoyuyao, line 218: NIT: Given we have pass keyArgs, we can reduce the number of arguments here to  avoid @SuppressWarnings("parameternumber") for prepareKeyInfo().

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
xiaoyuyao: Sounds good to me. Do we have a JIRA to track the leader lease?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
bharatviswa504: By removing the encryptionInfo and size, we cannot remove @SuppressWarnings("parameternumber"). I will leave for now. I will open a new jira to refactor this.

bharatviswa504: Jira to track this
https://issues.apache.org/jira/browse/HDDS-3407

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
bharatviswa504: I will open a jira for this also.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
xiaoyuyao: Ok to me.  Let's fix that later. +1.

bharatviswa504:

PR #832, opened 2020-04-16
HDDS-3401. Ozone audit entries could be consistent among volume creation with quota and update quota

bharatviswa504:
Use QUOTA_IN_BYTES for setVolumeRequest.

## How was this patch tested?

Existing tests. This is just a naming change in the Audit log.


bharatviswa504:

PR #833, opened 2020-04-16
HDDS-3314. FIx ContainerOperationClient#readContainer to use Grpc Client to read from datanode

sadanand48:
Ozone debug chunkinfo <key uri> command fails intermittently on killing  a datanode in a multinode setup as it uses the ratis pipeline.This was because the query happens only on the leader node.The change is that the scm now queries different datanodes in case of failure through the grpcClient.

https://issues.apache.org/jira/browse/HDDS-3314

## How was this patch tested?
Manual test

### DIFF
--- a/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/ContainerOperationClient.java
+++ b/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/ContainerOperationClient.java
@@ -382,7 +382,7 @@ public class ContainerOperationClient implements ScmClient {
       Pipeline pipeline) throws IOException {
     XceiverClientSpi client = null;
     try {
-      client = xceiverClientManager.acquireClient(pipeline);
+      client = xceiverClientManager.acquireClientForReadData(pipeline);
       ReadContainerResponseProto response =
           ContainerProtocolCalls.readContainer(client, containerID, null);
       if (LOG.isDebugEnabled()) {


hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/ContainerOperationClient.java
```diff
@@ -382,7 +382,7 @@ public ContainerDataProto readContainer(long containerID,
       Pipeline pipeline) throws IOException {
     XceiverClientSpi client = null;
     try {
-      client = xceiverClientManager.acquireClient(pipeline);
+      client = xceiverClientManager.acquireClientForReadData(pipeline);
```
bharatviswa504, line 385: I have not really understood this change, as we just changed from acquireClient  -> acquireClientForReadData.
How that will handle failures, and how it will retry on other datanodes.

And also InfoSubCommand Uses below code, and it is not calling readContainer.
      final ContainerWithPipeline container = scmClient.
          getContainerWithPipeline(containerID);

>       final ContainerWithPipeline container = scmClient.
>           getContainerWithPipeline(containerID);
>       Preconditions.checkNotNull(container, "Container cannot be null");

bshashikant: @bharat, if we use acquireClientForReadData, it will use XceiverClientGrpc to read the info rather than Ratis(which reads from the leader). XceiverClientGrpc automatically fails over from datanode to another if it sees a failure.

InfoSubCommand has recently been changed to not use readContainer call to read data from datanode. But , in Ozone, we only use XceiverClientGrpc to read from datanode , it make sense here to use the same while reading the containerData as well (acquireClient -> acquireClientForReadData). This function is used by ChunkKeyHandler.

PR #834, opened 2020-04-16
HDDS-3101. Depend on lightweight ConfigurationSource interface instead of Hadoop Configuration

elek:

To make it possible to create different client jars compiled with different version of Hadoop we need clear and Hadoop independent `hdds-common` (and `hdds-client`) projects.

(For more details about the motivation, check [this](https://lists.apache.org/thread.html/rd0ea00f958368e888db1947eb71e514fb977df0b7baaad928ac50e94%40%3Cozone-dev.hadoop.apache.org%3E) design doc.)

Our current blocker is the usage of `org.apache.hadoop.conf.Configuration`. Configuration class is a heavyweight object from `hadoop-common` which introduce a lot of unnecessary dependencies. It also violates multiple OOP principles, for example the **Dependency inversion** principle (most of the time we don't need all the methods, it would be enough to depend on a reduced set of the methods)

To make our components more independent I propose to depend on a lightweight `ConfigurationSource` interface which includes all the required `getXXX` methods from the existing `Configuration`. OzoneConfiguration can implement that interface (and with older Hadoop we can create direct adapters).

Other motivation: Some of the used helper classes (related to the storage / time units) are not available in Hadoop 2.7. With this approach we can use any old Hadoop configuration as the only required method would be the `.get` method.

https://issues.apache.org/jira/browse/HDDS-3101

## How was this patch tested?

https://github.com/elek/hadoop-ozone/actions/runs/79732132


elek:
Please review this PR even if it's conflicted. It's a small change, but big patch: I need to rebase it daily.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -81,11 +81,11 @@
    * passed to LevelDB and this memory is allocated in Native code space.
    * CacheSize is specified
    * in MB.
-   * @param conf - {@link Configuration}
+   * @param conf - {@link org.apache.hadoop.hdds.conf.ConfigurationSource}
```
adoroszlai, line 84:
```suggestion
   * @param conf - {@link ConfigurationSource}
```

as it's already imported.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/freon/OzoneGetConf.java
```diff
@@ -234,7 +236,7 @@ public Integer run() throws Exception {
     public int doWorkInternal(OzoneGetConf tool, String[] args)
         throws IOException {
       Collection<InetSocketAddress> addresses = HddsUtils
-          .getSCMAddresses(tool.getConf());
+          .getSCMAddresses(new LegacyHadoopConfigurationSource(tool.getConf()));
```
adoroszlai, line 239: Wouldn't it be OK to use `OzoneConfiguration.of` instead of `Legacy...` in this class?  Especially since we explicitly create `OzoneGetConf` with an `OzoneConfiguration`.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/freon/OzoneGetConf.java
```diff
@@ -250,10 +252,13 @@ public int doWorkInternal(OzoneGetConf tool, String[] args)
     @Override
     public int doWorkInternal(OzoneGetConf tool, String[] args)
         throws IOException {
-      if (OmUtils.isServiceIdsDefined(tool.getConf())) {
-        tool.printOut(OmUtils.getOmHAAddressesById(tool.getConf()).toString());
+      LegacyHadoopConfigurationSource configSource =
+          new LegacyHadoopConfigurationSource(tool.getConf());
```
adoroszlai, line 256: Same here.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/contract/ITestOzoneContractCreate.java
```diff
@@ -18,18 +18,19 @@
...
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 
-import java.io.IOException;
-
 /**
  * Ozone contract tests creating files.
  */
-public class ITestOzoneContractCreate extends AbstractContractCreateTest {
+public class  ITestOzoneContractCreate extends AbstractContractCreateTest {
```
adoroszlai, line 33: Seems unintended.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/fs/SpaceUsageCheckFactory.java
```diff
@@ -68,14 +69,16 @@ default SpaceUsageCheckFactory setConfiguration(Configuration conf) {
...
+  static SpaceUsageCheckFactory create(ConfigurationSource config) {
     Conf conf = OzoneConfiguration.of(config).getObject(Conf.class);
     Class<? extends SpaceUsageCheckFactory> aClass = null;
     String className = conf.getClassName();
     if (className != null && !className.isEmpty()) {
       try {
-        aClass = config.getClassByName(className)
-            .asSubclass(SpaceUsageCheckFactory.class);
+        aClass =
+            (Class<? extends SpaceUsageCheckFactory>)
```
adoroszlai, line 79: Can we keep `.asSubclass(SpaceUsageCheckFactory.class)` instead of the unchecked cast?

elek:
> I would like to mention that a nontrivial part of the patch is cleanup of imports. While I'm not against that, it does increase the surface of merge conflicts unnecessarily.

Agree. I tried to follow an existing convention, but It turned out that there is no unified convention. Would be great to start a discussion to ozone-dev to use the same convention by anybody...

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueHandler.java
```diff
@@ -114,16 +113,20 @@
...
     containerType = ContainerType.KeyValueContainer;
     blockManager = new BlockManagerImpl(config);
     chunkManager = ChunkManagerFactory.createChunkManager(config);
-    volumeChoosingPolicy = ReflectionUtils.newInstance(conf.getClass(
-        HDDS_DATANODE_VOLUME_CHOOSING_POLICY, RoundRobinVolumeChoosingPolicy
-            .class, VolumeChoosingPolicy.class), conf);
+    try {
+      volumeChoosingPolicy = conf.getClass(
+          HDDS_DATANODE_VOLUME_CHOOSING_POLICY, RoundRobinVolumeChoosingPolicy
+              .class, VolumeChoosingPolicy.class).newInstance();
```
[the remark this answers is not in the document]
elek: That's true, I didn't restore the support of this feature (for the sake of simplicity).

Long term I would implement a `@Configurable` annotation instead of `Configurable` interface and inject configuration values instead of annotations if required.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/freon/OzoneGetConf.java
elek: > Especially since we explicitly create OzoneGetConf with an OzoneConfiguration

I think we create the `OzoneGetConf` with `Configuration` which includes an `OzoneConfiguration`:

```java
    Configuration conf = new Configuration();
    conf.addResource(new OzoneConfiguration());
    int res = ToolRunner.run(new OzoneGetConf(conf), args);
```

`conf` is a `Configuration` not an `OzoneConfiguration`. 

But I changed it to use `OzoneConfiguration.of`

(BTW, would be great to switch to use GenericCli instead of `ToolRunner`)

adoroszlai: > I think we create the `OzoneGetConf` with `Configuration` which includes an `OzoneConfiguration`:

That's old code.  New one requires `OzoneConfiguration` in constructors.

adoroszlai: > (BTW, would be great to switch to use GenericCli instead of `ToolRunner`)

Agree.  Seems like you already created a task for that: https://issues.apache.org/jira/browse/HDDS-3102

elek:
Thanks the review @adoroszlai 

I merged it to the master.

PR #836, opened 2020-04-16
HDDS-3403. Generate ozone specific version from type in FSProto.proto

elek:

FSProtos.proto is copied from the Hadoop and used during the proto file generation *BUT* the types defined in FSProtos.proto are generated by Hadoop subproject.

This makes it hard to use Ozone with older Hadoop versions as the types of FSProtos are available only from Hadoop 3.x.

An easy fix is to generate our own version based on the existing FSProtos.proto

https://issues.apache.org/jira/browse/HDDS-3403

## How was this patch tested?

CI tests.


elek:
Merging to the master. Thanks @adoroszlai the review.

PR #813, opened 2020-04-13
HDDS-3309. Add TimedOutTestsListener to surefire and add timeout to integration tests

smengcl:
## What changes were proposed in this pull request?

Add `TimedOutTestsListener` as a listener to maven-surefire-plugin like Hadoop [does](https://github.com/apache/hadoop/blob/1189af4746919774035f5d64ccb4d2ce21905aaa/hadoop-hdfs-project/hadoop-hdfs/pom.xml#L233-L238). And add timeout to integration tests **that are not marked with `@Ignore`** at the time of the patch.

Also previously discussed under https://github.com/apache/hadoop-ozone/pull/750.

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-3309

## How was this patch tested?

The patch should make the timed out tests print out the stack trace (if any). Checks itself.

### Bonus

To test if the `TimedOutTestsListener` is working, you can change the timeout of a class to an impossibly small value like 5 sec (for an integration test to initialize). Then run that integration test. Make sure the test isn't ignored or it won't run with `mvn test`.

Note that the timeout must be smaller than `surefire.fork.timeout` shown below otherwise the fork will still be killed with no stack trace or thread dump.

https://github.com/apache/hadoop-ozone/blob/2e9bae2f8636974c5c0e11a4a3e04d9cbfbb2288/pom.xml#L228

For example, in `TestOzoneManagerRestart` I removed the `@Ignore` annotation and set the timeout to `5000` ms. Then ran:
```bash
mvn test -e -Dtest=TestOzoneManagerRestart
```

Maven output:
```bash
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.hadoop.ozone.om.TestOzoneManagerRestart
[ERROR] Tests run: 3, Failures: 0, Errors: 3, Skipped: 0, Time elapsed: 15.223 s <<< FAILURE! - in org.apache.hadoop.ozone.om.TestOzoneManagerRestart
[ERROR] testRestartOMWithVolumeOperation(org.apache.hadoop.ozone.om.TestOzoneManagerRestart)  Time elapsed: 5.019 s  <<< ERROR!
java.lang.Exception: test timed out after 5000 milliseconds
	at java.lang.Thread.sleep(Native Method)
	at org.apache.hadoop.test.GenericTestUtils.waitFor(GenericTestUtils.java:218)
...
[...]

PR #837, opened 2020-04-16
HDDS-3400. Extract test utilities to separate module

adoroszlai:

Create `hadoop-hdds/test-utils` module for test utilities (eg. `GenericTestUtils` and `LambdaTestUtils`).  This should not depend on any other HDDS/Ozone modules: the goal is to be able to use test listeners defined in this module (currently only `TimedOutTestsListener`) in all other modules.

https://issues.apache.org/jira/browse/HDDS-3400

## How was this patch tested?

CI:
https://github.com/adoroszlai/hadoop-ozone/runs/592403416

Applied the addition of listener from #813, [verified](https://github.com/adoroszlai/hadoop-ozone/runs/592999817) that no [`ClassNotFoundException`](https://github.com/apache/hadoop-ozone/pull/813/checks?check_run_id=589621137) happens.

### DIFF
--- a/hadoop-hdds/client/pom.xml
+++ b/hadoop-hdds/client/pom.xml
@@ -32,6 +32,11 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
   <dependencies>
     <dependency>
       <groupId>org.apache.hadoop</groupId>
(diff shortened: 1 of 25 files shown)


adoroszlai:

PR #839, opened 2020-04-17
HDDS-3411. Switch Recon SQL DB to Derby.

avijayanhwx:

Recon currently uses Sqlite as its defacto SQL DB with an option to configure other JDBC compatible databases. However, on some platforms like the IBM power pc, this causes problems from compile time since it does not have the sqlite native driver. This task aims to change the default SQL DB used by Recon to Derby, but retains the out of the box support (no need to supply the driver) for Sqlite as well.

https://issues.apache.org/jira/browse/HDDS-3411

## How was this patch tested?
Added unit tests.
Manually tested on docker with Derby and Sqlite.
Built the recon modules on IBM PPC host.


hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/ReconSqlDbConfig.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.hadoop.hdds.conf.Config;
+import org.apache.hadoop.hdds.conf.ConfigGroup;
+import org.apache.hadoop.hdds.conf.ConfigTag;
+import org.apache.hadoop.hdds.conf.ConfigType;
+
+/**
+ * The configuration class for the Recon SQL DB.
+ */
+@ConfigGroup(prefix = "ozone.recon.sql.db")
+public class ReconSqlDbConfig {
```
swagle, line 30: Why does this class only provide 1 configuration item? The JooqPersistence module properties should not be moved here?

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/JooqCodeGenerator.java
```diff
@@ -55,10 +60,11 @@
...
       LoggerFactory.getLogger(JooqCodeGenerator.class);
 
-  private static final String SQLITE_DB =
-      System.getProperty("java.io.tmpdir") + "/recon-generated-schema";
-  private static final String JDBC_URL = "jdbc:sqlite:" + SQLITE_DB;
-
+  private static final String DB = Paths.get(
+      System.getProperty("java.io.tmpdir"),
+      "recon-generated-schema-" + Time.monotonicNow()).toString();
+  public static final String RECON_SCHEMA_NAME = "RECON";
```
swagle, line 66: Generally, schemas are lowercase.

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/JooqCodeGenerator.java
```diff
@@ -55,10 +60,11 @@
   private static final Logger LOG =
       LoggerFactory.getLogger(JooqCodeGenerator.class);
 
-  private static final String SQLITE_DB =
-      System.getProperty("java.io.tmpdir") + "/recon-generated-schema";
-  private static final String JDBC_URL = "jdbc:sqlite:" + SQLITE_DB;
-
+  private static final String DB = Paths.get(
+      System.getProperty("java.io.tmpdir"),
+      "recon-generated-schema-" + Time.monotonicNow()).toString();
```
swagle, line 65: Any reason to use a timestamp-based path?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/persistence/DefaultDataSourceProvider.java
```diff
@@ -43,14 +52,26 @@
    */
   @Override
   public DataSource get() {
```
swagle, line 54: This is actually an antipattern :-) Although I am guilty of this as well. Instead of the if were should have a DerbyDataSourceProvider, SqliteDataSourceProvider and Default one.

vivekratnavel:
@avijayanhwx Thanks for working on this! Looks like the integration test failure might be related to this change.

```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 23.674 s <<< FAILURE! - in org.apache.hadoop.ozone.recon.TestReconWithOzoneManager
[ERROR] testOmDBSyncing(org.apache.hadoop.ozone.recon.TestReconWithOzoneManager)  Time elapsed: 0.491 s  <<< FAILURE!
```

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/persistence/TestReconWithDifferentSqlDBs.java
```diff
@@ -0,0 +1,142 @@
+package org.apache.hadoop.ozone.recon.persistence;
```
vivekratnavel, line 1: License is missing.

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/JooqCodeGenerator.java
```diff
@@ -109,20 +113,25 @@ private void generateSourceCode(String outputDir) throws Exception {
...
-      db = new SQLiteDataSource();
-      db.setUrl(JDBC_URL);
+      try {
+        createNewDerbyDatabase(JDBC_URL, RECON_SCHEMA_NAME);
+      } catch (Exception e) {
+        LOG.error("Error creating Recon Derby DB.", e);
+      }
+      dataSource = new EmbeddedDataSource();
+      dataSource.setDatabaseName(DB);
+      dataSource.setUser(RECON_SCHEMA_NAME);
```
vivekratnavel, line 125: Why should we set schema name as user?

hadoop-ozone/recon/pom.xml
```diff
@@ -331,6 +331,11 @@
       <artifactId>bonecp</artifactId>
       <version>0.8.0.RELEASE</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.derby</groupId>
+      <artifactId>derby</artifactId>
+      <version>10.14.2.0</version>
+    </dependency>
     <dependency>
```
elek, line 339: Thanks the patch @avijayanhwx 

Can we remove old sqlite dependency from here and from `hadoop-ozone/dist/src/main/license/LICENSE.txt`?

Is it still required?

One other question: do you have any performance numbers between sqlite vs derby? (I don't think it's a big problem, just interested if you know sg.)

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/JooqCodeGenerator.java
avijayanhwx: In derby (at least), user name becomes the schema.

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/persistence/TestReconWithDifferentSqlDBs.java
avijayanhwx: Thanks, will fix it.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/persistence/DefaultDataSourceProvider.java
avijayanhwx: Yeah, I agree. Will look into this change. Maybe I can create 2 different datasource providers and re-use in the test.

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/JooqCodeGenerator.java
avijayanhwx: If we have a failed build on a host (which I experienced), re-building fails with database already found always.

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/JooqCodeGenerator.java
avijayanhwx: Yes, but internally, I found that it becomes uppercase in the DB metadata. Hence, if I keep it as "recon", I have to do .toUpperCase() in Jooq code gen setSchemaName for it to find our classes.

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/ReconSqlDbConfig.java
avijayanhwx: Yes, but the Java configuration based API is for new config items generally. The code looks at ozone-default.xml and also Java based configuration classes and creates a master list for the default xml during assembly.

hadoop-ozone/recon/pom.xml
avijayanhwx: @elek Thanks for the review.  We have not done perf testing on our SQL DB yet since we don't have a scale issue there (hopefully :) ) We will take up a task to try perf analysis between the 2 DBs. Until then, we can assume the performance between Derby vs Sqlite is similar as seen in some online references. 

Regarding the sqlite dependency that has been left around, I wanted to retain the out of the box support for sqlite in Recon (without JAR being supplied by the user). Also, we have sqlite JAR dependency in the unit tests.

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/ReconTaskSchemaDefinition.java
```diff
@@ -48,7 +50,9 @@
   @Override
   public void initializeSchema() throws SQLException {
     Connection conn = dataSource.getConnection();
-    createReconTaskStatus(conn);
+    if (!TABLE_EXISTS_CHECK.test(conn, RECON_TASK_STATUS_TABLE_NAME)) {
+      createReconTaskStatus(conn);
```
vivekratnavel, line 54: Can we rename to createReconTaskStatusTable for consistency?

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/UtilizationSchemaDefinition.java
```diff
@@ -52,8 +59,12 @@
   @Transactional
   public void initializeSchema() throws SQLException {
     Connection conn = dataSource.getConnection();
-    createClusterGrowthTable(conn);
-    createFileSizeCount(conn);
+    if (!TABLE_EXISTS_CHECK.test(conn, FILE_COUNT_BY_SIZE_TABLE_NAME)) {
+      createFileSizeCount(conn);
```
vivekratnavel, line 63: Can we rename to createFileSizeCountTable for consistency?

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/ReconSqlDbConfig.java
elek: I am 100% happy with the answer...

... but as @anuengineer is not here I feel the responsibility to decorate pull requests with random external literature. :smile: 

 Slow migration to Java based configuration API can happen at any time:

> THE BOY SCOUTS HAVE A RULE: “Always leave the campground cleaner than you found it.” If you find a mess on the ground, you clean it up regardless of who might have made it. You intentionally improve the environment for the next group of campers. (Actually, the original form of that rule, written by Robert Stephenson Smyth Baden-Powell, the father of scouting, was “Try and leave this world a little better than you found it.”)

> What if we followed a similar rule in our code: “Always check a module in cleaner than when you checked it out”? Regardless of who the original author was, what if we always made some effort, no matter how small, to improve the module? What would be the result?

https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html

(Again, It's fine with me as is, I do the same. I skip the migration when it makes the patch harder to understand. But I never skip to add funny comments... :wink: )

anuengineer: @elek and here comes the random quote, You can always tag me and I will show up; even if it only for a random quote :facepalm:  

“Without change, something sleeps inside us, and seldom awakens. The sleeper must awaken -- Frank Herbert, Dune "

Thank you for doing your part :), and wake up all.

avijayanhwx: Ok guys I am convinced :) I will move all the Recon SQL DB related configs to Java in this PR. These configs are very good candidates for hiding inside the code.  They are not expected to change often.

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/UtilizationSchemaDefinition.java
avijayanhwx: Done.

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/ReconTaskSchemaDefinition.java
avijayanhwx: Done.

vivekratnavel:
@avijayanhwx Thanks for working on this! 
@swagle @elek Thanks for the reviews!
I have merged the change.

PR #840, opened 2020-04-17
HDDS-3398. Update grpc/netty dependencies

mukul1987:

Upgate grpc/netty dependencies in ozone

https://issues.apache.org/jira/browse/HDDS-3398

## How was this patch tested?
Ran MiniOzoneChaosCluster test.

### DIFF
--- a/hadoop-ozone/csi/pom.xml
+++ b/hadoop-ozone/csi/pom.xml
@@ -29,7 +29,7 @@ http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <packaging>jar</packaging>
 
   <properties>
-    <grpc-compile.version>1.17.1</grpc-compile.version>
+    <grpc-compile.version>1.24.0</grpc-compile.version>
     <protobuf-compile.version>3.5.0</protobuf-compile.version>
   </properties>
   <dependencies>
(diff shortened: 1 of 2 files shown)


elek:
Thanks to create this patch @mukul1987 

Would be great to test CSI server. I found that the CSI server is broken since the last netty upgrade in ratis-third-party (0.2.0->0.3.0)

elek:
@arp7 was the CSI server tested?

elek:
I tested it and it didn't work. Would you be so kind to use 4.1.38 instead of 4.1.48? I couldn't see any CVE with 4.1.38 and it doesn't break CSI server.

arp7:
Sorry looks like I jumped the gun on the commit.

mukul1987:
@elek, How can the CSI server can be tested, is there an acceptance test available ?

elek:
> @elek, How can the CSI server can be tested, is there an acceptance test available ?

The easiest way is just start the "ozone csi" command. If it can be started (and a socket is created) than it should be fine. 

You need to `ozone-site.xml` configuration:  `ozone.csi.owner=hadoop` and `ozone.csi.socket=/tmp/csi.socket`

I created an acceptance test, will create a PR soon. Are you happy with using 4.1.38 instead of 4.1.48?

PR #842, opened 2020-04-19
HDDS-3455. Change MiniLoadGenerator to a pluggable model.

mukul1987:
This jira proposes to change MiniOzoneLoadGenerator to a pluggable model, this allows different tests to specify different load generators to be used in the test.

https://issues.apache.org/jira/browse/HDDS-3455

## How was this patch tested?
Ran MiniOzoneChaosCluster

### DIFF
--- a/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/MiniOzoneLoadGenerator.java
+++ b/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/MiniOzoneLoadGenerator.java
@@ -109,23 +109,23 @@ public class MiniOzoneLoadGenerator {
       return this;
     }
 
-    public Builder setConf(OzoneConfiguration conf) {
-      this.conf = conf;
+    public Builder setConf(OzoneConfiguration configuration) {
+      this.conf = configuration;
       return this;
     }
 
-    public Builder setNumBuffers(int numBuffers) {
-      this.numBuffers = numBuffers;
+    public Builder setNumBuffers(int buffers) {
+      this.numBuffers = buffers;
       return this;
     }
 
-    public Builder setNumThreads(int numThreads) {
-      this.numThreads = numThreads;
+    public Builder setNumThreads(int threads) {
+      this.numThreads = threads;
       return this;
     }
 
-    public Builder setVolume(OzoneVolume volume) {
-      this.volume = volume;
+    public Builder setVolume(OzoneVolume vol) {
+      this.volume = vol;
       return this;
     }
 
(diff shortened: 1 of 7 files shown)


hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/loadgenerators/LoadExecutors.java
```diff
@@ -70,15 +73,19 @@ private void load(long runTimeMillis) {
...
+        gen.initialize();
+      } catch (Throwable t) {
+        LOG.error("Failed to initialize loadgen:{}", gen, t);
       }
-    } catch (Throwable t) {
-      LOG.error("Failed to initialize loadgen:{}", generator, t);
+    }
+
+    for (int i = 0; i < numThreads; i++) {
+      futures.add(CompletableFuture.runAsync(() -> load(time), executor));
```
adoroszlai, line 88: If some generator fails to `initialize()`, should we still consider it in `load(long)`?  I think it would be more safe to add each generator to a separate list, from which `load(long)` should choose, only after successful initialization.

hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/MiniOzoneLoadGenerator.java
```diff
@@ -44,66 +39,99 @@
...
-    LoadBucket bucket = new LoadBucket(volume.getBucket(bucketName), conf,
-        omServiceID);
-    LoadGenerator loadGenerator = function.apply(bucket);
-    loadExecutors.add(new LoadExecutors(numThreads, loadGenerator));
+    LoadBucket ozoneBucket = new LoadBucket(volume.getBucket(bucketName),
+        conf, omServiceID);
+
+    LoadGenerator loadGenerator = clazz
+        .getConstructor(DataBuffer.class, LoadBucket.class)
+        .newInstance(buffer, ozoneBucket);
```
adoroszlai, line 75: Can you please document the required signature for the constructors of `LoadGenerator` implementations in the interface?

hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/loadgenerators/LoadExecutors.java
mukul1987: done, if the initialization fails. ane exception is thrown to stop the chaos cluster.

hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/MiniOzoneLoadGenerator.java
mukul1987: done, great comment :)

mukul1987:
Hi @adoroszlai, Yes I will make the code into a config based model in later patches. This will be taken care of in a followup patch

hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/loadgenerators/LoadGenerator.java
```diff
@@ -22,6 +22,16 @@
  * Interface for load generator.
  */
 public abstract class LoadGenerator {
+  /**
```
adoroszlai, line 25: This is not a javadoc for `keyNameDelimiter`.

```suggestion
  /*
```

PR #843, opened 2020-04-20
HDDS-3223. Improve s3g read 1GB object efficiency by 100 times

runzhiwang:
**What's the problem ?**

Use dd to read 1000M object from ozone bucket mounted by goofys, it cost about 470 seconds, i.e. 2.2M/s, which is too slow. 
image

I also use tcpdump to capture the packet when I  read 200M object. As the image shows, the 1st GET request cost about 1 second, but the 10th GET request cost about 22 seconds. The GET request become slower and slower.
image

**What's the reason ?**
When read 1000M object, there are 50 GET requests, each GET request read 20M. When do GET, the stack is: [IOUtils::copyLarge](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/endpoint/ObjectEndpoint.java#L262) -> [IOUtils::skipFully](https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/IOUtils.java#L1190) -> [IOUtils::skip](https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/IOUtils.java#L2064) -> [InputStream::read](https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/IOUtils.java#L1957).

It means, the 50th GET request which should read 980M-1000M, but to skip 0-980M, it also [InputStream::read](https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/IOUtils.java#L1957) 0-980M. So the 1st GET request read 0-20M, the 2nd GET request read 0-40M, the 3rd GET request read 0-60M, ..., the 50th GET request read 0-1000M. So the GET  request from 1st-50th become slower and slower.

You can also refer [IO-203](https://issues.apache.org/jira/browse/IO-203) and [IO-355](https://issues.apache.org/jira/browse/IO-355) why IOUtils implement skip by read rather than real skip, e.g. seek.
image

**How to improve ?**
Copy IOUtils::copyLarge to S3WrapperInputStream, and replace [IOUtils::skipFully](https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/IOUtils.java#L1190)  with [S3WrapperInputStream::seek](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/io/S3WrapperInputStream.java#L67).  No other changes of IOUtils::copyLarge.
After improving, read 1000M object cost 4.79 seconds, i.e. 219M/s, about 100 times faster.
image

https://issues.apache.org/jira/browse/HDDS-3223

## How was this patch tested?

Existed UT and IT.

### DIFF
--- a/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/io/S3WrapperInputStream.java
+++ b/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/io/S3WrapperInputStream.java
@@ -18,17 +18,20 @@
 
 package org.apache.hadoop.ozone.s3.io;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.hadoop.fs.FSInputStream;
 import org.apache.hadoop.ozone.client.io.KeyInputStream;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 
 /**
  * S3Wrapper Input Stream which encapsulates KeyInputStream from ozone.
  */
 public class S3WrapperInputStream extends FSInputStream {
   private final KeyInputStream inputStream;
+  private static final int DEFAULT_BUFFER_SIZE = 32 * 1024;
 
   /**
    * Constructs S3WrapperInputStream with KeyInputStream.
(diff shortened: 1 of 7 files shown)


runzhiwang:
I will fix the failed UT

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/io/S3WrapperInputStream.java
```diff
@@ -76,4 +79,87 @@ public long getPos() throws IOException {
...
+   * </p>
+   * <p>
+   * This method uses the provided buffer, so there is no need to use a
+   * <code>BufferedInputStream</code>.
+   * </p>
+   *
+   * @param output the <code>OutputStream</code> to write to
+   * @param inputOffset : number of bytes to skip from input before copying
+   * -ve values are ignored
+   * @param length : number of bytes to copy. -ve means all
```
bharatviswa504, line 129: -ve means all. But I don't see that handled in the code.

runzhiwang: @bharatviswa504 Hi, thanks for your comments. I copy the code and java doc from commons-io and only change one line. i.e. `skipFully(input, inputOffset);` -> seek(inputOffset).
 And -ve read entire file till end. If `length < 0`, then bytesToRead equals to bufferLength and never change. So while loop will not break until `read == IOUtils.EOF`. So it call read entire file till end.
image

runzhiwang:
> Thank You @runzhiwang for the analysis and great improvement in read performance.
> 
> I see there is one more usage of IOUtils.copyLarge in createMultipartKey, can we update over there also with a similar change.
> 
> L534:
> 
> ```
>             if (range != null) {
>               RangeHeader rangeHeader =
>                   RangeHeaderParserUtil.parseRangeHeader(range, 0);
>               IOUtils.copyLarge(sourceObject, ozoneOutputStream,
>                   rangeHeader.getStartOffset(),
>                   rangeHeader.getEndOffset() - rangeHeader.getStartOffset());
> 
>             }
> ```
> 
> And also can we add test for S3WrapperInputStream which tests a new method with multiple seek values.
@bharatviswa504 I will handle this. Thanks for your comments.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/io/S3WrapperInputStream.java
bharatviswa504: Got it.

runzhiwang:
@bharatviswa504 Hi, I move copyLarge into KeyInputStream, and add integration test in TestKeyInputStream for it. And also change all the use of IOUtils.copyLarge.

bharatviswa504:
+1 LGTM.
Thank you @runzhiwang for the improvement.

PR #845, opened 2020-04-20
HDDS-3451. Add timeout to XceiverServerRatis#submitRequest call.

nandakumar131:
Add timeout to XceiverServerRatis#submitRequest call

https://issues.apache.org/jira/browse/HDDS-3451


mukul1987:
Thanks for updating the patch @nandakumar131 . +1 the patch looks good to me.

PR #844, opened 2020-04-20
HDDS-3456. Fix Acceptance test failures due to disk out of space.

adoroszlai:
## What changes were proposed in this pull request?

1. Remove the docker images `ozone-build` and its temporary child image after completion of Maven build for acceptance test.
2. Remove `apache/hadoop:3.1.2` image after `hadoop31` test.  `hadoop:3` (for `hadoop32`) is used later for secure tests, too.

https://issues.apache.org/jira/browse/HDDS-3456

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/602389143

```
Dockerfile for action: '/home/runner/work/hadoop-ozone/hadoop-ozone/./.github/buildenv/Dockerfile'.
##[command]/usr/bin/docker build -t 430c1a:5eb4503f65695cb5f3b88870f564260e "/home/runner/work/hadoop-ozone/hadoop-ozone/.github/buildenv"
...
Run docker image rm $(docker images -a -q | head -1) || true
Untagged: 430c1a:5eb4503f65695cb5f3b88870f564260e
Deleted: sha256:35a79ac289f9198f022d949e34bf10a36de0230c4016ee2c05c0348f66dd0acc
...
Run docker image rm apache/ozone-build || true
Untagged: apache/ozone-build:latest
Untagged: apache/ozone-build@sha256:1bb0075ead1b4a78e53860a21b7613ffbab7f317e1fb304ebdeb7fbbfda11db4
Deleted: sha256:ce5cb3e6669c11ecfb36a079aab583c5f8c9eaaa23a91128c170a1cfadcc4968
...
```

PR #846, opened 2020-04-20
HDDS-3464. Use dedicated build partition for acceptance tests

elek:

Even with HDDS-3456 it seems to be a good idea to run acceptance test from /mnt in the github actions environment where we have dedicated 14GB space.

https://issues.apache.org/jira/browse/HDDS-3464

## How was this patch tested?

Tested on my local fork. Test started to run without any problem, but I didn't wait until the end with opening this PR to fix master as soon as possible.

Note: this PR tries to solve the same problem as #844, but there is no conflict between them. We can have both to have additional safety.


elek:
Seems to be possible to use `/mnt` for integration test, too:

https://github.com/elek/hadoop-ozone/runs/604768716?check_suite_focus=true

If it works I will modify the description of the JIRA/PR and propose to do it in one step.

adoroszlai:
Finally a green run.

elek:
> Finally a green run.

Double checked the cause of the previous fails to be sure that we didn't introduce any new flakyness:

 1. TestDnRatisLogParser (fixed in HDDS-3373, it could be missing from the build)
 2. TestSCMNodeMetrics (rewritten in HDDS-3480)
 3. topology acceptance test (I saw the same in other builds, so this is not a flakyness which introduced by this patch. Will start to run in on my fork to check the state...)

It seems to be proven that they are not new errors (no new flakyness) and the last build is green.

I am merging it (to have stable results)

Thanks the review @adoroszlai

PR #849, opened 2020-04-21
HDDS-3468. Organize log4j dependency in pom.xml.

bshashikant:

Organized log4j dependency

https://issues.apache.org/jira/browse/HDDS-3468

## How was this patch tested?

Build  ozone source code.


mukul1987:
Thanks for the patch @bshashikant , I see that there is also another variable called
    <log4j2.version>2.11.0</log4j2.version>

Can both of them be moved to top level pom. ?

PR #854, opened 2020-04-21
HDDS-3393. Recon throws NPE in clusterState endpoint

vivekratnavel:

- Perform a null check on tables before calling getEstimatedKeyCount() on them.
- Update mock api response to match the expected json structure.

https://issues.apache.org/jira/browse/HDDS-3393

## How was this patch tested?

Patch was tested with docker-compose and verified that there is no NPE thrown in Recon logs after hitting the clusterState endpoint before first snapshot from OM.

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/ClusterStateEndpoint.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/ClusterStateEndpoint.java
@@ -22,7 +22,6 @@ import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.NodeState;
 import org.apache.hadoop.hdds.scm.container.placement.metrics.SCMNodeStat;
 import org.apache.hadoop.hdds.scm.server.OzoneStorageContainerManager;
-import org.apache.hadoop.hdds.utils.db.Table;
 import org.apache.hadoop.ozone.recon.api.types.ClusterStateResponse;
 import org.apache.hadoop.ozone.recon.api.types.DatanodeStorageReport;
 import org.apache.hadoop.ozone.recon.recovery.ReconOMMetadataManager;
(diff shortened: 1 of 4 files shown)


hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/ClusterStateEndpoint.java
```diff
@@ -81,20 +82,26 @@ public Response getClusterState() {
             stats.getScmUsed().get(), stats.getRemaining().get());
     ClusterStateResponse.Builder builder = ClusterStateResponse.newBuilder();
     try {
-      builder.setVolumes(
-          omMetadataManager.getVolumeTable().getEstimatedKeyCount());
+      Table volumeTable = omMetadataManager.getVolumeTable();
```
bharatviswa504, line 85: Instead of not null checks, can we have a state in MetadataManager like INITIALIZED, STOPPED? This will help be future proof, and if anywhere this check is required it will be generic.

vivekratnavel: Thanks for the review! I added a boolean flag to keep track of the initialized state to ReconOMMetadataManager.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/recovery/ReconOMMetadataManager.java
```diff
@@ -41,4 +41,10 @@
    * Database.
    */
   long getLastSequenceNumberFromDB();
+
+  /**
+   * Check if OM tables are initialized.
+   * @return if OM Tables are initialized
```
bharatviswa504, line 47: Minor NIT: return true if OM Tables are initialized, otherwise false.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/recovery/ReconOmMetadataManagerImpl.java
```diff
@@ -120,4 +122,11 @@ public long getLastSequenceNumberFromDB() {
     }
   }
 
+  /**
+   * Check if OM tables are initialized.
+   * @return if OM Tables are initialized
+   */
+  public boolean isOmTablesInitialized() {
```
bharatviswa504, line 129: Missed `@override` notion, and same comment as above related to Javadoc or it can be completely removed.

hadoop-ozone/recon/src/main/resources/webapps/recon/ozone-recon-web/api/db.json
```diff
@@ -378,11 +378,11 @@
           }
         ],
         "missingSince": 1578491371528,
-        "pipelineId": "05e3d908-ff01-4ce6-ad75-f3ec79bcc7982"
+        "pipelineID": "05e3d908-ff01-4ce6-ad75-f3ec79bcc7982"
       },
       {
-        "id": 2,
-        "keys": 5943,
+        "containerID": 2,
```
bharatviswa504, line 384: Not understood why these are changed, and how it is related to this change.

vivekratnavel: @bharatviswa504 These are not related to the change. But, I noticed that the json structure of missing containers api mock response does not match with what is expected by the UI and thought of fixing them along with this change. Since, these are mock responses, they will not affect any behavior of Recon, except when starting a local mock development server to serve mock responses from this db.json file.

bharatviswa504:
Thank You @vivekratnavel for the contribution.

PR #855, opened 2020-04-21
HDDS-3474. Create transactionInfo Table in OmMetadataManager.

bharatviswa504:

1. introduce a new transaction info table which stores transactionInfo.
Key = TRANSACTIONINFO
value = currentTerm-transactionIndex
2. Add new UT's for this table.
3. Provide utility/helper methods to parse the transaction info table value

https://issues.apache.org/jira/browse/HDDS-3474

## How was this patch tested?

Added UT.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
@@ -326,4 +326,8 @@ public final class OzoneConsts {
   public static final String GDPR_ALGORITHM = "algorithm";
 
 
+  // Transaction Info
+  public static final String TRANSACTION_INFO_KEY = "#TRANSACTIONINFO";
+  public static final String TRANSACTION_INFO_SPLIT_KEY = "#";
+
 }
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
@@ -42,6 +42,7 @@ import org.apache.hadoop.hdds.utils.db.cache.CacheValue;
 import org.apache.hadoop.hdds.utils.db.cache.TableCacheImpl;
 import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.common.BlockGroup;
+import org.apache.hadoop.ozone.om.codec.OMTransactionInfoCodec;
 import org.apache.hadoop.ozone.om.codec.OmBucketInfoCodec;
 import org.apache.hadoop.ozone.om.codec.OmKeyInfoCodec;
 import org.apache.hadoop.ozone.om.codec.OmMultipartKeyInfoCodec;
@@ -64,6 +65,7 @@ import org.apache.hadoop.ozone.om.helpers.OzoneFSUtils;
 import org.apache.hadoop.ozone.om.helpers.RepeatedOmKeyInfo;
 import org.apache.hadoop.ozone.om.helpers.S3SecretValue;
 import org.apache.hadoop.ozone.om.lock.OzoneManagerLock;
+import org.apache.hadoop.ozone.om.ratis.OMTransactionInfo;
 import org.apache.hadoop.ozone.protocol.proto
     .OzoneManagerProtocolProtos.UserVolumeInfo;
 import org.apache.hadoop.ozone.security.OzoneTokenIdentifier;
@@ -118,6 +120,9 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
    * |----------------------------------------------------------------------|
    * |  multipartInfoTable| /volumeName/bucketName/keyName/uploadId ->...   |
    * |----------------------------------------------------------------------|
+   * |----------------------------------------------------------------------|
+   * |  ratislogTable | #TRANSACTIONINFO -> OMTransactionInfo               |
+   * |----------------------------------------------------------------------|
    */
 
   public static final String USER_TABLE = "userTable";
@@ -131,6 +136,8 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
   public static final String S3_SECRET_TABLE = "s3SecretTable";
   public static final String DELEGATION_TOKEN_TABLE = "dTokenTable";
   public static final String PREFIX_TABLE = "prefixTable";
+  public static final String TRANSACTION_INFO_TABLE =
+      "transactionInfoTable";
 
   private DBStore store;
 
@@ -148,6 +155,7 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
   private Table s3SecretTable;
   private Table dTokenTable;
   private Table prefixTable;
+  private Table transactionInfoTable;
   private boolean isRatisEnabled;
 
   public OmMetadataManagerImpl(OzoneConfiguration conf) throws IOException {
@@ -282,6 +290,7 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
         .addTable(DELEGATION_TOKEN_TABLE)
         .addTable(S3_SECRET_TABLE)
         .addTable(PREFIX_TABLE)
+        .addTable(TRANSACTION_INFO_TABLE)
         .addCodec(OzoneTokenIdentifier.class, new TokenIdentifierCodec())
         .addCodec(OmKeyInfo.class, new OmKeyInfoCodec())
         .addCodec(RepeatedOmKeyInfo.class, new RepeatedOmKeyInfoCodec())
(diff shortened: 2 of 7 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
```diff
@@ -329,4 +329,8 @@ private OzoneConsts() {
   public static final String GDPR_ALGORITHM = "algorithm";
 
 
+  // Transaction Info
+  public static final String TRANSACTION_INFO_KEY = "TRANSACTIONINFO";
+  public static final String TRANSACTION_INFO_SPLIT_KEY = "-";
+
```
hanishakoneru, line 335: Question out of curiosity: In another patch (HDDS-3217), the Keys in DB are suffixed with a #. Is that a just convention followed for ContainerDBs or has any significance?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OMTransactionInfo.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.common.base.Preconditions;
+import org.apache.hadoop.ozone.OzoneConsts;
+
+/**
+ * TransactionInfo which is applied to OM DB.
+ */
+public class OMTransactionInfo {
+
+  private long currentTerm;
+  private long transactionIndex;
```
hanishakoneru, line 30: Can we add a comment here that the transactionIndex corresponds to the Ratis Log Index. Would help if someone is trying to correlate this to Ratis snapshots.

elek:
Thanks @bharatviswa504 to work on this. I am trying to understand the problem to help with the code review but still not clear the proposal. I added my comments to the parent JIRA.

Especially as this patch introduced `String`->`String` table.

It's also not clear the performance degradation caused by this problem. Can you please help me to understand?

bharatviswa504:
> Thanks @bharatviswa504 to work on this. I am trying to understand the problem to help with the code review but still not clear the proposal. I added my comments to the parent JIRA.
> 
> Especially as this patch introduced `String`->`String` table.
> 
> It's also not clear the performance degradation caused by this problem. Can you please help me to understand?

For the questions posted in jira, replied over there.

The main intention of this Jira is
1. Avoid additional DB looks up, which impact perf. One such jira. HDDS-3129
 2. To not handle replay code in actual request execution which was done in previous jiras. This has caused additional complexity in the code.
3. Optimize restart time.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
bharatviswa504: Technically we don't need it. As in container DB we have iterators to iterate the keys in container, but there are some keys like BCSID, blockCount which are not blocks, so to avoid this they have a filter to skip these keys during iteration (But the same can be done without # also. Might be there it is done to know that these are system defined keys and a way to distinguish. (This is just my understanding, and followed the same when implementing HDDS-3217). In OM, we have tables for each. But I can use the # prefix.

bharatviswa504: Done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OMTransactionInfo.java
bharatviswa504: Done.

bharatviswa504:
Thank You @hanishakoneru for the review. Addressed review comments.

arp7:
There is also a design note attached to https://issues.apache.org/jira/browse/HDDS-3354.

elek:
Thanks the answer @bharatviswa504 . Still working on my understanding but one small comments in the mean time:

If the table is TransactionInfo -> currentTerm-transactionIndex as you write in the Jira: can we modify `Table<String, String>` to have real types + Codec. Can help the understanding IMHO.

bharatviswa504:
> Thanks the answer @bharatviswa504 . Still working on my understanding but one small comments in the mean time:
> 
> If the table is TransactionInfo -> currentTerm-transactionIndex as you write in the Jira: can we modify `Table<String, String>` to have real types + Codec. Can help the understanding IMHO.

Key - TRANSACTIONINFO (string)
Value - currentTerm#transactionIndex (Also a string) not sure what is meant by real types here?

bharatviswa504:
> > Thanks the answer @bharatviswa504 . Still working on my understanding but one small comments in the mean time:
> > If the table is TransactionInfo -> currentTerm-transactionIndex as you write in the Jira: can we modify `Table<String, String>` to have real types + Codec. Can help the understanding IMHO.
> 
> Key - TRANSACTIONINFO (string)
> Value - currentTerm#transactionIndex (Also a string) not sure what is meant by real types here?
> 
> As both are strings, so used a String codec which already exists.

The reason for not creating a proto object is for every iteration I need to create a proto object and then it will be converted to bytes when flushed tooDB. I think this is proto creation is unnecessary and the same can be done with simple string i think this should be fine. If needed for understandability I can update java doc and by reading the OMTransactionInfo class will help. Let me know your thoughts.

bharatviswa504:
Hi @elek 
Any more comments?

I want to get this in, as this is blocking further jira's progress.

bharatviswa504:
Thank You @elek for the review.
I have addressed the review comments.

`Do we need a generic settings/ table instead of introducing a new table every time when we need to store something on disk? In that case it would be more reasonable to use <String,String>.`

I have created ratis log table, because in OM each information/entity is stored in a column family. And also we need to persist this information in DB because we want the operation to be atomic, so that when restart, we find lastAppliedIndex from DB, and start applying the transactions after this transaction so that we know this is the last transaction applied to DB and we don't need any special handling of replay in actual requests. (Right now in actual request logic, we have a code for replay, and it is difficult to maintain and the new requests implemented by developers need to know about this detail, with this the user can implement the actual logic for the request and donot need to worry about replay part)
https://github.com/facebook/rocksdb/issues/2426

elek:
I did a quick test with 

```
#!/usr/bin/env bash
for i in `seq 1 1000000`; do
    /home/elek/projects/ldb --db=/tmp/new.db --create_if_missing put TEST $RANDOM
done
```

(ldb is a rocksdb LDB tool).

The disk usage oscillated between 100MB and 200MB. I think all the new values are appended and during the compaction they are removed. Which means that we have a lot of unnecessary IO. 

I still don't fully understand the design. Most probably because my limited understanding it seems to be very complex for me and just wandering if there is any alternatives which were considered and more simple. Would be great to discuss sometimes...

arp7:
> The disk usage oscillated between 100MB and 200MB. I think all the new values are appended and during the compaction they are removed. Which means that we have a lot of unnecessary IO.

Do you happen to have comparable analysis without this change?

bharatviswa504:
> I did a quick test with
> 
> ```
> #!/usr/bin/env bash
> for i in `seq 1 1000000`; do
>     /home/elek/projects/ldb --db=/tmp/new.db --create_if_missing put TEST $RANDOM
> done
> ```
> 
> (ldb is a rocksdb LDB tool).
> 
> The disk usage oscillated between 100MB and 200MB. I think all the new values are appended and during the compaction they are removed. Which means that we have a lot of unnecessary IO.
> 
> I still don't fully understand the design. Most probably because my limited understanding it seems to be very complex for me and just wandering if there is any alternatives which were considered and more simple. Would be great to discuss sometimes...

We are not doing write with sync option on. And few questions is this with this change we see more disk usage compared to with out this change?

bharatviswa504:
Posted a question regarding this pattern, when we started a proposal for this.
https://github.com/facebook/rocksdb/issues/6595

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
```diff
@@ -131,6 +136,8 @@
   public static final String S3_SECRET_TABLE = "s3SecretTable";
   public static final String DELEGATION_TOKEN_TABLE = "dTokenTable";
   public static final String PREFIX_TABLE = "prefixTable";
+  public static final String TRANSACTION_INFO_TABLE =
+      "transactionInfoTable";
```
hanishakoneru, line 140: Can we rename RatisLogTable to TransactionInfoTable to avoid confusion.

bharatviswa504: Done.

bharatviswa504:
Thank You @hanishakoneru for the review and @elek for the discussion and review.

PR #856, opened 2020-04-21
HDDS-3475. Use transactionInfo table to persist transaction information.

bharatviswa504:

Commit the transaction info as part of double buffer flush.

This requires #855. Posted this for CI run.

https://issues.apache.org/jira/browse/HDDS-3475

## How was this patch tested?

Added UT.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
@@ -140,7 +140,7 @@ public final class OzoneManagerDoubleBuffer {
 
     public OzoneManagerDoubleBuffer build() {
       if (isRatisEnabled) {
-        Preconditions.checkNotNull(indexToTerm, "When ratis is enabled, " +
+        Preconditions.checkNotNull(rs, "When ratis is enabled, " +
                 "OzoneManagerRatisSnapshot should not be null");
         Preconditions.checkNotNull(indexToTerm, "When ratis is enabled " +
             "indexToTerm should not be null");
@@ -307,7 +307,7 @@ public final class OzoneManagerDoubleBuffer {
           }
 
           // When non-HA do the sort step here, as the sorted list is not
-          // require for flush to DB. As in non-HA We want to complete
+          // required for flush to DB. As in non-HA we want to complete
           // futures as quick as possible after flush to DB, to release rpc
           // handler threads.
           if (!isRatisEnabled) {
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
@@ -26,9 +26,11 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import org.apache.hadoop.hdds.function.SupplierWithIOException;
 import org.apache.hadoop.hdds.tracing.TracingUtil;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMResponse;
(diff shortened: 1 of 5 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
```diff
@@ -86,10 +89,10 @@
...
-      OzoneManagerRatisSnapshot ozoneManagerRatisSnapShot) {
-    this(omMetadataManager, ozoneManagerRatisSnapShot, true);
-  }
+  /**
+   * function which will get term associated with the transaction index.
+   */
+  private Function<Long, Long> termGet = null;
 
   public OzoneManagerDoubleBuffer(OMMetadataManager omMetadataManager,
       OzoneManagerRatisSnapshot ozoneManagerRatisSnapShot,
```
hanishakoneru, line 98: For this constructor, termGet will be null. But a daemon is initialized to flushTransactions. Could it lead to NullPointer?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
```diff
@@ -173,17 +196,19 @@ private void flushTransactions() {
...
-              readyBuffer.stream().map(DoubleBufferEntry::getTrxLogIndex)
-                  .max(Long::compareTo).get();
-
-          List<Long> flushedEpochs =
-              readyBuffer.stream().map(DoubleBufferEntry::getTrxLogIndex)
-                  .sorted().collect(Collectors.toList());
+          // When non-HA do this step here, as the sorted is not require in
+          // flush to DB. As in non-HA We want to complete futures as quick
+          // as possible after flush to DB, to release rpc handler threads.
+          if (!isRatisEnabled) {
```
hanishakoneru, line 202: Sorry this comment is not clear. Do you mean sorted Transaction list is not required here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
bharatviswa504: This constructor is removed, as now we use Builder. And also for non-HA this termGet is not used, so we don't need this to be set. So, skipped setting this in OzoneManagerProtocolServerSideTranslatorPB. (Previously this constructor used to be called from this class)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
bharatviswa504: Yes, at this point in time we don't require the list to be sorted for non-HA, the sorted list is required for cleanupCache only, so delaying this. The reason for this already explained in the comment.

bharatviswa504:
Thank You @hanishakoneru for the review. Code has been changed as HDDS-3474 completely changed the logic, so modified the PR according to HDDS-3474 changes.

bharatviswa504:
@hanishakoneru thanks for the review, addressed review comments and also rebased on top of apache/master as now HDDS-3474 went in.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
```diff
@@ -121,15 +133,27 @@ public Builder enableTracing(boolean enableTracing) {
...
 
+    public Builder setIndexToTerm(Function<Long, Long> termGet) {
+      this.indexToTerm = termGet;
+      return this;
+    }
+
     public OzoneManagerDoubleBuffer build() {
+      if (isRatisEnabled) {
+        Preconditions.checkNotNull(indexToTerm, "When ratis is enabled, " +
+                "OzoneManagerRatisSnapshot should not be null");
```
hanishakoneru, line 144: Checking indexToTerm not null instead of omRatisSnapshot

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
hanishakoneru: Time for this sorting should be inconsequential. The readyBuffer would have 10s of transactions at max. 
Do we really need to separate it for HA and non-HA? 
NIT: Can you please fix the typo require -> required and capital W in We.

bharatviswa504: Some time, we see this list grow up to 100-300. And for non-HA, all rpc handlers are blocked on flush futures, intention here is does not want to spend time on unnecessary operations and make the flush futures complete and unblock rpc handlers.

bharatviswa504:
Addressed/replied to the comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
bharatviswa504: Done

hanishakoneru:
+1 pending CI.

bharatviswa504:
Thank You @hanishakoneru for the review

PR #857, opened 2020-04-22
HDDS-3385. Simplify S3 -> Ozone volume mapping

adoroszlai:

Replace user-specific volume mapping with fix `/s3v` volume for S3 buckets.  Unfortunately `s3` is not a valid volume name, at least 3-chars length is currently required.

* Remove server-side bucket mapping
* Remove `ozone s3 path` command
* Verify volume name and bucket name separately, to be able to throw `OMException` with the correct result code instead of `IllegalArgumentException`

Originally I intended to handle on-demand `/s3v` volume creation in Ozone client, as it's used by both S3 Gateway and CSI service.  But volumes can only be created by administrators, while buckets can be created by others, too.  Previously this was circumvented at the Ozone Manager server side in the S3-specific bucket creation handler.  So we can either ask Ozone users to create the volume if they want to use S3 Gateway, or we have to keep the S3-specific OM API.  For this patch I chose the first approach.

Also:

* add smoke test for deleting buckets
* check location in bucket create response
* fix a few typos / leftover items in the doc

https://issues.apache.org/jira/browse/HDDS-3385

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/608414085


adoroszlai:

bharatviswa504:
Sorry for the late review. Since it is marked for draft until 2 days back, not looked into it.

1. Previously listBuckets owned by an authenticated user. Now we list all buckets in /s3v will be shown (Meaning buckets which are created by other user also will be shown now). This is a behavior change from previous release, is the semantics of this API have changed with this.
2. And also this is a breaking change from the previous release, where old buckets will not be read by new S3G. (Do we want to support older release 0.5 beta or break the compatibility)

cc @arp7

elek:
> And also this is a breaking change from the previous release, where old buckets will not be read by new S3G. (Do we want to support older release 0.5 beta or break the compatibility)

There was no promise about backward compatibility. We try to do our best to make storage scheme compatible to support early adopters.

In this specific case, I agree, this is an incompatible change (especially until 'bind mount' of volumes are not implemented).

That's why a warning was sent to the mailing list:

https://lists.apache.org/thread.html/r647c7564e6f60e065d0f45cf4bc34cbb9a14a94bc4389daf9577d27a%40%3Cozone-dev.hadoop.apache.org%3E

> !!! INCOMPATIBLE CHANGE ALERT !!!!!
> 
> When 1 will be implemented, but 2, not yet. For a limited time of period 
> we will share buckets only from one volume as s3 buckets. This is 
> different from the current implementation when you can use s3buckets 
> from multiple volumes.
> 
> If it's a blocker for you, please share your opinion and we can schedule 
> the implementation according to the feedback.

bharatviswa504:
> > And also this is a breaking change from the previous release, where old buckets will not be read by new S3G. (Do we want to support older release 0.5 beta or break the compatibility)
> 
> There was no promise about backward compatibility. We try to do our best to make storage scheme compatible to support early adopters.
> 
> In this specific case, I agree, this is an incompatible change (especially until 'bind mount' of volumes are not implemented).
> 
> That's why a warning was sent to the mailing list:
> 
> https://lists.apache.org/thread.html/r647c7564e6f60e065d0f45cf4bc34cbb9a14a94bc4389daf9577d27a%40%3Cozone-dev.hadoop.apache.org%3E
> 
> > !!! INCOMPATIBLE CHANGE ALERT !!!!!
> > When 1 will be implemented, but 2, not yet. For a limited time of period
> > we will share buckets only from one volume as s3 buckets. This is
> > different from the current implementation when you can use s3buckets
> > from multiple volumes.
> > If it's a blocker for you, please share your opinion and we can schedule
> > the implementation according to the feedback.

I have missed reading the email notification. 

Let me answer why this is brought up during implementation of HDDS-3217 it has come up we need to support older version like 0.5.0 beta release to newer version, as some of customers like Tencent who are using ozone might be affected. So, only handling upgrade compatibility in one jira will not be enough, we should not have any breaking/compatibility changes from 0.5.0 version, so during the code reading, I found this and brought it up. 

So, now is there any plan to support the buckets which are created using AWS Accesskey id for volumes, until the mount is supported, or is there any plan to complete the mount before the next release?

ChenSammi:
Sorry, didn't see this email thread earlies.
Besides the imcompatible s3 volume name change,  is there any ACL changes
after one volume one s3 bucket changed to one volume, multiple user's
bucket?


Bests,
Sammi

elek:
> So, now is there any plan to support the buckets which are created using AWS Accesskey id for volumes, until the mount is supported, or is there any plan to complete the mount before the next release?

Yes, I think it can be completed before the next release.

>   is there any ACL changes
after one volume one s3 bucket changed to one volume, multiple user's
bucket?

Only the mapping is changed, I assume all the existing bucket level permission is the same.

PR #861, opened 2020-04-23
HDDS-3483. Handle unhealthy replica state of an open container.

nandakumar131:
HDDS-3483. Handle unhealthy replica state of an open container.

It was assumed that when a container is in OPEN state SCM doesn't have to take any action on it and the Ratis pipeline in datanode should handle the consistency of that container.

But, there are cases where a replica of a container which is in OPEN state can be marked as UNHEALTHY in datanode. This should be handled in SCM.

## How was this patch tested?

Added unit test to verify the new functionality.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -252,9 +252,14 @@ public class ReplicationManager implements MetricsSource, SafeModeNotification {
       final LifeCycleState state = container.getState();
 
       /*
-       * We don't take any action if the container is in OPEN state.
+       * We don't take any action if the container is in OPEN state and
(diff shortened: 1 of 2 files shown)


elek:
Thanks to open this issue @nandakumar131, one minor comment:

{code}
+    // Wait for EventQueue to call the event handler
+    Thread.sleep(100L);
+    Mockito.verify(closeContainerHandler, Mockito.times(1))
+        .onMessage(id, eventQueue);
{code}

FYI: Instead of using `Thread.sleep` we have a better api: `eventQueue.processAll(5000);`

(Sorry, I was not fast enough to comment it during the 1-2 hours to window while it was open. Keeping it open more time can give everybody a chance to check it (especially as US is waking up just now))

nandakumar131:
Thanks for the review @elek.
Good to know that we have the API exposed in `EventQueue`. I'm also not a fan of `Thread#sleep` in test cases.

We should even refactor other tests in the class to use `EventQueue#processAll`. Created HDDS-3484 for the same.

PR #863, opened 2020-04-23
HDDS-3461. Add acceptance test to smoketest CSI service startup

elek:

Ozone CSI service implement Container Storage Interface specification to provide volumes for container orchestrators such as Yarn and Kubernetes.

We don't have any acceptance test which makes easy to break the classpath / functionality.

As a first step I would create a simple smoketest to check if the csi service can be started and the identitiy service can be called.

https://issues.apache.org/jira/browse/HDDS-3461

## How was this patch tested?

With executing `./test.sh` in the `compose/ozone-csi` folder. (Will be part of the acceptance test runs).


elek:
>  I'd prefer adding csi container to the base ozone environment instead of duplication, but we can discuss that later.

Yeah, I don't know what is the best approach. It seems to be an optional component. Earlier we had a conversation about creating `ozone-minimal` and `ozone-full` containers, but since that, we already added **recon** and *s3g** to the normal cluster as required components.

I think **recon** should be required, s3g is more like a very frequently used optional component. csi seems to be more optional, that's the reason why I created a separated cluster, but I am open to merge it with anything else. (In theory I agree to provide less cluster) 

I will merge it as is to have this safety line as soon as possible, but open to move in any direction.

Thanks the review @adoroszlai

elek:
RevertedFauked

elek:
/pending Failed in https://github.com/apache/hadoop-ozone/actions/runs/86963299

elek:
@adoroszlai Problem is fixed. Do you have any concern to recommit it?

adoroszlai:
> @adoroszlai Problem is fixed. Do you have any concern to recommit it?

So the fix is 0a8b3d68ff3db7a03b634ea68d03db3cc4e1c8fe?  LGTM.

elek:
> So the fix is 0a8b3d6? LGTM.

Yes. CSI socket was missing when we checked it.

Thanks to review it @adoroszlai and @dineshchitlangia Will merge it again.

PR #864, opened 2020-04-23
HDDS-3405.Tool for Listing keys from the OpenKeyTable

sadanand48:
The tool lists keys present in the Open Key Table and can be used to debug when keys are written into ozone. There is a chance that the key might not have gotten committed into the OM (KeyTable) and will be present in the OpenKeyTable which will be displayed by the tool.

https://issues.apache.org/jira/browse/HDDS-3405

## How was this patch tested?
written unit test in TestOmMetadataManager


elek:
Thanks to open this issue @sadanand48, I have some question regarding the concept itself.

Based on the description it seems to be debug option. I don't think that we need to extend the client RPC just for a debug tool.

There are two options, what I would prefer:

 * Use a generic tool which reads the local rocksdb. We had this one earlier (`ozone sql`) but it doesn't work. I would suggest to fix it and provide a generic way to select any table. I think it would add the power for the admins without adding more methods to the RPC.

 * An other option if we **really** need to get this info from remote host, to separate the admin and client calls and add this method to the admin calls. It was discussed earlier that we need a better separation of RPC calls:

 1. Client calls are available from any machine and public
 2. Admin calls can be published on different port, used by the admin tools and can be blocked for security reason (only for localhost)
 3. Server calls are intra-service calls

PR #867, opened 2020-04-24
HDDS-3373. Intermittent failure in TestOMRatisLogParser

adoroszlai:

Fix intermittent failure in `TestOMRatisLogParser`.  The problem was that regardless of which `i` index corresponded to non-`snapshot` `ratisDirs[i]`, index `1` was always used for `groupDir`.

Added an assertion to verify that we do not choose `snapshot`.  Also added a few other safety checks.

https://issues.apache.org/jira/browse/HDDS-3373

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/614616449


adoroszlai:
Thanks @elek for the review.  I'm merging it.

PR #868, opened 2020-04-24
HDDS-3457. Fix ozonefs put and mkdir KEY_NOT_FOUND issue when ACL enable 

captainzmc:

When ACL enabled(not enable security), while we put file of  mkdir using O3FS, FileSystem will check if it exists first. If the key not exist, checkAccess will report KEY_NOT_FOUND Exception when get keyInfo. 

Because the previous bucket permission check has passed, it’s currently reading a key that does not exist, so this should be allowed.

So when KEY_NOT_FOUND appears, we confirm whether it is a read ACLType, and if it’s a read ACLType we return true. To avoid KEY_NOT_FOUND exception.

https://issues.apache.org/jira/browse/HDDS-3457

## How was this patch tested?

Enable acl and put file or mkdir using o3fs.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -1614,8 +1614,15 @@ public class KeyManagerImpl implements KeyManager {
           OzoneFileStatus fileStatus = getFileStatus(args);
           keyInfo = fileStatus.getKeyInfo();
         } catch (IOException e) {
-          throw new OMException("Key not found, checkAccess failed. Key:" +
-              objectKey, KEY_NOT_FOUND);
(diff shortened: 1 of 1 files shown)


aryangupta1998:
Hi @captainzmc,
I enabled ACL and performed mkdir using o3fs, but 'checkAccess' function is never reached. I have put some System.out.println statements in 'checkAccess' function, but nothing is being printed. Can you please help!

captainzmc:
> Hi @captainzmc,
> I enabled ACL and performed mkdir using o3fs, but 'checkAccess' function is never reached. I have put some System.out.println statements in 'checkAccess' function, but nothing is being printed. Can you please help!

Hi @aryangupta1998 [Is that the method](https://github.com/apache/hadoop-ozone/blob/34ee8311b0d0a37878fe1fd2e5d8c1b91aa8cc8f/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L1633)? I suggest you print the log with LOG.xxx or enable debugging in IDEA.

PR #871, opened 2020-04-25
HDDS-3487. Ozone start fails with NullPointerException in TLS enabled cluster

avijayanhwx:
Fix the NPE in ozone components startup.
```
SCM start failed with exception
java.lang.NullPointerException
  at org.apache.hadoop.hdds.conf.ConfigurationSource.getPassword(ConfigurationSource.java:112)
  at org.apache.hadoop.hdds.server.http.BaseHttpServer.getPassword(BaseHttpServer.java:348)
```

**Problem** 
The LegacyHadoopConfiguration class needs to use the underlying getPassword API which reads the password from the CredentialManager, and then falls back to the config key. Instead it used the default implementation that just looked for the config key's corresponding value.

https://issues.apache.org/jira/browse/HDDS-3487

## How was this patch tested?
Manually tested.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/LegacyHadoopConfigurationSource.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/LegacyHadoopConfigurationSource.java
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.hdds.utils;
 
+import java.io.IOException;
 import java.util.Collection;
 
 import org.apache.hadoop.conf.Configuration;
@@ -37,6 +38,11 @@ public class LegacyHadoopConfigurationSource implements ConfigurationSource {
   @Override
   public String get(String key) {
     return configuration.getRaw(key);
+  }
+
+  @Override
+  public char[] getPassword(String key) throws IOException {
+    return configuration.getPassword(key);
   }
 
   @Override
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/conf/OzoneConfiguration.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/conf/OzoneConfiguration.java
@@ -262,5 +262,16 @@ public class OzoneConfiguration extends Configuration
     }
     return configMap;
   }
-  
+
+  @Override
+  public char[] getPassword(String key) {
+    String value = get(key);
+    if (value == null) {
+      throw new NullPointerException(
+          "Password entry is missing for key " + key
+              + ".Note: generic ConfigurationSource interface doesn't support"
+              + " Hadoop CredentialProvider implementations");
+    }
+    return value.toCharArray();
+  }
 }
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/conf/OzoneConfiguration.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/conf/OzoneConfiguration.java
@@ -267,7 +267,7 @@ public class OzoneConfiguration extends Configuration
   public char[] getPassword(String key) {
     String value = get(key);
     if (value == null) {
-      throw new NullPointerException(
+      throw new UnsupportedOperationException(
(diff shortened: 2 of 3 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/LegacyHadoopConfigurationSource.java
```diff
@@ -39,6 +40,11 @@ public String get(String key) {
     return configuration.getRaw(key);
   }
 
+  @Override
+  public char[] getPassword(String key) throws IOException {
```
mukul1987, line 44: Should we also remove the deault method on ConfigurationSource for getPassword, and have the get() method in OzoneConfiguration and the new method in LegacyHadoopConfig ?

avijayanhwx:
> I tested it with a TLS enabled cluster and it worked well:
> 
> ssl-conf:
> 
> ```
> <configuration>
> <property><name>ssl.server.keystore.location</name><value>/etc/keystore/keystore</value></property>
> <property><name>ssl.server.exclude.cipher.list</name><value>^.*MD5.*$,^TLS_DH_.*$,^.*RC4.*$,^.*CCM.*$,^TLS_DHE.*$,^.*SHA$,^TLS_RSA_WITH.*$</value></property>
> <property><name>ssl.server.keystore.password</name><value>Welcome1</value></property>
> <property><name>ssl.server.keystore.keypassword</name><value>Welcome1</value></property>
> <property><name>ssl.server.truststore.location</name><value>/etc/keystore/keystore</value></property>
> <property><name>ssl.server.truststore.password</name><value>Welcome1</value></property>
> </configuration>
> ```
> 
> OM/SCM can be started without any problem. Based on the code I think It's a problem with the missing support of `CredentialProviders`. If you use localjceks, I would suggest to add this information to the NPE:
> 
> ConfigurationSource.java:
> 
> ```
>   default char[] getPassword(String key) throws IOException {
>     String value = get(key);
>     if (value == null) {
>       throw new NullPointerException(
>           "Password entry is missing for key " + key
>               + ".Note: generic ConfigurationSource interface doesn't support"
>               + " Hadoop CredentialProvider implementations");
>     }
>     return value.toCharArray();
>   }
> ```

@elek In the cluster that I tried out, the 'ssl.server.truststore.password' config key is not defined. Instead the getPassword relies on the Hadoop credential provider to get the truststore password.

Before this change, we relied on the Configuration#getPassword method while starting up the HTTP Server whose implementation looks like this.

`  public char[] getPassword(String name) throws IOException {
    char[] pass = null;

    pass = getPasswordFromCredentialProviders(name);

    if (pass == null) {
      pass = getPasswordFromConfig(name);
    }

    return pass;
  }`

After this change, the code logic becomes 

` default char[] getPassword(String key) throws IOException {
    return get(key).toCharArray();
  }`

Hence, we have skipped the logic of checking in Credential provider.

elek:
@avijayanhwx Thanks to explain it. This is the same what I found, it's not related to the TLS config it's related to the credential provider. What do you think about adding more verbose exception message as I proposed?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/LegacyHadoopConfigurationSource.java
elek: Long-term we might need a `getPassword()` for `ConfigurationSource` and we can have different implementations (with Hadoop Configuration we can use `CredentialProviders`, with different `ConfigurationSource` we can have different approaches). 

Currently we have a very naive approach for `ConfigurationSource` we read the value as is, but using `getPassword` instead of `get` shows that we need an encryption. 

But as of know getPassword() is not called only when we use Hadoop Configuration. Agree, we can remove it temporary, until we clean-up the HTTP utilities and switch to use full `Configuration`.

elek:
> What do you think about adding more verbose exception message as I proposed?

Or throw an exception.

Or remove it from the interface as Mukul suggested.

avijayanhwx:
Thank you for the reviews @elek , @mukul1987.

PR #872, opened 2020-04-27
HDDS-3488. Remove unnecessary jackson version definition from Ozone.

mukul1987:
Remove unnecessary jackson dependencies from Ozone.

https://issues.apache.org/jira/browse/HDDS-3488

## How was this patch tested?
Build succeeds after removing dependencies.


elek:
> But this is just pedantry as we should remove this section anyway as these dependencies are not defined/used anywhere else...

I was wrong. We got the jackson jar files as transitive dependencies from hadoop-common. In this case this section works as a version enforcement. If we need to remove the jar files from the classpath an exclude rule should be added to `hdds/hadoop-hdds/hadoop-dependency-server/`

elek:
Thanks the updated. Now al the classes are fully excluded. I tested it with build + `find -name "*jackson*" and these jackson jars are excluded.

elek:
Merged. Thanks the quick fix @mukul1987

PR #877, opened 2020-04-28
HDDS-3496. Make S3_ADMIN_NAME configurable

captainzmc:

With buckets and volumes created by s3g, the admin user is now ‘OzoneS3Manager’. We need to change it to other users. This can be changed to configurable.

https://issues.apache.org/jira/browse/HDDS-3496

## How was this patch tested?
1、ozone.acl.enableo ozone acl and set ozone.om.s3g.administrator=anyone in ozone-site.xml
2、create bucket use s3api：
aws s3api --endpoint-url http://s3g-host:9878 create-bucket --bucket=test
3、You can see the acl info of the bucket(by default is 'OzoneS3Manager'):
ozone sh bucket getacl s375201e1bbbdba75e600fcd0273c304ca/test

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -123,6 +123,11 @@ public final class OzoneConfigKeys {
    * */
   public static final String OZONE_ADMINISTRATORS_WILDCARD = "*";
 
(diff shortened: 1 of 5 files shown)


elek:
Thanks to fix this issue @captainzmc 

I am not sure if it's still relevant. I think it's not a problem any more as HDDS-3385 removed the create volume call (`s3v` volume is used for all the buckets not s3 bucket is created)

captainzmc:
thanks @elek. I'm going to turn off the PR. And we will evaluate whether the HDDS-3385 has any impact on our online clusters.

PR #878, opened 2020-04-28
HDDS-3469. Add third party jar versions as properties in pom.xml

adoroszlai:

Move most of dependency version definitions to root `pom.xml`.

* Except `csi`, to avoid conflict with #863
* Except `commons-lang`, which I would like to replace in few places with `commons-lang3` soon

https://issues.apache.org/jira/browse/HDDS-3469

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/625552925


adoroszlai:
Thanks @bshashikant for reviewing and merging this.

PR #881, opened 2020-04-28
HDDS-3081. Replication manager should detect and correct containers which don't meet the replication policy

sodonnel:

Now we have network topology available in Ozone, we need the ability to identify and correct containers where the replicas do not meet the replication policy. For network aware clusters, the rule is that the replicas must be on at least two racks. In the future, it is possible there will be other policies, eg node groups and so on.

Ideally, the existing ReplicationManager in SCM should be extended to detect, and correct these mis-replicated containers.

This change starts by adding a new method to the PlacementPolicy interface:

```
ContainerPlacementStatus validateContainerPlacement(
      List<DatanodeDetails> dns, int replicas);
```

This accepts the list of DNs currently hosting the replicas for a container, and the desired replication factor. It return an instance of `ContainerPlacementStatus` which has two methods indicating if the policy is met and if not, how many more replicas are needed:

```
boolean isPolicySatisfied();
int additionalReplicaRequired();
```

Currently, all placement policies extend `SCMCommonPlacementPolicy` and there is a default impelementation there, which handles both rack aware and non rack aware clusters. If we have further policies in the future (eg node groups, upgrade domain) this can be overridden in new subclasses.

Replication manager can then call:

```
placementPolicy.validateContainerPlacement(List<DatanodeDetails>, replicationFactor)
```

To get the information it needs to check the container.

For over replicated containers, when removing replicas, it is important to also ensure removing the container does not violate the policy, so some changes will be needed around the over-replicated code path too.

https://issues.apache.org/jira/browse/HDDS-3081

## How was this patch tested?

New unit tests

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
@@ -213,16 +213,6 @@ public abstract class SCMCommonPlacementPolicy implements PlacementPolicy {
       List<DatanodeDetails> healthyNodes);
 
   /**
-   * Default implementation for basic placement policies that do not have a
-   * placement policy. If the policy has not network topology this method should
-   * return null.
-   * @return The networkTopology for the policy or null if none is configured.
-   */
-  protected NetworkTopology getNetworkTopology() {
-    return null;
-  }
-
-  /**
    * Default implementation to return the number of racks containers should span
    * to meet the placement policy. For simple policies that are not rack aware
    * we return 1, from this default implementation.
@@ -249,7 +239,7 @@ public abstract class SCMCommonPlacementPolicy implements PlacementPolicy {
   @Override
   public ContainerPlacementStatus validateContainerPlacement(
       List<DatanodeDetails> dns, int replicas) {
-    NetworkTopology topology = getNetworkTopology();
+    NetworkTopology topology = nodeManager.getClusterNetworkTopologyMap();
     int requiredRacks = getRequiredRackCount();
     if (topology == null || replicas == 1 || requiredRacks == 1) {
       if (dns.size() > 0) {
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
@@ -25,8 +25,10 @@ import java.util.stream.Collectors;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
+import org.apache.hadoop.hdds.scm.container.placement.algorithms.ContainerPlacementStatusDefault;
 import org.apache.hadoop.hdds.scm.container.placement.metrics.SCMNodeMetric;
 import org.apache.hadoop.hdds.scm.exceptions.SCMException;
+import org.apache.hadoop.hdds.scm.net.NetworkTopology;
 import org.apache.hadoop.hdds.scm.node.NodeManager;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -198,4 +200,66 @@ public abstract class SCMCommonPlacementPolicy implements PlacementPolicy {
    */
   public abstract DatanodeDetails chooseNode(
       List<DatanodeDetails> healthyNodes);
+
+  /**
+   * Default implementation for basic placement policies that do not have a
+   * placement policy. If the policy has not network toplogy this method should
+   * return null.
+   * @return The networkTopology for the policy or null if none is configured.
+   */
+  protected NetworkTopology getNetworkTopology() {
+    return null;
+  }
+
+  /**
+   * Default implementation to return the number of racks containers should span
+   * to meet the placement policy. For simple policies that are not rack aware
+   * we return 1, from this default implementation.
+   * should have
+   * @return The number of racks containers should span to meet the policy
+   */
+  protected int getRequiredRackCount() {
+    return 1;
+  }
+
+  /**
+   * This default implementation handles rack aware policies and non rack
+   * aware policies. If a future placement policy needs to check more than racks
+   * to validate the policy (eg node groups, HDFS like upgrade domain) this
+   * method should be overridden in the sub class.
+   * This method requires that subclasses which implement rack aware policies
+   * override the default method getRequiredRackCount and getNetworkTopology.
+   * @param dns List of datanodes holding a replica of the container
+   * @param replicas The expected number of replicas
+   * @return ContainerPlacementStatus indicating if the placement policy is
+   *         met or not. Not this only considers the rack count and not the
+   *         number of replicas.
+   */
+  @Override
+  public ContainerPlacementStatus validateContainerPlacement(
+      List<DatanodeDetails> dns, int replicas) {
+    NetworkTopology topology = getNetworkTopology();
+    int requiredRacks = getRequiredRackCount();
+    if (topology == null || replicas == 1 || requiredRacks == 1) {
+      // placement is always satisfied if there is at least one DN.
+      return new ContainerPlacementStatusDefault(dns.size(), 1, 1);
+    }
+    // We have a network topology so calculate if it is satisfied or not.
+    int numRacks = 1;
+    final int maxLevel = topology.getMaxLevel();
+    // The leaf nodes are all at max level, so the number of nodes at
+    // leafLevel - 1 is the rack count
+    numRacks = topology.getNumOfNodes(maxLevel- 1);
+    final long currentRackCount = dns.stream()
+        .map(d -> topology.getAncestor(d, 1))
+        .distinct()
+        .count();
+
+    if (replicas < requiredRacks) {
+      requiredRacks = replicas;
+    }
+
+    return new ContainerPlacementStatusDefault(
+        (int)currentRackCount, requiredRacks, numRacks);
+  }
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
@@ -32,7 +32,6 @@ import org.apache.hadoop.hdds.scm.net.NetworkTopology;
 import org.apache.hadoop.hdds.scm.node.NodeManager;
 
 import com.google.common.annotations.VisibleForTesting;
-import org.apache.hadoop.ozone.container.common.interfaces.Container;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -51,10 +50,14 @@ public abstract class SCMCommonPlacementPolicy implements PlacementPolicy {
 
   /**
    * Return for replication factor 1 containers where the placement policy
-   * is always met, rather than creating a new object each time.
+   * is always met, or not met (zero replicas available) rather than creating a
+   * new object each time. There are only used when there is no network topology
+   * or the replication factor is 1 or the required racks is 1.
    */
   private ContainerPlacementStatus validPlacement
       = new ContainerPlacementStatusDefault(1, 1, 1);
+  private ContainerPlacementStatus invalidPlacement
+      = new ContainerPlacementStatusDefault(0, 1, 1);
 
   /**
    * Constructor.
@@ -249,8 +252,12 @@ public abstract class SCMCommonPlacementPolicy implements PlacementPolicy {
     NetworkTopology topology = getNetworkTopology();
     int requiredRacks = getRequiredRackCount();
     if (topology == null || replicas == 1 || requiredRacks == 1) {
-      // placement is always satisfied if there is at least one DN.
-      return validPlacement;
+      if (dns.size() > 0) {
+        // placement is always satisfied if there is at least one DN.
+        return validPlacement;
+      } else {
+        return invalidPlacement;
+      }
     }
     // We have a network topology so calculate if it is satisfied or not.
     int numRacks = 1;
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
@@ -258,7 +258,6 @@ public abstract class SCMCommonPlacementPolicy implements PlacementPolicy {
     if (replicas < requiredRacks) {
       requiredRacks = replicas;
     }
-
     return new ContainerPlacementStatusDefault(
         (int)currentRackCount, requiredRacks, numRacks);
   }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
@@ -32,6 +32,7 @@ import org.apache.hadoop.hdds.scm.net.NetworkTopology;
 import org.apache.hadoop.hdds.scm.node.NodeManager;
 
 import com.google.common.annotations.VisibleForTesting;
+import org.apache.hadoop.ozone.container.common.interfaces.Container;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -47,6 +48,13 @@ public abstract class SCMCommonPlacementPolicy implements PlacementPolicy {
   private final NodeManager nodeManager;
   private final Random rand;
   private final ConfigurationSource conf;
+
+  /**
+   * Return for replication factor 1 containers where the placement policy
+   * is always met, rather than creating a new object each time.
+   */
+  private ContainerPlacementStatus validPlacement
+      = new ContainerPlacementStatusDefault(1, 1, 1);
 
   /**
    * Constructor.
@@ -203,7 +211,7 @@ public abstract class SCMCommonPlacementPolicy implements PlacementPolicy {
 
   /**
    * Default implementation for basic placement policies that do not have a
-   * placement policy. If the policy has not network toplogy this method should
+   * placement policy. If the policy has not network topology this method should
    * return null.
    * @return The networkTopology for the policy or null if none is configured.
    */
@@ -242,14 +250,14 @@ public abstract class SCMCommonPlacementPolicy implements PlacementPolicy {
     int requiredRacks = getRequiredRackCount();
     if (topology == null || replicas == 1 || requiredRacks == 1) {
       // placement is always satisfied if there is at least one DN.
-      return new ContainerPlacementStatusDefault(dns.size(), 1, 1);
+      return validPlacement;
     }
     // We have a network topology so calculate if it is satisfied or not.
     int numRacks = 1;
     final int maxLevel = topology.getMaxLevel();
     // The leaf nodes are all at max level, so the number of nodes at
     // leafLevel - 1 is the rack count
-    numRacks = topology.getNumOfNodes(maxLevel- 1);
+    numRacks = topology.getNumOfNodes(maxLevel - 1);
     final long currentRackCount = dns.stream()
         .map(d -> topology.getAncestor(d, 1))
         .distinct()
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -39,6 +39,7 @@ import org.apache.hadoop.hdds.conf.ConfigType;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.LifeCycleState;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerDatanodeProtocolProtos.ContainerReplicaProto.State;
+import org.apache.hadoop.hdds.scm.ContainerPlacementStatus;
 import org.apache.hadoop.hdds.scm.PlacementPolicy;
 import org.apache.hadoop.hdds.scm.events.SCMEvents;
 import org.apache.hadoop.hdds.scm.safemode.SCMSafeModeManager.SafeModeStatus;
@@ -379,7 +380,8 @@ public class ReplicationManager
    */
   private boolean isContainerHealthy(final ContainerInfo container,
                                      final Set<ContainerReplica> replicas) {
-    return container.getReplicationFactor().getNumber() == replicas.size() &&
+    return !isContainerUnderReplicated(container, replicas) &&
+        !isContainerOverReplicated(container, replicas) &&
         replicas.stream().allMatch(
             r -> compareState(container.getState(), r.getState()));
   }
@@ -393,8 +395,11 @@ public class ReplicationManager
    */
   private boolean isContainerUnderReplicated(final ContainerInfo container,
       final Set<ContainerReplica> replicas) {
+    boolean misReplicated = !getPlacementStatus(
+        replicas, container.getReplicationFactor().getNumber())
+        .isPolicySatisfied();
     return container.getReplicationFactor().getNumber() >
-        getReplicaCount(container.containerID(), replicas);
+        getReplicaCount(container.containerID(), replicas) || misReplicated;
   }
 
   /**
@@ -501,6 +506,11 @@ public class ReplicationManager
           .stream()
           .map(action -> action.datanode)
           .collect(Collectors.toList());
+      final List<DatanodeDetails> replicationInFlight = inflightReplication
+          .getOrDefault(id, Collections.emptyList())
+          .stream()
+          .map(action -> action.datanode)
+          .collect(Collectors.toList());
       final List<DatanodeDetails> source = replicas.stream()
           .filter(r ->
               r.getState() == State.QUASI_CLOSED ||
@@ -512,25 +522,61 @@ public class ReplicationManager
       if (source.size() > 0) {
         final int replicationFactor = container
             .getReplicationFactor().getNumber();
-        final int delta = replicationFactor - getReplicaCount(id, replicas);
+        // Want to check if the container is mis-replicated after considering
+        // inflight add and delete.
+        // Create a new list from source (healthy replicas minus pending delete)
+        List<DatanodeDetails> targetReplicas = new ArrayList<>();
+        targetReplicas.addAll(source);
+        // Then add any pending additions
+        targetReplicas.addAll(replicationInFlight);
+
+        int delta = replicationFactor - getReplicaCount(id, replicas);
+        final int additionalRacks
+            = containerPlacement.validateContainerPlacement(
+                targetReplicas, replicationFactor).additionalReplicaRequired();
+        final int replicasNeeded
+            = delta < additionalRacks ? additionalRacks : delta;
+
         final List<DatanodeDetails> excludeList = replicas.stream()
             .map(ContainerReplica::getDatanodeDetails)
             .collect(Collectors.toList());
+        excludeList.addAll(replicationInFlight);
         List<InflightAction> actionList = inflightReplication.get(id);
         if (actionList != null) {
           actionList.stream().map(r -> r.datanode)
               .forEach(excludeList::add);
         }
         final List<DatanodeDetails> selectedDatanodes = containerPlacement
-            .chooseDatanodes(excludeList, null, delta,
+            .chooseDatanodes(excludeList, null, replicasNeeded,
                 container.getUsedBytes());
-
-        LOG.info("Container {} is under replicated. Expected replica count" +
-                " is {}, but found {}.", id, replicationFactor,
-            replicationFactor - delta);
-
-        for (DatanodeDetails datanode : selectedDatanodes) {
-          sendReplicateCommand(container, datanode, source);
+        if (delta > 0) {
+          LOG.info("Container {} is under replicated. Expected replica count" +
+                  " is {}, but found {}.", id, replicationFactor,
+              replicationFactor - delta);
+        }
+        int newAdditionalRacks = additionalRacks;
+        if (additionalRacks > 0) {
+          LOG.info("Container {} is mis-replicated. Is should be on {} " +
+              "additional racks", id, additionalRacks);
+          // Check if the new target nodes (original plus newly selected nodes)
+          // makes the placement policy valid.
+          targetReplicas.addAll(selectedDatanodes);
+          newAdditionalRacks = containerPlacement.validateContainerPlacement(
+              targetReplicas, replicationFactor).additionalReplicaRequired();
+        }
+        if (delta > 0 || newAdditionalRacks < additionalRacks) {
+          // Only create new replicas if we are missing a replicas or
+          // the number of pending racks has improved. No point in creating
+          // new replicas for mis-replicated containers unless it improves
+          // things
+          for (DatanodeDetails datanode : selectedDatanodes) {
+            sendReplicateCommand(container, datanode, source);
+          }
+        } else {
+          LOG.warn("Container required {} additional racks and {} additional " +
+              "replicas. After selecting new nodes, {} racks are still " +
+              "required. No new replicas will be scheduled.",
+              additionalRacks, Integer.max(delta, 0), newAdditionalRacks);
         }
       } else {
         LOG.warn("Cannot replicate container {}, no healthy replica found.",
@@ -591,6 +637,21 @@ public class ReplicationManager
             eligibleReplicas.get(i).getDatanodeDetails(), true);
       }
     }
+  }
+
+  /**
+   * Given a set of ContainerReplica, transform it to a list of DatanodeDetails
+   * and then check if the list meets the container placement policy.
+   * @param replicas List of containerReplica
+   * @param replicationFactor Expected Replication Factor of the containe
+   * @return ContainerPlacementStatus indicating if the policy is met or not
+   */
+  private ContainerPlacementStatus getPlacementStatus(
+      Set<ContainerReplica> replicas, int replicationFactor) {
+    List<DatanodeDetails> replicaDns = replicas.stream()
+        .map(c -> c.getDatanodeDetails()).collect(Collectors.toList());
+    return containerPlacement.validateContainerPlacement(
+        replicaDns, replicationFactor);
   }
 
   /**
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -21,6 +21,7 @@ package org.apache.hadoop.hdds.scm.container;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -602,7 +603,7 @@ public class ReplicationManager
     final ContainerID id = container.containerID();
     final int replicationFactor = container.getReplicationFactor().getNumber();
     // Dont consider inflight replication while calculating excess here.
-    final int excess = replicas.size() - replicationFactor -
+    int excess = replicas.size() - replicationFactor -
         inflightDeletion.getOrDefault(id, Collections.emptyList()).size();
 
     if (excess > 0) {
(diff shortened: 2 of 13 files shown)


sodonnel:
Note, a further commit is required handle over replicated containers before this change is complete, but sharing the work in progress to give visibility of the change.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -512,25 +523,61 @@ private void handleUnderReplicatedContainer(final ContainerInfo container,
...
+        // Want to check if the container is mis-replicated after considering
+        // inflight add and delete.
+        // Create a new list from source (healthy replicas minus pending delete)
+        List<DatanodeDetails> targetReplicas = new ArrayList<>();
+        targetReplicas.addAll(source);
+        // Then add any pending additions
+        targetReplicas.addAll(replicationInFlight);
+
+        int delta = replicationFactor - getReplicaCount(id, replicas);
+        final int additionalRacks
```
avijayanhwx, line 535: If delta > 0, do we still need to validate the placement status? Cannot we just choose a DN in accordance to policy and issue the replication command? In that case, this logic can be executed only if we are "misreplicated" and not "underreplicated".

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -582,17 +629,71 @@ private void handleOverReplicatedContainer(final ContainerInfo container,
...
-      eligibleReplicas.addAll(0, unhealthyReplicas);
-
-      for (int i = 0; i < excess; i++) {
-        sendDeleteCommand(container,
-            eligibleReplicas.get(i).getDatanodeDetails(), true);
+      // If there are unhealthy replicas, then we should remove them even if it
+      // makes the container violate the placement policy, as excess unhealthy
+      // containers are not really useful. It will be corrected later as a
+      // mis-replicated container will be seen as under-replicated.
+      for (ContainerReplica r : unhealthyReplicas) {
```
avijayanhwx, line 636: Can we use something like findAny() here instead of breaking from the loop after first iteration? 'excess' is always > 0 here.

fapifta:
Hi @sodonnel, thank you for this work. As I see the code itself seems to be fairly correct as far as I can tell based on just reading through the changes I haven't seen a logical issue so far.

After the first overview, I went through some of the related code, and the current implementation, and then the changes again, and the following evolved in my mind during the process:
I really liked the idea for the first time to have a default implementation for container placement status in an immutable class, but thinking it through, I would drop the whole ContainerPlacementStatus interface and the related code.
My reason to tell this is a bit long and affects large portions of the solution.
As I get the intent, rack awareness when set needs to be satisfied, but in the code as it seems ReplicationManager goes far beyond its limits, and contains concrete rack awareness related logic. I think the placement policy itself should encapsulate the fact that there are racks, because if it does not, then node group based policy will pollute ReplicationManager with node group related code, and so on with any other specific implementation logic.

As.I look at the changeset from this angle, changes in ReplicationManager that mentions racks, should be internally inside the placement policy implementation.
What I would suggest, is to change the PlacementPolicy interface, and add the following methods:
boolean isValidPlacement(Set<ContainerReplica> replicas, int replicationFactor)
void removeExcessReplicas(Set<ContainerReplica> currentReplicas, int replicationFactor, BiConsumer<ContainerInfo, DatanodeDetails> purger)
void addMissingReplicas(Set<ContainerReplica> currentReplicas, int replicationFactor, BiConsumer<ContainerInfo, DatanodeDetails> replicator)

With this, the placement validity can be checked by the isValidPlacement method where needed, after that the two other method can be called if the placement is invalid.
The BiConsumers accept the containerId, and the DataNodeInfo, and removeExcessReplicas can take a BiConsumer like this: (container, datanode) -> sendDeleteCommand(container, datanode, true), while the addMissingReplicas can take a BiConsumer like this: (container, datanode) -> sendReplicateCommand(container, datanode, sources).

With this approach I believe all the racks related code can be part of the ContainerPlacementRackAware, and the PipelinePlacementPolicy classes. Also I suggest to use Set<ContainerReplica> as the parameter everywhere in the PlacementPolicy, as that is coming in, and DatanodeDetails and other things can be calculated only when they are needed from that.

After the change in the approach, the policies themselves have to handle the over and under replication via these methods, and these methods can be tested fairly similarly as in the TestContainerPlacementRackAware class it is added now, while I would suggest to change how we test the ReplicationManager, to make that part simpler as well, as there you can just provide a mock ContainerManager that provides the container ids, and replica informations as you wish, and then you can run the manager, and check if the proper events arrived to the queue ideally by spying on the EventQueue and verifying the events posted. I am suggesting this, as we purely want to check if the events are posted to the queue, and we do not want to validate the queue or anything else, so we want to provide containers for the ReplicationManager, and then check it it posts the expected commands given the policy and the container replicas we provided.

I know that this is a fundamental change, but at the moment I am -1 on the implementation because of the fact that in the current state we would make the ReplicationManager also rack aware which I think we would like to avoid.

fapifta:
Thinking it a bit further, the current implementation basically brings rack as a term into ReplicaManager just via integers and counts, so I was a bit harsh when stating that effectively we make ReplicationManager rack aware.
However I still think this whole logic of handle(Over|Under)ReplicatedContainers has to be part of the placement policy, as that should be the only point that handles the placement of containers, and if we leave this logic in ReplicationManager, then it will explode further and will be way harder to maintain as soon as a new placement policy is added, moreover it makes it impossible to really plug in a new placement policy by just a config change, as it is planned now.

sodonnel:
@fapifta, Thanks for looking at this.

As things stand prior to this change, the job of the placement policy is to return a list of nodes to create replicas on.

The job of the replication manager is to check the the replication, that replicas are all healthy and now placement, is met, and if it is not ask the policy for new nodes to schedule replicas on. In the future, it will most likely need to deal with maintenance mode replicas and decommission.

Similarly, the role of pipeline manager is to ask the policy for nodes to establish pipelines on and the pipeline manager actually creates them.

In that respect the placement policy is simply a selector of nodes and I don't think it makes sense to give it an additional responsibility to actually destroy (or create) replicas.

I do think I have made some unfortunate variable naming choices. As you said in your second comment, RM is not really rack aware. It is "placement aware" via the ContainerPlacementStatus.

There may be some merit to extending the placement policy to decide which replicas to remove, but then it will:

* Need to consider the health of them (like RM does now)
* Decommission related status in the future
* The logic doesn't really apply to Pipelines, only for closed containers, so pushing this logic into the common parent may not make sense, but it would need to go there due to the inheritance chain.

I think it would be good to clean up the naming to remove "rack" and use something more general to make it clear the rack logic is not baked into RM.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
```diff
@@ -198,4 +200,65 @@ public boolean hasEnoughSpace(DatanodeDetails datanodeDetails,
    */
   public abstract DatanodeDetails chooseNode(
       List<DatanodeDetails> healthyNodes);
+
+  /**
+   * Default implementation for basic placement policies that do not have a
+   * placement policy. If the policy has not network toplogy this method should
```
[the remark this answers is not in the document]
sodonnel: fixed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
```diff
@@ -198,4 +200,65 @@ public boolean hasEnoughSpace(DatanodeDetails datanodeDetails,
...
+   *         number of replicas.
+   */
+  @Override
+  public ContainerPlacementStatus validateContainerPlacement(
+      List<DatanodeDetails> dns, int replicas) {
+    NetworkTopology topology = getNetworkTopology();
+    int requiredRacks = getRequiredRackCount();
+    if (topology == null || replicas == 1 || requiredRacks == 1) {
+      // placement is always satisfied if there is at least one DN.
+      return new ContainerPlacementStatusDefault(dns.size(), 1, 1);
```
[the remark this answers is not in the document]
sodonnel: Are you suggesting that I should have a pre-created "policy is met object" and just return that same object here, rather than a new one each time? If so, that is a good idea. I have made that change assuming that is what you meant.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
```diff
@@ -198,4 +200,65 @@ public boolean hasEnoughSpace(DatanodeDetails datanodeDetails,
...
+    }
+    // We have a network topology so calculate if it is satisfied or not.
+    int numRacks = 1;
+    final int maxLevel = topology.getMaxLevel();
+    // The leaf nodes are all at max level, so the number of nodes at
+    // leafLevel - 1 is the rack count
+    numRacks = topology.getNumOfNodes(maxLevel- 1);
```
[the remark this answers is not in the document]
sodonnel: fixed.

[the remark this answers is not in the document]
sodonnel: That is a good point. This should be static except when nodes register or are removed, which is rare. 

I don't really want to pass "rack count" into this method, as in the future, we may have other implementations of this (eg node groups, upgrade domain etc), and putting that into the interface would not be ideal.

I wonder if it would make sense to cache the count for a given level in `NetworkTopologyImpl`, rather than passing around the count here? The only things that change it, are add() and remove(), so we could easily have a hash map or list to store the count for a given level, and upon any add or remove wipe the cached values. This would be fairly simple to do and would benefit any other parts of the code which happen to call this existing method.

Do you think that makes sense?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -512,25 +523,61 @@ private void handleUnderReplicatedContainer(final ContainerInfo container,
       if (source.size() > 0) {
         final int replicationFactor = container
             .getReplicationFactor().getNumber();
-        final int delta = replicationFactor - getReplicaCount(id, replicas);
+        // Want to check if the container is mis-replicated after considering
+        // inflight add and delete.
+        // Create a new list from source (healthy replicas minus pending delete)
+        List<DatanodeDetails> targetReplicas = new ArrayList<>();
```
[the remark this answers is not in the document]
sodonnel: Fixed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: For now, mis-replicated count can only return 0 or 1, as replicas must be on at most 2 racks, and if there is an existing replica, it must be on a rack already.
However, if we introduced a new policy "must be on 3 racks", then it would be possible to be under replicated with a delta of 1 and mis-replicated count could return 2, meaning we need to schedule more than 1 replica now.

You suggestion is valid if we only consider "two racks" as the only option, but I am hoping to cover off future enhancements here too.

Do you think that makes sense?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: There could be more unhealthyReplicas than excess, and in that case, we want to break if we have removed the excess count. The loop will always make at least one pass.

I am not sure how we can use findAny() to make this cleaner, while also decrementing excess.

Remember that it is possible to have an excess of, say 2, with one unhealthy. So we would use the logic here to remove that unhealthy one unconditionally. Then fall into the other logic to remove 1 more.

If you have a cleaner idea about how to do this please post a short code snippet.

fapifta:
Somehow I still feel bad about the current separation of concerns...

But let's go one by one:
It is not a problem to consider node state in the PlacementPolicy, as the policies are already depend on NodeManager, and selects just from healthy nodes. Decommissioning state will/should be available via the NodeManager also for the policies.
Pipeline placement can use the current interface, and we can introduce a new one called ContainerPlacementPolicy extending PlacementPolicy and defining the necessary methods.
Strictly speaking pipeline placement does not need the validateContainrePlacement method either.

My problem is a bit more fundamental, and I am mainly concerned about the complexity we would be introducing here to have a generalized-ish algorythm with a secondary condition that has to be met and checked for all placement policies, and for all except one placement policies we supply the dummy values to do not do anything differently then before. While on the other hand we are completely unsure whether we will ever need a tertiary condition for an other placement policy in the future or we won't. This is why I suggested a new approach, and reverse the control, as the way how we remove the excess replicas, or how we add new replicas via the ReplicationManager will not foreseeably change in the future much, and as we can provide a callback mechanism to the policies without introducing new classes and hard to understand indirections, but without coupling, I still think that is the better way to implement this.

What I fear the most if we let this through, is that the next one comes to this code after adding a policy either need to do a magic inside the policy, to provide the secondary condition properly, maybe mix together a secondary and tertiary condition to fit into the current interface, as we can not broke compatibility, and at the end of the day ReplicationManager gets modified again to accept the new stuff, as it is easier to add. I do not see this with my suggestion, moreover I do not see any reason after that to change the ReplicationManager ever again significantly.
Also in the suggested version, we can test the mis-replication detection and fix specifically with just the policies, and leave the ReplicationManager just an executor of what the policy suggests. I hope it is and understandable fear, I am not sure it is addressable within the current implementation though... :) But I am open to even chat about it.


This was how I summarised and understood first what was happening, and what will be introduced with the new code.
At the beginning we have ReplicationManager, we have NodeManager, we have the PlacementPolicy.
 - ReplicationManager is to ensure that the number of replicas "considered live" is exactly as much as the replication factor of a container.
 - NodeManager is to provide Node status, and information. I am unsure why it knows about NetworkTopology, and whether it should but for now it knows about topology also.
 - PlacementPolicy is there to provide information about the replica placement rules, based on NodeManager's data.

Now we want to extend a functionality that was hardcoded into ReplicationManager before.
The functionality works now as:
 - if a container is open, leave it as it is
 - if a container is closing, send the close command to the container replicas
 - if a container is quasy_closed, force the closure of the replicas
 - update container replica states about inflight and deleted replicas if needed based on current action
 - if the container is healthy leave it as it is
 - if the container is under replicated handle under-replication
 - if the container is over replicated handle over-replication

How we handle mis-replication before patch:
 - at under replication, we use the PlacementPolicy to get a set of nodes that are proper to hold the container replicas considering the exclusion list consists of unhealthy or in-weird-state replicas, and we replicate to the nodes that are coming back from the policy.
- at over replication, we calculate the number of excess replicas, and remove those that are not necessary/unhealthy/in-weird-state.


In the new code, what is changing is how we handle mis-replication, the rest of the code is unchanged. The new code suggest the following way:
 - at under-replication we start to have a secondary condition expressed by a number of missing something (in this case rack), and we check against this number and the number of missing replicas, and if none-of them is met and both of them are improving we actually re-replicate, otherwise next time we will check again.
 - at over-replication we start to have a secondary condition expressed by a number and we check against it again, to see if we remove a replica then the system still met a condition on this number, and remove only replicas that are not ruining the condition.

Until we only have the rack-aware placement with this secondary condition, it can be named as rackSomething, however at the point in time when we introduce a new one, this naming will not stand the probe of time for sure, and we might end up with more general more hard to understand names. The architecture probably will not stand either, as this generalism here is limiting our possibility to write a policy specific easy to understand algorithm to find the excess replicas and to add new replicas. Also the current algorithm is a bit harder to understand if we can not use the policy specific terms as I see it at the moment.

sodonnel:
I am still not convinced putting the actual remove or add logic into the placement policies is the correct way to go here. That would require fairly fundamental changes to the existing replication manager, which I would prefer to leave alone as much as possible. It would also change the intent of the policies, which are read only "selectors" right now.

> However I still think this whole logic of handle(Over|Under)ReplicatedContainers has to be part of the placement policy, as that should be the only point that handles the placement of containers, and if we leave this logic in ReplicationManager, then it will explode further and will be way harder to maintain as soon as a new placement policy is added, moreover it makes it impossible to really plug in a new placement policy by just a config change, as it is planned now.

I have removed any reference to rack from replication manager (there were only variable names and not concrete logic) - now it deals with under over replicated and mis-replication. I believe we could plug a new policy in that handles 3 racks, or node groups with zero changes to replication manager, and just by extending the relevant placement policy classes. Considering what we have in HDFS, its hard to imagine a placement policy that does not fit into this model.

We already have an API into the placement policies to pick new nodes to add (chooseDataNodes(...)) and this is used by the replication manager and pipeline manager, and perhaps elsewhere. Replication manager uses this to handle the under and mis-replication.

What is missing is, perhaps, is an API called `chooseNodesToRemove()` in the placement policies. That would allow us to decide to do things, such as remove a replica from a more used node in some policies, but for now we have just "remove any healthy node that does not violate the policy". With only that current requirement, the logic added in this PR, which simply tries removing each replica in turn is probably OK.

Remember that RM also has to keep track of the inflight replications, and consider them when running the calculations on the containers.

I also feel that some of your suggestions are wider refactors, and ideally we should tackle them later rather than combine them with this change:

 * Change how we test replication manager
 * Use Set<ContainerReplica> everywhere in the placement policies

fapifta:
I am still not convinced either, but the argument about inflight actions is a valid point I am unsure about but let's just drop this.

As I see it will be hard to get to a common ground as we have fairly different points of view, I can accept yours, though I still don't believe in it :) Let's leave it now as it is, and we will see how it goes in the future.

On the other hand if we go this way, I have a few concerns about the implementation details still, I am adding them inline.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
```diff
@@ -198,4 +211,69 @@ public boolean hasEnoughSpace(DatanodeDetails datanodeDetails,
...
   public abstract DatanodeDetails chooseNode(
       List<DatanodeDetails> healthyNodes);
+
+  /**
+   * Default implementation for basic placement policies that do not have a
+   * placement policy. If the policy has not network topology this method should
+   * return null.
+   * @return The networkTopology for the policy or null if none is configured.
+   */
+  protected NetworkTopology getNetworkTopology() {
```
fapifta, line 221: Considering that we have the NetworkTopology inside NodeManager, do we need this method?
It is used only in the validateContainerPlacement method where we can use NodeManager also.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
```diff
@@ -198,4 +211,69 @@ public boolean hasEnoughSpace(DatanodeDetails datanodeDetails,
...
+    NetworkTopology topology = getNetworkTopology();
+    int requiredRacks = getRequiredRackCount();
+    if (topology == null || replicas == 1 || requiredRacks == 1) {
+      if (dns.size() > 0) {
+        // placement is always satisfied if there is at least one DN.
+        return validPlacement;
+      } else {
+        return invalidPlacement;
+      }
+    }
```
fapifta, line 261: This part of the method is clearly dealing with non-topology-aware stuff, and returns, while the rest is dealing with topology-aware stuff. Why we don't use polymorphism here to separate the two? I understand that this requires copying the topology-aware stuff to two places, but that is an other problem we have to deal with at one point.
So I believe this if should be the default implementation in this class, and the rest should be inside the rack aware and pipeline placement policy.

This would render the getRequiredRackCount to be unnecessary as well, as here it would be possible to just remove the expression yielding into true always for non-topology-aware policies, while the topology aware ones can have and use the constant in their implementation.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -393,8 +396,11 @@ private boolean isContainerHealthy(final ContainerInfo container,
    */
   private boolean isContainerUnderReplicated(final ContainerInfo container,
       final Set<ContainerReplica> replicas) {
+    boolean misReplicated = !getPlacementStatus(
```
fapifta, line 399: I understand that this is here to check if we have less than 2 racks with replicas, and if so, consider the container to be under replicated.
I am unsure though if we want to ensure that a container is replicated to exactly 2 racks, it does not seems to be the case in the rack aware placement policy, but if so, then mis-replication should be checked somewhere else also to handle the case when a container is replicated to 3 racks due to for example removing a rack from config and place the nodes in a rack to other racks. (Probably this can go to isContainerOverReplicated(), if we want to check for this case.)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -512,25 +523,60 @@ private void handleUnderReplicatedContainer(final ContainerInfo container,
...
+        // Create a new list from source (healthy replicas minus pending delete)
+        List<DatanodeDetails> targetReplicas = new ArrayList<>(source);
+        // Then add any pending additions
+        targetReplicas.addAll(replicationInFlight);
+
+        int delta = replicationFactor - getReplicaCount(id, replicas);
+        final ContainerPlacementStatus placementStatus =
+            containerPlacement.validateContainerPlacement(
+                targetReplicas, replicationFactor);
+        final int misRepDelta = placementStatus.additionalReplicaRequired();
```
fapifta, line 537: This method in reality gives back the additional rack count required, not the additional replicas required.
This is why we need to check against the replication factor and the replica count difference also, while we provide the replication factor to the validation method, so it would be able to return the number of replicas really required.

Let see the different cases:
 - non-topology aware placement policies: if replica count == replication factor we are good, otherwise the required additional replica count is replication factor - replica count, and we need that many extra replicas later from chooseDataNodes.
 - a topology aware placement policy with replication factor of 1 in case it is under replicated we already miss the block
 - a topology aware placement policy with replication factor of 2 in case it is under replicated either we miss the block, or we have it in one rack, and chooseDatanodes will give us the required one in a different rack.
 - topology aware placement policies with replication factor of 3 (or more):
   - when we have 1 healthy replica, we need 2 additional replicas anyways, it is not really interesting in which rack.
   - when we have 2 healthy replica, we need 1 additional replica from either one of the racks in which we already have a replica or in an other rack instead if we have the two replica in the same rack, this the policy already handles as it uses the excluded node list to anchor to proper racks in chooseDatanodes().
   - when we have 3 healthy replica, but all three is in the same rack, then additionalReplicaRequired should yield to 1 still, and chooseDataNode should select a new set of DataNodes correctly, consisting of the nodes already have the replica, and one additional in a different rack. Which makes the container over replicated, so one excess replica will be removed later, after replication finished.

So if the additionalReplicaRequired works as the name suggests, we should not need to check replication factor and replica count here, but just rely on the value given back by the method.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -512,25 +523,60 @@ private void handleUnderReplicatedContainer(final ContainerInfo container,
...
+        final int replicasNeeded
+            = delta < misRepDelta ? misRepDelta : delta;
+
         final List<DatanodeDetails> excludeList = replicas.stream()
             .map(ContainerReplica::getDatanodeDetails)
             .collect(Collectors.toList());
+        excludeList.addAll(replicationInFlight);
```
fapifta, line 544: Do we add the in-flight replications twice here this way, or I am missing something?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -512,25 +523,60 @@ private void handleUnderReplicatedContainer(final ContainerInfo container,
...
+        }
+        int newMisRepDelta = misRepDelta;
+        if (misRepDelta > 0) {
+          LOG.info("Container: {}. {}",
+              id, placementStatus.misReplicatedReason());
+          // Check if the new target nodes (original plus newly selected nodes)
+          // makes the placement policy valid.
+          targetReplicas.addAll(selectedDatanodes);
+          newMisRepDelta = containerPlacement.validateContainerPlacement(
+              targetReplicas, replicationFactor).additionalReplicaRequired();
```
fapifta, line 566: The rack-aware policy's chooseDatanodes() method we use ensures that based on the excluded nodes list we get back a proper list, I believe we do not need to check it twice Are you aware of any cases where this is not true? Based on my understanding of the logic there I would expect all future placement policies to do so, as other existing ones implicitly do so now. If necessary we can add this to the contract of the interface, though this is an interesting property of the rack-aware policy that would worth some re-thinking but that is out of scope for now.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
sodonnel: The constructor for PipelinePlacementPolicy only receives nodeManager:

``` public PipelinePlacementPolicy(final NodeManager nodeManager,
      final PipelineStateManager stateManager, final ConfigurationSource conf) {
```

But the container one receives a networkTopology as part of its constructor:

```
 public SCMContainerPlacementRackAware(final NodeManager nodeManager,
      final ConfigurationSource conf, final NetworkTopology networkTopology,
      final boolean fallback, final SCMContainerPlacementMetrics metrics) {
```

The pipline Policy always asks the nodeManager for the policy, but the Container one uses the one passed. Ideally they should be the same, but I have a feeling some tests may make use of this "feature".

The other old placement policies (SCMContainerPlacementRandom, SCMContainerPlacementCapacity) have the same constructors, but they don't even use the topology at all.

So, I think we can get rid of this new method and then we should create another Jira to do the small refactor on all those older constructors to remove network topology from the parameter list and ensure they pull the topology from the node manager - what do you think?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
sodonnel: I kept it here, to ensure the logic is not duplicated. Ideally, we move all the rack aware logic into a common parent and have PipelinePlacementPolicy and ContainerPlacementRackAware inherit from that. However there was some push back on that earlier, and its probably a non-trivial task.

Rather than duplicating the logic, I think it is better where it is and then we have HDDS-3079 (which probably needs renamed) to extract the rack aware stuff into one common parent.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: My thinking on this, is that the policy states the replica "must be on AT LEAST two racks". I don't believe we should consider it over-replicated if it turns out to be on 3 somehow. That logic is consistent with HDFS, and probably provides better redundancy and locality (more racks is likely better in almost all cases).

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: The PlacementStatus.additionalReplicaRequired() method is intended to state how many replicas are need to meet the placement policy and not the replication factor. In other places (eg Recon) it might be useful to be able to get the "additional racks required" separately from under replication.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: Well spotted. I think I can remove the logic after this line, as I already have created that list earlier in the method now:

```
        if (actionList != null) {
          actionList.stream().map(r -> r.datanode)
              .forEach(excludeList::add);
        }
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: > The rack-aware policy's chooseDatanodes() method we use ensures that based on the excluded nodes list we get back a proper list

This is where it gets tricky. There is a fallback (certainly in the pipeline placement policy, but I have not checked the ContainerPlacementRackAware in complete detail) where if the placement policy cannot select enough racks, it will give back nodes which do not meet the placement.

This is to ensure (in the case of pipelinePlacementPolicy) that the cluster can keep on operating if only 1 rack is available for some reason. For the closed container placement, this should be the case too, as if you are down to 1 replica, it is better it gets replicated onto more nodes on the same rack, rather than risk losing the data.

We probably need a future change to scrub the bad pipelines if this fall back happens, while replication manager should deal with closed containers.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
fapifta: Sounds right, let's do so. 👍

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
fapifta: I am afraid that anyone who gets to HDDS-3079 will not find this out, maybe don't even check... It would be better to duplicate, as we feel the pain more and more of this, or at least to mention this method as well as the target for the JIRA ;)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
fapifta: Sounds reasonable for me, let's leave it this way, I wanted to be sure we considered this when it came to my minds as a possible starting state.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
fapifta: 👍

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
fapifta: Right, I see your point... I am not sure, haven't checked, but I guess based on NetworkTopology, and maybe NodeManager we can tell how many healthy racks we have, and if we have one, and the fallback is enabled, then we can report the number of required replicas to be zero in the case we got down to 1 live rack only, while if fallback is not enabled, then we should report 1 additional replica to be required, and give a warning in the logs as the policy can not be met.
I think it would be more clear to handle that in the validation of rack aware policies if we can. What do you think?

Note that we are hardcoding fallback = true at the moment at the only place where we call the policy factory in non-test code.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
fapifta: Where we have a difference is two cases with replication factor of 3:
 - we have only 1 replica, we need two replica in one rack (delta = 2, misRepDelta = 1)
 - we have 2 replica in different racks, we need one replica and zero rack (delta = 1, misRepDelta = 0)

I see this as something that still implicitly brings in rack awareness into ReplicationManager via an arbitrary int value, that comes from a method which has a name, that - for me at least - does not suggest this subtle but fundamental difference at all, that is why I started to check into this and suggested to get rid of it. I don't see the "it might be good to report this for Recon" as a good reason to increase the complexity, if we want this in Recon, we can add a new status implementation with this value and its accessor added for rack aware policies and we can grab the value from that, if it will be needed.
For the same reason (implicit rack awareness) I am unhappy about the check in lines from L562, and suggested a way there if we can achieve it.

I understand the statement "a container has a valid placement", as "we have the replication factor amount of replicas placed as we want them to be placed". And my suggestion reflects this understanding. Isn't this the good way to look at it? If not, why, what is the gain if we look it differently?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
sodonnel: I have reworded HDDS-3079 and added a specific note:

> Note that as part of this move, the rack specific logic in SCMCommonPlacementPolicy.validateContainerPlacement() should be extracted and moved into this common parent too.

I really don't like the idea of duplicating more code into both classes and I think the above Jira now makes it clear what should happen when the rack specific logic moved.

I don't think it is a trivial task to extract the rack specific part of both placement policies either, so I think it is best left to that other change rather than doing it with this change. It will make this change too big.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: The nodemanager + topology can easily tell us the total racks every registered since SCM was started. But if you have a case where the cluster had 2 racks and than a rack has stopped, that is not so easy to see. You would need to filter all the non-healthy nodes, which is basically what the placement policy does. Then it falls back to give you a non-placement valid node if it needs to. It is important it does this, as it needs to handle under-replication too via the same chooseNodes methods.

The whole area around the fallback needs further thought in general. If we fall to 1 rack, on a cluster that had 2 racks, then its safe to assume the other rack should come back, and until it does the containers are mis-replicated (and probably under-replicated too) and should be reported as such.

In both cases a warning would be written to the logs stating:

```
LOG.warn("Container {} is mis-replicated, requiring {} additional " +
              "replicas. After selecting new nodes, mis-replication has not " +
              "improved. No additional replicas will be scheduled",
              id, misRepDelta);
```
This is really an edge case (2 rack cluster going into 1 rack) and I think we should leave this as it is for now as there are surely other things need to be thought about here in the wider network topology logic.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/SCMCommonPlacementPolicy.java
fapifta: 👍

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: > I understand the statement "a container has a valid placement", as "we have the replication factor amount of replicas placed as we want them to be placed"

Placement and replication are two different things, and I think it is good to keep them that way.

If you consider HDFS fsck - if a container is under replicated, it will report that. If it is mis-replicated it will report that. If it is both, it reports them both separately, along with their "missing count" for each. In HDFS for racks, there can only ever be 1 additional rack needed. But with upgrade domain, there can be possibly two additional domains needed.

> I see this as something that still implicitly brings in rack awareness into ReplicationManager via an arbitrary int value

It brings mis-replication into the RM, not rack awareness, and this is OK. A container needs to meet several goals to be properly replicated - the correct number of healthy replicas and the correct placement and this is what the replication manager checks.

In Recon, we want to be able to give an indicator about mis-replication and under replication separately in a similar way and ideally with Recon we just load the exact same placement policy as the cluster is using and call the same methods.

>  I am unhappy about the check in lines from L562, and suggested a way there if we can achieve it.

This check is still useful. If we combined the two rules as you are suggesting so "addtitionalReplicaRequired" return a single number representing delta and misRep, then in the case where you have delta=1, misRep=1 (ie two replicas on one rack), then if the newly selected node, due to fallback, gives us 3 replicas on one rack. This is OK, but the next time this container is checked, we have 3 replicas on one rack, and the logic will return 1, telling us to create another. We need to be sure this improves the mis-replication before creating a 4th replica, as that will make the container over-replicated.

Of course, we can check:

 * If we test 2 replicas with 1 rack, the logic will return 1.
 * If we test 3 replicas with 1 rack, the logic will return 1.
 * If we test 4 replicas with 1 rack, the logic will return 1.

So if we push the under and mis-replication numbers together, we cannot really see what is going on, and we will need to test the replication count too I think anyway.

I can give these methods new names or improve the java docs a bit more to make it clear that placement != replication and the number returned from the ContainerPlacementStatus object does not reflect replicas. I've made a small change to the doc now and will push it up:

```
  /**
   * Returns a boolean indicating if the container replicas meet the desired
   * placement policy. That is, they are placed on a sufficient number of
   * racks, or node groups etc. This does not check if the container is over
   * or under replicated, as it is possible for a container to have enough
   * replicas and still not meet the placement rules.
   * @return True if the containers meet the policy. False otherwise.
   */
  boolean isPolicySatisfied();

  /**
   * Returns an String describing why a container does not meet the placement
   * policy.
   * @return String indicating the reason the policy is not satisfied or null if
   *         the policy is satisfied.
   */
  String misReplicatedReason();

  /**
   * If the container does not meet the placement policy, return an integer
   * indicating how many additional replicas are required so the container
   * meets the placement policy. Otherwise return zero.
   * Note the count returned are the number of replicas needed to meet the
   * placement policy. The container may need additional replicas if it is
   * under replicated. The container could also have sufficient replicas but
   * require more to make it meet the policy, as the existing replicas are not
   * placed correctly.
   * @return The number of additional replicas required, or zero.
   */
  int additionalReplicaRequired();
```

I could change the name of additionalReplicaRequired() to be something like replicasRequiredForPlacement(), if yo think it makes it more clear.

fapifta: Ah ok, finally I understood :) Sorry for being a bit hard headed on this one, but it seems I haven't thought of the ambiguity in the mentioned case which is needed to be handled unless we will run into an infinite loop with the current fallback logic. I still feel that this we might revise when/if we touch fallback logic, but unsure :)

I would be glad if the name can be changed though, I don't have a better idea then the suggested one though... maybe misReplicationCount()?

Also in the javadoc, in the last sentence before @return do you mean "placed incorrectly"?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
fapifta: Ok, let's do so for now, but if we can, try to keep this in mind as a possible improvement, as the current check feels a bit hacky still :)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: Do you mean this line:

> The container could also have sufficient replicas but require more to make it meet the policy, as the existing replicas are **not** placed correctly.

Did you miss the "not" in there? :)

I will change this method to be called misReplicationCount() which hopefully makes it more clear.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
sodonnel: Yea, there are improvements needed in this area for sure. I'd like to see how it works in practice before we jump in with too many changes, but its certain on my radar for improvements.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
fapifta: oh... I did, it was hiding behind a scrollbar I haven't scrolled here in the comment stream :)

Looks good this way ;)

fapifta:
Thank you for the long discussion and addressing a lot of concerns and discuss through the nit details @sodonnel, I am +1 for this to go in, and I hope we can fix soon those things we discussed to be dealt with later.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
avijayanhwx: Yes, thanks for the explanation.

PR #884, opened 2020-04-29
HDDS-2406. Ozone shell key get throws IllegalArgumentException if pipeline is empty.

flirmnave:
If we get exist key but pipeline's datanode is empty, it throws IllegalArgumentException.
I add ResultCodes to Exception so that we can get more clearly message.
https://issues.apache.org/jira/browse/HDDS-2406
## How was this patch tested?
Run on docker
Screenshot from 2020-04-29 16-07-30


hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientManager.java
```diff
@@ -168,7 +169,8 @@ private XceiverClientSpi acquireClient(Pipeline pipeline, boolean read)
       throws IOException {
     Preconditions.checkNotNull(pipeline);
     Preconditions.checkArgument(pipeline.getNodes() != null);
-    Preconditions.checkArgument(!pipeline.getNodes().isEmpty());
+    Preconditions.checkArgument(!pipeline.getNodes().isEmpty(),
+        FAILED_TO_FIND_HEALTHY_NODES);
```
adoroszlai, line 173: I think `NO_REPLICA_FOUND` would better reflect the situation.  There may be other, live datanodes, and "healthy nodes" seem more related to placement of new data.

flirmnave: Thanks @adoroszlai for review, I change it.

PR #886, opened 2020-04-30
HDDS-3473. Ozone chunkinfo CLI should display block file path info

sadanand48:
The debug tool CLI was not considering both the file per block and file per chunk strategies while displaying the file path .This change fixes the issue and depending on the strategy , the path is shown in a field called files.

https://issues.apache.org/jira/browse/HDDS-3473

## How was this patch tested?
Manually tested.


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ChunkKeyHandler.java
```diff
@@ -128,17 +131,23 @@ protected void execute(OzoneClient client, OzoneAddress address)
...
         chunkDetails.setChunkOffset(chunkInfo.getOffset());
         chunkDetailsList.add(chunkDetails);
-        chunkPaths.add(getChunkLocationPath(containerData.getContainerPath())
-              + File.separator
-              + chunkInfo.getChunkName());
+        if (chunkLayOutVersion == ChunkLayOutVersion.FILE_PER_CHUNK) {
+          chunkPaths.add(getChunkLocationPath(containerData.getContainerPath())
+                  + File.separator
+                  + chunkInfo.getChunkName());
+        } else {
```
mukul1987, line 138: lets convert this if into a switch case, in case new CHunkLayoutVersions are added later, there should be an option of adding it later. Also for default case lets throw an exception.

hadoop-ozone/dist/src/main/smoketest/basic/ozone-shell.robot
```diff
@@ -121,7 +121,7 @@ Test key handling
     ${result} =     Execute             ozone sh key info ${protocol}${server}/${volume}/bb1/key1 | jq -r '. | select(.name=="key1")'
                     Should contain      ${result}       creationTime
     ${result} =     Execute             ozone debug chunkinfo ${protocol}${server}/${volume}/bb1/key1 | jq -r '.[]'
-                    Should contain      ${result}       chunks
+                    Should contain      ${result}       files
```
mukul1987, line 124: Can this command be executed on the datanode container and lets do a check that the file exists.

sadanand48: I have included a separate robot test that checks whether the file is present in the datanode.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ChunkKeyHandler.java
```diff
@@ -128,17 +132,31 @@ protected void execute(OzoneClient client, OzoneAddress address)
...
+        case FILE_PER_BLOCK:
+          chunkPaths.add(getChunkLocationPath(containerData
+                  .getContainerPath())
+                  + File.separator
+                  + keyLocation.getLocalID() + ".block");
+          break;
+        default:
+          throw new StorageContainerException("chunk strategy does not exist",
+                  ContainerProtos.Result.UNABLE_TO_FIND_CHUNK);
+        }
```
bshashikant, line 151: I think we should move the code to give the chunk file path given the containerID/ContainerData into some common utilities class and use it across client/server so that any further changes won't break it..

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ChunkFileUtility.java
```diff
@@ -0,0 +1,39 @@
...
+                  + chunkInfo.getChunkName();
+        case FILE_PER_BLOCK:
+          return  getChunkLocationPath(data
+                  .getContainerPath())
+                  + File.separator
+                  + keyLocation.getLocalID() + ".block";
+        default:
+          throw new StorageContainerException("chunk strategy does not exist",
+                  ContainerProtos.Result.UNABLE_TO_FIND_CHUNK);
+        }
```
adoroszlai, line 37: I'd prefer taking chunk file path directly from `ChunkLayOutVersion.getChunkFile()` instead of the `switch`.  This way if a new enum constant is introduced, this code will work without any changes.

This needs a minor refactoring in `ChunkLayOutVersion`, which accepts a `ContainerData` object, to work directly with the chunk dir:

```diff
--- hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ChunkLayOutVersion.java
+++ hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ChunkLayOutVersion.java
@@ -40,17 +40,15 @@

   FILE_PER_CHUNK(1, "One file per chunk") {
     @Override
-    public File getChunkFile(ContainerData containerData, BlockID blockID,
-        ChunkInfo info) throws StorageContainerException {
-      File chunksLoc = verifyChunkDirExists(containerData);
-      return chunksLoc.toPath().resolve(info.getChunkName()).toFile();
+    public File getChunkFile(File chunkDir, BlockID blockID,
+        ChunkInfo info) {
+      return chunkDir.toPath().resolve(info.getChunkName()).toFile();
     }
   },
   FILE_PER_BLOCK(2, "One file per block") {
     @Override
-    public File getChunkFile(ContainerData containerData, BlockID blockID,
-        ChunkInfo info) throws StorageContainerException {
-      File chunkDir = verifyChunkDirExists(containerData);
+    public File getChunkFile(File chunkDir, BlockID blockID,
+        ChunkInfo info) {
       return new File(chunkDir, blockID.getLocalID() + ".block");
     }
   };
@@ -118,8 +116,14 @@ public String getDescription() {
     return description;
   }

-  public abstract File getChunkFile(ContainerData containerData,
-      BlockID blockID, ChunkInfo info) throws StorageContainerException;
+  public abstract File getChunkFile(File chunksDir,
+      BlockID blockID, ChunkInfo info);
+
+  public File getChunkFile(ContainerData containerData,
+      BlockID blockID, ChunkInfo info) throws StorageContainerException {
+    File chunksLoc = verifyChunkDirExists(containerData);
+    return getChunkFile(chunksLoc, blockID, info);
+  }

   @Override
   public String toString() {
```

sadanand48: Thankyou @adoroszlai for the review. Addressed review comments .

hadoop-ozone/dist/src/main/smoketest/debug/ozone-debug.robot
```diff
@@ -0,0 +1,37 @@
...
+Test Timeout        2 minute
+
+*** Variables ***
+
+
+*** Test Cases ***
+Create Volume,Bucket and put key
+   Execute             ozone sh volume create o3://om/vol1 --quota 100TB
+   Execute             ozone sh bucket create o3://om/vol1/bucket1
+   Execute             ozone sh key put o3://om/vol1/bucket1/debugKey /opt/hadoop/NOTICE.txt
```
adoroszlai, line 29: I think this should be a `Suite Setup` instead of a `Test Case`, but it's OK to improve in a followup Jira.

hadoop-ozone/dist/src/main/smoketest/debug/ozone-debug.robot
```diff
@@ -0,0 +1,37 @@
...
+
+Test ozone debug
+    ${result} =     Execute             ozone debug chunkinfo o3://om/vol1/bucket1/debugKey | jq -r '.[]'
+                    Should contain      ${result}       files
+    ${result} =     Execute             ozone debug chunkinfo o3://om/vol1/bucket1/debugKey | jq -r '.[].files[0]'
+    ${result3} =    Execute             echo "exists"
+    ${result2} =    Execute             test -f ${result} && echo "exists"
+                    Should Be Equal     ${result2}       ${result3}
```
adoroszlai, line 37: `Execute` checks the return code of the command, so I think this should be enough to verify that the file exists:

```
                    Execute             test -f ${result}
```

Better yet, there is a Robot keyword for the same: [`File Should Exist`](http://robotframework.org/robotframework/3.0.2/libraries/OperatingSystem.html#File%20Should%20Exist)

```suggestion
                    File Should Exist             ${result}
```

PR #887, opened 2020-04-30
HDDS-3492. With hdfs-fuse file implementation 'cp' command is not working - fixed.

aryangupta1998:

'cp' (copy) command doesn’t work with hdfs-fuse file system. It works by passing 'cp' command two times for the same file. It mainly happens when destination file doesn’t exists. It happens because when 'cp' command is performed for the first time it creates the file at the destination path, but as per ozone file system protocols any file has no access till its committed. So, 'getattr' function is not able to locate the file(though it’s present at the destination path). Hence, 'cp' command fails in the first go. Second time when 'cp' command is executed, file is already present in the destination path so 'cp' command simply override the data in the file.
This change introduces a new class 'PosixOzoneFileSystem' which creates the file, closes it and again opens it to make the 'cp' command work perfectly.

https://issues.apache.org/jira/browse/HDDS-3492

## How was this patch tested?

Tested manually

### DIFF
--- a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/PosixOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/PosixOzoneFileSystem.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.fs.ozone;
+
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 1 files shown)


elek:
Thanks to create this patch @aryangupta1998. Don't see any problem with this approach, but neither understand it fully. Is `getattr` supposed to be working according to POSIX before closing the file?

Do we need to fix it in our side? (Checking the open key tables when we do getFileStatus)

aryangupta1998:
> Thanks to create this patch @aryangupta1998. Don't see any problem with this approach, but neither understand it fully. Is `getattr` supposed to be working according to POSIX before closing the file?
> 
> Do we need to fix it in our side? (Checking the open key tables when we do getFileStatus)

@elek Yes 'getattr' is working according to POSIX before closing the file, but first time 'getattr' works before closing of file is to check whether the file exists at destination path or not. But second time 'getattr' doesn’t works before closing of file is when system creates empty file at destination path and opens it to write something but according to ozone filesystem, file is not visible till it’s committed so 'getattr' function fails this time before even closing of file. To address this issue we have raised this PR.

nandakumar131:
Thanks @aryangupta1998  for the PR and @elek  for review. I will merge this shorty.

elek:
Thanks the answer @aryangupta1998, we also discussed it offline, and I try to summarize it:

 * We have more serious problems, as this is just a workaround for NFS
 * Even this workaround doesn't solve all the problems. (It creates a 0 size file first, and add content later. This behavior may or may not be accepted by any other tools).
 * One example: when some configuration is copied to this file system, other service can found it as a zero size file and start with wrong parameters as real content is added later
 * HBase/Impala/... and other projects have the same problems with Ozone (Should be fixed it in our side) AND with other Object Stores (s3a)
 * Related Impala bug for exmaple: https://issues.apache.org/jira/browse/IMPALA-9437
 * But fixing is not easy as there could be multiple parallel file creation (no lease).


TLDR; This seems to be a dangerous workaround. I don't think that it's good solution for production-ready Fuse file system. But it can be a workaround to make possible the further testing of hdfs-fuse with Ozone. As hdfs-fuse is in alpha stage, I am fine to commit it.

But we need to continue to search for a generic, long-term fix (IMHO)

nandakumar131:
Thanks @elek  for the summary. Created HDDS-3569.

PR #888, opened 2020-04-30
HDDS-3502. Remove dependence on commons-lang

adoroszlai:

Few parts of Ozone still use `commons-lang`, while most are already on `commons-lang3`.  This change proposes to update those remaining usages and remove the unnecessary dependency.

https://issues.apache.org/jira/browse/HDDS-3502

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/633284638


adoroszlai:
> Should we also update the License Notice - https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/dist/src/main/license/bin/NOTICE.txt#L269 ?

Thanks for asking, I was not aware of this.

I don't think we need to update the notice.  We still use this library, but only v3 instead of both v2 and v3.

vivekratnavel:
@adoroszlai Thanks for working on this!

adoroszlai:
Thanks @vivekratnavel for reviewing and merging it.

PR #892, opened 2020-05-02
HDDS-3449. Enable TestSCMPipelineMetrics test cases

cku328:

Fixed and enabled this test case.

After the HDDS-2757 patch, the operation to close pipeline will run after timeout, so the results will not immediately respond to the metrics.
In the test method `TestSCMPipelineMetrics#testPipelineDestroy()` , I make the operation to close pipeline run immediately.

https://issues.apache.org/jira/browse/HDDS-3449

## How was this patch tested?

Ran UTs - Pass

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/pipeline/TestSCMPipelineMetrics.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/pipeline/TestSCMPipelineMetrics.java
@@ -45,7 +45,6 @@ import static org.apache.hadoop.test.MetricsAsserts.getMetrics;
 /**
  * Test cases to verify the metrics exposed by SCMPipelineManager.
  */
-@Ignore
 public class TestSCMPipelineMetrics {
 
   private MiniOzoneCluster cluster;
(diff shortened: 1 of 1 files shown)


avijayanhwx:
@cku328 Does this problem not exist for the testNumBlocksAllocated method? There is a 'closePipeline' call followed by metrics check there as well.

cku328:
Thanks @avijayanhwx for the review and comments.
I traced the code and checked testNumBlocksAllocated method doesn't have this problem.

The operation to close pipeline calls to [this method](https://github.com/apache/hadoop-ozone/blob/ade09f145f42b12b3b05f75580c1965bad3fe42a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java#L395-L410), first calling [finalizePipeline(...)](https://github.com/apache/hadoop-ozone/blob/ade09f145f42b12b3b05f75580c1965bad3fe42a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java#L537) to remove the pipeline metrics immediately, so it is not possible to get the metric value _(numBlocksAllocated)_ by pipeline.

After that, it will call the [destroyPipeline(...)](https://github.com/apache/hadoop-ozone/blob/ade09f145f42b12b3b05f75580c1965bad3fe42a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java#L558) method, which affects the metric of `testPipelineDestroy()` _(increments number of NumPipelineDestroyed)_, but whether it executes immediately depends on the 'onTimeout' flag.

avijayanhwx:
Thank you for the explanation @cku328.

cku328:
Thanks @avijayanhwx for merging it.

PR #893, opened 2020-05-04
HDDS-3517. Add a directory based Ozone Manager LoadGenerator.

mukul1987:

Most of the LoadGenerators in MiniOzoneLoadGenerator are all based on the read/write path. The purpose of this LoadGenerator is to add a directory-based load generator which stresses the IN.

https://issues.apache.org/jira/browse/HDDS-3517

## How was this patch tested?
Ran MiniOzoneChaosTests, also already found https://issues.apache.org/jira/browse/HDDS-3382, https://issues.apache.org/jira/browse/HDDS-3381 because of this load generator

### DIFF
--- a/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/loadgenerators/NestedDirLoadGenerator.java
+++ b/hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/loadgenerators/NestedDirLoadGenerator.java
@@ -1,0 +1,58 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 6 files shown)


hadoop-ozone/fault-injection-test/mini-chaos-tests/src/test/java/org/apache/hadoop/ozone/loadgenerators/NestedDirLoadGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A Load generator where nested directories are created and read them.
+ */
+public class NestedDirLoadGenerator extends LoadGenerator {
+  private final LoadBucket fsBucket;
+  private final int maxDirWidth;
```
avijayanhwx, line 32: nit. Shouldn't this be 'maxDirDepth'?

PR #894, opened 2020-05-04
HDDS-3501. OzoneFileStatus should not extend FileStatus

elek:

FileStatus is a Hadoop specific class. The return type of getFileStatus OM call should be Hadoop independent and a simple POJO can be used.

OzoneFileSystem can create the appropriate FileStatus implementation based on the information in this simple POJO.

image

https://issues.apache.org/jira/browse/HDDS-3501

## How was this patch tested?

With full CI build on my fork.


hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -454,19 +442,23 @@ public BasicKeyInfo next() {
...
         status.getBlockSize(),
-        status.getModificationTime(),
-        status.getAccessTime(),
-        status.getPermission().toShort(),
-        status.getOwner(),
-        status.getGroup(),
-        status.getPath(),
+        keyInfo.getModificationTime(),
+        keyInfo.getModificationTime(),
+        status.isDirectory() ? (short) 00777 : (short) 00666,
```
xiaoyuyao, line 458: How do we plan to map the dir/dir permission into Ozone ACLs? It is currently hard coded to 777 for dir and 666 for file.

elek:
> How do we plan to map the dir/dir permission into Ozone ACLs? It is currently hard coded to 777 for dir and 666 for file.

Yes, I just realized that we don't support it when I moved this code part from one place to other.

I am not sure what should we do based on POSIX. AFAIK ACL and directory permissions are different. Or do you know how does it work with HDFS?

xiaoyuyao:
> > How do we plan to map the dir/dir permission into Ozone ACLs? It is currently hard coded to 777 for dir and 666 for file.
> 
> Yes, I just realized that we don't support it when I moved this code part from one place to other.
> 
> I am not sure what should we do based on POSIX. AFAIK ACL and directory permissions are different. Or do you know how does it work with HDFS?

I think the current implementation matches with how s3 maps to HCFS ACLs. Since this is not new code introduced in this JIRA, we can follow up on it with separate JiRA.

xiaoyuyao:
LGTM, +1. Thanks @elek for the contribution.

PR #895, opened 2020-05-04
HDDS-3518: Add a freon generator to create directory tree and create …

rakeshadr:

This Jira proposes to add a functionality to freon tool to create directories in tree structure. 

**Proposed structure:**
  /*
      Nested directories will be created like this,
      suppose you pass depth=3, span=3 and number of tests=2

      Directory Structure:-
                            |-- Dir111
                            |
                |-- Dir11 --|-- Dir112
                |           |
                |           |-- Dir113
                |
                |
                |           |-- Dir121
                |           |
       Dir1   --|-- Dir12 --|-- Dir122
                |           |
                |           |-- Dir123
                |
                |
                |           |-- Dir131
                |           |
                |-- Dir13 --|-- Dir132
                            |
                            |-- Dir133

     In each directory 'c' number of files to be written in each directory with file size 'g' will be created.
   */

This is basically useful to determine the performance evaluation of ozone, when creating defined number of nested directories and also create given number of files into each directory.

https://issues.apache.org/jira/browse/HDDS-3518

## How was this patch tested?

Added unit test case..

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/freon/TestHadoopDirTreeGenerator.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/freon/TestHadoopDirTreeGenerator.java
@@ -109,7 +109,7 @@ public class TestHadoopDirTreeGenerator {
   }
 
   private void verifyDirTree(String volumeName, String bucketName, int depth,
-                             int span, int fileCount, int perFileSizeInKBs)
+                             int span, int fileCount, int perFileSizeInBytes)
           throws IOException {
 
     store.createVolume(volumeName);
@@ -120,7 +120,7 @@ public class TestHadoopDirTreeGenerator {
     new Freon().execute(
         new String[]{"-conf", confPath, "dtsg", "-d", depth + "", "-c",
             fileCount + "", "-s", span + "", "-n", "1", "-r", rootPath,
-                     "-g", perFileSizeInKBs + ""});
+                     "-g", perFileSizeInBytes + ""});
     // verify the directory structure
     FileSystem fileSystem = FileSystem.get(URI.create(rootPath),
             conf);
@@ -132,7 +132,7 @@ public class TestHadoopDirTreeGenerator {
       // as it has only one dir at root.
       verifyActualSpan(1, fileStatuses);
       int actualDepth = traverseToLeaf(fileSystem, fileStatus.getPath(),
-              1, depth, span, fileCount, perFileSizeInKBs);
+              1, depth, span, fileCount, perFileSizeInBytes);
       Assert.assertEquals("Mismatch depth in a path",
               depth, actualDepth);
     }
@@ -140,7 +140,7 @@ public class TestHadoopDirTreeGenerator {
 
   private int traverseToLeaf(FileSystem fs, Path dirPath, int depth,
                              int expectedDepth, int expectedSpanCnt,
-                             int expectedFileCnt, int perFileSizeInKBs)
+                             int expectedFileCnt, int perFileSizeInBytes)
           throws IOException {
     FileStatus[] fileStatuses = fs.listStatus(dirPath);
     // check the num of peer directories except root and leaf as both
@@ -153,10 +153,10 @@ public class TestHadoopDirTreeGenerator {
       if (fileStatus.isDirectory()) {
         ++depth;
         return traverseToLeaf(fs, fileStatus.getPath(), depth, expectedDepth,
-                expectedSpanCnt, expectedFileCnt, perFileSizeInKBs);
+                expectedSpanCnt, expectedFileCnt, perFileSizeInBytes);
       } else {
         Assert.assertEquals("Mismatches file len",
-                perFileSizeInKBs * 1024, fileStatus.getLen());
+                perFileSizeInBytes, fileStatus.getLen());
         actualNumFiles++;
       }
     }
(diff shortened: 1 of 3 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/freon/TestHadoopDirTreeGenerator.java
```diff
@@ -0,0 +1,182 @@
...
+    cluster.waitForClusterToBeReady();
+    cluster.waitTobeOutOfSafeMode();
+    return cluster;
+  }
+
+  @Test
+  public void testNestedDirTreeGeneration() throws Exception {
+    OzoneConfiguration conf = new OzoneConfiguration();
+    MiniOzoneCluster cluster = startCluster(conf);
+    ObjectStore store =
```
xiaoyuyao, line 92: Can we wrap this with try {} finally {shutdown (cluster)} so that the cluster is ensured to shutdown if any verification failed?

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/HadoopDirTreeGenerator.java
```diff
@@ -0,0 +1,218 @@
...
+    mixinStandardHelpOptions = true,
+    showDefaultValues = true)
+public class HadoopDirTreeGenerator extends BaseFreonGenerator
+    implements Callable<Void> {
+
+  private static final Logger LOG =
+      LoggerFactory.getLogger(HadoopDirTreeGenerator.class);
+
+  @Option(names = {"-r", "--rpath"},
+      description = "Hadoop FS directory system path",
```
xiaoyuyao, line 53: "Hadoop FS directory system path" is a bit confusing, do you mean "Hadoop FS root path"?

rakeshadr: Thanks @xiaoyuyao for the reviews.

Yes, its "Hadoop FS root path". I will modify to make it clear.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/freon/TestHadoopDirTreeGenerator.java
rakeshadr: Sure, I will update this in next PR commit.

rakeshadr: Done!

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/HadoopDirTreeGenerator.java
rakeshadr: Done!

rakeshadr:
Thanks a lot @elek for the comments with detailed explanation.

> 1. I guess you would like to test the correctness and not interested about the performance. But I love the numbers and I think it's always good to have.

Sure, I will add the timer and helpful during perf evaluation.

> 1. There is a `prefix` field in the `BaseFreonGenerator`. This is a random string, which can be specified. 
> (One easy way to use it to generate dirs in `rpath` + `/` + `prefix`)

This tool provides '--nameLen' option to the users to define the length of the random name of the directory.  Another point is, `/` + `prefix` will add additional depth(sub-dir), which HadoopDirTreeGenerator#makeDirWithGivenNumberOfFiles() is already doing it. 

I am not finding any usage of 'prefix' in this tool and skip this comment. Does that make sense to you? Thanks!


> 1. We have a `ContentGenerator` class which can do the the random key generation. I think it can be used here... If you see some limitation of it, we can improve it as well...
Yes, will use ContentGenerator for writing the content.

elek:
Thanks the answers @rakeshadr 

> I am not finding any usage of 'prefix' in this tool and skip this comment. Does that make sense to you? 

Yes, prefix is useful only if we implement the READ/validator side (a separated Freon test). I am fine with ignoring it now and start to use it only when we really need it...

Let's add the metrics + ContentGenerator and I will merge it.

rakeshadr:
Thanks @elek , latest commit addressing your comments.

elek:
Thanks @rakeshadr the update.

Unfortunately when I checked your update 3 more opinions came in to my mind (Sorry). Let me share them here:

1.  You introduced an atomic counter `totalDirsCnt`. But we already have multiple counters `successCounter`, `failureCounter`, `failureCounter` in the base class. The only problem is that his patch doesn't follow the existing approach where the overall test can be splitted to small tasks (which can be measured one by one). Personally I think it's possible even in this use case but the generation should be refactored.

2. For example you used some `System.out.println` in this patch. We had some problems with different printing earlier (sometimes freon is executed from shell sometimes from command line) `BaseFreonGenerator.printReport` already handle this, but this functionality is not used.

3. We already have a `HadoopNestedDirGenerator`, half of the functionality of this patch is already there. For me it seems to be better to combine the two tests.

But I have two additional notes:

 1. I might be wrong with any point, I can be convinced.

 2. Usually I am more interested about the agreement in long-term directions, I think we can merge this patch. It works well and (if we agree) we can further improve it (maybe merge with the `HadoopNestedDirGenerator`, maybe use the same structure as the other Freon tests.) --> **Let me merge it now, and let's continue the discussion**

elek:
Merged. Thank you the contribution @rakeshadr (and thanks for the patience for my long comments ;-) )

PR #899, opened 2020-05-06
HDDS-3549. TestKeyInputStream#testSeek fails intermittently.

lokeshj1703:

1. The test TestKeyInputStream#testSeek compares incorrect metric. This metric compared pendingOps.
`    Assert.assertEquals(readChunkCount, metrics
        .getContainerOpsMetrics(ContainerProtos.Type.ReadChunk));
`
The PR now compares the correct metric i.e. getContainerOpCountMetrics.
2. Changed name for metric value getter to make it more distinct. From getContainerOpsMetrics to getPendingContainerOpCountMetrics.

https://issues.apache.org/jira/browse/HDDS-3549

## How was this patch tested?

The patch fixes a unit test.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestKeyInputStream.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestKeyInputStream.java
@@ -257,7 +257,7 @@ public class TestKeyInputStream {
 
     // Seek operation should not result in any readChunk operation.
     Assert.assertEquals(readChunkCount, metrics
-        .getContainerOpsMetrics(ContainerProtos.Type.ReadChunk));
+        .getContainerOpCountMetrics(ContainerProtos.Type.ReadChunk));
 
     byte[] readData = new byte[chunkSize];
     keyInputStream.read(readData, 0, chunkSize);
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientMetrics.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientMetrics.java
@@ -96,7 +96,7 @@ public class XceiverClientMetrics {
     containerOpsLatency[type.ordinal()].add(latencyNanos);
   }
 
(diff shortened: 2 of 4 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestKeyInputStream.java
```diff
@@ -257,7 +257,7 @@ public void testSeek() throws Exception {
 
     // Seek operation should not result in any readChunk operation.
     Assert.assertEquals(readChunkCount, metrics
-        .getContainerOpsMetrics(ContainerProtos.Type.ReadChunk));
+        .getContainerOpCountMetrics(ContainerProtos.Type.ReadChunk));
```
adoroszlai, line 260: So this was comparing pending op count, and now it will compare completed op count.

But I don't think this would fix the test failure:

```
testSeek(org.apache.hadoop.ozone.client.rpc.TestKeyInputStream)  Time elapsed: 0.103 s  <<< FAILURE!
java.lang.AssertionError: expected:<0> but was:<-1>
  ...
  at org.apache.hadoop.ozone.client.rpc.TestKeyInputStream.testSeek(TestKeyInputStream.java:259)
```

Neither pending nor completed count should ever be -1.

lokeshj1703: I agree it should not be -1. But if we look at completedCount it doesn't have a decrement method whereas pendingOpCount does have it. So completedCount will never decrement from 0 but pendingOpCount can. There is a problem with how we are decrementing the pending metric (we are decrementing when a reply comes and in onError call) and that should be tracked via a separate jira. There should be a better mechanism to decrement pendingOpCount.

adoroszlai: You are right, I forgot that the other metric is not decremented, so it will not have this problem.

xiaoyuyao:
+1 to commit this. Do we have a separate follow up JIRA?

lokeshj1703:
Filed https://issues.apache.org/jira/browse/HDDS-3578 to fix the other issue.

PR #900, opened 2020-05-06
HDDS-3500. Hide OMFailoverProxyProvider usage behind an interface

elek:

`OzoneManagerProtocolClientSideTranslatorPB` uses `OmFailoverProxyProvider` to access OM HA, but this class is not supported in Hadoop 2.x environment.

It would be better to
1. separated `ProtocolClientSideTranslator` from the transport layer logic
2. Remove implementation specific method from the `OzoneManagerProtocol` (`getOMFailoverProxyProvider` should be removed)
3. Use a simple `OMTransport` interface to handle all the connection logic in one, isolated place

https://issues.apache.org/jira/browse/HDDS-3500

## How was this patch tested?

With my own form


elek:
@bharatviswa504 Can you please help me the review?

bharatviswa504:
@elek I will take a look at it today.
I have not completed my review, looked at high level.

@elek I will take a look at it today.
I have not completed my review, looked at a high level.

bharatviswa504:
> @bharatviswa504 Can you please help me the review?

@elek I will take a look at it today.
I have not completed my review, looked at a high level.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/Hadoop3OmTransport.java
```diff
@@ -0,0 +1,269 @@
...
+import org.apache.hadoop.security.token.SecretManager;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.protobuf.RpcController;
+import com.google.protobuf.ServiceException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Full-featured Hadoop RPC + HA implementation.
```
bharatviswa504, line 49: This class will be used for non-HA also right?
If only one node is configured, also this will work, right? So, do we want to update java doc here to reflect correctly

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/Hadoop3OmTransportFactory.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.apache.hadoop.ozone.om.protocolPB;
+
+import java.io.IOException;
+
+import org.apache.hadoop.hdds.conf.ConfigurationSource;
+import org.apache.hadoop.security.UserGroupInformation;
+
+/**
+ * Factory to crate the default Hadoop 3 + HA transport.
```
bharatviswa504, line 26: NIT: crate -> Create
Same as above this will be used irrespective of HA or not HA.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OmTransport.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.hadoop.io.Text;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMRequest;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMResponse;
+
+/**
+ * Transport responsible to send messages to OM.
+ */
+public interface OmTransport {
+
+  OMResponse submitRequest(OMRequest payload) throws IOException;
```
bharatviswa504, line 31: Can we add some java docs to these methods?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/OmFailoverProxyUtil.java
```diff
@@ -0,0 +1,50 @@
...
+      RpcClient clientProxy) {
+    OzoneManagerProtocolClientSideTranslatorPB ozoneManagerClient =
+        (OzoneManagerProtocolClientSideTranslatorPB) clientProxy
+            .getOzoneManagerClient();
+    Hadoop3OmTransport transport =
+        (Hadoop3OmTransport) ozoneManagerClient.getTransport();
+    return transport.getOmFailoverProxyProvider();
+  }
+
+  public static OMFailoverProxyProvider getFailoverProxyProvider(
```
bharatviswa504, line 46: One would have been sufficient right here, as anyway RpcClient extends ClientProtocol.

So, we can have code as below right?

  public static OMFailoverProxyProvider getFailoverProxyProvider(
      ClientProtocol clientProtocol)) {
    OzoneManagerProtocolClientSideTranslatorPB ozoneManagerClient =
        (OzoneManagerProtocolClientSideTranslatorPB) clientProxy
            .getOzoneManagerClient();
    Hadoop3OmTransport transport =
        (Hadoop3OmTransport) ozoneManagerClient.getTransport();
    return transport.getOmFailoverProxyProvider();
  }

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/snapshot/TestOzoneManagerSnapshotProvider.java
```diff
@@ -105,17 +106,15 @@ public void testDownloadCheckpoint() throws Exception {
     retVolumeinfo.createBucket(bucketName);
     OzoneBucket ozoneBucket = retVolumeinfo.getBucket(bucketName);
 
-    String leaderOMNodeId = objectStore.getClientProxy().getOMProxyProvider()
-        .getCurrentProxyOMNodeId();
-    OzoneManager ozoneManager = cluster.getOzoneManager(leaderOMNodeId);
+    OzoneManager ozoneManager = cluster.getOMLeader();
```
bharatviswa504, line 109: The reason for this is 

Because getOMLeader has chance to return null.
```

  public OzoneManager getOMLeader() {
    OzoneManager res = null;
    for (OzoneManager ozoneManager : this.ozoneManagers) {
      if (ozoneManager.isLeader()) {
        if (res != null) {
          // Found more than one leader
          // Return null, expect the caller to retry in a while
          return null;
        }
        // Found a leader
        res = ozoneManager;
      }
    }
    return res;
  }
```

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/protocol/ClientProtocol.java
```diff
@@ -481,9 +484,6 @@ void cancelDelegationToken(Token<OzoneTokenIdentifier> token)
    */
   S3SecretValue getS3Secret(String kerberosID) throws IOException;
 
-  @VisibleForTesting
```
bharatviswa504, line 484: I have a question, we are removing this public utility method from client protocol, this will break applications if they use this method right, after the upgrade. Or end users(clients) should not use the RpcClient class directly only they should access through FileSystem and S3 Interfaces is the contract?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/Hadoop3OmTransport.java
elek: Yes, I mean RPC + *failover*  not *HA* exactly. Fixed, thanks.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OmTransport.java
elek: :heavy_check_mark:

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -216,23 +214,6 @@ public static void setScmId(String scmId){
...
 
-  /**
-   * Test OM Proxy Provider.
-   */
-  @Test
-  public void testOMClientProxyProvider() {
-    OMFailoverProxyProvider omFailoverProxyProvider = store.getClientProxy()
-        .getOMProxyProvider();
-    List<OMProxyInfo> omProxies = omFailoverProxyProvider.getOMProxyInfos();
-
```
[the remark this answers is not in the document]
elek: restored

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/protocol/ClientProtocol.java
elek: End users shouldn't use any public method which are annotated with `@VisibleForTesting` or `@Private`. And they should be careful if `@Stable` is missing from the class or methods. So I think it's not a problem here.

But I agree. And this is the reason why I think this method (independent from the patch) should be added only for the implementation not for the interface. If it's `VisibleForTesting`, we can do an easy cast during the unit test.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/snapshot/TestOzoneManagerSnapshotProvider.java
elek: Thanks. Fixed, together with `hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMRatisSnapshots.java`

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OmTransportFactory.java
```diff
@@ -0,0 +1,63 @@
...
+      UserGroupInformation ugi, String omServiceId) throws IOException {
+    OmTransportFactory factory = createFactory();
+
+    return factory.createOmTransport(conf, ugi, omServiceId);
+  }
+
+  static OmTransportFactory createFactory() throws IOException {
+    ServiceLoader<OmTransportFactory> transportFactoryServiceLoader =
+        ServiceLoader.load(OmTransportFactory.class);
+    Iterator<OmTransportFactory> iterator =
```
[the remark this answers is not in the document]
elek: Good question. This is because we need a default and an option to choose a **different** implementation. 

If we create a service file for the default transport factory (`Hadoop3OmTransportFactory`) than it would be hard to override it. In case of the Hadoop2 client we would have two different implementation defined without an easy way to choose between them.

With this approach the default is always defined by line 50:61 but you have the option to define your own version.

elek:
Thanks the review and the suggestions @bharatviswa504, I addressed all the comments and pushed the improved version.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OmTransportFactory.java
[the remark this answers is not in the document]
bharatviswa504: Question: what is alternative way other then having meta-inf/services to load class for OmTransportFactory?

bharatviswa504:
I have one question, other than that it LGTM.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OmTransportFactory.java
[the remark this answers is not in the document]
elek: Currently there is no alternative way. I couldn't see any usage for any other approach, but if you see any use case, I am open to implement it. Let's create a new issue if you see some interesting direction...

elek:
Thanks the review @bharatviswa504 and @adoroszlai 

I am pushing it now. 

@bharatviswa504 If you think we need to improve the pluggability, let's continue this work and create new issues...

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OmTransportFactory.java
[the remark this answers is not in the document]
bharatviswa504: Then I am confused, what is the purpose of the code?
Because I think meta-inf/services is packaged along with jar right?

bharatviswa504:
I have a question, and posted my comments to the discussion.

elek:
@bharatviswa504 Please check #992 about the usage of META-INF/service. We can further improve in there, if you have any specific suggestion.

PR #906, opened 2020-05-08
HDDS-3494. Implement ofs://: Support volume and bucket deletion

smengcl:

Support volume and bucket deletion directly via ofs://

- Support empty volume and bucket deletion.
- Support recursive volume and bucket deletion. (see examples below)

https://issues.apache.org/jira/browse/HDDS-3494

## How was this patch tested?

Tested via contract tests.
Added new integration test case.

## Examples

```bash

```

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -25,6 +25,7 @@ import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathIsNotEmptyDirectoryException;
 import org.apache.hadoop.fs.contract.ContractTestUtils;
 import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
@@ -65,6 +66,7 @@ import static org.apache.hadoop.ozone.OzoneAcl.AclScope.ACCESS;
 import static org.apache.hadoop.ozone.OzoneConsts.OZONE_URI_DELIMITER;
 import static org.apache.hadoop.ozone.om.OMConfigKeys.OZONE_OM_ADDRESS_KEY;
 import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.BUCKET_NOT_FOUND;
+import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.VOLUME_NOT_FOUND;
 
 /**
  * Ozone file system tests that are not covered by contract tests.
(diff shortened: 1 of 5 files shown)


smengcl:
Not sure why the checks on the Apache repo are Queued for so long.
Tests on my fork passed: https://github.com/smengcl/hadoop-ozone/commit/9716c3b18343c07247951d4f787198ba136a5b91

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -469,15 +478,76 @@ public boolean delete(Path f, boolean recursive) throws IOException {
...
+      // Handle rm root
+      if (ofsPath.isRoot()) {
+        Iterator<? extends OzoneVolume> it =
+            adapterImpl.getObjectStore().listVolumes("");
+        while (it.hasNext()) {
+          OzoneVolume volume = it.next();
+          String nextVolume = addTrailingSlashIfNeeded(
+              f.toString()) + volume.getName();
+          delete(new Path(nextVolume), true);
+        }
```
smengcl, line 492: Replace the rm root logic with a warning. Effectively disallowing admin to use `rm -r /` to clear the cluster.

Admins should re-init the cluster if they really want to clear the cluster.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -469,15 +478,72 @@ public boolean delete(Path f, boolean recursive) throws IOException {
...
+
     } else {
       LOG.debug("delete: Path is a file: {}", f);
       result = adapter.deleteObject(key);
     }
 
     if (result) {
       // If this delete operation removes all files/directories from the
-      // parent direcotry, then an empty parent directory must be created.
+      // parent directory, then an empty parent directory must be created.
```
umamaheswararao, line 546: Why is this? Do you mind explaining why do we need to create parent dir. Parent dir should not have deleted right? I know its not related to this change, but just for my understanding.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -766,4 +768,103 @@ public void testTempMount() throws Exception {
...
+  private void deleteNonRecursivelyAndFail(Path f) throws IOException {
+    try {
+      fs.delete(f, false);
+      Assert.fail("Should have thrown PathIsNotEmptyDirectoryException!");
+    } catch (PathIsNotEmptyDirectoryException ignored) {
+    }
+  }
+
+  @Test
+  public void testDeleteVolumeAndBucket() throws IOException {
```
umamaheswararao, line 803: I think these test cases should be split into multiple cases?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -457,6 +462,10 @@ public boolean delete(Path f, boolean recursive) throws IOException {
       return false;
     }
 
+    if (status == null) {
+      return false;
+    }
+
     String key = pathToKey(f);
     boolean result;
```
umamaheswararao, line 471: Can we add javadoc to delete method here? parent abstract method delete does not provide any javadoc. SO, its good to provide here and explain the behavior?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
smengcl: Good idea

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
smengcl: It complies to the `FileSystem#delete` behavior: https://github.com/apache/hadoop/blob/263c76b678275dfff867415c71ba9dc00a9235ef/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L1634-L1643

But yes it would be a good idea to explain delete behavior for OFS.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -469,15 +478,72 @@ public boolean delete(Path f, boolean recursive) throws IOException {
         return false;
       }
 
+      OFSPath ofsPath = new OFSPath(key);
+
```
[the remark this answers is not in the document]
smengcl: Yes. Unfortunately I found that the original check is broken. we might just want to remove it:
```
$ ozone fs -rm -r ofs://omservice/
2020-05-14 18:51:50 INFO  deprecation:1394 - io.bytes.per.checksum is deprecated. Instead, use dfs.bytes-per-checksum
2020-05-14 18:51:50 INFO  BasicRootedOzoneFileSystem:476 - delete: key =
2020-05-14 18:51:50 WARN  BasicRootedOzoneFileSystem:488 - delete: OFS does not support rm root. To wipe the cluster, please re-init OM instead.
rm: `ofs://omservice/': Input/output error
```

`addTrailingSlashIfNeeded` doesn't really add '/' to empty string (root).

also I noticed the `Input/output error` part. looking into it.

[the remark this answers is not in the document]
smengcl: Indeed for o3fs this logic is fixed in HDDS-2973: https://github.com/apache/hadoop-ozone/commit/d1b8c08a71f141be32da459ffa377f7ebdddae27#diff-e48c4ce6b86d4a33c3f38ba8d6d06ea3R442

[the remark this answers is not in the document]
smengcl: I will remove the buggy check logic.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -469,15 +478,72 @@ public boolean delete(Path f, boolean recursive) throws IOException {
...
+      // Handle delete volume
+      if (ofsPath.isVolume()) {
+        String volumeName = ofsPath.getVolumeName();
+        if (recursive) {
+          // Delete all buckets first
+          OzoneVolume volume =
+              adapterImpl.getObjectStore().getVolume(volumeName);
+          Iterator<? extends OzoneBucket> it = volume.listBuckets("");
+          while (it.hasNext()) {
+            OzoneBucket bucket = it.next();
```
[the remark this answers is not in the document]
smengcl: done

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
smengcl: as far as I recall, this should relate to the fact that ozone doesn't create all dirs along the way if we create a key inside a deep directory. e.g. if we create `/volume1/bucket1/dir1/subdir1/key1`, key`/volume1/bucket1/dir1` and `/volume1/bucket1/dir1/subdir1` will NOT be actually created. This breaks some iteration logic (that I couldn't remember exactly which at the moment).

@hanishakoneru should have more info on this.

smengcl: We should remove this logic since HDDS-2940 is in, which solves the problem.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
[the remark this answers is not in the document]
smengcl: done in 5b5b7cae4a86f360ece9ae132444c63361e08f09

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
smengcl: done. I've appended some extra javadoc for delete in 6d7a21b911b7b1ecff3a60443fa72782c5425de0

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
smengcl: done in https://github.com/apache/hadoop-ozone/pull/906/commits/6d7a21b911b7b1ecff3a60443fa72782c5425de0

smengcl:
Thanks @umamaheswararao  for the review and comments. I have updated on all of them. Please take a look.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
umamaheswararao: some special behavior :-). Its good that we will create dirs along the path now. Thanks for the details.

umamaheswararao:
Thanks @smengcl for addressing them.
Latest changes LGTM +1

BTW, could you please comment about CI failures?

smengcl:
@umamaheswararao  Looks like a lot of flaky tests since the last commit is only javadoc change.
WIll trigger a retest empty commit just in case.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
smengcl: No problem :) Filed HDDS-3596.

smengcl:
Will merge this soon. Union of retest 1&2 passes all test suites.

PR #907, opened 2020-05-09
HDDS-3556 Refactor conf in SCMRatisServer to Java-based conf.

timmylicheng:
Refactor SCMRatisServer conf to java based.

https://issues.apache.org/jira/browse/HDDS-3556

## How was this patch tested?
UT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
@@ -386,6 +386,7 @@ public final class StorageContainerManager extends ServiceRuntimeInfoImpl
    * Create an SCM instance based on the supplied configuration.
    *
    * @param conf        HDDS configuration
+   * @param configurator SCM configurator
    * @return SCM instance
    * @throws IOException, AuthenticationException
    */
@@ -410,12 +411,7 @@ public final class StorageContainerManager extends ServiceRuntimeInfoImpl
    */
   public static StorageContainerManager createSCM(OzoneConfiguration conf)
       throws IOException, AuthenticationException {
-    StorageContainerManager scm = new StorageContainerManager(conf);
-    if (SCMHAUtils.isSCMHAEnabled(conf) && scm.getScmRatisServer() == null) {
-      SCMRatisServer scmRatisServer = initializeRatisServer(conf, scm);
-      scm.setScmRatisServer(scmRatisServer);
-    }
-    return scm;
+    return createSCM(conf, new SCMConfigurator());
   }
 
   /**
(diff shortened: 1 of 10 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -1141,9 +1142,11 @@ private void initializeRatisServer() throws IOException {
     if (scmRatisServer == null) {
       SCMNodeDetails scmNodeDetails = SCMNodeDetails
           .initStandAlone(configuration);
-      //TODO enable Ratis ring
-      scmRatisServer = SCMRatisServer.newSCMRatisServer(configuration, this,
-          scmNodeDetails, Collections.EMPTY_LIST);
+      //TODO enable Ratis group
+      scmRatisServer = SCMRatisServer.newSCMRatisServer(configuration
+              .getObject(SCMRatisServer.SCMRatisServerConfiguration.class),
+          this, scmNodeDetails, Collections.EMPTY_LIST,
```
xiaoyuyao, line 1148: Passing half created object (this) to SCMRatisServer constructor could cause subtle bugs that are tricky to get right. We should avoid it if possible.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/ratis/SCMRatisServer.java
```diff
@@ -479,4 +425,166 @@ private void setServerRole(RaftPeerRole currentRole,
       this.roleCheckLock.writeLock().unlock();
     }
   }
+
+  /**
+   * Configuration used by SCM Ratis Server.
+   */
+  @ConfigGroup(prefix = "ozone.scm.ratis")
```
xiaoyuyao, line 432: Is it possible to make these RATIS related configuration reusable among DN,SCM and OM? 
Just different prefix like, ozone.om.ratis, ozone.scm.ratis, ozone.dn.ratis as prefix?

timmylicheng: Yea it looks like we currently have ozone.om.ratis, ozone.scm.ratis, hdds.datanode.ratis, dfs.container.ratis and even dfs.ratis.

If we decide to refactor all config into Java-based configuration, we shall consider refactoring all config to be like ozone.ratis.*

Then we can take a look at reusable configuration among OM, SCM and container. I create Jira to track: https://issues.apache.org/jira/browse/HDDS-3577

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
timmylicheng: I have made updates to have SCM use static method for createSCM and make constructor private. SCMRatisServer will be constructed after SCM is constructed.

ChenSammi:
+1, LGTM.

PR #909, opened 2020-05-10
HDDS-3416. Enable TestSCMNodeManager#testScmStatsFromNodeReport

cku328:

Fixed and enabled this test method(`TestSCMNodeManager#testScmStatsFromNodeReport`).
Added the NodeReportEvent handler to get node reports from datanode.

https://issues.apache.org/jira/browse/HDDS-3416

## How was this patch tested?

Ran UTs - Pass.

### DIFF
--- a/hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestSCMNodeManager.java
+++ b/hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestSCMNodeManager.java
@@ -33,6 +33,7 @@ import org.apache.hadoop.hdds.DFSConfigKeysLegacy;
 import org.apache.hadoop.hdds.HddsConfigKeys;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
+import org.apache.hadoop.hdds.protocol.MockDatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.scm.server.SCMDatanodeHeartbeatDispatcher.NodeReportFromDatanode;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerDatanodeProtocolProtos.NodeReportProto;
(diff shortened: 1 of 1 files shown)


hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestSCMNodeManager.java
```diff
@@ -855,6 +860,10 @@ public void testScmStatsFromNodeReport()
         String storagePath = testDir.getAbsolutePath() + "/" + dnId;
         StorageReportProto report = TestUtils
             .createStorageReport(dnId, storagePath, capacity, used, free, null);
+        NodeReportProto nodeReportProto = TestUtils.createNodeReport(report);
+        nodeReportHandler.onMessage(
```
avijayanhwx, line 864: Shouldn't we find out why the existing NodeReportHandler that is part of the created SCM not getting triggered?

cku328:
Thanks @avijayanhwx for the review and comments.
I have modified this test case mainly by referring to the actions of other similar test cases.

I speculate that the report is only updated under specific conditions, it is not responded to through NodeReportHandler. This requires initializing a NodeReportHandler to process and update the new stored reports. 

A better approach, I think, should be to take nodeReport into the parameter when registering the node and first update it at that time, but in the `TestUtils:createRandomDatanodeAndRegister(...)` method _nodeReport_ is set to null by default.

I have modified the commit, which should be better...

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestSCMNodeManager.java
avijayanhwx: @cku328 Thanks for the refactoring. I looked at the description of the test, and the what we are trying to validate. The test is meant to validate the logic to process a NodeReport correctly. It is not meant to validate whether a Node report is processed IF a heartbeat is got. Hence, I am OK with either of your patches - Using nodeReportHandler.onMessage or using nodeManager.processNodeManager. Maybe it is better to use the onMessage to keep it consistent with the other test you recently fixed. what do you think?

cku328: Thanks @avijayanhwx for the comment, I changed some code so that NodeReport can be sent out with the initial heartbeat. I confirmed the NodeReportHandler in SCM is working properly and it can be triggered to process node reports from multiple nodes.

cku328:
Thanks @elek for telling me about this good helper utility.
I updated my patch, based on the review comments.

PR #912, opened 2020-05-11
WIP Patch - HDDS-2949: store dir/key entries in separate tables - first patch onl…

rakeshadr:
As of HDDS-2940, all the directories from the path prefix get created as entries in the key table. as per the namespace proposal attached to HDDS-2939, directory entries need to be stored in a separate "directory" table. Files will continue to be stored in the key table, which can be thought of as the "file" table.

The advantage of a separate directory table is to make directory lookup more efficient - the entire table would fit into memory for a typical file based dataset.

## How was this patch tested?

Presently used existing `TestOzoneFileSystem` unit test cases. I will add specific unit test cases in an incremental way, in my next commits.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OzoneFSUtils.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OzoneFSUtils.java
@@ -90,6 +90,13 @@ public final class OzoneFSUtils {
     return keyPath.getNameCount();
   }
 
+  public static String appendKeyName(String toKeyName, String fromFileName) {
+    StringBuilder newToKeyName = new StringBuilder(toKeyName);
+    newToKeyName.append(OZONE_URI_DELIMITER);
+    newToKeyName.append(fromFileName);
+    return newToKeyName.toString();
+  }
+
   public static String addTrailingSlashIfNeeded(String key) {
     if (!key.endsWith(OZONE_URI_DELIMITER)) {
       return key + OZONE_URI_DELIMITER;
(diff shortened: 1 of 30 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OzoneFSUtils.java
```diff
@@ -75,6 +75,24 @@ public static String getImmediateChild(String descendant, String ancestor) {
     return descendant;
   }
 
+  /**
+   * The function returns file name from the given absolute path. For
+   * example, the given key path '/a/b/c/d/e/file1' then it returns file name
+   * 'file1'.
+   */
+  public static String getDirName(String keyName) {
+    if (keyName.endsWith("/")) {
```
mukul1987, line 84: Can we use the import java.nio.file.Path here in place of string manipulation and then use path.getname here ?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMDirectoryCreateResponse.java
```diff
@@ -88,4 +96,34 @@ protected void addToDBBatch(OMMetadataManager omMetadataManager,
...
+          LOG.debug("putWithBatch parent : dir {} info : {}", parentKey,
+                  parentDirInfo);
+          omMetadataManager.getDirectoryTable()
+                  .putWithBatch(batchOperation, parentKey, parentDirInfo);
+        }
+      }
+
+      String dirKey = omMetadataManager.getOzoneKey(dirKeyInfo.getVolumeName(),
+              dirKeyInfo.getBucketName(), dirKeyInfo.getKeyName());
+      omMetadataManager.getKeyTable().putWithBatch(batchOperation, dirKey,
```
mukul1987, line 119: this should be directory table as well.

linyiqun:
Hi @rakeshadr , in this WIP patch, I see dir table is introduced but we still use the BUCKET lock to do the global control. However from design doc, we should have one additionally lock like dir lock to do more fine-grained lock control right? Since now we have two separate tables to store file/dir.

BTW, I left some detailed comments on JIRA HDDS-2949 for dir cache design that you mentioned in mailing list.

PR #913, opened 2020-05-11
HDDS-2694. HddsVolume#readVersionFile fails when reading older versions.

avijayanhwx:
HddsVolume#layoutVersion is a version number, supposed to be used for handling upgrades from older versions. Currently only one version is defined. But should a new version be introduced, HddsVolume would fail to read older version file. 

**Note** : I did not add a unit test to check multiple versions since that would mean opening up DatanodeLayoutVersion constructor visibility, or doing reflections to change the VERSION_INFOs. If we want to do that, I can add that test as well.

https://issues.apache.org/jira/browse/HDDS-2694

## How was this patch tested?
Unit test added.


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/HddsVolumeUtil.java
```diff
@@ -143,11 +145,16 @@ public static int getLayOutVersion(Properties props, File versionFile) throws
...
+            dataNodeLayoutVersion.getVersion() == lv));
+    if (!validVersion) {
+      int[] supportedVs =
+          Stream.of(DataNodeLayoutVersion.getAllVersions())
+              .mapToInt(DataNodeLayoutVersion::getVersion).toArray();
       throw new InconsistentStorageStateException("Invalid layOutVersion. " +
-          "Version file has layOutVersion as " + lv + " and latest Datanode " +
-          "layOutVersion is " +
-          DataNodeLayoutVersion.getLatestVersion().getVersion());
+          "Version file has layOutVersion as " + lv + " which is unsupported" +
```
hanishakoneru, line 156: Can we change layOut to layout. Made this mistake previously but since layout is single word, it would be good to remove the capital O.

avijayanhwx: Done.

avijayanhwx:
> Thanks @avijayanhwx for working on this.
> Since we only have 1 version currently, do we really need this fix?
> I am ok with getting this is in for cleaner code. I have a minor comment.

@hanishakoneru  Yes I agree this is not a problem now. However, if you look at the 'getLayOutVersion' method in isolation, it is supposed to check if it is a supported version, and not just the latest version. I was starting to look at Ozone upgrades in total, and found this to be a simple fix to start with.

bharatviswa504:
I have one more thing when looking to this patch, we have not used this layOutVersion in code any where, when generating ContainerDir with this layOut, should we use this version even over there?

And also one more question, we write this layOutVersion to a file. Now when layOutVersion is changed in a newer software version, will we rewrite it or how this is handled, as just changing this place is not a complete fix. Overall I think it should be understood how this layout version will be used when the format changed, any info can you provide to help my understanding.

PR #915, opened 2020-05-12
HDDS-3563. Make s3GateWay s3v volume configurable.

captainzmc:

HDDS-3385 has a big change to s3g for creating buckets. At present, the volume is fixed as s3v. This should be user-configurable to better accommodate the older cluster (0.5).

After rebase master, our production cluster will encounter the problem that the old s3g volume could not find. So we are very eager to solve this problem.

https://issues.apache.org/jira/browse/HDDS-3563

## How was this patch tested?

https://github.com/captainzmc/hadoop-ozone/runs/665703241

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/HddsClientUtils.java
@@ -219,6 +219,16 @@ public final class HddsClientUtils {
   }
 
   /**
+   * Returns the s3VolumeName configured in ConfigurationSource.
+   * @param conf Configuration object
+   * @return s3 volume name
+   */
+  public static String getS3VolumeName(ConfigurationSource conf) {
+    return conf.get(OzoneConfigKeys.OZONE_S3_VOLUME_NAME,
+            OzoneConfigKeys.OZONE_S3_VOLUME_NAME_DEFAULT);
+  }
+
+  /**
    * Returns the maximum no of outstanding async requests to be handled by
    * Standalone and Ratis client.
    */
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -390,6 +390,10 @@ public final class OzoneConfigKeys {
       "ozone.acl.enabled";
   public static final boolean OZONE_ACL_ENABLED_DEFAULT =
       false;
+  public static final String OZONE_S3_VOLUME_NAME =
+          "ozone.s3g.volume.name";
+  public static final String OZONE_S3_VOLUME_NAME_DEFAULT =
+          "s3v";
   public static final String OZONE_S3_AUTHINFO_MAX_LIFETIME_KEY =
       "ozone.s3.token.max.lifetime";
   public static final String OZONE_S3_AUTHINFO_MAX_LIFETIME_KEY_DEFAULT = "3m";
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
@@ -122,7 +122,6 @@ public final class OzoneConsts {
   public static final String STORAGE_DIR_CHUNKS = "chunks";
   public static final String OZONE_DB_CHECKPOINT_REQUEST_FLUSH =
       "flushBeforeCheckpoint";
-  public static final String S3_VOLUME_NAME = "s3v";
 
   /**
    * Supports Bucket Versioning.
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1435,6 +1435,14 @@
   </property>
 
   <property>
+    <name>ozone.s3g.volume.name</name>
+    <value>s3v</value>
+    <tag>OZONE, S3GATEWAY</tag>
+    <description>
+      The volume name to access through the s3gateway.
+    </description>
+  </property>
+  <property>
     <name>ozone.s3g.domain.name</name>
     <value/>
     <tag>OZONE, S3GATEWAY</tag>
--- a/hadoop-hdds/docs/content/interface/S3.md
+++ b/hadoop-hdds/docs/content/interface/S3.md
@@ -24,7 +24,7 @@ summary: Ozone supports Amazon's Simple Storage Service (S3) protocol. In fact, 
 
 Ozone provides S3 compatible REST interface to use the object store data with any S3 compatible tools.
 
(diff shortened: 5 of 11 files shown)


captainzmc:
hi @elek @adoroszlai,  This problem is related to the HDDS-3385 and HDDS-3503. Could you help to review it?

bharatviswa504:
Hi @captainzmc 
Thanks for the contibution.
I have one question, previously we used to compute credentials based on dummy credentials set when security not enabled, when security enabled credentials returned by getSecret command. 

So, volume name is generated based on credentials which will be set differently for different users. So, this config will be used S3Gateway with configured volume name by ozone.s3g.volume.name. So, how this will solve completely for older version like 0.5.0, as volume name will be different for different users, whereas with this approach we configure it to a single value.

bharatviswa504:
If we want to solve this, this should be a list of all volume names which are generated by older software, and compute volume name similar to older software version, and then compare it with list, and use that volume for this bucket. If it does not match with configured list, use the default "s3v" or another fallback config to use default, if the volume name from list does not match with credentials. This is a just an idea to solve this. Or we can wait for cleaner solution of bind mount as pointed by @elek in mailing list.

captainzmc:
hi @bharatviswa504, Thanks for your review.
As @elek suggested before.
[s3v should be configurable instead of constant (which would open a very interesting use case: install multiple s3g with different settings. I think it's good to keep open this possibility).](https://github.com/apache/hadoop-ozone/pull/857#pullrequestreview-401840305)
In tencent, we haven't enabled acl in our cluster at present, and there are only two AWS Accesskey id at present. We think elek's suggestion is advisable. Start different s3gateway for different users using ozone.s3g.volume.name (set it to Accesskey id). Each user use their own s3gateway.

ChenSammi:
> 
> 
> If we want to solve this, this should be a list of all volume names which are generated by older software, and compute volume name similar to older software version, and then compare it with list, and use that volume for this bucket. If it does not match with configured list, use the default "s3v" or another fallback config to use default, if the volume name from list does not match with credentials. This is a just an idea to solve this, might be there is a better alternative to solve this. Or we can wait for the cleaner solution of bind mount as pointed by @elek in the mailing list.

Agree with @bharatviswa504 , it would be better have a list of bucket->volume mapping, otherwise it's hard to maintain differrent configurations for different s3g.

captainzmc:
> > If we want to solve this, this should be a list of all volume names which are generated by older software, and compute volume name similar to older software version, and then compare it with list, and use that volume for this bucket. If it does not match with configured list, use the default "s3v" or another fallback config to use default, if the volume name from list does not match with credentials. This is a just an idea to solve this, might be there is a better alternative to solve this. Or we can wait for the cleaner solution of bind mount as pointed by @elek in the mailing list.
> 
> Agree with @bharatviswa504 , it would be better have a list of bucket->volume mapping, otherwise it's hard to maintain differrent configurations for different s3g.

hi @ChenSammi @bharatviswa504 @elek, Thanks for your discussion.
The bucket is now created with a constant volume(default s3v). If we configure a list of volumes, There are a few problems:
1. When creating a bucket, we don't know which volume to choose. Previously, the volume was determined by the md5Hex value of the AccesskeyId.
2. If there are two buckets with the same bucket name(volume1/bucket, volume2/bucket). When getting this bucket, we do not know which one to choose.
The root cause of the problems above is that, current implementations cannot relate user(AccesskeyId) to volume, unless we revert implementation binds AccesskeyId to volume. Maybe we can use this PR as a temporary solution, until ozone volume 
can be "bind mounted" to other volumes?

bharatviswa504:
> > > If we want to solve this, this should be a list of all volume names which are generated by older software, and compute volume name similar to older software version, and then compare it with list, and use that volume for this bucket. If it does not match with configured list, use the default "s3v" or another fallback config to use default, if the volume name from list does not match with credentials. This is a just an idea to solve this, might be there is a better alternative to solve this. Or we can wait for the cleaner solution of bind mount as pointed by @elek in the mailing list.
> > 
> > 
> > Agree with @bharatviswa504 , it would be better have a list of bucket->volume mapping, otherwise it's hard to maintain differrent configurations for different s3g.
> 
> hi @ChenSammi @bharatviswa504 @elek, Thanks for your discussion.
> The bucket is now created with a constant volume(default s3v). If we configure a list of volumes, There are a few problems:
> 
> 1. When creating a bucket, we don't know which volume to choose. Previously, the volume was determined by the md5Hex value of the AccesskeyId.
> 2. If there are two buckets with the same bucket name(volume1/bucket, volume2/bucket). When getting this bucket, we do not know which one to choose.
> 
> The root cause of the problems above is that, current implementations cannot relate user(AccesskeyId) to volume, unless we revert implementation bind AccesskeyId to volume. Maybe we can use this PR as a temporary solution, until ozone volume can be "bind mounted" to other volumes?

For any request to S3, what volume need to be used if we generate from accessKeyID, and if that volume is there in the list use that or fall back to new code. This should work for all the scenarios mentioned above.

Lets say we got request for 
bucket1 from user1 -> volumeName from accessKeyID lets say is xxxv1, we use xxxv1 as volume.
bucket2 from user1 -> volumeName from accessKeyID lets say is xxxv2, we use xxxv2 as volume.
Now we got a new request, which is bucket3, new one which is not from older version, when we generate volume name from accessKeyID, it does not match with list, so fall back.

So, with this approach, onlything is admin need to make sure that he configure all volumeNames for all users in the cluster, by getting their accessKeyID's.

bharatviswa504:
> If there are two buckets with the same bucket name(volume1/bucket, volume2/bucket). When getting this bucket, we do not know which one to choose
This is not possible with older software, s3 buckets are unique in the system. 

.

captainzmc:
> For any request to S3, what volume need to be used if we generate from accessKeyID, and if that volume is there in the list use that or fall back to new code. This should work for all the scenarios mentioned above.
> 
> Lets say we got request for
> bucket1 from user1 -> volumeName from accessKeyID lets say is xxxv1, we use xxxv1 as volume.
> bucket2 from user1 -> volumeName from accessKeyID lets say is xxxv2, we use xxxv2 as volume.
> Now we got a new request, which is bucket3, new one which is not from older version, when we generate volume name from accessKeyID, it does not match with list, so fall back.
> 
> So, with this approach, onlything is admin need to make sure that he configure all volumeNames for all users in the cluster, by getting their accessKeyID's.

Reuse the AccesskeyId to determine the value of the volume. This conflicts with the design of [hdds-3385](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/docs/content/design/ozone-volume-management.md), which was meant to simplify the implementation by removing the mapping between volume and AccessKeyId. We need more discussion here. cc @elek.

elek:
Thanks the discussion, I try to follow it, but not sure what is the latest proposal.

For me, the easy path seems to be accept this patch and quickly implement the bind mounts.

If I understand well this conversation is more related to the original design and if we can allow backward compatible change or not... (But fix me If I am wrong...)

bharatviswa504:
Yes agreed, long term solution to solve this will be bind mount.

If this PR solves temporarily for Tencent, I am fine with it.

elek:
Thanks @captainzmc the patch and @bharatviswa504 and @ChenSammi the discussionla 

@adoroszlai opened an issue for the bind mounts: https://issues.apache.org/jira/browse/HDDS-3612

I merged this to the master

PR #917, opened 2020-05-13
HDDS-3002. Make the Mountd work for Ozone.

prashantpogde:

-Support for NFS v3 mountd support.
These set of changes are enough to get the mountd started using
'# ozone nfs3'
and it should show up in
'# showmount -e'

https://issues.apache.org/jira/browse/HDDS-3002

## How was this patch tested?

It is tested manually by running a local nfs3 mount server for ozone.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeyRenameResponse.java
```diff
@@ -88,11 +88,22 @@ public void addToDBBatch(OMMetadataManager omMetadataManager,
...
     } else if (createToKeyAndDeleteFromKey()) {
       // If both from and toKeyName are equal do nothing
-      omMetadataManager.getKeyTable().deleteWithBatch(batchOperation,
-          omMetadataManager.getOzoneKey(volumeName, bucketName, fromKeyName));
-      omMetadataManager.getKeyTable().putWithBatch(batchOperation,
-          omMetadataManager.getOzoneKey(volumeName, bucketName, toKeyName),
-          newKeyInfo);
+      if (!toKeyName.equals(fromKeyName)) {
+        omMetadataManager.getKeyTable().deleteWithBatch(batchOperation,
+            omMetadataManager.getOzoneKey(volumeName, bucketName, fromKeyName));
```
mukul1987, line 93: Let have this as a different jira for this along with the Ozone filehandle ?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMDirectoryCreateResponse.java
```diff
@@ -80,7 +80,15 @@ protected void addToDBBatch(OMMetadataManager omMetadataManager,
           dirKeyInfo.getBucketName(), dirKeyInfo.getKeyName());
       omMetadataManager.getKeyTable().putWithBatch(batchOperation, dirKey,
           dirKeyInfo);
-
+      // We can also persist the fileHandle to KeyIdTable.
+      if (dirKeyInfo.getFileHandleInfo() != null) {
+        omMetadataManager.getKeyIdTable().putWithBatch(batchOperation,
+            omMetadataManager.getOzoneKeyIdTableKey(
```
mukul1987, line 86: Lets move this to a seperate change.

mukul1987: This change also needs to go into, OMFileCreateRespnse and OMKeyCreate response as well.

mukul1987:
Also this is same as #828. Can we close the other PR ?

elek:
> I think this patch can be subdivided into a change where FileHandle is introduced in OzoneManager.

Agree, it seems to be one generic patch. Would be easier to follow the development with separating to smaller patches based on the sub-issues of HDDS-3001

prashantpogde:
thanks, I will create smaller patches and upload them.

maobaolong:
@prashantpogde Thank you for working on this, it is great to see you are using OzoneFilesystem to access Ozone, but, i don't know if we have another better idea to avoid bring a new api `getFileHandleInfo`, and if we don't use OzoneFilesystem directly, it would be better to abstract this Ozone NFS into HCFS-NFS.

elek:
/pending "I will create smaller patches and upload them" + question from @maobaolong

PR #918, opened 2020-05-13
WIP HDDS-3572. Make sure ozone. administrators have access to all keys when acl enabled.

captainzmc:

In some cases, a particular user need to access all the keys.
When acl enabled, if  there is a user needed access all the keys. We need to increase the user's access to all the keys. Usually there are a lot of keys in a cluster, so it's very difficult to add permissions to all keys. 
I found that the current ozone Administrators cannot access all keys. Administrators of ozone are also checked for permissions. In HDFS, dfs.cluster.administrators can able to access all files. 

https://issues.apache.org/jira/browse/HDDS-3572

## How was this patch tested?

https://github.com/captainzmc/hadoop-ozone/runs/667241978

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
@@ -18,14 +18,17 @@ package org.apache.hadoop.ozone.om;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+
 
 import org.apache.hadoop.crypto.CipherSuite;
 import org.apache.hadoop.crypto.CryptoProtocolVersion;
 import org.apache.hadoop.crypto.key.KeyProvider;
 import org.apache.hadoop.crypto.key.KeyProviderCryptoExtension;
 import org.apache.hadoop.fs.CommonConfigurationKeys;
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.protocol.StorageType;
 import org.apache.hadoop.ozone.OzoneAcl;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
@@ -44,6 +47,7 @@ import org.iq80.leveldb.DBException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_ADMINISTRATORS;
 import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.BUCKET_NOT_FOUND;
 import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.INTERNAL_ERROR;
 import static org.apache.hadoop.ozone.om.lock.OzoneManagerLock.Resource.BUCKET_LOCK;
@@ -61,25 +65,29 @@ public class BucketManagerImpl implements BucketManager {
    */
   private final OMMetadataManager metadataManager;
   private final KeyProviderCryptoExtension kmsProvider;
+  private OzoneConfiguration conf;
 
   /**
    * Constructs BucketManager.
    *
    * @param metadataManager
    */
-  public BucketManagerImpl(OMMetadataManager metadataManager) {
-    this(metadataManager, null, false);
+  public BucketManagerImpl(OMMetadataManager metadataManager,
+      OzoneConfiguration conf) {
+    this(metadataManager, null, false, conf);
   }
 
   public BucketManagerImpl(OMMetadataManager metadataManager,
-                           KeyProviderCryptoExtension kmsProvider) {
-    this(metadataManager, kmsProvider, false);
+      KeyProviderCryptoExtension kmsProvider, OzoneConfiguration conf) {
+    this(metadataManager, kmsProvider, false, conf);
   }
 
   public BucketManagerImpl(OMMetadataManager metadataManager,
-      KeyProviderCryptoExtension kmsProvider, boolean isRatisEnabled) {
+      KeyProviderCryptoExtension kmsProvider, boolean isRatisEnabled,
+      OzoneConfiguration conf) {
     this.metadataManager = metadataManager;
     this.kmsProvider = kmsProvider;
+    this.conf = conf;
   }
 
   KeyProviderCryptoExtension getKMSProvider() {
@@ -555,6 +563,11 @@ public class BucketManagerImpl implements BucketManager {
       throws OMException {
     Objects.requireNonNull(ozObject);
     Objects.requireNonNull(context);
+    Collection<String> ozAdmins =
+        conf.getTrimmedStringCollection(OZONE_ADMINISTRATORS);
+    if(ozAdmins.contains(context.getClientUgi().getUserName())) {
+      return true;
+    }
 
     String volume = ozObject.getVolumeName();
     String bucket = ozObject.getBucketName();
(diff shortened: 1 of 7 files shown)


vivekratnavel:
Ozone supports Ranger Authorizer for ACLs and whitelisting users to access all keys can easily be achieved by creating a Ranger policy or adding users to an existing Ranger policy. For Ozone native authorizer, we already have ozone.administrators who should be able to access all keys. I don't understand the need for this new white list.

@xiaoyuyao Please correct me if I am wrong.

captainzmc:
Hi @vivekratnavel, Very thanks for your discussion. Your tips are very helpful for me.
Currently, ranger is not enabled in our cluster, we use native acl.
Ozone Administrators might overlap with this whitelist. Instead of adding a white list, we can improve the functionality of Ozone Administrators.
After testing, I found that the current ozone Administrators cannot access all keys. Administrators of ozone are also checked for permissions. Is this a bug? In HDFS, Administrators can able to access all files，and HDFS does not check permissions of Administrators.
I can change this PR to make sure Administrators can access all keys.

captainzmc:
> Ozone supports Ranger Authorizer for ACLs and whitelisting users to access all keys can easily be achieved by creating a Ranger policy or adding users to an existing Ranger policy. For Ozone native authorizer, we already have ozone.administrators who should be able to access all keys. I don't understand the need for this new white list.
> 
> @xiaoyuyao Please correct me if I am wrong.

Hi @vivekratnavel, Very thanks for your discussion. Your tips are very helpful for me.
Currently, ranger is not enabled in our cluster, we use native acl.
Ozone Administrators might overlap with this whitelist. Instead of adding a white list, we can improve the functionality of Ozone Administrators.
After testing, I found that the current ozone Administrators cannot access all keys. Administrators of ozone are also checked for permissions. Is this a bug? In HDFS, Administrators can able to access all files，and HDFS does not check permissions of Administrators.
I can change this PR to make sure Administrators can access all keys.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1589,6 +1593,15 @@ public boolean checkAccess(OzoneObj ozObject, RequestContext context)
     Objects.requireNonNull(ozObject);
     Objects.requireNonNull(context);
     Objects.requireNonNull(context.getClientUgi());
+    String whiteList = conf.get(OZONE_WRITE_LIST);
+    Set<String> whiteListSet = new HashSet<>();
+    LOG.warn("=====keymanagerImpl=======checkAccess=====");
+    if (whiteList != null) {
+      whiteListSet.addAll(Arrays.asList(whiteList.trim().split(",")));
```
timmylicheng: This is going to rely on config to set up white list users, which ends up being static and requires reboot when changes to the lists happen. 
We want some CLI tool to dynamic modify the user list while the cluster is running and the list should be persisted somewhere so that it could load the list after reboot. So then concurrentHashSet may be a better option. Any updates to the list will have to go the concurrent hash set as well as to persistence. 
Also think about OM HA. Would this change be replicate to OM followers? Change to the white list is effectively a state change for OM.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
```diff
@@ -555,6 +564,14 @@ public boolean checkAccess(OzoneObj ozObject, RequestContext context)
       throws OMException {
     Objects.requireNonNull(ozObject);
     Objects.requireNonNull(context);
+    String whiteList = conf.get(OZONE_WRITE_LIST);
+    Set<String> whiteListSet = new HashSet<>();
+    if (whiteList != null) {
+      whiteListSet.addAll(Arrays.asList(whiteList.trim().split(",")));
+    }
+    if(whiteListSet.contains(context.getClientUgi().getUserName())) {
```
timmylicheng: space

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1589,6 +1593,15 @@ public boolean checkAccess(OzoneObj ozObject, RequestContext context)
...
+    }
+    if(whiteListSet.contains(context.getClientUgi().getUserName())) {
```
timmylicheng: space

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1589,6 +1593,15 @@ public boolean checkAccess(OzoneObj ozObject, RequestContext context)
...
+      return true;
```
timmylicheng, line 1599: Consider to add comment for how the white list work here

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1589,6 +1593,15 @@ public boolean checkAccess(OzoneObj ozObject, RequestContext context)
     Objects.requireNonNull(ozObject);
     Objects.requireNonNull(context);
     Objects.requireNonNull(context.getClientUgi());
+    String whiteList = conf.get(OZONE_WRITE_LIST);
+    Set<String> whiteListSet = new HashSet<>();
+    LOG.warn("=====keymanagerImpl=======checkAccess=====");
```
timmylicheng: Is this warn log necessary?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1631,8 +1634,15 @@ public void checkAcls(ResourceType resType, StoreType storeType,
       ACLType aclType, String vol, String bucket, String key,
       UserGroupInformation ugi, InetAddress remoteAddress, String hostName)
       throws OMException {
-    checkAcls(resType, storeType, aclType, vol, bucket, key,
-        ugi, remoteAddress, hostName, true);
+    String whiteList = configuration.get(OZONE_WRITE_LIST);
+    Set<String> whiteListSet = new HashSet<>();
+    if (whiteList != null) {
+      whiteListSet.addAll(Arrays.asList(whiteList.trim().split(",")));
```
timmylicheng: Same here and the whitelist should be member OM. We cannot afford reading config every time the stack reaches here.

xiaoyuyao:
bq. I found that the current ozone Administrators cannot access all keys. Administrators of ozone are also checked for permissions. In HDFS, dfs.cluster.administrators can able to access all files.

Based on [HDDS-1303 design doc](https://issues.apache.org/jira/secure/attachment/12997587/Design%20Doc-%20Native%20ACL%20support%20for%20Ozone.pdf):
"An admin can always take ownership of an object, which means that all Admins always have READ_ACP and WRITE_ACP privileges in the system. "

If I remember correctly, ozone.administrators in native authorizer by default will have READ_ACL/WRITE_ACL permissions but may not have direct access to all the keys even though they can gain read/write access by modify the ACLs indirectly. 

If hdfs superuser bypass is desired, we can add this as an option to native ozone authorizer to honor it.

captainzmc:
> If hdfs superuser bypass is desired, we can add this as an option to native ozone authorizer to honor it. We can easily add that after [HDDS-3391] (I had the patch for [HDDS-3391] last week but somehow the PR was not sent out due to other issues).
Hi @xiaoyuyao Thanks for your reply.
Will the patch of [hdds-3391] be submitted in the near future? We need to use hdds-3572 for the recent upgrade of our cluster, of course we can implement an internal patch first.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
captainzmc: @timmylicheng Thanks for your review.
I think that's a good suggestion. Now the configuration of ozone. administrators is static, and om needs to be restarted for each update. I opened a task [HDDS-3591] to keep track of this.

xiaoyuyao:
@captainzmc Please check: https://github.com/apache/hadoop-ozone/pull/923 Thanks!

captainzmc:
[HDDS-3391](https://github.com/apache/hadoop-ozone/pull/923) solves the same problem. close this.

PR #920, opened 2020-05-14
HDDS-3581. Make replicationfactor can be set as a number

maobaolong:

Now replicationfactor is an enum, which has only two option ONE and THREE, i aim to refactor it into a number, and keep compatible with the earlier client version.

This step is a beginning and make it possible to config any number for replication

https://issues.apache.org/jira/browse/HDDS-3581

## How was this patch tested?

CI tests


xiaoyuyao:
My understanding is that we can change the replication count for closed container? Can we add some document or discussion before making this change as RATIS can't handle all the replication factors.

maobaolong:
@xiaoyuyao This PR aimed to refactor the protocol and relates code to support using integer to specify the count of replication. Now in the proto files, it use enum type to stand for the limit of replication count, i think the limitation is too earlier. I want to remove the limitation from the proto level as a first step, then, using some configuration key to limit it. BTW, any count replication support effort is doing by @runzhiwang and some ratis developer

maobaolong:
This is a document relates this PR, i am writing it, feel free to take a look and left some commets.
https://docs.qq.com/doc/DV2N6bWdCcnJVc3Rk

Any comments will help me, thank you for your review.

xiaoyuyao:
Thanks for the document, that helps. Instead of changing the Enum to int, have you consider adding the additional factors into existing Enum?  Here are some of my thoughts

1. We won't be able to support large replication factor or arbitrary replication factor (even) in near future due to raft limitations.
2. We will have better control over the replication factor that we can support. 
3. We will need to maintain backward and forward compatibility for future changes.

maobaolong:
@xiaoyuyao Thank you for your review, I think this PR is significant. 

> have you consider adding the additional factors into existing Enum (e.g. odd number < 32)

Yeah, this is my first thought, but i change my mind quickly, because, i think we should use configuration to limit the maximum number of replication factor instead of the Enum in the proto. We can imagine that each time we add the additional refactor, we should modify the protocol, and now we support ONE and THREE, after this PR, we support ONE, THREE,....THIRTY-ONE, some day, ratis support factor TWO, we should add TWO in the factor, and the client must be upgraded each time. Another things, many factor in a enum make the source code not pretty.

About your thoughts
> 1. We won't be able to support large replication factor or arbitrary replication factor (even) in near future due to raft limitations.

I think some of even factor like TWO will be support recently. we have strong request to have this feature, and the concept of RAFT support TWO member group, @runzhiwang as a developer of ratis is working on TWO member group ratis now

> 2. We will have better control over the replication factor that we can support.

Yeah, but I think we can have a support list in configuration, we can update our support list but no need to update source code and without client upgrade.

> 3. We will need to maintain backward and forward compatibility for future changes.

Sure, it is necessary, i see a PR to bring a maven plugin to check the proto compatibility, it is great.

As a conclusion, i want to remove the limitation by Enum  type in proto, update the proto, the client  have to upgrade. We really cannot expect a totally factor of replication we will support future, but we can do some efforts like this PR to update the support replication factor without client upgrade  through move the concept of replication factor limitation from Enum to configuration.

elek:
Thanks to create this patch @maobaolong 

It seems to be related to the storage class proposal published on the mailing list:

https://lists.apache.org/thread.html/r1e2a5d5581abe9dd09834305ca65a6807f37bd229a07b8b31bda32ad%40%3Cozone-dev.hadoop.apache.org%3E

With storage class you can remove any replication related information from the proto and you don't need any number just the information about the storage class. One storage class can be resolved as (RATIS/THREE --> CLOSED/TWO) but you can also support (RATIS/ONE --> CLOSED/TWO).

I would prefer to follow (or at least discuss) this more generic approach.

maobaolong:
@elek Thank you for watching my PR. I have read the mail list link you give me before, and i agree this approach is one of the approach to solve the multi replication factor support, i think there are no conflict between my PR and the storage-class approach.

This PR aimed to remove the limitation of replication factor through protocol, instead, i use an integer type to transfer the replication factor from client to server, if we want to set up some limitation, we can use some new configuration. i think this PR is one the way to the storage-class approach.

Lets talk about the storage-class. 
- if we set the storage class like this (RATIS/THREE --> CLOSED/TWO), not sure if we should define a message StorageClass in the proto file which also contains replicationType and replicationFactor, if this design follow this PR, it can be (RATIS/3 --> CLOSED/2). 
- BTW, i think (RATIS/THREE --> CLOSED/TWO) is a good idea to follow the ratis way, but it bring a extra replica writing work during the put operation, for write performance of replication factor TWO, Ozone can be slower than HDFS does.
- About (RATIS/ONE --> CLOSED/TWO). It is another idea, it looks pretty for performance, but for data redundancy, i think it really bring some losing data risk between the ONE->TWO.

So, i think storage-class approach is a good approach, but we should have more discussion work for its design, after that, it can be a pretty approach. And my proposal of this PR only want to open limitation of replication factor, the following work is support more replication factor while writing directly. If fact, i have proved Ozone can write 2 replica directly through ratis in my private branch, i will push more PR after this PR.

elek:
Thanks to explain it @maobaolong 

> not sure if we should define a message StorageClass in the proto file which also contains replicationType and replicationFactor, 

Not exactly. My proposal to include only one "enum/int" in the proto file and remove replicationFactor and replicationType. For example use `1=STANDARAD` instead of `RATIS/THREE` in the proto file.

The exact meaning of STANDARD(1) can be resolved on server side (and can be changed over time). 

> In fact, i have proved Ozone can write 2 replica directly through Ratis... 

I agree with your analyses about the effectiveness of `Ratis/THREE->Closed/TWO` or Ratis/ONE->Closed/TWO`. I was just not sure how is it possible to use TWO with Ratis. Do you use two nodes Ratis cluster? It has totally different availability guarantee than Ratis/THREE as there is no majority just full quorum. I guess it works only if you have two nodes all the time and It couldn't work with loosing any of the nodes.

I am interested if you have more details to share...

*In general*:

I wouldn't like to block this effort. Based on the `storage-class` proposal this change is not required any more as it can be replaced with an even more generic approach (propagate only the ID of the storage-class instead of replication type / factor). 

I am more interested about the consensus about the long-term divergence (long-term = until the next release...), if you need this change immediately, I am not against it.

** Footnote **: 

Just some notes, not closely related. Yesterday I read the [paper of the ChubaoFs](https://arxiv.org/pdf/1911.03001v1.pdf) and it has very interesting approach. The motivation is different, but they two different kind of replications for different use cases:

> The storage engine guarantees the strong consistencyamong the replicas through either primary-backup or Raft-based replication protocols. This design decision is basedon the observations that the former one is not suitable foroverwrite as the replication performance needs to be com-promised, and the latter one has write amplification issue asit introduces extra IO of writing the log files.

With introducing storage-class AND using specific containers / replication (like Erasure Coded containers) our approach can be somewhat similar

maobaolong:
@elek Thank you very mush for the long reply of you, it take me though more, the following is my reply:

> I was just not sure how is it possible to use TWO with Ratis. Do you use two nodes Ratis cluster? 
> I am interested if you have more details to share...

I'm very glad you are interested the replication `2` works what i did. In fact, the master branch cannot works well with replication `2`, but i really did it under the helping of @runzhiwang who is a developer of ratis. 

Firstly, i checkout the current master branch of ratis(commitId: 60643005d0a6faaa25c1810fac9fcb68cda88c3b) and execute `mvn install ...`, then, i did some modification like https://github.com/apache/hadoop-ozone/pull/885 did, it make me build successful. 

Finally, i did a test under a 3 node Ozone cluster. 

```bash
bin/ozone sh key put  -r 2 /myvol/mybucket/NOTICE.txt NOTICE.txt
bin/ozone admin pipeline  list
Pipeline[ Id: 69f66a9d-38e5-4c51-8b7e-bcc2052fa9df, Nodes: 27137460-e588-4d63-bb06-15123772331c{ip: 127.0.0.1, host: localhost, networkLocation: /default-rack, certSerialId: null}affc3600-f15c-4e42-b4ff-4748f8d0cd5d{ip: 127.0.0.1, host: localhost, networkLocation: /default-rack, certSerialId: null}, Type:RATIS, Factor:2, State:OPEN, leaderId:27137460-e588-4d63-bb06-15123772331c, CreationTimestamp2020-05-19T08:08:35.185Z]
bin/ozone sh key info /myvol/mybucket/NOTICE.txt 
{
  "volumeName" : "myvol",
  "bucketName" : "mybucket",
  "name" : "NOTICE.txt",
  "dataSize" : 17540,
  "creationTime" : "2020-05-19T08:02:28.275Z",
  "modificationTime" : "2020-05-19T08:02:29.095Z",
  "replicationType" : "RATIS",
  "replicationFactor" : 2,
  "ozoneKeyLocations" : [ {
    "containerID" : 10,
    "localID" : 104194070824550403,
    "length" : 17540,
    "offset" : 0
  } ],
  "metadata" : { },
  "fileEncryptionInfo" : null
}

```

The above test can be successfully too if there are only 2 datanode in Ozone cluster.


> It has totally different availability guarantee than Ratis/THREE as there is no majority just full quorum.

I don't think there is no majority in a Ratis/TWO. In the 2 nodes group, there are different timeout before election for each node, when one of them achieve the election timeout the node can be a candidate, and `Vote to itself`, then ask the peers for vote it, so it can have majority and leader in 2 node group.

> I guess it works only if you have two nodes all the time and It couldn't work with loosing any of the nodes.

As far as i know, we use ratis pipeline to write block in container, and read block though grpc, so if we lose one of two replication, it can be read as normal. if fact, i just did a test for this.
```bash
# setup a two node Ozone cluster.
bin/ozone sh key put  -r 2 /myvol/mybucket/NOTICE.txt NOTICE.txt 
# kill one of the HddsDatanodeService process.
bin/ozone sh key get /myvol/mybucket/NOTICE.txt /tmp/foo
md5sum /tmp/foo
bb915076158b5871fd2e409a6dbacb1f  tmp/foo
```

> In general ...
Thank you to let me know more about the storage-class approach,  i'm very glad to discuss this approach more detail, it is better to have a jira track this approach, maybe we need to write a design doc as an attachment.

As we don't know how long we will finish the storage-class approach, and i've support 2 replication already, further more, other number of replication can be supported under the further tests, i don't know if we can use my approach currently, there are 2 factor request in our scenario.

> ChubaoFs
It is a well known fs from JD.COM, we can get some idea from it, i'd love to study it more.

PR #922, opened 2020-05-14
HDDS-3588. Fix NPE while getPipelines if absent in query2OpenPipelines

maobaolong:

As title described, there maybe a NPE if get pipeline from query2OpenPipelines return null. It is possible if we can specify a replication number other than 1 and 3.

https://issues.apache.org/jira/browse/HDDS-3588

## How was this patch tested?

Now, this issue cannot be trigger until HDDS-3581 merged.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateMap.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateMap.java
@@ -210,13 +210,9 @@ class PipelineStateMap {
     Preconditions.checkNotNull(state, "Pipeline state cannot be null");
 
     if (state == PipelineState.OPEN) {
-      List<Pipeline> pipeline =
-          query2OpenPipelines.get(new PipelineQuery(type, factor));
-      if (pipeline == null) {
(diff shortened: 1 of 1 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateMap.java
```diff
@@ -210,6 +210,11 @@ Pipeline getPipeline(PipelineID pipelineID) throws PipelineNotFoundException {
     Preconditions.checkNotNull(state, "Pipeline state cannot be null");
 
     if (state == PipelineState.OPEN) {
+      List<Pipeline> pipeline =
+          query2OpenPipelines.get(new PipelineQuery(type, factor));
```
avijayanhwx, line 214: Can we use getOrDefault here?

maobaolong: Thank you, done

maobaolong:
@avijayanhwx Thank you for your review, i have addressed your comments, PTAL.

ChenSammi:
+1.  Thanks @avijayanhwx for the review and @maobaolong  for fix the issue.

PR #924, opened 2020-05-15
HDDS-3595. Add a maven proto file backward compatibility checker in Ozone.

mukul1987:
This patch adds a protobuf backward compatibility checker into Ozone.

https://issues.apache.org/jira/browse/HDDS-3595

## How was this patch tested?
Build works, and able to detect failure because of this.

diff --git a/hadoop-ozone/csi/src/main/proto/csi.proto b/hadoop-ozone/csi/src/main/proto/csi.proto
index 3bd53a075..b8a984de2 100644
--- a/hadoop-ozone/csi/src/main/proto/csi.proto
+++ b/hadoop-ozone/csi/src/main/proto/csi.proto
@@ -119,7 +119,7 @@ message GetPluginInfoResponse {
   // characters or less, beginning and ending with an alphanumeric
   // character ([a-z0-9A-Z]) with dashes (-), dots (.), and
   // alphanumerics between. This field is REQUIRED.
-  string name = 1;
+  string newName = 1;

[ERROR] CONFLICT: "GetPluginInfoResponse" field: "name" has been removed, but is not reserved [csi.proto]
[ERROR] CONFLICT: "GetPluginInfoResponse" field: "newName" ID: 1 has an updated name, previously "name" [csi.proto]

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -243,11 +243,6 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>com.salesforce.servicelibs</groupId>
-        <artifactId>proto-backwards-compatibility</artifactId>
-        <version>${proto-backwards-compatibility.version}</version>
-      </dependency>
-      <dependency>
         <groupId>com.squareup.okhttp</groupId>
         <artifactId>okhttp</artifactId>
         <version>${okhttp.version}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -1602,8 +1602,29 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
   </dependencyManagement>
 
   <build>
+    <extensions>
+      <extension>
+        <groupId>kr.motd.maven</groupId>
+        <artifactId>os-maven-plugin</artifactId>
+        <version>${os-maven-plugin.version}</version>
+      </extension>
+    </extensions>
     <pluginManagement>
       <plugins>
+        <plugin>
+          <groupId>com.salesforce.servicelibs</groupId>
+          <artifactId>proto-backwards-compatibility</artifactId>
+          <configuration>
+            <protoSourceRoot>${basedir}/src/main/proto/</protoSourceRoot>
+          </configuration>
+          <executions>
+            <execution>
+              <goals>
+                <goal>backwards-compatibility-check</goal>
+              </goals>
+            </execution>
+          </executions>
+        </plugin>
         <plugin>
           <groupId>com.github.eirslett</groupId>
           <artifactId>frontend-maven-plugin</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -1609,6 +1609,7 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
         <plugin>
           <groupId>com.salesforce.servicelibs</groupId>
           <artifactId>proto-backwards-compatibility</artifactId>
+          <version>${proto-backwards-compatibility.version}</version>
           <configuration>
             <protoSourceRoot>${basedir}/src/main/proto/</protoSourceRoot>
           </configuration>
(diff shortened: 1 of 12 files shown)


mukul1987:
I found the plugin to be useful to detect backward compatibility issues. However I still am not completely sure about LICENSE and checkstyle and other related issue.

avijayanhwx:
> I found the plugin to be useful to detect backward compatibility issues. However I still am not completely sure about LICENSE and checkstyle and other related issue.

This will be very useful @mukul1987. The license is '**BSD 3-Clause "New" or "Revised" License**'

_- Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
- Neither the name of Salesforce.com nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission._

maobaolong:
@mukul1987 This feature looks good, very useful for keep `backward compatibility`. 👍

elek:
Thanks to open this issue @mukul1987 

I have a few questions:

 * Do we need to check the backward compatibility for RPC or just for the persisted data? If only for the persisted data, it would be more useful to separated the two RPC set.

 * What is the suggested way to use this? I assume that we can do any incompatible change with upgrading the lock file together with the proto.

 * What should we put to the lock file? Is the lock file from the last release should be added and upgrade with each release? Or the first/(last?) version from the previous major line?

mukul1987:
@adoroszlai , Thanks for the review. Have addressed the comments in the followup patch.

mukul1987:
Do we need to check the backward compatibility for RPC or just for the persisted data? If only for the persisted data, it would be more useful to separated the two RPC set.
>> backward compatibility should be on both the protos both for express as well as rolling upgrades.

What is the suggested way to use this? I assume that we can do any incompatible change with upgrading the lock file together with the proto.
>> this fails the compilation with an error so that the user can look into the error and identify that the change is backward incompatible.

What should we put to the lock file? Is the lock file from the last release should be added and upgrade with each release? Or the first/(last?) version from the previous major line?
>> The lock file should be committed along with the source code to identify backward incompatibility issues.

mukul1987:
@elek, can you please have another look at this change ?

pom.xml
```diff
@@ -241,6 +242,11 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>com.salesforce.servicelibs</groupId>
+        <artifactId>proto-backwards-compatibility</artifactId>
+        <version>${proto-backwards-compatibility.version}</version>
+      </dependency>
```
adoroszlai, line 249:
```suggestion
```

pom.xml
```diff
@@ -1596,8 +1602,29 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
...
+        <groupId>kr.motd.maven</groupId>
+        <artifactId>os-maven-plugin</artifactId>
+        <version>${os-maven-plugin.version}</version>
+      </extension>
+    </extensions>
     <pluginManagement>
       <plugins>
+        <plugin>
+          <groupId>com.salesforce.servicelibs</groupId>
+          <artifactId>proto-backwards-compatibility</artifactId>
```
adoroszlai, line 1616:
```suggestion
          <groupId>com.salesforce.servicelibs</groupId>
          <artifactId>proto-backwards-compatibility</artifactId>
          <version>${proto-backwards-compatibility.version}</version>
```

elek:
Thanks the answers @mukul1987 

> backward compatibility should be on both the protos both for express as well as rolling upgrades.

As far as I know there is no plan to support rolling upgrade, we don't need to force backward compatibility for RPC.

>> What should we put to the lock file? Is the lock file from the last release should be added and upgrade with each release? Or the first/(last?) version from the previous major line?

> The lock file should be committed along with the source code to identify backward incompatibility issues.

I understand this part, The question was about the details. Which version is committed and when? 

During an offline discussion we agreed that we can keep the version from the previous release until the next incompatible change. In case of incompatible change (right now) we should update the lock file *and* explain in the PR why did we update it. By default we should keep the lock file untouched (IMHO).

We can also make the process of updating the lock file during the release process more automatic.

But these are just random thoughts. Let's merge this PR first, start to use it, and we can adjust (and document !!) our usage patterns based on the experiences...

elek:
One note: LICENSE.txt shouldn't be modified. We don't include neither the code nor the binary of the plugin in any of the releases. Will revert that line during the commit if no objection.

elek:
Merged. Thanks @mukul1987 the patch and @adoroszlai the review and suggestions.

PR #925, opened 2020-05-15
HDDS-3586. OM HA can be started with 3 isolated LEADER instead of one OM ring

hanishakoneru:

If DNS is unreliable, OM HA cluster could end up with all OMs starting their own single node Ratis servers. This could happen because if a host address in unresolved, OM does not add it to its ratis peers.

https://issues.apache.org/jira/browse/HDDS-3586

## How was this patch tested?

Will add a unit test in next commit.


elek:
Thanks the patch @hanishakoneru 

I started to test it and deployed it exactly the same way as before and now I have three om which are stuck as there are enough voters:

```
2020-05-21 16:39:39 INFO  LeaderElection:61 - om1@group-D66704EFC61C-LeaderElection139: Election REJECTED; received 0 response(s) [] and 2 exception(s); om1@group-D66704EFC61C:t139, leader=null, voted=om1, raftlog=om1@group-D66704EFC61C-SegmentedRaftLog:OPENED:c-1,f-1,i0, conf=-1: [om1:ozone-om-0:9872, om3:ozone-om-2:9862, om2:ozone-om-1:9862], old=null
```

This is exactly what I needed (clean error instead of starting 3 isolated leader).

I also tested with `ozone sh` which doesn't return, but I think it's also expected and will be fixed by @bharatviswa504 as part of refactoring of the replay structure (I would expected a No leader available exception, or sg. like this)

So it works very well. Thank you very much the quick fix.

Will check the code once again tomorrow, after that I am ready to merge it. Are you planning to add some unit tests @hanishakoneru? Or is it covered by the existing acceptance tests?

hanishakoneru:
Thanks @elek for the review. I was planning to add unit tests but got side-tracked. I will update the PR with unit tests today.

elek:
Thanks the update @hanishakoneru 

This pr is on me. It worked well in failure scenario, but when I started in normal scenario I noticed different errors (may or may not be related). I think it's related the DNS which is resolved too late. I have to check it again.

hanishakoneru:
Hi @elek, did you get a chance to check this again?

elek:
Shame on me, I didn't check it again, yet. I think if the address is unresolved at the beginning and resolved during the start up, it couldn't start, but not sure. I have to test it and because it's so complex (I need a full cluster + port forwards + remote debug) I postponed it.

Sorry for the delay, started it now, and will see some results tomorrow...

hanishakoneru:
Thanks @elek. 
Got green CI run. Merging the PR.

PR #926, opened 2020-05-15
HDDS-3596. Clean up unused code after HDDS-2940 and HDDS-2942

smengcl:

It seems some snippets of code should be removed as HDDS-2940 is committed.

For example [this](https://github.com/apache/hadoop-ozone/blob/ffb340e32460ccaa2eae557f0bb71fb90d7ebc7a/hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java#L495-L499)

(From https://github.com/apache/hadoop-ozone/pull/906#discussion_r424873030)

https://issues.apache.org/jira/browse/HDDS-3596

## How was this patch tested?

The existing tests should do.

## Note

This should be committed after HDDS-2942 since that code path isn't fixed yet.


arp7:
/pending We may need to delay fixing this until https://issues.apache.org/jira/browse/HDDS-2942 is committed.

smengcl:
> LGTM, can you rebase the patch? @smengcl

Sure.

BTW HDDS-2942 is still not committed so this won't check in anytime soon I guess.

PR #929, opened 2020-05-16
HDDS-3593. Set a configurable maximum limit of available pipeline

maobaolong:

Set a configurable maximum limit of available pipeline, 

https://issues.apache.org/jira/browse/HDDS-3593

## How was this patch tested?

A new cluster will not create pipeline more than confg  autoCreateLimit,

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
@@ -344,9 +344,10 @@ public final class ScmConfigKeys {
   public static final boolean
       OZONE_SCM_PIPELINE_AUTO_CREATE_FACTOR_ONE_DEFAULT = true;
 
-  public static final String OZONE_SCM_PIPELINE_AUTO_CREATE_LIMIT =
-      "ozone.scm.pipeline.creation.auto.limit";
-  public static final int OZONE_SCM_PIPELINE_AUTO_CREATE_LIMIT_DEFAULT = 3;
+  public static final String OZONE_SCM_PIPELINE_AUTO_CREATE_LIMIT_RATE =
+      "ozone.scm.pipeline.creation.auto.limit.rate";
+  public static final float OZONE_SCM_PIPELINE_AUTO_CREATE_LIMIT_RATE_DEFAULT =
+      0.0f;
 
   public static final String OZONE_SCM_BLOCK_DELETION_MAX_RETRY =
       "ozone.scm.block.deletion.max.retry";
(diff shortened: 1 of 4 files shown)


maobaolong:
@xiaoyuyao  Thanks for review my PR. 

`ozone.scm.ratis.pipeline.limit` is different with `ozone.scm.pipeline.creation.auto.limit`.
- `ozone.scm.ratis.pipeline.limit` is the total limit, if achieve this limit, no more pipeline would be created.
- `ozone.scm.pipeline.creation.auto.limit` is a limitation for `BackgroundPipelineCreator`, before `BackgroundPipelineCreator` want to create a pipeline periodically, it will check if the count of current available pipeline achieve the `ozone.scm.pipeline.creation.auto.limit`, if achieved already, `BackgroundPipelineCreator` will skip this loop.

timmylicheng:
Pipeline limit is better to define along with datanode number. ozone.scm.ratis.pipeline.limit is mainly used in test and please be very careful to set it in distributed cluster since it's static and doesn't consider cluster expansion or shrinkage. 
Also talk with baolong offline regarding this change.
@maobaolong @xiaoyuyao

maobaolong:
@timmylicheng  @xiaoyuyao Thanks for your review and the suggestion. I change the limit to a rate which consider healthy datanode number. PTAL.

xiaoyuyao:
Thanks @timmylicheng and @maobaolong for the clarification. If we really want to push a ratio/factor based on the number of live datanodes, can we move the logic down into RatisPipelineProvider with other limit check? This way, we don't need to introduce NodeManager to BackgroundPipelineCreator.

maobaolong:
Thanks @xiaoyuyao for your suggestion, i if put the limit check to RatisPipelineProvider, all the ratis pipeline creation will be limit, my key point is limit the pipeline creation from BackgroundPipelineCreator, and a bigger limit check for other pipeline creation, for example, the putkey rpc call request a unknown pipeline, like 4 replication, if it cannot find the unknown type and replication factor pipeline, it will create pipeline sync, without the limit for BackgroundPipelineCreator, there must be no quota for the unknown type or replication factor pipeline.

timmylicheng:
> Thanks @xiaoyuyao for your suggestion, i if put the limit check to RatisPipelineProvider, all the ratis pipeline creation will be limit, my key point is limit the pipeline creation from BackgroundPipelineCreator, and a bigger limit check for other pipeline creation, for example, the putkey rpc call request a unknown pipeline, like 4 replication, if it cannot find the unknown type and replication factor pipeline, it will create pipeline sync, without the limit for BackgroundPipelineCreator, there must be no quota for the unknown type or replication factor pipeline.


@maobaolong You still have FACTOR to help you classify different behaviors. RatisPipelineProvider is the best place if you would complex logic for different Ratis pipelines. If you just want to skip creation, you can take a look at skipCreation function in BackgroundPipelineCreator. But I would you suggest to have a complete strategy to allocate different factor's pipelines. 
At the end of the day, your purpose is to support multiple Ratis pipeline with different factors. This would include preallocation of pipelines, resource split and potential pipeline eviction. It may be more than  just stop creating pipelines to leave quota to other pipelines with different factors. It's hard to rely on configs to know the best setup of pipeline quota. And it's almost impossible to maintain to balance of pipeline quota if you consider cluster expansion or shrinkage.

PR #923, opened 2020-05-15
HDDS-3391. Delegate admin ACL checks to Ozone authorizer plugin.

xiaoyuyao:
## What changes were proposed in this pull request?
Delegate admin check (volume create and list) to authorizer plugin.
Bypass admin acl check for ozone native authorizer.  

## What is the link to the Apache JIRA
https://issues.apache.org/jira/browse/HDDS-3391

## How was this patch tested?
Added UT
Existing acceptance tests.

PR #930, opened 2020-05-17
HDDS-3601. Refactor TestOzoneManagerHA.java into multiple tests to avoid frequent timeout issues.

mukul1987:
Refactor TestOzoneManagerHA into multiple smaller tests

https://issues.apache.org/jira/browse/HDDS-3601

## How was this patch tested?
All the 3 new tests pass.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMRatisSnapshots.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMRatisSnapshots.java
@@ -42,7 +42,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 
-import static org.apache.hadoop.ozone.om.TestOzoneManagerHA.createKey;
+import static org.apache.hadoop.ozone.om.TestOzoneManagerHAWithData.createKey;
 
 /**
  * Tests the Ratis snaphsots feature in OM.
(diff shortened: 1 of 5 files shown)


xiaoyuyao:
+1 from me too. Thanks for fixing this.

xiaoyuyao:
Agree this will help the timeout issue of individual tests. But we have a 900s surefire timeout per test module. If they are still under the same module, we still might hit the surefire timeout as observed in latest #923 run.

bharatviswa504:
LGTM.

One question I have is we have a time out rule of 5 minutes, this is applicable for each test right? Splitting it to multiple tests how will it solve the issue?

elek:
Good point @bharatviswa504

adoroszlai:
> One question I have is we have a time out rule of 5 minutes, this is applicable for each test right?

Where is this 5-min timeout defined?

bharatviswa504:
> > One question I have is we have a time out rule of 5 minutes, this is applicable for each test right?
> 
> Where is this 5-min timeout defined?

I mean here for the tests of HA, we have a rule 
  @Rule
  public Timeout timeout = new Timeout(300_000);

bharatviswa504:
> https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html

Thank You @elek for the clear explanation. Now, it makes sense.

elek:
Thanks @mukul1987 the patch. (and @bharatviswa504, @adoroszlai, @xiaoyuyao, @hanishakoneru  the review and the discussion). 

I am merging it to the master...

PR #934, opened 2020-05-18
HDDS-3605. Support close all pipelines.

maobaolong:

Provide a CLI to close all pipeline for same kind or all kinds.

https://issues.apache.org/jira/browse/HDDS-3605

## How was this patch tested?

```bash
# expect close all pipelines
bin/ozone admin pipeline close ALL
# expect close all pipelines with replication factor 3
bin/ozone admin pipeline close ALL -ffc 3
# expect close all pipelines with the CLOSE state
bin/ozone admin pipeline close ALL -fst CLOSE
```

### DIFF
--- a/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/pipeline/ClosePipelineSubcommand.java
+++ b/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/pipeline/ClosePipelineSubcommand.java
@@ -53,12 +53,6 @@ public class ClosePipelineSubcommand implements Callable<Void> {
       required = false)
   private String factor;
 
-  @CommandLine.Option(names = {"-fst", "--filterByState"},
-      description = "Filter listed pipelines by State(OPEN/CLOSE)",
-      defaultValue = "",
-      required = false)
-  private String state;
-
   @Override
   public Void call() throws Exception {
     try (ScmClient scmClient = parent.getParent().createScmClient()) {
@@ -67,11 +61,8 @@ public class ClosePipelineSubcommand implements Callable<Void> {
         List<Pipeline> unsuccessList = new LinkedList<>();
 
         scmClient.listPipelines().stream()
-            .filter(p -> ((Strings.isNullOrEmpty(factor) ||
-                (p.getFactor().toString().compareToIgnoreCase(factor) == 0))
-                && (Strings.isNullOrEmpty(state) ||
-                (p.getPipelineState().toString().compareToIgnoreCase(state)
-                    == 0))))
+            .filter(p -> (Strings.isNullOrEmpty(factor) ||
+                (p.getFactor().toString().compareToIgnoreCase(factor) == 0)))
             .forEach(pipeline -> {
               try {
                 scmClient.closePipeline(
--- a/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/pipeline/ClosePipelineSubcommand.java
+++ b/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/pipeline/ClosePipelineSubcommand.java
@@ -18,11 +18,13 @@
 
 package org.apache.hadoop.hdds.scm.cli.pipeline;
 
+import com.google.common.base.Strings;
 import org.apache.hadoop.hdds.cli.HddsVersionProvider;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.scm.client.ScmClient;
 import picocli.CommandLine;
 
+import java.io.IOException;
 import java.util.concurrent.Callable;
 
 /**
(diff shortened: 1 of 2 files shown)


maobaolong:
@mukul1987 Pease take a look at this PR, thank you.

ChenSammi:
Hi, @maobaolong, can you explain in what situation this command is useful?

maobaolong:
@ChenSammi Thank you for your review.

> can you explain in what situation this command is useful?

The situation is that i want to close all pipeline and new some pipeline with my specify replication factor and replication type, so that, i can test which factor is supported by Ozone and ratis. And this command is very useful for me, i can't imagine without this command, how many works will i do to close pipeline one by one. 

> This command is too powerful, how can we prevent user from mis-operation?

In fact, this command is the combine operation of list pipelines and close them, without this command, we also can do the same operation by writing a bash script to achieve the same goal.
And, pipeline is already admin command, i think we can have a high level general works we should do is make the admin commands safe, such as verify the role of the user.  The following JIRA to track this issue https://issues.apache.org/jira/browse/HDDS-3617

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/pipeline/ClosePipelineSubcommand.java
```diff
@@ -38,14 +40,59 @@
...
-      scmClient.closePipeline(
-          HddsProtos.PipelineID.newBuilder().setId(pipelineId).build());
+      if (pipelineId.equalsIgnoreCase("ALL")) {
+        if (Strings.isNullOrEmpty(factor) && Strings.isNullOrEmpty(state)) {
+          scmClient.listPipelines().forEach(pipeline -> {
+            try {
+              scmClient.closePipeline(
+                  HddsProtos.PipelineID.newBuilder()
+                      .setId(pipeline.getId().getId().toString()).build());
+            } catch (IOException e) {
```
xiaoyuyao, line 69: Should we continue closing the remaining and print out a list of pipeline closed successfully and unsuccessfully at the end?

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/pipeline/ClosePipelineSubcommand.java
```diff
@@ -38,14 +40,59 @@
...
+              throw new IllegalStateException(
+                  "met a exception while closePipeline", e);
+            }
+          });
+        } else {
+          scmClient.listPipelines().stream()
+              .filter(p -> ((Strings.isNullOrEmpty(factor) ||
```
xiaoyuyao, line 76: NIT: can we take care of the filtering before the forEach loop of closePipeline to avoid code duplication?

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/pipeline/ClosePipelineSubcommand.java
```diff
@@ -38,14 +40,59 @@
   @CommandLine.ParentCommand
   private PipelineCommands parent;
 
-  @CommandLine.Parameters(description = "ID of the pipeline to close")
+  @CommandLine.Parameters(description = "ID of the pipeline to close,"
+      + "'ALL' means all pipeline")
   private String pipelineId;
 
+  @CommandLine.Option(names = {"-ffc", "--filterByFactor"},
```
xiaoyuyao, line 47: Can you add some acceptance tests for this as you've done in the description?

maobaolong: Is there some examples for me to reference? I'd like to follow our test framework.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/pipeline/ClosePipelineSubcommand.java
maobaolong: Yeah, thank you for point it, the first condition is redundancy, remove it.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/pipeline/ClosePipelineSubcommand.java
maobaolong: I've got some help from @timmylicheng

maobaolong:
@xiaoyuyao  Thank you for your suggestion
- i've add two list to show the result of this batch operation
- Add a tests
- Remove the redundancy condition.

PTAL.

Please take another look.

nandakumar131:
Thanks @maobaolong for working on this.

I see your point on why we need to support such command options. Will it be ok if we just add an option to pass multiple pipeline IDs to the close command?

These commands are for debugging purposes only. We should not make it easy for someone to close all the available pipelines in the system.

With respect to the current change, why do we need to close a pipeline which is already in closed state? (do we need `-fst` option)

ChenSammi:
> 
> 
> @ChenSammi Thank you for your review.
> 
> > can you explain in what situation this command is useful?
> 
> The situation is that i want to close all pipeline and new some pipeline with my specify replication factor and replication type, so that, i can test which factor is supported by Ozone and ratis. And this command is very useful for me, i can't imagine without this command, how many works will i do to close pipeline one by one.
> 
> > This command is too powerful, how can we prevent user from mis-operation?
> 
> In fact, this command is the combine operation of list pipelines and close them, without this command, we also can do the same operation by writing a bash script to achieve the same goal.
> And, pipeline is already admin command, i think we can have a high level general works we should do is make the admin commands safe, such as verify the role of the user. The following JIRA to track this issue https://issues.apache.org/jira/browse/HDDS-3617

@maobaolong , thanks for the explaination.  It it's only for the test purpose, I stronlgy suggest use a moderate way to achieve this.  One is as @nandakumar131 suggested, allow close pipeline command to accept a list of pipeline IDs or a file of pipeline IDs list. Another is switch on/off the current background pipeline creation process for production/test.  A mis-operation of close all pipeline action in produciton environement can lead serious impact, no matter it's an admin or not.

maobaolong:
@nandakumar131 @ChenSammi  Thank you for your review, let me have a think how to close all pipeline through this PR.

elek:
/pending "Thank you for your review, let me have a think how to close all pipeline through this PR."

PR #935, opened 2020-05-18
HDDS-3606. Add datanode port into the printTopology command output

maobaolong:

Now, i start 3 datanode in a node, the 3 datanode use differents port, but i cannot distinguish which one is HEALTHY, which one is not. A port print is necessary.

https://issues.apache.org/jira/browse/HDDS-3606

## How was this patch tested?

```bash
 bin/ozone  admin printTopology
State = HEALTHY
 127.0.0.1(localhost):RATIS=52789,STANDALONE=52790    /default-rack
 127.0.0.1(localhost):RATIS=52777,STANDALONE=52778    /default-rack
 127.0.0.1(localhost):RATIS=52752,STANDALONE=52753    /default-rack

```

### DIFF
--- a/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/TopologySubcommand.java
+++ b/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/TopologySubcommand.java
@@ -116,12 +116,24 @@ public class TopologySubcommand implements Callable<Void> {
     });
   }
 
+  private String formatPortOutput(List<HddsProtos.Port> ports) {
+    StringBuilder sb = new StringBuilder();
+    for (int i = 0; i < ports.size(); i++) {
+      HddsProtos.Port port = ports.get(i);
+      sb.append(port.getName() + "=" + port.getValue());
+      if (i < ports.size() -1) {
+        sb.append(",");
+      }
+    }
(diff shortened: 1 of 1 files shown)


maobaolong:
@ChenSammi  Please take a look, thank you.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/TopologySubcommand.java
```diff
@@ -116,15 +116,27 @@ private void printOrderedByLocation(List<HddsProtos.Node> nodes) {
     });
   }
 
+  private String formatPortOutput(List<HddsProtos.Port> ports) {
+    StringBuilder sb = new StringBuilder();
+    for (int i = 0; i < ports.size(); i++) {
+      HddsProtos.Port port = ports.get(i);
+      sb.append(port.getName() + "=" + port.getValue());
+      if (i < ports.size() -1) {
```
ChenSammi, line 124: need extra space between - and 1.

maobaolong:
@ChenSammi Thanks for correct me, sorry for my careless, i've address your comments, PTAL.

ChenSammi:
The failed acceptance is not relavent.  Commit to trunk.

PR #937, opened 2020-05-18
HDDS-3607. Lot of warnings at DN startup.

fapifta:

The log message which is at WARN level before the PR, is printed for every edit in the raft log since the last snapshot during DataNode startup, as we replay the edits, we don't need to spam with this many warnings, especially as this is part of normal operations. The proposal is to change it to DEBUG level.

https://issues.apache.org/jira/browse/HDDS-3607

## How was this patch tested?

No tests were done, seems to be too trivial to be wrong :)

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/impl/BlockManagerImpl.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/impl/BlockManagerImpl.java
@@ -108,9 +108,11 @@ public class BlockManagerImpl implements BlockManager {
         // transaction is reapplied in the ContainerStateMachine on restart.
         // It also implies that the given block must already exist in the db.
         // just log and return
-        LOG.warn("blockCommitSequenceId {} in the Container Db is greater than"
-            + " the supplied value {}. Ignoring it",
-            containerBCSId, bcsId);
+        if (LOG.isDebugEnabled()) {
+          LOG.debug("blockCommitSequenceId {} in the Container Db is greater"
+                  + " than the supplied value {}. Ignoring it",
+              containerBCSId, bcsId);
+        }
         return data.getSize();
       }
       // update the blockData as well as BlockCommitSequenceId here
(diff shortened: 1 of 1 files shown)


fapifta:
Hmm... there were an integration test failure that was not related (usual timeout we see in integration tests), I am unsure how to re-initiate testing, but it does not seem to work... :)

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/impl/BlockManagerImpl.java
```diff
@@ -108,9 +108,11 @@ public long putBlock(Container container, BlockData data) throws IOException {
         // transaction is reapplied in the ContainerStateMachine on restart.
         // It also implies that the given block must already exist in the db.
         // just log and return
-        LOG.warn("blockCommitSequenceId {} in the Container Db is greater than"
-            + " the supplied value {}. Ignoring it",
-            containerBCSId, bcsId);
+        if (LOG.isDebugEnabled()) {
```
sodonnel, line 111: Is it necessary to wrap a debug log entry in the `if (LOG.isDebugEnabled())` statement? It is my understanding, that so long as you are not doing string interpolation in the log message and are passing simple variables as any parameters to the log message (ie not doing any computation to calculate the values being passed), then LOG4J does the correct thing and there is no performance penalty to removing the IF wrapper.

sodonnel: http://www.slf4j.org/faq.html#logging_performance suggests the IF wrapper is not needed for performance reasons.

fapifta: yepp, you are right, muscle memory I guess :D Thank you for checking it, I pushed a fix.

vivekratnavel:
@fapifta Thanks for working on this!
@sodonnel @mukul1987 Thanks for the reviews!

PR #938, opened 2020-05-18
HDDS-3608. NPE while process a pipeline report when PipelineQuery absent in query2OpenPipelines

maobaolong:

Fix a NPE while SCM receive and handle a datanode pipeline report, PipelineStateMap will get the pipelineList from `query2OpenPipelines`, and put the `updatedPipeline` to the list, but if the list is NULL, NPE will occur. 

https://issues.apache.org/jira/browse/HDDS-3608

## How was this patch tested?

No need

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateMap.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateMap.java
@@ -374,7 +374,12 @@ class PipelineStateMap {
     PipelineQuery query = new PipelineQuery(pipeline);
     if (updatedPipeline.getPipelineState() == PipelineState.OPEN) {
       // for transition to OPEN state add pipeline to query2OpenPipelines
-      query2OpenPipelines.get(query).add(updatedPipeline);
+      List<Pipeline> pipelineList = query2OpenPipelines.get(query);
+      if (pipelineList == null) {
+        pipelineList = new CopyOnWriteArrayList<>();
+        query2OpenPipelines.put(query, pipelineList);
+      }
+      pipelineList.add(updatedPipeline);
     } else {
       // for transition from OPEN to CLOSED state remove pipeline from
       // query2OpenPipelines
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateMap.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateMap.java
@@ -372,9 +372,9 @@ class PipelineStateMap {
     Pipeline updatedPipeline = pipelineMap.compute(pipelineID,
         (id, p) -> Pipeline.newBuilder(pipeline).setState(state).build());
     PipelineQuery query = new PipelineQuery(pipeline);
+    List<Pipeline> pipelineList = query2OpenPipelines.get(query);
     if (updatedPipeline.getPipelineState() == PipelineState.OPEN) {
       // for transition to OPEN state add pipeline to query2OpenPipelines
-      List<Pipeline> pipelineList = query2OpenPipelines.get(query);
       if (pipelineList == null) {
         pipelineList = new CopyOnWriteArrayList<>();
         query2OpenPipelines.put(query, pipelineList);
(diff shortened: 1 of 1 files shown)


maobaolong:
Hey @linyiqun , would you like to take a look at this PR? Thank you.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateMap.java
```diff
@@ -374,7 +374,12 @@ Pipeline updatePipelineState(PipelineID pipelineID, PipelineState state)
     PipelineQuery query = new PipelineQuery(pipeline);
     if (updatedPipeline.getPipelineState() == PipelineState.OPEN) {
       // for transition to OPEN state add pipeline to query2OpenPipelines
-      query2OpenPipelines.get(query).add(updatedPipeline);
+      List<Pipeline> pipelineList = query2OpenPipelines.get(query);
+      if (pipelineList == null) {
+        pipelineList = new CopyOnWriteArrayList<>();
+        query2OpenPipelines.put(query, pipelineList);
+      }
+      pipelineList.add(updatedPipeline);
```
linyiqun, line 382: Besides above change, Can you also do the empty check for another condition branch?
    } else {
      // for transition from OPEN to CLOSED state remove pipeline from
      // query2OpenPipelines
      query2OpenPipelines.get(query).remove(pipeline);   <---
    }
There is also a chance that remove operation can throw NPE.

maobaolong:
@linyiqun Thanks for your review comments, i've add an additional empty check, PTAL.

linyiqun:
LGTM, +1. Thanks for addressing the comment, @maobaolong .
I didn't have the write access to merge the PR now. So please wait others for helping review and merging.

maobaolong:
@linyiqun Thank you for your review, it's important to receive your +1

maobaolong:
@elek Please take a look at this PR, thank you.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateMap.java
```diff
@@ -372,13 +372,20 @@ Pipeline updatePipelineState(PipelineID pipelineID, PipelineState state)
     Pipeline updatedPipeline = pipelineMap.compute(pipelineID,
         (id, p) -> Pipeline.newBuilder(pipeline).setState(state).build());
     PipelineQuery query = new PipelineQuery(pipeline);
+    List<Pipeline> pipelineList = query2OpenPipelines.get(query);
     if (updatedPipeline.getPipelineState() == PipelineState.OPEN) {
       // for transition to OPEN state add pipeline to query2OpenPipelines
-      query2OpenPipelines.get(query).add(updatedPipeline);
+      if (pipelineList == null) {
```
xiaoyuyao, line 378: query2OpenPipelines has been initialized in initializeQueryMap() based on the RepType/ReFactor and has never been removed from the map. Correct me if I'm wrong, the pipelineList should never be null unless different RepType/Factor are specified in the query, which is currently impossible.

maobaolong: @xiaoyuyao Yeah, you are right while we do something initialized in initializeQueryMap(), but i am working on query unknown replication query dynamically, so initializeQueryMap() just put the known replication factor into the query2OpenPipelines. Further more, i have an idea to remove initializeQueryMap, instead, using a lazy way to add and remove items from/to the query2OpenPipelines, it is only an idea, but for this PR, it makes NPE gone away.

maobaolong:
@xiaoyuyao Sorry, i don't find your comments inline? Could you show me where is the inline comments you mean?

xiaoyuyao:
> @xiaoyuyao Sorry, i don't find your comments inline? Could you show me where is the inline comments you mean?

The comments is just above:

query2OpenPipelines has been initialized in initializeQueryMap() based on the RepType/ReFactor and has never been removed from the map. Correct me if I'm wrong, the pipelineList should never be null unless different RepType/Factor are specified in the query, which is currently impossible.

maobaolong:
@xiaoyuyao Sorry, i saw your reply. You are right with the condition `query2OpenPipelines has been initialized in initializeQueryMap() based on the RepType/ReFactor and has never been removed from the map`, but for the logic of the method updatePipelineState itself, i think it is better to have a non-null check, and do something to fall back. Actually, i'm just doing something broken the condition you've mentioned before, so i met the NPE already.

xiaoyuyao:
Sounds good to me. Thanks @maobaolong and all for the discussion.

PR #940, opened 2020-05-18
HDDS-3614. Remove S3Table from OmMetadataManager.

bharatviswa504:

Remove S3Table from OmMetadataManager, which are not used any more after HDDS-3385.

https://issues.apache.org/jira/browse/HDDS-3614

## How was this patch tested?

Cleanup Jira.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMMetadataManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMMetadataManager.java
@@ -301,14 +301,6 @@ public interface OMMetadataManager {
   Table<OzoneTokenIdentifier, Long> getDelegationTokenTable();
 
   /**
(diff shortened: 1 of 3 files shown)


ChenSammi:
@bharatviswa504,  the patch overall looks good. 
The static final string "S3_TABLE" and it's usage in the addOMTablesAndCodecs is missed.

bharatviswa504:
Thank You @ChenSammi for the review.
Addressed review comments.

bharatviswa504:
Hi @adoroszlai 
To solve this incompatibility, it is said we shall use bind mounts, untill then using configurable volume for s3 gateway.

https://github.com/apache/hadoop-ozone/pull/915#issuecomment-630146061

Above is the link and refer to PR for discussion. 

If there is another alternative proposal on the table to use the information from table, I am fine with delaying it. Can you share what is the proposal and how do we use this table?

captainzmc:
LGTM +1. The S3Table is no longer useful after HDDS-3385, and the original mapping relationship should not be needed.

PR #941, opened 2020-05-18
HDDS-3574. Implement ofs://: Override getTrashRoot

smengcl:

Implement `getTrashRoot()` in OFS so deleting to trash will always move files/dirs within the same bucket.

https://issues.apache.org/jira/browse/HDDS-3574

## How was this patch tested?

Added new test. Manually tested in docker.

## To Check

1. Whether key path char set includes the char set for user name

## Before and After patch

Before the patch the FS client might attempt to move the file to another bucket when `-skipTrash` is not specified and `fs.trash.interval > 0`.

```bash
# Configure fs.trash.interval to a value larger than 0
bash-4.2$ vi /etc/hadoop/core-site.xml
bash-4.2$ ozone fs -mkdir -p ofs://om/vol1/bucket1/dir1/
bash-4.2$ ozone fs -touch ofs://om/vol1/bucket1/dir1/key1 
bash-4.2$ ozone fs -rm ofs://om/vol1/bucket1/dir1/key1
2020-05-18 17:57:44,953 [main] INFO Configuration.deprecation: io.bytes.per.checksum is deprecated. Instead, use dfs.bytes-per-checksum
2020-05-18 17:57:44,971 [main] INFO ozone.BasicRootedOzoneFileSystem: getTrashRoot: path = ofs://om/vol1/bucket1/dir1/key1
2020-05-18 17:57:45,069 [main] INFO rpc.RpcClient: Creating Volume: user, with hadoop as owner.
2020-05-18 17:57:45,104 [main] INFO rpc.RpcClient: Creating Bucket: user/hadoop, with Versioning false and Storage Type set to DISK and Encryption set to false 
rm: Failed to move to trash: ofs://om/vol1/bucket1/dir1/key1: Cannot rename a key to a different bucket. Consider using -skipTrash option
```

After the patch the client should move the deleted file to trash under the bucket.

```bash
bash-4.2$ vi /etc/hadoop/core-site.xml
bash-4.2$ ozone fs -mkdir -p ofs://om/vol1/bucket1/dir1/
2020-05-19 18:04:41,914 [main] INFO rpc.RpcClient: Creating Volume: vol1, with hadoop as owner.
2020-05-19 18:04:42,076 [main] INFO rpc.RpcClient: Creating Bucket: vol1/bucket1, with Versioning false and Storage Type set to DISK and Encryption set to false
bash-4.2$ ozone fs -touch ofs://om/vol1/bucket1/dir1/key1
bash-4.2$ ozone fs -rm ofs://om/vol1/bucket1/dir1/key1
2020-05-19 18:04:52,055 [main] INFO Configuration.deprecation: io.bytes.per.checksum is deprecated. Instead, use dfs.bytes-per-checksum
2020-05-19 18:04:52,274 [main] INFO fs.TrashPolicyDefault: Moved: 'ofs://om/vol1/bucket1/dir1/key1' to trash at: ofs://om/vol1/bucket1/.Trash/hadoop/Current/vol1/bucket1/dir1/key1
```

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
@@ -18,7 +18,11 @@
 package org.apache.hadoop.ozone.shell;
 
 import com.google.common.base.Strings;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.FileUtil;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.ozone.OFSPath;
 import org.apache.hadoop.fs.ozone.OzoneFsShell;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.ozone.MiniOzoneCluster;
@@ -56,6 +60,7 @@ import java.util.List;
 import java.util.UUID;
 
 import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_TRASH_INTERVAL_KEY;
+import static org.apache.hadoop.fs.FileSystem.FS_DEFAULT_NAME_KEY;
 import static org.apache.hadoop.ozone.OzoneConsts.OZONE_OFS_URI_SCHEME;
 import static org.junit.Assert.fail;
 
@@ -448,21 +453,39 @@ public class TestOzoneShellHA {
 
   @Test
   public void testDeleteToTrash() throws Exception {
+    final String hostPrefix = OZONE_OFS_URI_SCHEME + "://" + omServiceId;
     OzoneConfiguration confcli = new OzoneConfiguration(conf);
-    confcli.setInt(FS_TRASH_INTERVAL_KEY, 60);
     confcli.set("fs.ofs.impl",
         "org.apache.hadoop.fs.ozone.RootedOzoneFileSystem");
+    confcli.set(FS_DEFAULT_NAME_KEY, hostPrefix);
+    confcli.setInt(FS_TRASH_INTERVAL_KEY, 60);
     OzoneFsShell shell = new OzoneFsShell(confcli);
-    final String hostPrefix = OZONE_OFS_URI_SCHEME + "://" + omServiceId;
-    final String dirPrefix = hostPrefix + "/volumed2t/bucket1/dir1";
+    FileSystem fs = FileSystem.get(confcli);
+    final String strDir1 = hostPrefix + "/volumed2t/bucket1/dir1";
+    final String strKey1 = strDir1 + "/key1";
+    final Path pathKey1 = new Path(strKey1);
+    final OFSPath ofsPathKey1 = new OFSPath(strKey1);
+    final Path trashPathKey1 = Path.mergePaths(
+        new Path(ofsPathKey1.getTrashRoot(), "Current"), pathKey1);
     int res;
     try {
-      res = ToolRunner.run(shell, new String[]{"-mkdir", "-p", dirPrefix});
+      res = ToolRunner.run(shell, new String[]{"-mkdir", "-p", strDir1});
       Assert.assertEquals(0, res);
-      res = ToolRunner.run(shell, new String[]{"-touch", dirPrefix + "/key1"});
+      res = ToolRunner.run(shell, new String[]{"-touch", strKey1});
       Assert.assertEquals(0, res);
-      res = ToolRunner.run(shell, new String[]{"-rm", dirPrefix + "/key1"});
+      // Verify key1 creation
+      FileStatus statusPathKey1 = fs.getFileStatus(pathKey1);
+      Assert.assertEquals(strKey1, statusPathKey1.getPath().toString());
+      // rm without skipTrash. since trash interval > 0, should moved to trash
+      res = ToolRunner.run(shell, new String[]{"-rm", strKey1});
       Assert.assertEquals(0, res);
+      // Verify that the file is moved to the correct trash location
+      FileStatus statusTrashPathKey1 = fs.getFileStatus(trashPathKey1);
+      // It'd be more meaningful if we actually write some content to the file
+      Assert.assertEquals(
+          statusPathKey1.getLen(), statusTrashPathKey1.getLen());
+      Assert.assertEquals(
+          fs.getFileChecksum(pathKey1), fs.getFileChecksum(trashPathKey1));
     } finally {
       shell.close();
     }
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
@@ -19,6 +19,7 @@ package org.apache.hadoop.ozone.shell;
 
 import com.google.common.base.Strings;
 import org.apache.hadoop.fs.FileUtil;
+import org.apache.hadoop.fs.ozone.OzoneFsShell;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.ozone.MiniOzoneCluster;
 import org.apache.hadoop.ozone.MiniOzoneHAClusterImpl;
@@ -29,6 +30,7 @@ import org.apache.hadoop.ozone.om.OzoneManager;
 import org.apache.hadoop.ozone.shell.s3.S3Shell;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.test.GenericTestUtils;
+import org.apache.hadoop.util.ToolRunner;
 import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Assert;
@@ -53,6 +55,8 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
 
+import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_TRASH_INTERVAL_KEY;
+import static org.apache.hadoop.ozone.OzoneConsts.OZONE_OFS_URI_SCHEME;
 import static org.junit.Assert.fail;
 
 /**
@@ -442,6 +446,27 @@ public class TestOzoneShellHA {
     Assert.assertEquals(0, getNumOfBuckets("bucket"));
   }
 
+  @Test
+  public void testDeleteToTrash() throws Exception {
+    OzoneConfiguration confcli = new OzoneConfiguration(conf);
+    confcli.setInt(FS_TRASH_INTERVAL_KEY, 60);
+    confcli.set("fs.ofs.impl",
+        "org.apache.hadoop.fs.ozone.RootedOzoneFileSystem");
+    OzoneFsShell shell = new OzoneFsShell(confcli);
+    final String hostPrefix = OZONE_OFS_URI_SCHEME + "://" + omServiceId;
+    final String dirPrefix = hostPrefix + "/volumed2t/bucket1/dir1";
+    int res;
+    try {
+      res = ToolRunner.run(shell, new String[]{"-mkdir", "-p", dirPrefix});
+      Assert.assertEquals(0, res);
+      res = ToolRunner.run(shell, new String[]{"-touch", dirPrefix + "/key1"});
+      Assert.assertEquals(0, res);
+      res = ToolRunner.run(shell, new String[]{"-rm", dirPrefix + "/key1"});
+      Assert.assertEquals(0, res);
+    } finally {
+      shell.close();
+    }
+  }
 
   @Test
   public void testS3PathCommand() throws Exception {
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
@@ -54,6 +54,7 @@ import picocli.CommandLine.RunLast;
 
 import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.PrintStream;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -66,7 +70,13 @@
   static final String OFS_MOUNT_TMP_VOLUMENAME = "tmp";
 
   OFSPath(Path path) {
-    String pathStr = path.toUri().getPath();
+    URI uri = path.toUri();
+    // scheme is case-insensitive
+    if (!uri.getScheme().toLowerCase().equals(OZONE_OFS_URI_SCHEME)) {
+      throw new ParseException("Can't parse schemes other than ofs://.");
+    }
+    authority = uri.getAuthority();
```
bharatviswa504, line 78: Should we set authority in the initOFSPath??
As there we have checked for auth.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -78,8 +88,8 @@ private void initOFSPath(String pathStr) {
     // pathStr should not have authority
     try {
       URI uri = new URI(pathStr);
-      String authority = uri.getAuthority();
-      if (authority != null && !authority.isEmpty()) {
+      String auth = uri.getAuthority();
```
bharatviswa504, line 91: Thinking more, do we need this, from above uri, we can get authority and check if it is empty or nulll and then throw error. So, in this way, we don't need to convert path to URI again.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -238,4 +248,23 @@ static String getTempMountBucketNameOfCurrentUser() throws IOException {
...
   }
+
+  /**
+   * Return trash root for the given path.
+   * @return trash root for the given path
+   */
+  public Path getTrashRoot() {
+    try {
+      String username = UserGroupInformation.getCurrentUser().getUserName();
+      URI uri = new URIBuilder().setScheme(OZONE_OFS_URI_SCHEME)
```
umamaheswararao, line 259: Should it be first username and then .Trash ? like ofs:///vol1/bucket1/username/.Trash  ?

This is default iml at FileSystem.java
```

 /**
   * Get the root directory of Trash for current user when the path specified
   * is deleted.
   *
   * @param path the trash root of the path to be determined.
   * @return the default implementation returns {@code /user/$USER/.Trash}
   */
  public Path getTrashRoot(Path path) {
    return this.makeQualified(new Path(getHomeDirectory().toUri().getPath(),
        TRASH_PREFIX));
  }
```
CC: @mukul1987 @sadanand48 @arp7 : I think this should be checked with Sadanand implementation as well to be in sync?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -78,8 +88,8 @@ private void initOFSPath(String pathStr) {
     // pathStr should not have authority
```
xiaoyuyao, line 88: should we change initOFSPath to take URI as parameter so that all the URI check can be done inside initOFSPath?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
xiaoyuyao: Given we don't have a home directory per bucket, I think this PR took an approach similar to encryption zone where the .Trash is created per bucket, i.e., ofs://vol1/bucket1/.Trash/username, which have a common trash root per bucket. 

If we take ofs:///vol1/bucket1/username/.Trash, we won't have a common trash root. As a result, the trash clean up will be expensive as we don't have a common root per bucket to check the expired trash checkpoints.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
```diff
@@ -238,4 +248,23 @@ static String getTempMountBucketNameOfCurrentUser() throws IOException {
...
+
+  /**
+   * Return trash root for the given path.
+   * @return trash root for the given path
+   */
+  public Path getTrashRoot() {
+    try {
+      String username = UserGroupInformation.getCurrentUser().getUserName();
+      URI uri = new URIBuilder().setScheme(OZONE_OFS_URI_SCHEME)
+          .setHost(authority).setPath(OZONE_URI_DELIMITER + volumeName +
```
xiaoyuyao, line 260: NIT: we can use Path API to create the path safely instead of string concat.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
```diff
@@ -442,6 +446,27 @@ public void testOzoneShCmdList() {
...
+    final String hostPrefix = OZONE_OFS_URI_SCHEME + "://" + omServiceId;
+    final String dirPrefix = hostPrefix + "/volumed2t/bucket1/dir1";
+    int res;
+    try {
+      res = ToolRunner.run(shell, new String[]{"-mkdir", "-p", dirPrefix});
+      Assert.assertEquals(0, res);
+      res = ToolRunner.run(shell, new String[]{"-touch", dirPrefix + "/key1"});
+      Assert.assertEquals(0, res);
+      res = ToolRunner.run(shell, new String[]{"-rm", dirPrefix + "/key1"});
+      Assert.assertEquals(0, res);
```
xiaoyuyao, line 465: Can you also verify the key1 is moved to the proper Trash location?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
```diff
@@ -442,6 +446,27 @@ public void testOzoneShCmdList() {
...
+    OzoneFsShell shell = new OzoneFsShell(confcli);
+    final String hostPrefix = OZONE_OFS_URI_SCHEME + "://" + omServiceId;
+    final String dirPrefix = hostPrefix + "/volumed2t/bucket1/dir1";
+    int res;
+    try {
+      res = ToolRunner.run(shell, new String[]{"-mkdir", "-p", dirPrefix});
+      Assert.assertEquals(0, res);
+      res = ToolRunner.run(shell, new String[]{"-touch", dirPrefix + "/key1"});
+      Assert.assertEquals(0, res);
+      res = ToolRunner.run(shell, new String[]{"-rm", dirPrefix + "/key1"});
```
xiaoyuyao, line 464: Can you add a test with skipTrash and make sure it is not moved to the Trash location?

arp7:
Just one thought about permissions. How will we ensure data deleted by one user and moved to `.Trash` is not made visible to another user who doesn't have permissions to see it?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -658,6 +658,19 @@ public String getUsername() {
...
   }
 
+  /**
+   * Get the root directory of Trash for a path in OFS.
+   * Returns /<volumename>/<bucketname>/.Trash/<username>
+   * Caller appends either Current or checkpoint timestamp for trash destination
+   * @param path the trash root of the path to be determined.
+   * @return trash root
+   */
+  @Override
```
xiaoyuyao, line 668: You also need to Override getTrashRoots() which returns all trash locations under the root across multiple buckets. This should be very similar to how Trash under encryption zone is handled in DistributedFileSystem#getTrashRoots().

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
umamaheswararao: Thanks @xiaoyuyao for explanation.
My point is: after looking at Trash emptier impl demo today, in o3fs we seems to be getting <user>/.Trash
My point is, this o3fs and ofs trash directories should be in sync. so that emptier will not omit the  trash dirs created by o3fs?
Also inside OM Server, we get trashRoots from the fs configured as OM. That should default fs. When we configure ofs, it should also consider trash dirs created by o3fs? I think yes, so that they both should be in sync.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: `initOFSPath` was designed to take a path without scheme or authority though. I didn't intend to change the parsing logic too much but I might do a refactoring to make it more consistent overall since you mentioned.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: I agree o3fs and ofs trash locations should be the same. We should refactor the common parts to some new shared classes later.

I didn't get the "OM Server" part though. IMO `getTrashRoot` is purely a client-side method?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: @bharatviswa504  also mentioned this. Will do.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: Do you mean using constructor `new Path(prevPath, "child")` and building the path step by step?

smengcl:
Also the `getTrashRoot` code doesn't handle the case where the input path is root, volume or bucket. Will post update in a moment.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
umamaheswararao: Yes its client side logic. TrashEmptier uses fs to get the directories and this TrashEmptier runs inside NN, so here it will be in OM server.
If we wanted to reuse the logic, all dirs should be in sync and TrashEmptier will get the trashRoots and cleanup.

For your ref below is the TrashEmptier code: 
 ```
 try {
          now = Time.now();
          if (now >= end) {
            Collection<FileStatus> trashRoots;
            trashRoots = fs.getTrashRoots(true);      // list all trash dirs

            for (FileStatus trashRoot : trashRoots) {   // dump each trash
              if (!trashRoot.isDirectory())
                continue;
              try {
                TrashPolicyDefault trash = new TrashPolicyDefault(fs, conf);
                trash.deleteCheckpoint(trashRoot.getPath(), false);
                trash.createCheckpoint(trashRoot.getPath(), new Date(now));
              } catch (IOException e) {
                LOG.warn("Trash caught: "+e+". Skipping " +
                    trashRoot.getPath() + ".");
              } 
            }
          }
        } catch (Exception e) {
          LOG.warn("RuntimeException during Trash.Emptier.run(): ", e); 
        }
```

smengcl:
> Just one thought about permissions. How will we ensure data deleted by one user and moved to `.Trash` is not made visible to another user who doesn't have permissions to see it?

I think key ACL already kind of solves this problem?
When a user moves a key, the ACL should, in theory, be kept like it was before the move?

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
smengcl: Good catch. It also seems [`Emptier`](https://github.com/apache/hadoop/blob/f86de6f76a3079c2655df9b242fc968edfb17b9d/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/TrashPolicyDefault.java#L288) in `TrashPolicyDefault` utilizes `getTrashRoots()`, which @umamaheswararao believes might make trash clean up easier in Ozone.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
bharatviswa504: I am in favor of maintaining trash roots for ofs and o3fs the same, as @umamaheswararao said in the first phase, we can reuse the code of hdfs logic and run the same in OM with OFS filesystem object and also for this we need to implement fs.getTrashRoots return path for all buckets in the system(Which can be done in a separate Jira).

xiaoyuyao: @umamaheswararao I think we are on the same page. Both o3fs and ofs have Trash per bucket. And code should be reused.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/OFSPath.java
smengcl: Attempted in 53c19f1f1f543978f7e381d2160e35415756a859.

I also explored [`Paths.get()`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Paths.html) but it returns `java.nio.file.Paths`.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
smengcl: done in 31969f9066a3e81fdf44e40b04d676e341604648

arp7:
> But we might need to set the ACL on /.Trash/<username>/ to that user only though. @xiaoyuyao Sounds good, we can do this in a followup jira.

smengcl:
> > > Just one thought about permissions. How will we ensure data deleted by one user and moved to `.Trash` is not made visible to another user who doesn't have permissions to see it?
> > 
> > 
> > I think key ACL already kind of solves this problem.
> > When a user moves a key, the ACL should, in theory, be kept as-is?
> > But we might need to set the ACL on `/.Trash/<username>/` to that user only though. @xiaoyuyao Thoughts?
> 
> Yes. TrashPolicyDefault has those the parent path of the deleted file created under the trash root with owner has ALL access and group/others has NONE access.

Thanks for the insight. So `TrashPolicyDefault` has already handled the trash directory permission logic: https://github.com/apache/hadoop/blob/branch-3.2.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/TrashPolicyDefault.java#L153

xiaoyuyao:
+1, feel free to merge if there is no additional comments.

smengcl:
> +1, feel free to merge if there is no additional comments.

Thanks. Will merge shortly.

hadoop-ozone/ozonefs/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
smengcl: Created HDDS-3705 to track `getTrashRoots` impl.

PR #942, opened 2020-05-19
HDDS-2556. Handle InterruptedException in BlockOutputStream

dineshchitlangia:
Address sonar violations to handle InterruptedException appropriately.

https://issues.apache.org/jira/browse/HDDS-2556

## How was this patch tested?
Sonar plugin in IntelliJ, Clean build, unit tests

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
@@ -317,8 +317,14 @@ public class BlockOutputStream extends OutputStream {
       if (!commitWatcher.getFutureMap().isEmpty()) {
         waitOnFlushFutures();
       }
-    } catch (InterruptedException | ExecutionException e) {
+    } catch (ExecutionException e) {
       setIoException(e);
+      adjustBuffersOnException();
+      throw getIoException();
+    } catch (InterruptedException ex) {
+      LOG.error("Command execution was interrupted.");
+      Thread.currentThread().interrupt();
+      setIoException(ex);
       adjustBuffersOnException();
       throw getIoException();
     }
@@ -427,9 +433,14 @@ public class BlockOutputStream extends OutputStream {
         setIoException(ce);
         throw ce;
       });
-    } catch (IOException | InterruptedException | ExecutionException e) {
+    } catch (IOException | ExecutionException e) {
       throw new IOException(
           "Unexpected Storage Container Exception: " + e.toString(), e);
+    } catch (InterruptedException ex) {
+      LOG.error("Command execution was interrupted.");
+      Thread.currentThread().interrupt();
+      throw new IOException(
+          "Unexpected Storage Container Exception: " + ex.toString(), ex);
     }
     commitWatcher.getFutureMap().put(flushPos, flushFuture);
     return flushFuture;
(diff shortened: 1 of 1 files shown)


dineshchitlangia:
Test failure is reported in HDDS-3083

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -317,10 +317,16 @@ private void handleFullBuffer() throws IOException {
...
-    } catch (InterruptedException | ExecutionException e) {
+    } catch (ExecutionException e) {
       setIoException(e);
       adjustBuffersOnException();
       throw getIoException();
+    } catch (InterruptedException ex) {
+      LOG.error("Command execution was interrupted.");
+      Thread.currentThread().interrupt();
+      setIoException(ex);
+      adjustBuffersOnException();
```
bharatviswa504, line 328: Same code is in multiple places, can we have short utility to do all these in that. So, this will avoid code smell in sonar.

dineshchitlangia: @bharatviswa504  I have tried to move the duplicate lines to separate method and also perform other cleanups. 
At the very least, the thread interrupt must be done in the catch block and not in a separate method being called from the catch block. So, the thread.interrupt() is the only line that cannot be avoided from being duplicated.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -654,4 +662,34 @@ private void writeChunkToContainer(ChunkBuffer chunk) throws IOException {
...
+   * @param ex
+   * @param skipExecutionException is optional, if passed as TRUE, then
+   * handle ExecutionException else skip it.
+   * @throws IOException
+   */
+  private void handleInterruptedException(Exception ex,
+      boolean skipExecutionException)
+      throws IOException {
+    LOG.error("Command execution was interrupted.");
+    if(skipExecutionException) {
```
bharatviswa504, line 678: Minor NIT: variable name is called skipExecutionException, but when true, we are handling ExecutionException. I think it should be viceversa.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -654,4 +662,34 @@ private void writeChunkToContainer(ChunkBuffer chunk) throws IOException {
   public void setXceiverClient(XceiverClientSpi xceiverClient) {
     this.xceiverClient = xceiverClient;
   }
+
+  /**
+   * Handles InterruptedExecution.
+   *
+   * @param ex
+   * @param skipExecutionException is optional, if passed as TRUE, then
```
bharatviswa504, line 670: NIT: javadoc also need to be updated for this.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
dineshchitlangia: Addressed in latest commit.

dineshchitlangia:
Thanks @bharatviswa504 for review. I have merged this to master.

PR #943, opened 2020-05-19
HDDS-3615. Call cleanup on tables only when double buffer has transactions related to tables.

bharatviswa504:
Call cleanup on tables only when the double buffer has transactions related to tables.

https://issues.apache.org/jira/browse/HDDS-3615

## How was this patch tested?

Existing tests. Will see if I can add any tests.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
@@ -37,6 +37,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.apache.hadoop.hdds.function.SupplierWithIOException;
 import org.apache.hadoop.hdds.tracing.TracingUtil;
+import org.apache.hadoop.ozone.om.response.CleanupTableInfo;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMResponse;
 import org.apache.hadoop.util.Time;
 import org.slf4j.Logger;
@@ -390,16 +391,35 @@ public final class OzoneManagerDoubleBuffer {
         opName.toLowerCase().contains(BUCKET)) {
       if (DeleteBucket.name().equals(opName)
           || DeleteVolume.name().equals(opName)) {
-        entry.getResponse().operatedTables().forEach(
-            table -> cleanupEpochs.get(table)
-                .add(entry.getTrxLogIndex()));
+        addCleanupEntry(entry, cleanupEpochs);
       }
     } else {
-      entry.getResponse().operatedTables().forEach(
-          table -> cleanupEpochs.get(table)
-              .add(entry.getTrxLogIndex()));
+      addCleanupEntry(entry, cleanupEpochs);
     }
   }
+
+
+  private void addCleanupEntry(DoubleBufferEntry entry, Map<String,
+      List<Long>> cleanupEpochs) {
+    Class<? extends OMClientResponse > responseClass =
+        entry.getResponse().getClass();
+    CleanupTableInfo cleanupTableInfo =
+        responseClass.getAnnotation(CleanupTableInfo.class);
+    if (cleanupTableInfo != null) {
+      String[] cleanupTables = cleanupTableInfo.cleanupTables();
+      for (String table : cleanupTables) {
+        cleanupEpochs.get(table)
+            .add(entry.getTrxLogIndex());
+      }
+    } else {
+      // This is to catch early errors, when an new response class missed to
+      // add CleanupTableInfo annotation.
+      throw new RuntimeException("CleanupTableInfo Annotation is missing " +
+          "for" + responseClass);
+    }
+  }
+
+
 
   private void cleanupCache(Map<String, List<Long>> cleanupEpochs) {
     cleanupEpochs.forEach((tableName, epochs) -> {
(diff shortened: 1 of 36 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
```diff
@@ -286,30 +301,45 @@ private void flushTransactions() {
...
+    if (opName.toLowerCase().contains(VOLUME) ||
+        opName.toLowerCase().contains(BUCKET)) {
+      if (DeleteBucket.name().equals(opName)
+          || DeleteVolume.name().equals(opName)) {
+        entry.getResponse().operatedTables().forEach(
+            table -> cleanupEpochs.get(table)
+                .add(entry.getTrxLogIndex()));
+      }
+    } else {
+      entry.getResponse().operatedTables().forEach(
```
xiaoyuyao, line 330: What are the difference is the operation is not volume bucket delete? It seems the txIndex are added the same anyway.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/bucket/OMBucketCreateResponse.java
```diff
@@ -30,11 +32,16 @@
 import javax.annotation.Nullable;
 import javax.annotation.Nonnull;
 
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.BUCKET_TABLE;
+
 /**
  * Response for CreateBucket request.
  */
 public final class OMBucketCreateResponse extends OMClientResponse {
```
xiaoyuyao, line 40: As an alternative, you might achieve the same by adding a class annotation with table name string.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
bharatviswa504: When operattion is of type volume/bucket entry should be added only when deletevolume/deletebucket.  As volume/bucket is full cache, and cache should be cleaned up only for delete operations.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/bucket/OMBucketCreateResponse.java
bharatviswa504: Done.

bharatviswa504:
Addressed review comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/bucket/OMBucketDeleteResponse.java
```diff
@@ -19,20 +19,29 @@
...
 
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.BUCKET_TABLE;
+
 /**
  * Response for DeleteBucket request.
  */
+@CleanupTableInfo(cleanupTables = {BUCKET_TABLE})
 public final class OMBucketDeleteResponse extends OMClientResponse {
 
+  private static final List<String> OPERATED_TABLES =
```
xiaoyuyao, line 42: Should this be removed after we take the annotation?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/bucket/OMBucketSetPropertyResponse.java
```diff
@@ -18,22 +18,31 @@
...
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.BUCKET_TABLE;
+
 /**
  * Response for SetBucketProperty request.
  */
+@CleanupTableInfo(cleanupTables = {BUCKET_TABLE})
 public class OMBucketSetPropertyResponse extends OMClientResponse {
   private OmBucketInfo omBucketInfo;
 
+  private static final List<String> OPERATED_TABLES =
```
xiaoyuyao, line 43: Same as above.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
```diff
@@ -353,29 +368,62 @@ private void flushTransactions() {
...
+    if (cleanupTableInfo != null) {
+      String[] cleanupTables = cleanupTableInfo.cleanupTables();
+      for (String table : cleanupTables) {
+        cleanupEpochs.computeIfAbsent(table, list -> new ArrayList<>())
+            .add(entry.getTrxLogIndex());
+      }
+    } else {
+      // This is to catch early errors, when an new response class missed to
+      // add CleanupTableInfo annotation.
+      throw new RuntimeException("CleanupTableInfo Annotation is missing " +
```
xiaoyuyao, line 415: Can we add a unit test on response class that misses annotation?

bharatviswa504: TestCleanupTableInfo test is added for this purpose.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/bucket/OMBucketDeleteResponse.java
bharatviswa504: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/bucket/OMBucketSetPropertyResponse.java
bharatviswa504: Done

bharatviswa504:
Thanks for the review @xiaoyuyao 
Addressed review comments.

bharatviswa504:

PR #944, opened 2020-05-19
HDDS-3609. Avoid to use Hadoop3.x IOUtils in Ozone Client

elek:

To support Hadoop 2.x it's better to avoid to use Hadoop 3.x specific utility methods from IOUtils

https://issues.apache.org/jira/browse/HDDS-3609

## How was this patch tested?

With "push & hope" method on my fork :-P

https://github.com/elek/hadoop-ozone/actions/runs/108204343


elek:
Thanks the feedback @adoroszlai 

I started to write my first answer yesterday: "good idea, let me check the version of IO commons used by hadoop 2.x"

But I started to think. Do we really need one additional dependency? IOUtils is better than guava but we need to shade all the dependencies on the client side. I agree with the server side, let's use common, but client side, It can be better to maintain our own logic.

But *currently* we have IOUtils on the classpath anyway. I updated the patch based on the suggestions, but in the future we can revisit that approach (using IOUtils) when we agree to remove IOUtils from the dependencies.

elek:
It turned out that using IOUtils is unsafe. With using Hadoop 2.7.7:

With the final patch I got:

```
nm_1        | java.lang.NoSuchMethodError: org.apache.commons.io.IOUtils.closeQuietly([Ljava/io/Closeable;)V
```
I will recreate the original patch (forking the close utility) if no objections.

PR #945, opened 2020-05-19
HDDS-3622. Implement rocksdb tool to parse scm db

sadanand48:
This tool parses rocksdb file for SCM and dumps specified table data. Also there is a list command which parses any db file and lists the columnFamilies(tables).

https://issues.apache.org/jira/browse/HDDS-3622

## How was this patch tested?
Manually Tested.

Here are the commands:
1.bin/ozone debug ldb --db=/tmp/ozone/data/metadata/scm.db/ list_column_families
default
validCerts
deletedBlocks
pipelines
revokedCerts
containers

2.bin/ozone debug ldb --db=/tmp/ozone/data/metadata/scm.db/ scan --column_family=containers

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ListTables.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ListTables.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.hadoop.ozone.debug;
+
+import org.rocksdb.*;
+import picocli.CommandLine;
+
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.concurrent.Callable;
+
+/**
+ * List all column Families/Tables in db.
+ */
+@CommandLine.Command(
+        name = "list",
+        description = "list all tables in db."
+)
+public class ListTables implements Callable<Void> {
+
+  @CommandLine.ParentCommand
+  private RDBParser parent;
+
+  @Override
+  public Void call() throws Exception {
+    List<byte[]> cfList = getColumnFamilyList(parent.getDbPath());
+    for (byte[] b : cfList) {
+      System.out.println(new String(b, StandardCharsets.UTF_8));
+    }
+    return null;
+  }
+
+  private static List<byte[]> getColumnFamilyList(String dbPath) {
+    List<byte[]> cfList = null;
+    try {
+      cfList = RocksDB.listColumnFamilies(new Options(), dbPath);
+    } catch (RocksDBException e) {
+      e.printStackTrace();
+    }
+    return cfList;
+  }
+}
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ListTables.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ListTables.java
@@ -18,7 +18,9 @@
 
 package org.apache.hadoop.ozone.debug;
 
-import org.rocksdb.*;
+import org.rocksdb.Options;
+import org.rocksdb.RocksDB;
+import org.rocksdb.RocksDBException;
 import picocli.CommandLine;
 
 import java.nio.charset.StandardCharsets;
@@ -46,13 +48,10 @@ public class ListTables implements Callable<Void> {
     return null;
   }
 
-  private static List<byte[]> getColumnFamilyList(String dbPath) {
+  private static List<byte[]> getColumnFamilyList(String dbPath)
+      throws RocksDBException {
     List<byte[]> cfList = null;
-    try {
-      cfList = RocksDB.listColumnFamilies(new Options(), dbPath);
-    } catch (RocksDBException e) {
-      e.printStackTrace();
-    }
+    cfList = RocksDB.listColumnFamilies(new Options(), dbPath);
     return cfList;
   }
 }
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ListTables.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ListTables.java
@@ -20,7 +20,6 @@ package org.apache.hadoop.ozone.debug;
 
 import org.rocksdb.Options;
 import org.rocksdb.RocksDB;
-import org.rocksdb.RocksDBException;
 import picocli.CommandLine;
 
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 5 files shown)


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ListTables.java
```diff
@@ -0,0 +1,58 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.hadoop.ozone.debug;
+
+import org.rocksdb.*;
```
mukul1987, line 21: Please expand wildcard imports.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ListTables.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+    return null;
+  }
+
+  private static List<byte[]> getColumnFamilyList(String dbPath) {
+    List<byte[]> cfList = null;
+    try {
+      cfList = RocksDB.listColumnFamilies(new Options(), dbPath);
+    } catch (RocksDBException e) {
+      e.printStackTrace();
```
mukul1987, line 54: after extending GenericCli, we will not need this change.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/SCMDBParser.java
```diff
@@ -0,0 +1,120 @@
...
+  private RDBParser parent;
+
+
+  private static void displayTable(RocksDB rocksDB,
+        DBColumnFamilyDefinition dbColumnFamilyDefinition,
+        List<ColumnFamilyHandle> list) throws IOException {
+    ColumnFamilyHandle columnFamilyHandle = getColumnFamilyHandle(
+            dbColumnFamilyDefinition.getTableName()
+                    .getBytes(StandardCharsets.UTF_8), list);
+    RocksIterator iterator = rocksDB.newIterator(columnFamilyHandle);
```
mukul1987, line 55: lets check for null here as if the table name doesn't match, we can throw NPE.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/RDBParser.java
```diff
@@ -0,0 +1,74 @@
...
+    return dbPath;
+  }
+
+  static {
+    columnFamilyMap = constructColumnFamilyMap();
+  }
+
+  private static HashMap<String, DBColumnFamilyDefinition>
+      constructColumnFamilyMap() {
+    columnFamilyMap = new HashMap<String, DBColumnFamilyDefinition>();
```
mukul1987, line 61: This map should be constructed from SCM DB definition.

mukul1987: This can be replaced with getColumnFamilies, and then a table created out of that.

avijayanhwx:
@sadanand48 Thanks for working on this useful tool. 

There is an Ozone component called 'Recon' which is supposed to act as the "management" HQ for Ozone. Currently, it has access to SCM and OM metadata in the same format as they are in the respective components. The vision for Recon is to know everything on the cluster and understand what decisions are being taken by SCM, OM & DN. Please take a look at HDDS-1084 and HDDS-1996 for more information.

Since Recon already has the SCM DB, the same tool should also work against that. Can we add a flag to this CLI like --recon which allows the tool to read from the Recon server as well? That will be useful in understanding Recon's state.

swagle:
> @sadanand48 Thanks for working on this useful tool.
> 
> There is an Ozone component called 'Recon' which is supposed to act as the "management" HQ for Ozone. Currently, it has access to SCM and OM metadata in the same format as they are in the respective components. The vision for Recon is to know what is happening in Ozone and understand what decisions are being taken by SCM, OM & DN. Please take a look at [HDDS-1084](https://issues.apache.org/jira/browse/HDDS-1084) and [HDDS-1996](https://issues.apache.org/jira/browse/HDDS-1996) for more information.
> 
> Since Recon already has the SCM DB, the same tool should also work against that. Can we add a flag to this CLI like --recon which allows the tool to read from the Recon server as well? That will be useful in understanding Recon's state.

Why not add the code to a package in Recon?

elek:
Thank you @sadanand48, I am very happy to have this feature.

> Can we add a flag to this CLI like --recon which allows the tool to read from the Recon server as well?

I think it's a more generic question. How can I choose the appropriate db definition based on a given database / table? There can be multiple options:

 1. If the column families are unique, we can try to use just the name
 2. If not, we can try to identify the required db definition based on the name of the directory (scm.db, ...)

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ListTables.java
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  public Void call() throws Exception {
+    List<byte[]> cfList = getColumnFamilyList(parent.getDbPath());
+    for (byte[] b : cfList) {
+      System.out.println(new String(b, StandardCharsets.UTF_8));
+    }
+    return null;
+  }
+
+  private static List<byte[]> getColumnFamilyList(String dbPath)
```
elek, line 51: Nit1: I am not sure if we need a separated method just to call this function. 

```java
List<byte[]> columnFamilies = RocksDB.listColumnFamilies(new Options(), parent.getDbPath);
```

Nit2: I tend to agree with Uncle Bob, that we don't need to include the type of a variable in the name of the variable:

> Hungarian Notation was considered to be pretty important back in the Windows C API, when everything was an integer handle or a long pointer or a void pointer, or one of several implementations of “string” (with different uses and attributes). The compiler did not check types in those days, so the programmers needed a crutch to help them remember the types.

> In modern languages we have much richer type systems, and the compilers remember and enforce the types. What’s more, there is a trend toward smaller classes and shorter functions so that people can usually see the point of declaration of each variable they’re using.

> Java programmers don’t need type encoding. Objects are strongly typed, and editing environments have advanced such that they detect a type error long before you can run a compile! So nowadays HN and other forms of type encoding are simply impediments. They make it harder to change the name or type of a variable, function, or class. They make it harder to read the code. And they create the possibility that the encoding system will mislead the reader.

(In: Clean Code / Chapter 2. Meaningful name)

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/SCMDBParser.java
```diff
@@ -0,0 +1,123 @@
...
+  public Void call() throws Exception {
+    List<ColumnFamilyDescriptor> cfs = new ArrayList<>();
+    final List<ColumnFamilyHandle> columnFamilyHandleList =
+            new ArrayList<>();
+    List<byte[]> cfList = null;
+    try {
+      cfList = RocksDB.listColumnFamilies(new Options(),
+                    parent.getDbPath());
+    } catch (Exception e) {
+      e.printStackTrace();
```
elek, line 94: If there is an exception, we are in trouble anyway. I would throw the exception without printing out the problem and GenericCli will handle all the remaining parts.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/SCMDBParser.java
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Parser for scm.db file.
+ */
+@CommandLine.Command(
+        name = "scmdbparser",
+        description = "Parse specified metadataTable"
+)
+public class SCMDBParser implements Callable<Void> {
+
+  @CommandLine.Option(names = {"-table"},
```
elek, line 41: Nit: Can you please use `--table`. It's more standard (or use `-t` and `--table` as an alias)

sadanand48:
Thanks @elek for the comments. I have made the changes you suggested. Also, i am now choosing the DBDefinition according to the directory name. In this patch , i have done it only for scm. Will extend it for om in a follow up jira. 


> 2. If not, we can try to identify the required db definition based on the name of the directory (scm.db, ...)

sadanand48:
@elek Can you please review this ? I have made the changes you suggested.

PR #946, opened 2020-05-19
HDDS-3623. Implement getReadCopy in Table.

bharatviswa504:

Introduce a getReadCopy in table method.

Fixed only OMBucketCreateRequest and OMBucketDeleteRequest.
Will file a new Jira to see where all this API can be used in OM.

https://issues.apache.org/jira/browse/HDDS-3623

## How was this patch tested?

Tested it on pseudo deployed cluster.


bharatviswa504:
Thank You @dineshchitlangia for the review.

PR #948, opened 2020-05-20
HDDS-3625. BadRequestException interception doesn't take effect in s3g.

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-3625  

@Provider is missed in the previous fix， which causes the function is not recognized by the jax-ws infra。


bharatviswa504:
Thank You @ChenSammi for the catch and fix this issue.

PR #951, opened 2020-05-21
HDDS-3635. Archive jacoco coverage files for unit/integration tests

elek:

Earlier we configured jacoco maven plugin to calculate the code coverage. But the jacoco.exec files (which contain the coverage information) are not stored in the artifacts created by the github actions.

We can add it easily.

https://issues.apache.org/jira/browse/HDDS-3635

## How was this patch tested?

Branch is pushed to my fork and artifacts are downloaded after the build.

Full coverage is reported based on the partial data (unit + it-*) locally...


elek:
Merged. Thanks the review @vivekratnavel

PR #953, opened 2020-05-21
HDDS-3629. Fix TestBlockOutputStreamWithFailures#test2DatanodesFailure.

bshashikant:
Added a wait time for leader election to complete for atleast one pipeline before running a test.
Removed harcoded checks for streamEntries size check as it will vary in case any retry happens during write.

https://issues.apache.org/jira/browse/HDDS-3629

## How was this patch tested?
Verified running the unit test


bshashikant:
Thanks @smengcl for the review. I have committed this.

PR #954, opened 2020-05-21
HDDS-3638. Add a cat command to show the text of a file in the Ozone server

maobaolong:

When we deploy a hdfs cluster, our devops usually want to put a file into the cluster and cat the result to ensure anything is ok, so they want this command of ozone too.

Some time, we only want to print the content of the file in Ozone, we don't need to download it now.

https://issues.apache.org/jira/browse/HDDS-3638

## How was this patch tested?

```bash
bin/ozone sh key cat /myvol/mybucket/NOTIC.txt
```

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CatKeyHandler.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CatKeyHandler.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.hadoop.ozone.shell.keys;
+
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_CHUNK_SIZE_DEFAULT;
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_CHUNK_SIZE_KEY;
+
+import org.apache.hadoop.conf.StorageUnit;
+import org.apache.hadoop.io.IOUtils;
+import org.apache.hadoop.ozone.client.OzoneBucket;
+import org.apache.hadoop.ozone.client.OzoneClient;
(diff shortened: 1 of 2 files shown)


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CatKeyHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import picocli.CommandLine.Command;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Cat an existing key.
+ */
+@Command(name = "cat",
+    description = "Cat a specific key from ozone server")
```
adoroszlai, line 40:
```suggestion
    description = "Copies a specific Ozone key to standard output")
```

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CatKeyHandler.java
```diff
@@ -0,0 +1,60 @@
...
+    String bucketName = address.getBucketName();
+    String keyName = address.getKeyName();
+
+    int chunkSize = (int) getConf().getStorageSize(OZONE_SCM_CHUNK_SIZE_KEY,
+        OZONE_SCM_CHUNK_SIZE_DEFAULT, StorageUnit.BYTES);
+
+    OzoneVolume vol = client.getObjectStore().getVolume(volumeName);
+    OzoneBucket bucket = vol.getBucket(bucketName);
+    try (InputStream input = bucket.readKey(keyName)) {
+      IOUtils.copyBytes(input, System.out, chunkSize);
```
xiaoyuyao, line 57: Should we consider a small copy buffer size like (4k) instead of default OZONE_SCM_CHUNK_SIZE_DEFAULT(4M) for better memory efficiency?

maobaolong: Done.

maobaolong:
@adoroszlai  @xiaoyuyao  Thanks for your review and help, i've addressed your comments, PTAL.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CatKeyHandler.java
```diff
@@ -0,0 +1,59 @@
...
+  @Override
+  protected void execute(OzoneClient client, OzoneAddress address)
+      throws IOException, OzoneClientException {
+
+    String volumeName = address.getVolumeName();
+    String bucketName = address.getBucketName();
+    String keyName = address.getKeyName();
+
+    int chunkSize = (int) getConf().getStorageSize(OZONE_SCM_CHUNK_SIZE_KEY,
+        "4KB", StorageUnit.BYTES);
```
adoroszlai, line 51: This will use `4KB` buffer only if chunk size is not configured, which I don't think can happen.  I guess @xiaoyuyao's intention was to always use 4KB buffer, regardless of chunk size setting.

maobaolong: @adoroszlai Got it, done.

maobaolong:
@adoroszlai Thank you for the clarification, i use a const 4096 as chunk size, i think i have address @xiaoyuyao 's comments now, sorry for the misunderstand.

xiaoyuyao:
LGTM, +1 pending CI.

PR #956, opened 2020-05-22
HDDS-2720. Ozone Failure injection Service

prashantpogde:

Adding Fault Injection Service in Ozone:tools/FaultInjectionService directory

https://issues.apache.org/jira/browse/HDDS-2720

## How was this patch tested?

This is an independent tool which will be used to do fault injection testing for Ozone. Its build/test process is independent of Ozone and should not impact Ozone in any way.
For using this tool, please follow README in tools/FaultInjectionService.


tools/FaultInjectionService/AUTHORS
```diff
@@ -0,0 +1,12 @@
+Current Maintainer
```
elek, line 1: I am not sure if we need this file (but I can be convinced). Until now we didn't follow this pattern. There are no maintainers of SCM / Freon / OM. The community itself is the maintainers but we can check the git history for first contact.

This file is also requires additional maintenance. I would suggest to remove it.

tools/FaultInjectionService/README.md
```diff
@@ -0,0 +1,53 @@
+NoiseInjector
+==============
+
+About
+------
+TBD
```
elek, line 6: Just one sentence please

tools/FaultInjectionService/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Development Status                                                              
+------------------
+TBD
```
elek, line 10: One word, please ;-)

tools/FaultInjectionService/AUTHORS
prashantpogde: @elek Thank you for reviewing it. 
- Yes I will remove AUTHORS file and change the directory name to lower case.
- Its supported only on linux and some dependencies require building them from the source. I will add an automated build script as next set of enhancement and create a separate PR.

tools/FaultInjectionService/README.md
prashantpogde: yup, will change.

tools/FaultInjectionService/README.md
prashantpogde: yup, will change.

elek:
Thanks the update @prashantpogde 

 1. Can you please remove `tools/fault-injection-service/LICENSE`? As I wrote this is unnecessary as the whole project is distributed under the top level `LICENSE` file.

 2. License header check doesn't work for this dir (as I see). Can you please double check if ASF headers are added to everywhere? Seems to be missing from the `Test*.cc` files 

 3. I am fine to create the build script in the next PR. After 1-2, I will merge this.

prashantpogde:
@elek Thank you for the review. I have made the suggested changes. Please take a look.

tools/fault-injection-service/README.md
```diff
@@ -0,0 +1,56 @@
+Fault Injection Service
```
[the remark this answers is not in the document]
prashantpogde: I noticed that ASF license header was not there in Ozone README.md,  Should I put it in this one ?

elek:
I just found this one on the [faq page](https://apache.org/legal/src-headers.html)

> Other files may make sense to have no license header. Three examples are: ... Short informational text files; for example README, INSTALL files

It seems that we don't need LICENSE header for that specific README.

PR #957, opened 2020-05-22
HDDS-3477. Disable partial chunk write during flush() call in ozone client by default.

captainzmc:

Currently, Ozone client flushes the partial chunks as well during flush() call by default.

https://github.com/apache/hadoop-ozone/pull/716 proposes to add a configuration to disallow partial chunk flush during flush() call. This Jira aims to enable the config on by default to mimic the default hdfs flush() behaviour and fix any failing unit tests associated with the change.

https://issues.apache.org/jira/browse/HDDS-3477

## How was this patch tested?

The affected UT has been modified.


captainzmc:
Hi @bshashikant, Could you please help to review this PR?

captainzmc:
Hi @bshashikant, Could you please help to review this PR?

captainzmc:
> Thanks @captainzmc for working on this. The changes look ok. However, i would prefer to not fix the actual unit tests but set the flush config to false in these unit tests. We can rewrite the same test with the flush configuration set to true . This way, unit tests will have coverage for different behaviours of flush with/without the configuration turned on. What do you think?

 Agree, that will make the test more complete.
 I will set flush config to false in the original UT. And rewrite the same test with the flush configuration set to true .

captainzmc:
Had set flush config to false in the original UT. And rewrite the same test with the flush configuration set to true. cc @bshashikant

bshashikant:
Thanks @captainzmc for the contribution. I have committed this to master branch.

PR #959, opened 2020-05-22
HDDS-3186. Introduce generic SCMRatisRequest and SCMRatisResponse.

nandakumar131:

This PR will introduce generic `SCMRatisRequest` and `SCMRatisResponse` which will be used by all the Ratis operations inside SCM. We also have a generic `StateMachine` which will dispatch the request to registered handlers.

HDDS-3186

## How was this patch tested?
This patch is not tested properly.
The following jiras are created for adding unit tests to test the changes introduced in this PR
HDDS-3650, HDDS-3651 and HDDS-3652

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
@@ -1,0 +1,282 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership.  The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.hadoop.hdds.scm.container;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ContainerInfoProto;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.LifeCycleState;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.LifeCycleEvent;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationFactor;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationType;
+import org.apache.hadoop.hdds.scm.ha.SCMHAManager;
+import org.apache.hadoop.hdds.scm.pipeline.Pipeline;
+import org.apache.hadoop.hdds.scm.pipeline.PipelineManager;
+import org.apache.hadoop.hdds.utils.db.Table;
+import org.apache.hadoop.util.Time;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * TODO: Add javadoc.
+ */
+public class ContainerManagerImpl implements ContainerManagerV2 {
+
+  /**
+   *
+   */
+  private static final Logger LOG = LoggerFactory.getLogger(
+      ContainerManagerImpl.class);
+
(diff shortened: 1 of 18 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.ratis.server.RaftServerConfigKeys;
+import org.apache.ratis.util.SizeInBytes;
+import org.apache.ratis.util.TimeDuration;
+
+import java.io.File;
+import java.net.InetSocketAddress;
+import java.util.Collections;
+import java.util.concurrent.TimeUnit;
+
+public class RatisUtil {
```
timmylicheng, line 36: can we put this into SCMHAUtils?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServer.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.ratis.protocol.ClientId;
+import org.apache.ratis.protocol.RaftClientReply;
+import org.apache.ratis.protocol.RaftClientRequest;
+import org.apache.ratis.protocol.RaftGroup;
+import org.apache.ratis.protocol.RaftGroupId;
+import org.apache.ratis.protocol.RaftPeer;
+import org.apache.ratis.protocol.RaftPeerId;
+import org.apache.ratis.server.RaftServer;
+
+public class SCMRatisServer {
```
timmylicheng, line 41: I've merged SCMRatisServer and SCMStateMachine into one between /ha and /ratis in https://github.com/timmylicheng/hadoop-ozone/pull/1. We can use /ha as your did here, but we need to combine all methods into one.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
nandakumar131: I have moved this to `ReflectionUtil` class which can be used for all the utility methods related to reflection. I will update the PR soon.

I agree that we should use the existing `SCMHAUtil` instead of creating new `RatisUtil`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServer.java
nandakumar131: I don't think we need to copy all the code/methods from OzoneManagerHA implementation. We can add things to SCMHA related classes whenever required.
It's better not to have code that is not used or needed.

Let's add/update the SCMHA code when needed.

timmylicheng: How about Snapshot in SCMStateMachine?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
```diff
@@ -0,0 +1,282 @@
...
+          .setUsedBytes(0)
+          .setNumberOfKeys(0)
+          .setStateEnterTime(Time.now())
+          .setOwner(owner)
+          .setContainerID(containerID.getId())
+          .setDeleteTransactionId(0)
+          .setReplicationFactor(pipeline.getFactor())
+          .setReplicationType(pipeline.getType())
+          .build();
+      containerStateManager.addContainer(containerInfo);
```
timmylicheng, line 216: Do we need to addContainerToDB here?

timmylicheng:
+1 merge in dev branch and pending on following JIRAs to resolve remaining works.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
timmylicheng: https://issues.apache.org/jira/browse/HDDS-3660 Track it here

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServer.java
timmylicheng: Track the issue in https://issues.apache.org/jira/browse/HDDS-3661.
This is to split the work.

nandakumar131: Let's configure/enable Ratis snapshot after we have some design plan on how to implement the snapshot.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
nandakumar131: I feel it's ok to have `ReflectionUtil`, `RatisUtil`, and `SCMHAUtil` as they serve a different purpose.

nandakumar131:
@timmylicheng thanks for the merge!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
nandakumar131: It is handled inside `ContainerStateManager`.

PR #963, opened 2020-05-23
HDDS-3646. Add a copy command to copy key to a new one.

maobaolong:

A copy command can copy an exist key into a new one, It should support specify the replication factor and type. Cross-bucket copy is not supported.

https://issues.apache.org/jira/browse/HDDS-3646

## How was this patch tested?

```bash
ozone sh key cp <bucket URI> <fromKey> <toKey>
```

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
@@ -21,6 +21,7 @@ import org.apache.hadoop.conf.StorageUnit;
 import org.apache.hadoop.hdds.client.ReplicationFactor;
 import org.apache.hadoop.hdds.client.ReplicationType;
 import org.apache.hadoop.io.IOUtils;
+import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.client.OzoneBucket;
 import org.apache.hadoop.ozone.client.OzoneClient;
 import org.apache.hadoop.ozone.client.OzoneClientException;
@@ -34,6 +35,8 @@ import picocli.CommandLine.Parameters;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.Map;
 
 import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_CHUNK_SIZE_DEFAULT;
 import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_CHUNK_SIZE_KEY;
@@ -89,11 +92,17 @@ public class CopyKeyHandler extends BucketHandler {
               OZONE_REPLICATION_TYPE_DEFAULT));
     }
 
+    Map<String, String> keyMetadata = new HashMap<>();
+    String gdprEnabled = bucket.getMetadata().get(OzoneConsts.GDPR_FLAG);
+    if (Boolean.parseBoolean(gdprEnabled)) {
+      keyMetadata.put(OzoneConsts.GDPR_FLAG, Boolean.TRUE.toString());
+    }
+
     int chunkSize = (int) getConf().getStorageSize(OZONE_SCM_CHUNK_SIZE_KEY,
         OZONE_SCM_CHUNK_SIZE_DEFAULT, StorageUnit.BYTES);
     try (InputStream input = bucket.readKey(fromKey);
          OutputStream output = bucket.createKey(toKey, input.available(),
-             replicationType, replicationFactor, bucket.getMetadata())) {
+             replicationType, replicationFactor, keyMetadata)) {
       IOUtils.copyBytes(input, output, chunkSize);
     }
 
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
@@ -21,7 +21,6 @@ import org.apache.hadoop.conf.StorageUnit;
 import org.apache.hadoop.hdds.client.ReplicationFactor;
 import org.apache.hadoop.hdds.client.ReplicationType;
 import org.apache.hadoop.io.IOUtils;
-import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.client.OzoneBucket;
 import org.apache.hadoop.ozone.client.OzoneClient;
 import org.apache.hadoop.ozone.client.OzoneClientException;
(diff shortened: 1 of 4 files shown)


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CpKeyHandler.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_CHUNK_SIZE_DEFAULT;
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_CHUNK_SIZE_KEY;
+import static org.apache.hadoop.ozone.OzoneConfigKeys.*;
+import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_REPLICATION_TYPE_DEFAULT;
```
dineshchitlangia, line 44: Redundant import statements.
Avoid using * and instead import the specific classes/members needed.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CpKeyHandler.java
```diff
@@ -0,0 +1,110 @@
...
+          getConf().get(OZONE_REPLICATION_TYPE,
+              OZONE_REPLICATION_TYPE_DEFAULT));
+    }
+
+    Map<String, String> keyMetadata = new HashMap<>();
+    String gdprEnabled = bucket.getMetadata().get(OzoneConsts.GDPR_FLAG);
+    if (Boolean.parseBoolean(gdprEnabled)) {
+      keyMetadata.put(OzoneConsts.GDPR_FLAG, Boolean.TRUE.toString());
+    }
+
```
dineshchitlangia, line 97: Since we are assuming that the copy command is to copy a key in within the same bucket, this GDPR piece of code won't cause any problem.
Can we pls add comments in the javadoc to state the copy will only for copying keys within the same bucket?
This will be a safeguard for users attempting to doing something like:
`ozone sh key cp /vol1/bucket1/key1 /vol1/bucket5/newkey`

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CpKeyHandler.java
```diff
@@ -0,0 +1,110 @@
...
+    int chunkSize = (int) getConf().getStorageSize(OZONE_SCM_CHUNK_SIZE_KEY,
+        OZONE_SCM_CHUNK_SIZE_DEFAULT, StorageUnit.BYTES);
+    try (InputStream input = bucket.readKey(fromKey);
+         OutputStream output = bucket.createKey(toKey, 0,
```
dineshchitlangia, line 101: Why is the size hardcoded as 0 ?

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CpKeyHandler.java
maobaolong: Done

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CpKeyHandler.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_CHUNK_SIZE_KEY;
+import static org.apache.hadoop.ozone.OzoneConfigKeys.*;
+import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_REPLICATION_TYPE_DEFAULT;
+
+/**
+ * Copy an existing key to another one.
+ */
+@Command(name = "cp",
+    description = "copies an existing key to another one")
+public class CpKeyHandler extends BucketHandler {
```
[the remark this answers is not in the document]
maobaolong: Done.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CpKeyHandler.java
maobaolong: Done.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CpKeyHandler.java
maobaolong: Sorry, i replace it to the input stream size.

maobaolong:
@dineshchitlangia Thank you for your review, i updated my PR and jira ticket description, PTAL.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
```diff
@@ -89,17 +89,11 @@ protected void execute(OzoneClient client, OzoneAddress address)
...
-      keyMetadata.put(OzoneConsts.GDPR_FLAG, Boolean.TRUE.toString());
-    }
-
     int chunkSize = (int) getConf().getStorageSize(OZONE_SCM_CHUNK_SIZE_KEY,
         OZONE_SCM_CHUNK_SIZE_DEFAULT, StorageUnit.BYTES);
     try (InputStream input = bucket.readKey(fromKey);
-         OutputStream output = bucket.createKey(toKey, 0,
-             replicationType, replicationFactor, keyMetadata)) {
+         OutputStream output = bucket.createKey(toKey, input.available(),
+             replicationType, replicationFactor, bucket.getMetadata())) {
```
dineshchitlangia: Your previous code was fine. The bucket metadata object may have properties that may not be needed for the key. So, if we copy the bucket metadata to be saved as key metadata, that is basically the piece of data which is mostly irrelevant for the key.

So you can replace `bucket.getMetadata()` with `keyMetadata`.

maobaolong:
@dineshchitlangia  Thank you for review again, if take the keyMetadata back now, and create a jira ticket to track the document of this command. About the failed tests, the logs show some timeout and permission deny... not sure it is relates to this PR, lets have a look wait the new CI checks finish.

dineshchitlangia:
Yes, it should be fine to log new jira for documentation work.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
```diff
@@ -89,11 +92,17 @@ protected void execute(OzoneClient client, OzoneAddress address)
               OZONE_REPLICATION_TYPE_DEFAULT));
     }
 
+    Map<String, String> keyMetadata = new HashMap<>();
+    String gdprEnabled = bucket.getMetadata().get(OzoneConsts.GDPR_FLAG);
+    if (Boolean.parseBoolean(gdprEnabled)) {
+      keyMetadata.put(OzoneConsts.GDPR_FLAG, Boolean.TRUE.toString());
+    }
```
adoroszlai, line 99: One more thing: I think instead of building key metadata from scratch, `cp` should copy actual metadata of the existing key, which it can get via `OzoneKeyDetails` (except `GDPR_SECRET` and `GDPR_ALGORITHM` should be removed).  This would ensure that any user-supplied custom metadata is transferred, as well as reduce the need for future changes in case new default metadata is added in `PutKeyHandler`.

maobaolong: Thank you for your advice, done.

maobaolong:
@dineshchitlangia Thank you for your review, The documents will be updated, tracked by https://issues.apache.org/jira/browse/HDDS-3653

maobaolong:
@adoroszlai Thanks for your suggestion, i addressed your comments and wait for the CI build.

maobaolong:
@dineshchitlangia Thank you for merge this PR, @adoroszlai thank you for your reviews.

PR #973, opened 2020-05-27
HDDS-3663. Display uuid in recon datanode page to distinguish each datanode

maobaolong:

Without this PR, i don't distinguish each datanode when some datanode from same node.

https://issues.apache.org/jira/browse/HDDS-3663

## How was this patch tested?

image


maobaolong:
@vivekratnavel Thank you for your review.

PR #976, opened 2020-05-28
HDDS-3672. Ozone fs failed to list intermediate directory.

ChenSammi:

The root cause is replication factor and replication type are required fieldsin KeyInfo proto. 

While the way to create the OMKeyInfo object for intermediate directory doesn't provide all required fields.  

https://issues.apache.org/jira/browse/HDDS-3672

## How was this patch tested?

Unit test is updated + manual confirmation:

```
bin/ozone sh volume create /hadoop
bin/ozone sh bucket create /hadoop/ozone
bin/ozone sh key put /hadoop/ozone/benchmarks/TestDFSIO/dir1/test README.md
bin/ozone fs -ls o3fs://ozone.hadoop/
```

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -2037,11 +2037,11 @@ public class KeyManagerImpl implements KeyManager {
                 // if entry is a directory
                 if (!deletedKeySet.contains(entryInDb)) {
                   if (!entryKeyName.equals(immediateChild)) {
-                    OmKeyInfo fakeDirEntry = new OmKeyInfo.Builder()
-                        .setVolumeName(omKeyInfo.getVolumeName())
-                        .setBucketName(omKeyInfo.getBucketName())
-                        .setKeyName(immediateChild)
-                        .build();
+                    OmKeyInfo fakeDirEntry = createDirectoryKey(
+                        omKeyInfo.getVolumeName(),
+                        omKeyInfo.getBucketName(),
+                        immediateChild,
+                        omKeyInfo.getAcls());
                     cacheKeyMap.put(entryInDb,
                         new OzoneFileStatus(fakeDirEntry, scmBlockSize, true));
                   } else {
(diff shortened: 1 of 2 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
```diff
@@ -465,11 +464,11 @@ public static ObjectName registerWithJmxProperties(
...
   /**
-   * Get the current UTC time in milliseconds.
-   * @return the current UTC time in milliseconds.
+   * Get the current time in milliseconds.
+   * @return the current time in milliseconds.
    */
-  public static long getUtcTime() {
-    return Calendar.getInstance(UTC_ZONE).getTimeInMillis();
+  public static long getTime() {
+    return System.currentTimeMillis();
```
rakeshadr, line 471: Thanks @ChenSammi for the analysis and the changes.

It looks like the "current time" change is not related to this, please let me know your thoughts.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2037,11 +2037,11 @@ private void listStatusFindKeyInTableCache(
                 // if entry is a directory
                 if (!deletedKeySet.contains(entryInDb)) {
                   if (!entryKeyName.equals(immediateChild)) {
-                    OmKeyInfo fakeDirEntry = new OmKeyInfo.Builder()
-                        .setVolumeName(omKeyInfo.getVolumeName())
-                        .setBucketName(omKeyInfo.getBucketName())
-                        .setKeyName(immediateChild)
-                        .build();
+                    OmKeyInfo fakeDirEntry = createDirectoryKey(
```
rakeshadr, line 2040: It looks to me an opportunity to add a unit test, can we? Thanks!

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
ChenSammi: It's part of another PR.  Will upload a clean patch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
ChenSammi: Sure.

ChenSammi: I find an easy way to reproduce the issue.  Use "key put" to upload an object with a path like object name, for example "object-dir/object-name", then use "fs -ls" to list the bucket. 
Will try to add a UT too.

ChenSammi:
Thanks @elek  and @rakeshadr  for the review.

PR #977, opened 2020-05-28
HDDS-3673. Version information not printed correctly in SCM WebUI.

flirmnave:

Add property in pom so that can get correct version information.

https://issues.apache.org/jira/browse/HDDS-3673

## How was this patch tested?

Run on local docker-compose.
Screenshot from 2020-05-28 15-42-48

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,7 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <hdds.version>${ozone.version}</hdds.version>
     <ozone.version>0.6.0-SNAPSHOT</ozone.version>
     <ozone.release>Denali</ozone.release>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -75,6 +75,7 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <hdds.version>${ozone.version}</hdds.version>
     <ozone.version>0.6.0-SNAPSHOT</ozone.version>
     <ozone.release>Denali</ozone.release>
+    <declared.hdds.version>${ozone.version}</declared.hdds.version>
```
adoroszlai, line 78: Since the property`hdds.version` is also defined, I think this one should reference it instead of `ozone.version`.

```suggestion
    <declared.hdds.version>${hdds.version}</declared.hdds.version>
```

flirmnave:
Thanks @adoroszlai for reviewing and I change it.

PR #980, opened 2020-05-28
HDDS-3196 New PipelineManager interface to persist to RatisServer.

timmylicheng:
Construct new PipelineManager and PipelineStateManager interface for SCM.

https://issues.apache.org/jira/browse/HDDS-3196

## How was this patch tested?
TODO in https://issues.apache.org/jira/browse/HDDS-3679

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
@@ -33,7 +33,6 @@ import org.apache.hadoop.hdds.scm.safemode.SCMSafeModeManager;
 import org.apache.hadoop.hdds.server.events.EventPublisher;
 import org.apache.hadoop.hdds.utils.Scheduler;
 import org.apache.hadoop.hdds.utils.db.Table;
-import org.apache.hadoop.hdds.utils.db.TableIterator;
 import org.apache.hadoop.metrics2.util.MBeans;
 import org.apache.hadoop.util.Time;
 import org.slf4j.Logger;
@@ -68,7 +67,6 @@ public class PipelineManagerV2Impl implements PipelineManager {
   private PipelineStateManagerV2 stateManager;
   private Scheduler scheduler;
   private BackgroundPipelineCreator backgroundPipelineCreator;
-  private final NodeManager nodeManager;
   private final ConfigurationSource conf;
   // Pipeline Manager MXBean
   private ObjectName pmInfoBean;
@@ -82,12 +80,10 @@ public class PipelineManagerV2Impl implements PipelineManager {
   public PipelineManagerV2Impl(ConfigurationSource conf,
                                NodeManager nodeManager,
                                PipelineStateManagerV2 pipelineStateManager,
-                               PipelineFactory pipelineFactory)
-      throws IOException {
+                               PipelineFactory pipelineFactory) {
     this.lock = new ReentrantReadWriteLock();
     this.pipelineFactory = pipelineFactory;
     this.stateManager = pipelineStateManager;
-    this.nodeManager = nodeManager;
     this.conf = conf;
     this.pmInfoBean = MBeans.register("SCMPipelineManager",
         "SCMPipelineManagerInfo", this);
@@ -102,7 +98,6 @@ public class PipelineManagerV2Impl implements PipelineManager {
     // Pipeline creation is only allowed after the safemode prechecks have
     // passed, eg sufficient nodes have registered.
     this.pipelineCreationAllowed = new AtomicBoolean(!this.isInSafeMode.get());
-    initializePipelineState();
   }
 
   public static PipelineManager newPipelineManager(
@@ -112,7 +107,9 @@ public class PipelineManagerV2Impl implements PipelineManager {
     // Create PipelineStateManager
     PipelineStateManagerV2 stateManager = PipelineStateManagerV2Impl
         .newBuilder().setPipelineStore(pipelineStore)
-        .setRatisServer(scmhaManager.getRatisServer()).build();
+        .setRatisServer(scmhaManager.getRatisServer())
+        .setNodeManager(nodeManager)
+        .build();
 
     // Create PipelineManager
     PipelineManagerV2Impl pipelineManager = new PipelineManagerV2Impl(conf,
@@ -129,20 +126,6 @@ public class PipelineManagerV2Impl implements PipelineManager {
     return pipelineManager;
   }
 
-  protected void initializePipelineState() throws IOException {
-    if (stateManager.isPipelineStoreEmpty()) {
-      LOG.info("No pipeline exists in current db");
-      return;
-    }
-    TableIterator<PipelineID, ? extends Table.KeyValue<PipelineID, Pipeline>>
-        iterator = stateManager.getPipelineStoreIterator();
-    while (iterator.hasNext()) {
-      Pipeline pipeline = iterator.next().getValue();
-      stateManager.addPipeline(pipeline.getProtobufMessage());
-      nodeManager.addPipeline(pipeline);
-    }
-  }
-
   @Override
   public Pipeline createPipeline(ReplicationType type,
                                  ReplicationFactor factor) throws IOException {
@@ -156,7 +139,6 @@ public class PipelineManagerV2Impl implements PipelineManager {
     try {
       Pipeline pipeline = pipelineFactory.create(type, factor);
       stateManager.addPipeline(pipeline.getProtobufMessage());
-      nodeManager.addPipeline(pipeline);
       recordMetricsForPipeline(pipeline);
       return pipeline;
     } catch (IOException ex) {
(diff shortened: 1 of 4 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
```diff
@@ -0,0 +1,637 @@
...
+        HddsConfigKeys.HDDS_SCM_SAFEMODE_ENABLED,
+        HddsConfigKeys.HDDS_SCM_SAFEMODE_ENABLED_DEFAULT));
+    // Pipeline creation is only allowed after the safemode prechecks have
+    // passed, eg sufficient nodes have registered.
+    this.pipelineCreationAllowed = new AtomicBoolean(!this.isInSafeMode.get());
+    initializePipelineState();
+  }
+
+  public static PipelineManager newPipelineManager(
+      ConfigurationSource conf, SCMHAManager scmhaManager,
```
xiaoyuyao, line 109: NIT: we can just pass RatisServer instead of SCMHAManager here.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.NavigableSet;
+
+/**
+ * Manages the state of pipelines in SCM.
+ */
+public interface PipelineStateManagerV2 {
+
+  @Replicate
```
xiaoyuyao, line 37: NIT: can we group all the method with @Replicate annotation together either at the begining or end with some java doc?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
timmylicheng: We may have more stuff in SCMHAManager than just SCMRatisServer for PipelineManager use cases. The idea here is to have a manager interface for SCM HA so that we won't worry about passing more things into PipelineManager like configs or other things. @xiaoyuyao

nandakumar131: In the future, we have to do `isLeader` check inside `PipelineManager` for which we might need `SCMHAManager` instance.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2.java
```diff
@@ -0,0 +1,88 @@
...
+  void updatePipelineState(HddsProtos.PipelineID pipelineIDProto,
+                           HddsProtos.PipelineState newState)
+      throws IOException;
+
+  boolean isPipelineStoreEmpty() throws IOException;
+
+  TableIterator<PipelineID, ? extends Table.KeyValue<PipelineID, Pipeline>>
+      getPipelineStoreIterator() throws IOException;
```
nandakumar131, line 103: We should not expose `TableIterator` from `PipelineStateManager`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
```diff
@@ -0,0 +1,637 @@
...
+      LOG.debug("Pipeline creation is not allowed until safe mode prechecks " +
+          "complete");
+      throw new IOException("Pipeline creation is not allowed as safe mode " +
+          "prechecks have not yet passed");
+    }
+    lock.writeLock().lock();
+    try {
+      Pipeline pipeline = pipelineFactory.create(type, factor);
+      stateManager.addPipeline(pipeline.getProtobufMessage());
+      nodeManager.addPipeline(pipeline);
```
nandakumar131, line 159: `stateManager#addPipeline` is replicated via Ratis, which will make the pipeline added to in-memory and to DB on all the SCMs.
`nodeManager#addPipeline`, since executed outside of `PipelineStateManager` will only be executed on Leader SCM. This will be a problem as the followers will not have this information.

Move `nodeManager#addPipeline` call to `stateManager#addPipeline`.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2.java
timmylicheng: Sure.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
timmylicheng: OK Will do

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2.java
timmylicheng: Updated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2.java
```diff
@@ -0,0 +1,101 @@
...
+  NavigableSet<ContainerID> getContainers(PipelineID pipelineID)
+      throws IOException;
+
+  int getNumberOfContainers(PipelineID pipelineID) throws IOException;
+
+
+  void removeContainerFromPipeline(PipelineID pipelineID,
+                                   ContainerID containerID) throws IOException;
+
+  void initialize() throws IOException;
```
nandakumar131, line 98: We don't need a `initialize` method exposed, this can. be done called inside `PipelineStateManagerImpl` constructor.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2Impl.java
```diff
@@ -0,0 +1,233 @@
...
+  public void addPipeline(HddsProtos.Pipeline pipelineProto)
+      throws IOException {
+    Pipeline pipeline = Pipeline.getFromProtobuf(pipelineProto);
+    if (pipelineStore != null) {
+      pipelineStore.put(pipeline.getId(), pipeline);
+    }
+    pipelineStateMap.addPipeline(pipeline);
+    if (nodeManager != null) {
+      nodeManager.addPipeline(pipeline);
+    }
```
nandakumar131, line 88: We should not do `null` check here. If `pipelineStore` and `nodeManager` are `null` there is a problem/bug.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2Impl.java
```diff
@@ -0,0 +1,233 @@
...
+  public void removePipeline(HddsProtos.PipelineID pipelineIDProto)
+      throws IOException {
+    PipelineID pipelineID = PipelineID.getFromProtobuf(pipelineIDProto);
+    if (pipelineStore != null) {
+      pipelineStore.delete(pipelineID);
+    }
+    Pipeline pipeline = pipelineStateMap.removePipeline(pipelineID);
+    if (nodeManager != null) {
+      nodeManager.removePipeline(pipeline);
+    }
```
nandakumar131, line 164: We should not do `null` check here. If `pipelineStore` and `nodeManager` are `null` there is a problem/bug.

nandakumar131:
@xiaoyuyao, if you're ok with the changes I will go ahead and merge it.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2.java
timmylicheng: Updated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2Impl.java
timmylicheng: Update.

nandakumar131:
@timmylicheng thanks for the contribution.

PR #982, opened 2020-05-28
HDDS-3668. OzoneManager start fails with RocksDB error on downgrade to older version.

avijayanhwx:

OM start fails with RocksDB error when downgrading to older version that does not have all the column families that may have been created  in the newer version.
```
Caused by: org.rocksdb.RocksDBException: You have to open all column families. Column families not opened: transact
ionInfoTable
        at org.rocksdb.RocksDB.open(Native Method)
        at org.rocksdb.RocksDB.open(RocksDB.java:290)
        at org.apache.hadoop.hdds.utils.db.RDBStore.<init>(RDBStore.java:97)
        ... 20 more
```
https://issues.apache.org/jira/browse/HDDS-3668

## How was this patch tested?
Manually tested. Unit test added.


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
```diff
@@ -94,6 +97,14 @@ public RDBStore(File dbFile, DBOptions options,
     this.writeOptions = writeOptions;
 
     try {
+      List<TableConfig> columnFamiliesInDb = getColumnFamiliesInExistingDb();
+      List<TableConfig> extraCf = columnFamiliesInDb.stream().filter(
+          cf -> !families.contains(cf)).collect(Collectors.toList());
+      for (TableConfig family : extraCf) {
+        LOG.info("Adding column family {}", family.getName());
```
bharatviswa504, line 104: One question, can these extra column families can be dropped, if they are not required after upgrade/downgrade.

Because now we removed S3Table in #940. So, once after upgrade, technically we don't require the column family any more.

bharatviswa504, line 104: Can this comment be more clear saying ColumnFamily in DB, but not in the current version something like that?

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/utils/db/TestRDBStore.java
```diff
@@ -348,5 +348,40 @@ public void testGetDBUpdatesSince() throws Exception {
...
+    options.setCreateIfMissing(true);
+    options.setCreateMissingColumnFamilies(true);
+    configSet = new HashSet<>();
+    List<String> familiesMinusOne = families.subList(0, families.size() - 1);
+    for(String name : familiesMinusOne) {
+      TableConfig newConfig = new TableConfig(name, new ColumnFamilyOptions());
+      configSet.add(newConfig);
+    }
+    rdbStore = new RDBStore(rdbStore.getDbLocation(), options, configSet);
+    for (String family : familiesMinusOne) {
```
bharatviswa504, line 378: Question: If column family is missing also, but DBStore will open that column family right currently? So, do we need to check the columnFamily is opened or not also?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
avijayanhwx: Will do.

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/utils/db/TestRDBStore.java
avijayanhwx: Added check for the extra column family.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
avijayanhwx: @bharatviswa504 That is a good point. Based on our offline discussion, this will be handled on a case-by-case basis as outlined in the Ozone upgrades design doc (HDDS-3698).

bharatviswa504:
Thank You @avijayanhwx for the fix. For now this will help in downgrade/upgrade until we have our upgrade story correct. But one thing as discussed offline with this, we will uncover bugs later after upgrade if not properly handled. Can we add some TODO code comments to explain this and need to be removed later once upgrades are handled?

bharatviswa504:
@arp7 @hanishakoneru 
In case you missed this.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
```diff
@@ -94,6 +98,15 @@ public RDBStore(File dbFile, DBOptions options,
     this.writeOptions = writeOptions;
 
     try {
+      List<TableConfig> columnFamiliesInDb = getColumnFamiliesInExistingDb();
+      List<TableConfig> extraCf = columnFamiliesInDb.stream().filter(
+          cf -> !families.contains(cf)).collect(Collectors.toList());
+      if (CollectionUtils.isNotEmpty(extraCf)) {
```
adoroszlai, line 104: Nit: `extraCf` cannot be null, so I would prefer not adding a dependency:

```suggestion
      if (!extraCf.isEmpty()) {
```

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
```diff
@@ -94,6 +98,15 @@ public RDBStore(File dbFile, DBOptions options,
...
+        LOG.info("Found the following extra column families in existing DB : " +
+                "{}", Arrays.toString(extraCf.toArray()));
```
adoroszlai, line 106: Nit: I think `List.toString` is just as good:

```suggestion
                "{}", extraCf);
```

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
```diff
@@ -149,6 +162,26 @@ public RDBStore(File dbFile, DBOptions options,
...
+      throws RocksDBException {
+    List<byte[]> bytes = RocksDB.listColumnFamilies(new Options(),
+        dbLocation.getAbsolutePath());
+    List<TableConfig> columnFamiliesInDb = bytes.stream()
+        .map(cfbytes -> new TableConfig(StringUtils.bytes2String(cfbytes),
+            DBProfile.DISK.getColumnFamilyOptions()))
+        .collect(Collectors.toList());
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("Found column Families in DB : {}",
+          Arrays.toString(columnFamiliesInDb.toArray()));
```
adoroszlai, line 180: Nit: same here regarding arrays:

```suggestion
          columnFamiliesInDb);
```

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
```diff
@@ -24,11 +24,14 @@
 import java.io.IOException;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
```
adoroszlai, line 27: Corresponding to suggestion about list to string:

```suggestion
```

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
```diff
@@ -24,11 +24,14 @@
...
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
+import org.apache.commons.collections.CollectionUtils;
```
adoroszlai, line 34: Corresponding to suggestion about empty list:

```suggestion
```

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBStore.java
```diff
@@ -149,6 +162,26 @@ public RDBStore(File dbFile, DBOptions options,
...
+   * @return List of column families
+   * @throws RocksDBException on Error.
+   */
+  private List<TableConfig> getColumnFamiliesInExistingDb()
+      throws RocksDBException {
+    List<byte[]> bytes = RocksDB.listColumnFamilies(new Options(),
+        dbLocation.getAbsolutePath());
+    List<TableConfig> columnFamiliesInDb = bytes.stream()
+        .map(cfbytes -> new TableConfig(StringUtils.bytes2String(cfbytes),
+            DBProfile.DISK.getColumnFamilyOptions()))
```
adoroszlai, line 176:
```suggestion
            DBStoreBuilder.HDDS_DEFAULT_DB_PROFILE.getColumnFamilyOptions()))
```

avijayanhwx:
> Thank You @avijayanhwx for the fix. For now this will help in downgrade/upgrade until we have our upgrade story correct. But one thing as discussed offline with this, we will uncover bugs later after upgrade if not properly handled. Can we add some TODO code comments to explain this and need to be removed later once upgrades are handled?

Done.

avijayanhwx:
> Thanks @avijayanhwx for working on this. In addition to @bharatviswa504's comments, I would like to propose a few minor code changes.

Thanks for the review @adoroszlai.

bharatviswa504:
Thank You @avijayanhwx for the contribution and @adoroszlai for the review.

PR #988, opened 2020-05-28
HDDS-3681. Recon: Add support to store file size counts in each volum…

vivekratnavel:
…e/bucket

- update utilization schema to include volume and bucket information
- update filesizecount task to persist volume and bucket information
- update utilization endpoint

https://issues.apache.org/jira/browse/HDDS-3681

## How was this patch tested?

Added unit tests and tested the endpoint response from a docker-compose cluster.

### DIFF
--- a/hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/UtilizationSchemaDefinition.java
+++ b/hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/UtilizationSchemaDefinition.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -26,6 +26,7 @@ import javax.sql.DataSource;
 
 import com.google.inject.Singleton;
 
+import org.jooq.DSLContext;
 import org.jooq.impl.DSL;
 import org.jooq.impl.SQLDataType;
 import org.slf4j.Logger;
@@ -44,6 +45,7 @@ public class UtilizationSchemaDefinition implements ReconSchemaDefinition {
       LoggerFactory.getLogger(UtilizationSchemaDefinition.class);
 
   private final DataSource dataSource;
+  private DSLContext dslContext;
 
   public static final String CLUSTER_GROWTH_DAILY_TABLE_NAME =
       "CLUSTER_GROWTH_DAILY";
@@ -59,6 +61,7 @@ public class UtilizationSchemaDefinition implements ReconSchemaDefinition {
   @Transactional
   public void initializeSchema() throws SQLException {
     Connection conn = dataSource.getConnection();
+    dslContext = DSL.using(conn);
     if (!TABLE_EXISTS_CHECK.test(conn, FILE_COUNT_BY_SIZE_TABLE_NAME)) {
       createFileSizeCountTable(conn);
     }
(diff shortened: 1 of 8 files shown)


hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/tasks/TestFileSizeCountTask.java
```diff
@@ -206,10 +216,189 @@ public void testProcess() {
...
+    recordToFind.value3(16384L);
+    assertEquals(0, fileCountBySizeDao.findById(recordToFind)
+        .getCount().longValue());
+    recordToFind.value3(65536L);
+    assertEquals(1, fileCountBySizeDao.findById(recordToFind)
+        .getCount().longValue());
+  }
+
+  @Test
+  public void testReprocessAtScale() throws IOException {
```
avijayanhwx, line 235: Good test to see how much we can handle!

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/FileSizeCountTask.java
```diff
@@ -49,33 +52,28 @@
...
 
-  private int maxBinSize = -1;
-  private long maxFileSizeUpperBound = 1125899906842624L; // 1 PB
-  private long[] upperBoundCount;
-  private long oneKb = 1024L;
+  // 1125899906842624L = 1PB
+  private static final long MAX_FILE_SIZE_UPPER_BOUND = 1125899906842624L;
   private FileCountBySizeDao fileCountBySizeDao;
+  // Map to store file counts in each <volume,bucket,fileSizeUpperBound>
+  private Map<FileSizeCountKey, Long> fileSizeCountMap;
```
avijayanhwx, line 59: Since we don't read everything from the DB at init time, 'fileSizeCountMap' can be a local variable created inside the 'process' and 'reprocess' methods.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/UtilizationEndpoint.java
```diff
@@ -34,7 +34,7 @@
  */
 @Path("/utilization")
 @Produces(MediaType.APPLICATION_JSON)
-public class UtilizationService {
+public class UtilizationEndpoint {
```
avijayanhwx, line 37: Can we add a getFileCounts(volume, bucket, fileSize(Default = null)) here?

hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/UtilizationSchemaDefinition.java
```diff
@@ -83,11 +86,17 @@ private void createClusterGrowthTable(Connection conn) {
   }
 
   private void createFileSizeCountTable(Connection conn) {
-    DSL.using(conn).createTableIfNotExists(FILE_COUNT_BY_SIZE_TABLE_NAME)
+    dslContext.createTableIfNotExists(FILE_COUNT_BY_SIZE_TABLE_NAME)
+        .column("volume", SQLDataType.VARCHAR(64))
```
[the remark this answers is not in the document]
vivekratnavel: Yes, verified that the max character limit of volume name and bucket name is 63 characters.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/FileSizeCountTask.java
vivekratnavel: Moving it to local variable would then require `writeCountsToDB`, `handlePutKeyEvent` and `handleDeleteKeyEvent` methods to take `Map<FileSizeCountKey, Long> fileSizeCountMap` as another argument. Currently, initializing this map in reprocess and process is just enough since both methods will never be called at the same time.

avijayanhwx: IMO, it is OK to have the handleDeleteKeyEvent & handlePutKeyEvent and writeCountsToDB to take the map as a parameter. They are meant to be stateless helper methods that act on the input given to them. In the current implementation, there is an assumption that the task's implementation makes of the higher level caller, which may not be true in the future.

vivekratnavel:
@avijayanhwx Thanks for the review!

PR #990, opened 2020-05-29
HDDS-3653. Add documentation for Copy key command

maobaolong:

Add a documents for cp key command

https://issues.apache.org/jira/projects/HDDS/issues/HDDS-3653

## How was this patch tested?

Just documents debt. no need to test

### DIFF
--- a/hadoop-hdds/docs/content/shell/KeyCommands.md
+++ b/hadoop-hdds/docs/content/shell/KeyCommands.md
@@ -31,6 +31,7 @@ Ozone shell supports the following key commands.
   * [list](#list)
   * [rename](#rename)
   * [cat](#cat)
+  * [copy](#cp)
 
 
 ### Get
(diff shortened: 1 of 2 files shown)


maobaolong:
@elek Please take a look at this documents, it would be better if land this in 0.6.0 release.

hadoop-hdds/docs/content/shell/KeyCommands.md
```diff
@@ -155,3 +156,22 @@ The `key cat` command displays the contents of a specific Ozone key to standard
...
+| Arguments                      |  Comment                                |
+|--------------------------------|-----------------------------------------|
+|  Uri                           | The name of the bucket in **/volume/bucket** format.
+|  FromKey                       | The existing key to be copied
+|  ToKey                         | The name of the new key
+| -r, \-\-replication            | Optional, Number of copies, ONE or THREE are the options. Picks up the default from cluster configuration.
+| -t, \-\-type                   | Optional, replication type of the new key. RATIS and STAND_ALONE are the options. Picks up the default from cluster configuration.
+
+{{< highlight bash >}}
+ozone sh key cp /hive/jan sales.orc new_one.orc
```
xiaoyuyao, line 175: The parameter is a bit different from hadoop cp comamnd. My understanding is that the uri is to help the OM HA but seems here we are restricted to under the same URI. Does it take full uri as from and to key parameters?

elek: Based on the code, the answer is no. You can use a bucketAddress (url or vol/bucket) + fromKey (string) + toKey (string).

This patch looks good to me as it describe the current behavior.

@xiaoyuyao Do you have any suggestion to change the code before the next release for better usability (btw. `ozone sh key put DESTINATIONURL localfile` is also reported to be confusing.)

xiaoyuyao: @elek yes. Let's file a ticket to fix it separately.

PR #991, opened 2020-05-29
HDDS-3686. Insert an argument of ozone shell to accept jvm arguments

maobaolong:

This is a great feature, i believe everyone would like this new "--jvmargs" argument, we can specify some arguments you like, mostly, i put the "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005" into HADOOP_OPTS, now i doesn't have to do like that.

https://issues.apache.org/jira/browse/HDDS-3686

## How was this patch tested?

```bash
bin/ozone --jvmargs "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005" admin pipeline list
Listening for transport dt_socket at address: 5005
```
then, use intellij idea to attach it and enjoy debugging, have a nice day.

### DIFF
--- a/hadoop-ozone/dist/src/shell/ozone/ozone
+++ b/hadoop-ozone/dist/src/shell/ozone/ozone
@@ -31,8 +31,7 @@ function hadoop_usage
   hadoop_add_option "--hosts filename" "list of hosts to use in worker mode"
   hadoop_add_option "--loglevel level" "set the log4j level for this command"
   hadoop_add_option "--workers" "turn on worker mode"
-  hadoop_add_option "--jvmargs" "jvm arguments, append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in HADOOP_CLIENT_OPTS will be append after these jvmargs"
-
+  hadoop_add_option "--jvmargs jvm arguments", "append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in HADOOP_CLIENT_OPTS will be append after these jvmargs"
 
   hadoop_add_subcommand "auditparser" client "runs audit parser tool"
   hadoop_add_subcommand "classpath" client "prints the class path needed to get the hadoop jar and the required libraries"
--- a/hadoop-ozone/dist/src/shell/ozone/ozone
+++ b/hadoop-ozone/dist/src/shell/ozone/ozone
@@ -31,7 +31,8 @@ function hadoop_usage
   hadoop_add_option "--hosts filename" "list of hosts to use in worker mode"
   hadoop_add_option "--loglevel level" "set the log4j level for this command"
   hadoop_add_option "--workers" "turn on worker mode"
-  hadoop_add_option "--jvmargs" "jvm arguments"
+  hadoop_add_option "--jvmargs" "jvm arguments, append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in HADOOP_CLIENT_OPTS will be append after these jvmargs"
+
 
   hadoop_add_subcommand "auditparser" client "runs audit parser tool"
   hadoop_add_subcommand "classpath" client "prints the class path needed to get the hadoop jar and the required libraries"
--- a/hadoop-ozone/dist/src/shell/ozone/ozone
+++ b/hadoop-ozone/dist/src/shell/ozone/ozone
@@ -31,7 +31,7 @@ function hadoop_usage
   hadoop_add_option "--hosts filename" "list of hosts to use in worker mode"
   hadoop_add_option "--loglevel level" "set the log4j level for this command"
   hadoop_add_option "--workers" "turn on worker mode"
-  hadoop_add_option "--jvmargs jvm arguments", "append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in HADOOP_CLIENT_OPTS will be append after these jvmargs"
+  hadoop_add_option "--jvmargs arguments" "append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in HADOOP_CLIENT_OPTS will be append after these jvmargs"
 
   hadoop_add_subcommand "auditparser" client "runs audit parser tool"
   hadoop_add_subcommand "classpath" client "prints the class path needed to get the hadoop jar and the required libraries"
--- a/hadoop-ozone/dist/src/shell/ozone/ozone
+++ b/hadoop-ozone/dist/src/shell/ozone/ozone
@@ -31,6 +31,7 @@ function hadoop_usage
   hadoop_add_option "--hosts filename" "list of hosts to use in worker mode"
   hadoop_add_option "--loglevel level" "set the log4j level for this command"
   hadoop_add_option "--workers" "turn on worker mode"
+  hadoop_add_option "--jvmargs" "jvm arguments"
 
   hadoop_add_subcommand "auditparser" client "runs audit parser tool"
   hadoop_add_subcommand "classpath" client "prints the class path needed to get the hadoop jar and the required libraries"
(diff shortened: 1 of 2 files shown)


sodonnel:
I think this is a good idea. However we can already pass JVM args like this:

```
export HADOOP_CLIENT_OPTS="$HADOOP_CLIENT_OPTS -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"
bash-4.2$ ozone admin pipeline list
Listening for transport dt_socket at address: 5005
```
I guess the HADOOP_CLIENT_OPTS option is there as the shell commands use the hadoop libraries to help with the CLI.

Looking at the Ozone doc pages, I don't see this mentioned anywhere and a google search for "site:hadoop.apache.org/ozone HADOOP_CLIENT_OPTS" gives no results either.

I wonder if we should expose the HADOOP_CLIENT_OPTS in the documentation, or rename it to OZONE_CLIENT_OPTS instead ...

maobaolong:
@sodonnel  Thank you for telling this. In fact i used to use HADOOP_OPTS and HADOOP_CLIENT_OPTS to achieve my debug purpose, but it hard to support me to write the following script.

```bash
bin/ozone scm --init
bin/ozone --daemon start scm
sh /Users/mbl/projects/github/hadoop-ozone/.project/hadoop/copyconfig.sh om
bin/ozone om --init
bin/ozone --daemon start om
bin/ozone --daemon start recon
bin/ozone sh vol create myvol;  bin/ozone sh bucket create myvol/mybucket
bin/ozone --daemon start datanode

```

I want to start the debug agent of each ozone server with different port. If we have this PR, we can 
```bash

bin/ozone scm --init
bin/ozone --jvmargs " -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5000" --daemon start scm
bin/ozone om --init
bin/ozone --jvmargs " -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5001"  --daemon start om
bin/ozone --jvmargs " -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5002"  --daemon start recon
bin/ozone sh vol create myvol;  bin/ozone sh bucket create myvol/mybucket
bin/ozone --jvmargs " -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5003"  --daemon start datanode
```

sodonnel:
Another way of using the HADOOP_CLIENT_OPTS / HADOOP_OPTS is by not exporting the variable, but including it on the same line as the command, eg:

```
> HADOOP_CLIENT_OPTS="$HADOOP_CLIENT_OPTS -agentlib:jdwp..." ozone scm ...
```
However, this is not as discoverable as an explicit jvmargs flag like you have added here.

With your change, does this command append to existing HADOOP_OPTS or overwrite any existing arguments set by HADOOP_OPTS? 

```
hadoop_add_param HADOOP_OPTS "$1" "$1"
```

I am wondering how these two features would work together?

sodonnel:
After a bit of trial and error I figure out that:

Anything in HADOOP_OPTS comes first.

These new JVM Flags are appended to that.

Anything in HADOOP_CLIENT_OPTS is appended to that, eg:

"$HADOOP_OPTS --jvmflags $HADOOP_CLIENT_OPTS"

maobaolong:
@sodonnel  Thank you for your review.

> With your change, does this command append to existing HADOOP_OPTS or overwrite any existing arguments set by HADOOP_OPTS?

The function `hadoop_add_param <envvar> <checkstring> <appendstring>` will append the `appendstring` if `checkstring` is not set.
 
 $1 is what we are adding to
 $2 is the name of what we want to add (key)
 $3 is the key+value of what we're adding

As the jvmargs will bring a bundle of jvm arguments, so i just check the whole arguments string as lazy.

> After a bit of trial and error I figure out that:
> Anything in HADOOP_OPTS comes first.
> These new JVM Flags are appended to that.
> Anything in HADOOP_CLIENT_OPTS is appended to that, eg:
>
> "$HADOOP_OPTS --jvmflags $HADOOP_CLIENT_OPTS"

Thank you for trial this PR.

hadoop-ozone/dist/src/shell/ozone/ozone
```diff
@@ -31,6 +31,7 @@ function hadoop_usage
   hadoop_add_option "--hosts filename" "list of hosts to use in worker mode"
   hadoop_add_option "--loglevel level" "set the log4j level for this command"
   hadoop_add_option "--workers" "turn on worker mode"
+  hadoop_add_option "--jvmargs" "jvm arguments"
```
sodonnel, line 34: Could we add a bit more help text here? I'd like to make it clear the arguments here are appended to HADOOP_OPTS and if there are also any HADOOP_CLIENT_OPTS they are appended after. eg:

```
hadoop_add_option "--jvmargs" "jvm arguments" "append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in HADOOP_CLIENT_OPTS will be append after these jvmargs"
```

maobaolong:
@sodonnel Thank you for your suggestion, i push a new commit to address it, PTAL.

hadoop-ozone/dist/src/shell/ozone/ozone
```diff
@@ -31,6 +31,8 @@ function hadoop_usage
   hadoop_add_option "--hosts filename" "list of hosts to use in worker mode"
   hadoop_add_option "--loglevel level" "set the log4j level for this command"
   hadoop_add_option "--workers" "turn on worker mode"
+  hadoop_add_option "--jvmargs" "jvm arguments, append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in HADOOP_CLIENT_OPTS will be append after these jvmargs"
```
sodonnel, line 34: To keep this consistent with the other options, I think this new option should be:

```
hadoop_add_option "--jvmargs jvm arguments", "append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in HADOOP_CLIENT_OPTS will be append after these jvmargs"
```

Please also remove the extra blank line after this new line of code.

maobaolong:
@sodonnel Thank you for your review again, i push a new commit to address your comment. PTAL

hadoop-ozone/dist/src/shell/ozone/ozone
```diff
@@ -31,6 +31,7 @@ function hadoop_usage
   hadoop_add_option "--hosts filename" "list of hosts to use in worker mode"
   hadoop_add_option "--loglevel level" "set the log4j level for this command"
   hadoop_add_option "--workers" "turn on worker mode"
+  hadoop_add_option "--jvmargs jvm arguments", "append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in HADOOP_CLIENT_OPTS will be append after these jvmargs"
```
sodonnel, line 34: I just checked how this looks on the CLI:

```
--buildpaths                       attempt to add class files from build tree
--config dir                       Hadoop config directory
--daemon (start|status|stop)       operate on a daemon
--debug                            turn on shell script debug mode
--help                             usage information
--hostnames list[,of,host,names]   hosts to use in worker mode
--hosts filename                   list of hosts to use in worker mode
--jvmargs jvm arguments,           append JVM options to any existing options defined in the HADOOP_OPTS environment variable. Any defined in
                                   HADOOP_CLIENT_OPTS will be append after these jvmargs
--loglevel level                   set the log4j level for this command

```
It seems a "," character has made it into the output. This may be my fault as the example change I gave could have had a comma in it.

Could you just remove the comma between the arguments and then this should be good to commit.

Thanks!

maobaolong:
@sodonnel Sorry for the mistake,  BTW, i change the "jvm arguments" to "arguments", i think it would be better if there is only one word here, two word maybe make people think it could be two arguments, and from the context, we know the arguments stand for the jvm arguments.

sodonnel:
Thanks for the few iterations on this. Looks good now, so I will go ahead and merge it.

maobaolong:
@sodonnel Thanks for your review

PR #993, opened 2020-05-29
HDDS-3542.Ozone chunkinfo CLI cannot connect to OM when run from non-om node.

sadanand48:
The chunkinfo tool was not connecting to OM  on a non-om leader client node .The fix  here is to use existing ozoneManagerClient instance instead of creating a new one.

https://issues.apache.org/jira/browse/HDDS-3542

## How was this patch tested?
Tested manually.


adoroszlai:
@sadanand48 please trigger new CI check with empty commit:

```
git commit --allow-empty -m 'trigger new CI check' && git push
```

PR #994, opened 2020-05-29
HDDS-3082. Refactor recon missing containers task to detect under, over and mis-replicated containers.

sodonnel:
The current Recon "Missing Containers Task" only highlights missing containers in the cluster.

It is desired for it to also detect under, over and mis-replicated containers.

In order to do that, the existing database table MISSING_CONTAINERS has been renamed to UNHEALTHY_CONTAINERS, with the definition:

```
container_id bigint NOT NULL,
container_state varchar(16) NOT NULL,
in_state_since bigint not null,
expected_replica_count integer,
actual_replica_count integer,
replica_delta integer not null,
reason varchar(500)
```

The container state can be MISSING, UNDER_REPLICATED, OVER_REPLICATED or MIS_REPLICATED.

A design decision was made so that if a container is MISSING, then it is not in any of the other states.

However, it can be both under and mis-replicated or in theory over and mis-replicated at the same time and this would result in two rows in the database for a single container.

Each time the "Container Health task" runs, it scans all the existing records, updates any counts and removes any records that are no longer valid.

Then it processes all other containers without any records in the unhealthy_containters table.

The reason the job is split into two parts, is to avoid the need to query the database for every single container on each run.

This change only adjusts the job and the backend storage. An additional change is needed to change the rest endpoints to expose the new container states to the users and UI.

https://issues.apache.org/jira/browse/HDDS-3082

## How was this patch tested?

New and existing unit tests


hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
```diff
@@ -0,0 +1,345 @@
...
+        reconTaskConfig.getMissingContainerTaskInterval());
+  }
+
+  public synchronized void run() {
+    try {
+      while (canRun()) {
+        long start = Time.monotonicNow();
+        long currentTime = System.currentTimeMillis();
+        long existingCount = processExistingDBRecords(currentTime);
+        LOG.info("Missing Container task thread took {} milliseconds to" +
```
vivekratnavel, line 79:
```suggestion
        LOG.info("Container Health task thread took {} milliseconds to" +
```

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
```diff
@@ -0,0 +1,345 @@
...
+                " process {} existing database records.",
+            Time.monotonicNow() - start, existingCount);
+        start = Time.monotonicNow();
+        final List<ContainerInfo> containers = containerManager.getContainers();
+        containers.forEach(container ->
+            processContainer(container, currentTime));
+        recordSingleRunCompletion();
+        LOG.info("Missing Container task Thread took {} milliseconds for" +
```
vivekratnavel, line 87:
```suggestion
        LOG.info("Container Health task thread took {} milliseconds for" +
```

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
```diff
@@ -0,0 +1,345 @@
...
+          containerManager.getContainerReplicas(container.containerID());
+      ContainerHealthStatus h = new ContainerHealthStatus(
+          container, containerReplicas, placementPolicy);
+      if (h.isHealthy()) {
+        return;
+      }
+      containerSchemaManager.insertUnhealthyContainerRecords(
+          ContainerHealthRecords.generateUnhealthyRecords(h, currentTime));
+    } catch (ContainerNotFoundException e) {
+      LOG.error("Container not found while finding missing containers", e);
```
vivekratnavel, line 171:
```suggestion
      LOG.error("Container not found while processing container in Container Health task", e);
```

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
```diff
@@ -0,0 +1,345 @@
...
+    }
+  }
+
+  /**
+   * Helper methods to generate and update the required database records for
+   * unhealthy containers.
+   */
+  static public class ContainerHealthRecords {
+
+    static public boolean retainOrUpdateRecord(
```
vivekratnavel, line 181: Can you please add JavaDoc for this method?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
```diff
@@ -0,0 +1,345 @@
...
+      return false;
+    }
+
+    /**
+     * With a Jooq record, if you update any field in the record, the record
+     * is marked as changed, even if you updated it to the same value as it is
+     * already set to. We only need to run a DB update statement if the record
+     * has really changed. The methods below ensure we do not update the Jooq
+     * record unless the values have changed and hence save a DB execution
+     * when
```
vivekratnavel, line 314: Please fix the dangling statement

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/fsck/TestContainerHealthStatus.java
```diff
@@ -0,0 +1,181 @@
...
+
+import static junit.framework.TestCase.assertEquals;
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test to ensure the correct container state is return by a
+ * ContainerHealthyStatus instance.
```
vivekratnavel, line 42:
```suggestion
 * Test to ensure the correct container state is returned by a
 * ContainerHealthStatus instance.
```

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/fsck/TestContainerHealthTask.java
```diff
@@ -0,0 +1,280 @@
...
+import org.apache.hadoop.ozone.recon.persistence.AbstractReconSqlDBTest;
+import org.hadoop.ozone.recon.schema.tables.daos.ReconTaskStatusDao;
+import org.hadoop.ozone.recon.schema.tables.daos.UnhealthyContainersDao;
+import org.hadoop.ozone.recon.schema.tables.pojos.ReconTaskStatus;
+import org.hadoop.ozone.recon.schema.tables.pojos.UnhealthyContainers;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Class to test single run of Missing Container Task.
```
vivekratnavel, line 60:
```suggestion
 * Class to test single run of Container Health Task.
```

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
```diff
@@ -0,0 +1,345 @@
...
+      }
+
+      if (container.isMissing()
+          && !recordForStateExists.contains(
+              UnHealthyContainerStates.MISSING.toString())) {
+        records.add(
+            recordForState(container, UnHealthyContainerStates.MISSING, time));
+      }
+
+      if (container.isUnderReplicated()
```
avijayanhwx, line 231: Can we add an 'else' after the isMissing() check? A missing container cannot be in any other state right?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
sodonnel: Well spotted. I have fixed this.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
sodonnel: Well spotted. I have fixed this.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
sodonnel: Fixed.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
sodonnel: Done.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
sodonnel: Fixed.

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/fsck/TestContainerHealthStatus.java
sodonnel: Fixed.

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/fsck/TestContainerHealthTask.java
sodonnel: Fixed.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
```diff
@@ -0,0 +1,345 @@
...
+      Set<String> existingRecords, long currentTime) {
+    if (container != null) {
+      containerSchemaManager.insertUnhealthyContainerRecords(
+          ContainerHealthRecords.generateUnhealthyRecords(
+              container, existingRecords, currentTime));
+      processedContainers.add(container.getContainer());
+    }
+  }
+
+  private long processExistingDBRecords(long currentTime) {
```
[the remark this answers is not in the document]
sodonnel: Done.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
```diff
@@ -0,0 +1,345 @@
...
+      while (canRun()) {
+        long start = Time.monotonicNow();
+        long currentTime = System.currentTimeMillis();
+        long existingCount = processExistingDBRecords(currentTime);
+        LOG.info("Missing Container task thread took {} milliseconds to" +
+                " process {} existing database records.",
+            Time.monotonicNow() - start, existingCount);
+        start = Time.monotonicNow();
+        final List<ContainerInfo> containers = containerManager.getContainers();
+        containers.forEach(container ->
```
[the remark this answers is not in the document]
sodonnel: Good idea. I have changed this.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
sodonnel: You are correct. Rather than next the rest of the method in the else block, I added a return statement in the missing branch. Does that look ok?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
```diff
@@ -0,0 +1,345 @@
...
+    long recordCount = 0;
+    try (Cursor<UnhealthyContainersRecord> cursor =
+             containerSchemaManager.getAllUnhealthyRecordsCursor()) {
+      ContainerHealthStatus currentContainer = null;
+      Set<String> existingRecords = new HashSet<>();
+      while(cursor.hasNext()) {
+        recordCount++;
+        UnhealthyContainersRecord rec = cursor.fetchNext();
+        try {
+          if (currentContainer == null
```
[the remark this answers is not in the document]
sodonnel: On the first pass through, currentContainer will be null, so we cannot called `currentContainer.getContainerID()` on it, so I think we need the null check.

If could be changed to something like:

```
if (container == null) {
  currentContainer = setCurrentContainer(rec.getContainerId());
} else if (currentContainer.getContainerID() != rec.getContainerId()) {
  completeProcessingContainer(currentContainer, existingRecords, currentTime);
  existingRecords.clear(); 
  currentContainer = setCurrentContainer(rec.getContainerId());
}
```

I am not sure that is much better though.

We will also still need a null check in completeProcessingContainer or around the call to it, to handle the case where there are no containers processed in the cursor.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/fsck/ContainerHealthTask.java
avijayanhwx: +1

avijayanhwx:
Thank you @sodonnel. LGTM +1

PR #997, opened 2020-05-31
HDDS-3479. Use SCMMetadataStore high level abstraction instead of DBS…

jsoft88:
…toreBuilder and SCMDBDefinition directly

HDDS-3479. Rename SCMMetadataStoreRDBImpl to SCMMetadataStoreImpl as it is not longer RocksDB specific.

Use ScmMetadataStore everywhere inside SCM instead of DBStoreBuilder with separated DB definition (SCMDBDefinition)

https://issues.apache.org/jira/browse/HDDS-3479?jql=project%20in%20(HDDS)%20AND%20labels%20in%20(newbie)%20AND%20assignee%20is%20EMPTY%20AND%20status%20in%20(open%2C%20Reopened)

## How was this patch tested?
By executing test.sh under dist/


jsoft88:
The acceptance check has some errors, specifically, test cases related to S3 in this part  Run cd /mnt/ozone/compose && ./test-all.sh. I don't think these errors are caused by the changes introduced in this PR.

PR #1000, opened 2020-06-01
HDDS-3562. Datanodes should send ICR when a container replica deletion is successful.

nandakumar131:
Whenever a datanode executes the delete container command and deletes the container replica, it has to immediately send an ICR to update the container replica state in SCM.

HDDS-3562

## How was this patch tested?
Related unit tests added

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/AbstractContainerReportHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/AbstractContainerReportHandler.java
@@ -68,13 +68,6 @@ public class AbstractContainerReportHandler {
       throws IOException {
     final ContainerID containerId = ContainerID
         .valueof(replicaProto.getContainerID());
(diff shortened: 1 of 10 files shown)


elek:
Congratulation for the number #1000 ;-)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/AbstractContainerReportHandler.java
```diff
@@ -225,14 +218,35 @@ private void updateContainerState(final DatanodeDetails datanode,
...
+    }
+  }
+
   /**
    * Returns true if the container replica is not marked as UNHEALTHY.
    *
    * @param replicaState State of the container replica.
    * @return true if unhealthy, false otherwise
    */
   private boolean isUnhealthy(final Supplier<State> replicaState) {
```
dineshchitlangia, line 246: NIT: Update javadocs

nandakumar131:
> Congratulation for the number #1000 ;-)

Thanks @elek  😄

PR #1002, opened 2020-06-01
HDDS-3642. Stop/Pause Background services while replacing OM DB with checkpoint from Leader

hanishakoneru:

When a follower OM needs to replace its DB with a checkpoint from Leader (to catch up on the transactions), it should pause or stop services which read/ write to the DB. 

During OM HA testing, found that OM could crash with JVM error on RocksDB. This happened because KeyDeletingService was trying to access a memory which is already freed up.

Please see Jira link below for error logs.

https://issues.apache.org/jira/browse/HDDS-3642

## How was this patch tested?

Tested on a local docker cluster.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -3047,18 +3047,25 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
       return null;
     }
 
-    // Pause the State Machine so that no new transactions can be applied.
-    // This action also clears the OM Double Buffer so that if there are any
-    // pending transactions in the buffer, they are discarded.
-    // TODO: The Ratis server should also be paused here. This is required
-    //  because a leader election might happen while the snapshot
-    //  installation is in progress and the new leader might start sending
-    //  append log entries to the ratis server.
-    omRatisServer.getOmStateMachine().pause();
+    File oldDBLocation = metadataManager.getStore().getDbLocation();
+    try {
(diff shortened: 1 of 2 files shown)


bharatviswa504:
@hanishakoneru This PR needs a rebase.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3113,27 +3120,32 @@ private DBCheckpoint getDBCheckpointFromLeader(String leaderId) {
     return null;
   }
 
+  void stopServices() throws Exception {
+    keyManager.stop();
+    stopSecretManager();
+    metadataManager.stop();
+
+    // s3SecretManager should also be stopped
```
bharatviswa504, line 3137: Right now, we don't have any stop() for S3SecretManager and PrefixManager.
And also for these, only read/write operations will happen. Write will go through double buffer and read's will not happen anyway as this is not leader. Do we still see issue?

hanishakoneru: If writes can happen then we should stop it, right?

bharatviswa504: Thanks for confirming.
Then we can remove the comments.

bharatviswa504:
Looks like it is failing in compilation and also check style issues.

bharatviswa504:
Kicked off CI run, test has not run due to unavailability of artificats.

bharatviswa504:

PR #1005, opened 2020-06-02
HDDS-3350. Ozone Retry Policy Improvements.

lokeshj1703:

Currently any ozone client request can spend a huge amount of time in retries and ozone client can retry its requests very aggressively. The waiting time can thus be very high before a client request fails. Further aggressive retries by ratis client used by ozone can bog down a ratis pipeline leader. The Jira aims to make changes to the current retry behavior in Ozone client.

https://issues.apache.org/jira/browse/HDDS-3350

## How was this patch tested?

Teragen results were compared with and without the new retry policy. The results are uploaded in Apache jira.


bshashikant:
@lokeshj1703 , can you please update the patch?

pom.xml
```diff
@@ -79,7 +79,7 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <declared.ozone.version>${ozone.version}</declared.ozone.version>
 
     <!-- Apache Ratis version -->
-    <ratis.version>0.6.0-cac3336-SNAPSHOT</ratis.version>
+    <ratis.version>0.6.0-6ab75ae-SNAPSHOT</ratis.version>
```
arp7, line 82: Is this a duplicate of HDDS-3350? Let's decouple the Ratis version update from the rest of the changes.

lokeshj1703: Ratis snapshot version update is required for compilation. Once HDDS-3654 is committed, I'll remove these changes.

lokeshj1703: Marking it as pending until 3654 is committed.

lokeshj1703:
OM HA related tests seem related to ratis snapshot upgrade. They will be fixed with HDDS-3654.
There are test timeouts in TestOzoneRpcClient which are happening because watch requests are committed by majority. The test shows appendEntry timeout from leader to follower. It does not seem to be fixed with latest ratis snapshot.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
```diff
@@ -269,23 +282,76 @@ static GrpcTlsConfig createTlsClientConfig(SecurityConfig conf,
...
+        .setBaseSleepTime(
+            TimeDuration.valueOf(exponentialBaseSleep, TimeUnit.MILLISECONDS))
+        .setMaxSleepTime(
+            TimeDuration.valueOf(exponentialMaxSleep, TimeUnit.MILLISECONDS))
+        .build();
+  }
+
+  private static ExceptionDependentRetry createExceptionDependentPolicy(
+      ExponentialBackoffRetry exponentialBackoffRetry,
+      MultipleLinearRandomRetry multipleLinearRandomRetry,
```
bshashikant, line 338: Instead of explicitly adding no retry for specific exception , can we define a static list wof exceptions for which there will be no retry and iterate here. ?

bshashikant: I think we need to add AlreadyClosedException to the list of no retry here( as this can be generated from ratis server as well). Also, for RfatLogIOException , the policy should be of no retry.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
```diff
@@ -269,23 +282,76 @@ static GrpcTlsConfig createTlsClientConfig(SecurityConfig conf,
     return tlsConfig;
   }
```
bshashikant, line 284: can we add a table defining the exception no retry policy relationship for better understanding?

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
```diff
@@ -281,15 +281,33 @@
 
   public static final String DFS_CONTAINER_RATIS_DATANODE_STORAGE_DIR =
       "dfs.container.ratis.datanode.storage.dir";
-  public static final String DFS_RATIS_CLIENT_REQUEST_MAX_RETRIES_KEY =
-      ScmConfigKeys.DFS_RATIS_CLIENT_REQUEST_MAX_RETRIES_KEY;
-  public static final int DFS_RATIS_CLIENT_REQUEST_MAX_RETRIES_DEFAULT =
-      ScmConfigKeys.DFS_RATIS_CLIENT_REQUEST_MAX_RETRIES_DEFAULT;
-  public static final String DFS_RATIS_CLIENT_REQUEST_RETRY_INTERVAL_KEY =
-      ScmConfigKeys.DFS_RATIS_CLIENT_REQUEST_RETRY_INTERVAL_KEY;
+
```
bshashikant, line 284: i guess it would be better to define these configs in RatisClientConfig instead of defining here.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java
lokeshj1703: Adding NoRetry for AlreadyClosedException was leading to test failures. This exception is generated for closed connections in Ratis. I have removed NoRetry policy for this exception.

bshashikant:
Unit test failures are related. @lokeshj1703 can you please check?

lokeshj1703:
@bshashikant I have updated the PR. 

There is a test failure in Test2WayCommitInRatis in the latest run. I ran the test 10 times and it passes locally. The test failure can also be seen in https://elek.github.io/ozone-build-results/ in multiple PRs.

lokeshj1703:
I have created https://issues.apache.org/jira/browse/HDDS-3799 for tracking the failure related to Test2WayCommitInRatis.

bshashikant:
Thanks @lokeshj1703 for working on this.

PR #1006, opened 2020-06-02
HDDS-3699. Change write chunk failure logging level to ERROR in BlockOutputStream.

sadanand48:
Changing logging level  for a  write chunk  failure to INFO in BlockOutputStream.

https://issues.apache.org/jira/browse/HDDS-3699

## How was this patch tested?
not required.

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
@@ -640,11 +640,9 @@ public class BlockOutputStream extends OutputStream {
         }
         return e;
       }, responseExecutor).exceptionally(e -> {
-        if (LOG.isDebugEnabled()) {
-          LOG.debug(
-              "writing chunk failed " + chunkInfo.getChunkName() + " blockID "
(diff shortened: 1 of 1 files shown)


elek:
Thanks to open this issue @sadanand48. Unfortunately I don't have the context? (How did you notice this? What was the problem? How can I reproduce it?).

Therefore, I have a few questions to understand this patch:  

Why don't we use `ERROR`, why do we need `INFO`?

What is the expected number of log messages in case of any error? (Let's say disk is not writable). Is it possible to a lot of INFO message?

The exception seems to be propagated. Is it logged on higher level?

bshashikant:
> Thanks to open this issue @sadanand48. Unfortunately I don't have the context? (How did you notice this? What was the problem? How can I reproduce it?).
> 
> Therefore, I have a few questions to understand this patch:
> 
> Why don't we use `ERROR`, why do we need `INFO`?
> 
> What is the expected number of log messages in case of any error? (Let's say disk is not writable). Is it possible to a lot of INFO message?
> 
> The exception seems to be propagated. Is it logged on higher level?

It should be logged in error level. It will be useful to log this error for debug purpose. The putBlock/watchForCommit exceptions are logged not the write chunk failures on higher level.

elek:
Ups, we got green build, but we forgot to merge it. Sorry for that. Doing it now. Thanks for the contribution.

PR #1008, opened 2020-06-02
HDDS-3671. Introduce SCMStateMachineHandler marker interface.

nandakumar131:

Introduce `SCMStateMachineHandler` marker interface which can be used to identify the StateMachine handlers in SCM

HDDS-3671

## How was this patch tested?
This change doesn't require any additional tests.


timmylicheng:
+1. Pending on CI.
This should help the Invocation Handler to be more complete.

PR #1009, opened 2020-06-02
HDDS-3612. Document details of bucket mount design

adoroszlai:

Add proposed implementation details to Ozone S3 volume mapping design doc.  The primary goal is to be able to discuss it before going forward with big code changes.

https://issues.apache.org/jira/browse/HDDS-3612

## How was this patch tested?

Checked preview on GitHub.

### DIFF
--- a/hadoop-hdds/docs/content/design/ozone-volume-management.md
+++ b/hadoop-hdds/docs/content/design/ozone-volume-management.md
@@ -102,27 +102,30 @@ This is an easy an fast method, but with this approach not all the volumes are a
  To implement the second (expose ozone buckets as s3 buckets) we have multiple options:
 
    1. Store some metadata (** s3 bucket name **) on each of the buckets
-   2. Implement a **bind mount** mechanic which makes it possible to *mount* any volume/buckets to the specific "s3" volume.
+   2. Implement a **symbolic link** mechanism which makes it possible to *link* to any volume/buckets from the "s3" volume.
 
 The first approach required a secondary cache table and it violates the naming hierarchy. The s3 bucket name is a global unique name, therefore it's more than just a single attribute on a specific object. It's more like an element in the hierachy. For this reason the second option is proposed:
 
 For example if the default s3 volume is `s3v`
 
  1. Every new buckets created via s3 interface will be placed under the `/s3v` volume
- 2. Any existing **Ozone** buckets can be exposed with mounting it to s3: `ozone sh mount /vol1/bucket1 /s3v/s3bucketname`
+ 2. Any existing **Ozone** buckets can be exposed by linking to it from s3: `ozone sh bucket link /vol1/bucket1 /s3v/s3bucketname`
 
 **Lock contention problem**
 
 One possible problem with using just one volume is using the locks of the same volume for all the S3 buckets (thanks Xiaoyu). But this shouldn't be a big problem.
 
  1. We hold only a READ lock. Most of the time it can acquired without any contention (writing lock is required only to change owner / set quota)
- 2. For symbolic link / bind mounts the read lock is only required for the first read. After that the lock of the referenced volume will be used. In case of any performance problem multiple volumes + bind mounts can be used.
+ 2. For symbolic link the read lock is only required for the first read. After that the lock of the referenced volume will be used. In case of any performance problem multiple volumes and links can be used.
 
 Note: Sanjay is added to the authors as the original proposal of this approach.
 
 #### Implementation details
 
- * Let bucket mount operation create a link bucket.  Links are like regular buckets, stored in DB the same way, but with two new, optional pieces of information: source volume and bucket.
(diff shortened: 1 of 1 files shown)


hadoop-hdds/docs/content/design/ozone-volume-management.md
```diff
@@ -4,7 +4,7 @@ summary: A simplified version of mapping between S3 buckets and Ozone volume/buc
 date: 2020-04-02
 jira: HDDS-3331
 status: accepted
-author: Marton Elek, Arpit Agarwall, Sunjay Radia
+author: Marton Elek, Arpit Agarwal, Sanjay Radia
```
arp7, line 7: You can remove my name, I really cannot claim any authorship credit for this idea. 🙂

hadoop-hdds/docs/content/design/ozone-volume-management.md
```diff
@@ -106,19 +106,27 @@ This is an easy an fast method, but with this approach not all the volumes are a
...
  2. For symbolic link / bind mounts the read lock is only required for the first read. After that the lock of the referenced volume will be used. In case of any performance problem multiple volumes + bind mounts can be used.
 
-Note: Sunjay is added to the authors as the original proposal of this approach.
+Note: Sanjay is added to the authors as the original proposal of this approach.
+
+#### Implementation details
+
+ * Let bucket mount operation create a link bucket.  Links are like regular buckets, stored in DB the same way, but with two new, optional pieces of information: source volume and bucket.
+ * Existing bucket operations (info, delete, ACL) work on the link object in the same way as they do on regular buckets.  No new link-specific RPC is required.
+ * Links are followed for key operations (list, get, put, etc.).  Checks for existence of the source bucket, as well as ACL, are performed at this time (similar to symlinks).  This avoids the need for reverse checks for each bucket delete or ACL change.
```
arp7, line 127: Yeah this should work a lot like symlinks, so we shouldn't perform reverse checks on changes to the target bucket.

hadoop-hdds/docs/content/design/ozone-volume-management.md
```diff
@@ -106,19 +106,27 @@ This is an easy an fast method, but with this approach not all the volumes are a
...
+ * The same permission is required on both the link and the source bucket to be able to perform the operation via the link.  This allows finer-grained access control.
```
arp7, line 128: We should probably try to match the behavior of Unix symlinks wrt permissions.

hadoop-hdds/docs/content/design/ozone-volume-management.md
```diff
@@ -106,19 +106,27 @@ This is an easy an fast method, but with this approach not all the volumes are a
...
 
  1. We hold only a READ lock. Most of the time it can acquired without any contention (writing lock is required only to change owner / set quota)
  2. For symbolic link / bind mounts the read lock is only required for the first read. After that the lock of the referenced volume will be used. In case of any performance problem multiple volumes + bind mounts can be used.
 
-Note: Sunjay is added to the authors as the original proposal of this approach.
+Note: Sanjay is added to the authors as the original proposal of this approach.
+
+#### Implementation details
+
+ * Let bucket mount operation create a link bucket.  Links are like regular buckets, stored in DB the same way, but with two new, optional pieces of information: source volume and bucket.
```
arp7, line 125: > Let bucket mount operation create a link bucket
Didn't understand this sentence. Does it mean that when you try to mount a bucket in a new volume it silently creates a link under the covers? Is the link reused next time we try to mount again?

Also how do we handle name collisions? Can the user choose any name for the link/mount point?

adoroszlai: > > Let bucket mount operation create a link bucket
> 
> Didn't understand this sentence. Does it mean that when you try to mount a bucket in a new volume it silently creates a link under the covers?

Mount and link are the same.  (Even the high-level parts of the doc refer to it as "link" in some places and "mount" in others.  Maybe we should not mix terminology.  Since proposed behavior is closer to symlinks, I suggest using "link" instead of "mount".)

> Is the link reused next time we try to mount again?

The mount is "active" until the link is deleted via regular DeleteBucket request.  Until then, another attempt to mount (or create a bucket with same name) will result in a "bucket already exists" error.

> Also how do we handle name collisions? Can the user choose any name for the link/mount point?

Name collision is handled the same as if trying to create a regular bucket.

Basically, the extra link info does not change behavior until client starts working with keys in the "bucket".  Then it is redirected.

adoroszlai:
Thanks @arp7 for reviewing this doc.  I have updated it with some clarifications.  Also changed to use "symbolic link" terminology instead of "mount" for consistency.  This changes the text of the original proposal very slightly, but keeps its spirit/intentions, I think.  Eg. propose to call the command `ozone sh bucket link` instead of `ozone sh mount`.

hadoop-hdds/docs/content/design/ozone-volume-management.md
elek: @arp7 You wrote additional section to this document about the volume ownership problems.

adoroszlai:
Thanks @arp7 and @elek for the review.

PR #1012, opened 2020-06-03
HDDS-3658. Stop to persist container related pipeline info of each key into OM DB to reduce DB size

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-3658

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
@@ -377,13 +377,26 @@ public final class OmKeyInfo extends WithObjectID {
     }
   }
 
+  /**
+   * For network transmit.
+   * @return
+   */
   public KeyInfo getProtobuf() {
+    return getProtobuf(false);
+  }
+
+  /**
+   *
+   * @param ignorePipeline true for persist to DB, false for network transmit.
+   * @return
+   */
+  public KeyInfo getProtobuf(boolean ignorePipeline) {
     long latestVersion = keyLocationVersions.size() == 0 ? -1 :
         keyLocationVersions.get(keyLocationVersions.size() - 1).getVersion();
 
     List<KeyLocationList> keyLocations = new ArrayList<>();
     for (OmKeyLocationInfoGroup locationInfoGroup : keyLocationVersions) {
-      keyLocations.add(locationInfoGroup.getProtobuf());
+      keyLocations.add(locationInfoGroup.getProtobuf(ignorePipeline));
     }
 
     KeyInfo.Builder kb = KeyInfo.newBuilder()
@@ -393,8 +406,8 @@ public final class OmKeyInfo extends WithObjectID {
         .setDataSize(dataSize)
         .setFactor(factor)
         .setType(type)
-        .addAllKeyLocationList(keyLocations)
         .setLatestVersion(latestVersion)
+        .addAllKeyLocationList(keyLocations)
         .setCreationTime(creationTime)
         .setModificationTime(modificationTime)
         .addAllMetadata(KeyValueUtil.toProtobuf(metadata))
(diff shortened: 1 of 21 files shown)


ChenSammi:
The rat check failed in doc module.  Have we changed the rule recently?

adoroszlai:
> The rat check failed in doc module. Have we changed the rule recently?

Thanks @ChenSammi for flagging this problem.  Posted #1017 to fix it.

elek (approved):
+1 LGTM, but I would prefer to get feedback from others (@nandakumar131?)

Thanks to create this patch @ChenSammi 

For me it seems to be reasonable. Pipeline information can changed at any time, seems to be better strategy to investing in caching on OM side.

One comment:

This patch introduces two diffrent protobuf conversion for the same class. This is an additional confirmation for me that we need to separate the network and storage protobuf from each other. It would be better to have a separated `interface-project` for all the protobuf files which are used in rocksdb. Would be easier to follow the incompatible/compatible changes.

But not sure how is it possible to avoid unnecessary copying. (This idea is not related to this patch, I am fine with the flag, just thinking about a direction for long-term)

elek:
As there is no other objection / comment I will merge this after a green build.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
```diff
@@ -408,6 +421,15 @@ public KeyInfo getProtobuf() {
...
     }
+    if (ignorePipeline) {
+      kb.addAllKeyLocationList(keyLocationVersions.stream()
+          .map(OmKeyLocationInfoGroup::getCompactProtobuf)
+          .collect(Collectors.toList()));
+    } else {
+      kb.addAllKeyLocationList(keyLocationVersions.stream()
+          .map(OmKeyLocationInfoGroup::getProtobuf)
+          .collect(Collectors.toList()));
+    }
```
adoroszlai, line 432: OmKeyInfo was changed in #1034 to avoid streams.  `keyLocationVersions` is now converted in a loop:

https://github.com/apache/hadoop-ozone/blob/2af6198686d81daa3ad0513f723118637d2945cf/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java#L388-L391

That's where normal/compact protobuf conversion should happen depending on `ignorePipeline` parameter.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+ARG BASE=latest
+FROM flokkr/base:35
```
adoroszlai, line 2: This dockerfile seems to be unrelated to the change in this PR.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyLocationInfoGroup.java
```diff
@@ -58,11 +58,22 @@ public long getVersion() {
...
+  public KeyLocationList getCompactProtobuf() {
+    return KeyLocationList.newBuilder()
+        .setVersion(version)
+        .addAllKeyLocations(
+            locationList.stream().map(
+                OmKeyLocationInfo::getCompactProtobuf)
+                .collect(Collectors.toList()))
+        .build();
+  }
+
```
adoroszlai, line 70: Can we avoid duplication by extracting to a helper method that takes `OmKeyLocationInfo` conversion function (`OmKeyLocationInfo::getCompactProtobuf` or `OmKeyLocationInfo::getProtobuf`)?

adoroszlai (dismissed):
It seems this needs to be slightly updated due to recent change on `master`.

ChenSammi:
@elek and @adoroszlai , thanks for the review and feedback. I have rebased and updated the patch accordingly.

ChenSammi:
Thanks @elek  and @adoroszlai  for review the code.

PR #1013, opened 2020-06-03
HDDS-3703. Avoid file lookup calls in writeChunk hotpath

adoroszlai:

1. Skip the check for existence of chunks directory that happened on each `getChunkFile` call.  Read and delete operations already verify existence of the chunk file itself.
2. Instead of the above check, verify the existence of the block file after write once per block.  This is needed to allow marking the container unhealthy immediately if file is lost during write, see existing unit test in `TestContainerStateMachine`.
3. Perform `finishWriteChunk` in batch, avoiding unnecessary conversion of all chunks, operate on already converted block data.
4. Replace "absolute path" with "path" as key in `OpenFiles` (for file-per-block).  This is safe because chunk dir location is constructed relative to the container dir (and eventually relative to HDDS volume), so the result is the same.

https://issues.apache.org/jira/browse/HDDS-3703

## How was this patch tested?

Tested locally on docker-compose cluster.

https://github.com/adoroszlai/hadoop-ozone/runs/733420600

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ChunkLayOutVersion.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/ChunkLayOutVersion.java
@@ -41,14 +41,14 @@ public enum ChunkLayOutVersion {
   FILE_PER_CHUNK(1, "One file per chunk") {
     @Override
     public File getChunkFile(File chunkDir, BlockID blockID,
-        ChunkInfo info) throws StorageContainerException {
-      return chunkDir.toPath().resolve(info.getChunkName()).toFile();
+        ChunkInfo info) {
+      return new File(chunkDir, info.getChunkName());
     }
   },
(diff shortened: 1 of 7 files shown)


adoroszlai:
> "FilePerBlockStrategy::writeChunk-->validateChunkForOverwrite" possibly can be optimised, as "isOverWritePermitted" would be false most of the times.

Unfortunately `validateChunkForOverwrite` allows overwrite regardless of the `OverWriteRequested` flag.  The only difference is a warning in the log.

bshashikant:
> "FilePerBlockStrategy::writeChunk-->validateChunkForOverwrite" possibly can be optimised, as "isOverWritePermitted" would be false most of the times.

Overwrite can happen bcoz of ratis irrespective of whether isOverWritePermitted is permitted or not.

bshashikant:
Thanks @adoroszlai for the contribution. I have committed this.

adoroszlai:
Thanks @bshashikant for reviewing and merging this.

PR #1018, opened 2020-06-04
HDDS-3716. Compile of Ozone fails with JDK 11+

elek:

@PostConstruct annotation is removed from JDK (it's Java EE) in the recent JDKs.

It's used by the Configuration annotations, but we don't need to use Java EE annotations:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project hadoop-hdds-config: Compilation failure: Compilation failure: 
[ERROR] /Users/sbanerjee/ozone_fork/hadoop-hdds/config/src/main/java/org/apache/hadoop/hdds/conf/ConfigurationReflectionUtil.java:[20,24] cannot find symbol
[ERROR]  symbol:  class PostConstruct
[ERROR]  location: package javax.annotation
[ERROR] /Users/sbanerjee/ozone_fork/hadoop-hdds/config/src/main/java/org/apache/hadoop/hdds/conf/ConfigurationReflectionUtil.java:[139,38] cannot find symbol
[ERROR]  symbol:  class PostConstruct
```

You can guess it, but it's:

https://issues.apache.org/jira/browse/HDDS-3716

## How was this patch tested?

Compiled ozone locally with java13.

### DIFF
--- a/hadoop-hdds/common/pom.xml
+++ b/hadoop-hdds/common/pom.xml
@@ -47,7 +47,13 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <artifactId>guava</artifactId>
       <scope>compile</scope>
     </dependency>
-
+    <!-- for the annotations in OzoneConfiguration -->
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-impl</artifactId>
+      <version>2.1.13</version>
+      <scope>provided</scope>
+    </dependency>
     <dependency>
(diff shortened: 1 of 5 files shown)


elek:
Thanks the review @arp7 and @adoroszlai (and the report @bshashikant). I am merging it now.

BTW eventually we can switch to use java11 to compile (I am talking about the used JVM, required API level is a different discussion). Sonar is complaining that JDK8 won't be supported after September 2020

PR #1023, opened 2020-06-04
HDDS-3718: Improve OmKeyLocationInfoGroup internal data structure

esahekmat:
OmKeyLocationInfoGroup changed to store a Map between versions and list of OmKeyLocationInfos
instead of storing all locationInfos in a list, by this change, the performance of getting the list of the latest version improved.

https://issues.apache.org/jira/browse/HDDS-3718

## How was this patch tested?
ozone common module tests


arp7:
Thanks for the contribution @esahekmat. Do you know if this structure is passed on the wire or persisted to disk i.e. will the change be incompatible?

cc @bharatviswa504 @melek

esahekmat:
I have tried to preserve the getProtobuf and getFromProtobuf methods and I have kept the KeyLocationList structure intact, does it enough to prevent incompatibility?

xiaoyuyao:
I agree the structure pass over the wire does not change with this PR. However, given the version support is not fully added, we should have only limited(one) version here. Change the list to map will not bring in much performance improvement. Should we postpone  the change when adding version support?

esahekmat:
I agree the change can be postponed until the version support implementation, however storing different version in a heterogeneous List, and filter it for specific version led to poor design, for example, searching for specific version in the list, some use-cases of it had been done in OmKeyLocationInfoGroup itself and some of them had been done in other class (see getBlocksLatestVersionOnly in OmKeyLocationInfoGroup and addPreallocateBlocks in BlockOutputStreamEntryPool)

arp7:
@elek if you've confirmed there is no change to on disk structures and no potential incompatibility, then I am +1 to proceed.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyLocationInfoGroup.java
```diff
@@ -30,12 +30,23 @@
...
+    this.locationVersionList = locations.stream()
+        .collect(Collectors.groupingBy(OmKeyLocationInfo::getCreateVersion));
+    //prevent NPE
+    this.locationVersionList.putIfAbsent(version, new ArrayList<>());
+  }
+
+  public OmKeyLocationInfoGroup(long version,
+                                Map<Long, List<OmKeyLocationInfo>> locations) {
+    this.version = version;
+    this.locationVersionList = locations;
```
xiaoyuyao, line 47: should we change the order of 47/49 so that there is always an ArrayList for the version and the locations is added to that arraylist.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyLocationInfoGroup.java
```diff
@@ -30,12 +30,23 @@
  */
 public class OmKeyLocationInfoGroup {
   private final long version;
-  private final List<OmKeyLocationInfo> locationList;
+  private final Map<Long, List<OmKeyLocationInfo>> locationVersionList;
```
xiaoyuyao, line 33: NIT: should we rename to locationVersionMap to avoid confusion?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyLocationInfoGroup.java
esahekmat: It will cause a redundant instantiation of the map and provides no benefit because the map reassigned by the constructor parameter.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyLocationInfoGroup.java
esahekmat: Sure, I will change it

esahekmat:
@xiaoyuyao Thanks for reviewing it again, I have answered your concerns, would you please take another look to the it

xiaoyuyao:
+1, Thanks @esahekmat for the contribution. I will merge the PR shortly.

PR #1025, opened 2020-06-05
HDDS-3732. Avoid UUID#toString call in DatanodeDetails#getUuidString.

runzhiwang:
UUID.toString in DatanodeDetails cost 7.23% cpu because of so many datanodes.
image

https://issues.apache.org/jira/browse/HDDS-3732

## How was this patch tested?

Existed UT and IT.


nandakumar131:
Thanks @runzhiwang for the improvement.
We can also replace the `uuid.toString()` call inside `DatanodeDetails#toString()` to `strUuid`.

runzhiwang:
@nandakumar131 Thanks for review. I have updated.

nandakumar131:
Failure doesn't seem related, re-triggered the pre-commit checks.

nandakumar131:
Thanks @runzhiwang for the improvement.

PR #1026, opened 2020-06-05
HDDS-3330. TestDeleteWithSlowFollower is still flaky.

bshashikant:

Increased the watch timeout so that key.close() succeeds on the same pipeline with majority commit. Added some precondition checks to validate the follower.

https://issues.apache.org/jira/browse/HDDS-3330

## How was this patch tested?
manual running the test.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestDeleteWithSlowFollower.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestDeleteWithSlowFollower.java
@@ -66,7 +66,6 @@ import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_STALENODE_INTER
 import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
 import org.junit.Test;
 
 /**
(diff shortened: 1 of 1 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestDeleteWithSlowFollower.java
```diff
@@ -221,7 +223,9 @@ public void testDeleteKeyWithSlowFollower() throws Exception {
     }
     Assert.assertNotNull(follower);
     Assert.assertNotNull(leader);
-    // shutdown the slow follower
+    //ensure that the chosen follower is still a follower
+    Assert.assertTrue(ContainerTestHelper.isRatisFollower(follower, pipeline));
```
adoroszlai, line 227: Do we want to fail the test if `follower` is no longer a follower, or should we just ignore test result in that case?  If the latter, it may be better to change this to `Assume.assumeTrue`.

bshashikant: There are no failures being induced till this check so, follower should still be a follower. I would prefer to fail the test in that case.

bshashikant:
Thanks @adoroszlai for the review. I have committed this.

PR #1027, opened 2020-06-05
HDDS-3734. Improve the performance of SCM with 3.86% by avoid TreeSet.addAll

runzhiwang:

TreeSet.addAll cost 3.86% cpu because of so many datanodes.

we need not TreeSet.addAll, because the DatanodeDetails was got from NodeStateMap::ConcurrentHashMap<NodeState, Set<UUID>> stateMap, the value type of stateMap is a set, so there is no duplicate DatanodeDetails. we can replace TreeSet.addAll with list.addAll

企业微信截图_15913533425809

https://issues.apache.org/jira/browse/HDDS-3734

## How was this patch tested?

Existed UT and IT.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
@@ -27,8 +27,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 import org.apache.hadoop.fs.CommonConfigurationKeys;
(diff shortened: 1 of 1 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
```diff
@@ -589,15 +589,15 @@ public boolean getSafeModeStatus() {
...
+   * @return List of Datanodes that match the NodeState.
    */
-  private Set<DatanodeDetails> queryNodeState(HddsProtos.NodeState nodeState) {
-    Set<DatanodeDetails> returnSet = new TreeSet<>();
+  private List<DatanodeDetails> queryNodeState(HddsProtos.NodeState nodeState) {
+    List<DatanodeDetails> returnList = new ArrayList<>();
     List<DatanodeDetails> tmp = scm.getScmNodeManager().getNodes(nodeState);
     if ((tmp != null) && (tmp.size() > 0)) {
-      returnSet.addAll(tmp);
+      returnList.addAll(tmp);
```
adoroszlai, line 598: I think `returnList` is no longer needed, `tmp` can be returned from here directly.

bharatviswa504:
Thank You @runzhiwang for the improvement and @adoroszlai for the review.

PR #1028, opened 2020-06-05
HDDS-3735. Improve SCM performance with 3.7% by remove unnecessary lock and unlock

runzhiwang:

I start a ozone cluster with 1000 datanodes, and run two weeks with heavy workload, and perf scm.  lock and unlock in 3 places cost about 3.7% cpu, but the lock and unlock is unnecessary, because read lock is to avoid read incomplete data when write, but here get value from ConcurrentHashMap is thread safe, we do not need read lock. So this patch remove these lock.

1. lock and unlock in ContainerStateMap::getContainerInfo
image

image

2. lock and unlock in ContainerStateMap::getContainerReplicas
image

image

3. lock and unlock in NodeStateMap::getNodeInfo
image

image

https://issues.apache.org/jira/browse/HDDS-3735

## How was this patch tested?

Existed UT and IT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManager.java
@@ -495,6 +495,15 @@ public class ContainerStateManager {
     return containers.getContainerInfo(containerID);
   }
 
+  /**
+   * Returns whether container exist for the given container id.
+   * @param containerID id of the container
+   * @return true if container exists, otherwise false
+   */
+  boolean isContainerExist(final ContainerID containerID) {
+    return containers.isContainerExist(containerID);
+  }
+
   void close() throws IOException {
   }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
@@ -204,9 +204,7 @@ public class SCMContainerManager implements ContainerManager {
   public boolean exists(ContainerID containerID) {
     lock.lock();
     try {
-      return (containerStateManager.getContainer(containerID) != null);
-    } catch (ContainerNotFoundException e) {
-      return false;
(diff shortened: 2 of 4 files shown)


runzhiwang:

elek:
Actually I am not sure about the other part.

Using read / write locks in the `ContainerStateMap` provides a consistent view. Read calls can see any containers only if all steps from the write method are done (all the indexes are updated).

The patch seems to introduce a consistency problem. If you don't have a read lock the write may not be finished when you read. The  `containerMap` itself is ok, but if you do any operation with any other indexes from `ContainerStateMap` you might see a different view.

Did I miss something? @runzhiwang

runzhiwang:
@elek Thanks for review. You are right, if the write code like following, maybe after  `currentInfo.setState(newState)` and before `currentInfo.setOwner(owner)`, we read the currentInfo, so we read a half write ContainerInfo. 
```
lock.writeLock().lock();
final ContainerInfo currentInfo = containerMap.get(containerID);
currentInfo.setState(newState);
currentInfo.setOwner(owner);
lock.writeLock().unlock();
```

I think we can avoid this by the following code. Though write become more complicated because it needs to new an object, but read is pretty more than write, so it's worth. What do you think ?
```
lock.writeLock().lock();
final ContainerInfo currentInfo = containerMap.get(containerID);
ContainerInfo newInfo = new ContainerInfo(currentInfo);
newInfo.setState(newState);
newInfo.setOwner(owner);
containerMap.put(containerID, tmpInfo);
lock.writeLock().unlock();
```

elek:
It's not just about the `updateState` method, but (for example) the `removeContainer` method.

Without write lock the container might be removed from `ownerMap` and not from the `typeMap`, which can cause inconsistency.

(I think, but fix me if I am wrong)

runzhiwang:
> Without write lock the container might be removed from ownerMap and not from the typeMap, which can cause inconsistency.

I think this case can not be avoided by only add read lock. Unless we read containerMap, ownerMap, typeMap in the single method with a read lock, Otherwise for example 1. we read  containerMap, 2. we remove container, 3. we read ownerMap. It will also cause inconsistency, because container exists in step1 but not exists in step3.

elek:
> @elek I think this case can not be avoided by only add read lock

Agree. It seems to be necessary. But it means that we can't remove the lock in the ContainerStateMap in this patch.

PR #1030, opened 2020-06-06
HDDS-3742. Improve OM performance with 5.29% by avoid stream.collect

runzhiwang:

I start a ozone cluster with 1000 datanodes, and run two weeks with heavy workload, and perf om.

stream().collect(Collectors.toList())) in getProtoBuf cost 5.29% cpu.

image

https://issues.apache.org/jira/browse/HDDS-3742

## How was this patch tested?

Existed test


bharatviswa504:
Thank You @runzhiwang for the performance improvement.

PR #1031, opened 2020-06-06
HDDS-3745. Improve OM and SCM performance with 64% by avoid collect datanode information  to s3g

runzhiwang:

**What's the problem ?**
I start a ozone cluster with 1000 datanodes and 10 s3gateway, and run two weeks with heavy workload, and perf om and scm.
1. From om perf, getServiceList cost 63.75% cpu.
image

2.From scm perf, queryNode come from om::getServiceList cost 33.20% cpu
image

**What's the reason ?**
Now s3g create a client for each request (we should improve it in the future). when create each `RpcClient`, s3g will call `ozoneManagerClient.getServiceInfo()`, `getServiceInfo` will call `getServiceList`.  `getServiceList` is a very heavy call, because it need to get DatanodeDetails of 1000 Datanodes, so om and scm are very busy with `getServiceList`.

But s3g does not use the List(ServiceInfo) which got from `getServiceList` at all.

https://issues.apache.org/jira/browse/HDDS-3745

## How was this patch tested?

Existed tests.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerRestInterface.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerRestInterface.java
@@ -21,7 +21,6 @@ package org.apache.hadoop.ozone.om;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.core.type.TypeReference;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
-import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.net.NetUtils;
 import org.apache.hadoop.ozone.MiniOzoneCluster;
 import org.apache.hadoop.ozone.om.helpers.ServiceInfo;
@@ -114,30 +113,6 @@ public class TestOzoneManagerRestInterface {
     Assert.assertEquals(scmAddress.getHostName(), scmInfo.getHostname());
     Assert.assertEquals(scmAddress.getPort(),
         scmInfo.getPort(ServicePort.Type.RPC));
-
-    ServiceInfo datanodeInfo = serviceMap.get(HddsProtos.NodeType.DATANODE);
-    DatanodeDetails datanodeDetails = cluster.getHddsDatanodes().get(0)
-        .getDatanodeDetails();
-    Assert.assertEquals(datanodeDetails.getHostName(),
-        datanodeInfo.getHostname());
-
-    Map<ServicePort.Type, Integer> ports = datanodeInfo.getPorts();
-    for(ServicePort.Type type : ports.keySet()) {
-      switch (type) {
-      case HTTP:
-      case HTTPS:
-        Assert.assertEquals(
-            datanodeDetails.getPort(DatanodeDetails.Port.Name.REST).getValue(),
-            ports.get(type));
-        break;
-      default:
-        // OM only sends Datanode's info port details
-        // i.e. HTTP or HTTPS
-        // Other ports are not expected as of now.
-        Assert.fail();
-        break;
-      }
-    }
   }
 
 }
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -54,7 +54,6 @@ import org.apache.hadoop.hdds.HddsConfigKeys;
 import org.apache.hadoop.hdds.HddsUtils;
 import org.apache.hadoop.hdds.annotation.InterfaceAudience;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
-import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.protocol.proto.SCMSecurityProtocolProtos.SCMGetCertResponseProto;
 import org.apache.hadoop.hdds.protocolPB.SCMSecurityProtocolClientSideTranslatorPB;
@@ -175,7 +174,6 @@ import static org.apache.hadoop.hdds.HddsConfigKeys.HDDS_BLOCK_TOKEN_ENABLED;
 import static org.apache.hadoop.hdds.HddsConfigKeys.HDDS_BLOCK_TOKEN_ENABLED_DEFAULT;
 import static org.apache.hadoop.hdds.HddsUtils.getScmAddressForBlockClients;
 import static org.apache.hadoop.hdds.HddsUtils.getScmAddressForClients;
-import static org.apache.hadoop.hdds.protocol.proto.HddsProtos.NodeState.HEALTHY;
 import static org.apache.hadoop.hdds.security.x509.certificates.utils.CertificateSignRequest.getEncodedString;
 import static org.apache.hadoop.hdds.server.ServerUtils.getRemoteUserName;
 import static org.apache.hadoop.hdds.server.ServerUtils.updateRPCListenAddress;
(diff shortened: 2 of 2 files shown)


hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
```diff
@@ -170,10 +169,10 @@ public RpcClient(ConfigurationSource conf, String omServiceId)
         OzoneManagerProtocol.class, conf
     );
     dtService = omTransport.getDelegationTokenService();
-    ServiceInfoEx serviceInfoEx = ozoneManagerClient.getServiceInfo();
+    String certificate = ozoneManagerClient.getCaCertificate();
     String caCertPem = null;
     if (OzoneSecurityUtil.isSecurityEnabled(conf)) {
-      caCertPem = serviceInfoEx.getCaCertificate();
+      caCertPem = certificate;
```
GlenGeng, line 175: how about 
```
caCertPem = ozoneManagerClient.getCaCertificate();
```

only invoke when security is enabled.

runzhiwang: CI does not passed if we do this, maybe CI want ozoneManagerClient.getCaCertificate throw exception even if security is not enabled. I will find the reason.

runzhiwang:

nandakumar131:
**Some background**
`gerServiceInfo` call was introduced as part of `Service Discovery API` (HDFS-12868).
The idea behind `Service Discovery API` was to not worry about setting multiple properties at client to connect to Ozone.
Point the client to any one of the OzoneManager (even in case of HA), the client should be able to get all the required configuration and be able to connect to Ozone.
This will make the client configuration simple.

When `gerServiceInfo` was introduced, Ozone client just needed OM's IP and port to connect, nothing else. Over time we made lot of code change and never revisited `Service Discovery API`.
After OM HA, it is no longer true that we can discover other services and required configurations just with single OM's IP/Port. We need `ozone-site.xml` at client and set of properties that will allow us to interact with OM HA.

Now I'm not even sure where and why `gerServiceInfo` is getting used.

bharatviswa504:
> **Some background**
> `gerServiceInfo` call was introduced as part of `Service Discovery API` (HDFS-12868).
> The idea behind `Service Discovery API` was to not worry about setting multiple properties at client to connect to Ozone.
> Point the client to any one of the OzoneManager (even in case of HA), the client should be able to get all the required configuration and be able to connect to Ozone.
> This will make the client configuration simple.
> 
> When `gerServiceInfo` was introduced, Ozone client just needed OM's IP and port to connect, nothing else. Over time we made lot of code change and never revisited `Service Discovery API`.
> After OM HA, it is no longer true that we can discover other services and required configurations just with single OM's IP/Port. We need `ozone-site.xml` at client and set of properties that will allow us to interact with OM HA.
> 
> Now I'm not even sure where and why `gerServiceInfo` is getting used.

For OM HA, the reason for not using the getServiceInfo was if client knows one of the OzoneManager addresses, and if that is down, client will not be able to talk to the ozone cluster.
Even right now getServiceInfo, will now return OM HA address with role, but this information cannot be used for the above reason, so the client still need a quorum of OM address.

nandakumar131:
@bharatviswa504 I understand the reason behind it.

We should revisit `Service Discovery API` and make a decision on whether we can make it work or should we remove it completely.

bharatviswa504:
From the code point of I see the usage of getServiceInfo used to get CaCert and used by XceiverClientRatis and Grpc.

```
    ServiceInfoEx serviceInfoEx = ozoneManagerClient.getServiceInfo();
    String caCertPem = null;
    if (OzoneSecurityUtil.isSecurityEnabled(conf)) {
      caCertPem = serviceInfoEx.getCaCertificate();
    }
```

bharatviswa504:
> @bharatviswa504 I understand the reason behind it.
> 
> We should revisit `Service Discovery API` and make a decision on whether we can make it work or should we remove it completely.

>After OM HA, it is no longer true that we can discover other services and required >configurations just with single OM's IP/Port. We need ozone-site.xml at client and set of .properties that will allow us to interact with OM HA.

Agreed. I just want to give you a context for the above statement mentioned.

runzhiwang:
@nandakumar131 @bharatviswa504  Hi, the problem of getServiceInfo is that it include all the datanodes' information.
image

xiaoyuyao:
@runzhiwang the reason the serviceinfo request retrieve all the DNs is for the DN REST service address. After the Ozone REST support was removed a while back, those leftover are not cleaned up completely. Maybe a time to clean that up to make it lighter without introducing a new RPC?

runzhiwang:
@xiaoyuyao Thanks for explanation, I agree and I will clean that up.

bharatviswa504:
> @elek @nandakumar131 @bharatviswa504 Hi, Thanks for review. The problem of getServiceInfo is that it contains all the datanodes' information which cost too much if the cluster is big. If it only contain OM address, there is no need to replace it with `getCaCertificate`.
> image

Thank You @runzhiwang for the explanation of the issue. I like @xiaoyuyao proposal, if DN information is no more required, we can clean that up, and use getServiceInfo instead of a new RPC call.

runzhiwang:
@xiaoyuyao @bharatviswa504 Thanks for review. I have updated the patch.

bharatviswa504:
Thank You for the update.
Can we also make the change in RpcClient.java, calling getServiceInfo only when security is enabled?

elek:
>  Can we also make the change in RpcClient.java, calling getServiceInfo only when security is enabled?

I have some fear if the secre/unsecure cluster would have such difference. For example, I should run all the performance tests on secure cluster. We should be careful and execute all the tests on both secure / unsecure cluster. Seems to be safer to keep it as is. But I can be convinced.

xiaoyuyao:
The getServiceInfo could be used for non-secure configuration from the server in future. I agree with @elek to leave it as-is.

bharatviswa504:
> > Can we also make the change in RpcClient.java, calling getServiceInfo only when security is enabled?
> 
> I have some fear if the secre/unsecure cluster would have such difference. For example, I should run all the performance tests on secure cluster. We should be careful and execute all the tests on both secure / unsecure cluster. Seems to be safer to keep it as is. But I can be convinced.

From code, I see that we use getServiceInfo value only when security is enabled. So, from S3G perspective, for each request, we can save one Rpc Call.


```
    ServiceInfoEx serviceInfoEx = ozoneManagerClient.getServiceInfo();
    String caCertPem = null;
    if (OzoneSecurityUtil.isSecurityEnabled(conf)) {
      caCertPem = serviceInfoEx.getCaCertificate();
    }
```

I have not understood, what is the problem mentioned here, as from code we can see the value from getServiceInfo is only used when security is enabled.

If someone has mistakenly used some of the information from getServiceInfo mistakenly with out initializing it, it will be caught immediately as it will throw NPE. And also we are not initializing serviceinfo, to any class parameter, it is locally used in the RpcClient constructor only.

xiaoyuyao:
bq. I have not understood, what is the problem mentioned here, as from code we can see the value from getServiceInfo is only used when security is enabled.

getServiceInfo is the service discovery API for OM, which is not for security only. If we go with the other route, i.e., a dedicated API for getCaCertificate(), I'm OK with including check for security enabled.

bharatviswa504:
But right now getServiceInfo is used only to getCaCertificate, other than that in current code there is no usage of it. As now the client does not need an SCM address, for OM, we get a quorum of address from config.

If we can skip getServiceInfo call in a non-secure cluster where we can save one RPC Call. I am fine with reusing this API or with totally a new API.

elek:
So what's next here? I propose to merge this as is (clean simple patch, just remove the datanode part). 

If `getServiceInfo()` is not required anymore: we can remove it fully in a separated patch. (Not just the usage, but we can deprecate and/or remove the server side parts. 

@bharatviswa504, @xiaoyuyao  Are you ok with this approach?

bharatviswa504:
Fine with me. In this Jira let's focus on removing datanode part.

+1 LGTM.

bharatviswa504:
Thank You @runzhiwang for the contribution and @elek and @xiaoyuyao for the review.

PR #1032, opened 2020-06-06
HDDS-3682. Recon UI: Add interactive visualization for file size counts

vivekratnavel:

- Add visualization for file size count distribution
- Add volume & bucket level filters for a drill-down view of the file size distribution

https://issues.apache.org/jira/browse/HDDS-3682

## How was this patch tested?

Manual testing in a docker-compose cluster and using mock API server.

Screen Shot 2020-06-06 at 12 16 21 AM

### DIFF
--- a/hadoop-ozone/recon/src/main/resources/webapps/recon/ozone-recon-web/LICENSE
+++ b/hadoop-ozone/recon/src/main/resources/webapps/recon/ozone-recon-web/LICENSE
@@ -304,6 +304,32 @@ http://creativecommons.org/publicdomain/zero/1.0/.
 
 -----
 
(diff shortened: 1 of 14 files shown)


avijayanhwx:
Thank you for working on this @vivekratnavel. The widget looks good, and will be useful. I deployed the changes locally, and tested with a few inputs. A couple of questions

1. The only way to close the volume and bucket drop down is to click on the 'V' icon in the drop down box. IMO, the intuitive way would be to have the dropdown close if we click outside. Is this a plotly library limitation? 
2. Is it possible to have a refresh icon that refreshes the histogram with the current volume and bucket selections maintained? Is there a way to do this already?

vivekratnavel:
Thanks for the review @avijayanhwx 
> 1. The only way to close the volume and bucket drop down is to click on the 'V' icon in the drop down box. IMO, the intuitive way would be to have the dropdown close if we click outside. Is this a plotly library limitation?

I also experienced the same bug and upon researching a bit, it seems like a temporary bug in react-select for multi-select dropdowns. https://github.com/JedWatson/react-select/issues/3756

I will keep an eye on the open issue against react-select and fix it in Recon once that issue gets resolved. Unfortunately, react-select is one of the most popular select dropdown components for React and there are no other alternatives that offer the features of react-select. 

> 2. Is it possible to have a refresh icon that refreshes the histogram with the current volume and bucket selections maintained? Is there a way to do this already?

We can maintain the volume and bucket selection in browser's local storage, but the problem there will be to handle any changes to selected volume or bucket. For example, if the selected volume or bucket gets deleted, then it needs to be handled. And, for this use case alone, we will have to build a component to persist data to local storage. I prefer not to persist the selection information at-least at this point. We can consider adding persistence to local storage in the future.

avijayanhwx:
> Thanks for the review @avijayanhwx
> 
> > 1. The only way to close the volume and bucket drop down is to click on the 'V' icon in the drop down box. IMO, the intuitive way would be to have the dropdown close if we click outside. Is this a plotly library limitation?
> 
> I also experienced the same bug and upon researching a bit, it seems like a temporary bug in react-select for multi-select dropdowns. [JedWatson/react-select#3756](https://github.com/JedWatson/react-select/issues/3756)
> 
> I will keep an eye on the open issue against react-select and fix it in Recon once that issue gets resolved. Unfortunately, react-select is one of the most popular select dropdown components for React and there are no other alternatives that offer the features of react-select.
> 
> > 1. Is it possible to have a refresh icon that refreshes the histogram with the current volume and bucket selections maintained? Is there a way to do this already?
> 
> We can maintain the volume and bucket selection in browser's local storage, but the problem there will be to handle any changes to selected volume or bucket. For example, if the selected volume or bucket gets deleted, then it needs to be handled. And, for this use case alone, we will have to build a component to persist data to local storage. I prefer not to persist the selection information at-least at this point. We can consider adding persistence to local storage in the future.

Thanks for the explanation.

PR #1034, opened 2020-06-08
HDDS-3749. Improve OM performance with 3% by avoid stream.collect

runzhiwang:

I start a ozone cluster with 1000 datanodes and 10 s3gateway, and run two weeks with heavy workload, and perf om. There are 3 places use stream.collect which cost 3% cpu.

1. stream.collect in Pipeline::getFromProtobuf
image

2. stream.collect in OmKeyInfo::getProtobuf
image
image

3. stream.collect in OMKeyCommitRequest::validateAndUpdateCache
image

https://issues.apache.org/jira/browse/HDDS-3749

## How was this patch tested?

Existed tests.


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyInfo.java
```diff
@@ -411,13 +415,19 @@ public static OmKeyInfo getFromProtobuf(KeyInfo keyInfo) {
     if (keyInfo == null) {
       return null;
     }
+
+    List<OmKeyLocationInfoGroup> omKeyLocationInfos = new ArrayList<>();
+    List<KeyLocationList> keyLocationLists = keyInfo.getKeyLocationListList();
+    for (KeyLocationList keyLocationList : keyLocationLists) {
```
nandakumar131, line 421: Temporary variable `keyLocationLists` can be removed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -132,10 +134,13 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
 
-      List<OmKeyLocationInfo> locationInfoList = commitKeyArgs
-          .getKeyLocationsList().stream()
-          .map(OmKeyLocationInfo::getFromProtobuf)
-          .collect(Collectors.toList());
+      List<OmKeyLocationInfo> locationInfoList = new ArrayList<>();
+      List<KeyLocation> keyLocations = commitKeyArgs
+          .getKeyLocationsList();
+
+      for (KeyLocation keyLocation : keyLocations) {
```
nandakumar131, line 141: The temporary variable `keyLocations ` can be removed.

runzhiwang:
@nandakumar131 Thanks for review. I have updated.

elek:
Thanks the quick fix @bharatviswa504. 

Do you have any idea how was it possible? The last build seems to be green... What should we improve on our process to avoid some problems?

runzhiwang:
@elek @bharatviswa504    I think the reason there are code1 and code2 both use java.util.stream.Collectors.   In this patch, I remove code1, and trigger CI, because code2 still use java.util.stream.Collectors, CI does not fail. Then, another commit remove code2, but this patch does not trigger CI again, if trigger, CI will fail, because both code1 and code2 were removed, there are no code use java.util.stream.Collectors. So when merge this patch, CI fail.   Sorry for that.

PR #1038, opened 2020-06-08
HDDS-3754. Rename framework to common-server

elek:

I started to document the project hierarchy of Ozone and found that to subproject name is very confusing:

 1. framework: it supposed to be similar to the common but share the classes only between server side projects (om/scm/...). I propose to call it server-common

 2. container-service: While we have historical reasons why it's a separated project, but today we can call it datanode and remove hadoop-ozone/datanode.

While I can learn both the specific names, for new contributors it can help if we use more straightforward names.

In this patch I would like to fix the first item (framework --> common-server)

https://issues.apache.org/jira/browse/HDDS-3754

## How was this patch tested?

Full CI run is passed on my fork.


nandakumar131:
Having `framework` subproject makes sense as it can provide

* Lock service/framework
* Lease service/framework
* Security service/framework
* EventQueue service/framework
...

Which can be used by other modules.

`Common` project is a collection of reusable code or utilities which is used by other modules, but with `framework` we can have a clear interface defined which can expose generic services. I agree that the current code is not in this state yet, but we can always improve.

elek:
@nandakumar131 Not sure if understood your proposal. Do you propose to create separated projects for separated framework/services? Like `framework-eventqueue`, or `framework-dbstore`?

I am fine with that. But the current project (which is used only to collect server side utilities and classes) is more like `common` for the server side.

But I am fine to leave it as is, if you don't think it's confusing...

nandakumar131:
> Not sure if understood your proposal. Do you propose to create separated projects for separated framework/services? Like framework-eventqueue, or framework-dbstore?

Sorry if my comment was not very clear. The proposal is to retain `framework` and not rename it to `common`.

Once we move to java 9 or above we can use the `module` feature of java to expose `EventQueue`, `dbstore`, `Lock`, `Lease` etc., as a service from `framework` project.

It makes sense to retain the name `framework` for this sub-project.

elek:
Ok.

For me `framework` was not expressive enough. I found that multiple serverside shared dependency (`common`) are added the `hdds/common` (which is not good as it's shared with the client), and I though because the name is not express the goal of the project.

But I may be the only one, it if it's clear for you. Thanks the feedback. I am closing this PR.

PR #1039, opened 2020-06-08
HDDS-3753. remove redundant synchronized for SCMPipelineManager's createPipeline method

wuweiwei95:

Remove the synchronized qualifier from the createPipeline method since it already uses lock for critical zone protection inside createPipeline.

https://issues.apache.org/jira/browse/HDDS-3753

## How was this patch tested?

 manual tests


nandakumar131:
Thanks for the contribution @wuweiwei95

PR #1042, opened 2020-06-09
HDDS-3759. Avoid UUID#toString call in PipelineID#getProtobuf

runzhiwang:
`UUID.toString` in `PipelineID::getProtobuf` cost 0.13% cpu, and `getContainerWithPipelineBatch` cost 1.76%, so `UUID.toString` cost about 0.13 / 1.76 = 7.4% in `getContainerWithPipelineBatch`. Because `getContainerWithPipelineBatch` happens in the main chain of read, so we need to improve it.

image

image

https://issues.apache.org/jira/browse/HDDS-3759

## How was this patch tested?

Existed tests.


nandakumar131:
Thanks @runzhiwang for the contribution.

PR #1044, opened 2020-06-09
HDDS-3743. Avoid NetUtils#normalize when get DatanodeDetails from proto

runzhiwang:

NetUtils.normalize cost 0.17%, because it needs to do Matcher.replaceAll. It happen in the critical path, i.e. lookUpKey when read. we need not to do this when getFromProtoBuf, So we can avoid normalize by check the last character, if the last character is not `/`, we need not normalize.

image

https://issues.apache.org/jira/browse/HDDS-3743

## How was this patch tested?

Existed tests.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/net/NetUtils.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/net/NetUtils.java
@@ -49,7 +49,12 @@ public final class NetUtils {
    * If <i>path</i>is empty or null, then {@link NetConstants#ROOT} is returned
    */
   public static String normalize(String path) {
-    if (path == null || path.length() == 0) {
+    if (path == null) {
+      return NetConstants.ROOT;
+    }
+
+    int len = path.length();
+    if (len == 0) {
       return NetConstants.ROOT;
     }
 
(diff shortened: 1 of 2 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/net/NetUtils.java
```diff
@@ -59,9 +64,12 @@ public static String normalize(String path) {
               + NetConstants.PATH_SEPARATOR_STR + ": " + path);
     }
 
+    if (len == 1 || path.charAt(len - 1) != NetConstants.PATH_SEPARATOR) {
```
xiaoyuyao, line 67: Can we simplified the logic without any matcher by using apache common StringUtils?

return len == 1 ? path : StringUtils.removeEnd(path, "/");

runzhiwang: @xiaoyuyao Thanks for review. But the result of  StringUtils.removeEnd("/a/b/c//", "/") is "/a/b/c/" instead of "/a/b/c". I'm not sure whether we should consider the path end with multiple "/".

xiaoyuyao: You are correct. The node to location mapping is provided by script or mapping file configured and parsed by implementation of DNSToSwitchMapping. The script should usually normalize the path already. Otherwise,  
the mapping can't  be used consistently for the topology class to determine node locations. 

E.g., multiple slash: "///" => ""
multiple slash in the middle "/a//b/c/" => "/a//b/c" which is not equal to "/a/b/c", etc.

runzhiwang: @xiaoyuyao Thanks for review. Updated.

xiaoyuyao:
Thanks @runzhiwang  for the contribution. I've merged the patch to master.

PR #1050, opened 2020-06-10
HDDS-3757. Add test coverage of the acceptance tests to overall test coverage 

elek:

This patch adds the coverage data from the acceptance test to the generic coverage measurement.

There was one question during the implementation: I decided to add the required HADOOP_OPTS to all the docker-compose file without using tricky docker-compose extension. I found that I need to add a few lines anyway, and I preferred to keep it simple, even if a possible change would require slightly more work (but can be done with easy search and replace)

https://issues.apache.org/jira/browse/HDDS-3757

## How was this patch tested?

Pushed the branch to apache repo and checked sonar cloud.

https://sonarcloud.io/dashboard?branch=HDDS-3757&id=hadoop-ozone


hadoop-ozone/dev-support/checks/acceptance.sh
```diff
@@ -17,19 +17,32 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
 
 if [ ! -d "$DIST_DIR" ]; then
     echo "Distribution dir is missing. Doing a full build"
-    "$DIR/build.sh"
+    "$DIR/build.sh" -Pjacoco
 fi
 
+mkdir -p "$REPORT_DIR"
+
+export HADOOP_OPTS='-javaagent:/opt/hadoop/share/jacoco/jacoco-agent.jar=destfile=/tmp/jacoco.exec,includes=org.apache.hadoop.ozone.*:org.apache.hadoop.hdds'
```
adoroszlai, line 31: Now `acceptance.sh` requires `jacoco-agent.jar`, which is only built when invoked with `-Pjacoco`.  I would like to suggest moving this definition to the GitHub Actions definition (in the `env` section, where `KEEP_IMAGE` is defined, too).

hadoop-ozone/dist/src/main/compose/ozone-csi/docker-compose.yaml
```diff
@@ -23,6 +23,8 @@ services:
       - ../..:/opt/hadoop
     env_file:
       - docker-config
+    environment:
+      HADOOP_OPTS: ${HADOOP_OPTS}
```
adoroszlai, line 27: I think we can avoid these by passing the variable via `-e` option of `docker-compose`, similar to how it is done for some other variables (`SECURITY_ENABLED`, etc.):

https://github.com/apache/hadoop-ozone/blob/53395a0ed75d96575a7dc26a7adec4eefabb2b74/hadoop-ozone/dist/src/main/compose/testlib.sh#L107

elek: We need it for `docker-compose up`. Do you see any possibility to set generic environment variables for all the containers? This approach works only for `docker-compose exec` IMHO.

elek:
> I only see 0.7% increase in coverage between master and HDDS-3757 branches. Also, I see several errors related to jacoco.exec in the log:

Yes, I am also surprised. Let me try to double check the results.

hadoop-ozone/dist/src/main/compose/ozone-csi/docker-compose.yaml
adoroszlai: Sorry, I confused the two commands.

elek:
> Let me try to double check the results.

We don't collect all the coverage data right now. This patch collects to exec files after executions of a robot test. It means that only that specific server (eg. scm) will be measured.

I should modify it to collect the files when compose is shutting down...

Another problem (which is harder) to add the measurement from the CLI. But that can be solved later...

vivekratnavel:
@elek Can we skip building Recon UI by adding `-Dskip.yarn` flag to coverage step's build.sh command in both post-commit and pr workflows?
https://github.com/apache/hadoop-ozone/pull/1050/files#diff-500d30cad2a60d5e5aaa6a95f025ba5fR251

Now that we take arguments for build.sh, this is possible and should save us some time in the CI.

adoroszlai:
> @elek Can we skip building Recon UI by adding `-Dskip.yarn` flag to coverage step's build.sh command in both post-commit and pr workflows?
> Now that we take arguments for build.sh, this is possible and should save us some time in the CI.

Seems OK:
https://github.com/adoroszlai/hadoop-ozone/runs/762799116

vivekratnavel:
The coverage phase runs a full build and Recon takes ~4 mins to build. This can be optimized by including cache like in build or acceptance stage or by completely excluding Recon UI build by including the flag `-Dskip.npx` flag in build.sh command.

elek:
Thanks @vivekratnavel I added this flag.

Quick update:

I found that the current approach is not working very well:

 * Even if the servers are instrumented by agents I couldn't collect the coverage data as they are persisted only when the JVM is shutting down. With docker, it's hard to do as the `docker-compose down` triggers the JVM shutdown and there is no way to save the jacoco files at that time.

The latest patch tries to follow more generic approach: A tcp server is started (on host) to collect all the coverage data.

 1. With this approach we don't need any copy for the jacoco.file-s. It just works
 2. The final jacoco-combined.exec is written by the TCP server.
 3. All the client calls can be measured independent how many client JVM is started

Will see how does it work.

hadoop-ozone/dist/src/main/compose/test-all.sh
```diff
@@ -41,8 +46,13 @@ for test in $(find "$SCRIPT_DIR" -name test.sh | sort); do
       echo "ERROR: Test execution of $(dirname "$test") is FAILED!!!!"
   fi
   RESULT_DIR="$(dirname "$test")/result"
-  cp "$RESULT_DIR"/robot-*.xml "$RESULT_DIR"/docker-*.log "$RESULT_DIR"/*.out* "$ALL_RESULT_DIR"/
+  cp "$RESULT_DIR"/robot-*.xml "$RESULT_DIR"/docker-*.log "$RESULT_DIR"/*.out* "$RESULT_DIR"/*.exec "$ALL_RESULT_DIR"/
 done
 
 rebot -N "smoketests" -d "$SCRIPT_DIR/result" "$SCRIPT_DIR/result/robot-*.xml"
+if [ "$OZONE_WITH_COVERAGE" ]; then
+  cp /tmp/jacoco-combined.exec "$SCRIPT_DIR/result/"
```
adoroszlai, line 54: I don't see any `*.exec` files in [acceptance.zip](https://github.com/apache/hadoop-ozone/suites/812019480/artifacts/8877510) for the latest run.  Am I missing something?

hadoop-ozone/dist/src/main/compose/test-all.sh
```diff
@@ -19,17 +19,22 @@
...
 ALL_RESULT_DIR="$SCRIPT_DIR/result"
-
+PROJECT_DIR="$SCRIPT_DIR/.."
 mkdir -p "$ALL_RESULT_DIR"
-rm "$ALL_RESULT_DIR/*"
+rm "$ALL_RESULT_DIR/*" || true
+
+if [ "$OZONE_WITH_COVERAGE" ]; then
+   java -cp "$PROJECT_DIR"/share/coverage/$(ls "$PROJECT_DIR"/share/coverage | grep test-util):"$PROJECT_DIR"/share/coverage/jacoco-core.jar org.apache.hadoop.test.JacocoServer &
+   DOCKER_BRIDGE_IP=$(docker network inspect bridge --format='{{(index .IPAM.Config 0).Gateway}}')
```
adoroszlai, line 30: Some of the compose environments have a dedicated network.  I think those would have problems starting SCM:

```
hadoop-ozone/dist/src/main/compose/ozone-topology/docker-compose.yaml
hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/docker-compose.yaml
hadoop-ozone/dist/src/main/compose/ozonesecure-mr/docker-compose.yaml
```

(I'm not sure because I have trouble starting any of the environments with coverage enabled.)

elek: > (I'm not sure because I have trouble starting any of the environments with coverage enabled.)

How did you try? Docker bridge ip supposed to be available on all the containers. But can be wrong on OSX. But worked well on github environments.

adoroszlai: IP is available, can be pinged, but connection to port is refused.  Confirmed that Jacoco server is running on localhost:6300.

I'm [experimenting](https://github.com/adoroszlai/hadoop-ozone/commits/HDDS-3757) with running Jacoco in its own container.

adoroszlai:
`HTTP/1.1 403 Forbidden` in Recon Web UI (HDDS-3846) seems to be specific to this change.  Haven't seen it elsewhere.  I'm [running into it](https://github.com/adoroszlai/hadoop-ozone/runs/806226271) even with my changes.

vivekratnavel:
@adoroszlai If Recon Frontend is not being built, this is possible.

adoroszlai:
> If Recon Frontend is not being built, this is possible.

Thanks, I'll give it a try without `-Dskip.npx`.

adoroszlai:
> If Recon Frontend is not being built, this is possible.

Thanks for the hint.  I can confirm that this was the problem.  The reason I couldn't reproduce it locally was that `node_modules` etc. are not wiped by `mvn clean`.  So despite using `-Dskip.npx`, I did have some Recon UI locally and the test was passing.

hadoop-hdds/test-utils/src/main/java/org/apache/hadoop/test/JacocoServer.java
```diff
@@ -0,0 +1,83 @@
...
+          RemoteControlWriter writer =
+              new RemoteControlWriter(socket.getOutputStream());
+          RemoteControlReader reader =
+              new RemoteControlReader(socket.getInputStream());
+          reader.setSessionInfoVisitor(destination::visitSessionInfo);
+          reader.setExecutionDataVisitor(destination::visitClassExecution);
+          while (reader.read()) {
+            ;//read until the end of the stream.
+          }
+          destination.flush();
```
adoroszlai, line 68: I've found that these calls to `ExecutionDataWriter`'s methods should be [synchronized](https://github.com/adoroszlai/hadoop-ozone/commit/7d21d2c5f7147e879397453f57e3d8d67e91cc99) to avoid producing garbage files:

```
Exception in thread "main" java.io.UTFDataFormatException: malformed input around byte 2
  at java.io.DataInputStream.readUTF(DataInputStream.java:634)
  at java.io.DataInputStream.readUTF(DataInputStream.java:564)
  at org.jacoco.cli.internal.core.data.ExecutionDataReader.readExecutionData(ExecutionDataReader.java:149)
```

or similar `Unknown block type` error.

hadoop-ozone/dev-support/checks/acceptance.sh
elek: I think this is fixed. I prefer to keep the logic in the shell scripts to keep our build as github independent as possible, but with the improved approach this is done (ENV in the github file, logic in the script)

hadoop-hdds/test-utils/src/main/java/org/apache/hadoop/test/JacocoServer.java
elek: Thanks to investigate it. I uploaded a synchronized version. 

The main problem is that the same functionality is written in Jacoco project under LGPL which couldn't be imported. I wrote my own version from scratch, and this mistake clearly proves that it's independent, as I added my own mistakes ;-)

adoroszlai: `destination.flush();` should be synced, too.

elek:
Finally, acceptance tests seem to be stable.

hadoop-hdds/test-utils/src/main/java/org/apache/hadoop/test/JacocoServer.java
elek: added.

elek:
Thanks the review and the continuous support @adoroszlai (and @vivekratnavel). 

I am merging this now to the master. Weekend seems to be an ideal time for merge this as we can revert in case of any instability (we have 4 scheduled build until Monday)

elek:
Note: if it's stable on the master, we can enable it for the pr builds, too.

PR #1053, opened 2020-06-10
HDDS-3737. Avoid serialization between UUID and String

runzhiwang:
**What's the problem ?**

Serialization between UUID and String: `UUID.toString` and `UUID.fromString`, not only cost cpu, because encode/decode String and UUID.fromString/UUID.toString both cost cpu, but also make the proto bigger, because uuid is just a number which is 16Byte, covet it to string need 32Byte.

**How to fix ?**
Convert UUID to 2 long number in proto, which make serialization and deserialization UUID  faster, and make proto smaller. As the image shows, JDK implement UUID with two long number: `mostSigBits` and `leastSigBits`, when `UUID.fromString`, JDK get `mostSigBits` and `leastSigBits` from String, and new a object of UUID, so we can covert UUID to 2 long number directly.

I only change 3 places of UUID which is  hot spot.

image

image

https://issues.apache.org/jira/browse/HDDS-3763

## How was this patch tested?

Existed tests.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
@@ -59,7 +59,6 @@ public final class Pipeline {
   private ThreadLocal<List<DatanodeDetails>> nodesInOrder = new ThreadLocal<>();
   // Current reported Leader for the pipeline
   private UUID leaderId;
-  private String strLeaderID = "";
   // Timestamp for pipeline upon creation
   private Instant creationTimestamp;
 
@@ -147,9 +146,6 @@ public final class Pipeline {
    */
   void setLeaderId(UUID leaderId) {
     this.leaderId = leaderId;
-    if (leaderId != null) {
-      this.strLeaderID = leaderId.toString();
-    }
   }
 
   /**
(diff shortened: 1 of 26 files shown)


runzhiwang:
@elek @xiaoyuyao Could you help review this patch ? Thank you very much.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
```diff
@@ -301,14 +306,21 @@ public static Pipeline getFromProtobuf(HddsProtos.Pipeline pipeline)
     for (DatanodeDetailsProto member : pipeline.getMembersList()) {
       nodes.add(DatanodeDetails.getFromProtoBuf(member));
     }
+    UUID leaderId = null;
+    if (pipeline.hasLeaderID() &&
+        StringUtils.isNotEmpty(pipeline.getLeaderID())) {
+      leaderId = UUID.fromString(pipeline.getLeaderID());
```
xiaoyuyao, line 312: Correct me if I'm wrong, with current approach, the improvement will be limited as fromProtoBuf to String will prefer process UUID before ID128. And to protoBuf will need to add both String version of UUID and the long/long version of uuid 128.

elek:
/pending @runzhiwang can you please add answer to the comment?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/pipeline/Pipeline.java
runzhiwang: @xiaoyuyao  Sorry for missing the comment. @elek  Thanks for reminding.

> And to protoBuf will need to add both String version of UUID and the long/long version of uuid 128
This is for compatibility, in the future, I will remove String version of UUID.

> The improvement will be limited as fromProtoBuf to String will prefer process UUID before ID128
Do you mean we should prefer ID128 before process UUID.fromString ? If so, I think you are totally right, I will update it.  Besides I will also remove  UUID.fromString in the future.

runzhiwang: @xiaoyuyao Thanks for review. I have updated the patch, Could you help review it again ?

PR #1054, opened 2020-06-11
Hdds 3772. Add LOG to S3ErrorTable for easier problem locating. 

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-3772

### DIFF
--- a/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/exception/S3ErrorTable.java
+++ b/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/exception/S3ErrorTable.java
@@ -111,7 +111,7 @@ public final class S3ErrorTable {
     OS3Exception err =  new OS3Exception(e.getCode(), e.getErrorMessage(),
         e.getHttpCode());
     err.setResource(resource);
-    LOG.info(err.toXml());
+    LOG.error(err.toXml());
     return err;
   }
 }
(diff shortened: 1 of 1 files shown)


hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/exception/S3ErrorTable.java
```diff
@@ -29,6 +31,9 @@
  */
 public final class S3ErrorTable {
 
+  private final static Logger LOG = LoggerFactory.getLogger(
+      S3ErrorTable.class);
```
xiaoyuyao, line 35: This seems to be the main part matching scope of the title. Should we use separate tickets to track other changes?

ChenSammi: Sorry, Xiaoyu, I will upload a clean patch.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/exception/S3ErrorTable.java
```diff
@@ -106,6 +111,7 @@ public static OS3Exception newError(OS3Exception e, String resource) {
     OS3Exception err =  new OS3Exception(e.getCode(), e.getErrorMessage(),
         e.getHttpCode());
     err.setResource(resource);
+    LOG.info(err.toXml());
```
bharatviswa504, line 114: One comment:
This should be LOG.error

bharatviswa504:
Thank You @ChenSammi for the contribution and @xiaoyuyao for the review.

PR #1055, opened 2020-06-11
HDDS-3780. Replace the imagePullPolicy from always to IfNotPresent

maobaolong:

With the always imagePullPolicy, i cannot use the self-build local image. With the IfNotPresent policy, i can build an images to my local repository and use it.

https://jira.apache.org/jira/browse/HDDS-3780

## How was this patch tested?

Minor fix, no need to test.


maobaolong:
@elek Thank you for your review, i've address your suggestion, updated the two more yaml file in the definitions/ozone-csi folder, PTAL.

PR #1056, opened 2020-06-11
HDDS-3781. Update the apiVersion to the newer one

maobaolong:

For some apiVersion has been removed many years ago, so it is not forward-compatible, and i can use it in the newer version kubernetes cluster. For example, v1beta1, v1beta2, v1beta3 are removed in 2015.

https://jira.apache.org/jira/browse/HDDS-3781

## How was this patch tested?

Run the ozone demo in kubernetes cluster successfully.


maobaolong:
@elek Aha, so many years, Thank you for your review and merge this PR.

PR #1057, opened 2020-06-11
HDDS-3782. Remove podAntiAffinity from datanode-statefulset

maobaolong:

Now, i think the yaml files under examples folder are trial purpose, so we should not add the strict limitation to prevent people easy to taste the example, if who want reference this example and use it in production env, then they will do a lot of modification.

For this ticket, it suppose to remove the podAntiAffinity check to let on node kubernetes run the ozone cluster on kubernetes successfully.

https://jira.apache.org/jira/browse/HDDS-3782

## How was this patch tested?

No need, just loosing restrictions.


elek:
Thanks to open this issue @maobaolong 

This approach is already considered, and you can find the arguments in https://issues.apache.org/jira/browse/HDDS-1646 why do we have these persistence lines.

> Now, i think the yaml files under examples folder are trial purpose

Depends on how do you define *trial* purpose. I think the provided examples should work in a real, production-like environment. For example if somebody tries Ozone out in a small Kubernetes cluster, we should provide an example how can it be configured.

Removing the `podAntiAffinity` rules would result fake results of the initial measurements.

I understand that we have an other use case: To quickly try it out locally on Minikube, K3s or one node cluster. And I agree that we should support it. Today we have the `minikube` definitions for that specific use case.

 * I am sure that the real `ozone` example should have real `podAntiAffinity` as it provides an example to use real Ozone
 * `minikube` shouldn't have the `podAntiAffinity`
 * We can discuss `getting-started` and `ozone-dev`, I am open the adjust them.
 * But it seems to unclear what is the goal of different subdirectories. Would be great to do it with additional clarification: either we need to add more README to clarify the goals, or create different structure (or dir names?)

maobaolong:
@elek I agree with your suggestion,
>  I am sure that the real ozone example should have real podAntiAffinity as it provides an example to use real Ozone

Thank you to let me know well about the "ozone" and "ozone-dev" mean, now i agree to keep the podAntiAffinity for ozone dir

> minikube shouldn't have the podAntiAffinity
If fact, i don't know minikube at all, keep it.

> We can discuss getting-started and ozone-dev, I am open the adjust them.
I strongly recommend to remove  podAntiAffinity and let it easy to start successfully.

> But it seems to unclear what is the goal of different subdirectories. Would be great to do it with additional clarification: either we need to add more README to clarify the goals, or create different structure (or dir names?)

Yeah, I have created a ticket to track this work, https://issues.apache.org/jira/browse/HDDS-3790
maybe someone who are very familiar with these directories can write a better README

So, for this PR, i suppose to let ozone CSI run successfully and easily, so i will revert the unrelated and un unnecessary file.

elek:
Thanks the update @maobaolong 

My problem is that the examples are generated (see #1123). If you modify only the generated files, the changes will disappear.

If you propose to modify `ozone-dev` to run it on one single node by default, you should do the following:

1. copy [this file](https://github.com/flokkr/docker-ozone/blob/master/flekszible/ozone/definitions/onenode.yaml) to `hadoop-ozone/dist/src/main/k8s/definitions/ozone/definitions`

2. Add `ozone/onenode` transformations to the `Flekszible` file under `ozone-dev` and `getting-started`.

For example: 

```
diff --git hadoop-ozone/dist/src/main/k8s/examples/ozone-dev/Flekszible hadoop-ozone/dist/src/main/k8s/examples/ozone-dev/Flekszible
index 96e8c6254..350ea73c5 100644
--- hadoop-ozone/dist/src/main/k8s/examples/ozone-dev/Flekszible
+++ hadoop-ozone/dist/src/main/k8s/examples/ozone-dev/Flekszible
@@ -26,6 +26,7 @@ import:
     - type: ozone/profiler
     - type: ozone/emptydir
     - type: ozone/csi
+    - type: ozone/onenode
   - path: prometheus
   - path: jaeger
     transformations:
```

3. Run `regenerate-all.sh` from `#1123` (You might need to install [flekszible](https://github.com/elek/flekszible) which can be done with brew or putting the single binary to the path)

elek:
Update: when you import the file, the name of the statefulset should be updated in the trigger lines (datanode instead of ozone-datanode AFAIR)

maobaolong:
@elek Thank you for telling me how to change the example by `Flekszible`,  but, in fact, i still want to start 3 datanode by default, and remove the affinity limitation, so that I can start 3 datanodes on an only one node kubernetes cluster.

elek:
I got it, and I convinced by you to remove the affinity rules from `ozone-dev` and `getting-started` and we agreed to keep it in `ozone` (as it supposed to be a production-like example).

My problem is that the files under `example` are **generated** to simplify the management. For example, we can add any new labels to the master template and regenerate all the examples.(with #1123). 

I am convinced to remove the affinity rules *by default*, but if you do it **only** in the generated files, they will be overridden with the next generation. (To make all the files independent of the generation, I prefer to commit the end results, what you can see in the examples which makes the existence of the generation tool optional).

I tried to propose to modify BOTH the generated files AND the rules for the generation, to persist the change. It was not about following a different default. 

(Sorry if it was not clear or misunderstood your comment)

elek:
With other words: 

Try this: keep this commit, apply #1123, and run the new script. You will see that your changes are disappearing. I described the method to avoid this and save your changes ;-)

maobaolong:
@elek Thank you for your help, i catch you point is let me modify the source file of the Flekszible. I think I've done following your instruments. PTAL.

PR #1058, opened 2020-06-11
HDDS-3778. Block distribution in a pipeline among open containers is not uniform.

bshashikant:
Made allocate block call synchronised over pipeline object.

https://issues.apache.org/jira/browse/HDDS-3778

## How was this patch tested?
Added a unit test.


nandakumar131:
@bshashikant, thanks for the quick fix. There is a checkstyle issue, can you take care of that?

PR #1062, opened 2020-06-11
HDDS-3783. Upload coverage even if tests failed

adoroszlai:

PRs get code coverage feedback from Codecov thanks to changes implemented in HDDS-3726.  Codecov needs coverage data for each new revision on `master`.  However, _coverage_ check is currently [skipped](https://github.com/apache/hadoop-ozone/runs/759894317?check_suite_focus=true) if any tests fail.  This PR proposes to always execute coverage if basic compilation is OK.  It still waits for various test runs before execution.

https://issues.apache.org/jira/browse/HDDS-3783

## How was this patch tested?

Tested conditions on simplified workflow file with both:

* successful "compilation": [code](https://github.com/adoroszlai/hadoop-ozone/blob/3639f470b02bd7647b662f7d0435d2e776f92b87/.github/workflows/post-commit.yml), [run](https://github.com/adoroszlai/hadoop-ozone/runs/761754406?check_suite_focus=true)
* and failed "compilation": [code](https://github.com/adoroszlai/hadoop-ozone/blob/5856b413ca5a00c763e318428169588bbbb89d3e/.github/workflows/post-commit.yml), [run](https://github.com/adoroszlai/hadoop-ozone/runs/762117230?check_suite_focus=true)

Real CI ("unfortunately" all tests passed):
https://github.com/adoroszlai/hadoop-ozone/runs/761992786

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -224,6 +224,7 @@ jobs:
     name: coverage
     runs-on: ubuntu-18.04
     needs:
+       - build
        - it-ozone
        - it-hdds-om
        - it-client
@@ -231,6 +232,9 @@ jobs:
        - it-filesystem
        - it-freon
        - unit
+    if: |
+      !cancelled() &&
+      needs.build.result == 'success'
     steps:
        - uses: actions/checkout@v2
        - uses: ./.github/buildenv


elek:
I was thinking about the same, but I am afraid if we enable it, we got invalid coverage data. Let's say it-client is failing, at least the failing tests will be missing. For me it seems to be more reliable to upload only the full builds.

adoroszlai:
> I was thinking about the same, but I am afraid if we enable it, we got invalid coverage data. Let's say it-client is failing, at least the failing tests will be missing. For me it seems to be more reliable to upload only the full builds.

I think we get invalid coverage data with the current setup, too.  For example #1055 only changes some CSI yaml files, but coverage is said to decrease.  I think that's due to using wrong base coverage data.

elek:
>  I think that's due to using wrong base coverage data.

I am not sure. I thought `codecov` compares the branch with the master instead of `git merge-base`. Seems to be better to have less frequent but good coverage data on the master. With uploading failing test data I can't be sure that the coverage on the sonarcloud is valid or not.

It would be better to understand the problem of codecov and adjust it. Are you sure that this is the missing piece for the codecov?

elek:
BTW we can have different rules for codecov / sonarcloud AFAIK.

adoroszlai:
Codecov seems to agree with you. ;)  It will not compare against commits with failed CI (if it know that CI passed or failed).  From the [doc](https://docs.codecov.io/docs/comparing-commits#commits-with-failed-ci):

> If CI fails, one or more of the following assumptions can be made:
> 1. Not all tests were executed; therefore coverage is incomplete.
> 2. Exceptions may call new execution paths, resulting in different coverage metrics.
> 3. A failed test could produce different coverage than the same test ran successfully.

I think for Ozone:

1. is the worst problem, can happen if dependent modules are skipped by Maven due to earlier failure (eg. in `common`).
2. is true even for successful CI ([example](https://codecov.io/gh/apache/hadoop-ozone/compare/bd88eec9fec06a8f8adcbe4d257c5ccaca144f4c...96f07d5c27ba8f2ab9d77df25e10d4cb94489e5f/changes), where [only javadoc was updated](https://github.com/apache/hadoop-ozone/commit/96f07d5c27ba8f2ab9d77df25e10d4cb94489e5f)).
3. is something we should try to fix (eg. split test cases that perform a long series of steps, making assertions all the way).

PR #1066, opened 2020-06-12
HDDS-3770. Improve getPipelines performance

runzhiwang:

1. `getPipelines` is in the critical path of `allocateBlock`, and `discardPipeline` is the hot spot in `getPipelines`.
2. Change `List` to `Set` to make `add`  and `contains` faster.
3. Remove `stream().filter().collect()` and `parallelStream().anyMatch` in `discardPipeline` because `Set::contains` is fast enough.

https://issues.apache.org/jira/browse/HDDS-3770?filter=-2

## How was this patch tested?

Existed tests.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManager.java
@@ -180,7 +180,7 @@ public interface ContainerManager extends Closeable {
    * @return ContainerInfo for the matching container.
    */
   ContainerInfo getMatchingContainer(long size, String owner,
-      Pipeline pipeline, List<ContainerID> excludedContainerIDS);
+      Pipeline pipeline, Set<ContainerID> excludedContainerIDS);
 
   /**
    * Once after report processor handler completes, call this to notify
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManager.java
@@ -18,6 +18,7 @@ package org.apache.hadoop.hdds.scm.container;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -180,7 +181,7 @@ public interface ContainerManager extends Closeable {
    * @return ContainerInfo for the matching container.
    */
   ContainerInfo getMatchingContainer(long size, String owner,
-      Pipeline pipeline, Set<ContainerID> excludedContainerIDS);
+      Pipeline pipeline, Collection<ContainerID> excludedContainerIDS);
 
   /**
    * Once after report processor handler completes, call this to notify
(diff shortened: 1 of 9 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -416,13 +417,12 @@ public void updateDeleteTransactionId(Map<Long, Long> deleteTransactionMap)
    */
   public ContainerInfo getMatchingContainer(final long sizeRequired,
       String owner, Pipeline pipeline) {
-    return getMatchingContainer(sizeRequired, owner, pipeline, Collections
-        .emptyList());
+    return getMatchingContainer(sizeRequired, owner, pipeline, new HashSet<>());
```
xiaoyuyao, line 420: Collections.emptySet() to avoid unnecessary allocation

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManager.java
```diff
@@ -180,7 +180,7 @@ ContainerInfo getMatchingContainer(long size, String owner,
    * @return ContainerInfo for the matching container.
    */
   ContainerInfo getMatchingContainer(long size, String owner,
-      Pipeline pipeline, List<ContainerID> excludedContainerIDS);
+      Pipeline pipeline, Set<ContainerID> excludedContainerIDS);
```
xiaoyuyao, line 183: Can we change here to Collection<ContainerID> to make the interface neutral to implementation? The underlying can be List or Set.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManager.java
```diff
@@ -61,8 +61,8 @@ Pipeline createPipeline(ReplicationType type, ReplicationFactor factor,
       ReplicationFactor factor, Pipeline.PipelineState state);
 
   List<Pipeline> getPipelines(ReplicationType type, ReplicationFactor factor,
-      Pipeline.PipelineState state, Collection<DatanodeDetails> excludeDns,
-      Collection<PipelineID> excludePipelines);
```
xiaoyuyao, line 65: Can we keep the existing one parameter as Collection which allows passing Set here?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManager.java
```diff
@@ -81,8 +81,8 @@ Pipeline getPipeline(PipelineID pipelineID) throws PipelineNotFoundException {
   }
 
   List<Pipeline> getPipelines(ReplicationType type, ReplicationFactor factor,
-      PipelineState state, Collection<DatanodeDetails> excludeDns,
-      Collection<PipelineID> excludePipelines) {
```
xiaoyuyao, line 85: Same as above.

runzhiwang:
@xiaoyuyao Thanks for review. I have updated.

runzhiwang:
@dineshchitlangia Thanks for review.

elek:
Merging this as it has green build + comments from @xiaoyuyao addressed and @dineshchitlangia also approved it. (thanks the reviews).

Thanks the contribution @runzhiwang

PR #1067, opened 2020-06-12
HDDS-3787. Make CSI command configurable.

maobaolong:

It doesn't have to use goofys as the mountable tools to access Ozone. With this PR, we can config an other command to mount file system which could be a fuse program.

https://issues.apache.org/jira/browse/HDDS-3787

## How was this patch tested?

With default configuration, CI checks should be passed.


maobaolong:
@elek 😃 Thanks for your "Big +1", with this PR, we can continue to test whether another tools can bring performance improvement.

PR #1068, opened 2020-06-12
HDDS-3611. Ozone client should not consider closed container error as failure.

lokeshj1703:

ContainerNotOpen exception exception is thrown by datanode when client is writing to a non open container. Currently ozone client sees this as failure and would increment the retry count. If client reaches a configured retry count it fails the write. Map reduce jobs were seen failing due to this error with default retry count of 5.
Idea is to not consider errors due to closed container in retry count. This would make sure that ozone client writes do not fail due to closed container exceptions.

https://issues.apache.org/jira/browse/HDDS-3611

## How was this patch tested?

Existing Unit test


bshashikant:
Thanks @lokeshj1703 for working on this. The change looks ok. 
I think in this case, ozone client has already exhausted the max attempt count trying to write the data but eventually failed .

How about this case, where the client attempts to write and in turn run into ContainerNot open exception? In such cases, ozone client may keep on retrying for considerable amount of time which i think we should not do.

May be its ok to fail irrespective of exception after trying for maximum attempts.
Also, the test failure seems to be related. Can u plz check as well?

PR #1070, opened 2020-06-12
HDDS-3785. Update topology.aware.read parameter in ozone-topology compose config

sodonnel:

HDDS-1865 updated the parameter dfs.network.topology.aware.read.enable to ozone.network.topology.aware.read, but in the docker-compose config for ozone-topology, the old parameter is still used.

This Jira is to update it to the new value.

From OzoneConfigKeys.java, we can see the correct value for this parameter:

```
  public static final String OZONE_NETWORK_TOPOLOGY_AWARE_READ_KEY =
      "ozone.network.topology.aware.read";
  public static final boolean OZONE_NETWORK_TOPOLOGY_AWARE_READ_DEFAULT = false;
```

https://issues.apache.org/jira/browse/HDDS-3785?

## How was this patch tested?

existing tests and manual tests.


bharatviswa504:
Thank You @sodonnel for the fix and @vivekratnavel and @xiaoyuyao for the review.

PR #1074, opened 2020-06-14
HDDS-3793. Use Hadoop 2.7.3 for ozone-mr/hadoop27 acceptance tests

adoroszlai:

Change `ozone-mr/hadoop27` acceptance test to use Hadoop 2.7.3, the version `hadoop-ozone-filesystem-hadoop2` is built with (since HDDS-3715).

https://issues.apache.org/jira/browse/HDDS-3793

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/769907631

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-config
+++ b/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-config
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-CORE-SITE.xml_fs.AbstractFileSystem.o3fs.impl=org.apache.hadoop.fs.ozone.BasicOzFs
+CORE-SITE.xml_fs.AbstractFileSystem.o3fs.impl=org.apache.hadoop.fs.ozone.OzFs
 MAPRED-SITE.XML_mapreduce.application.classpath=/opt/hadoop/share/hadoop/mapreduce/*:/opt/hadoop/share/hadoop/mapreduce/lib/*:/opt/ozone/share/ozone/lib/hadoop-ozone-filesystem-hadoop2-@project.version@.jar
 
 no_proxy=om,scm,s3g,kdc,localhost,127.0.0.1
--- a/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-config
+++ b/hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-config
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-CORE-SITE.xml_fs.AbstractFileSystem.o3fs.impl=org.apache.hadoop.fs.ozone.OzFs
+CORE-SITE.xml_fs.AbstractFileSystem.o3fs.impl=org.apache.hadoop.fs.ozone.BasicOzFs
 MAPRED-SITE.XML_mapreduce.application.classpath=/opt/hadoop/share/hadoop/mapreduce/*:/opt/hadoop/share/hadoop/mapreduce/lib/*:/opt/ozone/share/ozone/lib/hadoop-ozone-filesystem-hadoop2-@project.version@.jar
 
 no_proxy=om,scm,s3g,kdc,localhost,127.0.0.1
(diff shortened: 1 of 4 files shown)


hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/test.sh
```diff
@@ -27,8 +27,8 @@ execute_robot_test scm createmrenv.robot
 
 
 #rm is the container name (resource manager) and not the rm command
-execute_command_in_container rm sudo -E apk add --update py-pip
-execute_command_in_container rm sudo -E pip install robotframework
+execute_command_in_container rm sudo -E yum install -y python3-pip
```
elek, line 30: This part can be removed. IMHO.

```
docker run flokkr/hadoop:2.7.3 robot --version
Flokkr launcher script 1.1-18-g58a3efc

===== Plugin is activated ENVTOCONF =====
======================================
*** Launching "robot --version"
Robot Framework 3.1.2 (Python 3.6.8 on linux)
```

hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-config
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-CORE-SITE.xml_fs.AbstractFileSystem.o3fs.impl=org.apache.hadoop.fs.ozone.OzFs
```
elek, line 17: What's wrong with `OzFS`? Based on my understanding `ozonefs-hadoop2` project includes an `OzFs` implementation which is the same as `BasicOzFs`.

Is there any motivation behind this change?

hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/test.sh
elek, line 30: This part can be removed. IMHO.

```
docker run flokkr/hadoop:2.7.3 robot --version
Flokkr launcher script 1.1-18-g58a3efc

===== Plugin is activated ENVTOCONF =====
======================================
*** Launching "robot --version"
Robot Framework 3.1.2 (Python 3.6.8 on linux)
```

hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-config
elek, line 17: What's wrong with `OzFS`? Based on my understanding `ozonefs-hadoop2` project includes an `OzFs` implementation which is the same as `BasicOzFs`.

Is there any motivation behind this change?

adoroszlai: > Is there any motivation behind this change?

Yes, with `OzFs` I ran into `Invalid path name Wrong FS` (HDDS-3482).

hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/test.sh
adoroszlai: Thanks, I'll remove it.

hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-config
elek: Got it. In that case it's my fault, the fix should be added to the hadoop2 specific `OzFs`. 

One important point of the new structure is that we don't need to use different classes (like `BasicOzFs`) we can use the same FS names all the time.

If it's not a big deal, can you please add the workaround instead of changing the name of the AbstractFileSystem implementation? 

```java
diff --git hadoop-ozone/ozonefs-hadoop2/src/main/java/org/apache/hadoop/fs/ozone/OzFs.java hadoop-ozone/ozonefs-hadoop2/src/main/java/org/apache/hadoop/fs/ozone/OzFs.java
index 914832e2c..b5012f95c 100644
--- hadoop-ozone/ozonefs-hadoop2/src/main/java/org/apache/hadoop/fs/ozone/OzFs.java
+++ hadoop-ozone/ozonefs-hadoop2/src/main/java/org/apache/hadoop/fs/ozone/OzFs.java
@@ -40,4 +40,9 @@ public OzFs(URI theUri, Configuration conf)
     super(theUri, new OzoneFileSystem(), conf,
         OzoneConsts.OZONE_URI_SCHEME, false);
   }
+
+  @Override
+  public int getUriDefaultPort() {
+    return -1;
+  }
 }
```

Tested and worked well.

adoroszlai: Thanks for the hint, workaround added.

adoroszlai:
Thanks @vivekratnavel for the review, and @elek for reviewing and committing this.

PR #1076, opened 2020-06-15
HDDS-3773. Add OMDBDefinition to define structure of om.db.

sadanand48:
In order to support the rocksdb tool which displays data for a given db file ([HDDS-3622](https://github.com/apache/hadoop-ozone/pull/945)) , the appropriate DBDefinition file must be defined. The change here is to extend the tool for om.db.

https://issues.apache.org/jira/browse/HDDS-3773

## How was this patch tested?
Manual test.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/codec/OMDBDefinition.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/codec/OMDBDefinition.java
@@ -1,0 +1,161 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


avijayanhwx:
Thank you for the patch @sadanand48. Any reason why we are not using the OMDBDefinition to actually create the OM DB, like in the SCM?

sadanand48:
> Thank you for the patch @sadanand48. Any reason why we are not using the OMDBDefinition to actually create the OM DB, like in the SCM?

@avijayanhwx I think the plan was to use DBStore interface instead of the MetadataDataStore and in order to do that DBDefinition must be defined [(HDDS-3172)] .(https://github.com/apache/hadoop-ozone/pull/700#issuecomment-615289480).
OM currently uses the MetadataStore Interface so there is no OMDBDefinition defined.

bshashikant:
> > Thank you for the patch @sadanand48. Any reason why we are not using the OMDBDefinition to actually create the OM DB, like in the SCM?
> 
> @avijayanhwx I think the plan was to use DBStore interface instead of the MetadataDataStore and in order to do that DBDefinition must be defined [([HDDS-3172](https://issues.apache.org/jira/browse/HDDS-3172))] .([#700 (comment)](https://github.com/apache/hadoop-ozone/pull/700#issuecomment-615289480)).
> OM currently uses the MetadataStore Interface so there is no OMDBDefinition defined.

I think we probably need to follow the same model as that in SCM. Though the patch looks good, other changes related to using DB interface rather than MetadataDataStore interface, can be done in separate jira after further analysis.

@avijayanhwx , what do you think?

elek:
> I think we probably need to follow the same model as that in SCM. Though the patch looks good, other changes related to using DB interface rather than MetadataDataStore interface, can be done in separate jira after further analysis.

+1 Sounds like a reasonable plan for me.

avijayanhwx:
@sadanand48  / @bshashikant I am OK with handling the next set of OM changes in a different JIRA.

elek:
Great. I think the proposal is to do something similar to #997 in the next patch.

Now I am merging this one. Thanks the patch @sadanand48 and the review @avijayanhwx and @bshashikant

PR #1078, opened 2020-06-15
HDDS-3794. Topology Aware read does not work correctly in XceiverClientGrpc

sodonnel:

In XceiverClientGrpc.java, the calls to read a block or chunks for a Datanode end up in the private method sendCommandWithRetry(). In this method it decides which datanode it should send the request to. To do that, it checks if there is a cached DN connection for the given block and if so it uses that. If there is no cached connection, it should take network topology into account or shuffle the nodes:

```
   List<DatanodeDetails> datanodeList = null;

    DatanodeBlockID blockID = null;
    if (request.getCmdType() == ContainerProtos.Type.ReadChunk) {
      blockID = request.getReadChunk().getBlockID();
    } else if (request.getCmdType() == ContainerProtos.Type.GetSmallFile) {
      blockID = request.getGetSmallFile().getBlock().getBlockID();
    }

    if (blockID != null) {
      // Check if the DN to which the GetBlock command was sent has been cached.
      DatanodeDetails cachedDN = getBlockDNcache.get(blockID);
      if (cachedDN != null) {
        datanodeList = pipeline.getNodes();
        int getBlockDNCacheIndex = datanodeList.indexOf(cachedDN);
        if (getBlockDNCacheIndex > 0) {
          // Pull the Cached DN to the top of the DN list
          Collections.swap(datanodeList, 0, getBlockDNCacheIndex);
        }
      } else if (topologyAwareRead) {
        datanodeList = pipeline.getNodesInOrder();
      }
    }
    if (datanodeList == null) {
      datanodeList = pipeline.getNodes();
      // Shuffle datanode list so that clients do not read in the same order
      // every time.
      Collections.shuffle(datanodeList);
    }
    <call to DN after here>
```

The normal flow for the client is to first make a getBlock() call to the DN and then a readChunk() call.

Due to the logic at the top of the block above, blockID is always going to be null for the getBlock() call, then it never checks the topologyAwareRead section and shuffles the node.

Then for readChunk, it will find the blockID, find a cached DN, which was the result of the shuffle, and then it reuses that DN.

Therefore the topologyAwareRead does not work as expected.

https://issues.apache.org/jira/browse/HDDS-3794

## How was this patch tested?

New unit tests added to reproduce the issue and then prove it is fixed.

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
@@ -298,7 +298,9 @@ public class XceiverClientGrpc extends XceiverClientSpi {
     List<DatanodeDetails> datanodeList = null;
 
     DatanodeBlockID blockID = null;
-    if (request.getCmdType() == ContainerProtos.Type.ReadChunk) {
+    if (request.getCmdType() == ContainerProtos.Type.GetBlock) {
+      blockID = request.getGetBlock().getBlockID();
+    } else if  (request.getCmdType() == ContainerProtos.Type.ReadChunk) {
       blockID = request.getReadChunk().getBlockID();
     } else if (request.getCmdType() == ContainerProtos.Type.GetSmallFile) {
       blockID = request.getGetSmallFile().getBlock().getBlockID();
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
@@ -314,15 +314,17 @@ public class XceiverClientGrpc extends XceiverClientSpi {
           // Pull the Cached DN to the top of the DN list
           Collections.swap(datanodeList, 0, getBlockDNCacheIndex);
         }
-      } else if (topologyAwareRead) {
-        datanodeList = pipeline.getNodesInOrder();
       }
     }
     if (datanodeList == null) {
-      datanodeList = pipeline.getNodes();
-      // Shuffle datanode list so that clients do not read in the same order
-      // every time.
-      Collections.shuffle(datanodeList);
+      if (topologyAwareRead) {
+        datanodeList = pipeline.getNodesInOrder();
+      } else {
+        datanodeList = pipeline.getNodes();
+        // Shuffle datanode list so that clients do not read in the same order
+        // every time.
+        Collections.shuffle(datanodeList);
+      }
     }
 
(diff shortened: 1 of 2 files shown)


hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
```diff
@@ -314,15 +314,17 @@ private XceiverClientReply sendCommandWithRetry(
           // Pull the Cached DN to the top of the DN list
           Collections.swap(datanodeList, 0, getBlockDNCacheIndex);
         }
-      } else if (topologyAwareRead) {
```
xiaoyuyao, line 317: Good catch @sodonnel . The fix LGTM overall. I just have one question:

Have you consider populating blockID properly for the first getBlock() call with additional else if on the top?

else if (request.getCmdType() == ContainerProtos.Type.GetBlock) {
      blockID = request.getGetBlock().getBlockID();
    }

sodonnel: Yea, I did wonder why getBlock() is not part of that IF block. I guess it makes sense to add it in just in case, as it will do no harm.

I am not really fully up-to-speed on the read path from the Ozone client - would it be normal for several getblock calls to be made against the same pipeline (XceiverClientGrpc instance)? Ie do we expect getBlock to benefit from the cached connection sometimes?

bshashikant: > Yea, I did wonder why getBlock() is not part of that IF block. I guess it makes sense to add it in just in case, as it will do no harm.
> 
> I am not really fully up-to-speed on the read path from the Ozone client - would it be normal for several getblock calls to be made against the same pipeline (XceiverClientGrpc instance)? Ie do we expect getBlock to benefit from the cached connection sometimes?

Yes, it is quite possible to use the same XceiverClientGrpc instanvce to read for multiple blocks for multiple keys.

sodonnel: Thanks for the feedback. I have made this change, so pending a clean CI run please give it a check.

sodonnel: @xiaoyuyao @bshashikant Are you happy with the revision and for me to commit this one? Thanks.

xiaoyuyao: LGTM, +1.

xiaoyuyao:
Thanks @sodonnel  for the contribution and all for the reviews. I've merged the PR.

PR #1079, opened 2020-06-16
HDDS-3802. Incorrect data returned by reading a FILE_PER_CHUNK block.

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-3802

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/impl/FilePerChunkStrategy.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/impl/FilePerChunkStrategy.java
@@ -34,11 +34,14 @@ import org.apache.hadoop.ozone.container.keyvalue.KeyValueContainerData;
 import org.apache.hadoop.ozone.container.common.volume.HddsVolume;
 import org.apache.hadoop.ozone.container.common.volume.VolumeIOStats;
 import org.apache.hadoop.ozone.container.keyvalue.helpers.ChunkUtils;
+import org.apache.hadoop.ozone.container.keyvalue.interfaces.BlockManager;
 import org.apache.hadoop.ozone.container.keyvalue.interfaces.ChunkManager;
 import org.apache.hadoop.ozone.container.common.interfaces.Container;
 
 import static org.apache.hadoop.hdds.protocol.datanode.proto.ContainerProtos.Result.IO_EXCEPTION;
 import static org.apache.hadoop.hdds.protocol.datanode.proto.ContainerProtos.Result.UNABLE_TO_FIND_CHUNK;
+
+import org.eclipse.jetty.util.StringUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -62,14 +65,37 @@ public class FilePerChunkStrategy implements ChunkManager {
       LoggerFactory.getLogger(FilePerChunkStrategy.class);
 
   private final boolean doSyncWrite;
+  private final BlockManager blockManager;
 
-  public FilePerChunkStrategy(boolean sync) {
+  public FilePerChunkStrategy(boolean sync, BlockManager manager) {
     doSyncWrite = sync;
+    blockManager = manager;
   }
 
   private static void checkLayoutVersion(Container container) {
     Preconditions.checkArgument(
         container.getContainerData().getLayOutVersion() == FILE_PER_CHUNK);
+  }
+
+  private static int extractChunkIndex(String chunkName) {
+    int index = -1;
+    if (StringUtil.isBlank(chunkName)) {
+      LOG.error("Failed to parse empty chuck name to get index");
+      return index;
+    }
+
+    String[] subNames = chunkName.split("_");
+    if (subNames.length != 3) {
+      LOG.error("Failed to parse chuck name to get index " + chunkName);
+      return index;
+    }
+
+    try {
+      index = Integer.parseInt(subNames[2]);
+    } catch (Exception e) {
+      LOG.error("Failed to parse chuck name to get index " + chunkName);
+    }
+    return index;
   }
 
   /**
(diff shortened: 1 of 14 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/impl/FilePerChunkStrategy.java
```diff
@@ -218,17 +244,36 @@ public ChunkBuffer readChunk(Container container, BlockID blockID,
...
+    }
+    // Chunk index start from 1
+    Preconditions.checkState(index > 0);
+
+    long chunkFileOffset;
+    try {
+      BlockData blockData = blockManager.getBlock(kvContainer, blockID);
+      List<ContainerProtos.ChunkInfo> chunks = blockData.getChunks();
+      Preconditions.checkState(index <= chunks.size());
+      chunkFileOffset = chunks.get(index - 1).getOffset();
```
adoroszlai, line 261: Instead of parsing the chunk name for an index, we could find the chunk with the exact same name.  I think it's less fragile, although might be slower if the block has too many chunks.  What's your opinion?

https://github.com/adoroszlai/hadoop-ozone/commit/398ea1d5991c44baa59c120dea8fa3c80727e9e2 (note: CI is still in progress, may need tweak based on the result)

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/impl/FilePerChunkStrategy.java
```diff
@@ -218,17 +244,36 @@ public ChunkBuffer readChunk(Container container, BlockID blockID,
...
+      throw new StorageContainerException(
+          "Chunk file name can't be parsed " + possibleFiles.toString(),
+          UNABLE_TO_FIND_CHUNK);
+    }
+    // Chunk index start from 1
+    Preconditions.checkState(index > 0);
+
+    long chunkFileOffset;
+    try {
+      BlockData blockData = blockManager.getBlock(kvContainer, blockID);
```
adoroszlai, line 258: I think we can skip trying to determine chunk file offset if `info.getOffset() == 0`.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/impl/FilePerChunkStrategy.java
ChenSammi: A good point. I think the performance overhead can be ignored.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestKeyInputStream.java
```diff
@@ -64,25 +71,38 @@
...
    * Ozone is made active by setting OZONE_ENABLED = true
    *
    * @throws IOException
    */
-  @BeforeClass
-  public static void init() throws Exception {
-    chunkSize = 100;
+  @Before
+  public void init() throws Exception {
+    chunkSize = 1024 * 1024 * 4;
```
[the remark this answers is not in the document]
ChenSammi: I will try to reduce the chunkSize. One limitation here is chunkSize cann't be less than the mininum bytes-per-checksume value which is 256KB, otherwise the TestKeyInputStream will always succeed event without this patch.

ChenSammi:
Thanks @adoroszlai  for review the code.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestKeyInputStream.java
```diff
@@ -331,4 +351,45 @@ public void testCopyLarge() throws Exception {
...
+      while (totalRead < dataLength) {
+        int numBytesRead = keyInputStream.read(data);
+        if (numBytesRead == -1 || numBytesRead == 0) {
+          break;
+        }
+        byte[] tmp1 =
+            Arrays.copyOfRange(originData, totalRead, totalRead + numBytesRead);
+        byte[] tmp2 =
+            Arrays.copyOfRange(data, 0, numBytesRead);
+        Arrays.equals(tmp1, tmp2);
```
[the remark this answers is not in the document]
ChenSammi: Thanks.

ChenSammi:
@adoroszlai thanks for the review.  Is it a +1?  Should I wait for @bshashikant 's review and feedback before commit?

adoroszlai:
> @adoroszlai thanks for the review. Is it a +1?

Yes.

> Should I wait for @bshashikant 's review and feedback before commit?

Let's wait a bit more.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/impl/FilePerChunkStrategy.java
```diff
@@ -218,17 +222,40 @@ public ChunkBuffer readChunk(Container container, BlockID blockID,
...
       try {
-        // use offset only if file written by old datanode
-        long offset;
-        if (file.exists() && file.length() == info.getOffset() + len) {
-          offset = info.getOffset();
-        } else {
-          offset = 0;
+        if (file.exists()) {
+          long offset = info.getOffset() - chunkFileOffset;
+          Preconditions.checkState(offset >= 0);
```
bshashikant, line 255: I think, the precondition here won't hold good if we are using older client (prior FILE_PER_BLOCK strategy where offset for each chunk would be 0). @ChenSammi do you have a requirement for this?

bshashikant:
Thanks @ChenSammi for the contribution and @adoroszlai for the review. I have committed this.

ChenSammi:
Thanks @adoroszlai  and @bshashikant

PR #1081, opened 2020-06-16
HDDS-3809. Make number of open containers on a datanode a function of no of volumes reported by it.

bshashikant:
Added logic to take into account max no of healthy disks reported among the sets
of datanodes of a pipeline to get the total no of open containers per pipeline.

https://issues.apache.org/jira/browse/HDDS-3809

## How was this patch tested?

Added unit tests.


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -608,6 +608,16 @@ public void close() throws IOException {
     pipelineFactory.shutdown();
   }
 
+  /**
+   * returns max count of healthy volumes from the set of
+   * datanodes constituting the pipeline
```
dineshchitlangia, line 613:
```suggestion
   * Returns count of healthy volumes from the set of
   * datanodes constituting the pipeline.
```

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/MockNodeManager.java
```diff
@@ -605,6 +616,12 @@ public void setNetworkTopology(NetworkTopology topology) {
       this.currentState = currentState;
     }
 
+    NodeData(long capacity, long used, long currentState, long healthyVolumes) {
```
adoroszlai, line 619: Parameter `healthyVolumes` is unused.

bshashikant: Addressed in the latest commit.

adoroszlai:
Thanks @bshashikant for the contribution, @elek and @dineshchitlangia for the review.  Merged to `master` after latest green run.

PR #1082, opened 2020-06-17
HDDS-3685. Remove replay logic from actual request logic.

bharatviswa504:
Remove the replay logic from actual requests.

https://issues.apache.org/jira/browse/HDDS-3685

## How was this patch tested?

Existing tests, will add a few more tests to cover this scenario.


bharatviswa504:
Rebased patch with the latest master.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/s3/multipart/S3MultipartUploadCommitPartResponse.java
```diff
@@ -69,58 +70,32 @@
...
       // multipart upload. So, delete this part information.
+
       RepeatedOmKeyInfo repeatedOmKeyInfo =
           omMetadataManager.getDeletedTable().get(openKey);
 
-      repeatedOmKeyInfo = OmUtils.prepareKeyForDelete(
-          openPartKeyInfoToBeDeleted, repeatedOmKeyInfo,
-          openPartKeyInfoToBeDeleted.getUpdateID(), isRatisEnabled);
+      OmUtils.prepareKeyForDelete(openPartKeyInfoToBeDeleted,
+          repeatedOmKeyInfo, omMultipartKeyInfo.getUpdateID(), isRatisEnabled);
```
hanishakoneru, line 98: Should this return value be set to repeatedOmKeyInfo?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/exceptions/OMException.java
```diff
@@ -223,6 +223,6 @@ public String toString() {
 
     INVALID_VOLUME_NAME,
 
-    REPLAY // When ratis logs are replayed.
+    REPLAY
```
hanishakoneru, line 226: Any reason for keeping this around?

bharatviswa504: Initially, the reason was  because proto.lock file, if any breaking changes, it will fail to compile. I have removed this, as anyway HA is part of this release, and we can remove this field.

cc @avijayanhwx

bharatviswa504:
OMKeysDeleteRequest will be fixed by HDDS-3930.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/exceptions/OMException.java
avijayanhwx: The general guideline is breaking API changes are allowed pre-GA, but not after that.

hanishakoneru:
+1 pending CI.

bharatviswa504:

PR #1084, opened 2020-06-17
HDDS-3817. Create a column family through debug ldb tool

maobaolong:

We need to create a column family with given name for test purpose.

https://issues.apache.org/jira/browse/HDDS-3817

## How was this patch tested?

An example as follow. We can verify the xxx table is created.

```bash
ozone debug ldb --db=/Users/mbl/ozone/om/metadata-dirs/om.db ls
ozone debug ldb --db=/Users/mbl/ozone/om/metadata-dirs/om.db create -n xxx
ozone debug ldb --db=/Users/mbl/ozone/om/metadata-dirs/om.db ls
```

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/CreateTable.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/CreateTable.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


maobaolong:
@sadanand48 Please take a look at this PR. Thanks.

bharatviswa504:
Hi @maobaolong 
Thank you for the patch.
In what kind of cases, do we need this kind of approach of adding a column family to existing DB. Because this is up to OM/SCM who manages the DB, they need to create those tables according to the requirements of code. If we expose this API, is there a chance users can use to add a columnFamily, and while OM opens DB, it does not know about it, as there is no code in OmMetadatManager to manage such column family?

PR #1085, opened 2020-06-17
HDDS-3815. Avoid buffer copy in ContainerCommandRequestProto.

mukul1987:

Avoid Buffer Copy in ContainerCommandRequestProto.

https://issues.apache.org/jira/browse/HDDS-3815

## How was this patch tested?
Ran unit tests.


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/ContainerCommandRequestMessage.java
```diff
@@ -61,9 +61,11 @@ public static ContainerCommandRequestMessage toMessage(
   public static ContainerCommandRequestProto toProto(
       ByteString bytes, RaftGroupId groupId)
       throws InvalidProtocolBufferException {
-    final int i = 4 + bytes.asReadOnlyByteBuffer().getInt();
+    final int i = 4 + bytes.substring(0, Integer.BYTES)
```
adoroszlai, line 64:
```suggestion
    final int i = Integer.BYTES + bytes.substring(0, Integer.BYTES)
```

PR #1086, opened 2020-06-17
HDDS-3818. Schedule daily 2 builds from master branch build

elek:

@mukul1987  suggested to schedule cron based build to have more frequent data points to identify flaky tests.

We can start with two additional daily build which can be independent of the commit frequency (today we build master only after the commits).

https://issues.apache.org/jira/browse/HDDS-3818

## How was this patch tested?

Pushed a similar change to my fork to check if the syntax is valid.

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -14,7 +14,9 @@
 # limitations under the License.
 name: build-branch
 on:
-  - push
+  push:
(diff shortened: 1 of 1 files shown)


adoroszlai:
Thanks @mukul1987 for the suggestion, @elek for implementing it, @vivekratnavel and @smengcl for the review.

PR #1087, opened 2020-06-17
HDDS-3819. OzoneManager#listVolumeByUser ignores userName parameter when ACL is enabled

smengcl:

Don't ignore `userName` input when ACL is enabled for `OzoneManager#listVolumeByUser`.

https://issues.apache.org/jira/browse/HDDS-3819

## How was this patch tested?

Fixed existing test `TestOzoneManagerListVolumes`.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
@@ -146,7 +146,6 @@ public class TestOzoneManagerListVolumes {
   private void checkUser(MiniOzoneCluster cluster, UserGroupInformation user,
       List<String> expectVol, boolean expectListAllSuccess) throws IOException {
 
-    UserGroupInformation.setLoginUser(user);
     OzoneClient client = cluster.getClient();
     ObjectStore objectStore = client.getObjectStore();
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
@@ -189,6 +189,59 @@ public class TestOzoneManagerListVolumes {
     }
   }
 
+  /**
+   * Check if listVolume of other users than the login user works as expected.
+   * ozone.om.volume.listall.allowed = true
+   * Everyone should be able to list other users' volumes with this config.
+   */
+  @Test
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1583,11 +1583,11 @@ private void checkAcls(ResourceType resType, StoreType store,
...
    * @return true if permission granted, false if permission denied.
    */
-  private boolean hasAcls(ResourceType resType, StoreType store,
-      ACLType acl, String vol, String bucket, String key) {
+  private boolean hasAcls(String userName, ResourceType resType,
+      StoreType store, ACLType acl, String vol, String bucket, String key) {
     try {
       return checkAcls(resType, store, acl, vol, bucket, key,
-          ProtobufRpcEngine.Server.getRemoteUser(),
+          UserGroupInformation.createRemoteUser(userName),
```
smengcl, line 1590: Question for @xiaoyuyao here: `UserGroupInformation.createRemoteUser()` create a new UGI with `AuthMethod.SIMPLE`. Does this matter?

IMO in this case only username matters.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerListVolumes.java
```diff
@@ -146,7 +146,6 @@ private void setVolumeAcl(ObjectStore objectStore, String volumeName,
   private void checkUser(MiniOzoneCluster cluster, UserGroupInformation user,
       List<String> expectVol, boolean expectListAllSuccess) throws IOException {
 
-    UserGroupInformation.setLoginUser(user);
```
smengcl, line 149: Just a note:

Removing this line enables the existing tests to detect the bug solved by this jira.
You see previously I set the login user to the user I'm calling listVolume with, which unintentionally hides the bug.

smengcl: I've also added new test cases.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
xiaoyuyao: That should be fine in this case given the user has passed the RPC layer Kerberos auth.

smengcl: Sounds good. Thanks!

smengcl:
Thanks @xiaoyuyao for the review and +1. Merging in a minute.

PR #1096, opened 2020-06-19
HDDS-3833. Use Pipeline choose policy to choose pipeline from exist pipeline list

maobaolong:

With this policy driven mode, we can develop various pipeline choosing policy to satisfy complex production environment. 

Now the default policy is Random policy.

https://issues.apache.org/jira/browse/HDDS-3833

## How was this patch tested?

CI checks.


maobaolong:
@nandakumar131 Would you like to take a look at this PR? Thanks.

maobaolong:
Hi @elek Would like to take a look at this PR?

maobaolong:
@xiaoyuyao Please take a look at this PR, thanks.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
```diff
@@ -287,6 +287,9 @@
   public static final String OZONE_SCM_PIPELINE_OWNER_CONTAINER_COUNT =
       "ozone.scm.pipeline.owner.container.count";
   public static final int OZONE_SCM_PIPELINE_OWNER_CONTAINER_COUNT_DEFAULT = 3;
+  // Pipeline choose policy:
+  public static final String OZONE_SCM_PIPELINE_CHOOSE_IMPL_KEY =
+      "ozone.scm.pipeline.choose.impl";
```
avijayanhwx, line 292: nit. Suggest rename ozone.scm.pipeline.choose.impl --> ozone.scm.pipeline.choose.policy.impl

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/PipelineChoosePolicy.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface PipelineChoosePolicy {
+
+  /**
+   * Given an initial list of pipelines, return one of the pipelines.
+   *
+   * @param pipelineList list of pipelines.
+   * @return one of the pipelines.
+   */
+  Pipeline choosePipeline(List<Pipeline> pipelineList);
```
avijayanhwx, line 35: @maobaolong For my knowledge, can you list a few pipeline selection policies that you envision? I am wondering if we are OK with just a List\<Pipeline\> as an argument.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/BlockManagerImpl.java
```diff
@@ -222,9 +224,7 @@ public AllocatedBlock allocateBlock(final long size, ReplicationType type,
       }
 
       if (null == pipeline) {
-        // TODO: #CLUTIL Make the selection policy driven.
-        pipeline = availablePipelines
-            .get((int) (Math.random() * availablePipelines.size()));
+        pipeline = pipelineChoosePolicy.choosePipeline(availablePipelines);
```
avijayanhwx, line 227: We could get a runtime exception from a plugged in policy. Can we have to fallback to default (Random) on error after logging?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/choose/algorithms/PipelineChoosePolicyFactory.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public static PipelineChoosePolicy getPolicy(
+      ConfigurationSource conf) throws SCMException {
+    final Class<? extends PipelineChoosePolicy> policyClass = conf
+        .getClass(ScmConfigKeys.OZONE_SCM_PIPELINE_CHOOSE_IMPL_KEY,
+            OZONE_SCM_PIPELINE_CHOOSE_IMPL_DEFAULT,
+            PipelineChoosePolicy.class);
+    Constructor<? extends PipelineChoosePolicy> constructor;
+    try {
+      constructor = policyClass.getDeclaredConstructor();
```
avijayanhwx, line 53: We are relying on default constructor here. And we pass in only List\<Pipeline\> through the API. We may have policies that need more information like the topology, client address etc as well. Can we make sure we support them without change of interface later?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
maobaolong: Done.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/BlockManagerImpl.java
maobaolong: Done.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/PipelineChoosePolicy.java
maobaolong: @avijayanhwx Thank you for your suggestion, I add a arguments map, for extension purpose, i use Object type as value type.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/choose/algorithms/PipelineChoosePolicyFactory.java
maobaolong: Now it cannot get the client address while allocate block. After this PR, we can create a new ticket to add client parameter to the allocateblock method, then, I can write a localFirstPolicy. Also, I plan to let datanode report its utilized to scm, so i can implement a LowLoadFirstPolicy, we can also collect the pipeline vote count, batch size, use these information, we can get write many kind of choose pipeline policy.

maobaolong:
@avijayanhwx Thanks for your pervious review and suggestion, I push a new commit and addressed you comments, PTAL.

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -814,6 +814,20 @@
       value.
     </description>
   </property>
+  <property>
```
elek, line 817: I would prefer to use the new, Java based configuration API:

https://cwiki.apache.org/confluence/display/HADOOP/Java-based+configuration+API

hadoop-hdds/interface-server/src/main/resources/proto.lock
```diff
@@ -1209,6 +1209,10 @@
               {
                 "name": "INTERNAL_ERROR",
                 "integer": 29
+              },
```
elek, line 1212: Proto.lock file change should be excluded. the `proto.lock` should represent and old state: the last released state to use it as a base for comparison.

Rebase to the latest master and you won't see the changed proto lock files (which makes it easy to add lock files accidentally)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/BlockManagerImpl.java
```diff
@@ -222,9 +224,11 @@ public AllocatedBlock allocateBlock(final long size, ReplicationType type,
       }
 
       if (null == pipeline) {
-        // TODO: #CLUTIL Make the selection policy driven.
-        pipeline = availablePipelines
-            .get((int) (Math.random() * availablePipelines.size()));
+        Map<String, Object> params = new HashMap<>();
+        params.put(
+            PipelineChoosePolicy.PIPELINE_CHOOSE_POLICY_PARAM_SIZE, size);
+        pipeline = pipelineChoosePolicy.choosePipeline(
```
elek, line 230: It seems to be a standard parameter which is added to all the implementation. Wouldn't be better to add it as a type-safe field? Or we can use a `PipelineRequestInformation` class instead of the `Map<String,Object>` if you would like to make it easier to add more information latest.

avijayanhwx:
@maobaolong There are still some unaddressed comments here.
- Using Java based config API
- Adding unit test(s).
PTAL

maobaolong:
@avijayanhwx @elek  I've addressed all your comments and add some ut, PTAL.

maobaolong:
@elek Thanks you very much for merge this PR, base this framework, we can create various pipeline choose policies.

PR #1099, opened 2020-06-19
HDDS-3704. Update all the documentation to use ozonefs-hadoop2/3 instead of legacy/current

elek:

Using legacy jar is legacy. We need to updated all the docs.

https://issues.apache.org/jira/browse/HDDS-3704

## How was this patch tested?

I followed the instructions from `docs/content/recipe/SparkOzoneFSK8S.md` to test if it still works.

Also tested with 

```
cd hadoop-hdds/docs/content
grep -r lib-legacy
grep -r lib-current
```

### DIFF
--- a/hadoop-hdds/docs/content/interface/OzoneFS.md
+++ b/hadoop-hdds/docs/content/interface/OzoneFS.md
@@ -25,7 +25,7 @@ The Hadoop compatible file system interface allows storage backends like Ozone
 to be easily integrated into Hadoop eco-system.  Ozone file system is an
 Hadoop compatible file system.
 
-## Setting up the Ozone file system
+## Setting up the Ozone file system (o3fs)
 
 To create an ozone file system, we have to choose a bucket where the file system would live. This bucket will be used as the backend store for OzoneFileSystem. All the files and directories will be stored as keys in this bucket.
 
@@ -42,10 +42,6 @@ Please add the following entry to the core-site.xml.
 
 {{< highlight xml >}}
 <property>
-  <name>fs.o3fs.impl</name>
-  <value>org.apache.hadoop.fs.ozone.OzoneFileSystem</value>
-</property>
-<property>
   <name>fs.AbstractFileSystem.o3fs.impl</name>
   <value>org.apache.hadoop.fs.ozone.OzFs</value>
 </property>
@@ -57,11 +53,13 @@ Please add the following entry to the core-site.xml.
 
 This will make this bucket to be the default file system for HDFS dfs commands and register the o3fs file system type.
 
-You also need to add the ozone-filesystem.jar file to the classpath:
+You also need to add the ozone-filesystem-hadoop3.jar file to the classpath:
 
 {{< highlight bash >}}
-export HADOOP_CLASSPATH=/opt/ozone/share/ozonefs/lib/hadoop-ozone-filesystem-lib-current*.jar:$HADOOP_CLASSPATH
+export HADOOP_CLASSPATH=/opt/ozone/share/ozonefs/lib/hadoop-ozone-filesystem-hadoop3-*.jar:$HADOOP_CLASSPATH
 {{< /highlight >}}
+
+(Note: with Hadoop 2.x, use the `hadoop-ozone-filesystem-hadoop3-*.jar`)
 
 Once the default Filesystem has been setup, users can run commands like ls, put, mkdir, etc.
 For example,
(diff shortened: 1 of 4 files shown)


hadoop-hdds/docs/content/interface/OzoneFS.md
```diff
@@ -57,12 +53,14 @@ Please add the following entry to the core-site.xml.
...
 
-You also need to add the ozone-filesystem.jar file to the classpath:
+You also need to add the ozone-filesystem-hadoop3.jar file to the classpath:
 
 {{< highlight bash >}}
-export HADOOP_CLASSPATH=/opt/ozone/share/ozonefs/lib/hadoop-ozone-filesystem-lib-current*.jar:$HADOOP_CLASSPATH
+export HADOOP_CLASSPATH=/opt/ozone/share/ozonefs/lib/hadoop-ozone-filesystem-hadoop3-*.jar:$HADOOP_CLASSPATH
 {{< /highlight >}}
 
+(Note: with Hadoop 2.x, use the `hadoop-ozone-filesystem-hadoop3-*.jar`)
```
xiaoyuyao, line 62: Do you mean hadoop-ozone-filesystem-hadoop2-*.jar?

hadoop-hdds/docs/content/interface/OzoneFS.zh.md
```diff
@@ -58,7 +54,7 @@ ozone sh bucket create /volume/bucket
 你还需要将 ozone-filesystem.jar 文件加入 classpath：
 
 {{< highlight bash >}}
-export HADOOP_CLASSPATH=/opt/ozone/share/ozonefs/lib/hadoop-ozone-filesystem-lib-current*.jar:$HADOOP_CLASSPATH
+export HADOOP_CLASSPATH=/opt/ozone/share/ozonefs/lib/hadoop-ozone-filesystem-hadoop3-*.jar:$HADOOP_CLASSPATH
 {{< /highlight >}}
```
xiaoyuyao, line 59: Please insert between 59-60 the following in Chinese for the new sentence added 
“(Note: with Hadoop 2.x, use the `hadoop-ozone-filesystem-hadoop3-*.jar`)”

(注意：当使用Hadoop 2.x时，应该在classpath上添加`hadoop-ozone-filesystem-hadoop2-*.jar`）

hadoop-hdds/docs/content/interface/OzoneFS.md
elek: :blush: 

It seems that I lost in the forests of jar files. Thanks, fixing it.

hadoop-hdds/docs/content/interface/OzoneFS.zh.md
elek: > Please insert between 59-60 the following in Chinese for the new sentence added

Thanks A LOT. 

BTW, what is at �? I installed Chinese fonts, but I couldn't see it. I am not sure if I it's because my copy-paste or Github removed from something from there...

elek:
Thanks the review @xiaoyuyao. I pushed the commit, but please check it as I am not sure about one character which is visible as  � for me:

image

BTW, it's an interesting question how can we maintain the doc with two languages. Do you know any good tool which can manage the translations (even if external). Something which can flag the modified parts / new paragraph to request and update on the documentation. (cc: @iamabug ).

hadoop-hdds/docs/content/interface/OzoneFS.zh.md
xiaoyuyao: I think that's either typos or github encoding issue when I switch from Chinese to English on the comment. You can ignore and remove them.

elek: done

xiaoyuyao:
bq. Do you know any good tool which can manage the translations (even if external). 

Not as far as I know. How about we generate a md5 of each english document and its corresponding documents in other languages. Then add a unit test that fail if check if the md5 of english document matches. If not, assert that md5 of doc for other languages must be updated.

iamabug:
How about we align two translations at some version, and afterwards set a trigger on commits that relate to English docs, the trigger creates tickets which point to the said commits under a long-term ticket. Then we can check and resolve these tickets. Just an idea, not sure if possible.

elek:
An approach from Flink:

https://github.com/rmetzger/flink-community-tools/blob/master/src/main/java/de/robertmetzger/flink/community/flinkbot/checks/DocumentationCheck.java

It can check if the zh translation is included OR the issue is created.

PR #1101, opened 2020-06-20
HDDS-3821. Disable Ozone SPNEGO should not fall back to hadoop.http.a…

xiaoyuyao:
…uthentication configuration.

When ozone http kerberos is disabled, ensure the authentication filter does not fallback to configuration from hadoop.http.authentication.*

https://issues.apache.org/jira/browse/HDDS-3821

## How was this patch tested?

Acceptance test and Manual test when ozone http kerberos is disabled but  hadoop.http.authentication.* is enabled.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/resources/log4j.properties
+++ b/hadoop-ozone/integration-test/src/test/resources/log4j.properties
@@ -19,3 +19,6 @@ log4j.appender.stdout.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} (%F:%
 
 log4j.logger.org.apache.hadoop.security.ShellBasedUnixGroupsMapping=ERROR
 log4j.logger.org.apache.hadoop.util.NativeCodeLoader=ERROR
+
+log4j.logger.org.apache.ratis.grpc.server.GrpcLogAppender=DEBUG
+log4j.logger.org.apache.ratis.server.impl.RaftServerImpl=DEBUG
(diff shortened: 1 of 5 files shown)


xiaoyuyao:
org.apache.hadoop.ozone.client.rpc.Test2WayCommitInRatis seems to be flaky. Pass in the first commit but not the second one.

vivekratnavel:
@xiaoyuyao Thanks for working on this and @dineshchitlangia for reviewing it!

hadoop-ozone/integration-test/src/test/resources/log4j.properties
```diff
@@ -19,3 +19,6 @@ log4j.appender.stdout.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} (%F:%
 
 log4j.logger.org.apache.hadoop.security.ShellBasedUnixGroupsMapping=ERROR
 log4j.logger.org.apache.hadoop.util.NativeCodeLoader=ERROR
+
+log4j.logger.org.apache.ratis.grpc.server.GrpcLogAppender=DEBUG
+log4j.logger.org.apache.ratis.server.impl.RaftServerImpl=DEBUG
```
adoroszlai, line 24: Hi @xiaoyuyao,

Were these Ratis log levels intentionally set to DEBUG in the final commit, or is it a leftover that should have been removed?

xiaoyuyao:
This should be removed. Thanks @adoroszlai

adoroszlai:
> This should be removed.

Thanks for the confirmation.  Created HDDS-3967 for this.  Will submit PR soon.

PR #1102, opened 2020-06-20
HDDS-3825. Show rule status of SCM safemode

maobaolong:

Without this PR, we can't get the reason why scm still in the safemode, now, we can get the each rule status, and know the condition to exit safemode.

https://issues.apache.org/jira/browse/HDDS-3825

## How was this patch tested?

```bash
ozone admin safemode status --verbose
SCM is in safe mode.
validated:false, DataNodeSafeModeRule, registeredDns 0 should >= requiredDns 1
validated:true, HealthyPipelineSafeModeRule, currentHealthyPipelineCount 0 should >= healthyPipelineThresholdCount 0
validated:true, ContainerSafeModeRule, currentContainerThreshold 1.0 should >= safeModeCutoff 0.99
validated:true, AtleastOneDatanodeReportedRule, currentReportedPipelineCount 0 should >= thresholdCount 0
```


maobaolong:
@elek Please take a look, thank you.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/client/ScmClient.java
```diff
@@ -210,6 +212,15 @@ Pipeline createReplicationPipeline(HddsProtos.ReplicationType type,
    */
   boolean inSafeMode() throws IOException;
 
+  /**
+   * Get the safe mode statuses of each rules.
```
dineshchitlangia, line 216:
```suggestion
   * Get the safe mode status of all rules.
```

maobaolong:
@dineshchitlangia Thanks for your review and the suggestion, i've commit all of your suggestion commit, thank for your help. PTAL.

maobaolong:
@dineshchitlangia Thanks for your +1. 😃

dineshchitlangia:
@maobaolong I will merge this. However, going forward could you pls also include any relevant tests with your changes as the code coverage is now being looked at more seriously in the community for the benefits it brings :)
For this patch, pls file another jira to add the tests.

dineshchitlangia:
@maobaolong I filed https://issues.apache.org/jira/browse/HDDS-3849 and assigned to you Thanks.

maobaolong:
@dineshchitlangia Thanks you for create the ticket for me, i'd love to do it with pleasure.

PR #1093, opened 2020-06-18
HDDS-3826. Split Ozone FS acceptance tests

adoroszlai:
## What changes were proposed in this pull request?

1. Split monolithic Ozone FS acceptance test case into several steps.  This improves feedback about test progress, and makes easier to find out what's wrong in case of a failure.  (Unfortunately most subsequent steps still depend on earlier ones.  Improving that would be nice, but would make the test slower due to the necessary extra setup steps.)
2. Run test setup in proper `Setup` instead of a test case.
3. Split test setup into reusable keywords.
4. Add functions for formatting Ozone FS URLs with both o3fs and ofs flavor.  Includes its own test cases.
5. Test execution for ofs and o3fs is controlled outside of the test script.  This is necessary to avoid duplication of test cases and/or keywords.
6. Pass `SECURITY_ENABLED` and similar variables as Robot variables (`-v`), not environment ones (`-e`).  This allows setting default value in Robot tests, which makes it easier to run the test from command-line.

Most of this is in preparation for bucket links, as the same test should be executed for regular buckets and links.  Submitting as a separate PR for easier review.

Note that this PR is based on top of the change for #1091, a minor improvement which is necessary to save output of multiple executions of the same test suite (`ozonefs.robot` in this case).

https://issues.apache.org/jira/browse/HDDS-3826

## How was this patch tested?

Ran tests for lib locally (no docker-compose):

```
$ robot lib
...
Lib.Fs Tests                                                          | PASS |
...
Lib.String Tests                                                      | PASS |
...
```

Ran Ozone FS tests inside docker container bypassing `test.sh`:

```
$ docker-compose exec scm robot smoketest/ozonefs/ozonefs.robot
...
Ozonefs :: Ozone FS tests                                             | PASS |
20 critical tests, 20 passed, 0 failed

$ docker-compose exec scm robot -v SCHEME:o3fs smoketest/ozonefs/ozonefs.robot
...
Ozonefs :: Ozone FS tests                                             | PASS |
20 critical tests, 20 passed, 0 failed
```

Full acceptance test suite:
https://github.com/adoroszlai/hadoop-ozone/runs/783679213
https://github.com/adoroszlai/hadoop-ozone/runs/784247369

PR #1091, opened 2020-06-18
HDDS-3094. Save each output of smoketest executed multiple times

adoroszlai:
## What changes were proposed in this pull request?

Add a numeric suffix to Robot test output file names to avoid overwriting previous runs.

https://issues.apache.org/jira/browse/HDDS-3094

## How was this patch tested?

Ran test in `ozone-topology` env, which calls `readdata` test twice (with different set of running datanodes).

```
$ cd hadoop-ozone/dist/target/ozone*/compose/ozone-topology
$ ./test.sh
...
Output:  /tmp/smoketest/ozone-topology/result/robot-ozone-topology-ozone-topology-readdata-scm.xml
...
Output:  /tmp/smoketest/ozone-topology/result/robot-ozone-topology-ozone-topology-readdata-scm-1.xml
...
```

https://github.com/adoroszlai/hadoop-ozone/runs/783624066

PR #1104, opened 2020-06-21
HDDS-3612. Allow mounting bucket under other volume

adoroszlai:

Implement [bucket links](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/docs/content/design/ozone-volume-management.md#implementation-details).

#### Implementation details (from the design doc)

 * `ozone sh bucket link` operation creates a link bucket.  Links are like regular buckets, stored in DB the same way, but with two new, optional pieces of information: source volume and bucket.  (The bucket being referenced by the link is called "source", not "target", to follow symlink terminology.)
 * Link buckets share the namespace with regular buckets.  If a bucket or link with the same name already exists, a `BUCKET_ALREADY_EXISTS` result is returned.
 * (Unlike previous per-user S3 volume mapping) Link buckets are not inherently specific to a user, access is restricted only by ACL.
 * Links are persistent, ie. they can be used until they are deleted.
 * Existing bucket operations (info, delete, ACL) work on the link object in the same way as they do on regular buckets.  No new link-specific RPC is required.
 * Links are followed for key operations (list, get, put, etc.).  Read permission on the link is required for this.
 * Checks for existence of the source bucket, as well as ACL, are performed only when following the link (similar to symlinks).  Source bucket is not checked when operating on the link bucket itself (eg. deleting it).  This avoids the need for reverse checks for each bucket delete or ACL change.
 * Bucket links are generic, not restricted to the `s3v` volume.

#### Audit

Audit message for link creation mentions the bucket to which the link points (`sourceVolume` and `sourceBucket`):

```
OMAudit | ... | op=CREATE_BUCKET {volume=60748-target, bucket=link1, gdprEnabled=null, acls=[user:hadoop:a[ACCESS], group:users:a[ACCESS]], isVersionEnabled=false, storageType=DISK, creationTime=1592631444545, bucketEncryptionKey=null, sourceVolume=60748-source, sourceBucket=bucket1} | ret=SUCCESS |  
```

As well as audit messages for key operations issued on links:

```
OMAudit | ... | op=ALLOCATE_KEY {volume=60748-target, bucket=link1, key=key1, dataSize=671, replicationType=RATIS, replicationFactor=THREE, sourceVolume=60748-source, sourceBucket=bucket1} | ret=SUCCESS |  
OMAudit | ... | op=DELETE_KEY {volume=60748-target, bucket=link1, key=key2, dataSize=0, replicationType=RATIS, replicationFactor=ONE, sourceVolume=60748-source, sourceBucket=bucket1} | ret=SUCCESS |  
```

Response for `ALLOCATE_KEY` contains the real (source) volume/bucket, so `COMMIT_KEY` does not know about the link (but it is prepared to handle it anyway):

```
OMAudit | ... | op=COMMIT_KEY {volume=60748-source, bucket=bucket1, key=key1, dataSize=671, replicationType=RATIS, replicationFactor=ONE} | ret=SUCCESS |  
```

Based on top of changes from #1091 and #1093.

https://issues.apache.org/jira/browse/HDDS-3612

## How was this patch tested?

Added acceptance tests for testing bucket links via Ozone Shell.

Slightly tweaked existing S3 acceptance tests are executed on both normal buckets and bucket links.

Existing Ozone FS acceptance tests (reorganized in #1093) are also executed on links.

https://github.com/adoroszlai/hadoop-ozone/runs/792817258


adoroszlai:
@bharatviswa504 may I ask you to review this?

elek:
Thanks to working on this @adoroszlai 

Checked only on high-level (yet) but structure looks good to me. This is exactly what we planned. Will try out in a deployment.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
```diff
@@ -136,54 +137,49 @@ public void createBucket(OmBucketInfo bucketInfo) throws IOException {
...
-          .addAllMetadata(bucketInfo.getMetadata());
+
+      boolean hasSourceVolume = bucketInfo.getSourceVolume() != null;
+      boolean hasSourceBucket = bucketInfo.getSourceBucket() != null;
+
+      if (hasSourceBucket != hasSourceVolume) {
+        throw new OMException("Both source volume and source bucket are " +
+            "required for bucket links",
+            OMException.ResultCodes.INVALID_REQUEST);
+      }
```
bharatviswa504, line 150: Old write code is not being used anymore.
This logic needs to be added to new Class OMBucketCreateRequest.java

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
```diff
@@ -136,54 +137,49 @@ public void createBucket(OmBucketInfo bucketInfo) throws IOException {
...
+
+      BucketEncryptionKeyInfo.Builder bekb =
+          createBucketEncryptionKeyInfoBuilder(bek);
+
+      OmBucketInfo.Builder omBucketInfoBuilder = bucketInfo.toBuilder()
+          .setCreationTime(Time.now());
+
+      List<OzoneManagerProtocolProtos.OzoneAclInfo> defaultAclList =
```
bharatviswa504, line 163: Question: No, where we checked source volume/bucket exists or not.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2020,60 +2026,72 @@ public OmBucketInfo getBucketInfo(String volume, String bucket)
    */
   @Override
   public OpenKeySession openKey(OmKeyArgs args) throws IOException {
+    ResolvedBucket bucket = resolveBucketLink(args);
+
     if (isAclEnabled) {
       try {
         checkAcls(ResourceType.KEY, StoreType.OZONE, ACLType.WRITE,
-            args.getVolumeName(), args.getBucketName(), args.getKeyName());
+            bucket.realVolume(), bucket.realBucket(), args.getKeyName());
```
bharatviswa504, line 2034: Same, for all, write requests old code is not used anymore.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2145,37 +2168,51 @@ public OmKeyLocationInfo allocateBlock(OmKeyArgs args, long clientID,
    */
   @Override
   public OmKeyInfo lookupKey(OmKeyArgs args) throws IOException {
+    ResolvedBucket bucket = resolveBucketLink(args);
```
bharatviswa504, line 2171: Now for each operation if it is link bucket where if it has sourceVolume/SourceBucket we do checkAcls twice. One with READ permission on sourceBucket/SourceVolume. It is not clear why do we need that. 

If it requires, do you think we need to have an API where we can check all required ACLS with a single checkAcl call?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2197,20 +2234,25 @@ public void renameKey(OmKeyArgs args, String toKeyName) throws IOException {
    */
   @Override
   public void deleteKey(OmKeyArgs args) throws IOException {
+    Map<String, String> auditMap = args.toAuditMap();
```
bharatviswa504, line 2237: The same comment for all write requests:
Can we remove the changes from old write code path, which is not required? It is unnecessary now.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
bharatviswa504: Okay, I see is in resolveBucketLink we check Acls for read on provided bucket, and in actual request check acls on sourceBucket/sourceVolume which it is resolved to. Why do we need this?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -174,6 +175,12 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
     Result result = null;
 
     try {
+      ResolvedBucket bucket = ozoneManager.resolveBucketLink(keyArgs);
+      keyArgs = bucket.update(keyArgs);
```
bharatviswa504, line 179: Looks like this logic is needed for all KeyRequests. Can we move this to a common method, instead of duplicating it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -149,6 +150,12 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
     List<OmKeyInfo> missingParentInfos;
 
     try {
+      ResolvedBucket bucket = ozoneManager.resolveBucketLink(keyArgs);
+      keyArgs = bucket.update(keyArgs);
```
bharatviswa504, line 154: And also do you think it would be useful to log actual bucket/volume request, and then resolvedBucket/resolvedVolume instead of just updating volumeName/bucketName directly.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
bharatviswa504: Reading more understood that if we create link for /vol1/buck1 -> /vol2/buck2 (source)

We create in DB /vol1/buck1 and they have sourceVolume as vol2 and sourceBucket as buck2.

Now, when someone calls lookupKey on unresolved bucket, during actual request of lookupKey, this will result in Bucket_NOT_FOUND.  Do you think, we need to make sure that source volume/source bucket exists during link creation to avoid such scenarios? 

My reason was this looks strange, the user thinks he created a link bucket with some source volume/source bucket that has passed without any issues, but now when creating key it is saying bucket does not exist.

Following ln -s <<source>> <<dest>> looks confusing in our scenario.

bharatviswa504:
Question: Now when list at the root "/" with ofs filesystem, for link buckets they will be displayed as a new entry of only resolved buckets will be shown. What should be the behavior here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
adoroszlai: Source bucket can be deleted any time after the link is created.  We would have to perform a reverse lookup to check if it leaves any dangling link.  Since this is not done, checking upon creation would be inconsistent.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
adoroszlai: Read permission on the link is required to follow it.  Do you propose to completely skip ACL on link and allow anyone to use it?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
adoroszlai: I would have preferred removing old write code path before implementing links to avoid duplicate work.  However, now that these changes are in place, I prefer keeping them, and removing old write code path completely in a separate step.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
adoroszlai: Audit log includes both.

adoroszlai:
> Question: Now when list at the root "/" with ofs filesystem, for link buckets they will be displayed as a new entry of only resolved buckets will be shown. What should be the behavior here.

I'm afraid I don't understand the question.  Can you please give an example?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
adoroszlai: Thanks, I missed that.  Will update the patch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
adoroszlai: Thanks, extracted to a method.

adoroszlai:
Thanks @bharatviswa504 for the review.  I have updated the patch based on your code comments.

adoroszlai:
@ChenSammi this feature is continuation of S3 volume mapping changes.  Can you please take a look?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
elek: This is the same as the behavior of sym links:

```
mkdir source
ln -s source dest
rm r source
```
Now you have a wrong pointer. You can list the content of the directory, but when you try to create a fie it will fail.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
bharatviswa504: I am not sure what is the expected behavior, how is  this semantics derived?
Because with this approach for all mounted buckets, we do 2 checkAcls, we might put pressure on Ranger.

Can we rely on underlying bucket acls, as anyway we verify already. Any downside/security issue?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
bharatviswa504: Makes sense to me.

bharatviswa504:
> > Question: Now when list at the root "/" with ofs filesystem, for link buckets they will be displayed as a new entry of only resolved buckets will be shown. What should be the behavior here.
> 
> I'm afraid I don't understand the question. Can you please give an example?

when I perform ls on volume/bucket, which has below buckets. It will show only resolved buckets or all the buckets? Or it follows normal ls semantics?

if it has mounted buckets

b1 -> b2
b2
b4 -> b5

what should be the behavior of ls in this case?

adoroszlai:
> > > Question: Now when list at the root "/" with ofs filesystem, for link buckets they will be displayed as a new entry of only resolved buckets will be shown. What should be the behavior here.
> > 
> > 
> > I'm afraid I don't understand the question. Can you please give an example?
> 
> when user perform `ls` on volume/bucket, which has below buckets. It will show only resolved buckets or all the buckets? Or it follows normal ls semantics?
> 
> if it has mounted buckets
> 
> b1 -> b2
> b2
> b4 -> b5
> 
> what should be the behavior of ls in this case?

Thanks for the clarification.  `ls` on volume shows buckets and links in that volume, will not resolve links.  `ls` on bucket shows the keys in that bucket.  `ls` on link shows keys in the bucket that the link points to.

Example created by the robot test:

```
$ ozone fs -ls ofs://om/63433-target/
Found 7 items
drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:50 ofs://om/63433-target/bucket3
drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:49 ofs://om/63433-target/dangling-link
drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:50 ofs://om/63433-target/link2
drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:51 ofs://om/63433-target/link3
drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:51 ofs://om/63433-target/loop1
drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:51 ofs://om/63433-target/loop2
drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:51 ofs://om/63433-target/loop3
```

```
$ ozone fs -ls ofs://om/63433-source/bucket1
Found 2 items
-rw-rw-rw-   3 hadoop hadoop        671 2020-07-07 05:49 ofs://om/63433-source/bucket1/key1
-rw-rw-rw-   3 hadoop hadoop        372 2020-07-07 05:51 ofs://om/63433-source/bucket1/key3
```

```
$ ozone fs -ls ofs://om/63433-target/link3
Found 2 items
-rw-rw-rw-   3 hadoop hadoop        671 2020-07-07 05:49 ofs://om/63433-target/link3/key1
-rw-rw-rw-   3 hadoop hadoop        372 2020-07-07 05:51 ofs://om/63433-target/link3/key3
```

```
$ ozone sh bucket info /63433-target/link3
{
  "volumeName" : "63433-target",
  "name" : "link3",
...
  "sourceVolume" : "63433-source",
  "sourceBucket" : "bucket1"
}
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
bharatviswa504: Okay error seems to be confusing because the link bucket is happening through create bucket, with parameters sourcevolume/sourceBucket.

I am not sure if someone will use our direct API's OzoneBucket/RpcClient. 

But the explanation makes sense to me.

bharatviswa504:
> > > > Question: Now when list at the root "/" with ofs filesystem, for link buckets they will be displayed as a new entry of only resolved buckets will be shown. What should be the behavior here.
> > > 
> > > 
> > > I'm afraid I don't understand the question. Can you please give an example?
> > 
> > 
> > when user perform `ls` on volume/bucket, which has below buckets. It will show only resolved buckets or all the buckets? Or it follows normal ls semantics?
> > if it has mounted buckets
> > b1 -> b2
> > b2
> > b4 -> b5
> > what should be the behavior of ls in this case?
> 
> Thanks for the clarification. `ls` on volume shows buckets and links in that volume, will not resolve links. `ls` on bucket shows the keys in that bucket. `ls` on link shows keys in the bucket that the link points to.
> 
> Example created by the robot test:
> 
> ```
> $ ozone fs -ls ofs://om/63433-target/
> Found 7 items
> drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:50 ofs://om/63433-target/bucket3
> drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:49 ofs://om/63433-target/dangling-link
> drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:50 ofs://om/63433-target/link2
> drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:51 ofs://om/63433-target/link3
> drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:51 ofs://om/63433-target/loop1
> drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:51 ofs://om/63433-target/loop2
> drwxrwxrwx   - hadoop hadoop          0 2020-07-07 05:51 ofs://om/63433-target/loop3
> ```
> 
> ```
> $ ozone fs -ls ofs://om/63433-source/bucket1
> Found 2 items
> -rw-rw-rw-   3 hadoop hadoop        671 2020-07-07 05:49 ofs://om/63433-source/bucket1/key1
> -rw-rw-rw-   3 hadoop hadoop        372 2020-07-07 05:51 ofs://om/63433-source/bucket1/key3
> ```
> 
> ```
> $ ozone fs -ls ofs://om/63433-target/link3
> Found 2 items
> -rw-rw-rw-   3 hadoop hadoop        671 2020-07-07 05:49 ofs://om/63433-target/link3/key1
> -rw-rw-rw-   3 hadoop hadoop        372 2020-07-07 05:51 ofs://om/63433-target/link3/key3
> ```
> 
> ```
> $ ozone sh bucket info /63433-target/link3
> {
>   "volumeName" : "63433-target",
>   "name" : "link3",
> ...
>   "sourceVolume" : "63433-source",
>   "sourceBucket" : "bucket1"
> }
> ```

Thanks for your clarification. Even ls on the filesystem is same behavior it shows link also. I think we should be fine here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
bharatviswa504: Any info on this question?

adoroszlai: > I am not sure what is the expected behavior, how is this semantics derived?
> Can we rely on underlying bucket acls, as anyway we verify already. Any downside/security issue?

This matches unix symlinks permissions: if you have read access on the link, you see where it points to and can follow it.  ACL was discussed in design doc update (#1009) and the current behavior was approved.  Please check with @arp7 if we need any changes.  Also pinging @ChenSammi for input.

arp7: > This matches unix symlinks permissions: if you have read access on the link, you see where it points to and can follow it. 
+1

bharatviswa504:
@adoroszlai This patch needs a rebase. (This has been due to one of my commit only).
I will take a final look at it today, and will post my comments.

adoroszlai:
> This patch needs a rebase. (This has been due to one of my commit only).

Done.

> I will take a final look at it today, and will post my comments.

Thanks.

hadoop-ozone/dist/src/main/smoketest/basic/links.robot
```diff
@@ -0,0 +1,152 @@
...
+
+ACL verified on source bucket
+    Run Keyword if    '${SECURITY_ENABLED}' == 'true'    ACL verified on source bucket
+
+Loop in link chain is detected
+                        Execute                     ozone sh bucket link ${target}/loop1 ${target}/loop2
+                        Execute                     ozone sh bucket link ${target}/loop2 ${target}/loop3
+                        Execute                     ozone sh bucket link ${target}/loop3 ${target}/loop1
+    ${result} =         Execute And Ignore Error    ozone sh key list ${target}/loop2
+                        Should Contain              ${result}    INTERNAL_ERROR
```
bharatviswa504, line 140: This needs to be updated, once error code is modified.

hadoop-ozone/dist/src/main/compose/ozone/test.sh
```diff
@@ -26,24 +26,24 @@ source "$COMPOSE_DIR/../testlib.sh"
...
+  for bucket in link bucket; do
+    execute_robot_test scm -v SCHEME:${scheme} -v BUCKET_TYPE:${bucket} ozonefs/ozonefs.robot
+  done
+done
 
 execute_robot_test scm security/ozone-secure-token.robot
 
-execute_robot_test scm s3
+for bucket in link generated; do
+  execute_robot_test scm -v BUCKET:${bucket} s3
```
bharatviswa504, line 45: S3 tests not using this bucket right?

For example

Delete existing bucket
# Bucket already is created in Test Setup.
    ${bucket} =                Create bucket
    Execute AWSS3APICli        delete-bucket --bucket ${bucket}

And bucket is returned from 
Delete existing bucket
# Bucket already is created in Test Setup.
    ${bucket} =                Create bucket
    Execute AWSS3APICli        delete-bucket --bucket ${bucket}

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
```diff
@@ -88,21 +88,20 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
 
-    OzoneManagerProtocolProtos.KeyArgs deleteKeyArgs =
+    OzoneManagerProtocolProtos.KeyArgs keyArgs =
         deleteKeyRequest.getKeyArgs();
+    Map<String, String> auditMap = buildKeyArgsAuditMap(keyArgs);
 
-    String volumeName = deleteKeyArgs.getVolumeName();
-    String bucketName = deleteKeyArgs.getBucketName();
-    String keyName = deleteKeyArgs.getKeyName();
+    String volumeName = keyArgs.getVolumeName();
```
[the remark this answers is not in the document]
adoroszlai: > We can skip getting from original Args, as anyway final Volume/Bucket we get is from resolvedBucket returned keyArgs.

Unfortunately not, because `volumeName` and `bucketName` are used later to log result.  If `resolveBucketLink` throws exception (eg. due to lack of permission), then we exit from `try` earlier than getting `resolvedBucket` and we need the original values from request.

https://github.com/apache/hadoop-ozone/blob/fbf04a33ecad2d23e8bd57ecceca4842a9cf5281/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java#L173-L183

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3314,4 +3396,59 @@ private void startJVMPauseMonitor() {
...
+
+    String volumeName = volumeAndBucket.getLeft();
+    String bucketName = volumeAndBucket.getRight();
+    OmBucketInfo info = bucketManager.getBucketInfo(volumeName, bucketName);
+    if (!info.isLink()) {
+      return volumeAndBucket;
+    }
+
+    if (!visited.add(volumeAndBucket)) {
+      throw new OMException("Detected loop in bucket links", INTERNAL_ERROR);
```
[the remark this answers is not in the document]
adoroszlai: Sure.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -85,10 +87,11 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
     OMMetrics omMetrics = ozoneManager.getMetrics();
     omMetrics.incNumKeyDeletes();
-    Map<String, String> auditMap = null;
     String volumeName = deleteKeyArgs.getVolumeName();
     String bucketName = deleteKeyArgs.getBucketName();
-    String keyName = "";
+    Map<String, String> auditMap = new LinkedHashMap<>();
+    auditMap.put(VOLUME, volumeName);
```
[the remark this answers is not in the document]
adoroszlai: Similar to `volumeName` and `bucketName` variables, this is also required for the case when we encounter exception before reaching `bucket.audit()`.

hadoop-ozone/dist/src/main/smoketest/basic/links.robot
adoroszlai: Done.

hadoop-ozone/dist/src/main/compose/ozone/test.sh
adoroszlai: > S3 tests not using this bucket right?

Other tests do use bucket or link created by setup.  Thanks for spotting this instance, fixed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadAbortRequest.java
```diff
@@ -152,27 +158,29 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-        OMAction.ABORT_MULTIPART_UPLOAD, buildKeyArgsAuditMap(keyArgs),
+        OMAction.ABORT_MULTIPART_UPLOAD, auditMap,
         exception, getOmRequest().getUserInfo()));
 
     switch (result) {
     case SUCCESS:
       LOG.debug("Abort Multipart request is successfully completed for " +
-              "KeyName {} in VolumeName/Bucket {}/{}", keyName, volumeName,
-          bucketName);
+              "KeyName {} in VolumeName/Bucket {}/{}", keyName, requestedVolume,
```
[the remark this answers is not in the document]
adoroszlai: Updated, thanks.

adoroszlai:
Thanks @bharatviswa504 for the review.  I have updated the patch based on your comments.

> When deleting a bucket, we check bucket is empty or not. Do we need to check resolvedBucket is empty or not?

Link can be deleted regardless of the existence or emptiness of the real bucket it points to.  So no need to check resolved bucket.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
```diff
@@ -111,6 +110,10 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
     OMClientResponse omClientResponse = null;
     Result result = null;
     try {
+      keyArgs = resolveBucketLink(ozoneManager, keyArgs, auditMap);
```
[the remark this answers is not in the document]
adoroszlai: >  But once HA becomes the default, we can optimize this, as in HA there is only a single thread executor.

Does that mean we can completely get rid of the locks?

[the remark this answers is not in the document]
bharatviswa504: we cannot completely get rid of locks, as there are readers.

bharatviswa504:
> Thanks @bharatviswa504 for the review. I have updated the patch based on your comments.
> 
> > When deleting a bucket, we check bucket is empty or not. Do we need to check resolvedBucket is empty or not?
> 
> Link can be deleted regardless of the existence or emptiness of the real bucket it points to. So no need to check resolved bucket.

This is the semantics for link buckets?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -172,6 +172,10 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
     IOException exception = null;
 
     try {
+      keyArgs = resolveBucketLink(ozoneManager, keyArgs, auditMap);
+      volumeName = keyArgs.getVolumeName();
+      bucketName = keyArgs.getBucketName();
```
bharatviswa504, line 177: openKeyName should be generated from resolveBucketLink.

bharatviswa504:
Overall LGTM.
Few minor comments, once those are addressed good to go.

bharatviswa504:
@adoroszlai Any comments on left over 2 questions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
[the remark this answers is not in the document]
adoroszlai: OK, I'll check if I can update the patch to address this.

adoroszlai:
> > > When deleting a bucket, we check bucket is empty or not. Do we need to check resolvedBucket is empty or not?
> > 
> > Link can be deleted regardless of the existence or emptiness of the real bucket it points to. So no need to check resolved bucket.
> 
> This is the semantics for link buckets, because this is not a delete link, it is same delete bucket op which OM has received.

The same "delete bucket" op works for deleting links.  Since links cannot have direct entries, the emptiness check will not find any, hence delete is always allowed.  We don't resolve the bucket the link points to, because its emptiness does not affect whether we can delete the link.

Please let me know if I misunderstood the question.

ChenSammi:
@adoroszlai, thanks for working on this feature. I did some simple tests locally and looks good. A few questions about the expected behavior, 
1.  When the source bucket is deleted, bucket link still exists.  It's an expected behavior, right?
2. (Unlike previous per-user S3 volume mapping) Link buckets are not inherently specific to a user, access is restricted only by ACL. Not very clear about  what it means.  would you explain it further? Can we set different ACL on source bucket and bucket link? 
3. We're working on Volume quota  implementation.  I suppose bucket link will require 0 quota from it's parent volume, only the source bucket size will be controlled by it's volume's quota,  right?

adoroszlai:
> 1. When the source bucket is deleted, bucket link still exists.  It's an expected behavior, right?

Yes.

> 2. (Unlike previous per-user S3 volume mapping) Link buckets are not inherently specific to a user, access is restricted only by ACL. Not very clear about  what it means.  would you explain it further?

Previously each user had its own S3 volume and only by sharing the AWS access key could others use it.  Now if you create a bucket link, other users can access it as long as ACL permits (similar to other Ozone objects).

> Can we set different ACL on source bucket and bucket link?

Yes.  When following links eg. to create a key, only read permission is checked on the link.  The source bucket is still checked for the permission required for the specific operation (eg. key list, create, or delete).

> 3. We're working on Volume quota  implementation.  I suppose bucket link will require 0 quota from it's parent volume, only the source bucket size will be controlled by it's volume's quota,  right?

I'm not familiar with quotas, but I guess links should work like empty buckets in this respect.  Keys are created in the source bucket, so they should not affect space usage in the link's volume.

ChenSammi:
> 
> 
> > I did some simple tests locally and looks good.
> 
> Thanks @ChenSammi for that.
> 
> > 1. When the source bucket is deleted, bucket link still exists.  It's an expected behavior, right?
> 
> Yes.
> 
> > 1. (Unlike previous per-user S3 volume mapping) Link buckets are not inherently specific to a user, access is restricted only by ACL. Not very clear about  what it means.  would you explain it further?
> 
> Previously each user had its own S3 volume and only by sharing the AWS access key could others use it. Now if you create a bucket link, other users can access it as long as ACL permits (similar to other Ozone objects).
> 
> > Can we set different ACL on source bucket and bucket link?
> 
> Yes. When following links eg. to create a key, only read permission is checked on the link. The source bucket is still checked for the permission required for the specific operation (eg. key list, create, or delete).
> 
> > 1. We're working on Volume quota  implementation.  I suppose bucket link will require 0 quota from it's parent volume, only the source bucket size will be controlled by it's volume's quota,  right?
> 
> I'm not familiar with quotas, but I guess links should work like empty buckets in this respect. Keys are created in the source bucket, so they should not affect space usage in the link's volume.

@adoroszlai , thanks for the explanation.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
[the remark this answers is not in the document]
adoroszlai: Opened https://issues.apache.org/jira/browse/HDDS-3971 for the improvement.

bharatviswa504:
> > > > When deleting a bucket, we check bucket is empty or not. Do we need to check resolvedBucket is empty or not?
> > > 
> > > 
> > > Link can be deleted regardless of the existence or emptiness of the real bucket it points to. So no need to check resolved bucket.
> > 
> > 
> > This is the semantics for link buckets, because this is not a delete link, it is same delete bucket op which OM has received.
> 
> The same "delete bucket" op works for deleting links. Since links cannot have direct entries, the emptiness check will not find any, hence delete is always allowed. We don't resolve the bucket the link points to, because its emptiness does not affect whether we can delete the link.
> 
> Please let me know if I misunderstood the question.

Makes sense to me. Thanks for the explanation.

bharatviswa504:
I have one suggestion, can we make this mount configurable option, as this is required only when old S3 buckets needs to be exposed via O3fs/ofs. In this way, we can avoid extra bucket checks. (I know that it is in-memory cache, but still I feel if some one does not require this feature, it is totally disabled). I think we can make that check in 2 places. One in create bucket/and other in resolveBucketLink.

Fine to do in another Jira also.

bharatviswa504:
> I have one suggestion, can we make this mount configurable option, as this is required only when old S3 buckets needs to be exposed via O3fs/ofs. In this way, we can avoid extra bucket checks. (I know that it is in-memory cache, but still I feel if some one does not require this feature, it is totally disabled). I think we can make that check in 2 places. One in create bucket/and other in resolveBucketLink.
> 
> Fine to do in another Jira also.

I take back this comment. This feature helps to expose buckets created via shell/ofs to S3.  So making it configurable is not needed. (As by default if user decides to expose any bucket created via shell, it can be exposed without any restart.) 

Thank You @arp7 for the offline discussion.

adoroszlai:
> I have one suggestion, can we make this mount configurable option, as this is required only when old S3 buckets needs to be exposed via O3fs/ofs. In this way, we can avoid extra bucket checks. (I know that it is in-memory cache, but still I feel if some one does not require this feature, it is totally disabled). I think we can make that check in 2 places. One in create bucket/and other in resolveBucketLink.
> 
> Fine to do in another Jira also.

Thanks for the suggestion, sounds good to me.  I think if we want to disable it by default, then it should be done as part of this task.  If it's enabled by default, then it's fine to add the config later in separate task.  Let me know what should be the default value.  CC @arp7 @elek

adoroszlai:
> I take back this comment. This feature helps to expose buckets created via shell/ofs to S3. So making it configurable is not needed. (As by default if user decides to expose any bucket created via shell, it can be exposed without any restart.)
> 
> Thank You @arp7 for the offline discussion.

OK.

ChenSammi:
Thanks @adoroszlai for the contribution and @bharatviswa504 for the review.

adoroszlai:
Thanks @arp7, @elek for high-level review, @bharatviswa504 for the detailed reviews, @ChenSammi for testing and committing it.

PR #1105, opened 2020-06-22
HDDS-3828. Configuration parsing of ozone insight should be based on fields

elek:

Ozone insight command can print out configuration related to a specific ozone component with parsing the @Config object.

But the usage of config annotations has been changed since HDDS-2661. We should check the annotated fields not methods.

https://issues.apache.org/jira/browse/HDDS-3828

## How was this patch tested?

With docker-compose based test:

https://www.youtube.com/watch?v=ZLknuMIl_jw&list=PLCaV-jpCBO8UK5Ged2A_iv3eHuozzMsYv&index=5&t=3s

### DIFF
--- a/hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/ConfigurationSubCommand.java
+++ b/hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/ConfigurationSubCommand.java
@@ -60,15 +60,26 @@ public class ConfigurationSubCommand extends BaseInsightSubCommand
     return null;
   }
 
-  private void showConfig(Class clazz, Type type) {
+  protected void showConfig(Class clazz, Type type) {
     OzoneConfiguration conf = new OzoneConfiguration();
     conf.addResource(getHost(conf, new Component(type)) + "/conf");
+    printConfig(clazz, conf);
+  }
+
+  /**
+   * Print all the configuration annotated on the class or any superclass.
+   */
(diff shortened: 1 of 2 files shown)


hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/ConfigurationSubCommand.java
```diff
@@ -71,9 +71,9 @@ private void showConfig(Class clazz, Type type) {
 
     String prefix = configGroup.prefix();
 
-    for (Method method : clazz.getMethods()) {
-      if (method.isAnnotationPresent(Config.class)) {
-        Config config = method.getAnnotation(Config.class);
+    for (Field field : clazz.getDeclaredFields()) {
```
adoroszlai, line 74: Shouldn't it handle inherited fields, too?  I don't think we have such a configuration object hierarchy currently, but injection is prepared for it:

https://github.com/apache/hadoop-ozone/blob/53549eea6f5dfc8e4cc9594583d7510b206a8acd/hadoop-hdds/config/src/main/java/org/apache/hadoop/hdds/conf/ConfigurationReflectionUtil.java#L37-L44

elek: Good point.

But I am afraid we need a unit test to here to prove if it works as it's a more complex logic. I added one to prove it work.

PR #1107, opened 2020-06-22
HDDS-3807. Propagate raft log disks info to SCM from datanode.

bshashikant:

Add ratis volume info to  node report

https://issues.apache.org/jira/browse/HDDS-3807

## How was this patch tested?

Added unit tests.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/XceiverServerSpi.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/XceiverServerSpi.java
@@ -22,7 +22,6 @@ import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.datanode.proto.ContainerProtos
     .ContainerCommandRequestProto;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
-import org.apache.hadoop.hdds.protocol.proto.StorageContainerDatanodeProtocolProtos;
 import org.apache.hadoop.hdds.protocol.proto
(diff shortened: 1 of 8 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/XceiverServerSpi.java
```diff
@@ -83,4 +85,12 @@ default void removeGroup(HddsProtos.PipelineID pipelineId)
    * @return list of report for each pipeline.
    */
   List<PipelineReport> getPipelineReport();
+
+  /**
+   * Get storage report for the XceiverServer instance.
+   * @return list of report for each storage location.
+   */
+  default List<MetadataStorageReportProto> getStorageReport() throws IOException {
```
elek, line 93: It seems to be a good time the remove XceiverServerSpi interface. Originally we had two separated implementation to connect to the datanode. Today we have only one. One interface is used between the client and the datanode, and the other one between datanode and ratis (datanode). As this example shows, the two interface shouldn't be the same.

elek:
Checkstyle violation seems to be reported.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/XceiverServerSpi.java
bshashikant: Definitely, yes..
Filed https://issues.apache.org/jira/browse/HDDS-3963 for the same

elek:
Merging it as checkstyle issues are fixed. Thanks @bshashikant the contribution.

PR #1111, opened 2020-06-23
HDDS-3854. Fix error return value of KeyValueBlockIterator#hasNext

runzhiwang:

hasNext use a recursion, but the recursion hasNext does not return value. So if the first level of hasNext does not  return true, though the  recursion hasNext return true, the first level of hasNext still return false. Besides I think while loop is easier.

image

https://issues.apache.org/jira/browse/HDDS-3854

## How was this patch tested?

Existed tests.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueBlockIterator.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueBlockIterator.java
@@ -124,7 +124,7 @@ public class KeyValueBlockIterator implements BlockIterator<BlockData>,
     if (nextBlock != null) {
       return true;
     }
-    if (blockIterator.hasNext()) {
+    while (blockIterator.hasNext()) {
       KeyValue block = blockIterator.next();
       if (blockFilter.filterKey(null, block.getKey(), null)) {
         nextBlock = BlockUtils.getBlockData(block.getValue());
@@ -134,7 +134,6 @@ public class KeyValueBlockIterator implements BlockIterator<BlockData>,
         }
         return true;
       }
-      hasNext();
     }
     return false;
   }
--- a/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/keyvalue/TestKeyValueBlockIterator.java
+++ b/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/keyvalue/TestKeyValueBlockIterator.java
@@ -209,7 +209,7 @@ public class TestKeyValueBlockIterator {
   @Test
   public void testKeyValueBlockIteratorWithFilter() throws Exception {
     long containerId = 103L;
-    int deletedBlocks = 5;
+    int deletedBlocks = 10;
     int normalBlocks = 5;
     createContainerWithBlocks(containerId, normalBlocks, deletedBlocks);
     String containerPath = new File(containerData.getMetadataPath())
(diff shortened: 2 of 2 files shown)


maobaolong:
@runzhiwang Thank you for this improvement, it make code more clear and easy to understand.

sodonnel:
Good catch @runzhiwang 

It seems to me that the iterator fails to work correctly if there is a filter which filters out any of the blocks. The first time a block is filtered out, even though there are recursive calls, it will still return false at the top level and the iterator will stop, even if there are more blocks available.

I was surprised a unit test did not catch this, so I checked TestKeyValueBlockIterator#testKeyValueBlockIteratorWithFilter and it fails to catch this:

```
  @Test
  public void testKeyValueBlockIteratorWithFilter() throws Exception {
    long containerId = 103L;
    int deletedBlocks = 5;
    int normalBlocks = 5;
    createContainerWithBlocks(containerId, normalBlocks, deletedBlocks);
    String containerPath = new File(containerData.getMetadataPath())
        .getParent();
    try(KeyValueBlockIterator keyValueBlockIterator = new KeyValueBlockIterator(
        containerId, new File(containerPath), MetadataKeyFilters
        .getDeletingKeyFilter())) {

      int counter = 5;
      while (keyValueBlockIterator.hasNext()) {
        BlockData blockData = keyValueBlockIterator.nextBlock();
        assertEquals(blockData.getLocalID(), counter++);
      }
    }
  }
```
The reason it fails to catch it is because it does not check `count` is incremented from 5 to 10, as it expects to get the 5 deleting blocks. If we add an assertEquals(10, count) at the end of the test, it would catch this issue. However there is another problem in the test, as the createContainerWithBlocks method never created the "deleting" blocks.

With the following change to the test I can reproduce the issue and your change fixes it:

```
diff --git a/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/keyvalue/TestKeyValueBlockIterator.java b/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/keyvalue/TestKeyValueBlockIterator.java
index ccd3227fa..1f1423264 100644
--- a/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/keyvalue/TestKeyValueBlockIterator.java
+++ b/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/keyvalue/TestKeyValueBlockIterator.java
@@ -223,6 +223,7 @@ public void testKeyValueBlockIteratorWithFilter() throws Exception {
         BlockData blockData = keyValueBlockIterator.nextBlock();
         assertEquals(blockData.getLocalID(), counter++);
       }
+      assertEquals(10, counter);
     }
   }
 
@@ -276,8 +277,8 @@ private void createContainerWithBlocks(long containerId, int
             .getProtoBufMessage().toByteArray());
       }
 
-      for (int i = normalBlocks; i < deletedBlocks; i++) {
-        BlockID blockID = new BlockID(containerId, i);
+      for (int i = 0; i < deletedBlocks; i++) {
+        BlockID blockID = new BlockID(containerId, i + normalBlocks);
         BlockData blockData = new BlockData(blockID);
         blockData.setChunks(chunkList);
         metadataStore.getStore().put(StringUtils.string2Bytes(OzoneConsts
```

Could you make that change to the tests and then we are good to commit this change. Thanks.

I was also wondering - how did you discover this? Did you discover a problem at runtime on your cluster which caused you to investigate this?

runzhiwang:
@sodonnel Thanks for review and suggestions. I have updated the patch. 

> I was also wondering - how did you discover this? Did you discover a problem at runtime on your cluster which caused you to investigate this?

I discover this when I read the code, because my recent work is related to this.

sodonnel:
Thanks for addressing the unit test and for finding this issue. This change LGTM. I will go ahead and commit it.

PR #1115, opened 2020-06-23
HDDS-3632. starter scripts can't manage Ozone and HDFS datandodes on the same machine 

elek:
…
since the service names are same and they both referring to same location for pid files, we can not start both services at once.

Tweak is to export HADOOP_PID_DIR to different location after starting one service and start other one.

It would be better to have different pid file names.

```
Umas-MacBook-Pro ozone-0.5.0-beta % bin/ozone --daemon start datanode
datanode is running as process 25167.  Stop it first.
```

https://issues.apache.org/jira/browse/HDDS-3632

## How was this patch tested?

 1. Started a new cluster with `compose/ozonescripts`
 2. Fixed the configuration (will be uploaded in a [separated jira](https://issues.apache.org/jira/browse/HDDS-3857))
 3. Used `./start.sh`
 4. Checked if cluster works (ozone sh + volume, bucket, key create)
 5. Checked if the /tmp/*.pid contains ozone name on the datanodes.

### DIFF
--- a/hadoop-ozone/dist/src/shell/hdds/hadoop-functions.sh
+++ b/hadoop-ozone/dist/src/shell/hdds/hadoop-functions.sh
@@ -2702,11 +2702,11 @@ function hadoop_generic_java_subcmd_handler
     priv_outfile="${HADOOP_LOG_DIR}/privileged-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.out"
     priv_errfile="${HADOOP_LOG_DIR}/privileged-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.err"
     priv_pidfile="${HADOOP_PID_DIR}/privileged-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}.pid"
-    daemon_outfile="${HADOOP_LOG_DIR}/hadoop-${HADOOP_SECURE_USER}-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.out"
-    daemon_pidfile="${HADOOP_PID_DIR}/hadoop-${HADOOP_SECURE_USER}-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}.pid"
+    daemon_outfile="${HADOOP_LOG_DIR}/ozone-${HADOOP_SECURE_USER}-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.out"
+    daemon_pidfile="${HADOOP_PID_DIR}/ozone-${HADOOP_SECURE_USER}-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}.pid"
   else
-    daemon_outfile="${HADOOP_LOG_DIR}/hadoop-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.out"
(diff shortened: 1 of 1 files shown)


hadoop-ozone/dist/src/shell/hdds/hadoop-functions.sh
```diff
@@ -2702,11 +2702,11 @@ function hadoop_generic_java_subcmd_handler
     priv_outfile="${HADOOP_LOG_DIR}/privileged-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.out"
     priv_errfile="${HADOOP_LOG_DIR}/privileged-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.err"
     priv_pidfile="${HADOOP_PID_DIR}/privileged-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}.pid"
```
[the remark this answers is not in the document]
arp7: Same comment as @adoroszlai. Should we change the filenames for secure cluster also?

hadoop-ozone/dist/src/shell/hdds/hadoop-functions.sh
```diff
@@ -2702,11 +2702,11 @@ function hadoop_generic_java_subcmd_handler
...
+    daemon_outfile="${HADOOP_LOG_DIR}/ozone-${HADOOP_SECURE_USER}-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.out"
+    daemon_pidfile="${HADOOP_PID_DIR}/ozone-${HADOOP_SECURE_USER}-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}.pid"
   else
-    daemon_outfile="${HADOOP_LOG_DIR}/hadoop-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.out"
-    daemon_pidfile="${HADOOP_PID_DIR}/hadoop-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}.pid"
+    daemon_outfile="${HADOOP_LOG_DIR}/ozone-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}-${HOSTNAME}.out"
+    daemon_pidfile="${HADOOP_PID_DIR}/ozone-${HADOOP_IDENT_STRING}-${HADOOP_SUBCMD}.pid"
   fi
 
   # are we actually in daemon mode?
```
umamaheswararao, line 2712: in below mode also: do we have any scenario where user sets their log DIR to a common folder for all services?
If yes, below case also may need update?

hadoop-ozone/dist/src/shell/hdds/hadoop-functions.sh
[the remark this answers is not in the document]
elek: Good point, thanks. Fixed.

hadoop-ozone/dist/src/shell/hdds/hadoop-functions.sh
elek: Hard to say. I guess it's not a common setup, but who knows, and it should be safer to change, anyway. 

Fixed, thanks the suggestion.

umamaheswararao: Thanks @elek for addressing my comment.

elek:
Merging. Thanks the review @adoroszlai @arp7 and @umamaheswararao (and the report @umamaheswararao)

PR #1117, opened 2020-06-23
HDDS-3858. Remove support to start Ozone and HDFS datanodes in the same JVM

elek:

With a few thousands issues ago Ozone was integral part of Hadoop/HDFS project. At that time there were two options to start datanode:

 1. Start in a separated JVM

 2. Start in the same JVM with the HDFS

Today only 1 is the standard way, this is tested and working. 2nd is not working any more but still documented.

I propose to drop the support of this use case as I can't see any benefit to support it anymore:

 1. I think 100% of the users will use Ozone as a separated process not as a HDFS Datanode plugin
 2. Fixing the classpath issues is significant effort as the classpath of HDFS and Ozone are diverged.

https://issues.apache.org/jira/browse/HDDS-3858

## How was this patch tested?

Not required. One compose folder + docs are removed.


elek:
Ping everybody (especially @jnp @arp7 @xiaoyuyao @mukul1987 @nandakumar131 )

dineshchitlangia:
This is a double edged sword :) 
Some of my customers that I speak to have indicated they would not directly go to Ozone in an independent fashion and a few of them have indicated they want to explore Ozone by itself.
How about we leave it as is and mention a cautionary statement to indicate that Ozone HDFS Datanode Plugin is now an experimental feature and it not recommended from Production use.
This will allow some users to get their feet wet with ozone in a limited way.

jnp:
The option to run them in same JVM was allowed, to support in-place upgrade from HDFS to Ozone in future. It was assumed that having the two Datanodes in the same JVM will certainly simplify Hdfs to Ozone upgrade.
Some design work happened in past, but not much progress since then. However, that is still a valid goal, IMO.
Once users are comfortable to switch to Ozone completely, an in-place upgrade option will be desirable to avoid data copy, and will be a way faster option to move to Ozone.

xiaoyuyao:
@elek thanks for bringing this up. Given the fact that it is broken currently, it makes sense to remove the document saying it is supported. Let's open a separate JIRA to track the issues such as classpath, etc. that prevent running HDDS datanode as plugin inside HDFS datanode JVM. They need to be fixed when we revisit in-place upgrade.

elek:
> The option to run them in same JVM was allowed, to support in-place upgrade from HDFS to Ozone in future

@jnp Base on the last design in-place upgrade can work with different JVM on the same host. And it's possible to run HDFS and Ozone datanodes on the same node.

My problem is that maintaining this feature requires significant effort. If we would like to do it, it's a release blocker:

```
2020-06-23 15:22:52 WARN  DataNode:983 - ServicePlugin org.apache.hadoop.ozone.HddsDatanodeService@5226e402 could not be started
java.lang.NoSuchMethodError: com.google.common.base.Preconditions.checkArgument(ZLjava/lang/String;Ljava/lang/Object;)V
```

elek:
/pending "Blocking merge ability as we definitely need an agreement first. Possible topic for Community Meeting."

jnp:
> Given the fact that it is broken currently, it makes sense to remove the document saying it is supported.

I agree.

arp7:
+1 for documenting that this is not supported.

maobaolong:
+1, it will make Ozone repo clearly.

elek:
> +1 for documenting that this is not supported.

Yes, it's important that I wouldn't like to modify interfaces or code. Just remove visibility (remove docs + compose examples).

We can make it work any time, or we can provide specific plugin for in-place upgrade.

It seems that we have an agreement that we can hide this feature, and we can further discuss long-term strategies **before any code change**.

As we have the consensus, I will merge this PR, soon.

PR #1119, opened 2020-06-24
HDDS-3263. Fix TestCloseContainerByPipeline.java.

bshashikant:
Enabled the test as it seems to be stable with latest ratis upgrade.

https://issues.apache.org/jira/browse/HDDS-3263

## How was this patch tested?
Manually verified.


adoroszlai:
I think it was a bit too early to merge this after a single run.  `TestCloseContainerByPipeline` failed on [master](https://github.com/apache/hadoop-ozone/runs/804094128) and in [#1120](https://github.com/apache/hadoop-ozone/pull/1120/checks?check_run_id=804217101) since then.

PR #1122, opened 2020-06-24
HDDS-3861. Fix handlePipelineFailure throw exception if role is follower

runzhiwang:
1. if role is follower, ozone will throw exception.
image

https://issues.apache.org/jira/browse/HDDS-3861

## How was this patch tested?

Existed tests

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
@@ -553,7 +553,11 @@ public final class XceiverServerRatis implements XceiverServerSpi {
           roleInfoProto.getCandidateInfo().getLastLeaderElapsedTimeMs() + "ms";
       break;
     case FOLLOWER:
-      msg = datanode + " is in follower state for " +
+      msg = datanode + " closes pipeline when installSnapshot from leader " +
+          "because leader snapshot doesn't contain any data to replay, " +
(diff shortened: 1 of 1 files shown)


runzhiwang:
@bshashikant Could you help review this patch ? Thank you very much.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
```diff
@@ -552,6 +552,10 @@ private void handlePipelineFailure(RaftGroupId groupId,
       msg = datanode + " is in candidate state for " +
           roleInfoProto.getCandidateInfo().getLastLeaderElapsedTimeMs() + "ms";
       break;
+    case FOLLOWER:
+      msg = datanode + " is in follower state for " +
```
bshashikant, line 556: Can we update the msg accordingly with the reason. The reason precisely was when installSnapshotFromLeader gets invoked on follower, as leader snapshot doesn't contain any data to replay. All the log entries prior to the snapshot might have been purged. So , follower should not try to install snapshot from leader but can close the pipeline here.

runzhiwang: @bshashikant Thanks for you review and explanation. I have updated the patch.

runzhiwang:
@adoroszlai  Could you help merge it ?  The comment is just about the message of reason, I have updated it, I think it does not matter. But if without this PR, the cluster is very unstable.

PR #1125, opened 2020-06-24
HDDS-3866. Support multi-part-upload with Freon S3 key generator

elek:

ozone freon `s3kg` is a key generator which uses the s3 interface. It uses simple put objects but it turned out that s3 mpu implementation has some specific problem. I would improve the key generator to make it possible to test mpu upload with AWS Java SDK.

https://issues.apache.org/jira/browse/HDDS-3866

## How was this patch tested?

````
ozone freon s3kg -n 1 -s 6000000 --multi-part-upload
```

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/S3KeyGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/S3KeyGenerator.java
@@ -16,6 +16,9 @@
  */
 package org.apache.hadoop.ozone.freon;
 
+import java.io.ByteArrayInputStream;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.Callable;
 
 import org.apache.hadoop.hdds.cli.HddsVersionProvider;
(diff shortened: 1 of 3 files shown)


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/S3KeyGenerator.java
```diff
@@ -54,7 +64,8 @@
   private String bucketName;
 
   @Option(names = {"-s", "--size"},
-      description = "Size of the generated key (in bytes)",
+      description = "Size of the generated key (in bytes) or size of one "
+          + "multipart upload part (in case of multipart upload)",
```
bharatviswa504, line 68: Minor: Do you think we need some check on the size, when multipart upload option is specified, as in MPU each part should be at least 5MB(except last part) But here we use same part size having that check would be useful, otherwise we shall get an error during complete MPU afteer uploading 10 parts.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/S3KeyGenerator.java
```diff
@@ -81,7 +103,8 @@ public Void call() throws Exception {
     if (endpoint.length() > 0) {
       amazonS3ClientBuilder
           .withPathStyleAccessEnabled(true)
-          .withEndpointConfiguration(new EndpointConfiguration(endpoint, ""));
+          .withEndpointConfiguration(
+              new EndpointConfiguration(endpoint, "us-east-1"));
```
bharatviswa504, line 107: Any reason for this change, setting to us-east-1?

elek: Just one reason: got an exception ;-) The AWS 4 signature handling failed in our code with having empty region. I tried to define it with environment variable but it was not enough.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/S3KeyGenerator.java
elek: The backend will fail anyway thanks to the implementation there...  (guess, you were more thoroughgoing ;-) )

But yes, it's slightly more polite to fail early (and creates less garbage). Added a simple check.

bharatviswa504:
Thank You @elek for the contribution and @adoroszlai for the review.

PR #1128, opened 2020-06-25
HDDS-3862. Prepare checks for running some tests multiple times

adoroszlai:

1. Improve `unit.sh` and `integration.sh` to allow running tests multiple times after a single compilation.
   * Number of runs is controlled by `ITERATIONS` environment variable (defaults to 1).
   * Failed test output from each iteration is saved into a separate directory (`iterationN`), but only if repetitions are requested, ie. no change for regular, single run.
   * Each run is followed by a one-liner summary, eg.: `Iteration 1 exit code: 0`
2. Reduce duplication by extracting common logic into separate `junit.sh` script.

https://issues.apache.org/jira/browse/HDDS-3862

## How was this patch tested?

Regular CI runs:
* [unit](https://github.com/adoroszlai/hadoop-ozone/runs/806245518)
* eg. [integration (client)](https://github.com/adoroszlai/hadoop-ozone/runs/806264131)

Repeated tests:
* [TestCloseContainerByPipeline](https://github.com/adoroszlai/hadoop-ozone/runs/806313923), failed 5/20
* [TestCSMMetrics](https://github.com/adoroszlai/hadoop-ozone/runs/804181186), passed 20/20

### DIFF
--- a/hadoop-ozone/dev-support/checks/junit.sh
+++ b/hadoop-ozone/dev-support/checks/junit.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -u -o pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "$DIR/../../.." || exit 1
+
+: ${CHECK:="unit"}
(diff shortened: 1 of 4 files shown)


hadoop-ozone/dev-support/checks/junit.sh
```diff
@@ -0,0 +1,63 @@
...
+  mvn ${MAVEN_OPTIONS} -fae "$@" test \
+    | tee "${REPORT_DIR}/output.log"
+  irc=$?
+
+  # shellcheck source=hadoop-ozone/dev-support/checks/_mvn_unit_report.sh
+  source "${DIR}/_mvn_unit_report.sh"
+
+  if [[ ${ITERATIONS} -gt 1 ]]; then
+    REPORT_DIR="${original_report_dir}"
+    echo "Iteration ${i} exit code: ${irc}" | tee -a "${REPORT_DIR}/output.log"
```
elek, line 49:
```suggestion
    echo "Iteration ${i} exit code: ${irc}" | tee -a "${REPORT_DIR}/summary.txt"
```

Based on the existing convention, output.log contains the full std output, summary.txt the information.

adoroszlai: I've updated the patch based on this suggestion.  Now that I see the results of a repeated run, a question occurred to me: should we print only failed iterations (to match how we only list failed tests in summary) or all of them?

elek: When we had the Kubernetes based build / notification logic, it worked based on the number of lines in the summary.txt. Today we have no such dependencies. I am fine with keeping it simple in this form (include all lines).

adoroszlai:
> +1 let's merge after a green build

Build is finally green, merging it.  Thanks @elek for the review.

PR #1129, opened 2020-06-25
HDDS-3741. Reload old OM state if Install Snapshot from Leader fails

hanishakoneru:

Follower OM issues a pause on its services before installing new checkpoint from Leader OM (Install Snapshot). If this installation fails for some reason, the OM stays in paused state. It should be unpaused and the old state should be reloaded.

https://issues.apache.org/jira/browse/HDDS-3741

## How was this patch tested?

Unit tests added.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
@@ -18,7 +18,6 @@ package org.apache.hadoop.ozone.om;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -256,7 +255,7 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
       // marker indicates that the DB is in an inconsistent state and hence
       // the OM process should be terminated.
       File markerFile = new File(metaDir, DB_TRANSIENT_MARKER);
-      if (Files.exists(markerFile.toPath())) {
+      if (markerFile.exists()) {
         LOG.error("File {} marks that OM DB is in an inconsistent state.");
         // Note - The marker file should be deleted only after fixing the DB.
         // In an HA setup, this can be done by replacing this DB with a
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
@@ -290,10 +290,16 @@ public class OmMetadataManagerImpl implements OMMetadataManager {
 
   public static DBStore loadDB(OzoneConfiguration configuration, File metaDir)
       throws IOException {
+    return loadDB(configuration, metaDir, OM_DB_NAME);
+  }
+
+  public static DBStore loadDB(OzoneConfiguration configuration, File metaDir,
+      String dbName)
+      throws IOException {
     RocksDBConfiguration rocksDBConfiguration =
         configuration.getObject(RocksDBConfiguration.class);
     DBStoreBuilder dbStoreBuilder = DBStoreBuilder.newBuilder(configuration,
-        rocksDBConfiguration).setName(OM_DB_NAME)
+        rocksDBConfiguration).setName(dbName)
         .setPath(Paths.get(metaDir.getPath()));
     DBStore dbStore = addOMTablesAndCodecs(dbStoreBuilder).build();
     return dbStore;
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
@@ -18,6 +18,7 @@ package org.apache.hadoop.ozone.om;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -77,9 +78,11 @@ import com.google.common.collect.Lists;
 import org.apache.commons.lang3.StringUtils;
 import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_OPEN_KEY_EXPIRE_THRESHOLD_SECONDS;
 import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_OPEN_KEY_EXPIRE_THRESHOLD_SECONDS_DEFAULT;
+import static org.apache.hadoop.ozone.OzoneConsts.DB_TRANSIENT_MARKER;
 import static org.apache.hadoop.ozone.OzoneConsts.OM_DB_NAME;
 import static org.apache.hadoop.ozone.OzoneConsts.OM_KEY_PREFIX;
 
+import org.apache.ratis.util.ExitUtils;
 import org.eclipse.jetty.util.StringUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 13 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2999,32 +3000,15 @@ public TermIndex installSnapshot(String leaderId) {
...
-    }
-
-    try {
-      omTransactionInfo =
-          OzoneManagerRatisUtils.getTransactionInfoFromDownloadedSnapshot(
-              configuration, dbDir);
-    } catch (Exception ex) {
-      LOG.error("Failed during opening downloaded snapshot from " +
-          "{} to obtain transaction index", newDBlocation, ex);
+    if (omTransactionInfo == null) {
```
arp7, line 3011: Let's log an error here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3040,45 +3024,61 @@ public TermIndex installSnapshot(String leaderId) {
...
-      LOG.error("OM DB checkpoint replacement with new downloaded checkpoint " +
-          "failed.", e);
-      return null;
-    }
-
     // Reload the OM DB store with the new checkpoint.
     // Restart (unpause) the state machine and update its last applied index
     // to the installed checkpoint's snapshot index.
     try {
       reloadOMState(leaderIndex, leaderTerm);
```
arp7, line 3076: Before calling reloadState check for existence of temp_marker file. Also on OM restart.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3040,45 +3024,61 @@ public TermIndex installSnapshot(String leaderId) {
...
+          omTransactionInfo, lastAppliedIndex, leaderId, newDBLocation);
+      if (!canProceed) {
+        return null;
+      }
 
-    boolean canProceed =
-        OzoneManagerRatisUtils.verifyTransactionInfo(omTransactionInfo,
-        lastAppliedIndex, leaderId, newDBlocation);
+      try {
+        dbBackup = replaceOMDBWithCheckpoint(lastAppliedIndex, oldDBLocation,
```
arp7, line 3053: `replaceOMDBWithCheckpoint` contract should be that it either leaves the current state or the old state. If it leaves inconsistent state then it should also leave some persistent marker on disk so OM doesn't try to restart with a bad checkpoint.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3040,45 +3024,61 @@ public TermIndex installSnapshot(String leaderId) {
...
+            "checkpoint failed.", e);
+        return null;
+      }
 
-    // If downloaded DB has transaction info less than current one, return.
-    if (!canProceed) {
-      return null;
+      loadSuccess = true;
+    } finally {
+      if (!loadSuccess) {
```
arp7, line 3063: Remove this unpause, we already do a reload and unpause below.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3040,45 +3024,61 @@ public TermIndex installSnapshot(String leaderId) {
...
+        omRatisServer.getOmStateMachine().unpause(lastAppliedIndex,
+            currentTerm);
+      }
     }
 
     long leaderIndex = omTransactionInfo.getTransactionIndex();
```
arp7, line 3069: Move these inside the try block, just after you invoke `replaceOMDBWithCheckpoint`. Else these variables should be left to values that reflect the old checkpoint.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3040,45 +3024,61 @@ public TermIndex installSnapshot(String leaderId) {
...
+      } catch (Exception e) {
+        LOG.error("OM DB checkpoint replacement with new downloaded " +
+            "checkpoint failed.", e);
+        return null;
+      }
 
-    // If downloaded DB has transaction info less than current one, return.
-    if (!canProceed) {
-      return null;
+      loadSuccess = true;
```
arp7, line 3061: Don't need `loadSuccess` any more.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
arp7: Also the marker file should be created before starting the move operations, and deleted on success.

hanishakoneru:
Thanks @arp7 for the review. Addressed your review comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3314,4 +3352,8 @@ private void startJVMPauseMonitor() {
     jvmPauseMonitor.init(configuration);
     jvmPauseMonitor.start();
   }
+
+  void setExitManagerForTesting(ExitManager exitManagerForTesting) {
```
arp7, line 3356: Can you add a `@VisibleForTesting` annotation here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
```diff
@@ -249,6 +252,20 @@ public void start(OzoneConfiguration configuration) throws IOException {
...
+      // the OM process should be terminated.
+      File markerFile = new File(metaDir, DB_TRANSIENT_MARKER);
+      if (Files.exists(markerFile.toPath())) {
+        LOG.error("File {} marks that OM DB is in an inconsistent state.");
+        // Note - The marker file should be deleted only after fixing the DB.
+        // In an HA setup, this can be done by replacing this DB with a
+        // checkpoint from another OM.
+        String errorMsg = "Cannot load OM DB as it is in an inconsistent " +
+            "state.";
+        ExitUtils.terminate(1, errorMsg, LOG);
```
arp7, line 266: Should we use `ExitManager` here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
```diff
@@ -249,6 +252,20 @@ public void start(OzoneConfiguration configuration) throws IOException {
     if (store == null) {
       File metaDir = OMStorage.getOmDbDir(configuration);
 
+      // Check if there is a DB Inconsistent Marker in the metaDir. This
+      // marker indicates that the DB is in an inconsistent state and hence
+      // the OM process should be terminated.
+      File markerFile = new File(metaDir, DB_TRANSIENT_MARKER);
+      if (Files.exists(markerFile.toPath())) {
```
arp7, line 259: Nitpick: why not just `markerFile.exists()`?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/utils/OzoneManagerRatisUtils.java
```diff
@@ -228,15 +228,34 @@ public static Status exceptionToResponseStatus(IOException exception) {
...
   /**
-   * Obtain Transaction info from downloaded snapshot DB.
+   * Obtain OMTransactionInfo from Checkpoint.
+   */
+  public static OMTransactionInfo getTrxnInfoFromCheckpoint(
+      OzoneConfiguration conf, Path dbPath) throws Exception {
+    Path dbDir = dbPath.getParent();
+    String dbName = dbPath.getFileName().toString();
+    if (dbDir == null) {
+      throw new IOException("Checkpoint {} does not have proper DB location");
```
arp7, line 238: Print the `dbPath` in the exception message to make debugging easier in case this happens.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/snapshot/OzoneManagerSnapshotProvider.java
```diff
@@ -113,8 +113,10 @@ public OzoneManagerSnapshotProvider(MutableConfigurationSource conf,
   public DBCheckpoint getOzoneManagerDBSnapshot(String leaderOMNodeID)
       throws IOException {
     String snapshotTime = Long.toString(System.currentTimeMillis());
-    String snapshotFileName = Paths.get(omSnapshotDir.getAbsolutePath(),
-        snapshotTime, OM_DB_NAME).toFile().getAbsolutePath();
+    String snapshotFileName = OM_DB_NAME + "-" + leaderOMNodeID
```
arp7, line 116: Did we change the snapshot file name convention in this jira?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3041,58 +3047,74 @@ public TermIndex installSnapshot(String leaderId) {
...
+      String errorMsg = "Failed to reload OM state and instantiate services.";
+      exitManager.exitSystem(1, errorMsg, ex, LOG);
     }
 
     // Delete the backup DB
     try {
-      FileUtils.deleteFully(dbBackup);
+      if (dbBackup != null) {
+        FileUtils.deleteFully(dbBackup);
+      }
```
arp7, line 3104: On successful restart we should clear any old backups.

arp7: Let's do this in a followup jira. Ok to commit this one.

PR #1130, opened 2020-06-25
HDDS-3871. Add resource core-site during loading of ozoneconfiguration.

bharatviswa504:

Add core-site resources during the loading of ozone configuration.

This is done because in few scenarios/environments spark executor pods(when spark on k8s, does not have any ozone-site.xml in the classpath) To help in environments where ozone-site.xml will not be available and where we add ozone properties to core-site, this PR will help.

https://issues.apache.org/jira/browse/HDDS-3871

## How was this patch tested?

Tested it on the secure cluster where ozone properties are added to core-site and ran spark job wordcount which reads from o3fs. Verified that it is successfully run.


bharatviswa504:
Thank You @avijayanhwx for the review.

PR #1132, opened 2020-06-26
HDDS-3872. Reduce output of coverage.sh

adoroszlai:

1. Suppress output of download progress
2. Only build top-level project (previously all submodules were `SKIPPED` by the plugin)

https://issues.apache.org/jira/browse/HDDS-3872

## How was this patch tested?

Output of maven-dependency-plugin copy command is now only [28 lines](https://github.com/adoroszlai/hadoop-ozone/runs/811182393#step:5:6) vs. [6K+ lines earlier](https://github.com/apache/hadoop-ozone/runs/811002033#step:5:6).


adoroszlai:
Thanks for the review @HorizonNet and @vivekratnavel.

PR #1133, opened 2020-06-26
HDDS-3875. Package classpath files to the jar files instead of uploading them as artifacts

elek:

During the earlier release we realized that Apache Nexus couldn't handle the classpath files well.

Classpath files are generated during the build and copied to the final distribution to use separated classpath definition for each of the services.

It turned out that the Apache Nexus couldn't handle it well (INFRA-18344) and HDDS-1510 was an attempt to fix this problem. But during 0.5.0 release we have seen the same problem.

To make the 0.6.0 release more seamless I propose to update the logic and instead of copy the classpath file via an artifact I would pack/unpack them to/from the final artifact jar.

https://issues.apache.org/jira/browse/HDDS-3875

## How was this patch tested?

It's a dist packaging change. CI test is failed which means it works as before:

https://github.com/elek/hadoop-ozone/actions/runs/148630517

### DIFF
--- a/hadoop-hdds/pom.xml
+++ b/hadoop-hdds/pom.xml
@@ -284,7 +284,7 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
               <goal>build-classpath</goal>
             </goals>
             <configuration>
-              <outputFile>${project.build.outputDirectory}/${pom.artifactId}.classpath</outputFile>
+              <outputFile>${project.build.outputDirectory}/${project.artifactId}.classpath</outputFile>
               <prefix>$HDDS_LIB_JARS_DIR</prefix>
               <outputFilterFile>true</outputFilterFile>
               <includeScope>runtime</includeScope>
--- a/hadoop-hdds/pom.xml
+++ b/hadoop-hdds/pom.xml
@@ -279,37 +279,15 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
         <executions>
           <execution>
             <id>add-classpath-descriptor</id>
-            <phase>package</phase>
+            <phase>prepare-package</phase>
             <goals>
               <goal>build-classpath</goal>
             </goals>
             <configuration>
-              <outputFile>${project.build.directory}/classpath</outputFile>
+              <outputFile>${project.build.outputDirectory}/${pom.artifactId}.classpath</outputFile>
               <prefix>$HDDS_LIB_JARS_DIR</prefix>
               <outputFilterFile>true</outputFilterFile>
               <includeScope>runtime</includeScope>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>build-helper-maven-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>attach-classpath-artifact</id>
-            <phase>package</phase>
-            <goals>
-              <goal>attach-artifact</goal>
-            </goals>
-            <configuration>
-              <artifacts>
-                <artifact>
-                  <file>${project.build.directory}/classpath</file>
-                  <type>cp</type>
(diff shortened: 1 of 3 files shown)


hadoop-hdds/pom.xml
```diff
@@ -279,41 +279,19 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
...
           <execution>
             <id>add-classpath-descriptor</id>
-            <phase>package</phase>
+            <phase>prepare-package</phase>
             <goals>
               <goal>build-classpath</goal>
             </goals>
             <configuration>
-              <outputFile>${project.build.directory}/classpath</outputFile>
+              <outputFile>${project.build.outputDirectory}/${pom.artifactId}.classpath</outputFile>
```
adoroszlai, line 287: Maven outputs a warning for each submodule:

```
The expression ${pom.artifactId} is deprecated. Please use ${project.artifactId} instead.
```

```suggestion
              <outputFile>${project.build.outputDirectory}/${project.artifactId}.classpath</outputFile>
```

hadoop-ozone/dist/pom.xml
```diff
@@ -41,103 +41,13 @@
...
-                <artifactItem>
-                  <groupId>org.apache.hadoop</groupId>
-                  <artifactId>hadoop-ozone-insight</artifactId>
-                  <version>${ozone.version}</version>
-                  <classifier>classpath</classifier>
-                  <type>cp</type>
-                  <destFileName>hadoop-ozone-insight.classpath</destFileName>
-                </artifactItem>
-              </artifactItems>
+              <includes>*.classpath</includes>
```
adoroszlai, line 50: Fix me if I'm wrong, but `ozone classpath` now lists several modules that were previously omitted.

Previously:

```
$ ozone classpath
Usage: ozone classpath <ARTIFACTNAME>
Where the artifact name is one of:

hadoop-hdds-server-scm
hadoop-hdds-tools
hadoop-ozone-common
hadoop-ozone-csi
hadoop-ozone-datanode
hadoop-ozone-filesystem
hadoop-ozone-insight
hadoop-ozone-ozone-manager
hadoop-ozone-recon
hadoop-ozone-s3gateway
hadoop-ozone-tools
hadoop-ozone-upgrade
```

With patch:

```
$ ozone classpath
Usage: ozone classpath <ARTIFACTNAME>
Where the artifact name is one of:

hadoop-hdds-client
hadoop-hdds-common
hadoop-hdds-config
hadoop-hdds-container-service
hadoop-hdds-docs
hadoop-hdds-hadoop-dependency-client
hadoop-hdds-hadoop-dependency-server
hadoop-hdds-interface-admin
hadoop-hdds-interface-client
hadoop-hdds-interface-server
hadoop-hdds-server-framework
hadoop-hdds-server-scm
hadoop-hdds-tools
hadoop-ozone-client
hadoop-ozone-common
hadoop-ozone-csi
hadoop-ozone-datanode
hadoop-ozone-filesystem-common
hadoop-ozone-filesystem
hadoop-ozone-insight
hadoop-ozone-interface-client
hadoop-ozone-ozone-manager
hadoop-ozone-recon
hadoop-ozone-reconcodegen
hadoop-ozone-s3gateway
hadoop-ozone-tools
hadoop-ozone-upgrade
```

Invoking `ozone classpath` for the extra modules results in an error (which is the same on `master`):

```
$ ozone classpath hadoop-hdds-config
Error: Could not find or load main class org.apache.hadoop.util.Classpath
Caused by: java.lang.ClassNotFoundException: org.apache.hadoop.util.Classpath
```

elek: Nice catch. I did a `grep` for `OZONE_RUN_ARTIFACT_NAME` in `ozone` shell script and selected only those artifacts.

elek:
Merging it. Hopefully we will have fewer problems with the 0.6.0. Thanks the review @adoroszlai

PR #1134, opened 2020-06-26
HDDS-3868. Implement getTrashRoot and getTrashRoots in o3fs

smengcl:

Override the default Hadoop FS impl.

Currently o3fs moves trash under:
```
o3fs://bucketName.volumeName.om/user/userName/.Trash/Current/...
```

New trash location will be:
```
o3fs://bucketName.volumeName.om/.Trash/userName/Current/...
```

https://issues.apache.org/jira/browse/HDDS-3868

## How was this patch tested?

Added a simple test `testGetTrashRoot`.

### DIFF
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -616,7 +616,8 @@ public class BasicOzoneFileSystem extends FileSystem {
   @Override
   public Path getTrashRoot(Path path) {
     final Path pathToTrash = new Path(OZONE_URI_DELIMITER, TRASH_PREFIX);
-    return new Path(pathToTrash, System.getProperty("user.name"));
+    return new Path(pathToTrash, getUsername());
+  }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -606,6 +607,52 @@ public String getUsername() {
...
+   *         Returns .Trash of users if {@code /.Trash/$USER} exists.
+   */
+  @Override
+  public Collection<FileStatus> getTrashRoots(boolean allUsers) {
+    Path trashRoot = new Path(OZONE_URI_DELIMITER, TRASH_PREFIX);
+    List<FileStatus> ret = new ArrayList<>();
+    try {
+      if (!allUsers) {
+        Path userTrash = new Path(trashRoot, userName);
+        if (exists(userTrash)) {
```
xiaoyuyao, line 637: should we check both existing and isDirectory before return?

smengcl: Done.

Interestingly Hadoop's [default implementation](https://github.com/apache/hadoop/blob/263c76b678275dfff867415c71ba9dc00a9235ef/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L3262-L3266) doesn't check directories. But it would be nice to have.

smengcl:
Thanks @xiaoyuyao  for the review. Will commit shortly.

PR #1139, opened 2020-06-28
HDDS-3885. Create Datanode home page

maobaolong:

We need a Datanode home page. After this, we can do more effort to see the starttime version, build info and more information about hddsdatanode service.

HDDS-3885

## How was this patch tested?

start datanode and see the root page.

http://localhost:9882/

image


maobaolong:
@adoroszlai Please take a look at this PR, thanks.

maobaolong:
@adoroszlai Thanks for your review. AFAIK, the reason why Rpc Metrics page is empty for datanode is that datanode has no rpc service.PTAL.

adoroszlai:
Thanks @maobaolong for updating the patch.  Let's wait for others to take a look.

maobaolong:
@adoroszlai  Thank you for your review.

maobaolong:
@xiaoyuyao Please take a look at this PR, if we take this PR into 0.6.0, it would be better, thanks.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/HddsDatanodeService.java
```diff
@@ -22,13 +22,11 @@
 import java.net.InetAddress;
 import java.security.KeyPair;
 import java.security.cert.CertificateException;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
ChenSammi, line 25: Import * is not recommended.

maobaolong: Done.

maobaolong:
@ChenSammi Thank you for correct me, PTAL.

ChenSammi:
Tested it locally.  +1.  Thanks @maobaolong for the contribution.

PR #1145, opened 2020-06-29
HDDS-3895. Implement container related operations in ContainerManagerImpl

nandakumar131:
SCM HA uses the new ContainerManagerImpl. This change introduces container related operations in ContainerManagerImpl.

HDDS-3895

## How was this patch tested?
Unit test will be added as part of HDDS-3896

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
@@ -23,12 +23,14 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.stream.Collectors;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ContainerInfoProto;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.LifeCycleState;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.LifeCycleEvent;
@@ -37,7 +39,9 @@ import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationType;
 import org.apache.hadoop.hdds.scm.ha.SCMHAManager;
 import org.apache.hadoop.hdds.scm.pipeline.Pipeline;
 import org.apache.hadoop.hdds.scm.pipeline.PipelineManager;
+import org.apache.hadoop.hdds.utils.UniqueId;
 import org.apache.hadoop.hdds.utils.db.Table;
+import org.apache.hadoop.ozone.common.statemachine.InvalidStateTransitionException;
 import org.apache.hadoop.util.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -60,6 +64,7 @@ public class ContainerManagerImpl implements ContainerManagerV2 {
   /**
    *
    */
+  //Can we move this lock to ContainerStateManager?
   private final ReadWriteLock lock;
 
   /**
@@ -93,67 +98,13 @@ public class ContainerManagerImpl implements ContainerManagerV2 {
   }
 
   @Override
-  public Set<ContainerID> getContainerIDs() {
-    lock.readLock().lock();
-    try {
-      return containerStateManager.getContainerIDs();
-    } finally {
-      lock.readLock().unlock();
-    }
-  }
-
-  @Override
-  public Set<ContainerInfo> getContainers() {
-    lock.readLock().lock();
-    try {
-      return containerStateManager.getContainerIDs().stream().map(id -> {
-        try {
-          return containerStateManager.getContainer(id);
-        } catch (ContainerNotFoundException e) {
-          // How can this happen? o_O
-          return null;
-        }
-      }).filter(Objects::nonNull).collect(Collectors.toSet());
-    } finally {
-      lock.readLock().unlock();
-    }
-  }
-
-  @Override
-  public ContainerInfo getContainer(final ContainerID containerID)
+  public ContainerInfo getContainer(final ContainerID id)
       throws ContainerNotFoundException {
     lock.readLock().lock();
     try {
-      return containerStateManager.getContainer(containerID);
-    } finally {
-      lock.readLock().unlock();
-    }
-  }
-
-  @Override
-  public Set<ContainerInfo> getContainers(final LifeCycleState state) {
-    lock.readLock().lock();
-    try {
-      return containerStateManager.getContainerIDs(state).stream().map(id -> {
-        try {
-          return containerStateManager.getContainer(id);
-        } catch (ContainerNotFoundException e) {
-          // How can this happen? o_O
-          return null;
-        }
-      }).filter(Objects::nonNull).collect(Collectors.toSet());
-    } finally {
-      lock.readLock().unlock();
-    }
-  }
-
-  @Override
-  public boolean exists(final ContainerID containerID) {
-    lock.readLock().lock();
-    try {
-      return (containerStateManager.getContainer(containerID) != null);
-    } catch (ContainerNotFoundException ex) {
-      return false;
+      return Optional.ofNullable(containerStateManager
+          .getContainer(id.getProtobuf()))
+          .orElseThrow(() -> new ContainerNotFoundException("ID " + id));
     } finally {
       lock.readLock().unlock();
     }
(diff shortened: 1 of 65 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
```diff
@@ -279,6 +257,24 @@ public void notifyContainerReportProcessing(final boolean isFullReport,
     throw new UnsupportedOperationException("Not yet implemented!");
   }
 
+  @Override
+  public void deleteContainer(final ContainerID containerID)
+      throws IOException {
+    final HddsProtos.ContainerID id = containerID.getProtobuf();
```
xiaoyuyao, line 263: Need to hold the writeLock?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
```diff
@@ -279,6 +257,24 @@ public void notifyContainerReportProcessing(final boolean isFullReport,
...
+    if (!containerStateManager.contains(id)) {
```
xiaoyuyao, line 264: Can we use checkIfContainerExist()?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
nandakumar131: Thanks for the review. Missed this somehow 😄

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
```diff
@@ -93,94 +98,45 @@ public ContainerManagerImpl(
...
-  @Override
-  public boolean exists(final ContainerID containerID) {
+  public List<ContainerInfo> listContainers(final ContainerID startID,
+                                            final int count) {
     lock.readLock().lock();
     try {
-      return (containerStateManager.getContainer(containerID) != null);
-    } catch (ContainerNotFoundException ex) {
-      return false;
+      final long startId = startID == null ? 0 : startID.getId();
```
xiaoyuyao, line 118: NIT: can we use a different name startId and startID is confusing.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerV2.java
```diff
@@ -113,7 +90,7 @@ void deleteContainer(ContainerID containerID)
    */
   void updateContainerState(ContainerID containerID,
                             LifeCycleEvent event)
-      throws ContainerNotFoundException;
+      throws IOException, InvalidStateTransitionException;
```
xiaoyuyao, line 93: NIT: javadoc update @throws

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
```diff
@@ -60,6 +64,7 @@
   /**
    *
    */
+  //Can we move this lock to ContainerStateManager?
```
xiaoyuyao, line 67: Same question here. I see we have a RWLock in ContainerMangerImpl and one RWLock in ContainerStateManager. Can we rely only on the lock in ContainerStateManager?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerV2.java
```diff
@@ -130,6 +141,21 @@ ContainerInfo getContainer(ContainerID containerID)
   void addContainer(ContainerInfoProto containerInfo)
       throws IOException;
 
+  /**
+   *
+   */
+  @Replicate
+  void updateContainerState(HddsProtos.ContainerID id,
```
xiaoyuyao, line 148: Can we add unit test if any the updateContainerState with @Replicate?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
nandakumar131: I would also like to move the lock to ContainerStateManager and not have any lock in ContainerMangerImpl. I'm still thinking of corner cases where this might cause issues.
State change calls to ContainerStateManager are made via Ratis, so we should not check some state in ContainerMangerImpl and make a Ratis call to change the state without holding any lock.

I will remove the lock in followup jira after making sure that we are not missing any corner cases.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerV2.java
nandakumar131: I'm planning to add detailed javadoc in follow up jiras. Will fix the @throws here.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerV2.java
nandakumar131: Sure.

nandakumar131:
Integration test failure is not related to this change. Created HDDS-4125 to fix the bug which is causing test failure.

timmylicheng:
@nandakumar131 Ratis One pipeline is not instantly removed in StaleNodeHandler. We will come up with a fix.

nandakumar131:
It is not just about one node pipeline, we have the same issue in three node pipeline as well. If all the three nodes goes stale (because of restart), we will mark the pipeline as closed and will not remove.

The problem is not just there is no instant removal, it is never deleted.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerImpl.java
```diff
@@ -291,23 +328,14 @@ ContainerInfo getMatchingContainer(final long size, String owner,
...
-      throws ContainerNotFoundException, ContainerReplicaNotFoundException {
-    throw new UnsupportedOperationException("Not yet implemented!");
-  }
-
-
-  void removeContainer(final ContainerID containerID)
-      throws ContainerNotFoundException {
-    throw new UnsupportedOperationException("Not yet implemented!");
+  public void removeContainer(final HddsProtos.ContainerID id) {
+    containers.removeContainer(ContainerID.getFromProtobuf(id));
```
GlenGeng, line 338: Show we update `containerStore` as well ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerImpl.java
```diff
@@ -254,32 +260,63 @@ public void addContainer(final ContainerInfoProto containerInfo)
...
 
-  void updateContainerState(final ContainerID containerID,
-                            final LifeCycleEvent event)
-      throws IOException {
-    throw new UnsupportedOperationException("Not yet implemented!");
+  public void updateContainerState(final HddsProtos.ContainerID containerID,
+                                   final LifeCycleEvent event)
+      throws IOException, InvalidStateTransitionException {
+    // TODO: Remove the protobuf conversion after fixing ContainerStateMap.
+    final ContainerID id = ContainerID.getFromProtobuf(containerID);
```
GlenGeng, line 286: Show we update containerStore as well ?

GlenGeng:
It seems `ContainerManager` is not replace by `ContainerManagerV2` in `StorageContainerManager`.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
timmylicheng: Agree with @xiaoyuyao 's suggestion.
@nandakumar131 Can we also track it in a separate Jira?

timmylicheng:
> It seems `ContainerManager` is not replaced by `ContainerManagerV2` in `StorageContainerManager` yet.

@nandakumar131 @GlenGeng Create https://issues.apache.org/jira/browse/HDDS-4132 to track switching to ContainerManagerV2.

@nandakumar131 Could you create other follow-up tasks?

nandakumar131:
Thanks @GlenGeng for the review. I will address them as part of HDDS-4133.

timmylicheng:
+1 overall. 

Comments on this PR will be addressed in following Jiras. HDDS-4132 and HDDS-4133.

Thanks @nandakumar131 for contribution. Merging...

PR #1146, opened 2020-06-29
HDDS-2978. Intermittent failure in TestResourceLimitCache

cku328:

Calling `CompletableFuture.cancel()` will mark the future as cancelled,but does not interrupt the running task.

The Javadoc for the method [CompletableFuture#cancel()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#cancel-boolean-) states:
```
Parameters:
mayInterruptIfRunning - this value has no effect in this implementation because interrupts are not used to control processing.
```

In this PR, I'm trying to make asynchronous tasks run in a specified thread pool by given the `executor` when using the `supplyAsync(...)` method.

Finally, use `Executor.shutdownNow()` to interrupt the operation of putting key (_put<4,a>_).

https://issues.apache.org/jira/browse/HDDS-2978

## How was this patch tested?

Manual unit test 17000+ times -> Tests passed.

image

### DIFF
--- a/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/utils/TestResourceLimitCache.java
+++ b/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/utils/TestResourceLimitCache.java
@@ -20,7 +20,11 @@ import org.apache.hadoop.test.GenericTestUtils;
 import org.junit.Assert;
 import org.junit.Test;
 
-import java.util.concurrent.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 /**
  * Test for ResourceLimitCache.
@@ -71,7 +75,7 @@ public class TestResourceLimitCache {
       } catch (InterruptedException e) {
         return null;
       }
-    });
+    },pool);
     Assert.assertTrue(!future.isDone());
     Thread.sleep(100);
     Assert.assertTrue(!future.isDone());
--- a/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/utils/TestResourceLimitCache.java
+++ b/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/utils/TestResourceLimitCache.java
@@ -75,7 +75,7 @@ public class TestResourceLimitCache {
       } catch (InterruptedException e) {
         return null;
       }
-    },pool);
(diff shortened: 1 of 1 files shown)


hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/utils/TestResourceLimitCache.java
```diff
@@ -60,25 +58,27 @@ public void testResourceLimitCache()
...
+    ExecutorService pool = Executors.newCachedThreadPool();
     future = CompletableFuture.supplyAsync(() -> {
       try {
         return resourceCache.put(4, "a");
       } catch (InterruptedException e) {
-        e.printStackTrace();
+        return null;
       }
-      return null;
     });
```
adoroszlai, line 74: `pool` is currently unused, I think you intended to pass it to `supplyAsync`:

```suggestion
    }, pool);
```

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/utils/TestResourceLimitCache.java
```diff
@@ -20,9 +20,7 @@
 import org.junit.Assert;
 import org.junit.Test;
 
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.*;
```
adoroszlai, line 23: Nit: please avoid `*` import.

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/utils/TestResourceLimitCache.java
cku328: Thanks @adoroszlai for the notification.
After testing, I want to re-apply this modification in a new branch, and I missed it.

cku328:
Thanks @adoroszlai  for the review.
I updated my patch, based on the review comments.

I retested it 1500 times -> passed.

image

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/utils/TestResourceLimitCache.java
cku328: Thanks! I'll avoid using it.

adoroszlai:
Thanks @cku328 for the fix and @lokeshj1703 for the review.

cku328:
Thanks @adoroszlai  @lokeshj1703  for the review.

PR #1147, opened 2020-06-29
HDDS-3892. Datanode initialization is too slow when there are thousan…

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-3892

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -104,9 +104,9 @@ public final class OzoneConfigKeys {
   public static final String OZONE_CONTAINER_CACHE_SIZE =
       "ozone.container.cache.size";
   public static final int OZONE_CONTAINER_CACHE_DEFAULT = 1024;
-  public static final String OZONE_CONTAINER_CACHE_LOCK_STRIPS =
-      "ozone.container.cache.lock.strips";
-  public static final int OZONE_CONTAINER_CACHE_LOCK_STRIPS_DEFAULT = 1024;
+  public static final String OZONE_CONTAINER_CACHE_LOCK_STRIPES =
+      "ozone.container.cache.lock.stripes";
+  public static final int OZONE_CONTAINER_CACHE_LOCK_STRIPES_DEFAULT = 1024;
 
   public static final String OZONE_SCM_BLOCK_SIZE =
       "ozone.scm.block.size";
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -104,6 +104,9 @@ public final class OzoneConfigKeys {
   public static final String OZONE_CONTAINER_CACHE_SIZE =
       "ozone.container.cache.size";
   public static final int OZONE_CONTAINER_CACHE_DEFAULT = 1024;
+  public static final String OZONE_CONTAINER_CACHE_LOCK_STRIPS =
+      "ozone.container.cache.lock.strips";
+  public static final int OZONE_CONTAINER_CACHE_LOCK_STRIPS_DEFAULT = 1024;
 
   public static final String OZONE_SCM_BLOCK_SIZE =
       "ozone.scm.block.size";
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import com.google.common.util.concurrent.Striped;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.utils.MetadataStore;
 import org.apache.hadoop.hdds.utils.MetadataStoreBuilder;
@@ -43,6 +44,7 @@ public final class ContainerCache extends LRUMap {
   private final Lock lock = new ReentrantLock();
   private static ContainerCache cache;
   private static final float LOAD_FACTOR = 0.75f;
+  private final static Striped<Lock> rocksDBLock = Striped.lazyWeakLock(1024);
   /**
    * Constructs a cache that holds DBHandle references.
    */
@@ -118,50 +120,56 @@ public final class ContainerCache extends LRUMap {
     Preconditions.checkState(containerID >= 0,
         "Container ID cannot be negative.");
     ReferenceCountedDB db;
-    lock.lock();
+    Lock containerLock = rocksDBLock.get(containerDBPath);
+    containerLock.lock();
     try {
-      db = (ReferenceCountedDB) this.get(containerDBPath);
-      if (db != null) {
-        db.incrementReference();
-        return db;
+      lock.lock();
+      try {
+        db = (ReferenceCountedDB) this.get(containerDBPath);
+        if (db != null) {
+          db.incrementReference();
+          return db;
+        }
+      } finally {
+        lock.unlock();
+      }
+
+      try {
+        MetadataStore metadataStore =
+            MetadataStoreBuilder.newBuilder()
+                .setDbFile(new File(containerDBPath))
+                .setCreateIfMissing(false)
+                .setConf(conf)
+                .setDBType(containerDBType)
+                .build();
+        db = new ReferenceCountedDB(metadataStore, containerDBPath);
+      } catch (Exception e) {
+        LOG.error("Error opening DB. Container:{} ContainerPath:{}",
+            containerID, containerDBPath, e);
+        throw e;
+      }
+
+      lock.lock();
+      try {
+        ReferenceCountedDB currentDB =
+            (ReferenceCountedDB) this.get(containerDBPath);
+        if (currentDB != null) {
+          // increment the reference before returning the object
+          currentDB.incrementReference();
+          // clean the db created in previous step
+          db.cleanup();
+          return currentDB;
+        } else {
+          this.put(containerDBPath, db);
+          // increment the reference before returning the object
+          db.incrementReference();
+          return db;
+        }
+      } finally {
+        lock.unlock();
       }
     } finally {
-      lock.unlock();
-    }
-
-    try {
-      MetadataStore metadataStore =
-          MetadataStoreBuilder.newBuilder()
-              .setDbFile(new File(containerDBPath))
-              .setCreateIfMissing(false)
-              .setConf(conf)
-              .setDBType(containerDBType)
-              .build();
-      db = new ReferenceCountedDB(metadataStore, containerDBPath);
-    } catch (Exception e) {
-      LOG.error("Error opening DB. Container:{} ContainerPath:{}",
-          containerID, containerDBPath, e);
-      throw e;
-    }
-
-    lock.lock();
-    try {
-      ReferenceCountedDB currentDB =
-          (ReferenceCountedDB) this.get(containerDBPath);
-      if (currentDB != null) {
-        // increment the reference before returning the object
-        currentDB.incrementReference();
-        // clean the db created in previous step
-        db.cleanup();
-        return currentDB;
-      } else {
-        this.put(containerDBPath, db);
-        // increment the reference before returning the object
-        db.incrementReference();
-        return db;
-      }
-    } finally {
-      lock.unlock();
+      containerLock.unlock();
     }
   }
 
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
@@ -48,10 +48,10 @@ public final class ContainerCache extends LRUMap {
   /**
    * Constructs a cache that holds DBHandle references.
    */
-  private ContainerCache(int maxSize, int strips, float loadFactor, boolean
+  private ContainerCache(int maxSize, int stripes, float loadFactor, boolean
       scanUntilRemovable) {
     super(maxSize, loadFactor, scanUntilRemovable);
-    rocksDBLock = Striped.lazyWeakLock(strips);
+    rocksDBLock = Striped.lazyWeakLock(stripes);
   }
 
   /**
@@ -66,10 +66,10 @@ public final class ContainerCache extends LRUMap {
     if (cache == null) {
       int cacheSize = conf.getInt(OzoneConfigKeys.OZONE_CONTAINER_CACHE_SIZE,
           OzoneConfigKeys.OZONE_CONTAINER_CACHE_DEFAULT);
-      int strips = conf.getInt(
-          OzoneConfigKeys.OZONE_CONTAINER_CACHE_LOCK_STRIPS,
-          OzoneConfigKeys.OZONE_CONTAINER_CACHE_LOCK_STRIPS_DEFAULT);
-      cache = new ContainerCache(cacheSize, strips, LOAD_FACTOR, true);
+      int stripes = conf.getInt(
+          OzoneConfigKeys.OZONE_CONTAINER_CACHE_LOCK_STRIPES,
+          OzoneConfigKeys.OZONE_CONTAINER_CACHE_LOCK_STRIPES_DEFAULT);
+      cache = new ContainerCache(cacheSize, stripes, LOAD_FACTOR, true);
     }
     return cache;
   }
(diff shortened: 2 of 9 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
```diff
@@ -117,28 +117,47 @@ public ReferenceCountedDB getDB(long containerID, String containerDBType,
...
+    }
+
+    lock.lock();
+    try {
+      ReferenceCountedDB currentDB =
+          (ReferenceCountedDB) this.get(containerDBPath);
+      if (currentDB != null) {
+        // increment the reference before returning the object
+        currentDB.incrementReference();
+        return currentDB;
```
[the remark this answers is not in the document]
ChenSammi: Thanks for the review. You're right.  The new DB instance should be explicitely closed in this case.

bharatviswa504:
Hi @ChenSammi 
Thanks for the PR.
I too have the same comment as @sodonnel. I think we need to close the DB instance if we get an instance from the cache.

And also do you have this version on your deployed cluster
https://github.com/apache/hadoop-ozone/commit/a2ab8d6e35f60af9762a191265942071755329be

This saves in iterating Db to compute block metadata, and get all the required metadata from db.get(). We have observed this issue during a billion object test.
For older containers created before HDDS-3217, we still iterate and set block metadata. (But this fixed a bug, where we used to iterate thrice for open containers)

ChenSammi:
@bharatviswa504  we have HDDS-3217 deployed.

ChenSammi:
> 
> 
> I see one issue with this approach.
> If the database is already opened, and if we try to open again we will get this error.
> 
> I think, with this change, we will throw an exception if we try to open the database again an already existing one.
> 
> java.io.IOException: Failed init RocksDB, db path : /Users/bviswanadham/workspace/hadoop-ozone/hadoop-hdds/container-service/target/test-dir/xCkBnsLVrc/cont1, exception :org.rocksdb.RocksDBException lock : /Users/bviswanadham/workspace/hadoop-ozone/hadoop-hdds/container-service/target/test-dir/xCkBnsLVrc/cont1/LOCK: No locks available

@bharatviswa504 , I get your point. It's an issue, but not introduced by this patch.  It's a currenlty existing issue and we need to carefully think about how to fix it with a new JIRA.

ChenSammi:
The failed integration test is irrevelant.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
```diff
@@ -117,28 +117,49 @@ public ReferenceCountedDB getDB(long containerID, String containerDBType,
...
+    }
+
+    try {
+      MetadataStore metadataStore =
+          MetadataStoreBuilder.newBuilder()
+              .setDbFile(new File(containerDBPath))
+              .setCreateIfMissing(false)
+              .setConf(conf)
+              .setDBType(containerDBType)
+              .build();
```
adoroszlai, line 139: > > I see one issue with this approach.
> > If the database is already opened, and if we try to open again we will get this error.
> > I think, with this change, we will throw an exception if we try to open the database again an already existing one.
> > java.io.IOException: Failed init RocksDB, db path : /Users/bviswanadham/workspace/hadoop-ozone/hadoop-hdds/container-service/target/test-dir/xCkBnsLVrc/cont1, exception :org.rocksdb.RocksDBException lock : /Users/bviswanadham/workspace/hadoop-ozone/hadoop-hdds/container-service/target/test-dir/xCkBnsLVrc/cont1/LOCK: No locks available
> 
> @bharatviswa504 , I get your point. It's an issue, but not introduced by this patch. It's a currenlty existing issue and we need to carefully think about how to fix it with a new JIRA.

@ChenSammi can you please clarify why you think it's an existing issue?  It seems to me that calling `build()` outside of the lock introduces this problem.

If I understand correctly, the performance issue is due to using a single lock for all containers.  Instead of moving the expensive part outside of the lock, how about using a separate lock per `containerDBPath`?

ChenSammi: @adoroszlai, OK, I get the point now.  During the datanode initialization phase, actually there is no change a DB will be opened concurrently.  But since this function is also used in late phase, there is possibility to trigger the exception.

ChenSammi:
Thanks @bharatviswa504  and @adoroszlai for the review. lock per containerDBPath in ContainerCache is a good idea. I tried it but cannot solve our problem given a volume has nearly 20K containers, it still take nearly an hour for the datanode to finish the initialization.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
```diff
@@ -113,22 +113,41 @@ protected boolean removeLRU(LinkEntry entry) {
...
-            .setConf(conf)
-            .setDBType(containerDBType)
-            .build();
+                .setDbFile(new File(containerDBPath))
+                .setCreateIfMissing(false)
+                .setConf(conf)
+                .setDBType(containerDBType)
+                .build();
         db = new ReferenceCountedDB(metadataStore, containerDBPath);
         this.put(containerDBPath, db);
```
sodonnel, line 152: If I understand correctly, you have created a boolean to either lock or not lock in this method. In the DN initialization logic, you pass false here to avoid locking, as the same container will not be initialized by multiple threads at once.

I am still not sure this is safe, specifically the line:

```
this.put(containerDBPath, db);
```

ContainerCache is an instance of LRUCache and the JavaDoc says:

> Note that LRUMap is not synchronized and is not thread-safe.</strong>
> If you wish to use this map from multiple threads concurrently, you must use
> appropriate synchronization.

Therefore I think you must lock around access and writes to the LRU Cache.

I also worry about removing locking completely incase someone passes "false" mistakenly to this method in the future.

@adoroszlai Earlier suggested using a lock per container path. We might be able to use a Guava Striped lock here with something like 256 buckets.

Could we do something like:

```
  public ReferenceCountedDB getDB(long containerID, String containerDBType,
                             String containerDBPath, ConfigurationSource conf)
      throws IOException {
    Preconditions.checkState(containerID >= 0,
        "Container ID cannot be negative.");
    ReferenceCountedDB db = null;
    Lock slock = stripedLock.get(containerDBPath);
    slock.lock();
    try {
      lock.lock();
      try {
        db = (ReferenceCountedDB) this.get(containerDBPath);
      } finally {
        lock.unlock();
      }
      if (db == null) {
        MetadataStore metadataStore =
            MetadataStoreBuilder.newBuilder()
                .setDbFile(new File(containerDBPath))
                .setCreateIfMissing(false)
                .setConf(conf)
                .setDBType(containerDBType)
                .build();
        db = new ReferenceCountedDB(metadataStore, containerDBPath);
        addDB(containerDBPath, db);
      }
      db.incrementReference();
      return db;
    } catch (Exception e) {
      LOG.error("Error opening DB. Container:{} ContainerPath:{}",
          containerID, containerDBPath, e);
      throw e;
    } finally {
      slock.unlock();
    }
  }
```

We still have the global lock to protect the ContainerCache structure, but we now have a stripped lock to ensure that the same ContainerPath cannot be initialized at the same time.

I am not sure if we need to worry about another thread calling addDB - if that is the case, we would need to add back the logic you had before, where you take the global lock again, and then test to see if an entry for the DB has been added to the cache, and if so close the new instance.

sodonnel:
> Thanks @bharatviswa504 and @adoroszlai for the review. lock per containerDBPath in ContainerCache is a good idea. I tried it but cannot solve our problem given a volume has nearly 20K containers, it still take nearly an hour for the datanode to finish the initialization.

With a lock per containerPath, where does the bottle neck come from? Is it around the lock, or does a single thread load all the containers for a volume, and therefore its the time taken to init all the DB instances that is the bottleneck?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
sodonnel: Thinking about this some more, it might be safest to wrap addDB and removeDB with the strippedLock too, as that way we ensure only one DB instance can be initialized and added to the map at any given time.

@adoroszlai might have some better idea than me on this too?

ChenSammi:
> > Thanks @bharatviswa504 and @adoroszlai for the review. lock per containerDBPath in ContainerCache is a good idea. I tried it but cannot solve our problem given a volume has nearly 20K containers, it still take nearly an hour for the datanode to finish the initialization.
> 
> With a lock per containerPath, where does the bottle neck come from? Is it around the lock, or does a single thread load all the containers for a volume, and therefore its the time taken to init all the DB instances that is the bottleneck?

The DB open takes time. The LOCK makes the DB open action one by one. Once it has more than 10K rocksdb in a volume, the initialization takes a considerable time, say more than 30minutes.   
There are several severe consequence of datanode slow initialization,
1. service unavailable
2. scm will treat datanode as DEAD(We have raised the DEAD threshold to 30m in our cluster), and close all the related pipeline and containers related with this datanode, which contributed to why there are so many small containers in our cluster. It's a negative cycle.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
ChenSammi: Thanks @sodonnel , I will take a look at the StrippedLock.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
```diff
@@ -117,30 +119,57 @@ public ReferenceCountedDB getDB(long containerID, String containerDBType,
...
+
+      lock.lock();
+      try {
+        ReferenceCountedDB currentDB =
+            (ReferenceCountedDB) this.get(containerDBPath);
+        if (currentDB != null) {
+          // increment the reference before returning the object
+          currentDB.incrementReference();
+          // clean the db created in previous step
+          db.cleanup();
```
sodonnel, line 160: This cleanup() call may be expensive, and ideally it should be outside the lock. However, I think it will be very rare where you hit this scenario and therefore I think the logic is OK as it is now.

ChenSammi: OK.

ChenSammi:
> 
> 
> These changes with the striped lock LGTM, +1.
> 
> One question - have you been able to test this change on your DNs with a large number of containers, and did it give a good improvement?

Yes, I have tested it on a datanode with 20K+ containers, it costs about 30m to finish the ContainerSet build.  Before the patch,  time spending is about 10 times longer.  If we ignore the lock during ContainerSet build,  it costs about 15m.

sodonnel:
> > These changes with the striped lock LGTM, +1.
> > One question - have you been able to test this change on your DNs with a large number of containers, and did it give a good improvement?
> 
> Thanks @sodonnel .
> Yes, I have tested it on a datanode with 20K+ containers, it costs about 30m to finish the ContainerSet build. Before the patch, time spending is about 10 times longer. If we ignore the lock during ContainerSet build, it costs about 15m.

This seems like a good improvement for a relatively simply change. I think removing the lock completely is risky and may cause some other problems, either now or later.

@bharatviswa504 @adoroszlai  - have you guys any further comments or concerns with the latest patch?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
```diff
@@ -43,6 +44,7 @@
   private final Lock lock = new ReentrantLock();
   private static ContainerCache cache;
   private static final float LOAD_FACTOR = 0.75f;
+  private final Striped<Lock> rocksDBLock = Striped.lazyWeakLock(1024);
```
adoroszlai, line 47: Would be nice to make the number of stripes configurable (later).

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestContainerCache.java
```diff
@@ -63,6 +70,8 @@ public void testContainerCacheEviction() throws Exception {
     conf.setInt(OzoneConfigKeys.OZONE_CONTAINER_CACHE_SIZE, 2);
 
     ContainerCache cache = ContainerCache.getInstance(conf);
+    cache.clear();
+    Assert.assertTrue(cache.size() == 0);
```
adoroszlai, line 74:
```suggestion
    Assert.assertEquals(0, cache.size());
```

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestContainerCache.java
```diff
@@ -123,4 +132,47 @@ public void testContainerCacheEviction() throws Exception {
...
+  public void testConcurrentDBGet() throws Exception {
+    File root = new File(testRoot);
+    root.mkdirs();
+    root.deleteOnExit();
+
+    OzoneConfiguration conf = new OzoneConfiguration();
+    conf.setInt(OzoneConfigKeys.OZONE_CONTAINER_CACHE_SIZE, 2);
+    ContainerCache cache = ContainerCache.getInstance(conf);
+    cache.clear();
+    Assert.assertTrue(cache.size() == 0);
```
adoroszlai, line 146:
```suggestion
    Assert.assertEquals(0, cache.size());
```

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestContainerCache.java
```diff
@@ -123,4 +132,47 @@ public void testContainerCacheEviction() throws Exception {
...
+        Assert.fail("Should get the DB instance");
+      }
+    }
+
+    ReferenceCountedDB db = cache.getDB(1, "RocksDB",
+        containerDir.getPath(), conf);
+    db.close();
+    db.close();
+    db.close();
+    Assert.assertTrue(cache.size() == 1);
```
adoroszlai, line 175:
```suggestion
    Assert.assertEquals(1, cache.size());
```

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestContainerCache.java
```diff
@@ -123,4 +132,47 @@ public void testContainerCacheEviction() throws Exception {
...
+    cache.clear();
+    Assert.assertTrue(cache.size() == 0);
+    File containerDir = new File(root, "cont1");
+    createContainerDB(conf, containerDir);
+    ExecutorService executorService = Executors.newFixedThreadPool(2);
+    Runnable task = () -> {
+      try {
+        ReferenceCountedDB db1 = cache.getDB(1, "RocksDB",
+            containerDir.getPath(), conf);
+        Assert.assertTrue(db1 != null);
```
adoroszlai, line 154:
```suggestion
        Assert.assertNotNull(db1);
```

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/ozoneimpl/TestContainerReader.java
```diff
@@ -219,4 +220,68 @@ public void testContainerReader() throws Exception {
...
+    for (int i = 0; i < volumeNum; i++) {
+      threads[i].start();
+    }
+    for (int i = 0; i < volumeNum; i++) {
+      threads[i].join();
+    }
+    System.out.println("Open " + volumeNum + " Volume with " + containerCount +
+        " costs " + (System.currentTimeMillis() - startTime) / 1000 + "s");
+    Assert.assertTrue(
+        containerSet.getContainerMap().entrySet().size() == containerCount);
```
adoroszlai, line 284:
```suggestion
    Assert.assertEquals(containerCount,
        containerSet.getContainerMap().entrySet().size());
```

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/ozoneimpl/TestContainerReader.java
```diff
@@ -219,4 +220,68 @@ public void testContainerReader() throws Exception {
...
+    Assert.assertTrue(cache.size() == containerCount);
```
adoroszlai, line 285:
```suggestion
    Assert.assertEquals(containerCount, cache.size());
```

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
```diff
@@ -104,6 +104,9 @@
   public static final String OZONE_CONTAINER_CACHE_SIZE =
       "ozone.container.cache.size";
   public static final int OZONE_CONTAINER_CACHE_DEFAULT = 1024;
+  public static final String OZONE_CONTAINER_CACHE_LOCK_STRIPS =
+      "ozone.container.cache.lock.strips";
+  public static final int OZONE_CONTAINER_CACHE_LOCK_STRIPS_DEFAULT = 1024;
```
adoroszlai, line 109: Note minor typo: `strips` -> `stripes`, but I'm fine with both.

ChenSammi:
Thanks @bharatviswa504 @sodonnel and @adoroszlai for the code review and great suggestions.

PR #1148, opened 2020-06-29
HDDS-3831. Enhance Recon ContainerEndPoint to report on different unhealty container states

sodonnel:

This change provides two new endpoints on the Recon "containers" endpoint:

/containers/unhealthy
/containers/unhealthy/{state}

The intention is for these endpoints to replace /containers/missing, but that end point is still in place as we need to change the front end of recon to use this new endpoint first.

Calling either of the above new endpoints will return a summary of all unhealthy containers in the response, eg:

```
  @JsonProperty("missingCount")
  private long missingCount = 0;

  @JsonProperty("underReplicatedCount")
  private long underReplicatedCount = 0;

  @JsonProperty("overReplicatedCount")
  private long overReplicatedCount = 0;

  @JsonProperty("misReplicatedCount")
  private long misReplicatedCount = 0;
```

Followed by either:

 * A list of all unhealthy containers, sorted by `container_id asc , state asc` for the /containers/unhealthy endpoint
 * A list of unhealthy containers in the passed state, sorted by `container_id asc , state asc` for the /containers/unhealthy/{state} endpoint

Note that even when /containers/unhealthy/{state} is passed, the full summary of all containers is returned - it would be possible to only give a summary of the passed state, but I think it makes sense to give the full summary on all calls and just filter the list based on the passed parameter.

For both endpoints, there is a parameter "limit" to control how many containers to return (default 1000) and "batchNum" starting a 1 to get each batch / page of results.

The reason it does not use the previous container_id to start the next batch, is because there may be more than 1 database entry per container, and therefore we would need to pass preContainerId and preState. However it would be possible to change to use this if it is desired.

https://issues.apache.org/jira/browse/HDDS-3831

## How was this patch tested?

New unit tests

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestReconTasks.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestReconTasks.java
@@ -37,6 +37,7 @@ import org.apache.hadoop.ozone.MiniOzoneCluster;
 import org.apache.hadoop.ozone.recon.scm.ReconContainerManager;
 import org.apache.hadoop.ozone.recon.scm.ReconStorageContainerManagerFacade;
 import org.apache.hadoop.test.LambdaTestUtils;
+import org.hadoop.ozone.recon.schema.ContainerSchemaDefinition;
 import org.hadoop.ozone.recon.schema.tables.pojos.UnhealthyContainers;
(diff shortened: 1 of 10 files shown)


vivekratnavel:
@sodonnel For some reason, there are no CI checks for the commit. Can you please try to trigger CI checks by pushing an empty commit? Thanks!

PR #1150, opened 2020-06-30
HDDS-3903. OzoneRpcClient support batch rename keys.

captainzmc:

Currently rename folder is to get all the keys, and then rename them one by one. This makes for poor performance.

HDDS-2939 can able to optimize this part, but at present the HDDS-2939 is slow and still a long way to go. So we optimized the batch operation based on the current interface. We were able to get better performance with this PR before the HDDS-2939 came in.

This PR is a subtask of Batch Rename and first makes OzoneRpcClient Support Batch Rename Keys.

https://issues.apache.org/jira/browse/HDDS-3903

## How was this patch tested?

UT has been added.


captainzmc:
Hi @xiaoyuyao，This PR implementation is consistent with HDDS-3286 batchDelete. And I split batchRename into two subtasks. This one, mainly OM side implementation. Could you help review it?

captainzmc:
Hi @xiaoyuyao，This PR implementation is consistent with HDDS-3286 batchDelete. And I split batchRename into two subtasks. This one, mainly OM side implementation. Could you help review it?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
```diff
@@ -216,6 +217,13 @@ OmKeyLocationInfo allocateBlock(OmKeyArgs args, long clientID,
    */
   void renameKey(OmKeyArgs args, String toKeyName) throws IOException;
 
+  /**
+   * Rename existing keys within a bucket.
+   * @param keyMap The key is new key name nad value is original key OmKeyArgs.
+   * @throws IOException
+   */
+  void renameKeys(Map<String, OmKeyArgs> keyMap) throws IOException;
```
xiaoyuyao, line 225: Should we make thie Map<OmKeyArgs, OmKeyArgs> so that the batch of the rename does not to have the same volume/bucket name, e.g. /vol1/bucket1/key1->/vol1/bucket2/key2 and /vol2/bucket1/key1->/vol2/bucket1/key2?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -672,6 +676,32 @@ public OmKeyInfo lookupKey(OmKeyArgs args) throws IOException {
...
+    List<RenameKeyRequest> renameKeyRequestList  = new ArrayList<>();
+    for (Map.Entry< String, OmKeyArgs > entry : keyMap.entrySet()) {
+      RenameKeyRequest.Builder reqKey = RenameKeyRequest.newBuilder();
+      OmKeyArgs args = entry.getValue();
+      KeyArgs keyArgs = KeyArgs.newBuilder()
+          .setVolumeName(args.getVolumeName())
+          .setBucketName(args.getBucketName())
+          .setKeyName(args.getKeyName())
+          .setModificationTime(Time.now())
+          .setDataSize(args.getDataSize()).build();
```
xiaoyuyao, line 691: Do we need to update the datasize here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmRenameKeyInfo.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.hadoop.ozone.om.helpers.OmKeyInfo;
+
+import javax.annotation.Nonnull;
+
+/**
+ * This class is used for rename keys.
+ */
+public class OmRenameKeyInfo {
+
+  private String toKeyName;
```
xiaoyuyao, line 30: newKeyInfo.keyName already cover toKeyName?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  @Override
+  public OMRequest preExecute(OzoneManager ozoneManager) throws IOException {
+
+    RenameKeysRequest renameKeysRequest = getOmRequest().getRenameKeysRequest();
+    Preconditions.checkNotNull(renameKeysRequest);
+
+    return getOmRequest().toBuilder()
+        .setRenameKeysRequest(renameKeysRequest.toBuilder())
```
xiaoyuyao, line 86: I think we need to set the modification time here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,289 @@
...
+          .getRenameKeyRequestList()) {
+        OzoneManagerProtocolProtos.KeyArgs renameKeyArgs =
+            renameKeyRequest.getKeyArgs();
+        volumeName = renameKeyArgs.getVolumeName();
+        bucketName = renameKeyArgs.getBucketName();
+        fromKeyName = renameKeyArgs.getKeyName();
+        String objectKey = omMetadataManager.getOzoneKey(volumeName, bucketName,
+            fromKeyName);
+        OmKeyInfo omKeyInfo = omMetadataManager.getKeyTable().get(objectKey);
+        unRenamedKeys.add(omKeyInfo);
```
xiaoyuyao, line 134: unRenamedKeys is not updated inside the loop. Can we add a unit test for the returned unRenamedKeys upon failure?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
captainzmc: I'll make this to Map<OmKeyArgs, String> to consistent with the renameKey() API.

In addition, the API was added to OzoneBucket and currently only supports renaming keys under the same bucket. 
`OzoneBucket bucket = volume.getBucket(bucketName);`
`Map<String, String> keyMap = new HashMap();`
`keyMap.put(keyName1, newKeyName1);`
`bucket.renameKeys(keyMap);`

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
captainzmc: Although we have put the list of unDeletedKeys and unRenamedKeys into the response. Currently, renameKeys and DeleteKeys in OzoneBucket.java and ClientProtocol.java are still void. 
I've added TODO to the OMClientRequest and created a jira(HDDS-3916), and I'll change those separately.

captainzmc:
Thanks @xiaoyuyao for the review. I have fixed review issues. Can you take another look?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,298 @@
...
+          createRenameKeysErrorOMResponse(omResponse, exception,
+              unRenamedKeys));
+    } finally {
+      addResponseToDoubleBuffer(trxnLogIndex, omClientResponse,
+          omDoubleBufferHelper);
+    }
+
+    if (result == Result.SUCCESS || result == Result.FAILURE) {
+      auditLog(auditLogger, buildAuditMessage(OMAction.RENAME_KEY, auditMap,
+          exception, getOmRequest().getUserInfo()));
```
adoroszlai, line 258: Shouldn't all renamed keys be logged to audit?

captainzmc: Thanks @adoroszlai  for the review. Fixed this issue.

adoroszlai: Thanks, but the new code only logs key names, other info eg. volume/bucket is missing.  Also, if result is failure, then all keys are logged as "failure", although only the last key failed (and we don't even know which one is last, since `auditMap` is a hashmap).

captainzmc: Thanks adoroszlai  for the tip，I'll add a volume/bucket to the key. 
Also, on how to locate the failed key. We can see this in the Exception in auditLog. Typically, the failed keys are displayed in the log with an Exception, such as:
`17:45:30.550 [OM StateMachine ApplyTransaction Thread - 0] ERROR OMAudit - user=micahzhao | ip=127.0.0.1 | op=RENAME_KEY {dir/key2=dir/file2, dir/file1=dir/file2} | ret=FAILURE`
`org.apache.hadoop.ozone.om.exceptions.OMException: Key not found /a88fb570-5b5d-43db-81e0-d6597f4ea81f/4ad1e6c3-41c3-439d-8082-ae24a601ba38/dir/file1`
`at org.apache.hadoop.ozone.om.request.key.OMKeysRenameRequest.validateAndUpdateCache(OMKeysRenameRequest.java:228)`

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeysRenameResponse.java
```diff
@@ -0,0 +1,135 @@
...
+      Table<String, OmKeyInfo> keyTable = omMetadataManager
+          .getKeyTable();
+      try {
+        acquiredLock =
+            omMetadataManager.getLock().acquireWriteLock(BUCKET_LOCK,
+                volumeName, bucketName);
+        // If toKeyName is null, then we need to only delete the fromKeyName
+        // from KeyTable. This is the case of replay where toKey exists but
+        // fromKey has not been deleted.
+        if (deleteFromKeyOnly()) {
```
bharatviswa504, line 89: Adding to the cache should be done as part of ValidateAndUpdateCache.
As in HA, we return the response without DB flush to be completed. So, if we don't update the cache before returning the response, subsequent requests might thing still key exists.

Posted PR for DeleteKeys regarding the same issue #1169

bharatviswa504: And also replay code is now not required, as it is taken care of by HDDS-3354.

captainzmc: Thanks Bharat for the suggestion, I have taken a close look at the implementation of #1169 with some very nice changes. In this PR I will synchronize the #1169 changes here to make sure they are implemented the same.

captainzmc:
Status updates:
I have updated this PR based on DeleteKeys #1195. Make sure RenameKeys and DeleteKeys can be implemented in the same way.

captainzmc:
> I have one comment regarding updateCache which i have seen during OMKeysDeleteRequest, same applies to OMKeysRenameRequest also.

Hi @bharatviswa504. Thanks for fixing OMKeysDeleteRequest. I have updated this PR based on DeleteKeys #1195. Make sure RenameKeys and DeleteKeys can be implemented in the same way. Could you please review this PR again?

captainzmc:
hi @xiaoyuyao @adoroszlai. I have updated this PR based on DeleteKeys #1195, that Bharat had fixed some issues.  Could you please review this PR again?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
```diff
@@ -750,6 +751,25 @@ public void renameKey(String volumeName, String bucketName,
...
+    HddsClientUtils.checkNotNull(keyMap);
+    Map <OmKeyArgs, String> keyArgsMap = new HashMap<>();
+    for (Map.Entry< String, String > entry : keyMap.entrySet()) {
+      OmKeyArgs keyArgs = new OmKeyArgs.Builder()
+          .setVolumeName(volumeName)
+          .setBucketName(bucketName)
+          .setKeyName(entry.getKey())
+          .build();
+      keyArgsMap.put(keyArgs, entry.getValue());
+    }
```
adoroszlai, line 768: What's the benefit of building a `Map<OmKeyArgs, String>` compared to passing plain `Map<String, String> keyMap` directly?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,360 @@
...
+                // Add to cache. Only fromKey should be deleted. ToKey already
+                // exists in DB as this transaction is a replay.
+                Table<String, OmKeyInfo> keyTable = omMetadataManager
+                    .getKeyTable();
+                keyTable.addCacheEntry(new CacheKey<>(fromKey),
+                    new CacheValue<>(Optional.absent(), trxnLogIndex));
+                renameKeyInfoList.add(new OmRenameKeyInfo(
+                    null, fromKeyValue));
+              }
+            }
```
adoroszlai, line 233: Replay logic was removed in 90e8211bb39d7ecbbdd0a4b4df55e5a5aa64c932.  Can you please update the PR (also to resolve conflicts)?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
captainzmc: The main purpose here is to keep up with the [renameKey implementation](https://github.com/apache/hadoop-ozone/pull/1150/files#diff-ddfee0637dda7038422139fc5a3cbfdeR751) and help make the code implementation clearer.

captainzmc:
> Thanks @captainzmc for updating the PR based on #1195. I would like to suggest waiting until that PR is merged before updating this one again, to avoid having to resolve merge conflicts multiple times.

Thanks @adoroszlai  for your suggestion. I’ll update this PR after #1195 merged and the conflict will be resolved together.

captainzmc:
> > Thanks @captainzmc for updating the PR based on #1195. I would like to suggest waiting until that PR is merged before updating this one again, to avoid having to resolve merge conflicts multiple times.
> 
> Thanks @adoroszlai for your suggestion. I’ll update this PR after #1195 merged and the conflict will be resolved together.

Update  status:
Now #1195 had merged. I had updated this PR, removed Replay Logic(as as #1082) and Resolved Conflicts. Now, this PR can continue to be reviewed.

captainzmc:
> > Thanks @captainzmc for updating the PR based on #1195. I would like to suggest waiting until that PR is merged before updating this one again, to avoid having to resolve merge conflicts multiple times.
> 
> Thanks @adoroszlai for your suggestion. I’ll update this PR after #1195 merged and the conflict will be resolved together.

captainzmc:
> > Thanks @captainzmc for updating the PR based on #1195. I would like to suggest waiting until that PR is merged before updating this one again, to avoid having to resolve merge conflicts multiple times.
> 
> Thanks @adoroszlai for your suggestion. I’ll update this PR after #1195 merged and the conflict will be resolved together.

Hi @adoroszlai, Now #1195 had merged. I had updated this PR, removed Replay Logic(as #1082) and Resolved Conflicts. Please take another look?

captainzmc:
> > Thanks @captainzmc for updating the PR based on #1195. I would like to suggest waiting until that PR is merged before updating this one again, to avoid having to resolve merge conflicts multiple times.
> 
> Thanks @adoroszlai for your suggestion. I’ll update this PR after #1195 merged and the conflict will be resolved together.

Hi @adoroszlai @xiaoyuyao @bharatviswa504  Now #1195 had merged. I had updated this PR, removed Replay Logic(as #1082) and Resolved Conflicts. Please take another look?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
adoroszlai: The signature of `OzoneManagerProtocol#renameKeys(Map<OmKeyArgs, String> keyMap)` accepts a batch of keys in different volumes/buckets.  But `RpcClient#renameKeys` only supports renaming multiple keys in a single volume/bucket.

My concern is:

1. lots of object conversion `String -> OmKeyArgs -> KeyArgs`
2. volume/bucket existence check performed for each key

I think given the limitation in current `RpcClient` the above is unnecessary overhead.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -675,6 +678,31 @@ public OmKeyInfo lookupKey(OmKeyArgs args) throws IOException {
...
+    RenameKeysRequest.Builder reqKeys = RenameKeysRequest.newBuilder();
+    List<RenameKeyRequest> renameKeyRequestList  = new ArrayList<>();
+    for (Map.Entry< OmKeyArgs, String> entry : keyMap.entrySet()) {
+      RenameKeyRequest.Builder reqKey = RenameKeyRequest.newBuilder();
+      OmKeyArgs args = entry.getKey();
+      KeyArgs keyArgs = KeyArgs.newBuilder()
+          .setVolumeName(args.getVolumeName())
+          .setBucketName(args.getBucketName())
+          .setKeyName(args.getKeyName())
+          .setModificationTime(Time.now()).build();
```
adoroszlai, line 692: Shouldn't modification time be updated on OM server-side?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -1227,12 +1231,76 @@ public void testRenameKey()
...
+
+    // rename nonexistent key
+    Map<String, String> keyMap1 = new HashMap();
+    keyMap1.put(keyName1, keyName2);
+    keyMap1.put(newKeyName2, keyName2);
+    try {
+      bucket.renameKeys(keyMap1);
+    } catch (OMException ex) {
+      Assert.assertEquals(PARTIAL_RENAME, ex.getResult());
+    }
```
adoroszlai, line 1298: I think this belongs to a separate test case.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -1227,12 +1231,76 @@ public void testRenameKey()
...
+  }
+
+  // Listing all volumes in the cluster feature has to be fixed after HDDS-357.
+  // TODO: fix this
+  @Ignore
```
adoroszlai, line 1303: This was recently removed in HDDS-3062, so it seems to be accidentally being added back (due to merge conflict?).

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,315 @@
...
+    boolean acquiredLock = false;
+    boolean renameStatus = true;
+    try {
+      for (RenameKeyRequest renameKeyRequest : renameKeysRequest
+          .getRenameKeyRequestList()) {
+        OzoneManagerProtocolProtos.KeyArgs keyArgs =
+            renameKeyRequest.getKeyArgs();
+
+        volumeName = keyArgs.getVolumeName();
+        bucketName = keyArgs.getBucketName();
```
adoroszlai, line 138: Please add support for bucket links.  Example:

https://github.com/apache/hadoop-ozone/blob/3571d7e72688b008c2667997cb4d824ccc9a8a3e/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java#L134-L136

`auditMap` needs to be created beforehand, eg.:

https://github.com/apache/hadoop-ozone/blob/3571d7e72688b008c2667997cb4d824ccc9a8a3e/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRenameRequest.java#L106

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -1227,12 +1231,76 @@ public void testRenameKey()
...
+    String value = "sample value";
+    store.createVolume(volumeName);
+    OzoneVolume volume = store.getVolume(volumeName);
+    volume.createBucket(bucketName);
+    OzoneBucket bucket = volume.getBucket(bucketName);
+    OzoneOutputStream out1 = bucket.createKey(keyName1,
+        value.getBytes().length, STAND_ALONE,
+        ONE, new HashMap<>());
+    out1.write(value.getBytes());
+    out1.close();
```
adoroszlai, line 1259: Can you please extract to a helper method (eg. `createTestKey`) to avoid duplication?  (Also include following `bucket.getKey` and following `assertEquals` calls.)

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -1227,12 +1231,76 @@ public void testRenameKey()
...
+
+    Assert.assertEquals(newKeyName1, bucket.getKey(newKeyName1).getName());
+    Assert.assertEquals(newKeyName2, bucket.getKey(newKeyName2).getName());
+
+    // old key should not exist
+    try {
+      bucket.getKey(keyName1);
+    } catch (OMException ex) {
+      Assert.assertEquals(KEY_NOT_FOUND, ex.getResult());
+    }
```
adoroszlai, line 1283: Can you please extract to helper method (eg. `assertKeyRenamed`), and also include preceding `assertEquals(newKeyName...)` call?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
captainzmc: Here I'll modify RPCClient to use Map <OmKeyArgs, String> keyArgsMap directly.

adoroszlai: It seems none of the methods implementing `ClientProtocol` have `OmKeyArgs` in the signature.  So I don't think `renameKeys` should do that either.

captainzmc: Thanks for @adoroszlai 's reply. At present, our client calls RpcClient#renameKeys through [OzoneBucket#renameKeys](https://github.com/apache/hadoop-ozone/pull/1150/files#diff-a4a0be1d5df61f0de3bdce909bb3a4a3R1273) to batch renameKeys.  OzoneBucket is a bucket-level API, we can only rename Keys in a single volume/bucket use it.
Do we need to change the RpcClient#renameKeys? If need change here any Suggestions here?

captainzmc:
Status updates. Fixed review issues. For RpcClient#renameKeys, currently we only support keys rename for the same bucket. To improve performance, I've removed the logic to double-check whether volume/bucket exists.

captainzmc:
> Thanks @captainzmc for updating the patch. It looks better now. However, I'm still not convinced about adding half-baked support for multi-bucket batch rename. Let's wait for others' feedback.

Thanks @adoroszlai ’s feedback.  Fixed auditMap update issues.
This PR is a subtask of HDDS-3620. The goal is to support batch operation when BasicOzoneFileSystem Rename folder to improve performance. At present, the rename operation under OFS and O3FS only supports in the same bucket. This is similar to the previous implementation of [deleteKeys](https://github.com/apache/hadoop-ozone/pull/814).

captainzmc:
> > Thanks @captainzmc for updating the patch. It looks better now. However, I'm still not convinced about adding half-baked support for multi-bucket batch rename. Let's wait for others' feedback.
> 
> Thanks @adoroszlai ’s feedback. Fixed auditMap update issues.
> This PR is a subtask of [HDDS-3620](https://issues.apache.org/jira/browse/HDDS-3620). The goal is to support batch operation when BasicOzoneFileSystem Rename folder to improve performance. At present, the rename operation under OFS and O3FS only supports in the same bucket. This is similar to the previous implementation of [deleteKeys](https://github.com/apache/hadoop-ozone/pull/814).

Updates PR to resolve conflict issues.
Hi, @xiaoyuyao @bharatviswa504 Could you help review this PR again?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
adoroszlai: Thanks @captainzmc for removing the existence check.  However, bucket link resolution is still performed for each key, which seems unnecessary since all keys are in the same volume/bucket.

captainzmc: Thanks @adoroszlai's feedback.  I had added new commit to fix the problem.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,304 @@
...
+        OzoneManagerProtocolProtos.KeyArgs keyArgs =
+            renameKeyRequest.getKeyArgs();
+
+        // resolve Bucket Link at the first time.
+        if (bucketName == null) {
+          keyArgs = resolveBucketLink(ozoneManager, keyArgs, auditMap);
+          volumeName = keyArgs.getVolumeName();
+          bucketName = keyArgs.getBucketName();
+          auditMap.put(VOLUME, volumeName);
+          auditMap.put(BUCKET, bucketName);
```
adoroszlai, line 144: Volume and bucket are already added to `auditMap` in `resolveBucketLink`, also accounting for the case when the requested bucket is a link.  Please remove these two lines to avoid overwriting the requested volume/bucket with the resolved one.

```suggestion
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,304 @@
...
+    boolean renameStatus = true;
+
+    try {
+      for (RenameKeyRequest renameKeyRequest : renameKeysRequest
+          .getRenameKeyRequestList()) {
+        OzoneManagerProtocolProtos.KeyArgs keyArgs =
+            renameKeyRequest.getKeyArgs();
+
+        // resolve Bucket Link at the first time.
+        if (bucketName == null) {
```
adoroszlai, line 139: For all other keys I think we should verify that the volume/bucket is the same as for the first key.

hadoop-ozone/interface-client/src/main/resources/proto.lock
```diff
@@ -83,6 +83,10 @@
                 "name": "DeleteKeys",
                 "integer": 38
               },
+              {
+                "name": "RenameKeys",
+                "integer": 39
+              },
```
adoroszlai, line 89: Please remove changes to `proto.lock` from the patch.  See HDDS-3991 for details.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
bharatviswa504: I agree with @adoroszlai here, we don't need KeyArgs.

HDDS-3930 created a new message DeleteKeyArgs {
    required string volumeName = 1;
    required string bucketName = 2;
    repeated string keys = 3;
}

we can come up with a new proto structure here.


message RenameKeysRequest {
    optional RenameKeyArgs deleteKeys = 1;
}

message RenameKeyArgs {
    required string volumeName = 1;
    required string bucketName = 2;
    repeated string keys = 3;
}

I am not completely sure what is the benefit to be consistent with RenameKeyRequest message structure.

bharatviswa504: sorry missed this, it has been already addressed. But I see we pass volumeName/bucketName with each Renamekey, we can set them only once, as right now rename is supported within a single volume/bucket

captainzmc: Thanks for @bharatviswa504 's suggestion.  Agree with your idea, I had already updated this PR. KeyArgs is no longer used. And set volumeName and bucketName only once. Can you help take another look?

hadoop-ozone/interface-client/src/main/resources/proto.lock
captainzmc: Thanks for @adoroszlai 's advice. Fixed the issues mentioned above. Can you help take another look?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    auditMap = buildAuditMap(auditMap, renamedKeys, unRenamedKeys);
+    auditLog(auditLogger, buildAuditMessage(OMAction.RENAME_KEYS, auditMap,
+        exception, getOmRequest().getUserInfo()));
+
+    switch (result) {
+    case SUCCESS:
+      LOG.debug("Rename Keys is successfully completed for auditMap:{}.",
+          auditMap.toString());
```
adoroszlai, line 245: Avoid eager `toString()` call in logging (especially debug).

```suggestion
          auditMap);
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,279 @@
...
+      break;
+    case FAILURE:
+      ozoneManager.getMetrics().incNumKeyRenameFails();
+      LOG.error("Rename keys failed for auditMap:{}.", auditMap.toString());
```
adoroszlai, line 249: Avoid eager `toString()` call in logging.

```suggestion
      LOG.error("Rename keys failed for auditMap:{}.", auditMap);
```

captainzmc:
> Thanks @captainzmc for changing the protocol. I think it is much improved now. I only have couple of minor comments left.

Thanks for @adoroszlai 's  feedback. Updated PR and Accepted suggestions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysRenameRequest.java
```diff
@@ -0,0 +1,279 @@
...
+
+        fromKeyValue.setUpdateID(trxnLogIndex, ozoneManager.isRatisEnabled());
+
+        fromKeyValue.setKeyName(toKeyName);
+
+        //Set modification time
+        fromKeyValue.setModificationTime(Time.now());
+
+        acquiredLock =
+            omMetadataManager.getLock().acquireWriteLock(BUCKET_LOCK,
```
bharatviswa504, line 187: I think here we should acquire lock and release lock at end of the operation. 

Lets say this thread checked the fromKey and toKey and other delete thread acquire the lock and deleted the fromKey and this thread waiting for lock, once delete completes we update the cache and rename it to toKey.

So, now we renamed a deleted Key. 

And also there can be other scenarios like commitKey committed toKey and this thread assumes there is no such Key.

To avoid these kinds of scenarios check of Key from Table and add to Response should be done holding lock.

captainzmc: Thank @bharatviswa504 . Agreed and  updated  this PR  to fix the problem.

captainzmc:
> Just a question will this change cause any compatibility issues? (As this is new API, only new client code will use it, just seeing if I missed anything)

Thanks for @bharatviswa504 's review.  RenameKeys is a new API and does not cause compatibility issues. Older clients can continue to use renameKey.

captainzmc:
hi @adoroszlai @bharatviswa504 Is there any progress here? I see that PR has been approved but the status hasn't been updated yet.

arp7:
The downside of the new API is it will hold the lock for a long time. This will have a significant perf impact on concurrent operations and that will come back to bite us. I think the correct long term fix for this is HDDS-2939 which will support atomic rename and delete operations.

captainzmc:
> @captainzmc is it a viable option to wait for [HDDS-2939](https://issues.apache.org/jira/browse/HDDS-2939) which will add atomic rename capability to OM?

Thanks @arp7  for being able to discuss this PR.
Yes, HDDS-2939 may be a good solution to the above problem. But HDDS-2939 still has a long way to merge into master. Currently our cluster use Rename and Delete old directories, each time taking a long time. So we modified the existing interface to get better performance. This can be seen as an interim solution.
[We talked about this in the addition of batchDelete.](https://github.com/apache/hadoop-ozone/pull/814#issuecomment-630790204)

xiaoyuyao:
I'm +1 to merge this patch as-is and we can revisit this when HDDS-2939 is merged. 

@arp7 we can mitigate the long lock holding issue by adding a limit on the max batch size (e.g., default to 1000). This can be added as follow up JIRA for both batch rename and delete.

arp7:
Thanks @xiaoyuyao, let's get this in. Can we file a follow up jira for the periodic lock release? HDDS-2939 will eventually implement atomic and efficient rename however it will take some time.

bharatviswa504:
When ratis is enabled, periodic lock release on the server is of no help, as we have a single thread executor.

Only the batch size from the client should be reduced that will help when ratis is enabled in OM.

xiaoyuyao:
bq. When ratis is enabled, periodic lock release on the server is of no help, as we have a single thread executor. (It can help readers, but not writers)

Good point. I agree with that. Blocking readers are not good for throughput as well. Let's open a separate to tune the default batch size to minimize the perf impact on readers and non-HA cases.

I will merge this one shortly. Thanks @captainzmc for the patience/contribution and all for the reviews and discussion.

bharatviswa504:
>Good point. I agree with that. Blocking readers are not good for throughput as well. Let's open a separate to tune the >default batch size to minimize the perf impact on readers and non-HA cases.

I have not understood What do we want to change.

1. Default Batch size controlled in client or have another parameter on server to release lock?

PR #1152, opened 2020-07-01
HDDS-3905. Show status of OM in the OM web ui

maobaolong:

Add a Status table for OM, after this PR, we can bring more status item to the status table.

HDDS-3905.

## How was this patch tested?

image


maobaolong:
@adoroszlai  Please take a look at this PR, the rpc  port is the only one of Status now, after this PR, we can bring more status to this table.

maobaolong:
@dineshchitlangia Please take a look, thanks.

PR #1155, opened 2020-07-01
HDDS-3908. Duplicate dot in Prometheus endpoint config name

adoroszlai:

Remove trailing dot from config prefix for `hdds.prometheus`.  I think this should be done before 0.6.0 release.

https://issues.apache.org/jira/browse/HDDS-3908

## How was this patch tested?

```
$ mvn -DskipTests clean package
$ grep hdds.prometheus hadoop-hdds/common/target/classes/ozone-default-generated.xml
    <name>hdds.prometheus.endpoint.token</name>
```


vivekratnavel:
Thanks @adoroszlai for the patch and @avijayanhwx for the review

PR #1156, opened 2020-07-01
HDDS-3879. Introduce SCM and OM layoutVersion zero to the VERSION file

sodonnel:

When SCM or OM is initialized, ensure a property like the following is written to the version file:

layoutVersion=0

https://issues.apache.org/jira/browse/HDDS-3879

## How was this patch tested?

Manually inspected the version file after the change.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
@@ -48,6 +48,7 @@ public abstract class Storage {
   public static final String STORAGE_DIR_CURRENT = "current";
   protected static final String STORAGE_FILE_VERSION = "VERSION";
   public static final String CONTAINER_DIR = "containerDir";
+  private static final int LAYOUT_VERSION = 0;
 
   private final NodeType nodeType;
   private final File root;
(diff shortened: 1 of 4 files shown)


sodonnel:
@avijayanhwx Is this the sort of change you were intending with this Jira? Note that I think it also covers HDDS-3830 too, as both OM and SCM use the same code from reading and writing the VERSION file.

avijayanhwx:
@sodonnel Thanks for working on this. I am OK with this change. In the future, we may also have to track a different version number which denotes software version. But, for now, this is OK with me.

swagle:
> @sodonnel Thanks for working on this. I am OK with this change. In the future, we may also have to track a different version number which denotes software version. But, for now, this is OK with me.

@avijayanhwx Shouldn't we add whatever we think would be needed now? Introducing it later will not cause incompat?

sodonnel:
I started on the changes to add in the software version too, but in doing that I starting to think about whether we need to actually store it. 

I will discuss with @avijayanhwx and see what conclusion we come to. If we need it, we may as well get it in as part of this change.

avijayanhwx:
> I started on the changes to add in the software version too, but in doing that I starting to think about whether we need to actually store it.
> 
> I will discuss with @avijayanhwx and see what conclusion we come to. If we need it, we may as well get it in as part of this change.

Yes, this change as it is can go in.

@swagle We will not have any backward compat issues with respect to layout/software version since the default behavior can handle that.

avijayanhwx:
LGTM +1

PR #1159, opened 2020-07-01
HDDS-3913. Recon build should ignore proxies

vivekratnavel:

Recon frontend-build should ignore proxies to avoid build failures when proxies are configured for maven.

https://issues.apache.org/jira/browse/HDDS-3913

## How was this patch tested?

Tested locally with proxies configured in ~/.m2/settings.xml

### DIFF
--- a/hadoop-ozone/recon/pom.xml
+++ b/hadoop-ozone/recon/pom.xml
@@ -88,6 +88,7 @@
         <artifactId>frontend-maven-plugin</artifactId>
         <version>${frontend-maven-plugin.version}</version>
         <configuration>
+          <npmInheritsProxyConfigFromMaven>false</npmInheritsProxyConfigFromMaven>
           <installDirectory>target</installDirectory>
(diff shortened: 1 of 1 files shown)


elek:
/ping @ChenSammi 

I guess, you have inernal builds. Do you cache npm artifacts in local maven proxy. Do you see any problem with this change?

GlenGeng:
> /ping @ChenSammi
> 
> I guess, you have inernal builds. Do you cache npm artifacts in local maven proxy. Do you see any problem with this change?

@elek 
We do see this problem in our internal environment, since need add proxies for `http` and `https`, and also use `<npmInheritsProxyConfigFromMaven>false</npmInheritsProxyConfigFromMaven>` to handle this problem.

This fix is fine from our(Tencent) side.

adoroszlai:
Thanks @vivekratnavel for the fix and @GlenGeng for the review.

PR #1161, opened 2020-07-03
HDDS-3917. Add recon to no_proxy of docker-config for acceptance test

GlenGeng:

For internal acceptance test, which have configured proxy for http, test cases for recon will be affected.

https://issues.apache.org/jira/browse/HDDS-3917

## How was this patch tested?

CI


GlenGeng:
@elek  Thanks for pointing this out !

It looks like for now all `docker-config` share the same `no_proxy` config. Changing them all can avoid the case that in future someone add `recon` to an existing docker test but forget change the `docker-config`, since this is a easily overlooked issue. I can revert the changed `docker-config` which does not consist of a `recon` component. Both are fine to me.

elek:
> Changing them all can avoid the case that in future someone add recon to an existing docker test but forget change the docker-config, since this is a easily overlooked issue

Yes, I can accept this argument. Especially as the problem can be detected only in an internal environment with http proxies. Easy to forget the update.

> Both are fine to me.

Me, too. Let me merge this in this form. It's more verbose, but as you wrote, it can provide additional safety....

PR #1162, opened 2020-07-04
HDDS-3921. IllegalArgumentException triggered in SCMContainerPlacemen…

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-3921

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -547,14 +547,14 @@ public class ReplicationManager
             containerPlacement.validateContainerPlacement(
                 targetReplicas, replicationFactor);
         int delta = replicationFactor - getReplicaCount(id, replicas);
-        if (placementStatus.isPolicySatisfied() && delta <= 0) {
-          LOG.debug(
-              "Container {} meets placement policy with inflight replicas", id);
-          return;
-        }
         final int misRepDelta = placementStatus.misReplicationCount();
         final int replicasNeeded
             = delta < misRepDelta ? misRepDelta : delta;
+        if (replicasNeeded <= 0) {
+          LOG.debug("Container {} meets replication requirement with " +
+              "inflight replicas", id);
+          return;
+        }
 
         final List<DatanodeDetails> excludeList = replicas.stream()
             .map(ContainerReplica::getDatanodeDetails)
@@ -615,7 +615,7 @@ public class ReplicationManager
 
     final ContainerID id = container.containerID();
     final int replicationFactor = container.getReplicationFactor().getNumber();
-    // Dont consider inflight replication while calculating excess here.
+    // Don't consider inflight replication while calculating excess here.
     int excess = replicas.size() - replicationFactor -
         inflightDeletion.getOrDefault(id, Collections.emptyList()).size();
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -512,7 +512,7 @@ public class ReplicationManager
    */
   private void handleUnderReplicatedContainer(final ContainerInfo container,
       final Set<ContainerReplica> replicas) {
-    LOG.debug("Handling underreplicated container: {}",
+    LOG.debug("Handling under-replicated container: {}",
         container.getContainerID());
     try {
       final ContainerID id = container.containerID();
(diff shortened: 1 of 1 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -543,11 +543,15 @@ private void handleUnderReplicatedContainer(final ContainerInfo container,
         List<DatanodeDetails> targetReplicas = new ArrayList<>(source);
         // Then add any pending additions
         targetReplicas.addAll(replicationInFlight);
-
-        int delta = replicationFactor - getReplicaCount(id, replicas);
         final ContainerPlacementStatus placementStatus =
             containerPlacement.validateContainerPlacement(
                 targetReplicas, replicationFactor);
+        int delta = replicationFactor - getReplicaCount(id, replicas);
```
xiaoyuyao, line 549: NIT: this line to move the line to calculate delta from 547->549 is not needed.  Only the line 550 is the core change that breaks out when we've already handle the under replicate container via inflight replication.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -543,11 +543,15 @@ private void handleUnderReplicatedContainer(final ContainerInfo container,
...
+        if (placementStatus.isPolicySatisfied() && delta <= 0) {
```
sodonnel, line 550: Rather than this new IF block here, would it make sense to simply add:

```
if (replicasNeeded <= 0) {
  LOG.debug(...);
  return;
}

At line 554 / 558, just after the line:

```
        final int replicasNeeded
            = delta < misRepDelta ? misRepDelta : delta;
```

That would avoid needing to check call `placementStatus.isPolicySatisfied()` and then `placementStatus.misReplicationCount()`.

sodonnel:
Thanks for this change. I also wonder if there is a bug in the method `isContainerUnderReplicated(...)` which leads to this problem, and results in more processing that necessary for mis-replicated containers with inFlight Additions.

In `isContainerUnderReplicated` is uses only the live replicas to check for mis-replication, but then it considers inflight add and delete for under replication. Should we also include the inflightAdds when considering mis-replication in that method?

For `isContainerOverReplicated` it also uses the inflightAdds and deletes to check for over replication.

ChenSammi:
> Thanks @ChenSammi for reporting the issue and propose the fix. The change LGTM.
> 
> I have a question wrt the excess calculation in handleOverReplicationContainer where we only consider inflightDeletion without inflightReplication at around line 615. Could this contribute the over replication issue as we will break out when a smaller excess value reaches 0 but the inflighreplication is still going?

@xiaoyuyao , I think inflightReplication not considered here is safer since replication has the change to fail.  Imaging we have 2 healthy replicas and 2 inflight replications, this case, send the command to delete the extra 1 replica until we are sure that we have 4 healthy replicas in hand.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
ChenSammi: This delta calculation and check in line 550 is required.  Currenlty placementStatus.isPolicySatisfied() only checks whether topology requirement is satisfied or not. It doesn't check whether it has enough replicas.

ChenSammi:
> Thanks for this change. I also wonder if there is a bug in the method `isContainerUnderReplicated(...)` which leads to this problem, and results in more processing that necessary for mis-replicated containers with inFlight Additions.
> 
> In `isContainerUnderReplicated` is uses only the live replicas to check for mis-replication, but then it considers inflight add and delete for under replication. Should we also include the inflightAdds when considering mis-replication in that method?
> 
> For `isContainerOverReplicated` it also uses the inflightAdds and deletes to check for over replication.

It's a good point. I don't think it's a real bug, but we can improve it.  HDDS-3942 to track it.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
ChenSammi: Make sense.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
xiaoyuyao: Got it. I agree delta calculation is needed. I mean move delta calculation after placementStatus is not necessary.

ChenSammi:
testDeleteKeyWithSlowFollower failed at leader membership check step. The test passed locally. It seems a timing issue, not relevant to this patch.

ChenSammi:
Thanks @sodonnel  and @xiaoyuyao for the review.

xiaoyuyao:
LGTM, +1. Thanks @ChenSammi for the contribution and all for the reviews. I will merge the PR shortly.

PR #1163, opened 2020-07-04
HDDS-3920. Too many redudant replications due to fail to get node's a…

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-3920

With the patch, over-replicated replicas are deleted by ReplicationManager. The total containers on each datanode drops from 80K~ to 8K~

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerReportHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerReportHandler.java
@@ -102,8 +102,15 @@ public class ContainerReportHandler extends AbstractContainerReportHandler
   public void onMessage(final ContainerReportFromDatanode reportFromDatanode,
                         final EventPublisher publisher) {
 
-    final DatanodeDetails datanodeDetails =
+    final DatanodeDetails dnFromReport =
         reportFromDatanode.getDatanodeDetails();
+    DatanodeDetails datanodeDetails =
+        nodeManager.getNodeByUuid(dnFromReport.getUuidString());
+    if (datanodeDetails == null) {
+      LOG.warn("Received container report from unknown datanode {}",
+          dnFromReport);
+      return;
+    }
     final ContainerReportsProto containerReport =
         reportFromDatanode.getReport();
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
@@ -241,43 +241,43 @@ public class SCMNodeManager implements NodeManager {
       DatanodeDetails datanodeDetails, NodeReportProto nodeReport,
       PipelineReportsProto pipelineReportsProto) {
 
-    InetAddress dnAddress = Server.getRemoteIp();
-    if (dnAddress != null) {
-      // Mostly called inside an RPC, update ip and peer hostname
-      datanodeDetails.setHostName(dnAddress.getHostName());
-      datanodeDetails.setIpAddress(dnAddress.getHostAddress());
-    }
-    try {
-      String dnsName;
-      String networkLocation;
-      datanodeDetails.setNetworkName(datanodeDetails.getUuidString());
-      if (useHostname) {
-        dnsName = datanodeDetails.getHostName();
-      } else {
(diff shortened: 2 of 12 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -260,12 +260,21 @@ public RegisteredCommand register(
...
-      clusterMap.add(datanodeDetails);
-      addEntryTodnsToUuidMap(dnsName, datanodeDetails.getUuidString());
-      // Updating Node Report, as registration is successful
-      processNodeReport(datanodeDetails, nodeReport);
-      LOG.info("Registered Data node : {}", datanodeDetails);
+      if (!isNodeRegistered(datanodeDetails)) {
+        clusterMap.add(datanodeDetails);
+        nodeStateManager.addNode(datanodeDetails);
+        // Check that datanode in nodeStateManager has topology parent set
+        DatanodeDetails dn =
```
xiaoyuyao, line 267: Can we use nodeStateManager.getNode(datanodeDetails) to avoid the UUID->String->UUID conversion?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -260,12 +260,21 @@ public RegisteredCommand register(
       if (networkLocation != null) {
         datanodeDetails.setNetworkLocation(networkLocation);
       }
-      nodeStateManager.addNode(datanodeDetails);
-      clusterMap.add(datanodeDetails);
-      addEntryTodnsToUuidMap(dnsName, datanodeDetails.getUuidString());
-      // Updating Node Report, as registration is successful
-      processNodeReport(datanodeDetails, nodeReport);
-      LOG.info("Registered Data node : {}", datanodeDetails);
+      if (!isNodeRegistered(datanodeDetails)) {
```
xiaoyuyao, line 263: Can we move isNodeRegistered check at the begin of register() to avoid unnecessary dnsname lookup?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerReportHandler.java
```diff
@@ -102,8 +102,15 @@ public ContainerReportHandler(final NodeManager nodeManager,
   public void onMessage(final ContainerReportFromDatanode reportFromDatanode,
                         final EventPublisher publisher) {
 
-    final DatanodeDetails datanodeDetails =
+    final DatanodeDetails dnFromReport =
         reportFromDatanode.getDatanodeDetails();
+    DatanodeDetails datanodeDetails =
```
xiaoyuyao, line 107: When SCM does not have information about a datanode, it is usually the datanode is dead. So skip processing those delayed report makes sense to me .

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
ChenSammi: Sure.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
ChenSammi: Sure.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerReportHandler.java
ChenSammi: Xiaoyu, the key point here is we need the DatanodeDetails object from the nodeManager which has topology parent  correctly set.  The DatanodeDetails object parsed from container report always has no parent which will make the Replication Manager treat it as a datanode under /default-rack, further treat the container as under-replicated since cross rack requirement is not satisfied.

xiaoyuyao: That's a good catch. Thanks for the details.

xiaoyuyao:
LGTM, +1 pending CI.

ChenSammi:

PR #1164, opened 2020-07-06
HDDS-3824: OM read requests should make SCM#refreshPipeline outside BUCKET_LOCK

rakeshadr:

Refresh pipeline info does a call to SCM and it can be moved outside the BUCKET_LOCK, this would help to improve the performance of read/write mix workloads.

https://issues.apache.org/jira/browse/HDDS-3824

## How was this patch tested?

Using TestOzoneFileSystem unit test cases.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -1701,13 +1701,15 @@ public class KeyManagerImpl implements KeyManager {
     String keyName = args.getKeyName();
 
     return getOzoneFileStatus(volumeName, bucketName, keyName,
-            args.getRefreshPipeline());
+            args.getRefreshPipeline(), false, null);
   }
 
   private OzoneFileStatus getOzoneFileStatus(String volumeName,
                                              String bucketName,
                                              String keyName,
-                                             boolean refreshPipeline)
+                                             boolean refreshPipeline,
+                                             boolean sortDatanodes,
+                                             String clientAddress)
       throws IOException {
     OmKeyInfo fileKeyInfo = null;
     metadataManager.getLock().acquireReadLock(BUCKET_LOCK, volumeName,
@@ -1742,6 +1744,9 @@ public class KeyManagerImpl implements KeyManager {
       if (fileKeyInfo != null) {
         if (refreshPipeline) {
           refreshPipeline(fileKeyInfo);
+        }
+        if (sortDatanodes) {
+          sortDatanodeInPipeline(fileKeyInfo, clientAddress);
         }
         return new OzoneFileStatus(fileKeyInfo, scmBlockSize, false);
       }
@@ -1898,15 +1903,10 @@ public class KeyManagerImpl implements KeyManager {
     String bucketName = args.getBucketName();
     String keyName = args.getKeyName();
     OzoneFileStatus fileStatus = getOzoneFileStatus(volumeName, bucketName,
-            keyName, false);
+            keyName, args.getRefreshPipeline(), args.getSortDatanodes(),
+            clientAddress);
       //if key is not of type file or if key is not found we throw an exception
-    if (fileStatus != null && fileStatus.isFile()) {
-      if (args.getRefreshPipeline()) {
-        refreshPipeline(fileStatus.getKeyInfo());
-      }
-      if (args.getSortDatanodes()) {
-        sortDatanodeInPipeline(fileStatus.getKeyInfo(), clientAddress);
-      }
+    if (fileStatus.isFile()) {
       return fileStatus.getKeyInfo();
     }
     throw new OMException("Can not write to directory: " + keyName,
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -1747,8 +1747,8 @@ public class KeyManagerImpl implements KeyManager {
     // Key is not found, throws exception
     if (LOG.isDebugEnabled()) {
       LOG.debug("Unable to get file status for the key: volume: {}, bucket:" +
-                      " {}, key: {}, with error: No such file exists.", volumeName,
-              bucketName, keyName);
+                      " {}, key: {}, with error: No such file exists.",
+              volumeName, bucketName, keyName);
     }
     throw new OMException("Unable to get file status: volume: " +
             volumeName + " bucket: " + bucketName + " key: " + keyName,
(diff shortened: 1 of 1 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1706,36 +1718,41 @@ public OzoneFileStatus getFileStatus(OmKeyArgs args) throws IOException {
...
+    } finally {
+      metadataManager.getLock().releaseReadLock(BUCKET_LOCK, volumeName,
+              bucketName);
+
+      // if the key is a file then do refresh pipeline info in OM by asking SCM
       if (fileKeyInfo != null) {
-        if (args.getRefreshPipeline()) {
+        if (refreshPipeline) {
           refreshPipeline(fileKeyInfo);
         }
```
xiaoyuyao, line 1742: should we order datanodes when the key is a file?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -659,15 +660,6 @@ public OmKeyInfo lookupKey(OmKeyArgs args, String clientAddress)
           });
         }
       }
-      // Refresh container pipeline info from SCM
```
bharatviswa504, line 662: Can we also move the generation of the secret token also outside of the lock?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Thanks for the review. Good point. For sorting dns, we need to introduce a new arg **clientAddress** to APIs [listStatus](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/fs/OzoneManagerFS.java#L52), [getFileStatus](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/fs/OzoneManagerFS.java#L34) interfaces.

Need to pass a new argument **clientAddress** like [lookupFile](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/fs/OzoneManagerFS.java#L45) API. Can we do interface changes in a separate jira task ?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Thanks for the review. Sure, will do it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: @xiaoyuyao, IMHO, your proposal is unrelated to this perf changes and to do focussed review/commit, I have created HDDS-3947 jira to handle your comment. Hope this is fine with you. Thanks!

rakeshadr:
@xiaoyuyao @bharatviswa504,
Please review the updated changes when you get a chance. Thanks!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1877,26 +1897,18 @@ public OmKeyInfo lookupFile(OmKeyArgs args, String clientAddress)
...
-      }
+    OzoneFileStatus fileStatus = getOzoneFileStatus(volumeName, bucketName,
+            keyName, false);
       //if key is not of type file or if key is not found we throw an exception
-    } finally {
-      metadataManager.getLock().releaseReadLock(BUCKET_LOCK, volumeName,
-          bucketName);
+    if (fileStatus != null && fileStatus.isFile()) {
+      if (args.getRefreshPipeline()) {
+        refreshPipeline(fileStatus.getKeyInfo());
```
xiaoyuyao, line 1905: The last parameter of getOzoneFileStatus() should have refreshPipeline handled already. Can we pass args.getRefreshPipeline() on line 1901?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1877,26 +1897,18 @@ public OmKeyInfo lookupFile(OmKeyArgs args, String clientAddress)
...
+      }
+      if (args.getSortDatanodes()) {
```
xiaoyuyao, line 1907: NIT: sortDatanodes can be handled similarly in getOzoneFileStatus?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Done

rakeshadr:
Thanks @xiaoyuyao for the comments. I have updated PR, kindly review it again.

rakeshadr:
_Note_:- TestWatchForCommit#test2WayCommitForTimeoutException failure is unrelated to my changes. It is passing in my local env.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2083,6 +2092,11 @@ private void listStatusFindKeyInTableCache(
       metadataManager.getLock().releaseReadLock(BUCKET_LOCK, volumeName,
           bucketName);
     }
+    if (args.getRefreshPipeline()) {
+      for(OzoneFileStatus fileStatus : fileStatusList){
+        refreshPipeline(fileStatus.getKeyInfo());
```
xiaoyuyao, line 2097: I'm +1 on this JIRA. 

Can we file a follow up JIRA to allow batching refreshPipeline call instead of individual calls to save RPC traffic.

rakeshadr: Thanks @xiaoyuyao , I have raised HDDS-3961 to handle this case.

rakeshadr:
Thanks @xiaoyuyao and @bharatviswa504  for the reviews and committing the PR.

PR #1165, opened 2020-07-06
HDDS-3923. Display the safemode status on scm page

maobaolong:

https://issues.apache.org/jira/browse/HDDS-3923

## How was this patch tested?

image


maobaolong:
@dineshchitlangia thanks for merge the PR https://github.com/apache/hadoop-ozone/pull/1109 , it really helps us able to see the reason why scm stay in the safemode, but for convenience, it's would be better to see the safemode rule statuses from webpage. Please take a look at this PR, thanks

PR #1166, opened 2020-07-06
HDDS-3914. Remove LevelDB configuration option for DN Metastore

hanishakoneru:

LevelDB support was removed for OM and SCM DBs but DN Metastore can still be configured to use LevelDB or RocksDB. 
This Jira proposes to remove LevelDB configuration option for DN Metastore (ozone.metastore.impl) and use RocksDB only.

https://issues.apache.org/jira/browse/HDDS-3914

## How was this patch tested?

Deprecation patch. No unit tests required.


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainerCheck.java
```diff
@@ -186,8 +186,8 @@ private void checkContainerFile() throws IOException {
     }
 
     dbType = onDiskContainerData.getContainerDBType();
-    if (!dbType.equals(OZONE_METADATA_STORE_IMPL_ROCKSDB) &&
-        !dbType.equals(OZONE_METADATA_STORE_IMPL_LEVELDB)) {
+    if (!dbType.equals(CONTAINER_DB_TYPE_ROCKSDB) &&
```
xiaoyuyao, line 189: Thanks @hanishakoneru  for working on this. Not sure if I understand correctly, since we are removing LevelDB support on DN, do we still need to define two types of DB for container here?

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/utils/TestMetadataStore.java
```diff
@@ -90,12 +93,12 @@ public void init() throws IOException {
         + "-" + storeImpl.toLowerCase());
 
     OzoneConfiguration conf = new OzoneConfiguration();
-    conf.set(OzoneConfigKeys.OZONE_METADATA_STORE_IMPL, storeImpl);
 
     store = MetadataStoreBuilder.newBuilder()
         .setConf(conf)
         .setCreateIfMissing(true)
         .setDbFile(testDir)
+        .setDBType(storeImpl)
```
xiaoyuyao, line 101: Do we still need this?

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/utils/TestMetadataStore.java
```diff
@@ -166,12 +170,12 @@ public void testIterator() throws Exception {
...
 
     OzoneConfiguration conf = new OzoneConfiguration();
-    conf.set(OzoneConfigKeys.OZONE_METADATA_STORE_IMPL, storeImpl);
+
     String dbType;
     GenericTestUtils.setLogLevel(MetadataStoreBuilder.LOG, Level.DEBUG);
     GenericTestUtils.LogCapturer logCapturer =
         GenericTestUtils.LogCapturer.captureLogs(MetadataStoreBuilder.LOG);
-    if (storeImpl.equals(OzoneConfigKeys.OZONE_METADATA_STORE_IMPL_LEVELDB)) {
+    if (storeImpl.equals(CONTAINER_DB_TYPE_LEVELDB)) {
```
xiaoyuyao, line 178: The logic needs to be reversed, maybe a typo.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainerCheck.java
hanishakoneru: We are deprecating LevelDB support. I kept it there in case there are old containers using LevelDB.

I am up for removing LevelDB support completely. I think it is safe to assume that there would not be a Container using LevelDB.

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/utils/TestMetadataStore.java
hanishakoneru: For the same reason as above.

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/utils/TestMetadataStore.java
hanishakoneru: Thanks. Will fix it.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainerCheck.java
xiaoyuyao: That makes sense to me.

vivekratnavel:
@hanishakoneru Thanks for working on this and @xiaoyuyao thanks for the review!

PR #1169, opened 2020-07-07
HDDS-3930. Fix OMKeyDeletesRequest.

bharatviswa504:

1. The cache of the key should be updated in ValidateAndUpdateCache, as we return response once after adding to cache, and before DoubleBuffer flushes to disk using OmClientResponse#addToDBBatch.
2. Changed return type of DeleteKeysResponse, as anyway when success we are not returning delete list and also when failure, we fail complete batch. So we can just return status.

https://issues.apache.org/jira/browse/HDDS-3930

## How was this patch tested?

Existing tests.

### DIFF
--- a/hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
+++ b/hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
@@ -303,6 +303,8 @@ enum Status {
 
     INVALID_VOLUME_NAME = 61;
 
+    PARTIAL_DELETE = 62;
+
     // When transactions are replayed
     REPLAY = 100;
 }
@@ -848,7 +850,18 @@ message DeleteKeyRequest {
 }
 
 message DeleteKeysRequest {
-    repeated KeyArgs keyArgs = 1;
+    optional DeleteKeyArgs deleteKeys = 1;
+}
+
+message DeleteKeyArgs {
+    required string volumeName = 1;
+    required string bucketName = 2;
+    repeated string keys = 3;
+}
+
+message DeleteKeysResponse {
+    optional DeleteKeyArgs unDeletedKeys = 1;
+    optional bool status = 2;
 }
 
 message DeleteKeyResponse {
@@ -866,9 +879,6 @@ message DeletedKeys {
     repeated string keys = 3;
 }
 
-message DeleteKeysResponse {
-    optional bool status = 1;
-}
 
 
 message PurgeKeysRequest {
--- a/hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
+++ b/hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
@@ -867,9 +867,9 @@ message DeletedKeys {
 }
 
 message DeleteKeysResponse {
-    repeated KeyInfo deletedKeys = 1;
-    repeated KeyInfo unDeletedKeys = 2;
+    optional bool status = 1;
 }
+
 
 message PurgeKeysRequest {
     repeated DeletedKeys deletedKeys = 1;
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -17,15 +17,34 @@
 
 package org.apache.hadoop.ozone.om;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.ObjectReader;
-import com.fasterxml.jackson.databind.ObjectWriter;
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
-import com.google.protobuf.BlockingService;
-import com.google.protobuf.ProtocolMessageEnum;
-import org.apache.commons.lang3.NotImplementedException;
-import org.apache.commons.lang3.StringUtils;
+import javax.management.ObjectName;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.security.KeyPair;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.cert.CertificateException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+
 import org.apache.hadoop.conf.StorageUnit;
 import org.apache.hadoop.crypto.key.KeyProvider;
 import org.apache.hadoop.crypto.key.KeyProviderCryptoExtension;
@@ -143,41 +162,16 @@ import org.apache.hadoop.util.JvmPauseMonitor;
 import org.apache.hadoop.util.KMSUtil;
 import org.apache.hadoop.util.ReflectionUtils;
 import org.apache.hadoop.util.ShutdownHookManager;
-import org.apache.ratis.proto.RaftProtos.RaftPeerRole;
-import org.apache.ratis.server.protocol.TermIndex;
-import org.apache.ratis.util.FileUtils;
-import org.apache.ratis.util.LifeCycle;
-import org.bouncycastle.pkcs.PKCS10CertificationRequest;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.ObjectWriter;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.protobuf.BlockingService;
+import com.google.protobuf.ProtocolMessageEnum;
+import org.apache.commons.lang3.NotImplementedException;
+import org.apache.commons.lang3.StringUtils;
 
-import javax.management.ObjectName;
-import java.io.BufferedWriter;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStreamWriter;
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.StandardCopyOption;
-import java.security.KeyPair;
-import java.security.PrivateKey;
-import java.security.PublicKey;
-import java.security.cert.CertificateException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Timer;
-import java.util.TimerTask;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.TimeUnit;
 
 import static org.apache.hadoop.hdds.HddsConfigKeys.HDDS_BLOCK_TOKEN_ENABLED;
 import static org.apache.hadoop.hdds.HddsConfigKeys.HDDS_BLOCK_TOKEN_ENABLED_DEFAULT;
@@ -216,6 +210,14 @@ import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.INVA
 import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.KEY_NOT_FOUND;
 import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.TOKEN_ERROR_OTHER;
 import static org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OzoneManagerService.newReflectiveBlockingService;
+
+import org.apache.ratis.proto.RaftProtos.RaftPeerRole;
+import org.apache.ratis.server.protocol.TermIndex;
+import org.apache.ratis.util.FileUtils;
+import org.apache.ratis.util.LifeCycle;
+import org.bouncycastle.pkcs.PKCS10CertificationRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Ozone Manager is the metadata manager of ozone.
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -2229,8 +2229,8 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
    */
   @Override
   public void deleteKeys(OmDeleteKeys deleteKeys) throws IOException {
-    throw new NotImplementedException("OzoneManager does not require this to " +
-        "be implemented. As write requests use a new approach");
+    throw new UnsupportedOperationException("OzoneManager does not require " +
+        "this to be implemented. As write requests use a new approach");
   }
 
   @Override
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -169,7 +169,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.BlockingService;
 import com.google.protobuf.ProtocolMessageEnum;
-import org.apache.commons.lang3.NotImplementedException;
 import org.apache.commons.lang3.StringUtils;
 
 
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -17,34 +17,15 @@
 
 package org.apache.hadoop.ozone.om;
 
-import javax.management.ObjectName;
-import java.io.BufferedWriter;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStreamWriter;
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.StandardCopyOption;
-import java.security.KeyPair;
-import java.security.PrivateKey;
-import java.security.PublicKey;
-import java.security.cert.CertificateException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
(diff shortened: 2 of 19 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -116,89 +111,112 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+        validateBucketAndVolume(omMetadataManager, volumeName, bucketName);
+      }
+
+
+      // Check if any of the key in the batch cannot be deleted. If exists the
+      // batch delete will be failed.
+
+      for (indexFailed = 0; indexFailed < deleteKeyArgsList.size();
+           indexFailed++) {
+        KeyArgs deleteKeyArgs = deleteKeyArgsList.get(0);
```
smengcl, line 141: Why do we always get the first element here? Typo?

Also, I didn't find existing test for `OMKeysDeleteRequest` or `OMKeysDeleteResponse`. It'd be a good idea to add some tests for sanity checks.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -867,10 +867,10 @@ message DeletedKeys {
 }
 
 message DeleteKeysResponse {
-    repeated KeyInfo deletedKeys = 1;
-    repeated KeyInfo unDeletedKeys = 2;
```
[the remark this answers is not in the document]
bharatviswa504: I don't see any usage of it, and also I see no real usage of this result in the client. 
And this is a new API adding, and will be released in next ozone release.

One more reason, as we fail whole batch, I don't see real use case for it as of current behavior.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -116,89 +111,112 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-      auditLog(auditLogger, buildAuditMessage(
-          OMAction.DELETE_KEY, auditMap, exception, userInfo));
+    // When we get any error during iteration build the remaining audit map
+    // from deleteKeyArgsList.
+    for (int i = indexFailed; i < deleteKeyArgsList.size(); i++) {
+      buildKeyArgsAuditMap(deleteKeyArgsList.get(i));
     }
 
+    auditLog(auditLogger, buildAuditMessage(
+        OMAction.DELETE_KEYS, auditMap, exception, userInfo));
```
xiaoyuyao, line 211: As @smengcl pointed out, we only audit the first delete and miss all the others here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
bharatviswa504: yes, it is a typo. Not sure why any test has caught this bug.
I will add some UT's to test this.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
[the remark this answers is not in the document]
xiaoyuyao: HDDS-3916 is in progress to use it on client side to resume delete failures.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
bharatviswa504: Fixed above issue, that will take care of this issue too.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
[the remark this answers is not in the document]
bharatviswa504: Will restore the original behavior.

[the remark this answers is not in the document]
smengcl: Thanks. Makes sense to me.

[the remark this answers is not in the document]
bharatviswa504: Added tests.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -116,93 +111,116 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-        bucketName = deleteKeyArgs.getBucketName();
-        keyName = deleteKeyArgs.getKeyName();
-        auditMap = buildKeyArgsAuditMap(deleteKeyArgs);
+
+        // Do we need to fail the batch if one of the key does not exist?
+        // For now following the previous code behavior. If this code changes
+        // behavior, this will be incompatible change across upgrades, and we
+        // need to version the Requests and do logic accordingly.
+
+        if (omKeyInfo == null) {
```
xiaoyuyao, line 156: Client may get an outdated list of keys and send the list for delete. Instead of fail the complete deletion, I suggest we update the undelete list that returned to the client.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -116,93 +111,116 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+          LOG.error("Key does not exist {}", objectKey);
+          throw new OMException("Key Not Found " + objectKey, KEY_NOT_FOUND);
+        }
+
         // check Acl
         checkKeyAcls(ozoneManager, volumeName, bucketName, keyName,
```
xiaoyuyao, line 162: The Acl check could also failed delete for some files, this could also be reflected in the undeleted list.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeysDeleteResponse.java
```diff
@@ -69,65 +64,36 @@ public OMKeysDeleteResponse(@Nonnull OMResponse omResponse) {
   public void addToDBBatch(OMMetadataManager omMetadataManager,
                            BatchOperation batchOperation) throws IOException {
 
+    String volumeName = "";
+    String bucketName = "";
+    String keyName = "";
     for (OmKeyInfo omKeyInfo : omKeyInfoList) {
```
xiaoyuyao, line 70: we should add only those that can be deleted (pass existence check and acl check, etc.) to the batch.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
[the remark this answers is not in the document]
xiaoyuyao: bq. One more reason, as we fail whole batch, I don't see real use case for it as of current behavior.

I think the writebatch should contain only those that can be deleted. Those undeleted will be returned to client.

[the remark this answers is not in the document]
bharatviswa504: The current behavior is we fail the entire batch if we are not able to delete the key.

So, proposal is to change this behavior? 

And also I see we fail to delete only when key not found. In this case, there is no real point in returning them as unDeletedKeys.

2nd case we fail is when checking ACLS,  when user does not have permission, in this case even if we return unDeletedKeys, not sure how it will be used.

3rd case, when reading DB failure, then also whole batch will be failed. As anyway, at that time OM will be terminated, as this is a critical error to avoid DB divergence.

So, still, I am not really sure what is real use case of returning unDeletedKeys and how that will be used. Just trying to understand it. 

The current PR, kept the behavior of existing behavior, where it used to fail entire batch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
bharatviswa504: Yes agree with this part, will change it to return batch delete with ignoring KEY_NOT_FOUND,

But I still I don't get what is the purpose of returning this as unDeletedKeys, do we want to send a status also for each Key then?

Like Key1, KEY_NOT_FOUND
        Key2, ACL_ERROR

What is the plan/how this will be used??

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
adoroszlai: Sorry, but I don't think it's fixed.  `auditMap` still contains info about only one key.

It is updated for each key in the first loop, overwriting previous values:

https://github.com/apache/hadoop-ozone/blob/62eded87415377b617d69b195f519b7fcb594e40/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java#L139-L142

as the map is not used until this `auditLog()` call.

Furthermore, audit map is built but discarded (not even stored in local var) for keys starting with `indexFailed`:

https://github.com/apache/hadoop-ozone/blob/62eded87415377b617d69b195f519b7fcb594e40/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java#L204-L207

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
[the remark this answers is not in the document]
xiaoyuyao: bq. The current behavior is we fail the entire batch if we are not able to delete the key. So, proposal is to change this behavior?

The goal is to not fail the whole deletion when something failed in the middle and return the list of deleted and undeleted to the client. We can't get there if delete and undelete list are removed from the proto. Here is the reason for keeping them for HDDS-3916. 

1. It can be expensive to lookup or retry the whole list again if one of the deletion failed. 
2. Returning undelete file list reduce the amount of unnecessary processing.

Also holding the bucket lock once for all may not be a good idea to large delete as that will prevent other clients operation on the same bucket. grab and release individual or in small batches give other clients opportunities to access buckets.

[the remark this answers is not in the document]
bharatviswa504: Got it. So, when this information is returned if the plan is to retry, I will add them.

The reason for my question is, we fail when we don't have permission/key not found we return them, but if this information is useful, will add that back.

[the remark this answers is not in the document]
bharatviswa504: As in HA, and when ratis is enabled as default for non-HA, it is a single thread executor now. So, this optimization is not really required. For non-HA, it might help, I will open a new Jira for this improvement.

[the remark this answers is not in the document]
bharatviswa504: Created https://issues.apache.org/jira/browse/HDDS-3948 for this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
bharatviswa504: Good catch. Thanks fixed it

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
bharatviswa504: Done. For now, made it simple returning unDeletedKeys.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
bharatviswa504: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeysDeleteResponse.java
bharatviswa504: Done

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -848,7 +850,18 @@ message DeleteKeyRequest {
 }
 
 message DeleteKeysRequest {
-    repeated KeyArgs keyArgs = 1;
+    optional DeleteKeyArgs deleteKeys = 1;
```
xiaoyuyao, line 853: Should this be required?

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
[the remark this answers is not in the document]
xiaoyuyao: Looks good to me.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
bharatviswa504: In proto 3 onwards all fields are optional.
So, followed that approach and declared optional. (In future if something is changed, we can still be backward compatible)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2220,16 +2224,13 @@ public void deleteKey(OmKeyArgs args) throws IOException {
...
    */
   @Override
-  public void deleteKeys(List<OmKeyArgs> args) throws IOException {
-    if (args != null) {
-      for (OmKeyArgs keyArgs : args) {
-        deleteKey(keyArgs);
-      }
-    }
+  public void deleteKeys(OmDeleteKeys deleteKeys) throws IOException {
+    throw new NotImplementedException("OzoneManager does not require this to " +
```
adoroszlai, line 2232: Please consider changing to `UnsupportedOperationException` as
> [`NotImplementedException`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/NotImplementedException.html) represents the case where the author has yet to implement the logic at this point in the program

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -116,93 +98,112 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-                unDeletedKeys));
+      result = Result.FAILURE;
+      exception = ex;
+      createErrorOMResponse(omResponse, ex);
+
+      // reset deleteKeys as request failed.
+      deleteKeys = new ArrayList<>();
+      // Add all keys which are failed due to any other exception .
+      for (int i = indexFailed; i < length; i++) {
+        unDeletedKeys.addKeys(deleteKeyArgs.getKeys(indexFailed));
```
adoroszlai, line 174: Should not repeatedly add the same key:

```suggestion
      for (int i = indexFailed; i < length; i++) {
        unDeletedKeys.addKeys(deleteKeyArgs.getKeys(i));
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -116,93 +98,112 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+    auditMap = buildDeleteKeysAuditMap(volumeName, bucketName, deleteKeys,
+        unDeletedKeys.getKeysList());
+
+    auditLog(auditLogger, buildAuditMessage(
+        OMAction.DELETE_KEYS, auditMap, exception, userInfo));
+
 
     switch (result) {
     case SUCCESS:
       omMetrics.decNumKeys();
```
adoroszlai, line 199: Should decrease number of keys by `deleteKeys.size()`.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -116,93 +98,112 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
       LOG.debug("Key deleted. Volume:{}, Bucket:{}, Key:{}", volumeName,
           bucketName, keyName);
```
adoroszlai, line 201: Might want to log all keys:

```suggestion
      LOG.debug("Key deleted. Volume:{}, Bucket:{}, Key:{}", volumeName,
          bucketName, deleteKeys);
```

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
xiaoyuyao: Good to know. LGTM.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
xiaoyuyao, line 201: In the success case, the omMetrics.decNumKeys() only -1, which does not reflect the deleted keys if there are more than one key in the request. 

NIT: in success case, we only put the keyName of the last successful delete. If we want to put the names of all deleted keys. We should guard it with if (LOG.isDebugEnabled())

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -116,93 +98,112 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
       break;
-    case REPLAY:
-      LOG.debug("Replayed Transaction {} ignored. Request: {}",
-          trxnLogIndex, deleteKeyRequest);
-      break;
     case FAILURE:
       omMetrics.incNumKeyDeleteFails();
-      LOG.error("Key delete failed. Volume:{}, Bucket:{}, Key{}." +
-          " Exception:{}", volumeName, bucketName, keyName, exception);
+      LOG.error("Key delete failed. Volume:{}, Bucket:{}, Key:{}",
```
xiaoyuyao, line 205: Should we reword the error message to differentiate with the single delete.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -116,93 +98,112 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+          volumeName, bucketName, keyName, exception);
       break;
     default:
       LOG.error("Unrecognized Result for OMKeyDeleteRequest: {}",
```
xiaoyuyao, line 209: OMKeyDeleteRequest=> OMKeysDeleteRequest

bharatviswa504:
Thank You @xiaoyuyao and @adoroszlai for the review.
Addressed review comments.

bharatviswa504:
This is being closed and opened as #1195 
As the branch for this accidentally created in apache repo.

PR #1170, opened 2020-07-07
HDDS-3910. JooqCodeGenerator interrupted but still alive

cku328:

Set `cleanupDaemonThreads` to false so that the maven doesn't interrupt the daemon threads on quit.
_(Disabling cleanupDaemonThreads suppresses this warning)_

When maven has no more work to do, the VM will normally terminate any remaining daemon threads.
After the jooq code generator generates classes for recon, it is inactive, so I don't think it will interfere with subsequent activity.

https://issues.apache.org/jira/browse/HDDS-3910

## How was this patch tested?

Compiled successfully.

In my test results, the time spent on the affected part is reduced from `17 s 994 ms` to `2 s 943 ms`.

image

### DIFF
--- a/hadoop-ozone/recon/pom.xml
+++ b/hadoop-ozone/recon/pom.xml
@@ -48,6 +48,7 @@
           <executable>java</executable>
           <classpathScope>compile</classpathScope>
           <mainClass>org.hadoop.ozone.recon.codegen.JooqCodeGenerator</mainClass>
+          <cleanupDaemonThreads>false</cleanupDaemonThreads>
           <arguments>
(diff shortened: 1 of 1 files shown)


cku328:
Thanks @adoroszlai  for the review.

adoroszlai:
@avijayanhwx please let us know if you have any concerns about this change, otherwise I would like to merge it.

avijayanhwx:
Thanks for fixing this @cku328. LGTM +1

adoroszlai:
Thanks @avijayanhwx for the review.

PR #1174, opened 2020-07-08
HDDS-3918. ConcurrentModificationException in ContainerReportHandler.…

xiaoyuyao:
…onMessage.

Use thread safe HashSet for NodeStateMap#nodeToContainer map

https://issues.apache.org/jira/browse/HDDS-3918

## How was this patch tested?

Unit test ([TestCME.java](https://issues.apache.org/jira/secure/attachment/13007245/TestCME.java) attached to the linked JIRA.)

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/states/NodeStateMap.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/states/NodeStateMap.java
@@ -67,7 +67,7 @@ public class NodeStateMap {
    */
   private void initStateMap() {
     for (NodeState state : NodeState.values()) {
-      stateMap.put(state, new HashSet<>());
+      stateMap.put(state, ConcurrentHashMap.newKeySet());
     }
   }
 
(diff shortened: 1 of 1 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/states/NodeStateMap.java
```diff
@@ -88,7 +88,7 @@ public void addNode(DatanodeDetails datanodeDetails, NodeState nodeState)
         throw new NodeAlreadyExistsException("Node UUID: " + id);
       }
       nodeMap.put(id, new DatanodeInfo(datanodeDetails));
-      nodeToContainer.put(id, new HashSet<>());
+      nodeToContainer.put(id, new ConcurrentHashMap().newKeySet());
```
adoroszlai, line 91: Nit: [newKeySet](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#newKeySet--) is `static`, so creating new `ConcurrentHashMap` instance is unnecessary.

```suggestion
      nodeToContainer.put(id, ConcurrentHashMap.newKeySet());
```

xiaoyuyao: Good point. I also make change to stateMap which has similar issue.

xiaoyuyao:
Thanks @adoroszlai for the review and @ChenSammi for reporting the issue. PR has been merged.

PR #1178, opened 2020-07-08
HDDS-3939. Update proto.lock files

vivekratnavel:

update proto.lock files that were missed in HDDS-426

https://issues.apache.org/jira/browse/HDDS-3939

## How was this patch tested?

No testing needed.


vivekratnavel:
@bharatviswa504 Thanks for the review!

PR #1182, opened 2020-07-09
HDDS-3926. OM Token Identifier table should use in-house serialization.

prashantpogde:

Ozone Manager Token Identifier table should use in-house serialization rather than rely on proto serialization for key.

https://issues.apache.org/jira/browse/HDDS-3926

## How was this patch tested?

CleanBuild and Integration Tests

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
@@ -21,20 +21,21 @@ import java.io.DataInput;
 import java.io.DataInputStream;
 import java.io.DataOutput;
 import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
-import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.hadoop.hdds.annotation.InterfaceAudience;
 import org.apache.hadoop.hdds.annotation.InterfaceStability;
+import org.apache.hadoop.io.DataInputBuffer;
+import org.apache.hadoop.io.DataOutputBuffer;
 import org.apache.hadoop.io.Text;
+import org.apache.hadoop.io.WritableUtils;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMTokenProto;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMTokenProto.Type;
 import org.apache.hadoop.security.token.delegation.AbstractDelegationTokenIdentifier;
 
 import static org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMTokenProto.Type.S3AUTHINFO;
+
 
 /**
  * The token identifier for Ozone Master.
@@ -85,115 +86,50 @@ public class OzoneTokenIdentifier extends
    * @return byte[]
    */
   public byte[] toUniqueSerializedKey() {
-    ByteBuffer result =
-        ByteBuffer.allocate(4096);
-    result.order(ByteOrder.BIG_ENDIAN);
+    DataOutputBuffer buf = new DataOutputBuffer();
     try {
-      result.putLong(getIssueDate());
-      result.putInt(getMasterKeyId());
-      result.putInt(getSequenceNumber());
-
-      result.putLong(getMaxDate());
-
-      result.putInt(getOwner().toString().length());
-      result.put(getOwner().toString().getBytes(StandardCharsets.UTF_8));
-
-      result.putInt(getRealUser().toString().length());
-      result.put(getRealUser().toString().getBytes(StandardCharsets.UTF_8));
-
-      result.putInt(getRenewer().toString().length());
-      result.put(getRenewer().toString().getBytes(StandardCharsets.UTF_8));
-
-      result.putInt(getTokenType().getNumber());
+      super.write(buf);
+      WritableUtils.writeVInt(buf, getTokenType().getNumber());
       // Set s3 specific fields.
       if (getTokenType().equals(S3AUTHINFO)) {
-        result.putInt(getAwsAccessId().length());
-        result.put(getAwsAccessId().getBytes(StandardCharsets.UTF_8));
-
-        result.putInt(getSignature().length());
-        result.put(getSignature().getBytes(StandardCharsets.UTF_8));
-
-        result.putInt(getStrToSign().length());
-        result.put(getStrToSign().getBytes(StandardCharsets.UTF_8));
+        WritableUtils.writeString(buf, getAwsAccessId());
+        WritableUtils.writeString(buf, getSignature());
+        WritableUtils.writeString(buf, getStrToSign());
       } else {
-        result.putInt(getOmCertSerialId().length());
-        result.put(getOmCertSerialId().getBytes(StandardCharsets.UTF_8));
+        WritableUtils.writeString(buf, getOmCertSerialId());
         if (getOmServiceId() != null) {
-          result.putInt(getOmServiceId().length());
-          result.put(getOmServiceId().getBytes(StandardCharsets.UTF_8));
-        } else {
-          result.putInt(0);
+          WritableUtils.writeString(buf, getOmServiceId());
         }
       }
-    } catch (IndexOutOfBoundsException e) {
+    } catch (java.io.IOException e) {
       throw new IllegalArgumentException(
           "Can't encode the the raw data ", e);
     }
-    return result.array();
+    return buf.getData();
   }
 
   /** Instead of relying on proto deserialization, this
    *  provides  explicit deserialization for OzoneTokenIdentifier.
    * @return byte[]
    */
-  public static OzoneTokenIdentifier fromUniqueSerializedKey(byte[] rawData) {
-    OzoneTokenIdentifier result = newInstance();
-    ByteBuffer inbuf = ByteBuffer.wrap(rawData);
-    inbuf.order(ByteOrder.BIG_ENDIAN);
-    result.setIssueDate(inbuf.getLong());
-    result.setMasterKeyId(inbuf.getInt());
-    result.setSequenceNumber(inbuf.getInt());
-
-    result.setMaxDate(inbuf.getLong());
-
-    int strsize = 0;
-    strsize = inbuf.getInt();
-    byte[] ownerBytes = new byte[strsize];
-    inbuf.get(ownerBytes);
-    result.setOwner(new Text(ownerBytes));
-
-    strsize = inbuf.getInt();
-    byte[] ruserBytes = new byte[strsize];
-    inbuf.get(ruserBytes);
-    result.setRealUser(new Text(ruserBytes));
-
-    strsize = inbuf.getInt();
-    byte[] renewerBytes = new byte[strsize];
-    inbuf.get(renewerBytes);
-    result.setRenewer(new Text(renewerBytes));
-
+  public OzoneTokenIdentifier fromUniqueSerializedKey(byte[] rawData)
+      throws IOException {
+    DataInputBuffer in = new DataInputBuffer();
+    in.reset(rawData, rawData.length);
+    super.readFields(in);
+    int type = WritableUtils.readVInt(in);
     // Set s3 specific fields.
-    if (inbuf.getInt() == S3AUTHINFO.getNumber()) {
-      strsize = inbuf.getInt();
-      byte[] awsAccessIdBytes = new byte[strsize];
-      inbuf.get(awsAccessIdBytes);
-      result.setAwsAccessId(new String(awsAccessIdBytes,
-          StandardCharsets.UTF_8));
-
-      strsize = inbuf.getInt();
-      byte[] signatureBytes = new byte[strsize];
-      inbuf.get(signatureBytes);
-      result.setSignature(new String(signatureBytes, StandardCharsets.UTF_8));
-
-      strsize = inbuf.getInt();
-      byte[] strToSignBytes = new byte[strsize];
-      inbuf.get(strToSignBytes);
-      result.setStrToSign(new String(strToSignBytes, StandardCharsets.UTF_8));
+    if (type == S3AUTHINFO.getNumber()) {
+      this.tokenType = Type.S3AUTHINFO;
+      setAwsAccessId(WritableUtils.readString(in));
+      setSignature(WritableUtils.readString(in));
+      setStrToSign(WritableUtils.readString(in));
     } else {
-      strsize = inbuf.getInt();
-      byte[] omCertIdBytes = new byte[strsize];
-      inbuf.get(omCertIdBytes);
-      result.setOmCertSerialId(new String(omCertIdBytes,
-          StandardCharsets.UTF_8));
-      int nextStrsize = inbuf.getInt();
-      if (nextStrsize != 0) {
-        byte[] omServiceIdBytes = new byte[nextStrsize];
-        inbuf.get(omServiceIdBytes);
-        result.setOmServiceId(new String(omServiceIdBytes,
-            StandardCharsets.UTF_8));
-      }
+      this.tokenType = Type.DELEGATION_TOKEN;
+      setOmCertSerialId(WritableUtils.readString(in));
+      setOmServiceId(WritableUtils.readString(in));
     }
-    return result;
+    return this;
   }
 
   /**
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
@@ -21,6 +21,9 @@ import java.io.DataInput;
 import java.io.DataInputStream;
 import java.io.DataOutput;
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 
 import org.apache.commons.lang3.builder.EqualsBuilder;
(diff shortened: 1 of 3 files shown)


hadoop-hdds/interface-client/src/main/proto/proto.lock
```diff
@@ -1476,6 +1476,21 @@
           }
         ],
         "messages": [
+          {
```
xiaoyuyao, line 1479: The proto change seems unrelated to this PR.

prashantpogde: It was auto generated. I did not modify anything here in this file.

prashantpogde: Integration test failure seems unrelated to the changes here.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
```diff
@@ -77,6 +79,119 @@ public Text getKind() {
     return KIND_NAME;
   }
 
+  /** Instead of relying on proto serialization, this
+   *  provides  explicit serialization for OzoneTokenIdentifier.
+   * @return byte[]
+   */
+  public byte[] toUniqueSerilizedKey() {
```
adoroszlai, line 86:
```suggestion
  public byte[] toUniqueSerializedKey() {
```

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
```diff
@@ -77,6 +79,119 @@ public Text getKind() {
...
+          "Can't encode the the raw data ", e);
+    }
+    return result.array();
+  }
+
+  /** Instead of relying on proto deserialization, this
+   *  provides  explicit deserialization for OzoneTokenIdentifier.
+   * @return byte[]
+   */
+  public static OzoneTokenIdentifier fromUniqueSerilizedKey(byte[] rawData) {
```
adoroszlai, line 138:
```suggestion
  public static OzoneTokenIdentifier fromUniqueSerializedKey(byte[] rawData) {
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/codec/TokenIdentifierCodec.java
```diff
@@ -33,7 +33,7 @@
   public byte[] toPersistedFormat(OzoneTokenIdentifier object) {
     Preconditions
         .checkNotNull(object, "Null object can't be converted to byte array.");
-    return object.getBytes();
+    return object.toUniqueSerilizedKey();
```
adoroszlai, line 36:
```suggestion
    return object.toUniqueSerializedKey();
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/codec/TokenIdentifierCodec.java
```diff
@@ -42,8 +42,8 @@ public OzoneTokenIdentifier fromPersistedFormat(byte[] rawData)
     Preconditions.checkNotNull(rawData,
         "Null byte array can't converted to real object.");
     try {
-      return OzoneTokenIdentifier.readProtoBuf(rawData);
-    } catch (InvalidProtocolBufferException e) {
+      return OzoneTokenIdentifier.fromUniqueSerilizedKey(rawData);
```
adoroszlai, line 45:
```suggestion
      return OzoneTokenIdentifier.fromUniqueSerializedKey(rawData);
```

hadoop-hdds/interface-client/src/main/proto/proto.lock
adoroszlai: > It was auto generated. I did not modify anything here in this file.

Please exclude both proto.lock files from the patch.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
```diff
@@ -77,6 +79,119 @@ public Text getKind() {
...
+    result.order(ByteOrder.BIG_ENDIAN);
+    try {
+      result.putLong(getIssueDate());
+      result.putInt(getMasterKeyId());
+      result.putInt(getSequenceNumber());
+
+      result.putLong(getMaxDate());
+
+      result.putInt(getOwner().toString().length());
+      result.put(getOwner().toString().getBytes());
```
adoroszlai, line 98: Findbugs:

```
H I Dm: Found reliance on default encoding in org.apache.hadoop.ozone.security.OzoneTokenIdentifier.toUniqueSerilizedKey(): String.getBytes()  At OzoneTokenIdentifier.java:[line 98]
```

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
```diff
@@ -77,6 +79,119 @@ public Text getKind() {
...
+    byte[] renewerBytes = new byte[strsize];
+    inbuf.get(renewerBytes);
+    result.setRenewer(new Text(renewerBytes));
+
+    // Set s3 specific fields.
+    if (inbuf.getInt() == S3AUTHINFO.getNumber()) {
+      strsize = inbuf.getInt();
+      byte[] awsAccessIdBytes = new byte[strsize];
+      inbuf.get(awsAccessIdBytes);
+      result.setAwsAccessId(new String(awsAccessIdBytes));
```
adoroszlai, line 169: Findbugs:

```
H I Dm: Found reliance on default encoding in org.apache.hadoop.ozone.security.OzoneTokenIdentifier.fromUniqueSerilizedKey(byte[]): new String(byte[])  At OzoneTokenIdentifier.java:[line 169]
```

hadoop-hdds/interface-client/src/main/proto/proto.lock
prashantpogde: > > It was auto generated. I did not modify anything here in this file.
> 
> Please exclude both proto.lock files from the patch.

done

prashantpogde: > The proto.lock change seems unrelated to this PR.

these files are reverted.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
prashantpogde: done

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
prashantpogde: done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/codec/TokenIdentifierCodec.java
prashantpogde: done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/codec/TokenIdentifierCodec.java
prashantpogde: done

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
prashantpogde: done

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
prashantpogde: done

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
```diff
@@ -77,6 +80,122 @@ public Text getKind() {
     return KIND_NAME;
   }
 
+  /** Instead of relying on proto serialization, this
+   *  provides  explicit serialization for OzoneTokenIdentifier.
+   * @return byte[]
+   */
+  public byte[] toUniqueSerializedKey() {
+    ByteBuffer result =
+        ByteBuffer.allocate(4096);
```
ChenSammi, line 89: I see there is StrToSign if it's S3 auth. I 'm curious about how big this StrToSign can be? And whether 4096 bytes can hold all these.

prashantpogde: > I see there is StrToSign if it's S3 auth. I 'm curious about how big this StrToSign can be? And whether 4096 bytes can hold all these.

It should not be too big and not cause overflow.

https://docs.aws.amazon.com/general/latest/gr/sigv4-create-string-to-sign.html

I saw something similar in other places too e.g. 
org/apache/hadoop/security/token/TokenIdentifier.class with 4096 bytes limit.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
```diff
@@ -77,6 +80,122 @@ public Text getKind() {
     return KIND_NAME;
   }
 
+  /** Instead of relying on proto serialization, this
+   *  provides  explicit serialization for OzoneTokenIdentifier.
+   * @return byte[]
+   */
+  public byte[] toUniqueSerializedKey() {
+    ByteBuffer result =
```
xiaoyuyao, line 88: Thanks @prashantpogde for working on this. The patch LGTM overall. Have a few questions:
Have you consider leverage the readFields from parent class to handle the basic non-protoc serialization which seems to be much simpler. 

```  
public byte[] toUniqueSerializedKey() throws IOException {
    DataOutputBuffer buf = new DataOutputBuffer();
    super.write(buf);
    WritableUtils.writeEnum(buf, getTokenType());
    // Set s3 specific fields.
    if (getTokenType().equals(S3AUTHINFO)) {
      WritableUtils.writeString(buf, getAwsAccessId());
      WritableUtils.writeString(buf, getSignature());
      WritableUtils.writeString(buf, getStrToSign());
    } else {
      WritableUtils.writeString(buf, getOmCertSerialId());
    }
    return buf.getData();
  }
```

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
```diff
@@ -77,6 +80,122 @@ public Text getKind() {
...
+        ByteBuffer.allocate(4096);
+    result.order(ByteOrder.BIG_ENDIAN);
+    try {
+      result.putLong(getIssueDate());
+      result.putInt(getMasterKeyId());
+      result.putInt(getSequenceNumber());
+
+      result.putLong(getMaxDate());
+
+      result.putInt(getOwner().toString().length());
```
xiaoyuyao, line 98: NIT: The Text class is a UTF-8 bytes + length wrapper class. You can use getLength()/getBytes() without toString() conversion. Same apply to owner, realuser and renewer.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
```diff
@@ -77,6 +80,122 @@ public Text getKind() {
...
+    }
+    return result.array();
+  }
+
+  /** Instead of relying on proto deserialization, this
+   *  provides  explicit deserialization for OzoneTokenIdentifier.
+   * @return byte[]
+   */
+  public static OzoneTokenIdentifier fromUniqueSerializedKey(byte[] rawData) {
+    OzoneTokenIdentifier result = newInstance();
```
xiaoyuyao, line 140: We could simplify the code by leveraging the readFields from parent class to deserialize non-protoc token id.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
prashantpogde: didn't try this. This looks simpler. Let me try this.

prashantpogde: I wanted to do explicit serialization. But this also looks safe serialization.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
prashantpogde: this will change after I use super.write()

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
prashantpogde: Earlier I wanted to do explicit serialization, but your suggestion looks simpler. I will try this.

prashantpogde: I have uploaded the changes as suggested by @xiaoyuyao Please take a look..

prashantpogde: One failure in coverage check looks unreleted.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
```diff
@@ -77,6 +81,57 @@ public Text getKind() {
...
+      super.write(buf);
+      WritableUtils.writeVInt(buf, getTokenType().getNumber());
+      // Set s3 specific fields.
+      if (getTokenType().equals(S3AUTHINFO)) {
+        WritableUtils.writeString(buf, getAwsAccessId());
+        WritableUtils.writeString(buf, getSignature());
+        WritableUtils.writeString(buf, getStrToSign());
+      } else {
+        WritableUtils.writeString(buf, getOmCertSerialId());
+        if (getOmServiceId() != null) {
```
xiaoyuyao, line 100: Thanks @prashantpogde  for the update. LGTM overall. Only one question: the null check on line100 should be removed. WritableUtils.writeString will write len=-1 for null and the WritableUtils.readString will populate null back during de-serialization. This way, we will have matched deserialization code on the code in line129-130. 

+1 after that being fixed.

xiaoyuyao:
+1 pending CI.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/OzoneTokenIdentifier.java
avijayanhwx: @ChenSammi Are you OK with resolving this issue?

avijayanhwx: This issue is no longer valid in the newer version of the patch. cc @ChenSammi.

avijayanhwx:
Please don't merge this since we are testing out upgrade paths.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/codec/TokenIdentifierCodec.java
```diff
@@ -42,8 +42,9 @@ public OzoneTokenIdentifier fromPersistedFormat(byte[] rawData)
     Preconditions.checkNotNull(rawData,
         "Null byte array can't converted to real object.");
     try {
-      return OzoneTokenIdentifier.readProtoBuf(rawData);
-    } catch (InvalidProtocolBufferException e) {
+      OzoneTokenIdentifier object = OzoneTokenIdentifier.newInstance();
+      return object.fromUniqueSerializedKey(rawData);
+    } catch (BufferUnderflowException e) {
```
xiaoyuyao, line 47: We should catch IOException here and fall back to reading the old format to avoid OM failures to load tokens in old format upon upgrade/restart. We can rewrite the loaded token with the new format like HDDS-3925 upon hitting tokens in old format to avoid future fallbacks. But I'm OK without it given old tokens in old format will be expired in 7 days anyway (unlike the pipeline id) and the new tokens will be written in new format. 

I also suggest we add a unit test like below that can repro the loading failure cases in TestOzoneTokenIdentifier.java. 

  @Test
  public void testTokenPersistence() throws IOException {
    OzoneTokenIdentifier idWrite = getIdentifierInst();
    idWrite.setOmServiceId("defaultServiceId");

    byte[] oldIdBytes = idWrite.getBytes();
    TokenIdentifierCodec idCodec = new TokenIdentifierCodec();

    OzoneTokenIdentifier idRead = null;
    try {
      idRead =  idCodec.fromPersistedFormat(oldIdBytes);
    } catch (IOException ex) {
      Assert.fail("Should not fail to load old token format");
    }
    Assert.assertEquals("Deserialize Serialized Token should equal.",
        idWrite, idRead);
  }
}

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/codec/TokenIdentifierCodec.java
```diff
@@ -42,8 +42,9 @@ public OzoneTokenIdentifier fromPersistedFormat(byte[] rawData)
     Preconditions.checkNotNull(rawData,
         "Null byte array can't converted to real object.");
     try {
-      return OzoneTokenIdentifier.readProtoBuf(rawData);
-    } catch (InvalidProtocolBufferException e) {
+      OzoneTokenIdentifier object = OzoneTokenIdentifier.newInstance();
```
xiaoyuyao, line 45:
```suggestion
      } catch (IOException ex) {
      return OzoneTokenIdentifier.readProtoBuf(rawData);
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/codec/TokenIdentifierCodec.java
xiaoyuyao, line 47:
```suggestion
    } catch (IOException ex) {
      return OzoneTokenIdentifier.readProtoBuf(rawData);
    } catch (BufferUnderflowException e) {
```

prashantpogde: done

prashantpogde: done

prashantpogde:
Made the changes as suggested by @xiaoyuyao .

PR #1185, opened 2020-07-10
HDDS-3933. Fix memory leak because of too many Datanode State Machine Thread

runzhiwang:

**What's problem ?**
Datanode creates more than 20K Datanode State Machine Thread, then OOM happened.
image

**What's the reason ?**
20K Datanode State Machine Thread were created by newCachedThreadPool
image

Almost all of them were wait lock.
image

Only one Datanode State Machine Thread got the lock, and block when submitRequest. Because this thread was blocked and can not free the lock, newCachedThreadPool will create new thread infinitely.
image

**How to fix ?**
1. Avoid use `newCachedThreadPool`, because it will create new thread infinitely, if no thread available in pool.
2. Cancel future when task time out.
3. Besides there is bug in such following code which want to loop until timeout or returned >= count. 
    But timeout can not work, for example when init startTime = 0 and timeLeft = 100, if monotonicNow = 10, timeLeft = 100 - (10 - 0) = 90, and then if monotonicNow = 11, timeLeft = 90 - (11 - 0) = 79 which actually should be 89.
```
 while (returned < count && timeLeft > 0) {
    timeLeft = timeLeft - (Time.monotonicNow() - startTime);
 }
```

https://issues.apache.org/jira/browse/HDDS-3933

## How was this patch tested?

Existed UT.

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -269,13 +269,6 @@
       defined with postfix (ns,ms,s,m,h,d)</description>
   </property>
   <property>
-    <name>hdds.statemachine.endpoint.task.thread.count</name>
-    <value>2</value>
-    <tag>OZONE, DATANODE, MANAGEMENT</tag>
-    <description>Maximum number of threads in the thread pool that Datanode
-      will use for GETVERSION, REGISTER, HEARTBEAT to SCMs.</description>
-  </property>
-  <property>
     <name>hdds.container.report.interval</name>
     <value>60000ms</value>
     <tag>OZONE, CONTAINER, MANAGEMENT</tag>
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -269,6 +269,13 @@
       defined with postfix (ns,ms,s,m,h,d)</description>
   </property>
   <property>
+    <name>hdds.statemachine.endpoint.task.thread.count</name>
+    <value>2</value>
+    <tag>OZONE, DATANODE, MANAGEMENT</tag>
+    <description>Maximum number of threads in the thread pool that Datanode
+      will use for GETVERSION, REGISTER, HEARTBEAT to SCMs.</description>
+  </property>
+  <property>
     <name>hdds.container.report.interval</name>
     <value>60000ms</value>
     <tag>OZONE, CONTAINER, MANAGEMENT</tag>
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
@@ -28,6 +28,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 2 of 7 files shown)


runzhiwang:
I will check the failed ut.

ChenSammi:
Don't find any unit tests here. Would you please add some?  @runzhiwang

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -268,6 +268,13 @@
       datanode periodically send node report to SCM. Unit could be
       defined with postfix (ns,ms,s,m,h,d)</description>
   </property>
+  <property>
+    <name>hdds.statemachine.endpoint.task.thread.count</name>
+    <value>2</value>
+    <tag>OZONE, DATANODE, MANAGEMENT</tag>
+    <description>Maximum number of threads in the thread pool that Datanode
+      will use for GETVERSION, REGISTER, HEARTBEAT to SCMs.</description>
+  </property>
```
ChenSammi, line 277: Can we add more details about what's the proper value for this proerpty if user want to change it?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
```diff
@@ -200,22 +203,31 @@ public void execute(ExecutorService executor) {
...
+        }
       }
-      timeLeft = timeLeft - (Time.monotonicNow() - startTime);
+
+      Thread.sleep(durationMS / 10);
     }
+
+    for (Future<EndPointStates> future : futures) {
+      if (future != null && !future.isDone()) {
+        future.cancel(true);
```
ChenSammi, line 227: Instead of cancel the future, I would suggest to skip the next time execution, and let the current future finish it.  Also, please add LOG if this case happens.

runzhiwang: Maybe the future can not finish forever, because the rpc call blocked forever, as we found in production cluster.

hadoop-hdds/common/src/main/resources/ozone-default.xml
runzhiwang: I remove this config, and get thread pool size by add number of scm and recon.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
```diff
@@ -415,7 +431,14 @@ public void execute(ExecutorService service, long time, TimeUnit unit)
       if (this.isEntering()) {
         task.onEnter();
       }
-      task.execute(service);
+
+      if (isThreadPoolAvailable(service)) {
+        task.execute(service);
+      } else {
+        LOG.warn("No available thread in pool, duration:" +
```
xiaoyuyao, line 438: In a similar case of endpoint task locked up, will we end up with flooding of warning logs here?

runzhiwang: @xiaoyuyao Thanks for review and reasonable suggestion. How about add metrics for thread pool available and unavailable times ?

ChenSammi: There is a  EndpointStateMachine#logIfNeeded function.   We can leverage this function to reduce the same warning logs logged.  @runzhiwang

runzhiwang: @ChenSammi @xiaoyuyao  Thanks for review. I have updated the PR, because  EndpointStateMachine#logIfNeeded can not be accessed there, and it's can not be refactored to use, so I did not use it.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
```diff
@@ -434,9 +438,14 @@ public void execute(ExecutorService service, long time, TimeUnit unit)
 
       if (isThreadPoolAvailable(service)) {
         task.execute(service);
+        threadPoolNotAvailableCount.set(0);
       } else {
-        LOG.warn("No available thread in pool, duration:" +
-            unit.toMillis(time));
+        threadPoolNotAvailableCount.incrementAndGet();
```
ChenSammi, line 443: Can we switch theset two statement?

runzhiwang: Updated

ChenSammi:
LGTM + 1.  Thanks @runzhiwang for the contribution and @xiaoyuyao for the review.

PR #1186, opened 2020-07-10
HDDS-3941. Enable core dump when crash in C++

runzhiwang:

**What's the problem ?**
This PR is related to HDDS-3933.

When memory leak because of the reason explained in HDDS-3933, Datanode most time generates core.pid because it crash in Rocksdb when create new thread, as the image shows, but generates crash log rarely.
image

But because the default value of `core file size` is zero, so core.pid can not be generated. So when Datanode crash in Rocksdb, we can not get any information about why it crashed. 

**How to fix ?**
Set `ulimit -c unlimited` to enable core dump when crash in RocksDB.

https://issues.apache.org/jira/browse/HDDS-3941

## How was this patch tested?

Existed UT.


ChenSammi:
The patch looks good to me. + 1.

PR #1187, opened 2020-07-10
HDDS-3951. Rename the num.write.chunk.thread key.

maobaolong:

Reanme `dfs.container.ratis.num.write.chunk.thread` to `dfs.container.ratis.num.write.chunk.thread.per.disk `

to clearly express the meaning of this key, if it configured to 10, then the thread num would be 10 * numOfDisk. 

HDDS-3951

## How was this patch tested?

NO Need, just rename config key.


ChenSammi:
@maobaolong,  suggest change from " per disk" to " per volume" to be consistent with other datanode property naming style.

maobaolong:
@ChenSammi Thanks for your suggestion，I have addressed your comment，PTAL

xiaoyuyao:
LGTM, +1. Thanks @maobaolong  for the contribution and all for the reviews.

maobaolong:
@ChenSammi @xiaoyuyao Thank you for your review.

PR #1191, opened 2020-07-10
HDDS-3837 Add isLeader check in SCMHAManager.

timmylicheng:
Add isLeader check in SCMHAManager.

https://issues.apache.org/jira/browse/HDDS-3837

## How was this patch tested?
UT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAManagerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAManagerImpl.java
@@ -17,7 +17,17 @@
 
 package org.apache.hadoop.hdds.scm.ha;
 
+import com.google.common.base.Preconditions;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
+import org.apache.ratis.protocol.NotLeaderException;
+import org.apache.ratis.protocol.RaftGroupMemberId;
+import org.apache.ratis.protocol.RaftPeer;
+import org.apache.ratis.protocol.RaftPeerId;
+import org.apache.ratis.server.RaftServer;
+import org.apache.ratis.server.impl.RaftServerImpl;
+import org.apache.ratis.server.impl.RaftServerProxy;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 
@@ -31,14 +41,17 @@ import java.io.IOException;
  */
 public class SCMHAManagerImpl implements SCMHAManager {
 
-  private static boolean isLeader = true;
+  private static final Logger LOG =
+      LoggerFactory.getLogger(SCMHAManagerImpl.class);
 
   private final SCMRatisServerImpl ratisServer;
+  private final ConfigurationSource conf;
 
   /**
    * Creates SCMHAManager instance.
    */
   public SCMHAManagerImpl(final ConfigurationSource conf) throws IOException {
+    this.conf = conf;
     this.ratisServer = new SCMRatisServerImpl(
         conf.getObject(SCMHAConfiguration.class), conf);
   }
@@ -56,7 +69,28 @@ public class SCMHAManagerImpl implements SCMHAManager {
    */
   @Override
   public boolean isLeader() {
-    return isLeader;
+    if (!SCMHAUtils.isSCMHAEnabled(conf)) {
+      // When SCM HA is not enabled, the current SCM is always the leader.
+      return true;
+    }
+    RaftServer server = ratisServer.getServer();
+    Preconditions.checkState(server instanceof RaftServerProxy);
+    RaftServerImpl serverImpl = null;
+    try {
+      // SCM only has one raft group.
+      serverImpl = ((RaftServerProxy) server)
+          .getImpl(ratisServer.getRaftGroupId());
+    } catch (IOException ioe) {
+      LOG.error("Fail to get RaftServer impl and therefore it's not clear " +
+          "whether it's leader. ", ioe);
+    } finally {
+      if (serverImpl != null) {
+        // Only when it's sure the current SCM is the leader, otherwise
+        // it should all return false.
+        return serverImpl.isLeader();
+      }
+    }
+    return false;
   }
 
   /**
@@ -67,6 +101,44 @@ public class SCMHAManagerImpl implements SCMHAManager {
     return ratisServer;
   }
 
+  private RaftPeerId getPeerIdFromRoleInfo(RaftServerImpl serverImpl) {
+    if (serverImpl.isLeader()) {
+      return RaftPeerId.getRaftPeerId(
+          serverImpl.getRoleInfoProto().getLeaderInfo().toString());
+    } else if (serverImpl.isFollower()) {
+      return RaftPeerId.getRaftPeerId(
+          serverImpl.getRoleInfoProto().getFollowerInfo()
+              .getLeaderInfo().getId().getAddress());
+    } else {
+      return null;
+    }
+  }
+
+  @Override
+  public RaftPeer getSuggestedLeader() {
+    RaftServer server = ratisServer.getServer();
+    Preconditions.checkState(server instanceof RaftServerProxy);
+    RaftServerImpl serverImpl = null;
+    try {
+      // SCM only has one raft group.
+      serverImpl = ((RaftServerProxy) server)
+          .getImpl(ratisServer.getRaftGroupId());
+    } catch (IOException ioe) {
+      LOG.error("Fail to get RaftServer impl and therefore it's not clear " +
+          "whether it's leader. ", ioe);
+    } finally {
+      if (serverImpl != null) {
+        RaftPeerId peerId =  getPeerIdFromRoleInfo(serverImpl);
+        if (peerId != null) {
+          return new RaftPeer(peerId);
+        } else {
+          return null;
+        }
+      }
+    }
+    return null;
+  }
+
   /**
    * {@inheritDoc}
    */
@@ -75,4 +147,15 @@ public class SCMHAManagerImpl implements SCMHAManager {
     ratisServer.stop();
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public NotLeaderException triggerNotLeaderException() {
+    return new NotLeaderException(RaftGroupMemberId.valueOf(
+        ratisServer.getServer().getId(),
+        ratisServer.getRaftGroupId()),
+        getSuggestedLeader(),
+        ratisServer.getRaftPeers());
+  }
 }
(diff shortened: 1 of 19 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
```diff
@@ -327,6 +349,9 @@ public void openPipeline(PipelineID pipelineId) throws IOException {
    * @throws IOException
    */
   protected void removePipeline(Pipeline pipeline) throws IOException {
+    if (!scmhaManager.isLeader()) {
```
GlenGeng, line 352: Would it be cleaner by using `InvocationHander` to inject these `isLeader()` check ?

```
StateManager pipelineStateManager = (StateManager) Proxy.newProxyInstance(
    null,
    new Class<?>[] {StateManager.class}, 
    (proxy, method, args) -> {
      if (method.getName().equals("xxx")) {
        // do is leader check;
        // throw exception if needed;
      }
      return method.invoke(args);
    });
```

You may use `hashmap` or `annotation` to speed up the method name check.
Just a recommendation.

timmylicheng: If we want to refactor the code, I would prefer annotation just like @Replicate in which we would create something like CheckLeaderInvocationHandler extends InvocationHandler and do isLeader check there.
It would be convenient in interface. However, We also use isLeader check in some local methods instead of override methods. Let's keep it this fashion for now and we see if we can polish this.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAManagerImpl.java
```diff
@@ -56,7 +69,28 @@ public void start() throws IOException {
    */
   @Override
   public boolean isLeader() {
-    return isLeader;
+    if (!SCMHAUtils.isSCMHAEnabled(conf)) {
+      // When SCM HA is not enabled, the current SCM is always the leader.
+      return true;
+    }
```
GlenGeng, line 75: Add a todo please
```
// will change from RaftServerImpl:::isLeader() to RaftServerImpl::getRoleInfoProto()
// after RATIS-1001 is merged.
```

timmylicheng: https://issues.apache.org/jira/browse/HDDS-3962 
Created a JIRA to track this.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAManagerImpl.java
```diff
@@ -56,7 +69,28 @@ public void start() throws IOException {
...
+    RaftServerImpl serverImpl = null;
+    try {
+      // SCM only has one raft group.
+      serverImpl = ((RaftServerProxy) server)
+          .getImpl(ratisServer.getRaftGroupId());
+    } catch (IOException ioe) {
+      LOG.error("Fail to get RaftServer impl and therefore it's not clear " +
+          "whether it's leader. ", ioe);
+    } finally {
+      if (serverImpl != null) {
```
xiaoyuyao, line 87: should we move this to the try block?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAManagerImpl.java
```diff
@@ -67,6 +101,44 @@ public SCMRatisServer getRatisServer() {
...
   }
 
+  private RaftPeerId getPeerIdFromRoleInfo(RaftServerImpl serverImpl) {
+    if (serverImpl.isLeader()) {
+      return RaftPeerId.getRaftPeerId(
+          serverImpl.getRoleInfoProto().getLeaderInfo().toString());
+    } else if (serverImpl.isFollower()) {
+      return RaftPeerId.getRaftPeerId(
+          serverImpl.getRoleInfoProto().getFollowerInfo()
+              .getLeaderInfo().getId().getAddress());
```
xiaoyuyao, line 111: getId().getAddress() seems unnecessary.

timmylicheng: I update with what OM does.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAManagerImpl.java
timmylicheng: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAManagerImpl.java
```diff
@@ -67,6 +101,44 @@ public SCMRatisServer getRatisServer() {
...
+    RaftServerImpl serverImpl = null;
+    try {
+      // SCM only has one raft group.
+      serverImpl = ((RaftServerProxy) server)
+          .getImpl(ratisServer.getRaftGroupId());
+    } catch (IOException ioe) {
+      LOG.error("Fail to get RaftServer impl and therefore it's not clear " +
+          "whether it's leader. ", ioe);
+    } finally {
+      if (serverImpl != null) {
```
xiaoyuyao, line 130: Same as above, can we move this to the try block?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -110,4 +111,18 @@ public void stop() throws IOException {
...
+  }
+
+  @Override
+  public RaftGroupId getRaftGroupId() {
+    return raftGroupId;
+  }
+
+  @Override
+  public List<RaftPeer> getRaftPeers() {
+    return Collections.singletonList(new RaftPeer(raftPeerId));
```
xiaoyuyao, line 126: why ScmRatiSserver only have one entry in the raft peers, is this something we plan to implement in future?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NewNodeHandler.java
```diff
@@ -41,6 +46,11 @@ public NewNodeHandler(PipelineManager pipelineManager,
   @Override
   public void onMessage(DatanodeDetails datanodeDetails,
       EventPublisher publisher) {
-    pipelineManager.triggerPipelineCreation();
+    try {
+      pipelineManager.triggerPipelineCreation();
+    } catch (NotLeaderException ex) {
+      LOG.warn("Not the current leader SCM and cannot start pipeline" +
```
xiaoyuyao, line 52: Can we put this as debug log to avoid spamming the SCM log?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NonHealthyToHealthyNodeHandler.java
```diff
@@ -42,6 +47,11 @@ public NonHealthyToHealthyNodeHandler(
   @Override
   public void onMessage(DatanodeDetails datanodeDetails,
       EventPublisher publisher) {
-    pipelineManager.triggerPipelineCreation();
+    try {
+      pipelineManager.triggerPipelineCreation();
+    } catch (NotLeaderException ex) {
+      LOG.warn("Not the current leader SCM and cannot start pipeline" +
```
xiaoyuyao, line 53: Same as above.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
```diff
@@ -136,6 +139,9 @@ public static PipelineManagerV2Impl newPipelineManager(
   @Override
   public Pipeline createPipeline(ReplicationType type,
                                  ReplicationFactor factor) throws IOException {
+    if (!scmhaManager.isLeader()) {
```
xiaoyuyao, line 142: NIT: Can we add a helper function like checkLeader() to dedup the code?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
```diff
@@ -431,15 +464,21 @@ public void scrubPipeline(ReplicationType type, ReplicationFactor factor)
    * Schedules a fixed interval job to create pipelines.
    */
   @Override
-  public void startPipelineCreator() {
+  public void startPipelineCreator() throws NotLeaderException {
+    if (!scmhaManager.isLeader()) {
```
xiaoyuyao, line 468: when scm leader changes, how to ensure the creator and scrubber threads are started on the new leader and stopped on the old leader?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
timmylicheng: Yes. And this class will be deprecated soon in SCM group feature.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
timmylicheng: My thought:
1. Every SCM starts a pipeline creator thread. So I remove checkLeader in startPipelineCreator
2. For triggerPipelineCreator, if we decide to let follower go thru safemode once it steps up to be leader, we can have checkLeader there as well.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
timmylicheng: Updated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NonHealthyToHealthyNodeHandler.java
timmylicheng: Updated

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NewNodeHandler.java
timmylicheng: Updated

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAManagerImpl.java
timmylicheng: Updated

xiaoyuyao:
+1, Thanks @timmylicheng  for the patience.

PR #1193, opened 2020-07-10
HDDS-3953. Audit LOG action misused in OzoneManager.commitMultipartUp…

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-3953


ChenSammi:
> 
> 
> This new code is not being used. This is dead code. There is a Jira to clean up the old code.
> 
> In used code, I see it is having correctly. [#codelink](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCommitPartRequest.java#L240)

@bharatviswa504 , do you konw the JIRA number to clean up this part of code?  I find HDDS-3505 which aims to remove bucket related operation.

bharatviswa504:
This is the parent Jira https://issues.apache.org/jira/browse/HDDS-2353 for the cleanup.

ChenSammi:
Thanks @bharatviswa504 . Close this MR.

PR #1194, opened 2020-07-10
HDDS-3062. Fix TestOzoneRpcClientAbstract.testListVolume.

lokeshj1703:

TestOzoneRpcClientAbstract.testListVolume is disabled due to intermittent issues.

https://issues.apache.org/jira/browse/HDDS-3062

## How was this patch tested?

Fixes UT


lokeshj1703:
@adoroszlai Thanks for reviewing the PR! I agree. The appendEntries timeout issue will affect all the ratis client related tests that is why we see failures in TestWatchForCommit, TestBlockDeletion etc. But I think we can still enable tests which do not use ratis client or the tests not involved in write path. The test enabled in the PR creates a bunch of volumes and then checks if listVolumes works well. What do you think?

adoroszlai:
Thanks @lokeshj1703 for the fix.

PR #1195, opened 2020-07-10
HDDS-3930. Fix OMKeyDeletesRequest.

bharatviswa504:

1. The cache of the key should be updated in ValidateAndUpdateCache, as we return response once after adding to cache, and before DoubleBuffer flushes to disk using OmClientResponse#addToDBBatch.
2. Changed return type of DeleteKeysResponse, as anyway when success we are not returning delete list and also when failure, we fail complete batch. So we can just return status.

https://issues.apache.org/jira/browse/HDDS-3930

## How was this patch tested?
Added tests


xiaoyuyao:
Thanks @bharatviswa504  for the update. LGTM, +1 from me.

bharatviswa504:
I have addressed review comments.

bharatviswa504:
Thank You @adoroszlai @xiaoyuyao and @smengcl for the review.

PR #1196, opened 2020-07-10
HDDS-3955. Unable to list intermediate paths on keys created using S3G.

bharatviswa504:

Keys created via the S3 Gateway currently use the createKey OM API to create the ozone key. Hence, when using a hdfs client to list intermediate directories in the key, OM returns key not found error. This was encountered while using fluentd to write Hive logs to Ozone via the s3 gateway.

Added a new config in OM to control the behavior of key create to create intermediate directories or not. In this way, keys created via ozone S3G can be accessed through fs API.

https://issues.apache.org/jira/browse/HDDS-3955

## How was this patch tested?

Added UT and IT.

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2483,9 +2483,10 @@
     <name>ozone.om.enable.filesystem.paths</name>
     <tag>OZONE, OM</tag>
     <value>false</value>
-    <description>If true, OM will normalize the key paths and also create
-      intermediate directories. This flag will be helpful when objects
-      created by S3G need to be accessed using OFS/O3Fs.
+    <description>If true, key names will be interpreted as file system paths.
+      "/" will be treated as a special character and paths will be normalized
+      and must follow Unix filesystem path naming conventions. This flag will
+      be helpful when objects created by S3G need to be accessed using OFS/O3Fs.
       If false, it will fallback to default behavior of Key/MPU create
       requests where key paths are not normalized and any intermediate
       directories will not be created or any file checks happens to check
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2480,15 +2480,16 @@
   </property>
 
   <property>
-    <name>ozone.om.create.intermediate.directories</name>
+    <name>ozone.om.enable.filesystem.paths</name>
     <tag>OZONE, OM</tag>
     <value>false</value>
-    <description>If true, then intermediate directories will be created
-      during Key/MPU create requests. This flag will be helpful when objects
+    <description>If true, OM will normalize the key paths and also create
+      intermediate directories. This flag will be helpful when objects
       created by S3G need to be accessed using OFS/O3Fs.
       If false, it will fallback to default behavior of Key/MPU create
-      requests where any intermediate directories will not be created or any
-      file checks happens to check filesystem semantics.
+      requests where key paths are not normalized and any intermediate
+      directories will not be created or any file checks happens to check
+      filesystem semantics.
     </description>
   </property>
 </configuration>
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2478,4 +2478,17 @@
       rules in Amazon S3's object key naming guide.
     </description>
   </property>
+
+  <property>
+    <name>ozone.om.create.intermediate.directories</name>
+    <tag>OZONE, OM</tag>
+    <value>false</value>
+    <description>If true, then intermediate directories will be created
+      during Key/MPU create requests. This flag will be helpful when objects
+      created by S3G need to be accessed using OFS/O3Fs.
+      If false, it will fallback to default behavior of Key/MPU create
+      requests where any intermediate directories will not be created or any
+      file checks happens to check filesystem semantics.
+    </description>
+  </property>
 </configuration>
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -242,8 +242,8 @@ public final class OMConfigKeys {
 
   // This config needs to be enabled, when S3G created objects used via
   // FileSystem API.
-  public static final String OZONE_OM_CREATE_INTERMEDIATE_DIRECTORY =
-      "ozone.om.create.intermediate.directories";
-  public static final boolean OZONE_OM_CREATE_INTERMEDIATE_DIRECTORY_DEFAULT =
+  public static final String OZONE_OM_ENABLE_FILESYSTEM_PATHS =
+      "ozone.om.enable.filesystem.paths";
+  public static final boolean OZONE_OM_ENABLE_FILESYSTEM_PATHS_DEFAULT =
       false;
 }
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -239,4 +239,11 @@ public final class OMConfigKeys {
           "ozone.om.keyname.character.check.enabled";
   public static final boolean OZONE_OM_KEYNAME_CHARACTER_CHECK_ENABLED_DEFAULT =
           false;
+
+  // This config needs to be enabled, when S3G created objects will be used
+  // FileSystem.
+  public static final String OZONE_OM_CREATE_INTERMEDIATE_DIRECTORY =
+      "ozone.om.create.intermediate.directories";
+  public static final boolean OZONE_OM_CREATE_INTERMEDIATE_DIRECTORY_DEFAULT =
+      false;
 }
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -240,8 +240,8 @@ public final class OMConfigKeys {
   public static final boolean OZONE_OM_KEYNAME_CHARACTER_CHECK_ENABLED_DEFAULT =
           false;
 
-  // This config needs to be enabled, when S3G created objects will be used
-  // FileSystem.
+  // This config needs to be enabled, when S3G created objects used via
+  // FileSystem API.
   public static final String OZONE_OM_CREATE_INTERMEDIATE_DIRECTORY =
       "ozone.om.create.intermediate.directories";
   public static final boolean OZONE_OM_CREATE_INTERMEDIATE_DIRECTORY_DEFAULT =
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
@@ -38,10 +38,13 @@ import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMRespo
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer;
 import org.apache.hadoop.ozone.security.acl.OzoneObj;
 import org.apache.hadoop.security.UserGroupInformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.net.InetAddress;
+import java.nio.file.Paths;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -52,6 +55,8 @@ import java.util.Map;
  */
 public abstract class OMClientRequest implements RequestAuditor {
 
+  private static final Logger LOG =
+      LoggerFactory.getLogger(OMClientRequest.class);
   private OMRequest omRequest;
 
   /**
@@ -264,5 +269,23 @@ public abstract class OMClientRequest implements RequestAuditor {
     Map<String, String> auditMap = new LinkedHashMap<>();
     auditMap.put(OzoneConsts.VOLUME, volume);
     return auditMap;
+  }
+
+  public static String getNormalizedKey(boolean enableFileSystemPaths,
+      String keyName) {
+    if (enableFileSystemPaths) {
+      String normalizedKeyName;
+      if (keyName.startsWith("/")) {
+        normalizedKeyName = Paths.get(keyName).toUri().normalize().getPath();
+      } else {
+        normalizedKeyName = Paths.get("/", keyName).toUri().normalize().getPath();
+      }
+      if (!keyName.equals(normalizedKeyName)) {
+        LOG.debug("Normalized key {} to {} ", keyName, normalizedKeyName);
+      }
+      return normalizedKeyName.substring(1);
+    } else {
+      return keyName;
+    }
   }
 }
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
@@ -271,6 +271,7 @@ public abstract class OMClientRequest implements RequestAuditor {
     return auditMap;
   }
 
+  @SuppressWarnings("HardcodedFileSeparator")
   public static String getNormalizedKey(boolean enableFileSystemPaths,
       String keyName) {
     if (enableFileSystemPaths) {
@@ -278,7 +279,8 @@ public abstract class OMClientRequest implements RequestAuditor {
       if (keyName.startsWith("/")) {
         normalizedKeyName = Paths.get(keyName).toUri().normalize().getPath();
       } else {
-        normalizedKeyName = Paths.get("/", keyName).toUri().normalize().getPath();
+        normalizedKeyName = Paths.get("/", keyName).toUri()
+            .normalize().getPath();
       }
       if (!keyName.equals(normalizedKeyName)) {
         LOG.debug("Normalized key {} to {} ", keyName, normalizedKeyName);
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
@@ -50,6 +50,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 
 import static org.apache.hadoop.ozone.OzoneConsts.OM_KEY_PREFIX;
+import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.INVALID_KEY_NAME;
 
 /**
  * OMClientRequest provides methods which every write OM request should
@@ -274,17 +275,18 @@ public abstract class OMClientRequest implements RequestAuditor {
   }
 
 
-  public static String getNormalizedKey(boolean enableFileSystemPaths,
-      String keyName) {
+  public static String validateAndNormalizeKey(boolean enableFileSystemPaths,
+      String keyName) throws OMException {
     if (enableFileSystemPaths) {
-      return getNormalizedKey(keyName);
+      return validateAndNormalizeKey(keyName);
     } else {
       return keyName;
     }
   }
 
   @SuppressFBWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
-  public static String getNormalizedKey(String keyName) {
+  public static String validateAndNormalizeKey(String keyName)
+      throws OMException {
     String normalizedKeyName;
     if (keyName.startsWith(OM_KEY_PREFIX)) {
       normalizedKeyName = Paths.get(keyName).toUri().normalize().getPath();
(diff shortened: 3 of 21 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
```diff
@@ -239,4 +239,11 @@ private OMConfigKeys() {
           "ozone.om.keyname.character.check.enabled";
   public static final boolean OZONE_OM_KEYNAME_CHARACTER_CHECK_ENABLED_DEFAULT =
           false;
+
+  // This config needs to be enabled, when S3G created objects will be used
+  // FileSystem.
```
arp7, line 244: used via FileSystem API.

arp7:
Can you update the PR template a bit to describe the high-level approach (was fix made in OM or S3G)? 🙂

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
```diff
@@ -221,8 +233,45 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+        OMFileRequest.OMPathInfo pathInfo =
+            OMFileRequest.verifyFilesInPath(omMetadataManager, volumeName,
+                bucketName, keyName, Paths.get(keyName));
+        OMFileRequest.OMDirectoryResult omDirectoryResult =
+            pathInfo.getDirectoryResult();
+        inheritAcls = pathInfo.getAcls();
+
+        // Check if a file or directory exists with same key name.
+        if (omDirectoryResult == DIRECTORY_EXISTS) {
+          throw new OMException("Can not write to directory: " + keyName,
```
arp7, line 253: Let's make the exception message slightly more descriptive to explain what is really going on. We can mention that createIntermediateDirs behavior is enabled and hence `/` has special interpretation.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
```diff
@@ -221,8 +233,45 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+          if (omDirectoryResult == FILE_EXISTS_IN_GIVENPATH) {
+            throw new OMException("Can not create file: " + keyName +
+                " as there is already file in the given path", NOT_A_FILE);
+          }
+
+        missingParentInfos = OMDirectoryCreateRequest
+            .getAllParentInfo(ozoneManager, keyArgs,
+                pathInfo.getMissingParents(), inheritAcls, trxnLogIndex);
+
+        // Add cache entries for the prefix directories.
```
arp7, line 265: @bharatviswa504 , one q. We are adding cache entries right away instead of at the end on success. Could this be a potential issue on the failure path?

arp7:
The approach looks really good. It is surprisingly concise and elegant because I was expecting it to be a lot more code! Added a few review comments.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyCreateRequest.java
```diff
@@ -328,6 +342,73 @@ private OMRequest createKeyRequest(boolean isMultipartKey, int partNumber) {
...
+        100L, ozoneManagerDoubleBufferHelper);
+
+    Assert.assertEquals(omClientResponse.getOMResponse().getStatus(), OK);
+
+    Path keyPath = Paths.get(keyName);
+
+    // Check intermediate paths are created
+    keyPath = keyPath.getParent();
+    while(keyPath != null) {
+      Assert.assertNotNull(omMetadataManager.getKeyTable().get(
```
mukul1987, line 377: Can we change this to fs.getFileStatus(keypath) so that returns status.isDir() ?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
bharatviswa504: Yes, there will be some intermediate directories left in DB, key commit failed. There will be intermediate directories created.
But I don't see that causing an issue. But cleaner thing is cleaning up entries. 
But this is a common issue for FileCreate and KeyCreate.

And in the case lets say key create itself failed in DB flush, we terminate DB.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
bharatviswa504: Done

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
bharatviswa504: done

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyCreateRequest.java
bharatviswa504: Added test

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
```diff
@@ -265,4 +272,24 @@ public AuditMessage buildAuditMessage(AuditAction op,
...
+    if (enableFileSystemPaths) {
+      String normalizedKeyName;
+      if (keyName.startsWith(OM_KEY_PREFIX)) {
+        normalizedKeyName = Paths.get(keyName).toUri().normalize().getPath();
+      } else {
+        normalizedKeyName = Paths.get(OM_KEY_PREFIX, keyName).toUri()
+            .normalize().getPath();
+      }
+      if (!keyName.equals(normalizedKeyName)) {
+        LOG.debug("Normalized key {} to {} ", keyName, normalizedKeyName);
```
avijayanhwx, line 288: Can we print normalizedKeyName.substring(1) given we return that?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
```diff
@@ -265,4 +272,24 @@ public AuditMessage buildAuditMessage(AuditAction op,
     auditMap.put(OzoneConsts.VOLUME, volume);
     return auditMap;
   }
+
+  @SuppressFBWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
+  public static String getNormalizedKey(boolean enableFileSystemPaths,
```
avijayanhwx, line 277: What about moving the 'enableFileSystemPaths' flag outside, and call the method conditionally? In that way, the getNormalizedKey will stay a simple util function without knowing about application logic.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestNormalizePaths.java
```diff
@@ -0,0 +1,69 @@
...
+        getNormalizedKey(true, "/a/b/c/d"));
+    Assert.assertEquals("a/b/c/d",
+        getNormalizedKey(true, "////a/b/c/d"));
+    Assert.assertEquals("a/b/c/d",
+        getNormalizedKey(true, "////a/b/////c/d"));
+    Assert.assertEquals("a/b/c/...../d",
+        getNormalizedKey(true, "////a/b/////c/...../d"));
+    Assert.assertEquals("a/b/d",
+        getNormalizedKey(true, "/a/b/c/../d"));
+    Assert.assertEquals("a/d",
```
avijayanhwx, line 47: Nit. Can we add a test for key with length = 1? Since we return .substring(1) in the method, trying to be cautious here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
bharatviswa504: That requires in all over the code, we need to check this flag and call this. I feel this is better way.
If needed I can move the code of actual Normalize, to another utility if you expect this to be used else where.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestNormalizePaths.java
bharatviswa504: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
bharatviswa504: Moved common logic to a new method.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
bharatviswa504: Done

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -2478,4 +2478,18 @@
       rules in Amazon S3's object key naming guide.
     </description>
   </property>
+
+  <property>
+    <name>ozone.om.enable.filesystem.paths</name>
+    <tag>OZONE, OM</tag>
+    <value>false</value>
+    <description>If true, OM will normalize the key paths and also create
```
arp7, line 2486: I would rephrase first sentence a bit to state that key names will be interpreted as file system paths. `/` will be treated as a special character and paths will be normalized and must follow Unix filesystem path naming conventions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
```diff
@@ -265,4 +272,30 @@ public AuditMessage buildAuditMessage(AuditAction op,
...
+  public static String getNormalizedKey(String keyName) {
+    String normalizedKeyName;
+    if (keyName.startsWith(OM_KEY_PREFIX)) {
+      normalizedKeyName = Paths.get(keyName).toUri().normalize().getPath();
+    } else {
+      normalizedKeyName = Paths.get(OM_KEY_PREFIX, keyName).toUri()
+          .normalize().getPath();
+    }
+    if (!keyName.equals(normalizedKeyName)) {
+      LOG.debug("Normalized key {} to {} ", keyName,
```
arp7, line 296: Thanks for adding this log message. It will be super useful!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -79,15 +79,11 @@ public static OMPathInfo verifyFilesInPath(
...
-      if (omMetadataManager.getKeyTable().isExist(dbKeyName)) {
-        // Found a file in the given path.
-        // Check if this is actual file or a file in the given path
-        if (dbKeyName.equals(fileNameFromDetails)) {
-          result = OMDirectoryResult.FILE_EXISTS;
-        } else {
-          result = OMDirectoryResult.FILE_EXISTS_IN_GIVENPATH;
-        }
-      } else if (omMetadataManager.getKeyTable().isExist(dbDirKeyName)) {
+      // Check first for dir. This is to handle leading "/" in path, which
```
arp7, line 82: Leading `/` in path should be dropped at this point, right?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -79,15 +79,11 @@ public static OMPathInfo verifyFilesInPath(
       String dbDirKeyName = omMetadataManager.getOzoneDirKey(volumeName,
           bucketName, pathName);
 
-      if (omMetadataManager.getKeyTable().isExist(dbKeyName)) {
```
arp7, line 82: I am a little bit worried about changes to OMFileRequest. There is a risk of changing the behavior and invalidating the app-compat testing we have done so far.

bharatviswa504: This change can be reverted as now we don't allow leading / in path names when OMConfig enable.filesystems is set to true. Will updated in my next PR update.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
bharatviswa504: Yes. This change was required, when we are allowing leading / in keyName. I will revert the change

bharatviswa504:
Thank You @arp7 and @avijayanhwx for the review.
I have addressed review comments.

bharatviswa504:
Rebased with latest apache main branch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
```diff
@@ -265,4 +273,72 @@ public AuditMessage buildAuditMessage(AuditAction op,
...
+      for (int i = 0; i < components.length; i++) {
+        String element = components[i];
+        if (element.equals(".") ||
+            (element.contains(":")) ||
+            (element.contains("/") || element.equals(".."))) {
+          isValid = false;
+          break;
+        }
+
+        // The string may end with a /, but not have
```
arp7, line 330: Does this mean we are allowing trailing `/`?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -141,7 +141,6 @@ public OMRequest preExecute(OzoneManager ozoneManager) throws IOException {
               ozoneManager.getPreallocateBlocksMax(),
               ozoneManager.isGrpcBlockTokenEnabled(),
               ozoneManager.getOMNodeId());
-
```
arp7, line 144: Spurious change?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
arp7: I realized this is checked later in `OmKeyCreateRequest`.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
```diff
@@ -91,6 +99,19 @@ public OMRequest preExecute(OzoneManager ozoneManager) throws IOException {
...
+
+    String keyPath = keyArgs.getKeyName();
+    if (ozoneManager.getEnableFileSystemPaths()) {
+      // If enabled, disallow keys with trailing /. As in fs semantics
+      // directories end with trailing /.
+      keyPath = validateAndNormalizeKey(
+          ozoneManager.getEnableFileSystemPaths(), keyPath);
+      if (keyPath.endsWith("/")) {
+        throw new OMException("Invalid KeyPath: " + keyPath,
+            OMException.ResultCodes.INVALID_KEY_NAME);
```
arp7, line 111: Let's also mention why it is invalid, else user may be scratching their head. 🙂

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyCreateRequest.java
```diff
@@ -327,7 +344,164 @@ private OMRequest createKeyRequest(boolean isMultipartKey, int partNumber) {
...
+
+    // converted to a/b/c/.../file3
+    keyName = "///a/b///c//.../file3";
+    createAndCheck(keyName);
+
+    // converted to r1/r2
+    keyName = "././r1/r2/";
+    createAndCheck(keyName);
+
+    // converted to ..d1/d2/d3
```
arp7, line 390: Why is this converted to `../d1/d2/d3`? Should it be failed instead due to trailing `/`?

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyCreateRequest.java
```diff
@@ -327,7 +344,164 @@ private OMRequest createKeyRequest(boolean isMultipartKey, int partNumber) {
...
+    checkNotAValidPath(keyName);
+
+    // Key name ends with /
+    keyName = "/a/./";
+    checkNotAValidPath(keyName);
+
+    keyName = "/////";
+    checkNotAValidPath(keyName);
+
+    keyName = "../../b/c";
```
arp7, line 410: Also let's add a check for `../a/b`.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyCreateRequest.java
```diff
@@ -327,7 +344,164 @@ private OMRequest createKeyRequest(boolean isMultipartKey, int partNumber) {
...
+            101L, ozoneManagerDoubleBufferHelper);
+
+    Assert.assertEquals(OK, omClientResponse.getOMResponse().getStatus());
+
+    checkCreatedPaths(omKeyCreateRequest, omRequest, keyName);
+  }
+
+  private void checkCreatedPaths(OMKeyCreateRequest omKeyCreateRequest,
+      OMRequest omRequest, String keyName) throws Exception {
+    keyName = omKeyCreateRequest.validateAndNormalizeKey(true, keyName);
```
arp7, line 482: Let's add a separate test case for `validateAndNormalizeKey` itself. All it does is call `validateAndNormalizeKey` with bunch of hard-coded input and asserts on the expected output.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyCreateRequest.java
```diff
@@ -327,7 +344,164 @@ private OMRequest createKeyRequest(boolean isMultipartKey, int partNumber) {
...
+    // Check intermediate directories created or not.
+    Path keyPath = Paths.get(keyName);
+    checkIntermediatePaths(keyPath);
```
arp7, line 485: Nice!

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyCreateRequest.java
arp7: Ignore this comment, I see it above now. Was reviewing files out of order.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestNormalizePaths.java
```diff
@@ -0,0 +1,109 @@
...
+    Assert.assertEquals("a/b/c/d",
+        validateAndNormalizeKey(true, "a/b/c/d"));
+    Assert.assertEquals("a/b/c/d",
+        validateAndNormalizeKey(true, "/a/b/c/d"));
+    Assert.assertEquals("a/b/c/d",
+        validateAndNormalizeKey(true, "////a/b/c/d"));
+    Assert.assertEquals("a/b/c/d",
+        validateAndNormalizeKey(true, "////a/b/////c/d"));
+    Assert.assertEquals("a/b/c/...../d",
+        validateAndNormalizeKey(true, "////a/b/////c/...../d"));
```
arp7, line 51: Bharat, more than two dots should normalize to itself. In Unix I can create a dir named `...` or `.....`.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
bharatviswa504: Yes, we don't allow for a keyName with trailing /. This method can return after normalization in some cases. So, we checked in OmKeyCreateRequest

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestNormalizePaths.java
bharatviswa504: yes, you are right. It is the samething right, we are checking a/b/c/...../d, just it has removed additional and leading /

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyCreateRequest.java
bharatviswa504: ..d1/d2/d3/ -> ../d1/d2/d3 it removed trailing /. Normalize will take care of removing additional /

bharatviswa504: we will fail, even after normalize if we have a case like /a/. -> a/

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/key/TestOMKeyCreateRequest.java
bharatviswa504: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
bharatviswa504: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
bharatviswa504: Removed it

bharatviswa504:
> Few minor comments, and couple of questions on the expected normalizations.
> 
> Overall the change looks good. Thanks for adding comprehensive test cases.

Thank You @arp7 for the review. addressed review comments.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestNormalizePaths.java
arp7: Right, I got confused.

PR #1137, opened 2020-06-27
HDDS-3060. /retest github comment does not work

adoroszlai:
## What changes were proposed in this pull request?

Currently `/retest` command creates an empty commit, but unlike manual commits it does not re-trigger CI.  It [turns out](https://help.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token) that this is a safety measure on GitHub:

> events triggered by the GITHUB_TOKEN will not create a new workflow run

This PR proposes to make `/retest` command useful by adding a comment to the PR describing steps to trigger CI manually.  Actual steps depend on the comment author and PR setting _Allow edits and access to secrets by maintainers_.

https://issues.apache.org/jira/browse/HDDS-3060

## How was this patch tested?

[Tested](https://github.com/adoroszlai/hadoop-ozone/pull/7) various conditions in my fork.

PR #1197, opened 2020-07-11
HDDS-3925. SCM Pipeline DB should directly use UUID bytes for key rather than rely on proto serialization for key.

fapifta:
As we recently learned, protoBuf serialization is not needed to produce the same byte array over the wire, and therefore the toByteArray method's return value should not be used to byte array based comparison of protobuf serializations.
In the SCM's RocksDB we use the PipelineID as a key in the Pipeline table, and the key is created using the byte array representation of the protobuf serialization of the PipelineID, which can lead to mismatches when we access anything from the table based on a key. At the moment this can prevent us to delete a Pipeline from the table if there is a change in the byte array representation.

In order to avoid this situation, the way how we create the key to this table from the PipelineID needs to be changed, this is one part of this PR, and covered in the PipelineIDCodec related changes.
The other part is to clean up the DB from the old keys, as SCM will not be able to close and remove those Pipeline based on the ID as after we change the key serialization the byte array matched and stored will not be the same for the same PipelineID, that is used in the delete method which the code uses. So SCM can end up with pipelines that will never be cleaned up from the DB, and are polluting the in memory structures at startup until they are considered invalid.
In order to avoid this, SCMPipelineManager from now on checks if the key deserialization results in the same PipelineID as the one stored in the value in the table which is a Pipeline object. If the two are different, we remove the old version from the table, and add it with the new key, so that the pipelines are still preserved, but later can be deleted from RocksDB as well by the SCM when appropriate.

https://issues.apache.org/jira/browse/HDDS-3925

## How was this patch tested?
JUnit tests are added to new things. And to check key migration happens properly.

### DIFF
--- a/hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/utils/db/TestRDBStoreIterator.java
+++ b/hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/utils/db/TestRDBStoreIterator.java
@@ -1,0 +1,201 @@
+package org.apache.hadoop.hdds.utils.db;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.InOrder;
+import org.rocksdb.RocksIterator;
+
+import java.util.NoSuchElementException;
+import java.util.function.Consumer;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyCollectionOf;
+import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class TestRDBStoreIterator {
+
+  private RocksIterator rocksDBIteratorMock;
+  private RDBTable rocksTableMock;
+
+  @Before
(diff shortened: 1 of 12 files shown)


avijayanhwx:
Thank you working on this pifta. The patch looks good to me. I have verified the working using docker based testing. Can we add a unit test for to verify that removeFromDb actually removes the entry?

fapifta:
Hi @avijayanhwx,

thank you for the review, I have pushed the requested test, and a bit more.

At the end of the day, I have added tests to verify the behaviour and interactions of RDBStoreIterator with the underlying RockIterator, and the RocksDBTable. I hope this should sufficiently address the test request, let me know if you thought about something different.

As the TypedTable.TypedTableIterator class purely delegates to the raw RDbStoreIterator, I think that does not require too much tests.

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/utils/db/TestRDBStoreIterator.java
```diff
@@ -0,0 +1,224 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
```
avijayanhwx, line 4: Thanks for adding this test @fapifta.

fapifta:
Thank you for the review @avijayanhwx, I haven't seen the CI checks running, so I pushed the branch once more, let's see, all failures seems irrelevant, especially after a clean build was there already before I have added the new test.

adoroszlai:
`/retest` does not work (#1137)

avijayanhwx:
Thank you for fixing this @fapifta. I have merged your patch.

PR #1199, opened 2020-07-14
HDDS-3957. Fix mixed use of Longs.toByteArray and Ints.fromByteArray

runzhiwang:
**What's the problem ?**
When write `DB_PENDING_DELETE_BLOCK_COUNT_KEY` to rocksdb, most code convert value to byte array by `Longs.toByteArray`. But when read,  `parseKVContainerData` use `Ints.fromByteArray`. The result is always wrong, unless the value is zero.  
image

image

For example, `Longs.toByteArray(1)` return byte array in which only byte[7] is 1, but `Ints.fromByteArray` only parse the first 4 bytes of byte array. So write 1, but read out 0.
image
image

https://issues.apache.org/jira/browse/HDDS-3957

## How was this patch tested?

add assert in current ut.


hanishakoneru:
Thanks @runzhiwang for fixing this and @adoroszlai for reviewing it.
+1. Will merge the PR.

PR #1211, opened 2020-07-16
HDDS-3969. Add validName check for FileSystem requests

bharatviswa504:

This Jira is to validate KeyNames which are created with OzoneFileSystem.
Similar to how hdfs handles using DFSUtil. isValidName()

Made only changes in the client.

https://issues.apache.org/jira/browse/HDDS-3969

## How was this patch tested?

Added tests.

### DIFF
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -28,6 +28,7 @@ import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileAlreadyExistsException;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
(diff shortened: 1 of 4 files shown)


hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -721,7 +723,13 @@ public String pathToKey(Path path) {
       path = new Path(workingDir, path);
     }
     // removing leading '/' char
-    String key = path.toUri().getPath().substring(1);
+    String key = path.toUri().getPath();
+
+    if (OzoneFSUtils.isValidName(key)) {
+      key = path.toUri().getPath().substring(1);
```
avijayanhwx, line 729: Can we just do this here, instead of toUri().getPath() again?
`    LOG.trace("path for key:{} is:{}", key, path);
    return key.substring(1);
`

bharatviswa504: Done.

PR #1212, opened 2020-07-17
HDDS-3979. Make bufferSize configurable for stream copy

maobaolong:

- Refactor the code related copy large by using seek to enforce skip method of KeyInputStream and S3WrapperInputStream, the related codes.
  - We can use `IOUtils#copyLarge` to clarify and reduce the duplicated code.
  - After this, It's easy to specify a buffer by using `IOUtils#copyLarge`.
- Make bufferSize configurable for stream copy.
  - So that user like s3g can specify a buffer size to balance the memory and performance. 

HDDS-3979

## How was this patch tested?

Using awscli or goofys

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -465,6 +465,12 @@ public final class OzoneConfigKeys {
   public static final String  OZONE_CLIENT_HTTPS_NEED_AUTH_KEY =
       "ozone.https.client.need-auth";
   public static final boolean OZONE_CLIENT_HTTPS_NEED_AUTH_DEFAULT = false;
+
+  public static final String OZONE_CLIENT_BUFFER_SIZE_KEY =
+      "ozone.client.buffer.size";
+  public static final String OZONE_CLIENT_BUFFER_SIZE_DEFAULT =
+      "4KB";
+
   /**
    * There is no need to instantiate this class.
    */
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -466,11 +466,6 @@ public final class OzoneConfigKeys {
       "ozone.https.client.need-auth";
   public static final boolean OZONE_CLIENT_HTTPS_NEED_AUTH_DEFAULT = false;
 
-  public static final String OZONE_CLIENT_BUFFER_SIZE_KEY =
-      "ozone.client.buffer.size";
-  public static final String OZONE_CLIENT_BUFFER_SIZE_DEFAULT =
-      "4KB";
-
   /**
    * There is no need to instantiate this class.
    */
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2422,6 +2422,14 @@
     </description>
   </property>
   <property>
+    <name>ozone.client.buffer.size</name>
+    <tag>OZONE, S3GATEWAY</tag>
+    <value>4KB</value>
+    <description>
+      The size of the buffer which is for read block. (4KB by default).
+    </description>
+  </property>
+  <property>
     <name>ssl.server.keystore.keypassword</name>
     <tag>OZONE, SECURITY, MANAGEMENT</tag>
     <value></value>
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2422,7 +2422,7 @@
     </description>
   </property>
   <property>
-    <name>ozone.client.buffer.size</name>
+    <name>ozone.s3g.client.buffer.size</name>
     <tag>OZONE, S3GATEWAY</tag>
     <value>4KB</value>
     <description>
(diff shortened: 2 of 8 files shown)


hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyInputStream.java
```diff
@@ -325,62 +323,14 @@ public long getRemainingOfIndex(int index) throws IOException {
...
-  public long copyLarge(final OutputStream output,
-      final long inputOffset, final long len, final byte[] buffer)
-      throws IOException {
-    if (inputOffset > 0) {
-      seek(inputOffset);
-    }
-
-    if (len == 0) {
+  @Override
+  public long skip(long n) throws IOException {
```
runzhiwang, line 327: add unit test for skip.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestKeyInputStream.java
```diff
@@ -395,4 +334,64 @@ public void testReadChunk() throws Exception {
     }
     keyInputStream.close();
   }
+
+  @Test
+  public void testSkip() throws Exception {
+    {
```
runzhiwang, line 340: why { ?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestKeyInputStream.java
```diff
@@ -395,4 +334,64 @@ public void testReadChunk() throws Exception {
...
+      XceiverClientManager.resetXceiverClientMetrics();
```
runzhiwang, line 341: 4 space ?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyInputStream.java
maobaolong: Ok, done.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestKeyInputStream.java
maobaolong: Sorry, my bad, already fixed it.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestKeyInputStream.java
maobaolong: Sorry, my bad, already fixed it.

maobaolong:
@runzhiwang Thanks for the review, addressed all of your suggestion, PTAL.

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -2421,6 +2421,14 @@
       information will be extracted
     </description>
   </property>
+  <property>
+    <name>ozone.client.buffer.size</name>
+    <tag>OZONE, S3GATEWAY</tag>
+    <value>4KB</value>
+    <description>
+      The size of the buffer which is for read block. (4KB by default).
```
runzhiwang, line 2429: We can describe what size can achieve best performance.

maobaolong:
@runzhiwang Thank you for your review and `LGTM`. @bharatviswa504 Please take a look at this PR, thanks.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
```diff
@@ -465,6 +465,12 @@
   public static final String  OZONE_CLIENT_HTTPS_NEED_AUTH_KEY =
       "ozone.https.client.need-auth";
   public static final boolean OZONE_CLIENT_HTTPS_NEED_AUTH_DEFAULT = false;
+
```
elek, line 468: Two things:

 * `ozone.client.buffer.size` seems to be confusing. This configuration for the Ozone client of the **s3g**. As we have a global list of configuration keys, I guess we need a `s3g` prefix here. (For example `ozone.s3g.client.buffer.size`) 
 * Usually, I prefer to use annotation based configuration, if possible. (It can be harder here as S3g uses CDI injections. If it's too complex, you can ignore...)

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/endpoint/ObjectEndpoint.java
```diff
@@ -114,6 +120,13 @@ public ObjectEndpoint() {
     customizableGetHeaders.add("Content-Encoding");
   }
 
+  @PostConstruct
+  public void init() {
```
elek, line 124: NIT: It can be slightly better to use

```
  @Inject
  private OzoneConfiguration ozoneConfiguration;
```

And you don't need to expose `getClient()` in the `EndpointBase`. 

(Didn't try, but it should work, IMHO).

maobaolong: Thank you for let me know this high technology

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
maobaolong: Thank you for give me this task,  i have create a ticket for it, and take it, https://issues.apache.org/jira/browse/HDDS-4049

For this PR, lets just correct the key name.

maobaolong:
@elek Thanks you for your review and suggestion, i've addressed your comments, PTAL.

PR #1214, opened 2020-07-17
HDDS-3981. Add more debug level log to XceiverClientGrpc for debug purpose

maobaolong:

When I profile the s3g performance, i find in some log of XceiverClientGrpc, I cannot find the related datanode ip, so i use datanode instead of the UUID and networkFullPath.

Also, i add a debug level log to show the request has been executed and output the try counts.

HDDS-3981

## How was this patch tested?

No need, all changes only related to log.

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
@@ -58,7 +58,6 @@ import com.google.common.base.Preconditions;
 import io.opentracing.Scope;
 import io.opentracing.Span;
 import io.opentracing.util.GlobalTracer;
-import org.apache.hadoop.util.Time;
 import org.apache.ratis.thirdparty.io.grpc.ManagedChannel;
 import org.apache.ratis.thirdparty.io.grpc.Status;
 import org.apache.ratis.thirdparty.io.grpc.netty.GrpcSslContexts;
@@ -330,11 +329,8 @@ public class XceiverClientGrpc extends XceiverClientSpi {
       }
     }
 
-    int index = -1;
     for (DatanodeDetails dn : datanodeList) {
-      index++;
       try {
-        long startTime = Time.monotonicNow();
         if (LOG.isDebugEnabled()) {
           LOG.debug("Executing command {} on datanode {}", request, dn);
         }
@@ -343,11 +339,6 @@ public class XceiverClientGrpc extends XceiverClientSpi {
         // in case these don't exist for the specific datanode.
         reply.addDatanode(dn);
         responseProto = sendCommandAsync(request, dn).getResponse().get();
-        if (LOG.isDebugEnabled()) {
-          LOG.debug("Executed command {} on datanode {}, cost = {}, "
-                  + "retryCount = {}" + ", cmdType = {}", request, dn,
-              Time.monotonicNow() - startTime, index, request.getCmdType());
-        }
         if (validators != null && !validators.isEmpty()) {
           for (CheckedBiFunction validator : validators) {
             validator.apply(request, responseProto);
@@ -461,8 +452,14 @@ public class XceiverClientGrpc extends XceiverClientSpi {
               public void onNext(ContainerCommandResponseProto value) {
                 replyFuture.complete(value);
                 metrics.decrPendingContainerOpsMetrics(request.getCmdType());
+                long cost = System.nanoTime() - requestTime;
                 metrics.addContainerOpsLatency(request.getCmdType(),
-                    System.nanoTime() - requestTime);
+                    cost);
+                if (LOG.isDebugEnabled()) {
+                  LOG.debug("Executed command {} on datanode {}, cost = {}, "
+                          + "cmdType = {}", request, dn,
+                      cost, request.getCmdType());
+                }
                 semaphore.release();
               }
 
@@ -470,8 +467,14 @@ public class XceiverClientGrpc extends XceiverClientSpi {
               public void onError(Throwable t) {
                 replyFuture.completeExceptionally(t);
                 metrics.decrPendingContainerOpsMetrics(request.getCmdType());
+                long cost = System.nanoTime() - requestTime;
                 metrics.addContainerOpsLatency(request.getCmdType(),
                     System.nanoTime() - requestTime);
+                if (LOG.isDebugEnabled()) {
+                  LOG.debug("Executed command {} on datanode {}, cost = {}, "
+                          + "cmdType = {}", request, dn,
+                      cost, request.getCmdType());
+                }
                 semaphore.release();
               }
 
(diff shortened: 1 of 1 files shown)


hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
```diff
@@ -354,7 +363,7 @@ private XceiverClientReply sendCommandWithRetry(
         responseProto = null;
       } catch (ExecutionException e) {
         LOG.debug("Failed to execute command {} on datanode {}",
-            request, dn.getUuid(), e);
```
xiaoyuyao, line 357: Full DatanodeDetails.toString() include a lot of information and can hurt the perf on critical path. Consider using either ip or uuid instead.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
```diff
@@ -354,7 +363,7 @@ private XceiverClientReply sendCommandWithRetry(
         responseProto = null;
       } catch (ExecutionException e) {
         LOG.debug("Failed to execute command {} on datanode {}",
```
xiaoyuyao, line 365: This needs to be guarded by if (LOG.isDebugEnabled())

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
```diff
@@ -339,6 +343,11 @@ private XceiverClientReply sendCommandWithRetry(
         // in case these don't exist for the specific datanode.
         reply.addDatanode(dn);
         responseProto = sendCommandAsync(request, dn).getResponse().get();
+        if (LOG.isDebugEnabled()) {
```
xiaoyuyao, line 346: In sendCommandAsync() below we already have a LOG.debug(cmdType, Dn). Can you move this there to avoid duplicate debug log?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
maobaolong: I hope to using ip and uuid like ip/uuid, do you think it is ok for you?
Because, i want to get the ip so that i can clearly know the exception related to which datanode, but in local mode, all of ips of datanode are the same, for this situation, an addition uuid can help me to distinguish.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
maobaolong: @xiaoyuyao I hope so, but I think it would be difficult, because, i want to log the retry index and cost time.
when the `sendCommandAsync ` return back, the client still doesn't get the result from datanode. the result return until the `get` return.

If i have something wrong, please correct me.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
maobaolong: OK

maobaolong:
@xiaoyuyao Thanks to review my PR, i reply to your comments, please check it again, thanks.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
```diff
@@ -433,7 +442,7 @@ public XceiverClientReply sendCommandAsync(
     UUID dnId = dn.getUuid();
     if (LOG.isDebugEnabled()) {
       LOG.debug("Send command {} to datanode {}",
-          request.getCmdType(), dn.getNetworkFullPath());
+          request.getCmdType(), dn);
```
xiaoyuyao, line 445: Correct me if I'm wrong, I think we have already print the same debug message on Line 338-340 from the caller. Print just the IP address should be sufficient.

maobaolong: OK, you are right, the redundancy log content is useless, I've change it to `dn.getIpAddress()`.

maobaolong:
@xiaoyuyao Thank you for the review suggestion, PTAL.

elek:
@bharatviswa504 What do we need to make it possible to merge this commit?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
xiaoyuyao: Good point, you are right on this.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
xiaoyuyao: I think ip address should be sufficient for production environment. The local mode is for mini clsuter test only.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
xiaoyuyao: I mean if we could move the LOG for latency around Line 464 where the container op latency metrics are updated. This way, we can piggyback LOG with the existing metrics update. 

metrics.addContainerOpsLatency(request.getCmdType(),
                    System.nanoTime() - requestTime);

xiaoyuyao (commented):
Thanks @maobaolong for the update. One more comment added inline.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
maobaolong: @xiaoyuyao Thanks you for your reply and suggestion. I tried to add the LOG into the L464, like following
```java
              public void onNext(ContainerCommandResponseProto value) {
                replyFuture.complete(value);
                metrics.decrPendingContainerOpsMetrics(request.getCmdType());
                long cost = System.nanoTime() - requestTime;
                metrics.addContainerOpsLatency(request.getCmdType(),
                    cost);
                if (LOG.isDebugEnabled()) {
                  LOG.debug("Executed command {} on datanode {}, cost = {}, "
                          + "retryCount = {}" + ", cmdType = {}", request, dn,
                      cost, index, request.getCmdType());
                }
                semaphore.release();
              }
```

But, there are some problem
- It cannot print the `Index` which stand for the retried datanode index of the `datanodeList`.
- I should add the same log logic to the `onError` also? Otherwise, when we met error, the `onNext` won't be executed.

As `index` can be calculate from the log context, I drop it.

maobaolong:
@xiaoyuyao Thanks @xiaoyuyao for the suggestion, I pushed a new commit to address your comment, I reply your comment and ask some question. PTAL.

xiaoyuyao:
Thanks @maobaolong  for the update. LGTM, +1. Will merge it shortly.

PR #1215, opened 2020-07-17
HDDS-3982. Disable moveToTrash in o3fs and ofs temporarily

smengcl:

A proper server-side trash cleanup solution (HDDS-3915) might not land any time soon.

This jira aims to completely disable "move to trash" when a client is deleting files and {{fs.trash.interval > 0}} by intercepting the deprecated {{fs.rename(src, dst, options)}} call used by {{TrashPolicyDefault#moveToTrash}}.

This should be reverted when trash cleanup is implemented.

https://issues.apache.org/jira/browse/HDDS-3982

## How was this patch tested?

1. Added new tests.

2. Also tested manually:

As [usual](https://github.com/apache/hadoop-ozone/pull/1181), compile -> launch docker-compose cluster -> exec bash on om

```
$ mvn clean install -Pdist -DskipTests -e -Dmaven.javadoc.skip=true -DskipShade
$ cd hadoop-ozone/dist/target/ozone-*-SNAPSHOT/compose/ozone
$ docker-compose up -d
$ docker-compose exec om /bin/bash
```

Add `fs.trash.interval` to client config:
```
$ vi /etc/hadoop/core-site.xml
# Add a line in <configuration> tag and save:
<property><name>fs.trash.interval</name><value>1</value></property>
```

### o3fs

```
$ ozone sh volume create volume1
$ ozone sh bucket create /volume1/bucket2
$ ozone fs -mkdir -p o3fs://bucket2.volume1.om/dir3
$ ozone fs -touch o3fs://bucket2.volume1.om/dir3/key5
$ ozone fs -rm o3fs://bucket2.volume1.om/dir3/key5
2020-07-17 19:22:29,778 [main] INFO Configuration.deprecation: io.bytes.per.checksum is deprecated. Instead, use dfs.bytes-per-checksum
2020-07-17 19:22:29,867 [main] INFO ozone.BasicOzoneFileSystem: Move to trash is disabled for o3fs, deleting instead: o3fs://bucket2.volume1.om/dir3/key5
2020-07-17 19:22:29,881 [main] INFO fs.TrashPolicyDefault: Moved: 'o3fs://bucket2.volume1.om/dir3/key5' to trash at: /.Trash/hadoop/Current/dir3/key5
```

Verification: key is deleted, but trash directory is created:
```
$ ozone fs -ls -R o3fs://bucket2.volume1.om/.Trash/hadoop/Current/
drwxrwxrwx   - hadoop hadoop          0 2020-07-17 19:22 o3fs://bucket2.volume1.om/.Trash/hadoop/Current/dir3
```

### OFS

```
$ ozone fs -mkdir -p ofs://om/volume1/bucket2/dir3
$ ozone fs -touch ofs://om/volume1/bucket2/dir3/key5
$ ozone fs -rm ofs://om/volume1/bucket2/dir3/key5
2020-07-17 19:26:59,193 [main] INFO Configuration.deprecation: io.bytes.per.checksum is deprecated. Instead, use dfs.bytes-per-checksum
2020-07-17 19:26:59,306 [main] INFO ozone.BasicRootedOzoneFileSystem: Move to trash is disabled for ofs, deleting instead: ofs://om/volume1/bucket2/dir3/key5
2020-07-17 19:26:59,334 [main] INFO fs.TrashPolicyDefault: Moved: 'ofs://om/volume1/bucket2/dir3/key5' to trash at: ofs://om/volume1/bucket2/.Trash/hadoop/Current/volume1/bucket2/dir3/key5
```

Verification: key is deleted, but trash directory is created:
```
$ ozone fs -ls -R ofs://om/volume1/bucket2/.Trash/hadoop/Current/
drwxrwxrwx   - hadoop hadoop          0 2020-07-17 19:26 ofs://om/volume1/bucket2/.Trash/hadoop/Current/volume1
drwxrwxrwx   - hadoop hadoop          0 2020-07-17 19:26 ofs://om/volume1/bucket2/.Trash/hadoop/Current/volume1/bucket2
drwxrwxrwx   - hadoop hadoop          0 2020-07-17 19:26 ofs://om/volume1/bucket2/.Trash/hadoop/Current/volume1/bucket2/dir3
```

## Known Issues

* `TrashPolicyDefault` already creates the empty directories in trash before calling `fs.rename(src, dst, options)`.
  * These are empty dirs so they shouldn't be a big problem.
  * We could do something about `fs.mkdirs` as well to solve this but I'm not sure if we should go down this path.
* `TrashPolicyDefault` still prints "Moved ... to trash" message, as shown above, which may be confusing to users.

### DIFF
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -29,6 +29,7 @@ import org.apache.hadoop.fs.FileAlreadyExistsException;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.LocatedFileStatus;
+import org.apache.hadoop.fs.Options.Rename;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathIsNotEmptyDirectoryException;
 import org.apache.hadoop.fs.permission.FsPermission;
@@ -395,6 +396,32 @@ public class BasicOzoneFileSystem extends FileSystem {
       createFakeParentDirectory(src);
     }
     return result;
+  }
+
+  /**
+   * Intercept rename to trash calls from TrashPolicyDefault,
+   * convert them to delete calls instead.
+   */
+  @Deprecated
+  protected void rename(final Path src, final Path dst,
+      final Rename... options) throws IOException {
+    boolean hasMoveToTrash = false;
+    if (options != null) {
+      for (Rename option : options) {
+        if (option == Rename.TO_TRASH) {
+          hasMoveToTrash = true;
+          break;
+        }
+      }
+    }
+    if (!hasMoveToTrash) {
+      // if doesn't have TO_TRASH option, just pass the call to super
+      super.rename(src, dst, options);
+    } else {
+      // intercept when TO_TRASH is found
+      LOG.info("Move to trash is disabled for o3fs, deleting instead: {}", src);
+      delete(src, true);
+    }
   }
 
   private class DeleteIterator extends OzoneListingIterator {
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -421,7 +421,7 @@ public class BasicOzoneFileSystem extends FileSystem {
       // intercept when TO_TRASH is found
       LOG.info("Move to trash is disabled for o3fs, deleting instead: {}. "
           + "Files or directories will NOT be retained in trash. "
-          + "Ignore the following TrashPolicyDefault message.", src);
+          + "Ignore the following TrashPolicyDefault message, if any.", src);
       delete(src, true);
     }
   }
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -419,7 +419,9 @@ public class BasicOzoneFileSystem extends FileSystem {
       super.rename(src, dst, options);
     } else {
       // intercept when TO_TRASH is found
-      LOG.info("Move to trash is disabled for o3fs, deleting instead: {}", src);
+      LOG.info("Move to trash is disabled for o3fs, deleting instead: {}. "
+          + "Files or directories will NOT be retained in trash. "
+          + "Ignore the following TrashPolicyDefault message.", src);
       delete(src, true);
     }
   }
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
@@ -29,6 +29,7 @@ import org.apache.hadoop.fs.FileAlreadyExistsException;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.LocatedFileStatus;
+import org.apache.hadoop.fs.Options;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathIsNotEmptyDirectoryException;
 import org.apache.hadoop.fs.permission.FsPermission;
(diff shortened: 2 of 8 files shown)


umamaheswararao:
This is a good idea.

- But the below log message from could confuse people?
```
 fs.rename(path, trashPath,
            Rename.TO_TRASH);
        LOG.info("Moved: '" + path + "' to trash at: " + trashPath);
```
I don't see a way to avoid though. :-(
Probably we will say: A generic statement in previous log  "We will not retain any data in trash. This may just reduce confusion that, they will think data moved but deleted immediately. :-) 

Probably your logs will looks like:
INFO ozone.BasicOzoneFileSystem: Move to trash is disabled for o3fs, deleting instead: o3fs://bucket2.volume1.om/dir3/key5.
                                                              Files/dirs will not be retained in Trash.
INFO fs.TrashPolicyDefault: Moved: 'o3fs://bucket2.volume1.om/dir3/key5' to trash at: /.Trash/hadoop/Current/dir3/key5
I am not this is confusing more. Think about some generic message to convey that next message is just fake.

- I think we need to add test case to make sure no files moving under trash folder.


One another thought could be that: ( I am not proposing it to do it now, just for discussion):
Currently in Hadoop side, the trash policy config is common for all kinds of fs.
```
 Class<? extends TrashPolicy> trashClass = conf.getClass(
        "fs.trash.classname", TrashPolicyDefault.class, TrashPolicy.class);
    TrashPolicy trash = ReflectionUtils.newInstance(trashClass, conf);
    trash.initialize(conf, fs); // initialize TrashPolicy
    return trash
```
If user wants different policies based on their FS behaviors, that's not possible. Probably if we have confg like fs.<scheme>.trash.classname and by default we can use TrashPolicyDefault.class. if user wants, they can configure per fs specific policy. 
So, If config was defined this way, our life would have been easier now. In our case we could have configured our own policy and simply delete the files in it. 
Anyway what we are doing is temporary until we have proper cleanup. So, this option will not help as we need changes in Hadoop side.

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -397,6 +398,32 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+          hasMoveToTrash = true;
+          break;
+        }
+      }
+    }
+    if (!hasMoveToTrash) {
+      // if doesn't have TO_TRASH option, just pass the call to super
+      super.rename(src, dst, options);
+    } else {
+      // intercept when TO_TRASH is found
```
umamaheswararao, line 421: Please check my above comment.

smengcl:
@umamaheswararao  Thanks for the review.

One question about the `fs.trash.classname` though: I actually thought about overriding it. But once this is configured to a custom class on the client it seems it will use that Trash Policy class for ALL HCFS?

umamaheswararao:
@smengcl yes, fs.trash.classname configuration is for all HCFS.
If we wanted to have different impls for HCFSs, we may need to introduce new advanced config, that should be like fs.<scheme>.trash.classname.

The impl can be something like below:
  ```
public static TrashPolicy getInstance(Configuration conf, FileSystem fs) {
    Class<? extends TrashPolicy> defaultTrashClass = conf.getClass(
        "fs.trash.classname", TrashPolicyDefault.class, TrashPolicy.class);
    Class<? extends TrashPolicy>  trashClass = conf.getClass(String.format(
        "fs.%s.trash.classname",
        fs.getScheme()), defaultTrashClass, TrashPolicy.class);
    TrashPolicy trash = ReflectionUtils.newInstance(trashClass, conf);
    trash.initialize(conf, fs); // initialize TrashPolicy
    return trash;
  }
```

```
@Test
  public void testPluggableFileSystemSpecificTrash() throws IOException {
    Configuration conf = new Configuration();
    // Test plugged TrashPolicy
    conf.setClass("fs.file.trash.classname", TestTrashPolicy.class, TrashPolicy.class);
    Trash trash = new Trash(conf);
    assertTrue(trash.getTrashPolicy().getClass().equals(TestTrashPolicy.class));
  }
```

So, by no change in behavior. If some one wants to override the behaviors for a specific fs, provide your TrashClass name in config. Ex: if you want to provide a trashPolicy for ofs separately, then configure like:
fs.ofs.trash.classname = OfsTrashPolicyDefault.class.getName();
Otherwise same TrashPolicyDefault.class will be used.

File: hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java (hunk 1 above)
Comment (smengcl) [reply hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java:421]: done in https://github.com/apache/hadoop-ozone/pull/1215/commits/a929357c62a60bdfccfb48f1c65f5adf47c30069

**Comment by smengcl** [comment]
Tests added.

File: hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java [hunk 2]
```diff
@@ -372,6 +373,34 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+      }
+    }
+    if (!hasMoveToTrash) {
+      // if doesn't have TO_TRASH option, just pass the call to super
+      super.rename(src, dst, options);
+    } else {
+      // intercept when TO_TRASH is found
+      LOG.info("Move to trash is disabled for ofs, deleting instead: {}. "
+          + "Files or directories will NOT be retained in trash. "
+          + "Ignore the following TrashPolicyDefault message, if any.", src);
```
umamaheswararao, line 399: One last thing: Currently we are straight a way deleting files. Since users coming from Hadoop shell, there could be assumptions that files may move to trash if no -skipTrash flag. But we will be deleting instead of trashing them.
User may realize after looking at this log message, but by this time already things happened.
There should be a way to shout out this behavior to users. One way is documenting and any other better way?

umamaheswararao:
Changes looks good to me. +1
Pending CI-Checks.

smengcl:
Merged to master. Thanks @umamaheswararao  for reviewing.

PR #1219, opened 2020-07-19
HDDS-3985. Update proto.lock files.

bharatviswa504:

update proto.lock files that were missed in HDDS-3807 and HDDS-3612

https://issues.apache.org/jira/browse/HDDS-3985

## How was this patch tested?

No testing needed.


hadoop-hdds/interface-client/src/main/proto/proto.lock
```diff
@@ -1935,4 +1935,4 @@
       }
     }
   ]
-}
+}
```
bharatviswa504, line 1938: This is a generated file by proto-backwards-compatibility.
This is a related change.

adoroszlai:
I wanted to ask this earlier on other PRs: what's the policy for updating `proto.lock` files?  Shouldn't we keep a version generated for "previous release" (ie. 0.5.0 currently)?  If they are always updated, how do they guarantee backwards compatibility?

CC @elek

elek:
> I wanted to ask this earlier on other PRs: what's the policy for updating proto.lock files? Shouldn't we keep a version generated for "previous release" (ie. 0.5.0 currently)? If they are always updated, how do they guarantee backwards compatibility?

Agree, the original version should be kept. Unfortunately the maven plugin always refresh the lock files but we can ignore the proto files. I will create another PR to ignore lock files.

elek:
I propose to use #1224 instead of updating the lock files.

bharatviswa504:
Thank You @adoroszlai for the question.
As these are optional fields, I think it will not break backward compatibility.

But I am also not completely sure when this file needs to be updated? (My main reason for adding this is, as @elek said it will generated new proto.lock file from current proto, and it adds those files during PR commit.)

@mukul1987 @avijayanhwx Can you help answer the above question of when this file needs to be updated.

elek:
`proto.lock` file represents and older state of the protobuf files. When build is executed the plugin compares the current `*.proto` files and the `proto.lock` files. 

If we would like to be backward compatible with the previous release (in our case, it's 0.5.0) the `proto.lock` file should represent the `0.5.0` state. Therefore we shouldn't update it.

After the 0.6.0 release we can update all the `proto.lock` files (once!), which means during the build the master state will be compared with the 0.6.0.

As a thumb rule, `proto.lock` file should generated from the old proto which supposed to be backward compatible with the current proto files.

bharatviswa504:
> `proto.lock` file represents and older state of the protobuf files. When build is executed the plugin compares the current `*.proto` files and the `proto.lock` files.
> 
> If we would like to be backward compatible with the previous release (in our case, it's 0.5.0) the `proto.lock` file should represent the `0.5.0` state. Therefore we shouldn't update it.
> 
> After the 0.6.0 release we can update all the `proto.lock` files (once!), which means during the build the master state will be compared with the 0.6.0.
> 
> As a thumb rule, `proto.lock` file should generated from the old proto which supposed to be backward compatible with the current proto files.

Thank You @elek for your reply. I understood the process now. I agree with you, that before every release these proto.lock files should be updated at once. 

I will close this PR, but it creates a problem of proto.lock files getting updated during the build. I think this needs to be solved, as this causes problem these files are also showing up during PR commit.

adoroszlai:
> I will close this PR, but it creates a problem of proto.lock files getting updated during the build. I think this needs to be solved, as this causes problem these files are also showing up during PR commit.

Thanks @bharatviswa504.  This is being addressed in #1224.

PR #1220, opened 2020-07-20
HDDS-3986. Frequent failure in TestCommitWatcher#testReleaseBuffersOnException

adoroszlai:

1. Fix frequent failure in `TestCommitWatcher`, caused by the following.

Watch request times out with `TimeoutIOException`:

https://github.com/apache/incubator-ratis/blob/6ab75ae9da4b380056279b02f208dc9b1329325b/ratis-grpc/src/main/java/org/apache/ratis/grpc/client/GrpcClientProtocolClient.java#L336-L339

which results in no retry per the policy:

https://github.com/apache/hadoop-ozone/blob/caf471111cdb9168ec013f4526bb997aa513e079/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisHelper.java#L291-L309

The original exception is then propagated to client instead of `RaftRetryFailureException`:

https://github.com/apache/incubator-ratis/blob/6ab75ae9/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java#L324-L330

Thus the assertion for exception type fails.

2. Also reduce request timeout to reduce the time required for the test.

https://issues.apache.org/jira/browse/HDDS-3986

## How was this patch tested?

Ran TestCommitWatcher 20 times:
https://github.com/adoroszlai/hadoop-ozone/runs/888869761#step:5:4

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestCommitWatcher.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestCommitWatcher.java
@@ -46,6 +46,7 @@ import org.apache.hadoop.ozone.container.ContainerTestHelper;
 import org.apache.ratis.protocol.AlreadyClosedException;
 import org.apache.ratis.protocol.NotReplicatedException;
 import org.apache.ratis.protocol.RaftRetryFailureException;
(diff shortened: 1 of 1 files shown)


lokeshj1703:
@adoroszlai Thanks for the contribution! I have committed the PR to master branch.

adoroszlai:
Thanks @lokeshj1703 for reviewing and committing it.

PR #1223, opened 2020-07-20
HDDS-3990. Test Kubernetes examples with acceptance tests

elek:

`hadoop-ozone/dist/src/main/k8s/example` directory contains example Kubernetes resources to start Ozone in kubernetes environment. To make sure those resources are working and up-to-date I propose to test them during standard build.

K3s project provides a lightweight Kubernetes distribution which can be installed easily in Github Actions environment and Kubernetes based clusters can be tested.

https://issues.apache.org/jira/browse/HDDS-3990

## How was this patch tested?

New type of acceptance tests are executed 5 times on my fork and passed.


hadoop-ozone/dist/src/main/k8s/examples/testlib.bats
```diff
@@ -0,0 +1,61 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+
+#
+# Can be executed with bats (https://github.com/bats-core/bats-core)
+# bats gc_opts.bats (FROM THE CURRENT DIRECTORY)
```
adoroszlai, line 21:
```suggestion
# bats testlib.bats (FROM THE CURRENT DIRECTORY)
```

hadoop-ozone/dist/src/main/k8s/examples/testlib.sh
```diff
@@ -0,0 +1,125 @@
...
+
+  if [ $(basename $PARENT_OF_PARENT) == "k8s" ]; then
+    #running from src dir
+    OZONE_ROOT=$(realpath ../../../../../target/ozone-0.6.0-SNAPSHOT)
+  else
+    #running from dist
+    OZONE_ROOT=$(realpath ../../..)
+  fi
+
+  flekszible generate -t mount:hostPath="$OZONE_ROOT",path=/opt/hadoop -t image:image=apache/ozone-runner:20200420-1 -t ozone/onenode
```
adoroszlai, line 97: Is there any way to avoid hard-coding the image name?

hadoop-ozone/dist/src/main/k8s/examples/testlib.sh
```diff
@@ -0,0 +1,125 @@
...
+   CONTAINER="$1"
+   shift 1 #Remove first argument which was the container name
+
+   # shellcheck disable=SC2206
+   ARGUMENTS=($@)
+
+   kubectl exec -it "${CONTAINER}" -- bash -c 'rm -rf /tmp/report'
+   kubectl exec -it "${CONTAINER}" -- bash -c 'mkdir -p  /tmp/report'
+   kubectl exec -it "${CONTAINER}" -- robot --nostatusrc -d /tmp/report ${ARGUMENTS[@]} || true
+   kubectl cp "${CONTAINER}":/tmp/report/output.xml "result/$CONTAINER-$RANDOM.xml" || true
```
adoroszlai, line 113: `$RANDOM`: TIL, thanks. 👍

.github/workflows/post-commit.yml
```diff
@@ -224,3 +224,60 @@ jobs:
...
+            ${{ runner.os }}-pnpm-
+      - name: checkout to /mnt/ozone
+        run: |
+          sudo chmod 777 /mnt
+          git clone https://github.com/${GITHUB_REPOSITORY}.git /mnt/ozone
+          cd /mnt/ozone
+          git fetch origin "${GITHUB_REF}"
+          git checkout FETCH_HEAD
+          git reset --hard
+      - name: insstall robotframework
```
adoroszlai, line 251: Nit: 
```suggestion
      - name: install robotframework
```

hadoop-ozone/dist/src/main/k8s/examples/testlib.bats
adoroszlai: One more: doesn't this test belong to `src/test`?

hadoop-ozone/dist/src/main/k8s/examples/getting-started/test.sh
```diff
@@ -0,0 +1,38 @@
...
+
+rm -rf result
+
+regenerate_resources
+
+start_k8s_env
+
+execute_robot_test scm-0 smoketest/basic/basic.robot
+
+combine_reports 'getting-started'
```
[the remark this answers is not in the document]
elek: Right, thanks. Removing them.

hadoop-ozone/dist/src/main/k8s/examples/testlib.sh
elek: It can be replaced easily during build time, but today it's possible to execute the tests from `src/min/k8s`, which makes it harder. I can parse it from the `pom.xml` but it's only for development.

I am not sure what is the most simple solution...

hadoop-ozone/dist/src/main/k8s/examples/getting-started/test.sh
```diff
@@ -0,0 +1,38 @@
...
+
+stop_k8s_env
+
+flekszible generate
```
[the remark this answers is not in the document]
elek: Here we have a main problem: some of the Kubernetes examples couldn't be started in the test environment. For example when the example is configured to schedule one datanode per real node, but we have only one node.

Therefore, before the tests the resource files are heavily modified:

 1. instead of using latest image, the current build is mounted to the `/opt/ozone` (similar to the docker tests)
 2. anti-affinity rules are removed (enable to run multiple datanode on the same node)
 3. real persistence is removed

These are executed by the `regenerate_resources` step at the beginning of the test. Which is something like this:

```
flekszible generate -t mount:hostPath="$OZONE_ROOT",path=/opt/hadoop -t image:image=apache/ozone-runner:20200420-1 -t ozone/onenode
```

As you can see we define three new transformations on the fly:

 1. `-t mount:hostPath="$OZONE_ROOT",path=/opt/hadoop` --> use the files from the current build
 2. `-t image:image=apache/ozone-runner:20200420-1`, use standard runner
 3.  `-t ozone/onenode` enable to schedule more datanode to the same node

The line which is commented by you:

```
flekszible generate
```

Is 100% optional. After the test, it restores the original state of the files. It can be added to the stop_k8s_env (as it can always be added). 

But based on the experience with the docker `test.sh` files, I would prefer to use more explicit lines in the test.sh.

As `test.sh` files are read frequently but modified only by a few times, I tend to make it slightly more verbose, but easier to understand (you can assume that `stop_k8s_env` stops kubernetes pods, but I wouldn't like to hide any hidden functionlaty there. 

But theses are just my thought, as it's a beginning of new test, I am really open to modify to any direction.

hadoop-ozone/dist/src/main/k8s/examples/testlib.sh
adoroszlai: OK, I'm fine with including it here.  Maybe a note near the version property in `pom.xml` would be nice to avoid someone forgetting to update this.

hadoop-ozone/dist/src/main/k8s/examples/getting-started/test.sh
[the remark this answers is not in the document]
adoroszlai: Thanks for confirming my guess.

I'm fine with making `test.sh` more verbose at the cost of some duplication.  However, I propose extracting this optional `flekszible generate` to a function, and let it check if it's being run from `src` dir, to address my first point.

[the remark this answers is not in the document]
elek: > and let it check if it's being run from src dir, to address my first point

> I ran tests from target/..., had some errors and tried to find out what's wrong.

I am not sure what was the problem in your case. It supposed to work from the target directory. In fact `hadoop-ozone/dev-support/checks/kubernetes.sh` executes all the tests from target by default.

`apache/ozone:0.6.0-SNAPSHOT` is used because during the release it becomes `apache/ozone:0.6.0` which is the distributed container.

`./test.sh` (with the first flekszible generation) replaces all the images with a dev version (ozone-runner + mount).

It seems to be working for me and passed on the github CI, but if you see any error, please let me know as it should work everywhere (do you have the latest released flekszible?)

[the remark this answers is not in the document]
adoroszlai: Let me clarify: I think the script would have worked fine, but I tried to use k3d (k3s in docker) instead of plain k3s.  Since the local dir was not available in the container, it failed.

The confusing part was that the resource files did not mention volume mounts, since they were converted back at the end.  So only by looking at `regenerate_resources` in `testlib` did I realize what's wrong.

adoroszlai:
Thanks @elek for updating the patch.  It's fine as it is.

.github/workflows/post-commit.yml
```diff
@@ -238,3 +238,61 @@ jobs:
...
+          key: ${{ runner.os }}-pnpm-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-
+      - name: checkout to /mnt/ozone
+        run: |
+          sudo chmod 777 /mnt
+          git clone https://github.com/${GITHUB_REPOSITORY}.git /mnt/ozone
+          cd /mnt/ozone
+          git fetch origin "${GITHUB_REF}"
+          git checkout FETCH_HEAD
```
adoroszlai, line 263: Can you please update this with the fix from fd47f915c65512de6618811e6c247871066a8351?

hadoop-ozone/dist/src/main/k8s/examples/getting-started/test.sh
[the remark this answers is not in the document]
elek: Yes, that's what came into my mind, too.

Today, it's not possible to run test on remote clusters (k3d is a remote cluster) as local mount doesn't work.

I can introduce some options in the future (would be great to execute tests on real environments, too).

For now, I will add a log message to the `regenerate_resource` to print a warning to the stdout.

.github/workflows/post-commit.yml
elek: fair point. Replaced with the new version. (One more reason to move checkout logic to a separated script like https://github.com/elek/hadoop-ozone/blob/smart-checkout/.github/checkout.sh)

adoroszlai:
Thanks @elek for improving it further.

PR #1224, opened 2020-07-20
HDDS-3991. Ignore protobuf lock files

elek:

HDDS-3595 introduced a new build time check to make sure protobuf files are changed only with backward compatible way. A lock file which contains the structure of the proto files are committed to the repository and during the build the new state of the proto files are compared with the committed version.

Unfortunately the plugin always generate a new lock file, even if it should be updated only after the releases. To make it more safe (and less confusing) I suggest putting the lock files to the git ignore list

https://issues.apache.org/jira/secure/RapidBoard.jspa?rapidView=260

## How was this patch tested?

`git status`


adoroszlai:
@elek how do we manually commit proto.lock updates?

elek:
> @elek how do we manually commit proto.lock updates?

Files can be committed even if they are on the `.gitignore` list with the `--force` option of `git add`.

But there is another problem here: git ignores ignore the changes only until they are added to the repo. If they are committed they will be tracked independent of the `.gitignore` file.

I updated the patch to follow a different approach: the check is executed from `target/classes` after the proto files and the lock files are copied (by maven resporce and proto plugins). At that time the update of the proto file is safe as target is ignored, but still we use the committed version of the lock file (which is copied).

Execute the latest patch and will see:

```
341   │ [INFO] --- proto-backwards-compatibility:1.0.5:backwards-compatibility-check (default) @ hadoop-hdds-interface-client ---
 342   │ [INFO] Backwards compatibility check passed.
```

(When `proto.lock` is missing, a different message is printed out, sg. about initialization).

bharatviswa504:
Looks like PR needs rebase.

elek:
Thanks the review @bharatviswa504 

Instead of the rebase I reverted the (accidentally) modified proto.lock on the master ;-)

I am merging it as we have full green build.

PR #1227, opened 2020-07-21
HDDS-3993. Create volume required for S3G during OM startup.

bharatviswa504:

Create volume required for S3G during Om startup

https://issues.apache.org/jira/browse/HDDS-3993

## How was this patch tested?

Tested to verify on docker-compose om-ha-s3 cluster, whether s3v volume is created or not. 
```
bash-4.2$ ozone sh volume list o3://id1
{
  "metadata" : { },
  "name" : "s3v",
  "admin" : "hadoop",
  "owner" : "hadoop",
  "creationTime" : "2020-07-21T00:01:53.739Z",
  "modificationTime" : "2020-07-21T00:01:53.739Z",
  "acls" : [ {
    "type" : "USER",
    "name" : "hadoop",
    "aclScope" : "ACCESS",
    "aclList" : [ "ALL" ]
  }, {
    "type" : "GROUP",
    "name" : "users",
    "aclScope" : "ACCESS",
    "aclList" : [ "ALL" ]
  } ],
  "quota" : 1152921504606846976
}
```


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3502,4 +3516,95 @@ public boolean getEnableFileSystemPaths() {
...
+    String s3VolumeName = HddsClientUtils.getS3VolumeName(configuration);
+    String dbVolumeKey = metadataManager.getVolumeKey(s3VolumeName);
+
+    if (!s3VolumeName.equals(OzoneConfigKeys.OZONE_S3_VOLUME_NAME_DEFAULT)) {
+      LOG.warn("Make sure that all S3Gateway use same volume name." +
+          " Otherwise user need to manually create/configure Volume " +
+          "configured by S3Gateway");
+    }
+    if (!metadataManager.getVolumeTable().isExist(dbVolumeKey)) {
+      long transactionID = 1L;
```
arp7, line 3534: What is the implication of hard-coding the transaction id = 1, especially for existing clusters where the s3 volume may not exist?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3502,4 +3516,95 @@ public boolean getEnableFileSystemPaths() {
...
+      long objectID = OMFileRequest.getObjIDFromTxId(transactionID);
```
arp7, line 3535: I looked into this function. I think hard-coding the transaction ID may be risky in a cluster where there are already some Ratis transactions however the s3v volume was not yet created. Perhaps we can set it to a really high value, something that is very unlikely to be hit in a real cluster. Like `(MAX_ULONG - 1) >> 8`

swagle:
Can we add a unit test?

bharatviswa504:
Thank You @arp7, @adoroszlai and @swagle  for the review.
Addressed review comments, fixed test cases and also added a test.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3502,4 +3517,95 @@ public boolean getEnableFileSystemPaths() {
...
+    String dbVolumeKey = metadataManager.getVolumeKey(s3VolumeName);
+
+    if (!s3VolumeName.equals(OzoneConfigKeys.OZONE_S3_VOLUME_NAME_DEFAULT)) {
+      LOG.warn("Make sure that all S3Gateway use same volume name." +
+          " Otherwise user need to manually create/configure Volume " +
+          "configured by S3Gateway");
+    }
+    if (!metadataManager.getVolumeTable().isExist(dbVolumeKey)) {
+      long transactionID = (Long.MAX_VALUE - 1) >> 8;
+      long objectID = OMFileRequest.getObjIDFromTxId(transactionID);
```
arp7, line 3535: Let's assert in a test case that the generated object ID is correct.

hadoop-hdds/interface-server/src/main/proto/proto.lock
```diff
@@ -457,6 +457,12 @@
                 "name": "storageReport",
                 "type": "StorageReportProto",
                 "is_repeated": true
+              },
+              {
+                "id": 2,
```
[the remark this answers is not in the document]
bharatviswa504: Yes, because proto.lock file getting updated. It is causing issues. I have opened a Jira but it is said that we should consider updating proto.lock files should be updated only at end of release.

bharatviswa504:
Updated test case and addressed review comments.

PR #1230, opened 2020-07-21
HDDS-2949: mkdir : store directory entries in a separate table

rakeshadr:

All the directories from the path prefix should be created on 'prefixTable'

## How was this patch tested?

Verified using TestOMDirectoryCreateResponse.java and TestOzonePrefixTable.java unit test cases.
Note: As this is an incremental change, it is expected to have failures related to file, key ops test cases. Will fix this later through subsequent sub-tasks.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMMetadataManager.java
```diff
@@ -116,6 +116,25 @@
    */
   String getOzoneDirKey(String volume, String bucket, String key);
 
+  /**
+   * Given a volume, bucket and a key, return the corresponding DB key.
+   *
```
linyiqun, line 121: The Comment needs to be updated, here we don't pass volume, bucket and key. The same problem in getOzoneLeafNodeKey method.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
```diff
@@ -170,37 +171,37 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       // Need to check if any files exist in the given path, if they exist we
       // cannot create a directory with the given key.
+      // Verify the path against prefix table
       OMFileRequest.OMPathInfo omPathInfo =
-          OMFileRequest.verifyFilesInPath(omMetadataManager, volumeName,
```
linyiqun, line 174: As we store the dir in prefix table now, we need to acquire PREFIX_LOCK instead of BUCKET_LOCK.
`
 volumeName, bucketName);
`

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzonePrefixTable.java
```diff
@@ -0,0 +1,183 @@
...
+      prefixKeys.remove(value.getKey());
+    }
+    Assert.assertEquals("Unexpected prefix table entries!", 0, count);
+
+    // verify entries in key table
+    TableIterator<String, ? extends
+            Table.KeyValue<String, OmKeyInfo>> keyTableItr =
+            omMgr.getKeyTable().iterator();
+    while (keyTableItr.hasNext()) {
+      fail("Shouldn't add any entries in KeyTable!");
```
linyiqun, line 111: Can you add an additionally test cases for current entries in KeyTable? Now the new file key would be like objId/fileName rather than /volume/bucket/keyName.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMMetadataManager.java
rakeshadr: Will fix it in new commit.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OmMetadataManagerImpl.java
```diff
@@ -469,6 +469,22 @@ public String getOzoneDirKey(String volume, String bucket, String key) {
...
+  @Override
+  public String getOzoneLeafNodeKey(long parentObjectId, String leafNodeName) {
+    StringBuilder builder = new StringBuilder();
+    builder.append(parentObjectId);
+    builder.append(OM_KEY_PREFIX).append(leafNodeName);
+    return builder.toString();
+  }
+
+  @Override
+  public String getOzonePrefixKey(long parentObjectId, String prefixName) {
```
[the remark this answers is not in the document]
rakeshadr: Agreed, will fix it in new commit

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzonePrefixTable.java
rakeshadr: #verifyPrefixKey() function is doing the same by fetching entries using "omMgr.getPrefixTable().get(dbKey);" call.
Are you suggesting something different from this verification check?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmPrefixInfo.java
```diff
@@ -37,7 +37,14 @@
...
+
   private String name;
+  private String volumeName;
+  private String bucketName;
+
+  private long creationTime;
+  private long modificationTime;
   private List<OzoneAcl> acls;
 
   public OmPrefixInfo(String name, List<OzoneAcl> acls,
```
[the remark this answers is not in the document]
rakeshadr: Agreed, will fix it new commit

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java
rakeshadr: Very good point. Will improve this logic.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -254,4 +281,141 @@ public static void addKeyTableCacheEntries(
...
+        continue;
+      }
+
+      // For example, /vol1/buck1/a/b/c/d/e/f/file1.txt
+      // 1. Do lookup on directoryTable. If not exists goto next step.
+      // 2. Do look on keyTable. If not exists goto next step.
+      // 3. Add 'sub-dir' to missing parents list
+      String dbNodeName = omMetadataManager.getOzonePrefixKey(lastKnownParentId,
+              fileName);
+      OmPrefixInfo omPrefixInfo = omMetadataManager.getPrefixTable().
```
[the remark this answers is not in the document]
rakeshadr: Very good point. Based on the upgrade path discussion, the plan is to introduce new directory table. In that case, all the dir ops will be under BUKCET_LOCK. Will update the patch soon with that changes.

rakeshadr (commented):
Thanks a lot @bharatviswa504,  @linyiqun for the reviews

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzonePrefixTable.java
linyiqun: @rakeshadr , actually I mean we need to add logic to verify new entry on key table as well in utl. Current unit test only check the dir entry in prefix table.

Key table format before:
KeyName:
/vol1/buck1/a/b/c/d/file1

Now like this:
1028/file1

rakeshadr: Thanks for the details. I have created new PR https://github.com/apache/hadoop-ozone/pull/1404 and incorporated your comments. Please review it.

rakeshadr:
Thanks @bharatviswa504 @linyiqun  for the reviews. I am sorry for multiple PRs. I have created new PR https://github.com/apache/hadoop-ozone/pull/1404 now. Basically, the idea is to create new set of classes V1 for easy migration of this feature branch with rolling upgrade. Also, I have created new table Directory_Table instead of Prefix Table. This is to simplifies the caching logic later.

Please review this PR and thanks for your patience!

mukul1987:
@rakeshadr Can we close this PR ? as there is a new one available ?

PR #1231, opened 2020-07-21
HDDS-3994. Make retry policy can be set by configuration.

maobaolong:

Write object when met exception can be slower than before, and i realized HDDS-3350 using a newer retry policy replace the existing retry policy, and the old configure items has been removed, so the behavior has been changed, i guess that is a behavior incompatible, so i trying to use a policy framework to let users can specify it to which policy they want, default to the `RetryLimitedPolicy`

HDDS-3994

## How was this patch tested?

Set configure key `dfs.ratis.client.request.retry.policy` to `org.apache.hadoop.hdds.ratis.retrypolicy.RetryLimitedPolicyCreator` or   `org.apache.hadoop.hdds.ratis.retrypolicy.RequestTypeDependentRetryPolicyCreator`

set the OrderedAsync logLevel to TRACE by execute the following command.

curl 'http://0.0.0.0:19878/logLevel?log=org.apache.ratis.client.impl.OrderedAsync&level=TRACE'

put a file to s3g and see the behavior of the retry policy by view the s3g log.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
@@ -112,6 +112,18 @@ public final class ScmConfigKeys {
   public static final String
       DFS_CONTAINER_RATIS_LEADER_PENDING_BYTES_LIMIT_DEFAULT = "1GB";
 
+  public static final String DFS_RATIS_CLIENT_REQUEST_RETRY_POLICY =
+      "dfs.ratis.client.request.retry.policy";
+
+  public static final String DFS_RATIS_CLIENT_REQUEST_MAX_RETRIES_KEY =
+      "dfs.ratis.client.request.max.retries";
+  public static final int DFS_RATIS_CLIENT_REQUEST_MAX_RETRIES_DEFAULT = 180;
+  public static final String DFS_RATIS_CLIENT_REQUEST_RETRY_INTERVAL_KEY =
+      "dfs.ratis.client.request.retry.interval";
+  public static final TimeDuration
+      DFS_RATIS_CLIENT_REQUEST_RETRY_INTERVAL_DEFAULT =
+      TimeDuration.valueOf(1000, TimeUnit.MILLISECONDS);
+
   public static final String DFS_RATIS_SERVER_RETRY_CACHE_TIMEOUT_DURATION_KEY =
       "dfs.ratis.server.retry-cache.timeout.duration";
   public static final TimeDuration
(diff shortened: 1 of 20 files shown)


maobaolong:
@lokeshj1703 Please take a look at this PR, thank you.

maobaolong:
@ChenSammi Please take a look, thank you very mush.

maobaolong:
@xiaoyuyao @lokeshj1703 Thanks for the discussion and glad to make consensual for us. Now, I change the default policy to the newer RequestTypeDependentRetryPolicy, PTAL.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
```diff
@@ -112,6 +112,18 @@
...
+
+  public static final String DFS_RATIS_CLIENT_REQUEST_MAX_RETRIES_KEY =
+      "dfs.ratis.client.request.max.retries";
+  public static final int DFS_RATIS_CLIENT_REQUEST_MAX_RETRIES_DEFAULT = 180;
+  public static final String DFS_RATIS_CLIENT_REQUEST_RETRY_INTERVAL_KEY =
+      "dfs.ratis.client.request.retry.interval";
+  public static final TimeDuration
+      DFS_RATIS_CLIENT_REQUEST_RETRY_INTERVAL_DEFAULT =
+      TimeDuration.valueOf(1000, TimeUnit.MILLISECONDS);
+
```
lokeshj1703, line 126: Can we add these configs to RatisClientConfig class? I would also suggest renaming some of the configs to match the existing configs like hdds.ratis.client.retrylimited.retry.interval or sth similar. This would make it easier to distinguish between the configs.

maobaolong: Yeah，Sure.

maobaolong:
@lokeshj1703 Thanks for your review. I put the config into ratisclientconfig, and changed the names of limtedretrypolicy ralated configs, it make the old config key cannot works from then on, so are you sure we should modify the key name of limitedretrypolicy?

lokeshj1703:
> it make the old config key cannot works from then on, so are you sure we should modify the key name of limitedretrypolicy?

@maobaolong I think we should rename these configs before GA. The old config names are very abstract and seem like applies to all retry policies.

maobaolong:
@lokeshj1703 

> I think we should rename these configs before GA. The old config names are very abstract and seem like applies to all retry policies.

OK, this fine. Thank you for your review and suggestion, now the CI checks have been passed.

lokeshj1703:
@maobaolong Thanks for the contribution! I have committed the PR to master branch.

PR #1232, opened 2020-07-21
HDDS-3431. Enable TestContainerReplication test cases

sadanand48:
Fix TestContainerReplication testcases.

https://issues.apache.org/jira/browse/HDDS-3431

## How was this patch tested?
Manually . Tested by creating a custom workflow to run 20 iterations . 20/20 [passed](https://github.com/sadanand48/hadoop-ozone/actions/runs/176961558)

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/container/TestContainerReplication.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/container/TestContainerReplication.java
@@ -37,6 +37,7 @@ import org.apache.hadoop.hdds.scm.XceiverClientGrpc;
 import org.apache.hadoop.hdds.scm.XceiverClientSpi;
 import org.apache.hadoop.hdds.scm.pipeline.MockPipeline;
 import org.apache.hadoop.hdds.scm.pipeline.Pipeline;
+import org.apache.hadoop.hdds.server.events.EventPublisher;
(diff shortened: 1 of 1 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/container/TestContainerReplication.java
```diff
@@ -150,6 +154,8 @@ public void testContainerReplication() throws Exception {
         -> destinationDatanodeDatanodeStateMachine.getSupervisor()
         .getReplicationRequestCount() > 0, 1000, 20_000);
 
+    Assert.assertEquals(1, destinationDatanodeDatanodeStateMachine.
+            getSupervisor().getReplicationSuccessCount());
```
adoroszlai, line 158: Is there any guarantee that if we have a request (ie. requestCount > 0), then it is also successfully completed?  I wouldn't think so.

I think we should wait for success:

```suggestion
        .getReplicationSuccessCount() > 0, 1000, 20_000);
```

sadanand48:
Thanks @adoroszlai for reviewing this. Sorry for the late reply . I had followed this comment ` Since datanode commands are added through event queue, onMessage method
 should take care of adding commands to command queue ` from `SCMNodeManager`. The test still works for me and i have also run it [here](https://github.com/sadanand48/hadoop-ozone/actions/runs/176961558).

PR #1234, opened 2020-07-22
HDDS-3996. Missing TLS client configurations to allow ozone.grpc.tls.…

xiaoyuyao:
…enabled.

Adding proper TLS client configuration for Client to talk to SSL enabled GRPC server (Ratis).

https://issues.apache.org/jira/browse/HDDS-3996

## How was this patch tested?
Testing with real cluster and ensure the Ratis Three Pipeline can be created by SCM when ozone.grpc.tls.enabled=True. 
This needs additional fix from HDDS-3997 to successfully pass SSL handshake between GRPC/Netty client/server.


jnp:
+1 for the patch.

xiaoyuyao:
Thansk @jnp  for the review. I will merge it shortly.

PR #1235, opened 2020-07-22
HDDS-3997. Ozone certificate needs additional flags and SAN extension…

xiaoyuyao:
… for GRPC TLS.

Adding SAN extension for SCM issued certificate and add DN certificate with sign/encrypt flag to meet requirement of SSL handshake with GRPC/Netty. 

https://issues.apache.org/jira/browse/HDDS-3997

## How was this patch tested?
Test with real cluster before patch HDDS-3996 and HDDS-3997. Certificate does not have the required extension and flag, which fail the SSL handshake for Ratis operations like addGroup, leader election when ozone.grpc.tls.enabled is true. 

After the patch, verified the RATIS-Three pipeline can be created successfuly without SSL error.


jnp:
+1 for the patch.

xiaoyuyao:
Thanks @jnp for the review. I will merge the PR shortly.

PR #1236, opened 2020-07-22
HDDS-4000. Split acceptance tests to reduce CI feedback time

adoroszlai:

Reduce total CI time:

1. Split acceptance tests into 3 groups executed in parallel using GitHub Actions matrix build.  Groups are defined to have similar run times.  `misc` has several unrelated items, some of which can be moved to another group if it grows too long.
 * secure: 50m 36s
 * unsecure: 48m 23s
 * misc: 50m 6s
2. Merge some integration test groups, both to make them similar length and to reduce the number of checks.  Currently some groups have much shorter run time.  Since `client` is the longest one, others can take more time, too.  New run times:
 * client: 54m 37s
 * filesystem-hdds: 43m 58s
 * ozone: 55m 37s
3. Start integration checks earlier: do not wait for `build` check to finish.  Both acceptance and integration checks perform a full build anyway.  The only reason to wait was that if there is a compile error, we wanted to avoid spawning too many checks that fail for the same reason.  But compile error is the least frequent CI problem, checkstyle, rat and various test failures are much more common.  So being optimistic helps in the most common scenarios.  (Ideally we should cache the results of `build` for later stages.  Then we could both depend on it and avoid duplicate builds.)

https://issues.apache.org/jira/browse/HDDS-4000

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/897297500


elek:
Nice idea (and nice Jira number BTW ;-) ).

I like this tagging system especially because it can support optional tests. For example, I can create some optional ..._test.sh files with specific tags (like `restart-tests`) which are excluded by default pr runs but run by the daily builds.

<brainstorming-mode on>

It can be easy to forget adding `#suite:misc` to a new file. What do you think to support execution all the test files *without* suite tags. (--> misc is the default suite)?

<brainstorming-mode off>

adoroszlai:
> I like this tagging system especially because it can support optional tests. For example, I can create some optional ..._test.sh files with specific tags (like `restart-tests`) which are excluded by default pr runs but run by the daily builds.

Note that they still need to be named exactly `test.sh`.

> It can be easy to forget adding `#suite:misc` to a new file.

Who writes such files from scratch? :)

> What do you think to support execution all the test files _without_ suite tags. (--> misc is the default suite)?

Agree, this is more safe.

hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/test.sh
```diff
@@ -15,6 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
```
cxorm, line 18: Seems we could not add this line here :wink:

adoroszlai:

> BTW, I don't know why the overall time (from 1:10) is increased recently. Do you have any information about that?

We now have 4x FS tests ((o3fs, ofs) x (bucket, link)) and 2x S3 tests (bucket, link).  Upgrade is also new.

PR #1239, opened 2020-07-22
HDDS-3827. Intermittent failure in TestKeyManagerUnit#listMultipartUploads

adoroszlai:

1. Fix intermittent failure in `TestKeyManagerUnit#listMultipartUploads`.  If the test is executed quickly, the timestamps may be the same, so check for "not before" instead of "after".
2. Cleanup test directory.  While repeatedly executing the test I found that 1.1MB of data was left over for each iteration.

https://issues.apache.org/jira/browse/HDDS-3827

## How was this patch tested?

Regular: https://github.com/adoroszlai/hadoop-ozone/runs/898216776
100x single test: https://github.com/adoroszlai/hadoop-ozone/runs/898217969


bharatviswa504:
Thank You @adoroszlai for the fix.

adoroszlai:

PR #1242, opened 2020-07-22
HDDS-4007. Generate encryption info for the bucket outside bucket lock.

bharatviswa504:

Generate encryption info for the bucket outside bucket lock

https://issues.apache.org/jira/browse/HDDS-4007

## How was this patch tested?

Perf improvement, it is order of code is change.


bharatviswa504:
Thank You @adoroszlai for the review and test.

PR #1243, opened 2020-07-22
HDDS-4008. Recon should fallback to ozone.om.service.ids when the internal service id is not defined.

avijayanhwx:
Recon connects to OM via RPC using the "ozone.om.internal.service.id" to get updates. If the above config is not defined, but the ozone.om.service.ids is defined, Recon should use the latter as a fallback. Currently, a single Recon instance supports only 1 OM HA cluster at a time. Hence, if multiple ids are defined and the internal service ID is not defined, we throw an error.

https://issues.apache.org/jira/browse/HDDS-4008

## How was this patch tested?
Manually tested.
Added unit test.


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -527,4 +529,45 @@ public static void validateKeyName(String keyName)
...
+      if (omServiceIds.size() > 1) {
+        throw new IOException(String.format(
+            "More than 1 OzoneManager ServiceID (ozone.om.service.ids) " +
+                "configured : %s, but ozone.om.internal.service.id is not " +
+                "configured.", omServiceIds.toString()));
+      }
+    } else {
+      omServiceIds = Collections.singletonList(localOMServiceId);
+    }
+
```
bharatviswa504, line 563: I think we also need to check if localServiceID value is one of the ozone.om.service.ids value.
To find if there is a mistake in the config.

hadoop-ozone/recon/src/main/resources/webapps/recon/ozone-recon-web/src/views/overview/overview.tsx
```diff
@@ -180,7 +180,7 @@ export class Overview extends React.Component<Record<string, object>, IOverviewS
             <OverviewCard loading={loading} title='Buckets' data={buckets.toString()} icon='folder-open'/>
           </Col>
           <Col xs={24} sm={18} md={12} lg={12} xl={6}>
-            <OverviewCard loading={loading} title='Keys' data={keys.toString()} icon='file-text'/>
+            <OverviewCard loading={loading} title='Keys (Estimated)' data={keys.toString()} icon='file-text'/>
```
bharatviswa504, line 183: Not related to this change??

avijayanhwx: I made this trivial change to cover for HDDS-4009 until that is fixed.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
avijayanhwx: Thanks @bharatviswa504. Will fix this.

hadoop-ozone/recon/src/main/resources/webapps/recon/ozone-recon-web/src/views/overview/overview.tsx
bharatviswa504: That is fine. But a new Jira should be nicer here.
I am fine with it.

avijayanhwx:
Thank you for the reviews @bharatviswa504 & @vivekratnavel.

PR #1244, opened 2020-07-22
HDDS-3999. OM Shutdown when Commit part tries to commit the part, after abort upload.

bharatviswa504:
This is caused by HDDS-3685. And the tests have not caught this because there is no test for the below scenario in our test suite.
1. Initiate MPU
2. Part Upload in progress
3. Abort Upload
4. Part upload close to commit the part.

https://issues.apache.org/jira/browse/HDDS-3999

## How was this patch tested?
Added a test and confirmed now OM does not crash.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.BitSet;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -1911,10 +1912,45 @@ public void testAbortUploadFail() throws Exception {
...
+    bucket.abortMultipartUpload(keyName, omMultipartInfo.getUploadID());
+
+    try {
+      ozoneOutputStream.close();
+    } catch (IOException ex) {
+      assertTrue(ex instanceof OMException);
+      assertEquals(NO_SUCH_MULTIPART_UPLOAD_ERROR,
+          ((OMException) ex).getResult());
+    }
+
```
maobaolong, line 1951: Please remove the extra blank line.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -1911,10 +1912,45 @@ public void testAbortUploadFail() throws Exception {
...
+    byte[] data = "data".getBytes(UTF_8);
+    OzoneOutputStream ozoneOutputStream = bucket.createMultipartKey(keyName,
+        data.length, 1, omMultipartInfo.getUploadID());
+    ozoneOutputStream.write(data, 0, data.length);
+
+    // Abort before completing part upload.
+    bucket.abortMultipartUpload(keyName, omMultipartInfo.getUploadID());
+
+    try {
+      ozoneOutputStream.close();
```
maobaolong, line 1945: Please add a fail() here

bharatviswa504:
>As the parameter omMultipartKeyInfo of S3MultipartUploadCommitPartResponse is annotation to Nullable, so, we should >consider possible NPE, for this, i have two suggestion
>Mark omMultipartKeyInfo Nonnull, and keep the callee never give a null to it.

When NO_SUCH_MULTIPART_UPLOAD_ERROR omMultipartKeyInfo will be null. So, in this case we should not use omMultipartKeyInfo. The previous code has bug that caused this issue.

>Also give a null pointer check for omMultipartKeyInfo int S3MultipartUploadCommitPartResponse#addToDBBatch, to avoid >NPE.
And when Status is OK, omMultipartKeyInfo will not be null, so we don't need an additional null check here. We have already checked the Status, and we access omMultipartKeyInfo only when the status is OK. I don't see any possibility when the status is OK, omMultipartKeyInfo to be null. Let me know if you still see any possibility. We can add a null check, but trying to understand here if there is any missed case.

Tagged omMultipartKeyInfo as nullable because when error is NO_SUCH_MULTIPART_UPLOAD_ERROR, the value will be null.

bharatviswa504:
>BTW, i don't think i have a abortMPU operation by executed the following command

>$ zcat logs/om-audit-host-*log.gz | grep "COMPLETE_MULTIPART_UPLOAD" | grep -v "multipartList"
>$ zcat logs/om-audit-host-*log.gz | grep "COMPLETE_MULTIPART_UPLOAD" | wc -l 
>188

Might be this is one of the scenario to catch this error.
On thinking more this can happen when the commit part tries to commit a part, after complete multipart upload.

arp7:
Thanks for contributing this patch @bharatviswa504  and thanks for the review @maobaolong. I plan to commit this shortly. If there's any remaining unaddressed comments we can address them in a followup Jira.

PR #1246, opened 2020-07-23
HDDS-4016. Change AbortMultipartUpload's audit log type to ABORT_MULTIPART_UPLOAD

captainzmc:

Currently, the audit log of AbortMultipartUpload is COMPLETE_MULTIPART_UPLOAD, which cannot be expressed clearly. We need to change it to ABORT_MULTIPART_UPLOAD.

https://issues.apache.org/jira/browse/HDDS-4016

## How was this patch tested?

[existing UT](https://github.com/captainzmc/hadoop-ozone/runs/901459583)


bharatviswa504:
@captainzmc This old code is no more used. We have recently closed another similar PR #1193  which is opened to fix the same. 

This is the parent Jira https://issues.apache.org/jira/browse/HDDS-2353 for the cleanup.

captainzmc:
Thanks for @bharatviswa504 's  tip.  Close this PR

PR #1249, opened 2020-07-23
HDDS-4017. Acceptance check may run against wrong commit

adoroszlai:

Use commit SHA to checkout a specific commit on `push` and `schedule` events.  Keep using name for `pull_request` events (eg. `refs/pull/1234/merge`), because merge commit may not be available by SHA.

Use `github` context instead of environment variables.  Github substitutes these values before running the command, so we can see them in the log.  Comparison:

```
git clone https://github.com/${GITHUB_REPOSITORY}.git /mnt/ozone
cd /mnt/ozone
git fetch origin "${GITHUB_REF}"
git checkout FETCH_HEAD
```

vs.

```
git clone https://github.com/adoroszlai/hadoop-ozone.git /mnt/ozone
cd /mnt/ozone
git fetch origin "refs/heads/HDDS-4017"
git checkout FETCH_HEAD
```

https://issues.apache.org/jira/browse/HDDS-4017

## How was this patch tested?

`push` event:
https://github.com/adoroszlai/hadoop-ozone/runs/903007668#step:4:8

`pull_request` event:
https://github.com/apache/hadoop-ozone/pull/1249/checks?check_run_id=903076398#step:4:6

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -134,9 +134,13 @@ jobs:
       - name: checkout to /mnt/ozone
         run: |
           sudo chmod 777 /mnt
-          git clone https://github.com/${GITHUB_REPOSITORY}.git /mnt/ozone
+          git clone 'https://github.com/${{ github.repository }}.git' /mnt/ozone
           cd /mnt/ozone
-          git fetch origin "${GITHUB_REF}"
+          if [[ '${{ github.event_name }}' == 'pull_request' ]]; then
+            git fetch --verbose origin '${{ github.ref }}'
+          else
+            git fetch --verbose origin '${{ github.sha }}'
+          fi
           git checkout FETCH_HEAD
           git reset --hard
       - name: run a full build


adoroszlai:
> Based on my understanding for pull requests the merge commit points to a potential merge. In case of using github `checkout ` action, not the `github.ref` but the merge commit is used.

Both `${GITHUB_REF}` and `${{ github.ref }}` point to the merge commit for PRs, and that's fine.

The problem is with push builds, because in that case both refs are something like `refs/heads/master`.  The commit this points to can change in the meantime, so we need to use the SHA.  (Your original solution was fine for push commits, my [fix for PRs](https://github.com/apache/hadoop-ozone/pull/1050/commits/31b2f0e3a98efe91635ce114f8b99d6387356623) introduced this problem.)

> Do you have any example when it was wrong?

The Jira issue HDDS-4017 has an example.

elek:
>  The commit this points to can change in the meantime, so we need to use the SHA.

Thanks to explain it. Now I got it. :bulb: :bulb:  :bulb: 

Let me merge it.

adoroszlai:

PR #1250, opened 2020-07-23
HDDS-4018. Datanode log spammed by NPE

adoroszlai:

Avoid calling `task.await()` without first calling `task.execute()`.  The former uses variable initialized in the latter.

https://issues.apache.org/jira/browse/HDDS-4018

## How was this patch tested?

Logs from acceptance test are normal size:
https://github.com/adoroszlai/hadoop-ozone/runs/902827380

Without the patch they grow to 800-1000MB:
https://github.com/apache/hadoop-ozone/runs/898489257

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
@@ -436,18 +436,17 @@ public class StateContext {
         task.onEnter();
       }
 
-      if (isThreadPoolAvailable(service)) {
-        task.execute(service);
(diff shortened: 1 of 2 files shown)


runzhiwang:
LGTM +1

runzhiwang:
@adoroszlai Hi, There is one more comment.
The reason of so many logs when NPE happen is when NPE happen, the thread did not sleep and start another while loop immediately.  I think we should also avoid this.
image

adoroszlai:
Thanks @runzhiwang and @xiaoyuyao for the review.

> The reason of so many logs when NPE happen is when NPE happen, the thread did not sleep and start another while loop immediately. I think we should also avoid this.

@runzhiwang can you please file a new issue for that?

runzhiwang:
@adoroszlai  Thanks. I will file a new issue for that.

PR #1252, opened 2020-07-24
HDDS-4010. S3G startup fails when multiple service ids are configured.

bharatviswa504:

This Jira is to fix this TODO.

**OzoneServiceProvider.java L59:

      // HA cluster.
      //For now if multiple service id's are configured we throw exception.
      // As if multiple service id's are configured, S3Gateway will not be
      // knowing which one to talk to. In future, if OM federation is supported
      // we can resolve this by having another property like
      // ozone.om.internal.service.id.
      // TODO: Revisit this later.
      if (serviceIdList.size() > 1) {
        throw new IllegalArgumentException("Multiple serviceIds are " +
            "configured. " + Arrays.toString(serviceIdList.toArray()));**

https://issues.apache.org/jira/browse/HDDS-4010

## How was this patch tested?

HDDS-4008 added test for configuration, this Jira reuses the same.
Existing ozone-om-ha docker running s3 test suite should cover S3 tests.

### DIFF
--- a/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneServiceProvider.java
+++ b/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneServiceProvider.java
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.ozone.s3;
 
+import org.apache.hadoop.hdds.conf.ConfigurationException;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.ozone.OmUtils;
@@ -28,11 +29,10 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
 
-import java.util.Arrays;
+import java.io.IOException;
 import java.util.Collection;
 
 import static org.apache.hadoop.ozone.om.OMConfigKeys.OZONE_OM_NODES_KEY;
-import static org.apache.hadoop.ozone.om.OMConfigKeys.OZONE_OM_SERVICE_IDS_KEY;
 
 /**
  * This class creates the OM service .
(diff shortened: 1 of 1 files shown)


hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneServiceProvider.java
```diff
@@ -49,34 +49,24 @@
...
   @PostConstruct
   public void init() {
-    Collection<String> serviceIdList =
-        conf.getTrimmedStringCollection(OZONE_OM_SERVICE_IDS_KEY);
-    if (serviceIdList.size() == 0) {
-      // Non-HA cluster
+    try {
+      omserviceID = OmUtils.getOzoneManagerServiceId(conf);
+    } catch (IOException ex) {
+      throw new ConfigurationException(ex.getMessage(), ex);
```
xiaoyuyao, line 55: IIUC: if ozone-site.xml has more than one service id configured, s3g startup will fail as well based on OmUtils.getOzoneMangerServiceId.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneServiceProvider.java
```diff
@@ -49,34 +49,24 @@
...
-        String serviceId = serviceIdList.iterator().next();
-        Collection<String> omNodeIds = OmUtils.getOMNodeIds(conf, serviceId);
-        if (omNodeIds.size() == 0) {
-          throw new IllegalArgumentException(OZONE_OM_NODES_KEY
-              + "." + serviceId + " is not defined");
-        }
-        omServiceAddr = new Text(OzoneS3Util.buildServiceNameForToken(conf,
-            serviceId, omNodeIds));
-        omserviceID = serviceId;
+      Collection<String> omNodeIds = OmUtils.getOMNodeIds(conf, omserviceID);
```
xiaoyuyao, line 63: existing om-ha-s3g or om-ha docker-compose don't have multiple service id configured. Can we add a unit or acceptance test to validate the behavior?

elek:
It seems to be related to #1149 

See my related comments:

https://github.com/apache/hadoop-ozone/pull/1149#issuecomment-664814018

https://github.com/apache/hadoop-ozone/pull/1149#issuecomment-664815366

Today, when only one serviceId is defined, it is used by s3g. This is th exact behavior which is suggested by #1149 for ozon shell commands.

Why I like the ne aproach to make the default serviceId configurable in case of multiple serviceId present, I would make it optional if one (and only one) serviceId is defined.

I would consider to keep the current behavior in this specific use case. Do you see any risk to keep it?

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneServiceProvider.java
bharatviswa504: When it has more than one service id, it will fail only when internal service id is not configured.

PR #1253, opened 2020-07-24
HDDS-4024. Avoid while loop too soon when exception happen

runzhiwang:

When exception throw in `context.execute(executorService, heartbeatFrequency, TimeUnit.MILLISECONDS)`, thread will not sleep and start another while loop immediately, it will cause cpu very high and huge log size.

image

https://issues.apache.org/jira/browse/HDDS-4024

## How was this patch tested?

No need add test.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeStateMachine.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeStateMachine.java
@@ -229,6 +229,7 @@ public class DatanodeStateMachine implements Closeable {
         // 1. Trigger heartbeat immediately
         // 2. Shutdown has be initiated.
         LOG.warn("Interrupt the execution.", e);
+        Thread.currentThread().interrupt();
       } catch (Exception e) {
         LOG.error("Unable to finish the execution.", e);
       }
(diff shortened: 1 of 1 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeStateMachine.java
```diff
@@ -224,19 +224,25 @@ private void start() throws IOException {
...
-        if (now < nextHB.get()) {
-          if(!Thread.interrupted()) {
-            Thread.sleep(nextHB.get() - now);
-          }
-        }
       } catch (InterruptedException e) {
         // Some one has sent interrupt signal, this could be because
         // 1. Trigger heartbeat immediately
         // 2. Shutdown has be initiated.
+        LOG.warn("Interrupt the execution.", e);
```
adoroszlai, line 231: Now that sleep is moved outside this `try`, I think interrupted state needs to be restored here.  Otherwise if `execute` is interrupted, this catch clears interrupted state and the thread will sleep below.

runzhiwang: @adoroszlai Thanks for review. I have updated the patch.

PR #1257, opened 2020-07-25
HDDS-3970. Enabling TestStorageContainerManager with all failures add…

prashantpogde:

ContainerStateManager: Addressing test failures

https://issues.apache.org/jira/browse/HDDS-3970

## How was this patch tested?
Testing TestStorageContainerManager several times in a row for failures.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
@@ -627,7 +627,6 @@ public class TestStorageContainerManager {
       modifiersField.setAccessible(true);
       modifiersField.setInt(f, f.getModifiers() & ~Modifier.FINAL);
       f.set(replicationManager, publisher);
-      Thread.sleep(20000);
 
       UUID dnUuid = cluster.getHddsDatanodes().iterator().next()
           .getDatanodeDetails().getUuid();
@@ -638,6 +637,13 @@ public class TestStorageContainerManager {
 
       CommandForDatanode commandForDatanode = new CommandForDatanode(
           dnUuid, closeContainerCommand);
+
+      GenericTestUtils.waitFor(() -> {
+        return replicationManager.isRunning();
+      }, 1000, 25000);
+
+      // Give ReplicationManager some time to process the containers.
+      Thread.sleep(5000);
 
       verify(publisher).fireEvent(eq(SCMEvents.DATANODE_COMMAND), argThat(new
           CloseContainerCommandMatcher(dnUuid, commandForDatanode)));
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
@@ -26,7 +26,6 @@ import static org.apache.hadoop.hdds.HddsConfigKeys
 import static org.apache.hadoop.hdds.HddsConfigKeys
     .HDDS_SCM_SAFEMODE_PIPELINE_CREATION;
 import static org.junit.Assert.fail;
-import org.junit.Ignore;
 import static org.mockito.Matchers.argThat;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.mock;
@@ -109,7 +108,6 @@ import com.google.common.collect.Maps;
 /**
  * Test class that exercises the StorageContainerManager.
  */
-@Ignore
 public class TestStorageContainerManager {
   private static XceiverClientManager xceiverClientManager;
   private static final Logger LOG = LoggerFactory.getLogger(
@@ -119,7 +117,7 @@ public class TestStorageContainerManager {
    * Set the timeout for every test.
    */
   @Rule
-  public Timeout testTimeout = new Timeout(300000);
+  public Timeout testTimeout = new Timeout(900000);
 
   @Rule
   public ExpectedException thrown = ExpectedException.none();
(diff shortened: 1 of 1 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
```diff
@@ -525,7 +523,7 @@ public void testScmInfo() throws Exception {
   /**
    * Test datanode heartbeat well processed with a 4-layer network topology.
    */
-  @Test(timeout = 60000)
+  @Test(timeout = 180000)
   public void testScmProcessDatanodeHeartbeat() throws Exception {
```
adoroszlai, line 527: Sleep lengths in `testCloseContainerCommandOnRestart` are being increased.  Why do we need to increase timeout on another test method?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
```diff
@@ -593,7 +591,7 @@ public void testCloseContainerCommandOnRestart() throws Exception {
           new TestStorageContainerManagerHelper(cluster, conf);
 
       helper.createKeys(10, 4096);
-      Thread.sleep(5000);
+      Thread.sleep(10000);
```
adoroszlai, line 594: Repeated check (eg. `GenericTestUtils.waitFor`) should be preferred over fixed length sleep:

1. allow quicker test execution if possible
2. explicit condition documents what the test is waiting for

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
```diff
@@ -604,8 +602,13 @@ public void testCloseContainerCommandOnRestart() throws Exception {
       // Stop processing HB
       scm.getDatanodeProtocolServer().stop();
 
-      scm.getContainerManager().updateContainerState(selectedContainer
-          .containerID(), HddsProtos.LifeCycleEvent.FINALIZE);
+      LoggerFactory.getLogger(TestStorageContainerManager.class).info(
```
adoroszlai, line 605: Why not `LOG.info(`?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
```diff
@@ -604,8 +602,13 @@ public void testCloseContainerCommandOnRestart() throws Exception {
...
+          "Current Container State is" + selectedContainer.getState());
```
adoroszlai, line 606: Please use placeholder in log message instead of `+` to avoid new code warning (sonar).

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
```diff
@@ -604,8 +602,13 @@ public void testCloseContainerCommandOnRestart() throws Exception {
...
+      if (selectedContainer.getState() == HddsProtos.LifeCycleState.OPEN) {
+        scm.getContainerManager().updateContainerState(selectedContainer
+            .containerID(), HddsProtos.LifeCycleEvent.FINALIZE);
```
adoroszlai, line 609: Do you know what other thread changed container state to CLOSING?  Don't we still have a race condition here: in the time window between `getState` and `updateContainerState` that other thread could still change state of the container?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
prashantpogde: I tested it multiple times on my laptop and tests use to timeout occasionally depending on other applications on my laptop. I kept it long enough that there should be no flakiness in the tests.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
prashantpogde: done

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
prashantpogde: no, I couldn't find the other thread closing the container. one theory could be, We stopped processing the heartbeat from the datanode and if it assumes data node is dead, it could close the container. 
Yes there is a race condition still unless it could be done atomically. I changed it to try/catch to avoid any race condition.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
prashantpogde: done

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
prashantpogde: done

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestStorageContainerManager.java
adoroszlai: Thanks.  What about the other sleep?

prashantpogde: I spent some time looking around in the code. Not sure how we can do this cleanly for the other sleep.

prashantpogde: Changed this to ->
 - waiting till the replication manager comes up using GenericTestUtils.waitFor and then
 - wait some more to give it enough time to process containers

prashantpogde:
Unrelated failure with coverage

PR #1258, opened 2020-07-25
HDDS-4029. Recon unable to add a new container which is in CLOSED state.

avijayanhwx:
Whenever there is a state change (OPEN -> CLOSING/CLOSED) while Recon is down, when it starts up again, it is not able to process reports for that container.

https://issues.apache.org/jira/browse/HDDS-4029

## How was this patch tested?
Manually tested.
Added unit tests.


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManager.java
```diff
@@ -325,8 +325,12 @@ public void addContainerInfo(long containerID,
                                Pipeline pipeline) throws IOException {
     Preconditions.checkNotNull(containerInfo);
     containers.addContainer(containerInfo);
-    pipelineManager.addContainerToPipeline(pipeline.getId(),
-        ContainerID.valueof(containerID));
+    if (pipeline != null) {
```
xiaoyuyao, line 328: I'm trying to understand why this is not an issue in SCM? 
Does Recon have similar idea like SCM safemode to ensure Pipeline meet the minimal requirement?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
```diff
@@ -77,22 +82,42 @@ public ReconContainerManager(
...
       DatanodeDetails datanodeDetails)
       throws IOException {
     if (!exists(containerID)) {
       LOG.info("New container {} got from {}.", containerID,
           datanodeDetails.getHostName());
       ContainerWithPipeline containerWithPipeline =
           scmClient.getContainerWithPipeline(containerID.getId());
-      LOG.debug("Verified new container from SCM {} ",
-          containerWithPipeline.getContainerInfo().containerID());
+      LOG.info("Verified new container from SCM {}, part of {} ",
```
xiaoyuyao, line 93: NIT: "part of " => "pipeline"

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
```diff
@@ -77,22 +82,42 @@ public ReconContainerManager(
...
         addNewContainer(containerID.getId(), containerWithPipeline);
       }
+    } else {
+      // Check if container state is not open. In SCM, container state
+      // changes to CLOSING first, and then the close command is pushed down
+      // to Datanodes. Recon 'learns' this from DN, and hence replica state
+      // will move container state to 'CLOSING'.
+      ContainerInfo containerInfo = getContainer(containerID);
+      if (containerInfo.getState().equals(HddsProtos.LifeCycleState.OPEN)
+          && !replicaState.equals(ContainerReplicaProto.State.OPEN)
```
xiaoyuyao, line 106: should we more preciously check for ContainerReplicaProto.State.CLOSE state?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
```diff
@@ -105,18 +130,32 @@ public void addNewContainer(long containerId,
...
+        PipelineID pipelineID = containerWithPipeline.getPipeline().getId();
+        if (getPipelineManager().containsPipeline(pipelineID)) {
+          getContainerStateManager().addContainerInfo(containerId,
+              containerInfo, getPipelineManager(),
+              containerWithPipeline.getPipeline());
+          success = true;
+        } else {
+          // Get open container for a pipeline that Recon does not know
+          // about yet. Cannot update internal state until pipeline is synced.
+          LOG.warn(String.format(
```
xiaoyuyao, line 144: This can be simplified as 
LOG.warn("Pipeline {} not found. Cannot add container {}",
              pipelineID, containerInfo.containerID());

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
avijayanhwx: @xiaoyuyao The container could be in any non-open & healthy state (CLOSING, QUASI_CLOSED, CLOSED). When the container was OPEN Recon may have added it to its metadata, but then it may have gone down. When it comes back up, the container could be in any state. A unit test (testProcessICRStateMismatch ) has been added for all the states.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
avijayanhwx: Sure, will do.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
avijayanhwx: A sample flow of events.

- Recon has containerState = OPEN, Recon gets container report with replicaState = QUASI_CLOSED
- Before processing the replica, it calls on the above logic (ReconContainerManager#checkAndAddNewContainer) which updates containerState to ‘CLOSING’ through the "FINALIZE" event. This is like a pre-processing step for Recon where the container replica state is showing up as unexpected. We have to do this since SCM is where the state transition was initiated, but Recon has to "learn" it later from DNs. 
- Further along in the same process container report flow, there is a process replica step which sees that the containerState = ‘CLOSING’, and replicaState = QUASI_CLOSED, hence moves container state to ‘QUASI_CLOSED’ (AbstractContainerReportHandler#updateContainerState). This is the same code as used by SCM to transition container state by looking at the replica state.

The same sequence of events would be true if starting replicaState = “CLOSING" or "CLOSED” as well

xiaoyuyao: Thanks for the details of state transition handling on Recon side. It makes sense to me.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManager.java
avijayanhwx: @xiaoyuyao Good question. Although we get pipeline report for a new pipeline through DN, it does not contain all the information that was used by SCM to create the pipeline (Factor, other nodes). While implementing this, we considered 2 ways to handle this, either through an elaborate pipeline creation ACK command which Recon can pick up, or through Recon requesting SCM for a confirmation whenever it sees a new pipeline. We went with the latter.  

In general, any state change that "originates" in the SCM needs to be synced with Recon. But anything that originates in the Datanode works automatically.

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/scm/AbstractReconContainerManagerTest.java
```diff
@@ -129,4 +132,46 @@ private StorageContainerServiceProvider getScmServiceProvider()
...
+
+  protected Table<ContainerID, ContainerInfo> getContainerTable()
+      throws IOException {
+    return CONTAINERS.getTable(store);
+  }
+
+  protected ContainerWithPipeline getTestContainer(LifeCycleState state)
+      throws IOException {
+    ContainerID containerID = new ContainerID(100L);
+    Pipeline pipeline = getRandomPipeline();
```
xiaoyuyao, line 144: Can we reuse the getTestContainer below to reduce duplicated code?

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/scm/TestReconContainerManager.java
```diff
@@ -86,12 +104,39 @@ public void testCheckAndAddNewContainer() throws IOException {
...
+        getContainerManager().getContainer(containerID).getState());
+
+    DatanodeDetails datanodeDetails = randomDatanodeDetails();
+
+    // First report with "CLOSED" replica state moves container state to
+    // "CLOSING".
+    getContainerManager().checkAndAddNewContainer(containerID, State.CLOSED,
+        datanodeDetails);
+    assertEquals(CLOSING,
+        getContainerManager().getContainer(containerID).getState());
```
xiaoyuyao, line 140: Can we add an additional case from CLOSING to QUASI_CLOSED upon receiving State.QUASI_CLOSED?

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/scm/AbstractReconContainerManagerTest.java
avijayanhwx: Sure, will do.

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/scm/TestReconContainerManager.java
avijayanhwx: @xiaoyuyao I have added the tests for all the states in TestReconICR. Is that OK?

xiaoyuyao: LGTM.

xiaoyuyao:
+1. Thanks @avijayanhwx  for the patience. The latest LGTM, +1.

avijayanhwx:
Thank you for the review @xiaoyuyao.

PR #1264, opened 2020-07-27
HDDS-4035. Update logs of HadoopDirGenerator.

aryangupta1998:

1. Added logs to HadoopNestedDirGenerator.

2. Edit logs of HadoopDirTreeGenerator. 

3. Use BaseFreonGenerator printReport() for shell and command-line execution of freon commands like ddsg, dtsg, etc.

https://issues.apache.org/jira/browse/HDDS-4035

## How was this patch tested?

Tested Manually.

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
@@ -305,12 +305,10 @@ public class BaseFreonGenerator {
    * Print out reports with the given message.
    */
   public void print(String msg){
-    List<String> messages = new LinkedList<>();
-    messages.add(msg);
     Consumer<String> print = freonCommand.isInteractive()
             ? System.out::println
             : LOG::info;
-    messages.forEach(print);
+    print.accept(msg);
   }
 
   /**
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
@@ -301,6 +301,15 @@ public class BaseFreonGenerator {
     messages.forEach(print);
   }
 
+  public void print(String s){
+    List<String> messages = new LinkedList<>();
+    messages.add(s);
+    Consumer<String> print = freonCommand.isInteractive()
+            ? System.out::println
+            : LOG::info;
+    messages.forEach(print);
+  }
+
   /**
    * Create the OM RPC client to use it for testing.
    */
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
@@ -302,8 +302,7 @@ public class BaseFreonGenerator {
   }
 
   /**
-   * Print out the messages for HadoopNestedDirGenerator
-   * and HadoopDirTreeGenerator tools.
+   * Print out reports with the given message.
    */
   public void print(String msg){
     List<String> messages = new LinkedList<>();
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
@@ -301,9 +301,13 @@ public class BaseFreonGenerator {
     messages.forEach(print);
   }
 
-  public void print(String s){
+  /**
+   * Print out the messages for HadoopNestedDirGenerator
+   * and HadoopDirTreeGenerator tools.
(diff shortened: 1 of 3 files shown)


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/HadoopNestedDirGenerator.java
```diff
@@ -109,5 +117,7 @@ private void createDir(long counter) throws Exception {
       Path dir = new Path(rootPath.concat("/").concat(childDir));
       fileSystem.mkdirs(dir.getParent());
     }
+    System.out.println("\nSuccessfully created directories. Total " +
```
rakeshadr, line 120: Sometimes freon is executed from 'shell' sometimes from 'command line', BaseFreonGenerator.printReport already handle this and we can follow this approach. 

Along with this patch, can you change the "System.out" in https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/HadoopDirTreeGenerator.java#L154 as well. Thanks!

aryangupta1998: Thanks for the review Rakesh. I have updated the PR and also the logs of HadoopTreeDirGenerator as per your suggestions.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
```diff
@@ -301,6 +301,15 @@ public void printReport() {
     messages.forEach(print);
   }
 
+  public void print(String s){
```
rakeshadr, line 304: Latest patch looks good. Adding few minor comments:

Please add javadoc and change the parameter name 's' to 'msg', just to improve more readable.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/HadoopDirTreeGenerator.java
```diff
@@ -97,17 +97,25 @@
...
+    } else {
+      init();
+      OzoneConfiguration configuration = createOzoneConfiguration();
+      fileSystem = FileSystem.get(URI.create(rootPath), configuration);
+
+      contentGenerator = new ContentGenerator(fileSizeInBytes, bufferSize);
+      timer = getMetrics().timer("file-create");
+
+      runTests(this::createDir);
+      return null;
```
rakeshadr, line 116: minor comment: please remove 'return null' inside the else block as below return is sufficient.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
aryangupta1998: Added javadoc, changed 's' to 'msg'.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/HadoopDirTreeGenerator.java
aryangupta1998: Removed 'return null' statement inside else block.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
```diff
@@ -301,6 +301,19 @@ public void printReport() {
     messages.forEach(print);
   }
 
+  /**
+   * Print out the messages for HadoopNestedDirGenerator
```
rakeshadr, line 305: As this is base class, please remove the inherited class details here.

you can write something like,
"Print out reports with the given message."

aryangupta1998: Changed the javadoc.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/BaseFreonGenerator.java
```diff
@@ -301,6 +301,18 @@ public void printReport() {
     messages.forEach(print);
   }
 
+  /**
+   * Print out reports with the given message.
+   */
+  public void print(String msg){
+    List<String> messages = new LinkedList<>();
+    messages.add(msg);
```
xiaoyuyao, line 309: Given we only have one String msg here, we can eliminate thew List allocation here by just simply call
print.accept(msg) in line 313.

aryangupta1998: Thanks @xiaoyuyao. Fixed the print function.

rakeshadr:
@aryangupta1998, Thanks for updating the PR!

The changes look good to me. Pending GREEN CI

rakeshadr:
Merged the changes to master branch. Thanks @aryangupta1998 for the contribution!

PR #1271, opened 2020-07-28
HDDS-4020. ACL commands like getacl and setacl should return a response only when Native Authorizer is enabled.

bharatviswa504:

Currently, the getacl and setacl commands return wrong information when an external authorizer such as Ranger is enabled. There should be a check to verify if Native Authorizer is enabled before returning any response for these two commands.

If an external authorizer is enabled, it should show a nice message about managing acls in external authorizer.  

https://issues.apache.org/jira/browse/HDDS-4020

## How was this patch tested?

Added UT

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/web/utils/OzoneUtils.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/web/utils/OzoneUtils.java
@@ -30,11 +30,15 @@ import java.util.concurrent.TimeUnit;
 
 import org.apache.hadoop.hdds.annotation.InterfaceAudience;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.scm.client.HddsClientUtils;
 import org.apache.hadoop.ozone.OzoneConsts;
 
 import com.google.common.base.Preconditions;
 import org.apache.ratis.util.TimeDuration;
+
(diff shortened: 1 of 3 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/web/utils/OzoneUtils.java
```diff
@@ -166,4 +170,22 @@ public static long getTimeDurationInMS(ConfigurationSource conf, String key,
...
+   * Return true, when Authorizer class is configured with non-default value.
+   * @param configuration
+   * @return boolean
+   */
+  public static boolean checkExternalAuthorizer(
+      OzoneConfiguration configuration) {
+    String authorizerClass = configuration.get(OZONE_ACL_AUTHORIZER_CLASS);
+    if (authorizerClass != null &&
+        !authorizerClass.equals(OZONE_ACL_AUTHORIZER_CLASS_DEFAULT)) {
+      System.out.print(String.format("When External Authorizer %s is " +
```
xiaoyuyao, line 183: Can we move System.out.print to AclHandler#execute based on the checkExternalAuthorizer return?

xiaoyuyao: This configuration seem like a server side configuration and should come from authoritative source such as OM discovery. 

What if the client does know what is configured on OM but OM is actually using native authorizer? If check based on client, we will not be able to do acl operations.

bharatviswa504:
Closing this as HDDS-4089 is trying to fix acl commands when external authorizer is configured.

PR #1272, opened 2020-07-29
HDDS-2660. Create insight point for datanode container protocol

elek:

The goal of this task is to create a new insight point for the datanode container protocol (HddsDispatcher) to be able to debug client<->datanode communication.

https://issues.apache.org/jira/browse/HDDS-2660

## How was this patch tested?

Manually. See the video:

https://www.youtube.com/watch?v=msQgfF95ivc&list=PLCaV-jpCBO8UK5Ged2A_iv3eHuozzMsYv&index=7&t=0s 

;-)


hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/datanode/DatanodeDispatcherInsight.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+    return result;
+  }
+
+  @Override
+  public List<MetricGroupDisplay> getMetrics() {
+    List<MetricGroupDisplay> metrics = new ArrayList<>();
+
+    addProtocolMessageMetrics(metrics, "hdds_dispatcher",
+        Type.SCM, ScmBlockLocationProtocolProtos.Type.values());
```
adoroszlai, line 79: Seems to be leftover from `ScmProtocolBlockLocationInsight`.  I think it should be something like `Type.DATANODE, ContainerProtos.Type.values()` instead.  (`BaseInsightPoint` needs to be tweaked to allow it.)

hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/datanode/DatanodeDispatcherInsight.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Override
+  public String getDescription() {
+    return "Datanode client protocol";
+  }
+
+  @Override
+  public boolean filterLog(Map<String, String> filters, String logLine) {
+    return true;
```
adoroszlai, line 91: Can you please explain why this is needed?  I think this may cause logs from other pipelines to be printed.

For example, I ran `ozone insight logs` on a 3-node pipeline, then:

```
$ ozone sh key put -r ONE /vol1/bucket1/passwd /etc/passwd
```

and the logs from the single datanode to which this block was written (part of the 3-node pipeline, too) appeared in the console.

hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/datanode/DatanodeDispatcherInsight.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.hadoop.ozone.insight.Component.Type;
+import org.apache.hadoop.ozone.insight.InsightPoint;
+import org.apache.hadoop.ozone.insight.LoggerSource;
+import org.apache.hadoop.ozone.insight.MetricGroupDisplay;
+
+import static org.apache.hadoop.ozone.insight.datanode.PipelineComponentUtil.getPipelineIdFromFilters;
+import static org.apache.hadoop.ozone.insight.datanode.PipelineComponentUtil.withDatanodesFromPipeline;
+
+/**
+ * Insight definition for datanode/pipline metrics.
```
adoroszlai, line 40: Seems to be leftover from `RatisInsight`.

hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/datanode/DatanodeDispatcherInsight.java
elek: There are two parts of the filtering:

 1. where to connect
 2. what to display (based on `[...]` text from the log

Unfortunately the pipeline id is not available in the  `HddsDispatcher`, therefore we couldn't filter out the other pipelines (and we need this code segment to show all the lines)

hadoop-ozone/insight/src/main/java/org/apache/hadoop/ozone/insight/datanode/DatanodeDispatcherInsight.java
elek: This is something which couldn't be easily fixed. `Type.DATANODE` is not supported for metrics. I can throw and UnsupportedOperationException instead of showing the metrics.

elek:
Closing this pr temporary. This is on my list but today, it's pending. 

I will improve the insight first to support multiple nodes, or making the datanode insight is to be a one-node insight point...

elek:
Both of the open questions are answered with an improvement: instead of using `pipeline` as a filter I started to use `datanode` (host or ip) as a filter. With this approach we can check the metrics and it's more obvious what is printed out.

elek:
@adoroszlai 

I think the new approach addressed all the earlier problems. Can we merge it now?

adoroszlai:
Thanks @elek for updating the patch.  Interestingly `log -v` stopped working, even if I execute `log` first (as mentioned previously).

elek:
> Thanks @elek for updating the patch. Interestingly log -v stopped working, even if I execute log first (as mentioned previously).

I double-checked and it worked for me. When you use leader, the messages are displayed immediately, when you use follower, the messages will be appeared only after the commit...

adoroszlai:
> > Thanks @elek for updating the patch. Interestingly log -v stopped working, even if I execute log first (as mentioned previously).
> 
> I double-checked and it worked for me. When you use leader, the messages are displayed immediately, when you use follower, the messages will be appeared only after the commit...

It seems to depend on the content.  Plain text files work fine, but it stops working on the first binary, eg. `ozone freon ockg -n1 -t1`.  I guess it's caused by control chars in the random data.

I think we should avoid logging chunk content.  `ContainerCommandRequestMessage` implements related logic to clear data.

elek:
> I think we should avoid logging chunk content. ContainerCommandRequestMessage implements related logic to clear data.

I checked but it's not something which can be added easily. The `OzoneProtocolMessageDispatcher` is very generic. To support this one, it should be modified to accept a very specific function as an argument which transforms the original message just for the tracing. 

And would also introduce significant overhead in case of tracing is turned on.

I am not sure if it's worth it, but I can add it if you think so.

adoroszlai (approved):

adoroszlai:
> > I think we should avoid logging chunk content. ContainerCommandRequestMessage implements related logic to clear data.
> 
> I checked but it's not something which can be added easily.

Thanks for checking.  I'm fine with doing it as a follow-up.

elek:
>> I checked but it's not something which can be added easily.

> Thanks for checking. I'm fine with doing it as a follow-up (created HDDS-4271).

You were right, you did it easily ;-)

PR #1273, opened 2020-07-29
HDDS-4045. Add more ignore rules to the RAT ignore list

elek:

We have separated rat ignore list for Hdds and Ozone projects but some files (which are ignored by .gitignore) not added to there. It's not a problem on github as rat is executed on a clean repo, but locally it can make the execution easier (exclude files which are already ignored).

Rats can read ignore list from VCS ignore file, unfortunately only from directory of the current project not from the root of the git repository.

https://issues.apache.org/jira/browse/HDDS-4045

## How was this patch tested?

```
./hadoop-ozone/dev-support/checks/rat.sh
```

### DIFF
--- a/hadoop-hdds/pom.xml
+++ b/hadoop-hdds/pom.xml
@@ -252,7 +252,15 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
         <configuration>
           <excludes>
             <exclude>**/hs_err*.log</exclude>
+            <exclude>**/.attach_*</exclude>
+            <exclude>**/**.rej</exclude>
+            <exclude>**/.factorypath</exclude>
+            <exclude>public</exclude>
+            <exclude>**/*.iml</exclude>
             <exclude>**/target/**</exclude>
+            <exclude>output.xml</exclude>
+            <exclude>log.html</exclude>
+            <exclude>report.html</exclude>
             <exclude>.gitattributes</exclude>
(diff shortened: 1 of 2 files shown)


hadoop-hdds/pom.xml
```diff
@@ -252,7 +252,15 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
...
             <exclude>**/hs_err*.log</exclude>
+            <exclude>**/.attach_*</exclude>
+            <exclude>**/**.rej</exclude>
+            <exclude>**/.factorypath</exclude>
+            <exclude>public</exclude>
+            <exclude>**/*.iml</exclude>
             <exclude>**/target/**</exclude>
+            <exclude>output.xml</exclude>
+            <exclude>log.html</exclude>
+            <exclude>report.html</exclude>
```
adoroszlai, line 263: Nit: `hadoop-ozone/pom.xml` has slightly different pattern for these 3 files:

```
            <exclude>**/output.xml</exclude>
            <exclude>**/log.html</exclude>
            <exclude>**/report.html</exclude>
```

elek: Yes, that's look better.

elek:
BTW, it would be useful to disable some of the tests at some specific case. For example, I am not sure if I need full retest for this patch/commit.

PR #1274, opened 2020-07-30
HDDS-3810. Add the logic to distribute open containers among the pipelines of a datanode.

bshashikant:

Added logic for open container distribution among multiple pipelines of a dn

https://issues.apache.org/jira/browse/HDDS-3810

## How was this patch tested?
Added unit tests.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
@@ -308,6 +308,10 @@ public final class ScmConfigKeys {
       OZONE_SCM_KEY_VALUE_CONTAINER_DELETION_CHOOSING_POLICY =
       "ozone.scm.keyvalue.container.deletion-choosing.policy";
 
+  public static final String OZONE_SCM_PIPELINE_PER_RAFT_LOG_DISK =
+      "ozone.scm.pipeline.per.raft.log.disk";
+
+  public static final int OZONE_SCM_PIPELINE_PER_RAFT_LOG_DISK_DEFAULT = 1;
   // Max timeout for pipeline to stay at ALLOCATED state before scrubbed.
   public static final String OZONE_SCM_PIPELINE_ALLOCATED_TIMEOUT =
       "ozone.scm.pipeline.allocated.timeout";
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
@@ -308,10 +308,10 @@ public final class ScmConfigKeys {
       OZONE_SCM_KEY_VALUE_CONTAINER_DELETION_CHOOSING_POLICY =
       "ozone.scm.keyvalue.container.deletion-choosing.policy";
 
-  public static final String OZONE_SCM_PIPELINE_PER_METADATA_DISK =
+  public static final String OZONE_SCM_PIPELINE_PER_METADATA_VOLUME =
       "ozone.scm.pipeline.per.metadata.disk";
 
-  public static final int OZONE_SCM_PIPELINE_PER_METADATA_DISK_DEFAULT = 2;
+  public static final int OZONE_SCM_PIPELINE_PER_METADATA_VOLUME_DEFAULT = 2;
   // Max timeout for pipeline to stay at ALLOCATED state before scrubbed.
   public static final String OZONE_SCM_PIPELINE_ALLOCATED_TIMEOUT =
       "ozone.scm.pipeline.allocated.timeout";
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
@@ -308,10 +308,10 @@ public final class ScmConfigKeys {
       OZONE_SCM_KEY_VALUE_CONTAINER_DELETION_CHOOSING_POLICY =
       "ozone.scm.keyvalue.container.deletion-choosing.policy";
 
-  public static final String OZONE_SCM_PIPELINE_PER_RAFT_LOG_DISK =
-      "ozone.scm.pipeline.per.raft.log.disk";
+  public static final String OZONE_SCM_PIPELINE_PER_METADATA_DISK =
+      "ozone.scm.pipeline.per.metadata.disk";
 
-  public static final int OZONE_SCM_PIPELINE_PER_RAFT_LOG_DISK_DEFAULT = 2;
+  public static final int OZONE_SCM_PIPELINE_PER_METADATA_DISK_DEFAULT = 2;
   // Max timeout for pipeline to stay at ALLOCATED state before scrubbed.
   public static final String OZONE_SCM_PIPELINE_ALLOCATED_TIMEOUT =
       "ozone.scm.pipeline.allocated.timeout";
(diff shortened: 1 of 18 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
```diff
@@ -307,6 +307,10 @@
       OZONE_SCM_KEY_VALUE_CONTAINER_DELETION_CHOOSING_POLICY =
       "ozone.scm.keyvalue.container.deletion-choosing.policy";
 
+  public static final String OZONE_SCM_PIPELINE_PER_RAFT_LOG_DISK =
+      "ozone.scm.pipeline.per.raft.log.disk";
```
adoroszlai, line 311: Should be added to `ozone-default.xml`, too (indicated by failure of `TestOzoneConfigurationFields`).

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -821,6 +821,13 @@
     <description>Number of containers per owner per disk in a pipeline.
     </description>
   </property>
+  <property>
+    <name>ozone.scm.pipeline.per.raft.log.disk</name>
+    <value>3</value>
```
adoroszlai, line 826: Shouldn't this be same as `OZONE_SCM_PIPELINE_PER_RAFT_LOG_DISK_DEFAULT`?

elek:
Green build + approval. Will merge this after a new build.

bshashikant:
This needs some more changes and i will be updating it very soon.

bshashikant:
@nandakumar131 @ChenSammi can you plz review?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/DatanodeInfo.java
```diff
@@ -121,6 +143,19 @@ public int getHealthyVolumeCount() {
     }
   }
 
+  /**
+   * Returns count of healthy raft log volumes reported from datanode.
+   * @return count of healthy raft log volumes
+   */
+  public int getRaftLogVolumeCount() {
```
nandakumar131, line 150: Rename suggestion: `getMetadataVolumeCount()`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -101,12 +103,23 @@ public SCMContainerManager(
...
   }
 
+  private int getOpenContainerCountPerPipeline(Pipeline pipeline) {
+    int totalContainerCountPerDn = numContainerPerVolume *
+        pipelineManager.getNumHealthyVolumes(pipeline);
+    int maxPipelineCountPerDn = pipelineManager.maxPipelineLimit(pipeline);
+    return (int) Math.ceil(
+        ((double) totalContainerCountPerDn / maxPipelineCountPerDn));
+  }
+
```
nandakumar131, line 122: Will this work in case of heterogeneous datanodes, where one datanode has 1 Raft log disk with 2 data disk and the other datanode has 5 Raft log disk with 10 data disk?

According to the current logic `getOpenContainerCountPerPipeline` will return 10, if `numContainerPerVolume` and `numPipelinesPerRaftLogDisk` are set to 2.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
```diff
@@ -308,6 +308,10 @@
       OZONE_SCM_KEY_VALUE_CONTAINER_DELETION_CHOOSING_POLICY =
       "ozone.scm.keyvalue.container.deletion-choosing.policy";
 
+  public static final String OZONE_SCM_PIPELINE_PER_RAFT_LOG_DISK =
```
GlenGeng, line 311: How about `OZONE_SCM_PIPELINE_PER_METADATA_DISK` ?  
There are both raft log disk and meta data storage report existing in the code context, they are similar to each other, bring in some redundancy.
BTW, the former one may lead misunderstanding. One might think that each raft log disk contains one raft log, which is straightforwad, nevertheless, the raft log disk and the raft log is a OneToMany relationship.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -530,6 +540,43 @@ public int getNumHealthyVolumes(List<DatanodeDetails> dnList) {
...
+          dn.getUuid());
+    }
+    return 0;
+  }
+
+  /**
+   * Returns the pipeline limit for set of datanodes.
+   */
+  @Override
+  public int maxPipelineLimit(List<DatanodeDetails> dnList) {
```
GlenGeng, line 571: method name is `maxPipelineLimit`, but the logic calculates the min, which is a little bit weird. How about `minPipelineLimit `

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -101,12 +103,23 @@ public SCMContainerManager(
...
+            ScmConfigKeys.OZONE_SCM_PIPELINE_PER_RAFT_LOG_DISK_DEFAULT);
 
     loadExistingContainers();
 
     scmContainerManagerMetrics = SCMContainerManagerMetrics.create();
   }
 
+  private int getOpenContainerCountPerPipeline(Pipeline pipeline) {
+    int totalContainerCountPerDn = numContainerPerVolume *
+        pipelineManager.getNumHealthyVolumes(pipeline);
```
GlenGeng, line 117: `pipelineManager.getMaxHealthyVolumeNum(pipeline) / pipelineManager.getMinPipelineLimit(pipeline)` may be more expressive.

GlenGeng (commented):
LGTM. It give Ozone some possibility to handle heterogenous machines.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
GlenGeng: might need change pipeline placement policy to make sure that only allocating pipelines on homogeneous machiens.

It one pipeline connects a strong DN and weak DN, its open container number will be `volumeNumOfStrongDN * 3 / pipelineNumOfWeakDN.`.

bshashikant: I agree with @GlenGeng here. The pipeline placement should choose homogeneous datanodes. The choice being made here, have many containers open on a minimal set of pipelines that we can have out of the set of datanodes in the pipleines.

bshashikant: Changed the logic to take the min of the data disks count of the set of dns in a pipeline while determining the open container limit.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
```diff
@@ -308,6 +308,10 @@
       OZONE_SCM_KEY_VALUE_CONTAINER_DELETION_CHOOSING_POLICY =
       "ozone.scm.keyvalue.container.deletion-choosing.policy";
 
+  public static final String OZONE_SCM_PIPELINE_PER_METADATA_DISK =
```
GlenGeng, line 311: name suggestion: OZONE_SCM_PIPELINE_PER_METADATA_VOLUME.

We'd better use concept volume instead of concept disk in SCM, the former one is a logical concept, the latter is physical one, and volume is already widely used in context.

GlenGeng (commented):
LGTM. just several inline comments. Thanks for the work!

nandakumar131:
+1, LGTM.

PR #1275, opened 2020-07-30
HDDS-4009. Recon Overview page: The volume, bucket and key counts are not accurate

vivekratnavel:

Instead of relying on Rocks DB's "rocksdb.estimate-num-keys" to get the counts of volume, bucket, and key, a new task is added to persist object counts in Recons SQL db.

https://issues.apache.org/jira/browse/HDDS-4009

## How was this patch tested?

Manually tested in a real cluster and also added unit tests.


avijayanhwx:
/pending Patch is being reworked on after offline conversation with @vivekratnavel.

PR #1276, opened 2020-07-30
HDDS-4046. Extensible subcommands for CLI applications

elek:

HDDS-3814 proposed a new subcommand which deletes the column families from the rocksdb tables.

But during the discussion there was no consensus if it's safe to add or not.

This patch makes the sub-commands easy to extend. Anybody can extend the main ozone shell commands with new sub-commands.

Sub-commands can be added to the classpath and activated by Service Provider Interface (meta-inf/services/...).

This is an optional feature, current approach works forward (annotation based sub-command definition).

And it's not only about HDDS-3814.

It makes it easier to organize the sub-commands. (For example RDBStore related commands can be moved to the rdb classes, they will be picked up).

It makes possible to provide additional, ad-hoc tools during debug or support which can help to solve/debug specific problems.

https://issues.apache.org/jira/browse/HDDS-4046

## How was this patch tested?

With CI. The changed subcommands are covered by acceptance tests. Also tested manually with `ozone sh`.


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/OMAdmin.java
```diff
@@ -50,7 +51,7 @@
     subcommands = {
         GetServiceRolesSubcommand.class
     })
-public class OMAdmin extends GenericCli {
+public class OMAdmin extends GenericCli implements SubcommandWithParent {
```
adoroszlai, line 54: Won't we need

```
@MetaInfServices(SubcommandWithParent.class)
```

too?

Currently `OMAdmin` is added to `OzoneAdmin` from the parent, so it's not a problem.  (#1285 will allow inverting subcommand registration for `OzoneAdmin`, too.)

elek: Yes, it's missing. :eagle: :eye: :trophy: 

Also, I can remove the `OmAdmin` from the annotations of `OzoneAdmin`. (Fortunately annotations and subcommands added programmatically can work together).

adoroszlai:
Thanks @elek for updating the patch.  Tested that `ozone admin` lists `om` as subcommand and `ozone admin om` works.

PR #1280, opened 2020-07-31
HDDS-4054. Remove duplicate judgments in allocateBlockInKey.

captainzmc:

In HDDS-1666 we have [changed the value of the key size](https://github.com/apache/hadoop/pull/943/files#diff-2d061b57a9838854d07da9e0eca64f31R439). So this size must be >0.

final long size = args.getDataSize() > 0 ?
 args.getDataSize() : scmBlockSize;
We [pass this size when we call allocateBlockInKey](https://github.com/apache/hadoop/pull/943/files#diff-2d061b57a9838854d07da9e0eca64f31R480). The key must be greater than 0, so the "if (size > 0){}" in allocateBlockInKey is redundant and should be removed.

https://issues.apache.org/jira/browse/HDDS-4054

## How was this patch tested?

use existing UT

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -468,6 +468,9 @@ public class KeyManagerImpl implements KeyManager {
     openVersion = keyInfo.getLatestVersionLocations().getVersion();
     LOG.debug("Key {} allocated in volume {} bucket {}",
         keyName, volumeName, bucketName);
+
+    // Previously we initialized the size of the size greater than 0. Here we
+    // call allocateBlock.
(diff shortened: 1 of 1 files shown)


captainzmc:
Hi @bharatviswa504  @arp7 Could you help to review this？

bharatviswa504:
@captainzmc 
Old write request code is no more used, there is a cleanup Jira.
https://issues.apache.org/jira/browse/HDDS-3371

captainzmc:
> @captainzmc
> Old write request code is no more used, there is a cleanup Jira.
> https://issues.apache.org/jira/browse/HDDS-3371
> 
> I believe this is causing the issue when reading code, multiple Jira's are recently opened to fix the issue in old code.
> I think we need to prioritize cleanup or mark old methods as deprecated to avoid the confusion.

Thanks @bharatviswa504 . Recently I have noticed that this part of the code is no longer used. 
Close this PR.

PR #1284, opened 2020-08-03
HDDS-4058. Wrong use of AtomicBoolean in HddsDatanodeService

adoroszlai:

* `AtomicBoolean isStopped` should be `final`, not `volatile`, since the reference is not being changed
* `stop()` should use atomic `getAndSet()` instead of `get()` followed by `set()`

https://issues.apache.org/jira/browse/HDDS-4058

## How was this patch tested?

Not really tested, trivial code change.

CI passed:
https://github.com/adoroszlai/hadoop-ozone/runs/939478014


adoroszlai:
Thanks @arp7 for reviewing and committing it.

PR #1285, opened 2020-08-03
HDDS-4056. Convert OzoneAdmin to pluggable model

adoroszlai:

1. `OzoneAdmin` is a collection of loosely coupled admin tools.  This change converts it to the pluggable model introduced in HDDS-4046, where subcommands register themselves declaratively, instead of the parent command listing each subcommand.
2. Move `--scm` parameter from `ozone admin` command to the subcommands that actually need it.  Currently all except `OMAdmin` do, but this is likely to change by adding more `om` and other non-`scm` subcommands in the future.  This is made easier by:
   * extracting `--scm`-related logic to the new `ScmOption` mixin,
   * extracting abstract parent class for `ScmClient`-based subcommands (similar to `Handler` for `ozone shell` subcommands).
3. Upgrade picocli to 4.4.0.  (Needed for `MIXEE`, but useful anyway.)

Note that moving the `--scm` parameter is backwards incompatible.  Previously the following command was valid:

```
ozone admin --scm scm:9860 pipeline list
```

and this one was not:

```
ozone admin pipeline list --scm scm:9860
```

but now it's the other way around.  The same applies to all SCM-related admin commands.  If necessary, we could provide backwards compatibility, but I don't think it's worth the effort and extra code complexity.  By default SCM address is read from config, so command-line override should be rare.

https://issues.apache.org/jira/browse/HDDS-4056

## How was this patch tested?

Existing acceptance tests cover some of the commands.  Also added new test cases for several `ozone admin` subcommands.

CI pending


hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/ScmOption.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_CLIENT_ADDRESS_KEY;
+import static picocli.CommandLine.Spec.Target.MIXEE;
+
+/**
+ * Defines command-line option for SCM address.
+ */
+public class ScmOption {
+
+  @CommandLine.Spec(MIXEE)
+  private CommandLine.Model.CommandSpec spec;
```
elek, line 40: Today I learned it. Nice. Thanks.

elek:
```
ozone admin pipeline list --scm scm:9860
```

Personally I feel it more natural. Moving some parameters before the last subcommand is strange syntax for me. (= I think the new one is easier to use)

elek (approved):
+1 LGTM

Good to have this cleanup (and the tests)

And sorry for the late answer, as I remember I started to check it earlier but somehow forget to finish it.

adoroszlai:

PR #1287, opened 2020-08-04
HDDS-4059: SCMStateMachine::applyTransaction() should not invoke TransactionContext.getClientRequest()

GlenGeng:

`applyTransaction()` should not call `trx.getClientRequest()`. Since client request will not be replicated from leader to follower, follower will not be able to update its state machine. 

We should call `trx.getStateMachineLogEntry()`, since content of client request will be injected to `StateMachineEntryProto` at leader, and extracted from log entry at follower.

https://issues.apache.org/jira/browse/HDDS-4059

## How was this patch tested?

CI


timmylicheng:
+1. Thanks for @GlenGeng contribution. Will merge soon.

PR #1289, opened 2020-08-05
HDDS-4039. Reduce the number of fields in hdds.proto to improve performance

runitao:

Reduce the number of fields in hdds.proto to improve performance. Because these fields are required only for Recon UI and don't have to be part of hdds.proto.

Refer to https://github.com/apache/hadoop-ozone/pull/1226#issuecomment-663416483

HDDS-4039

## How was this patch tested?

Test with docker-compose development enviroment and check the Recon UI works well.
Also test with `npx pnpm run dev` and check the UI is OK.
image

And test by CI
https://github.com/apache/hadoop-ozone/pull/1289/checks

### DIFF
--- a/hadoop-hdds/interface-client/src/main/proto/hdds.proto
+++ b/hadoop-hdds/interface-client/src/main/proto/hdds.proto
@@ -43,11 +43,6 @@ message DatanodeDetailsProto {
     // network name, can be Ip address or host name, depends
     optional string networkName = 6;
     optional string networkLocation = 7; // Network topology location
-    // TODO(runitao): remove it in next release
-    // both version and setupTime are deprecated and moved into ExtendedDatanodeDetailsProto
-    // keep here just for backward compatibility
-    optional string version = 8;
-    optional int64 setupTime = 9;
     // TODO(runzhiwang): when uuid is gone, specify 1 as the index of uuid128 and mark as required
     optional UUID uuid128 = 100; // UUID with 128 bits assigned to the Datanode.
 }
--- a/hadoop-hdds/interface-client/src/main/proto/hdds.proto
+++ b/hadoop-hdds/interface-client/src/main/proto/hdds.proto
@@ -49,14 +49,15 @@ message DatanodeDetailsProto {
 
 /**
  The DatanodeDetailsProto is a basic type that will be shared by many Proto,
- and we just need some more detail information to register with SCM, so I use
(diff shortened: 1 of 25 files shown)


hadoop-hdds/interface-client/src/main/proto/hdds.proto
```diff
@@ -43,14 +43,22 @@ message DatanodeDetailsProto {
     // network name, can be Ip address or host name, depends
     optional string networkName = 6;
     optional string networkLocation = 7; // Network topology location
-    optional string version = 8;         // Datanode version
```
avijayanhwx, line 46: @runitao Can we mark these fields as 'deprecated', and not set them in the code, rather than removing them. This will likely backward compatibility when we have a case to add a new field in the future.

runitao: @avijayanhwx  Thanks for review. No problem. But I need to confirm. 
Should I only keep version and setupTime(these two fields added in 0.6.0) or keep these four fields?

avijayanhwx: @runitao When were these fields added? Were they part of 0.6.0 release?

runitao: I got it, I will keep version and setupTime fields.

runitao: Fix that, PTAL @avijayanhwx, thanks.

avijayanhwx: @runitao HDDS-3798 is marked for 0.7.0, and hence I will revert it in branch **ozone-0.6.0**.

runitao: @avijayanhwx Great. The code will be more clean, I have removed the four fields from DatanodeDetailsProto.

runitao:
@elek Would you help me review this PR if you have time?

avijayanhwx (commented):
Thanks for addressing the review comments @runitao. Apologies for the delay (This PR fell off my radar).
Changes look good to me now.  Can you merge latest master to your patch branch to re-run CI with the latest code?

runitao:
@avijayanhwx Could you PTAL again?

avijayanhwx (approved):
Thanks for working on this patiently @runitao. LGTM +1

PR #1290, opened 2020-08-05
HDDS-4064. Show container verbose info with verbose option

maobaolong:

Show container verbose info with verbose option

HDDS-4064

## How was this patch tested?

```bash
bash-4.2$ ozone admin container info 1 --verbose
Container id: 1
Pipeline Info: Pipeline[ Id: 63fe01d7-45a0-42b9-ad4c-05993a8d7998, Nodes: e8c01f98-4355-43a8-9158-de658e9b62f2{ip: 172.22.0.8, host: ozone_datanode_2.ozone_default, networkLocation: /default-rack, certSerialId: null}c1455dca-5c03-48ac-a1ef-64ceb00b9cb9{ip: 172.22.0.2, host: ozone_datanode_1.ozone_default, networkLocation: /default-rack, certSerialId: null}aaa5575f-37c7-485f-bb5c-cae2b1fa52cd{ip: 172.22.0.7, host: ozone_datanode_3.ozone_default, networkLocation: /default-rack, certSerialId: null}, Type:RATIS, Factor:THREE, State:OPEN, leaderId:e8c01f98-4355-43a8-9158-de658e9b62f2, CreationTimestamp2020-08-05T09:35:25.102Z]
Container State: OPEN
Datanodes: [e8c01f98-4355-43a8-9158-de658e9b62f2/ozone_datanode_2.ozone_default,
c1455dca-5c03-48ac-a1ef-64ceb00b9cb9/ozone_datanode_1.ozone_default,
aaa5575f-37c7-485f-bb5c-cae2b1fa52cd/ozone_datanode_3.ozone_default]

```


hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
```diff
@@ -52,6 +53,10 @@
   @Parameters(description = "Decimal id of the container.")
   private long containerID;
 
+  @CommandLine.Option(names = {"--verbose"},
```
xiaoyuyao, line 56: Can we leverage the existing "--verbose" option defined in parent class GenericCli here?

maobaolong: @xiaoyuyao 
```java
boolean verbose = false;
if (parent.getParent() instanceof GenericCli) {
        if (((GenericCli) parent.getParent()).isVerbose()) {
            verbose = true;
        }
}
```

Do you mean that i should add this code to L66

xiaoyuyao: I agree uses GenericParentCommand seems better. Also, I would suggest we change printError to printStackTrace only if there is no meaningful error message? This way, we will have a clean verbose output as long as the error message is meaningful.

maobaolong: @adoroszlai Thanks for your tips, I commit a new PR, PTAL.

adoroszlai:
Sorry @maobaolong, my recent change for HDDS-4056 caused a conflict.  I went ahead and resolved it for you.  I also added an acceptance test case for the verbose mode.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
adoroszlai: > Also, I would suggest we change printError to printStackTrace only if there is no meaningful error message? This way, we will have a clean verbose output as long as the error message is meaningful.

Good idea @xiaoyuyao.  I think we can split this to a separate improvement issue, so I created HDDS-4160.

maobaolong:
@adoroszlai Thanks for resolve the conflict for me. I check the failed misc acceptance test, which caused by `ozonesecure-mr` cannot exit the safemode, I'll check the reason.

maobaolong:
@adoroszlai Please take another look.

adoroszlai (approved):
Thanks @maobaolong for updating the patch.

@xiaoyuyao can you please take another look?

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
elek: Ping @xiaoyuyao.

Please let me know if you are happy with this approach.

xiaoyuyao:
LGTM, +1. I will merge it shortly.

maobaolong:
@adoroszlai @xiaoyuyao @elek Thanks for your review.

PR #1293, opened 2020-08-05
HDDS-1889 Add support for verifying multiline log entry

llemec:

Added multi line support for Audit messages in TestOzoneAuditLogger

https://issues.apache.org/jira/browse/HDDS-1889

## How was this patch tested?

added messageIncludesMultilineException in TestOzoneAuditLogger

### DIFF
--- a/hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/audit/TestOzoneAuditLogger.java
+++ b/hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/audit/TestOzoneAuditLogger.java
@@ -25,13 +25,20 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import static org.apache.hadoop.ozone.audit.AuditEventStatus.FAILURE;
 import static org.apache.hadoop.ozone.audit.AuditEventStatus.SUCCESS;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.StringContains.containsString;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import org.hamcrest.Matcher;
+import org.hamcrest.collection.IsIterableContainingInOrder;
+
 
 /**
  * Test Ozone Audit Logger.
@@ -149,7 +156,7 @@ public class TestOzoneAuditLogger {
 
   @Test
   public void messageIncludesMultilineException() throws IOException {
-    String exceptionMessage = "Dummuy exception message";
+    String exceptionMessage = "Dummy exception message";
     TestException testException = new TestException(exceptionMessage);
     AuditMessage exceptionAuditMessage =
         new AuditMessage.Builder()
@@ -160,10 +167,11 @@ public class TestOzoneAuditLogger {
             .withResult(FAILURE)
             .withException(testException).build();
     AUDIT.logWriteFailure(exceptionAuditMessage);
-    verifyLog(exceptionMessage,
-          "org.apache.hadoop.ozone.audit."
-              + "TestOzoneAuditLogger."
-              + "messageIncludesMultilineException");
+    verifyLog(
+        "ERROR | OMAudit | user=john | ip=192.168.0.1 | op=CREATE_VOLUME {key1=value1, key2=value2} | ret=FAILURE",
+        "org.apache.hadoop.ozone.audit.TestOzoneAuditLogger$TestException: Dummy exception message",
+        "at org.apache.hadoop.ozone.audit.TestOzoneAuditLogger.messageIncludesMultilineException(TestOzoneAuditLogger.java:160) [test-classes/:?]");
+
 
   }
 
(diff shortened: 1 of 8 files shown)


fapifta:
Hi @llemec, 

first of all thank you for the contribution, acceptance test failures are related to recon, after they are fixed, it would be necessary to have a clean build before a commiter can merge the changeset, please ensure we have one.

I would argue that we should check the order of the messages as well, which would not be that hard to implement neither with your approach nor with the approach I am suggesting inline.

I have some simplification suggestions in the inline comments, please consider them, and if you like the idea, please include them in the PR.

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/audit/TestOzoneAuditLogger.java
```diff
@@ -158,20 +184,34 @@ private void verifyLog(String expected) throws IOException {
...
       i++;
     }
-
-    // When log entry is expected, the log file will contain one line and
-    // that must be equal to the expected string
-    assertTrue(lines.size() != 0);
-    assertTrue(expected.equalsIgnoreCase(lines.get(0)));
+    //check if every expected string can be found in the log entry
+    for(String expected : expectedStrings) {
+      assertTrue(contains(lines, expected));
```
fapifta, line 189: instead of writing a contains method we can use the following combination of assertThat and matchers to gain the same functionality:
```java
assertThat(lines, hasItem(containsString(expected)));
```

If we change to check that lines starts with the lines specified in expectedStrings in order, then an other approach like this solves the problem:
```java
assertThat(
    lines.subList(0,expectedStrings.length),
    containsInOrder(expectedStrings)
);
```

where containsInOrder is defined as:
```java
private Matcher<Iterable<? extends String>> containsInOrder(
    String[] expectedStrings) {
  return IsIterableContainingInOrder.contains(
      Arrays.stream(expectedStrings)
          .map(str -> containsString(str))
          .collect(Collectors.toList())
  );
}
```

In order for this to work it is necessary to add org-harmcrest:harmcrest-all as a test scope dependency to the hdds-common module, and change the messageIncludesMultilineException test, to call verifyLog as:
```java
verifyLog(
    "ERROR | OMAudit | user=john | ip=192.168.0.1 | op=CREATE_VOLUME " +
        "{key1=value1, key2=value2} | ret=FAILURE",,
    "TestException: " + exceptionMsg,
    "org.apache.hadoop.ozone.audit."
        + "TestOzoneAuditLogger."
        + "messageIncludesMultilineException"
);
```
assuming that the message text given to the TestException is extracted to a local variable called exceptionMsg ;)

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/audit/TestOzoneAuditLogger.java
```diff
@@ -143,7 +143,33 @@ public void notLogReadEvents() throws IOException {
...
 
-  private void verifyLog(String expected) throws IOException {
+  /**
+   * Test to verify if multiline entries can be checked.
+   */
+
+  @Test
+  public void messageIncludesMultilineException() throws IOException {
+    try{
+      throw new TestException("Dummy exception");
```
fapifta, line 153: One don't need to throw the exception in order to pass it to the withException method and log it correctly, so the test could be simpler by just specifying a new TestExecption("Dummy") to the withException method on the builder.

llemec: agreed and fixed.

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/audit/TestOzoneAuditLogger.java
```diff
@@ -143,7 +143,31 @@ public void notLogReadEvents() throws IOException {
...
   }
 
-  private void verifyLog(String expected) throws IOException {
+  /**
+   * Test to verify if multiline entries can be checked.
+   */
+
+  @Test
+  public void messageIncludesMultilineException() throws IOException {
+    String exceptionMessage = "Dummuy exception message";
```
fapifta, line 152: there is a typo here in the string Dummuy ;)

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/audit/TestOzoneAuditLogger.java
llemec: Added hamcrest suggestion. I find the legibility of hamcrest messages useful especially for evaluating log entries

hadoop-hdds/common/src/test/java/org/apache/hadoop/ozone/audit/TestOzoneAuditLogger.java
llemec: fixed

fapifta:
Hi @llemec,

thank you for working further on this PR, and addressing my suggestions.
It seems that a few commits somehow mixed into the branch from where the PR is initiated...
Not sure what went wrong, but this one unfortunately needs to be fixed... worst case cancel this PR, and add a new one from a clean branch containing your changes only ;)

@vivekratnavel might be able to help with the coverage problem in the checks, I am not sure but this one seems to be a non-generic issue

Besides that,
- please address the checkstyle issues remained in the code
- the contains method is not needed anymore, can we please remove that from the test class?

llemec:
Hello @fapifta,

Thank you for your review and suggestions. I am proceeding to cancel this PR and re-submit it via a clean branch, together with the checkstyle fixes and the removal of unnecessary methods.

PR #1294, opened 2020-08-05
HDDS-4063. Fix InstallSnapshot in OM HA

hanishakoneru:

OzoneManagerStateMachine#notifyInstallSnapshotFromLeader() checks the incoming roleInfoProto and proceeds with install snapshot request only if the role is Leader. This check is wrong and the roleInfoProto will contain the self node ID and not the leaders.

https://issues.apache.org/jira/browse/HDDS-4063

## How was this patch tested?

Testes manually on a docker cluster.


bharatviswa504:

PR #1295, opened 2020-08-06
HDDS-4037. Incorrect container numberOfKeys and usedBytes in SCM after key deletion

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-4037

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
@@ -110,7 +110,7 @@ public class SCMBlockDeletingService extends BackgroundService {
     for (PendingDeleteStatusList.PendingDeleteStatus deletionStatus :
         deletionStatusList.getPendingDeleteStatuses()) {
       LOG.debug(
-          "Block deletion txnID behinds in datanode {} for containerID {}."
+          "Block deletion txnID lagging in datanode {} for containerID {}."
               + " Datanode delete txnID: {}, SCM txnID: {}",
           dnDetails.getUuid(), deletionStatus.getContainerId(),
           deletionStatus.getDnDeleteTransactionId(),
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
@@ -109,8 +109,8 @@ public class SCMBlockDeletingService extends BackgroundService {
     DatanodeDetails dnDetails = deletionStatusList.getDatanodeDetails();
     for (PendingDeleteStatusList.PendingDeleteStatus deletionStatus :
         deletionStatusList.getPendingDeleteStatuses()) {
-      LOG.info(
-          "Block deletion txnID mismatch in datanode {} for containerID {}."
+      LOG.debug(
+          "Block deletion txnID behinds in datanode {} for containerID {}."
               + " Datanode delete txnID: {}, SCM txnID: {}",
           dnDetails.getUuid(), deletionStatus.getContainerId(),
           deletionStatus.getDnDeleteTransactionId(),
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -317,7 +317,7 @@ public class ReplicationManager
        * and bytes used statistics if needed.
        */
       if (state == LifeCycleState.CLOSED) {
-        checkAndUpdateContainerState(container, replicas);
+        checkAndUpdateContainerInfo(container, replicas);
       }
 
       /*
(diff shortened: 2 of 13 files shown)


avijayanhwx:
@ChenSammi Thanks for fixing this. Can you add a short description on what the root cause is, and how this patch fixes it? That will help the review easier.

ChenSammi:
> 
> 
> @ChenSammi Thanks for fixing this. Can you add a short description on what the root cause is, and how this patch fixes it? That will help make the review easier.
> 

Sure. Currently there are two issues.      1. Container keyCount in datanode side is incorrent.  KeyCount is increased by 1 for every putblock command.  This is true for FilePerChunk container layout while it's incorrect for FilePerBlock layout. With FilePerBlock layout, you will find keyCount value is higher than accutal block number.
    2. Container keyCount and bytesUsed hold in SCM side is based on the container report data from each datanode.   AbstractContainerReportHandler#updateContainerStats,  currently SCM only updates the in-memory Container data if the reported container data is bigger.  This's true for OPEN containers. But doesn't stand for CLOSED containers. 
 In our cluster, there are a lot of containers, all blocks are deleted. But I can still get the keyCount and BytesUsed data through CLI "sh container info" command.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
```diff
@@ -109,8 +109,8 @@ public void handlePendingDeletes(PendingDeleteStatusList deletionStatusList) {
     DatanodeDetails dnDetails = deletionStatusList.getDatanodeDetails();
     for (PendingDeleteStatusList.PendingDeleteStatus deletionStatus :
         deletionStatusList.getPendingDeleteStatuses()) {
-      LOG.info(
-          "Block deletion txnID mismatch in datanode {} for containerID {}."
+      LOG.debug(
+          "Block deletion txnID behinds in datanode {} for containerID {}."
```
adoroszlai, line 113:
```suggestion
          "Block deletion txnID lagging in datanode {} for containerID {}."
```

ChenSammi:
> However, there is another case where similar problem still exists: if key is so large that it spans multiple blocks, `numberOfKeys` will include each block separately.
> 
> Test with patch using default block size of 256MB (after previous 64MB test):
> 
> ```
> $ ozone freon ockg -n1 -t1 -s 536870912 -p 512MB
> ...
> Successful executions: 1
> 
> $ ozone admin container list
> {
>   "state" : "OPEN",
>   "replicationFactor" : "THREE",
>   "replicationType" : "RATIS",
>   "usedBytes" : 600834048,
>   "numberOfKeys" : 3,
>   "lastUsed" : "2020-08-10T08:32:43.112Z",
>   "stateEnterTime" : "2020-08-10T08:22:39.855Z",
>   "owner" : "f430cf45-c4ba-4271-8b3a-22bd314be60d",
>   "containerID" : 1,
>   "deleteTransactionId" : 0,
>   "sequenceId" : 0,
>   "open" : true
> }
> ```
> 
> Here `numberOfKeys` should be 2, but it is 3 = 1 (for 64MB key) + 2 (for 512MB key).

@adoroszlai  thanks for the review and comments.  The keyCount in each container is different from the keyCount from OM point of view.  The keyCount in container is actually the block count(FilePerBlock) or the chunk count(FilePerChunk).  So in the above case, 3 numberOfKeys is an expected value.

ChenSammi:
> I don't think this problem is specific to the FilePerBlock layout. PutBlock may be called for the same block multiple times in both layouts, depending on the buffer flush and block size configuration.
> 
> Config:
> 
> ```
> ozone.client.stream.buffer.size=1MB
> ozone.client.stream.buffer.flush.size=4MB
> ozone.scm.chunk.size=1MB
> ozone.scm.container.size=1GB
> ```
> 
> Test with Ozone 0.5.0, before FilePerBlock was introduced:
> 
> ```
> $ ozone freon ockg -n1 -t1 -s 67108864 -p 64MB
> ...
> Successful executions: 1
> 
> $ ozone scmcli container list
> {
>   "state" : "OPEN",
>   "replicationFactor" : "THREE",
>   "replicationType" : "RATIS",
>   "usedBytes" : 67108864,
>   "numberOfKeys" : 16,
>   "lastUsed" : 156717083,
>   "stateEnterTime" : 156589360,
>   "owner" : "5b4981a0-61f4-42d1-af2d-b990e9a77741",
>   "containerID" : 1,
>   "deleteTransactionId" : 0,
>   "sequenceId" : 0,
>   "open" : true
> }
> ```
> 
> I have verified that this case is fixed by the patch.

@adoroszlai , here I think WriteBlock command will be called multiple times instead of putBlock command, right?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -766,6 +773,32 @@ private void handleUnstableContainer(final ContainerInfo container,
 
   }
 
+  /**
+   * Check and update Container key count and used bytes based on it's replica's
+   * data.
+   */
+  private void checkAndUpdateContainerState(final ContainerInfo container,
```
adoroszlai, line 780: I think something along the lines of `updateContainerStatsFromReplicas` would be more descriptive, especially since container state (closed, etc.) is not being changed.

ChenSammi:
Thanks @adoroszlai for review the code.

PR #1296, opened 2020-08-06
HDDS-4053. Volume space: add quotaUsageInBytes and update it when write and delete key.

captainzmc:

First, we increase quotaUsageInBytes of volume. Later, we will judge whether the volume can be written based on this when we write the key.

https://issues.apache.org/jira/browse/HDDS-4053

## How was this patch tested?

UT is added

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
@@ -269,6 +269,7 @@ public final class OzoneConsts {
   public static final String KEY = "key";
   public static final String SRC_KEY = "srcKey";
   public static final String DST_KEY = "dstKey";
+  public static final String USED_BYTES = "usedBytes";
   public static final String QUOTA_IN_BYTES = "quotaInBytes";
   public static final String QUOTA_IN_COUNTS = "quotaInCounts";
   public static final String OBJECT_ID = "objectID";
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java
@@ -83,6 +83,8 @@ public class OzoneVolume extends WithMetadata {
 
   private int listCacheSize;
 
+  private long usedBytes;
+
   /**
    * Constructs OzoneVolume instance.
    * @param conf Configuration object.
@@ -129,6 +131,17 @@ public class OzoneVolume extends WithMetadata {
     this(conf, proxy, name, admin, owner, quotaInBytes, quotaInCounts,
         creationTime, acls, metadata);
     this.modificationTime = Instant.ofEpochMilli(modificationTime);
+  }
+
+  @SuppressWarnings("parameternumber")
+  public OzoneVolume(ConfigurationSource conf, ClientProtocol proxy,
+      String name, String admin, String owner, long quotaInBytes,
+      long quotaInCounts, long creationTime, long modificationTime,
+      List<OzoneAcl> acls, Map<String, String> metadata,
+      long usedBytes) {
+    this(conf, proxy, name, admin, owner, quotaInBytes, quotaInCounts,
+        creationTime, acls, metadata);
+    this.usedBytes = usedBytes;
   }
 
   @SuppressWarnings("parameternumber")
@@ -252,6 +265,10 @@ public class OzoneVolume extends WithMetadata {
    */
   public List<OzoneAcl> getAcls() {
     return acls;
+  }
+
+  public long getUsedBytes() {
+    return usedBytes;
   }
 
   /**
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmVolumeArgs.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmVolumeArgs.java
@@ -47,7 +47,7 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
   private long quotaInBytes;
   private long quotaInCounts;
   private final OmOzoneAclMap aclMap;
-  private LongAdder usedBytes = new LongAdder();
+  private final LongAdder usedBytes = new LongAdder();
 
   /**
    * Private constructor, constructed via builder.
@@ -181,9 +181,6 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
     return usedBytes;
   }
 
-  public void setUsedBytes(long usedBytes) {
-    this.usedBytes.add(usedBytes);
-  }
   /**
    * Returns new builder class that builds a OmVolumeArgs.
    *
@@ -392,7 +389,8 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
         ", admin='" + adminName + '\'' +
         ", owner='" + ownerName + '\'' +
         ", creationTime='" + creationTime + '\'' +
-        ", quota='" + quotaInBytes + '\'' +
+        ", quotaInBytes='" + quotaInBytes + '\'' +
+        ", usedBytes='" + usedBytes.sum() + '\'' +
         '}';
   }
 
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmVolumeArgs.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmVolumeArgs.java
@@ -23,6 +23,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.atomic.LongAdder;
 
 import org.apache.hadoop.ozone.OzoneAcl;
 import org.apache.hadoop.ozone.OzoneConsts;
@@ -46,6 +47,7 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
   private long quotaInBytes;
   private long quotaInCounts;
   private final OmOzoneAclMap aclMap;
+  private LongAdder usedBytes = new LongAdder();
 
   /**
    * Private constructor, constructed via builder.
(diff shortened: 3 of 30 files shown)


captainzmc:
hi @bharatviswa504 @xiaoyuyao Could you help review the PR about volume quota？

captainzmc:
hi @bharatviswa504 @arp7 Could you help review the PR about volume quota？

captainzmc:
hi @bharatviswa504 @arp7 @xiaoyuyao Could you help review the PR about volume quota？

captainzmc:
Hi @elek @adoroszlai Could you help review the PR about volume quota？

captainzmc:
@ChenSammi CC

cxorm:
Hi @captainzmc , 
Could you be so kind to rebase this PR ?

I would take a look on this.

captainzmc:
> Hi @captainzmc ,
> Could you be so kind to rebase this PR ?
> 
> I would take a look on this.

Thanks for yisheng's attention, PR  has been updated. @cxorm Now this PR can be reviewed.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java
```diff
@@ -131,6 +133,18 @@ public OzoneVolume(ConfigurationSource conf, ClientProtocol proxy,
     this.modificationTime = Instant.ofEpochMilli(modificationTime);
   }
 
+  @SuppressWarnings("parameternumber")
```
xiaoyuyao, line 136: NIT: Can we add a builder to avoid managing constructors with different parameters?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -182,8 +179,16 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           new CacheKey<>(dbOzoneKey),
           new CacheValue<>(Optional.of(omKeyInfo), trxnLogIndex));
 
+      long scmBlockSize = ozoneManager.getScmBlockSize();
+      int factor = omKeyInfo.getFactor().getNumber();
+      omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
+      long updateNum = omKeyInfo.getDataSize() * factor -
+          locationInfoList.size() * scmBlockSize * factor;
```
cxorm, line 186: I'm confused about the meaning of the variable. 
(Please let me know the meaning of the variable if I miss something.)

Could you please add some comments about it ?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -292,14 +289,21 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           bucketName, Optional.absent(), Optional.of(missingParentInfos),
           trxnLogIndex);
 
+      long scmBlockSize = ozoneManager.getScmBlockSize();
+      omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
+      long updateNum = newLocationList.size() * scmBlockSize
```
cxorm, line 294:
```suggestion
      long preAllocatedSpace = newLocationList.size() * scmBlockSize
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -210,10 +206,17 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           new CacheKey<>(openKeyName),
           new CacheValue<>(Optional.of(openKeyInfo), trxnLogIndex));
 
+      long scmBlockSize = ozoneManager.getScmBlockSize();
+      omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
+      long updateNum = newLocationList.size() * scmBlockSize
```
cxorm, line 211:
```suggestion
      long preAllocatedSpace = newLocationList.size() * scmBlockSize
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCommitPartRequest.java
```diff
@@ -207,13 +209,21 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           new CacheKey<>(openKey),
           new CacheValue<>(Optional.absent(), trxnLogIndex));
 
+      long scmBlockSize = ozoneManager.getScmBlockSize();
+      int factor = omKeyInfo.getFactor().getNumber();
+      omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
+      long updateNum = omKeyInfo.getDataSize() * factor -
+          keyArgs.getKeyLocationsList().size() * scmBlockSize * factor;
```
cxorm, line 216: The same as `OMKeyCommitRequest`

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -210,10 +206,17 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+          * openKeyInfo.getFactor().getNumber();
+      // atom update quotaUsage.
```
cxorm, line 213:
```suggestion
      // update quotaUsage atomically.
```

cxorm (changes requested):
Thank you @captainzmc for the work.

Some comments are added inline.
1. All variable-naming comments could be discussed if you have an idea.
2. A little confuse on `OMKeyCommitRequest.java` should be addressed. 

The test-part are not reviewed, if someone would take a look on this PR, you could start from here.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java
captainzmc: I think that's good advice, and that's a problem that also exists in OzoneBucket. We can optimize these problems separately, I opened a JIRA HDDS-4223 to follow and optimize this part.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
```diff
@@ -269,6 +269,7 @@ private OzoneConsts() {
   public static final String KEY = "key";
   public static final String SRC_KEY = "srcKey";
   public static final String DST_KEY = "dstKey";
+  public static final String QUOTA_USAGE_IN_BYTES = "quotaUsageInBytes";
```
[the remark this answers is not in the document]
captainzmc: Using quotaUsage here may not be appropriate because it makes it impossible to distinguish between QuotaUsageInBytes and later QuotaUsageInCount. 
Here we can use the usage in ContainerInfo, [use usedBytes](https://github.com/apache/hadoop-ozone/blob/971a36eea16b38257892244d6de862b4f9461138/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerInfo.java#L55).

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmVolumeArgs.java
```diff
@@ -46,6 +47,7 @@
   private long quotaInBytes;
   private long quotaInCounts;
   private final OmOzoneAclMap aclMap;
+  private LongAdder quotaUsageInBytes = new LongAdder();
```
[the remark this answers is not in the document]
captainzmc: QuotaUsageInBytes is a property of Volume that needs to be updated each time when CreateKey, AllocateBlock, CommitKey, DeleteKey, and the lock of Volume is usually used. 
Previously, only Bucket locks were used for these operation, and Volume locks can greatly affect the concurrency performance of different buckets. 
So to avoid Volume locking for performance, LongAdder is used here to complete the atomic update of quotaUsageInBytes.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
captainzmc: I will add comment to describe the calculation here.

captainzmc:
Thanks for @cxorm @xiaoyuyao to the review.  The new commit has been updated. Could you help take another look?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java
xiaoyuyao: sounds good to me.

xiaoyuyao (approved):
Thanks @captainzmc  for working on this. The patch LGTM, +1. Only concern I have is the cost of volume usage update per key write/delete. If you can post the perf diff with/wo update using freon, that will be great.

captainzmc:
> Thanks @captainzmc for working on this. The patch LGTM, +1. Only concern I have is the cost of volume usage update per key write/delete. If you can post the perf diff with/wo update using freon, that will be great.

Thanks for @xiaoyuyao 's +1. Here are the freon test results:
Using 10-100 threads to write data to different buckets under the same volume, adding quota Usage has little impact on performance. (LongAdder disperses the concurrency of a single cell to each cell, improving concurrency efficiency compared to atomicLong).  In my test, I used three virtual machines, each with a key size of 10K.
image

captainzmc:
Had updated PR to fix review issues and resolve conflict. @ChenSammi Could you help take another look?

ChenSammi:
Thanks @captainzmc  for the contribution and @cxorm @xiaoyuyao for review the patch.

PR #1303, opened 2020-08-07
HDDS-4075. Retry request on different OM on AccessControlException

hanishakoneru:

If a client attempts a request on an OM which has not caught up with leader OM and hence does have the delegation token, the request could fail with AccessControlException without trying it on other OMs.
On AccessControlException, all OMs must be tried once before the request is failed.

https://issues.apache.org/jira/browse/HDDS-4075

## How was this patch tested?

No new unit tests

### DIFF
--- a/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/failover/TestOMFailovers.java
+++ b/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/failover/TestOMFailovers.java
@@ -1,3 +1,21 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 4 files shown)


hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/failover/TestOMFailovers.java
```diff
@@ -0,0 +1,148 @@
...
+    GenericTestUtils.setLogLevel(OMFailoverProxyProvider.LOG, Level.DEBUG);
+    GenericTestUtils.LogCapturer logCapturer = GenericTestUtils.LogCapturer
+        .captureLogs(OMFailoverProxyProvider.LOG);
+
+    MockFailoverProxyProvider failoverProxyProvider =
+        new MockFailoverProxyProvider(conf);
+
+    OzoneManagerProtocolPB proxy = (OzoneManagerProtocolPB) RetryProxy
+        .create(OzoneManagerProtocolPB.class, failoverProxyProvider,
+            failoverProxyProvider.getRetryPolicy(9));
```
bharatviswa504, line 65: Minor: Can we use the default, if there is no special reason to use value 9

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/failover/TestOMFailovers.java
```diff
@@ -0,0 +1,148 @@
...
+ * Tests OM failover protocols using a Mock Failover provider and a Mock OM
+ * Protocol.
+ */
+public class TestOMFailovers {
+
+  private ConfigurationSource conf = new OzoneConfiguration();
+  private Exception testException;
+
+  @Test
+  public void test1() throws Exception {
```
bharatviswa504, line 52: Minor: Can we name test with the test it is trying to perform like testAccessControlException

bharatviswa504 (commented):
Overall LGTM.

bharatviswa504:
Thank You @hanishakoneru for the update and refactor code, and making it easy for Unit test FailOverProxy Provider.

bharatviswa504 (approved):
+1 LGTM

bharatviswa504:

PR #1304, opened 2020-08-08
HDDS-1745. Add integration test for createDirectory for OM HA

amaliujia:

Add an integration test for createDirectory which is implemented as part of HDDS-1730 for OM HA.

https://issues.apache.org/jira/browse/HDDS-1745

## How was this patch tested?

Integration Test.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAMetadataOnly.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAMetadataOnly.java
@@ -61,7 +61,6 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.Iterator;
 
-import static org.apache.hadoop.hdds.HddsConfigKeys.OZONE_METADATA_DIRS;
 import static org.apache.hadoop.ozone.MiniOzoneHAClusterImpl.NODE_FAILURE_TIMEOUT;
 import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_CLIENT_WAIT_BETWEEN_RETRIES_MILLIS_DEFAULT;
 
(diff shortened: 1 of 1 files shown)


amaliujia:
@bharatviswa504 

Can I ask you some suggestions on this PR:
1. Am I writing the integration test in the expected way?
2. Any suggestion that what kind of cases in your mind that this PR should test?
        - Right now I am just testing a normal successful CreateDirectory call.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAMetadataOnly.java
```diff
@@ -348,6 +362,57 @@ public void testJMXMetrics() throws Exception {
...
+  public void testOMCreateDirectory() throws Exception {
+    ObjectStore objectStore = getCluster().getRpcClient().getObjectStore();
+    String volumeName = "vol";
+    String bucketName = "buk";
+    String keyName = "test_dir";
+
+    objectStore.createVolume(volumeName);
+    objectStore.getVolume(volumeName).createBucket(bucketName);
+
+    OMRequest request = OMRequest.newBuilder().setCreateDirectoryRequest(
```
bharatviswa504, line 375: We do not need this kind of tests, they are covered in UT.

We can use `createDirectory `API and test create Directory functionality.

Test cases can be: 
1. single level path ('/dir)
2. Multi-level ('/dir1/dir2')
3. Few parents exist in multi-level path
4. Already a directory exists with the same name

any other cases you want to add :)

amaliujia: @bharatviswa504 thank you! 

I was confused on which `createDirectory` API is. Can you paste me a link here?

bharatviswa504: https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java#L588

amaliujia: Ah interesting.. 

If I run 
```
    bucket.createDirectory("/dir1");
    bucket.createDirectory("/dir1");
```
The test will still pass without a complain. I expect an exception should be thrown. 

I will the internal of OM to see how does CreateDirectory API work, and see if there is anything can be improved.

amaliujia: @bharatviswa504  

I am currently have a problem to verify that a directory is created successfully. 

What is the API to retrieve the information of directory? Seems that it is not considered as a key or a file so either `getKey` and `getFileStatus` does not work.

Basically I can do 

`bucket.createDirectory("/dir1");` then I need to do a `bucket.getDirectoryStatus`.

I also tried another direction: after a `createDirectory` call, try to put a file to that directory and then read that file to justify that `createDirectory` succeed. Later I realized this also does not work because without ``createDirectory`, the create file call will always work even for multiple level directory (e.g. `/dir1/dir2/dir3/file`).

PR #1305, opened 2020-08-08
HDDS-4009. Recon Overview page: The volume, bucket and key counts are not accurate

vivekratnavel:

 - Add a new task in Recon to keep track of the count of all the OM DB tables
 - Add / update unit tests
 - Add a new method "getSkipCache" in Table interface to get value from the table by skipping the cache. This is required for Recon, since recon never adds entries to cache of any table.
 - Update ClusterStateEndpoint to get the counts of volume, bucket, and key from global stats table and not rely on "rocksdb.estimate-num-keys"

https://issues.apache.org/jira/browse/HDDS-4009

## How was this patch tested?

Manually tested in a real cluster and also added unit tests.

### DIFF
--- a/hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/StatsSchemaDefinition.java
+++ b/hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/StatsSchemaDefinition.java
@@ -66,13 +66,4 @@ public class StatsSchemaDefinition implements ReconSchemaDefinition {
             .primaryKey("key"))
         .execute();
   }
-
-  /**
-   * Returns the DSL context.
-   *
-   * @return dslContext
-   */
-  public DSLContext getDSLContext() {
-    return dslContext;
-  }
 }
--- a/hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/StatsSchemaDefinition.java
+++ b/hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/StatsSchemaDefinition.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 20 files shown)


hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/schema/StatsSchemaDefinition.java
```diff
@@ -46,22 +48,31 @@
...
             .primaryKey("key"))
         .execute();
   }
+
+  /**
+   * Returns the DSL context.
+   *
+   * @return dslContext
+   */
+  public DSLContext getDSLContext() {
```
avijayanhwx, line 75: The 'dslContext' is common across tables. Do we need to add specific getters inside the individual table definitions? Can we just get the 'datasource' instance from the test injector and get hold of the dslcontext?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/OMDBUpdatesHandler.java
```diff
@@ -262,12 +270,13 @@ public void markCommit(byte[] bytes) throws RocksDBException {
...
-   * @param name table name.
-   * @return String.class by default.
+   * Return Key type class for the given table.
+   *
+   * @return keyType class.
    */
-  private Class<String> getKeyType() {
-    return String.class;
+  @VisibleForTesting
+  Optional<Class> getKeyType(String name) {
```
avijayanhwx, line 278: Nit. We can get rid of these single line methods.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/TableCountTask.java
```diff
@@ -0,0 +1,185 @@
...
+ * Class to iterate over the OM DB and store the total counts of volumes,
+ * buckets, keys, open keys, deleted keys, etc.
+ */
+public class TableCountTask implements ReconOmTask {
+  private static final Logger LOG =
+      LoggerFactory.getLogger(TableCountTask.class);
+
+  private GlobalStatsDao globalStatsDao;
+  private Configuration sqlConfiguration;
+  private HashMap<String, Long> objectCountMap;
```
avijayanhwx, line 52: This can be a method local variable in process().

vivekratnavel:
@avijayanhwx Thanks for the review! I have updated the patch addressing all your comments. Please take another look. Thanks!

PR #1310, opened 2020-08-10
HDDS-4094. Support byte-level write in Freon HadoopFsGenerator

elek:

Teragen seems to use the byte level write method of `FSDataOutputStream` (`write(byte)` is used instead of `write(byte[], int, int)`).

It seems to be a good idea to extend existing `ContentGenerator` of ozone to support the write in smaller chunks to make it easier to reproduce performance problems.

Note: statistics from `FileSystem` instance:

```
Closing file system instance: 1257412274
   write.call: 100001066
   write.allTime: 215951
   hsync.call: 1
   hsync.allTime: 3
   hflush.call: 0
   hflush.allTime: 0
   close.call: 4
   close.allTime: 62
```

This was a teragen test with 1GB data (and statistics from one container). write method seems to be called multiple times which means smaller write buffer.

https://issues.apache.org/jira/browse/HDDS-4094

## How was this patch tested?

Executed Freon test with or without turning on the new flag: 

https://elek.github.io/ozone-notes/performance/19_hcfs/

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ContentGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ContentGenerator.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang3.RandomStringUtils;
 
 /**
@@ -77,7 +78,8 @@ public class ContentGenerator {
     }
   }
 
-  public byte[] getBuffer() {
+  @VisibleForTesting
+  byte[] getBuffer() {
     return buffer;
   }
 }
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ContentGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ContentGenerator.java
@@ -40,7 +40,7 @@ public class ContentGenerator {
   private int bufferSize;
 
   /**
-   * Number of bytes to write in one call. Should be less than the bufferSize.
+   * Number of bytes to write in one call. Should be no larger than the bufferSize.
    */
   private final int copyBufferSize;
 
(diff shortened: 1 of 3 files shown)


hadoop-ozone/tools/src/test/java/org/apache/hadoop/ozone/freon/TestContentGenerator.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Test
+  public void writeWithSmallBuffer() throws IOException {
+    ContentGenerator generator = new ContentGenerator(1024, 1024, 10);
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+
+    generator.write(output);
+    Assert.assertArrayEquals(generator.getBuffer(), output.toByteArray());
+  }
+}
```
adoroszlai, line 56:
```suggestion

  @Test
  public void writeWithDistinctSizes() throws IOException {
    ContentGenerator generator = new ContentGenerator(20, 8, 3);
    ByteArrayOutputStream output = new ByteArrayOutputStream();

    generator.write(output);

    byte[] expected = new byte[20];
    byte[] buffer = generator.getBuffer();
    System.arraycopy(buffer, 0, expected, 0, buffer.length);
    System.arraycopy(buffer, 0, expected, 8, buffer.length);
    System.arraycopy(buffer, 0, expected, 16, 4);
    Assert.assertArrayEquals(expected, output.toByteArray());
  }
}
```

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ContentGenerator.java
```diff
@@ -38,15 +39,23 @@
    */
   private int bufferSize;
 
+  /**
+   * Number of bytes to write in one call. Should be less than the bufferSize.
```
adoroszlai, line 43:
```suggestion
   * Number of bytes to write in one call. Should be no larger than the bufferSize.
```

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ContentGenerator.java
```diff
@@ -56,7 +64,20 @@ public void write(OutputStream outputStream) throws IOException {
...
       int curSize = (int) Math.min(bufferSize, nrRemaining);
-      outputStream.write(buffer, 0, curSize);
+      if (copyBufferSize == 1) {
+        for (int i = 0; i < curSize; i++) {
+          outputStream.write(buffer[i]);
+        }
+      } else {
+        for (int i = 0; i < nrRemaining; i += copyBufferSize) {
+          outputStream.write(buffer, i,
+              Math.min(copyBufferSize, (int) (nrRemaining - i)));
```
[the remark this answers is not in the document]
elek: Thanks the suggestions. I also created a unit test when I saw that you already suggested one, slightly different so I kept both of them ;-)

adoroszlai:

PR #1311, opened 2020-08-10
HDDS-4095. Byteman script to debug HCFS performance

elek:
## What changes were proposed in this pull request

Teragen with Ozone is reported to be slower with Ozone than Hdfs with some specific configuration.

To make it easier to compare different file system, we need an easy measurement on the same point (`FSDataOutputStream`.write) to compare the time spent in that method with Hdfs/Ozone/other file systems.

It can be done with a byteman instrumentation rule

https://issues.apache.org/jira/browse/HDDS-4095

## How was this patch tested?

https://elek.github.io/ozone-notes/performance/20_teragen/

https://elek.github.io/ozone-notes/performance/19_hcfs/

Both tests used this btm file. FS usage was reported on the stdout:

For example: 

```
Closing file system instance: 1398604678
   write.call: 30000
   write.allTime: 692
   hsync.call: 0
   hsync.allTime: 0
   hflush.call: 0
   hflush.allTime: 0
   close.call: 10000
   close.allTime: 411171
```

### DIFF
--- a/dev-support/byteman/hcfs-write.btm
+++ b/dev-support/byteman/hcfs-write.btm
@@ -1,0 +1,111 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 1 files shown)


adoroszlai:
Thanks @elek for the contribution.

PR #1312, opened 2020-08-10
HDDS-2411. add a datanode chunk validator fo datanode chunk generator

frischHWC:

Solve HDDS-2411 by adding a Datanode chunk Validator for Ozone Freon test.

Note that a new checksum is computed for each chunk.

https://issues.apache.org/jira/browse/HDDS-2411

## How was this patch tested?

Manual tests were made by setting up a docker-compose cluster of 3 nodes (they are required as generator check for a pipeline with a factor THREE). 
A login to a node with following commands:
ozone freon rk --replicationType RATIS --factor THREE
ozone freon dcg -p tester
ozone freon dcgv -p tester

To show that test was well working, a change in chunk size for validator involves (as expected) a failure:
ozone freon dcg -p test -s 1000


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/DatanodeChunkValidator.java
```diff
@@ -0,0 +1,243 @@
...
+import picocli.CommandLine.Option;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.Callable;
+
+/**
+ * Data validator of chunks to use pure datanode XCeiver interface.
+ */
+@Command(name = "dcgv",
```
elek: NIT: g usually used as an acronym of "validator", `dcv` might be a better abbrev.

elek (commented):
Thanks the patch @frischHWC (and sorry for the late review).

Based on the code it looks good to me. One minor comment: I prefer to remove the `g` from the `name` of the command (the current convention is using `g` for generate and `v` for validate). To make the merge faster I pushed this change (hope, you don't mind).

I am trying to test but I have some problem with my local env. Will merge it soon when I am able to test it...

elek (approved):
+ 1

Finally, it worked well. I had some wrong ratis version locally. 

Thanks the contribution, will merge it after a green build...

frischHWC:
One check unsuccessful as Ozone Freon testing on Minikube did not passed. 
Error is due to SCM not exiting safe mode when setting up the k8s cluster for minikube, however other tests are working when setting up k8s cluster. 
Don't know if it's related to code added (at first sight, it should not).

adoroszlai:
> One check unsuccessful as Ozone Freon testing on Minikube did not passed.
> Error is due to SCM not exiting safe mode when setting up the k8s cluster for minikube, however other tests are working when setting up k8s cluster.
> Don't know if it's related to code added (at first sight, it should not).

No, it's not related, it's happening elsewhere (other PRs and commits), too.

bshashikant (approved):

elek:
> No, it's not related, it's happening elsewhere (other PRs and commits), too.

Yes, but it doesn't seem to be a false positive. I checked and somehow the SCM couldn't come out from the safe mode under 90 sec. Which is strange...

(I mean unrelated, but not a test problem)

elek:
Kubernetes problem is already tracked (HDDS-4152/HDDS-4153). Usually I hate to merge PR, even with unrelated failures, but this is so isolated, I wouldn't like to burn more CPU...

elek:
@frischHWC Would you be so kind to add ANY comment to the issue?

(I need your Jira username to add you to the project to make it possible to be assigned.)

frischHWC:
@elek : my jira username: frisch. 
No problem to comment the issue, assigned it to me and changes its status.

PR #1313, opened 2020-08-10
Revert HDDS-3989 & HDDS-3960 from 0.6.0.

avijayanhwx:
Following up on my comment in https://github.com/apache/hadoop-ozone/pull/1289/files#r468065882. Since the patches added a proto change which we want to be handled differently, I want to revert the changes from GA. That way, we dont need to live with those extra fields for ever in DatanodeDetails.proto.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@
 .classpath
 .project
 .settings
-*.factorypath
 target
 build
 dependency-reduced-pom.xml
@@ -62,6 +61,5 @@ output.xml
 report.html
 
 hadoop-hdds/docs/public
-hadoop-ozone/recon/node_modules
 
 .mvn
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
@@ -49,8 +49,6 @@ public class DatanodeDetails extends NodeImpl implements
   private String hostName;
   private List<Port> ports;
   private String certSerialId;
-  private String version;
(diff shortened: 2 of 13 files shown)


swagle:
So, these changes need to be brought in with a corresponding MLV (Metadata Layout Version) change?

avijayanhwx:
/retest

> So, these changes need to be brought in with a corresponding MLV (Metadata Layout Version) change?

In HDDS-3978, to show more information on Recon, a few “static” fields like version and setup time were added to the core DatanodeDetails proto that is used for every heartbeat. In a subsequent JIRA (HDDS-4039), these fields are being moved to DN Register proto instead of heartbeat proto. If the changes in HDDS-3978 are shipped with 0.6.0 (GA), we could never remove them due to backward compatibility. If we revert HDDS-3978 instead, then the subsequent JIRA does not need to worry about backward compatibility. cc @runitao .

runitao:
+1 LGTM

ChenSammi:
Hi @avijayanhwx , I'm OK with the revert.  Just need to confirm with you that,  the reverted commits are HDDS-3989 and HDDS-3960, not the HDDS-3798, or the HDDS-3978, right?

avijayanhwx:
> Hi @avijayanhwx , I'm OK with the revert. Just need to confirm with you that, the reverted commits are [HDDS-3989](https://issues.apache.org/jira/browse/HDDS-3989) and [HDDS-3960](https://issues.apache.org/jira/browse/HDDS-3960), not the [HDDS-3798](https://issues.apache.org/jira/browse/HDDS-3798), or the [HDDS-3978](https://issues.apache.org/jira/browse/HDDS-3978), right?

Yes, that is correct. Thanks!

PR #1314, opened 2020-08-11
HDDS-3988: DN can distinguish SCMCommand from stale leader SCM

GlenGeng:

As part of SCMCommand SCM will also send its current term, which will be used in Datanode to identify if the command was sent by the latest leader SCM.
 
Datanode will maintain the highest term that it has seen and compare it with the term that is received as part of SCMCommand.
If the term in the Datanode and SCMCommand are same, the command is added to the command queue for processing.
If the term in the Datanode is less than the term received in SCMCommand, Datanode will update its term and add the command to the command queue for processing.
If the term in the Datanode is greater than the term received in SCMCommand, Datanode will ignore the command.

https://issues.apache.org/jira/browse/HDDS-3988

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -31,6 +32,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -79,6 +81,18 @@ public class StateContext {
   private boolean shutdownOnError = false;
   private boolean shutdownGracefully = false;
   private final AtomicLong threadPoolNotAvailableCount;
+
+  /**
+   * term of latest leader SCM, extract from SCMCommand.
+   *
+   * Only leader SCM (both latest and stale) can send out SCMCommand,
+   * which will save its term in SCMCommand. Since latest leader SCM
+   * always has the highest term, term can be used to detect SCMCommand
+   * from stale leader SCM.
+   *
+   * For non-HA mode, term of SCMCommand will be 0.
+   */
+  private Optional<Long> termOfLeaderSCM = Optional.empty();
 
   /**
    * Starting with a 2 sec heartbeat frequency which will be updated to the
@@ -471,6 +485,65 @@ public class StateContext {
   }
 
   /**
+   * After startup, datanode needs detect latest leader SCM before handling
+   * any SCMCommand, so that it won't be disturbed by stale leader SCM.
+   *
+   * The rule is: after majority SCMs are in HEARTBEAT state and has
+   * heard from leader SCMs (commandQueue is not empty), datanode will init
+   * termOfLeaderSCM with the max term found in commandQueue.
+   *
+   * The init process also works for non-HA mode. In that case, term of all
+   * SCMCommands will be 0.
+   */
+  private void initTermOfLeaderSCM() {
+    // only init once
+    if (termOfLeaderSCM.isPresent()) {
+      return;
+    }
+
+    AtomicInteger scmNum = new AtomicInteger(0);
+    AtomicInteger activeScmNum = new AtomicInteger(0);
+
+    getParent().getConnectionManager().getValues()
+        .forEach(endpoint -> {
+          if (endpoint.isPassive()) {
+            return;
+          }
+          scmNum.incrementAndGet();
+          if (endpoint.getState()
+              == EndpointStateMachine.EndPointStates.HEARTBEAT) {
+            activeScmNum.incrementAndGet();
+          }
+        });
+
+    // majority SCMs should be in HEARTBEAT state.
+    if (activeScmNum.get() < scmNum.get() / 2 + 1) {
+      return;
+    }
+
+    // if commandQueue is not empty, init termOfLeaderSCM
+    // with the largest term found in commandQueue
+    commandQueue.stream()
+        .mapToLong(SCMCommand::getTerm)
+        .max()
+        .ifPresent(term -> termOfLeaderSCM = Optional.of(term));
+  }
+
+  /**
+   * monotonically increase termOfLeaderSCM.
+   * Always record the latest term that has seen.
+   */
+  private void updateTermOfLeaderSCM(SCMCommand<?> command) {
+    if (!termOfLeaderSCM.isPresent()) {
+      LOG.error("should init termOfLeaderSCM before update it.");
+      return;
+    }
+
+    termOfLeaderSCM = Optional.of(
+        Long.max(termOfLeaderSCM.get(), command.getTerm()));
+  }
+
+  /**
    * Returns the next command or null if it is empty.
    *
    * @return SCMCommand or Null.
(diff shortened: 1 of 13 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
```diff
@@ -470,6 +484,65 @@ public void execute(ExecutorService service, long time, TimeUnit unit)
...
+        .mapToLong(SCMCommand::getTerm)
+        .max()
+        .ifPresent(term -> termOfLeaderSCM = Optional.of(term));
+  }
+
+  /**
+   * monotonically increase termOfLeaderSCM.
+   * Always record the latest term that has seen.
+   */
+  void updateTermOfLeaderSCM(SCMCommand<?> command) {
```
timmylicheng, line 536: Can be a private method

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
```diff
@@ -470,6 +484,65 @@ public void execute(ExecutorService service, long time, TimeUnit unit)
...
+   * any SCMCommand, so that it won't be disturbed by stale leader SCM.
+   *
+   * The rule is: after majority SCMs are in HEARTBEAT state and has
+   * heard from leader SCMs (commandQueue is not empty), datanode will init
+   * termOfLeaderSCM with the max term found in commandQueue.
+   *
+   * The init process also works for non-HA mode. In that case, term of all
+   * SCMCommands will be 0.
+   */
+  void initTermOfLeaderSCM() {
```
timmylicheng, line 498: Can be a private method

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
```diff
@@ -470,6 +484,65 @@ public void execute(ExecutorService service, long time, TimeUnit unit)
...
+   * monotonically increase termOfLeaderSCM.
+   * Always record the latest term that has seen.
+   */
+  void updateTermOfLeaderSCM(SCMCommand<?> command) {
+    if (!termOfLeaderSCM.isPresent()) {
+      LOG.error("should init termOfLeaderSCM before update it.");
+      return;
+    }
+
+    termOfLeaderSCM = Optional.of(
```
timmylicheng, line 542: Shall we use AtomicLong for termOfLeaderSCM?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -130,6 +135,14 @@ public SCMNodeManager(OzoneConfiguration conf,
     this.useHostname = conf.getBoolean(
         DFSConfigKeysLegacy.DFS_DATANODE_USE_DN_HOSTNAME,
         DFSConfigKeysLegacy.DFS_DATANODE_USE_DN_HOSTNAME_DEFAULT);
+    this.scmhaManager = scmhaManager;
+  }
+
+  public SCMNodeManager(OzoneConfiguration conf,
```
timmylicheng, line 141: Any case SCMNodeManager would not require SCMHAManager?
Note that PipelineManager will always rely on SCMHAManager

amaliujia:
cc @amaliujia

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
```diff
@@ -478,7 +551,26 @@ public void execute(ExecutorService service, long time, TimeUnit unit)
...
+      }
+
+      while (true) {
+        SCMCommand<?> command = commandQueue.poll();
+        if (command == null) {
+          return null;
+        }
+
+        updateTermOfLeaderSCM(command);
+        if (command.getTerm() == termOfLeaderSCM.get()) {
```
amaliujia, line 566: I am confused on when `termOfLeaderSCM` is updated to the newest leader term. Is `termOfLeaderSCM` updated in during leader selection?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/MockSCMHAManager.java
```diff
@@ -78,8 +79,8 @@ public void start() throws IOException {
    * {@inheritDoc}
    */
   @Override
-  public boolean isLeader() {
-    return isLeader;
+  public Optional<Long> isLeader() {
```
amaliujia, line 82: another challenge is how to test such change. ideally we can have a minicluster setup with configurable nodes so we can simulate the scenario of split brian.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
GlenGeng: Do you mean whether `termOfLeaderSCM` is updated in leader election of SCM ? No, it won't. Datanode detects the latest SCM term by heartbeat with SCMs, whose interval is larger than 30s.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/MockSCMHAManager.java
GlenGeng: Good point. We need add a lot of test cases, including UT and acceptance test, before merging 2823 back to master.

amaliujia: Agreed. I have been thinking about it and we might be some testing infrastructure to simulate these complicated consensus cases. E.g. split-brian.    

OM HA might have done something similar.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
GlenGeng: Sure

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
GlenGeng: Sure

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
GlenGeng: Good Point. 
Since this is executed in a single thread, let's do this first. We surely need to consider the thread safety in future change.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
GlenGeng: Quite a lot of unit tests rely on this old Ctor.

GlenGeng:
@ChenSammi @timmylicheng @nandakumar131  CI has been passed upon the latest HDDS-2823. Please revise this PR, Thanks!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -656,8 +669,57 @@ public void setContainers(DatanodeDetails datanodeDetails,
   // Since datanode commands are added through event queue, onMessage method
   // should take care of adding commands to command queue.
   // Refactor and remove all the usage of this method and delete this method.
+  /**
+   * Only leader SCM can send SCMCommand to datanode, and needs record its
+   * term in the command so that datanode can distinguish commands from stale
+   * leader SCM by comparing term.
+   *
+   * There are 7 SCMCommands:
```
timmylicheng, line 677: Nit: Try not to list all the commands by name. It would be hard to maintain the correctness of the comments.

timmylicheng:
LGTM overall. Left comments inline on a minor issue.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
GlenGeng: Good point. I've removed the comments for good.

As discussed in design of SCMContext, we will query and inject the term before firing event of DATANODE_COMMAND explicitly for all related places. I will revisit this code in dev of phase 2.0.

GlenGeng:
@timmylicheng  Thanks for looking at this patch! 

We surely need unit tests for this feature. I've file a new for this task https://issues.apache.org/jira/browse/HDDS-4544

timmylicheng:
+1. Merging

PR #1316, opened 2020-08-12
HDDS-4109. Tests in TestOzoneFileSystem should use the existing MiniOzoneCluster

smengcl:

4 new tests have been added since HDDS-2833 and are not sharing that MiniOzoneCluster.

I am able to cut down the run time of TestOzoneFileSystem from 3m18s to 1m2s on my Mac. It would only save more run time on GitHub Workflow.

https://issues.apache.org/jira/browse/HDDS-4109

## How was this patch tested?

No new tests needed.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -107,7 +107,6 @@ public class TestOzoneFileSystem {
   private int rootItemCount;
   private Trash trash;
 
-  @Test(timeout = 300_000)
   public void testCreateFileShouldCheckExistenceOfDirWithSameName()
       throws Exception {
     /*
@@ -120,7 +119,6 @@ public class TestOzoneFileSystem {
      *
      * Op 3. create file -> /d1/d2/d3 (d3 as a file inside /d1/d2)
      */
-    setupOzoneFileSystem();
 
     Path parent = new Path("/d1/d2/d3/d4/");
     Path file1 = new Path(parent, "key1");
@@ -154,6 +152,9 @@ public class TestOzoneFileSystem {
     } catch (FileAlreadyExistsException fae) {
       // ignore as its expected
     }
+
+    // Cleanup
+    fs.delete(new Path("/d1/"), true);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


adoroszlai:
> 4 new tests have been added since [HDDS-2883](https://issues.apache.org/jira/browse/HDDS-2883) and are not sharing that MiniOzoneCluster.

I think it's HDDS-2833.

smengcl:
> > 4 new tests have been added since [HDDS-2883](https://issues.apache.org/jira/browse/HDDS-2883) and are not sharing that MiniOzoneCluster.
> 
> I think it's [HDDS-2833](https://issues.apache.org/jira/browse/HDDS-2833).

Fixed typo. Thanks!

smengcl:
> Thanks @smengcl for this speedup. Can you please add a comment that new tests should be added as private methods invoked from `testFileSystem()` to help avoid regression next time?

Sure thing. I've added a note in the javadoc of the test class.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -154,21 +155,21 @@ public void testCreateFileShouldCheckExistenceOfDirWithSameName()
     } catch (FileAlreadyExistsException fae) {
       // ignore as its expected
     }
+
+    // Cleanup
+    fs.delete(new Path("/d1/"), true);
```
ayushtkn, line 160: Shouldn't the cleanup be in a `finally` block? post `mkdir` if the test fails the directory won't get cleaned up. Wouldn't that affect other tests?

smengcl: Thanks for taking a look at this @ayushtkn .

In this case it doesn't matter IMO. Cause if this test (`testCreateFileShouldCheckExistenceOfDirWithSameName`) fails and throws, `testFileSystem` as a whole will fail at this point thus the consecutive tests won't be executed.

ayushtkn: hmm, correct the following tests will fail. Do you think, thats a good behaviour, Ideally only the affected test should fail. If there is a common cluster, that could have been initialised in a `BeforeClass` and all other tests could have ran independently.
Anyway not related to what you are doing here. Thanx for confirming.

ayushtkn (approved):
LGTM.
Initialising cluster can be though done in `BeforeClass` and the `tests` can be made independent, preventing failure of one affecting subsequent tests. May be sometime in future? if there were so specific reasons for doing it this way.
 
Anyway that isn't something being targeted here. just wanted to share my thoughts on this. :-)

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
smengcl: Yes, I also think `@BeforeClass` is better. Maybe another jira to refactor the test class this way.

smengcl: Filed https://issues.apache.org/jira/browse/HDDS-4162

smengcl:
Thanks @ayushtkn  for reviewing this. I have filed https://issues.apache.org/jira/browse/HDDS-4162 for the follow-up refactoring (use `BeforeClass`).

Will merge shortly.

PR #1319, opened 2020-08-12
HDDS-4107: replace scmID with clusterID for container and volume at Datanode side

GlenGeng:

The disk layout per volume is as follows:
```
../hdds/VERSION
../hdds/<<scmUuid>>/current/<<containerDir>>/<<containerID>>/metadata
../hdds/<<scmUuid>>/current/<<containerDir>>/<<containerID>>/<<dataDir>>
```

However, after SCM-HA is enabled, a typical SCM group will consists of 3 SCMs, each of the SCMs has its own scmUuid, meanwhile share the same clusterID.
Since federation is not supported yet, only one cluster is supported now, this Jira will change scmID to clusterID for container and volume at Datanode side.

The disk layout after the change will be as follows:
```
../hdds/VERSION
../hdds/<<clusterID>>/current/<<containerDir>>/<<containerID>>/metadata
../hdds/<<clusterID>>/current/<<containerDir>>/<<containerID>>/<<dataDir>>
```

https://issues.apache.org/jira/browse/HDDS-4107

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/HddsDatanodeService.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/HddsDatanodeService.java
@@ -282,9 +282,8 @@ public class HddsDatanodeService extends GenericCli implements ServicePlugin {
    * datanode.
    */
   private void startRatisForTest() throws IOException {
-    String scmId = "scm-01";
     String clusterId = "clusterId";
-    datanodeStateMachine.getContainer().start(scmId);
+    datanodeStateMachine.getContainer().start(clusterId);
     MutableVolumeSet volumeSet =
         getDatanodeStateMachine().getContainer().getVolumeSet();
 
@@ -292,7 +291,7 @@ public class HddsDatanodeService extends GenericCli implements ServicePlugin {
 
     for (Map.Entry<String, HddsVolume> entry : volumeMap.entrySet()) {
       HddsVolume hddsVolume = entry.getValue();
-      boolean result = HddsVolumeUtil.checkVolume(hddsVolume, scmId,
+      boolean result = HddsVolumeUtil.checkVolume(hddsVolume,
           clusterId, LOG);
       if (!result) {
         volumeSet.failVolume(hddsVolume.getHddsRootDir().getPath());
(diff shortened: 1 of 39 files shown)


elek:
Thanks the patch @GlenGeng 

I don't follow SCM-HA development very closely, but is it a backward-compatible change?

GlenGeng:
Thanks for looking at this PR @elek 

No, this is not a backward-compatible change. We need solve these compatibllity issues before merging 2823 back to master. @nandakumar131 @timmylicheng

elek:
> No, this is not a backward-compatible change. We need solve these compatibllity issues before merging 2823 back to master. 

Thanks to make it clear:

 1. Do we have any Jira issue to track this work?
 2. Do we have any high-level idea how the compatibility issue can be solved?

GlenGeng:
Above all, if we accept that all SCMs in a raft group share the same scmId and clusterId, the storage layout of data node would not be affected, and we can abandon this PR. However if they share different scmId but the same clusterId, I guess we have to address this backward incompatible issue.

During last sync up of SCM-HA, @nandakumar131 said he would like to talk with you and other guys for this compatibility problem. I wonder if any agreement has been achieved?

elek:
> During last sync up of SCM-HA, @nandakumar131 said he would like to talk with you and other guys for this compatibility problem. I wonder if any agreement has been achieved?

We didn't talk (yet) but I am fine with any option. I was just interested if we have any high level plan.

But now I understood the problem (thanks to your clean problem description). 

What about this: 

When datanode registers to an SCM HA server it can get both cluster and scm id. If there is an existing directory under scm id but not under the cluster id, datanode can rename the dir.

I am not familiar with the latest SCM HA design, but if we believe it's something which is possible, just create a Jira and merge this as is...

GlenGeng:
Hey @elek , I found that scmId not only exists in layout of volume dir, but also in container meta data.

I consider It will make the upgrade infeasible for a huge cluster consisting of millions of containers.

elek:
(Sorry, late answer)

Why do we need to updated container files? I see no problems to keep (the valid) clusterId in the container metadata files. 

Or do you need to add scm id in addition to the cluster id?

elek:
/pending @GlenGeng What is the plan with this issue? Do you need help? How can I help to move it forward?

GlenGeng:
> /pending @GlenGeng What is the plan with this issue? Do you need help? How can I help to move it forward?

Hey, @elek, sorry for the late reply!

I've discussed with @nandakumar131, several weeks ago, we decided to abandon this solution, and will figure out a new one.
Since currently this will not block our SCM HA test, we give it a low priority.

I will close this PR.

PR #1322, opened 2020-08-12
HDDS-3829. Introduce Layout Feature interface in Ozone.

avijayanhwx:
- Implement the concept of a 'Layout Feature' in Ozone (with sample usage in Ozone Manager), which defines a specific change in on-disk layout in Ozone.
- Every feature is associated with a layout version, and an API corresponding to the feature cannot be invoked (throws NOT_SUPPORTED_OPERATION) before finalization.
- Created an annotation based 'aspect' for "guarding" new APIs that are introduced by Layout Features. Check out TestOMLayoutFeatureAspect#testCheckLayoutFeature. 
- Added sample features and tests for ease of review (To be removed before commit).
- Created an abstract VersionManager and an inherited OM Version manager to initialize features, check if feature is allowed, check need to finalize, do finalization.

https://issues.apache.org/jira/browse/HDDS-3829

## How was this patch tested?
Unit tested.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/s3/security/package-info.java
```diff
@@ -19,4 +19,4 @@
 /**
  * Package contains classes related to S3 security responses.
  */
-package org.apache.hadoop.ozone.om.request.s3.security;
+package org.apache.hadoop.ozone.om.response.s3.security;
```
fapifta, line 22: This might go into a separate pull request and we may commit it way before the branch is commited.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/upgrade/TestOMVersionManager.java
```diff
@@ -0,0 +1,63 @@
...
+  @Test
+  public void testOMLayoutVersionManager() throws IOException {
+    OMStorage omStorage = mock(OMStorage.class);
+    when(omStorage.getLayoutVersion()).thenReturn(0);
+    OMVersionManager.init(omStorage);
+    assertTrue(OMVersionManager.isAllowed(INITIAL_VERSION));
+    assertFalse(OMVersionManager.isAllowed(CREATE_EC));
+    assertEquals(0, OMVersionManager.getMetadataLayoutVersion());
+    assertTrue(OMVersionManager.needsFinalization());
+    OMVersionManager.doFinalize(mock(OzoneManager.class));
```
fapifta, line 51: Can you please elaborate what will this line do in the test, and why it has to be here? It is unclear whether it tests anything, or this is just a leftover from a thought.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMLayoutFeatureCatalog.java
```diff
@@ -0,0 +1,71 @@
...
+package org.apache.hadoop.ozone.om.upgrade;
+
+import java.util.Optional;
+
+import org.apache.hadoop.ozone.upgrade.LayoutFeature;
+
+/**
+ * Catalog of Ozone Manager features.
+ */
+public class OMLayoutFeatureCatalog {
```
fapifta, line 28: If we use an enum to store the list of features in a component, why do we need the enclosing class here? Is there a future intent hidden here, or it is just a leftover from an earlier implementation?

fapifta:
Hi @avijayanhwx,

thank you for starting this work, and posting an inital version of the core code for the upgrade support.
I have a few general questions and concerns, also added a few comments after a quick review.

In HDFS the layoutversion is a monotonically decreasing number, we chose to use monotonically increasing version numbers, I am unsure why HDFS chose the negative numbers, can there be some hidden considerations, we might go through, before committing to the positive layoutversions and monotonic increase? I haven't found one.

Also, in HDFS one layoutversion covers one feature, would it be better to do not let two feetures associated with one layoutversion number? What is the benefit of having two features mapped to the same layout version? I don't feel good about it, though I don't have a specific example yet where it can cause trouble.

LayoutVersionManager is implemented in a way that it is fully static. I am unsure if this is a good design, I understand the intent that there has to be only one instance of this per component, and seeing it this way it is a reasonable choice to use static a implementation, but it can fire back later when we want to implement tests that involves changing something in this logic, and we can not freely and easily change the behaviour in tests as I fear, also it can introduce invisible interdependencies later that might be hard to detect/factor out. Implementing it in a non-static way does not seem to cause any drawback, even we can be fine with multiple instances for the same component, as the used values will be anyway hardcoded in the real system. What do you think, I would consider making it non-static, as I think it has more possibilities and less limitations later.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMLayoutFeatureCatalog.java
avijayanhwx: Good point. I wanted to have an enclosing class for "ease" of understanding. We have a 'catalog' of features in OM. In the future, we can expect some helper methods to be added at the catalog level which may not be needed at the Feature level.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/s3/security/package-info.java
avijayanhwx: Agreed. I had to include this since the plugin that was added for scanning annotations fails the build due to this discrepancy. I will create a separate JIRA to fix this on master, but since it is a tiny change, I would like to leave it here until then to keep the build green.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/upgrade/TestOMVersionManager.java
avijayanhwx: I am sure it is a leftover line ;) I will remove it.

avijayanhwx:
> Hi @avijayanhwx,
> 
> thank you for starting this work, and posting an inital version of the core code for the upgrade support.
> I have a few general questions and concerns, also added a few comments after a quick review.
> 
> In HDFS the layoutversion is a monotonically decreasing number, we chose to use monotonically increasing version numbers, I am unsure why HDFS chose the negative numbers, can there be some hidden considerations, we might go through, before committing to the positive layoutversions and monotonic increase? I haven't found one.
> 
> Also, in HDFS one layoutversion covers one feature, would it be better to do not let two feetures associated with one layoutversion number? What is the benefit of having two features mapped to the same layout version? I don't feel good about it, though I don't have a specific example yet where it can cause trouble.
> 
> LayoutVersionManager is implemented in a way that it is fully static. I am unsure if this is a good design, I understand the intent that there has to be only one instance of this per component, and seeing it this way it is a reasonable choice to use static a implementation, but it can fire back later when we want to implement tests that involves changing something in this logic, and we can not freely and easily change the behaviour in tests as I fear, also it can introduce invisible interdependencies later that might be hard to detect/factor out. Implementing it in a non-static way does not seem to cause any drawback, even we can be fine with multiple instances for the same component, as the used values will be anyway hardcoded in the real system. What do you think, I would consider making it non-static, as I think it has more possibilities and less limitations later.



> Hi @avijayanhwx,
> 
> thank you for starting this work, and posting an inital version of the core code for the upgrade support.
> I have a few general questions and concerns, also added a few comments after a quick review.
> 
> In HDFS the layoutversion is a monotonically decreasing number, we chose to use monotonically increasing version numbers, I am unsure why HDFS chose the negative numbers, can there be some hidden considerations, we might go through, before committing to the positive layoutversions and monotonic increase? I haven't found one.
> 
> Also, in HDFS one layoutversion covers one feature, would it be better to do not let two feetures associated with one layoutversion number? What is the benefit of having two features mapped to the same layout version? I don't feel good about it, though I don't have a specific example yet where it can cause trouble.
> 
> LayoutVersionManager is implemented in a way that it is fully static. I am unsure if this is a good design, I understand the intent that there has to be only one instance of this per component, and seeing it this way it is a reasonable choice to use static a implementation, but it can fire back later when we want to implement tests that involves changing something in this logic, and we can not freely and easily change the behaviour in tests as I fear, also it can introduce invisible interdependencies later that might be hard to detect/factor out. Implementing it in a non-static way does not seem to cause any drawback, even we can be fine with multiple instances for the same component, as the used values will be anyway hardcoded in the real system. What do you think, I would consider making it non-static, as I think it has more possibilities and less limitations later.

Regarding the HDFS layout feature versions, I could not find any document on why that was the implementation choice. So, I went with what I felt was more intuitive.

I can change the implementation to have only 1 Layout Feature per version.

I am looking into making the Version Manager class non-static. The only issue I see is that it may be used in many different code areas within a component, and hence may need to be passed around (if it cannot be instantiated without external help).

fapifta:
Hi @avijayanhwx,

the current test failures seems to be unrelated, as this will go into the feature branch and we still will work on a lot of stuff and form this further, I am +1 committing the changes to the branch as they are, and so we can start work on the initial idea further.
Thank you for the explanation, and addressing my concerns partly here, and partly in other forms.

As we discussed, the current aspect implementation requires the version manager to be static in order to be able to check versions without major changes in the code, we might address it later, other things seems to be reasonable.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
```diff
@@ -0,0 +1,110 @@
...
+  public int getMetadataLayoutVersion() {
+    return metadataLayoutVersion;
+  }
+
+  public int getSoftwareLayoutVersion() {
+    return softwareLayoutVersion;
+  }
+
+  public boolean needsFinalization() {
+    return metadataLayoutVersion < softwareLayoutVersion;
```
linyiqun, line 70: Sometimes metadataLayoutVersion is not always updated during rolling, we should have one variable called lastSoftwareLayoutVersion to compare with softwareLayoutVersion. And then to decide if we need to finalized.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
```diff
@@ -0,0 +1,110 @@
...
+  protected int softwareLayoutVersion; // SLV.
+  protected TreeMap<Integer, LayoutFeature> features = new TreeMap<>();
+  protected Map<String, LayoutFeature> featureMap = new HashMap<>();
+  protected volatile boolean isInitialized = false;
+
+  protected void init(int version, LayoutFeature[] lfs) {
+    if (!isInitialized) {
+      metadataLayoutVersion = version;
+      initializeFeatures(lfs);
+      softwareLayoutVersion = features.lastKey();
```
linyiqun, line 47: Here softwareLayoutVersion is extracted from LayoutFeature, this seems also like a metadataLayoutVersion. is this true? It looks a little confused here.

linyiqun (commented):
@avijayanhwx , sorry for the delayed review. 
As I see this PR was merged,  let's follow up this in follow-up JIRA.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
avijayanhwx: @linyiqun The lastSoftwareLayoutVersion is what we write into the OM Version file on finalize. This is what we read into the MLV during init. Hence, MLV is same as 'lastSoftwareLayoutVersion'. 

Let's say you finalized when you have features 1,2,3. So you write the value "3" into the Version file on finalize. The next software has layout features 1,2,3,4,5. When you "upgrade" the software bits and start OM, then it will see MLV = 3 and SLV = max(version(feature)). Since 3 < 5, it needs to be finalized again.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
avijayanhwx: Same as I have explained in the last comment.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
avijayanhwx: After finalize, the MLV becomes same as SLV (after writing to Version file), and then all feature APIs > 3 and <= 5 are now allowed.

avijayanhwx: In the v2 Upgrade doc, we were talking about LFV (Last Finalized Version) in terms of handling the Ratis "Apply Transaction" changes. For the first version, since we are adding 'Apply Transaction' changes also as Layout Features, we do not need that. 

Based on how often apply transaction changes are needed, in the next version of implementation, we can maintain "software version" separately from metadata layout version. The only extra added benefit for that is that we can support downgrades with just apply transaction changes, and no layout version changes.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
```diff
@@ -0,0 +1,110 @@
...
+  public int getSoftwareLayoutVersion() {
+    return softwareLayoutVersion;
+  }
+
+  public boolean needsFinalization() {
+    return metadataLayoutVersion < softwareLayoutVersion;
+  }
+
+  public boolean isAllowed(LayoutFeature layoutFeature) {
+    return layoutFeature.layoutVersion() <= metadataLayoutVersion;
```
linyiqun, line 74: I am looking into for this logic again. If new version is not finalized, metadataLayoutVersion will still older than new SLV = max(version(feature)), that means this new introduced features in new Ozone version cannot work until we do the finalize. But actually sometimes, some features can have a minimal compatible version, it can allow to run >=  minimal compatible MLV.

For the first step implementation, we could make each minimal compatible MLV as their current layout version, like


  public enum OMLayoutFeature implements LayoutFeature {
    INITIAL_VERSION(0, 0, "Initial Layout Version"),
    CREATE_EC(1, 1, ""),
    NEW_FEATURE(2, 2, "new feature", new NewOmFeatureUpgradeAction());


    private int layoutVersion;
    private int minCompatLayoutVersion

And then use minCompatLayoutVersion to do compatibility check.

linyiqun (commented):
Thanks @avijayanhwx  for the comments. One more comment from me below.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
avijayanhwx: @linyiqun It is by design that newer "layout" features cannot work in Ozone until finalize. We will have a concept of Minimum Compatible layout version in the future, which is used to denote the minimum Layout version from which you can upgrade to the current layout version.

PR #1327, opened 2020-08-13
HDDS-4098. Improve om admin getserviceroles error message

dshavkanihort:

Improve the Error Messages in Display.

https://issues.apache.org/jira/browse/HDDS-4098?filter=-1

## How was this patch tested?

$ cd hadoop-ozone/dist/target/ozone-*/compose/ozone
$ docker-compose up -d
$ docker-compose exec scm bash
bash-4.2$ ozone admin om getserviceroles --service-id=om
Error: This command works only on OzoneManager HA cluster. Service ID specified does not match with ozone.om.service.ids defined in the configuration. Configured ozone.om.service.ids are 
[]
bash-4.2$

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/OMAdmin.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/OMAdmin.java
@@ -80,8 +80,8 @@ public class OMAdmin extends GenericCli implements SubcommandWithParent {
       throw new OzoneClientException("This command works only on OzoneManager" +
           " HA cluster. Service ID specified does not match" +
           " with " + OZONE_OM_SERVICE_IDS_KEY + " defined in the " +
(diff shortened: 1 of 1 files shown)


dineshchitlangia:
@dshavkanihort Thank you for working on this.
Overall the changes look good to me, I see a checkstyle issue has been reported that needs to be fixed.

@adoroszlai do you have any suggestion for improving the wording of the message?

adoroszlai:
Thanks @dshavkanihort for working on this.

The checkstyle issue @dineshchitlangia mentioned is that line 83 exceeds maximum allowed length of 80 characters:

https://github.com/apache/hadoop-ozone/blob/773af08368d339a75703caef42f74d9ab8cf6bb2/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/OMAdmin.java#L83

You can find the problem in the _Summary of failures_ section after opening _Details_ of the failed _checkstyle_ check.  Here is the direct link: https://github.com/apache/hadoop-ozone/pull/1327/checks?check_run_id=981251536#step:4:5

PR #1334, opened 2020-08-17
HDDS-4125: Pipeline is not removed when a datanode goes stale

GlenGeng:

Let scrubPipeline() handles the destroy of all kinds of pipeline. 

https://issues.apache.org/jira/browse/HDDS-4125

## How was this patch tested?

CI


timmylicheng:
+1. Thanks @GlenGeng for the fix. Merging

PR #1336, opened 2020-08-17
HDDS-4119. Improve performance of the BufferPool management of Ozone client

elek:

Teragen reported to be slow with low number of mappers compared to HDFS.

In my test (one pipeline, 3 yarn nodes) 10 g teragen with HDFS was ~3 mins but with Ozone it was 6 mins. It could be fixed with using more mappers, but when I investigated the execution I found a few problems reagrding to the BufferPool management.

 1. IncrementalChunkBuffer is slow and it might not be required as BufferPool itself is incremental
 2. For each write operation the bufferPool.allocateBufferIfNeeded is called which can be a slow operation (positions should be  calculated).
 3. There is no explicit support for write(byte) operations

In the [flamegraphs](https://github.com/elek/ozone-notes/tree/master/profiles) it's clearly visible that with low number of mappers the client is busy with buffer operations. After the patch the Rpc call and the checksum calculation give the majority of the time.

Overall write performance is improved with at least 30% when minimal number of threads/mappers are used. 

## Thanks

Special thanks to @lokeshj1703, who helped me find the small mistakes in the original verison of the patch.

https://issues.apache.org/jira/browse/HDDS-4119

## How was this patch tested?

Teragen 10/100g with 2/30 mappers.

(https://github.com/elek/ozone-perf-env/tree/master/teragen-hdfs)


hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBuffer.java
```diff
@@ -44,9 +45,6 @@ static ChunkBuffer allocate(int capacity) {
    *   When increment <= 0, entire buffer is allocated in the beginning.
    */
   static ChunkBuffer allocate(int capacity, int increment) {
-    if (increment > 0 && increment < capacity) {
-      return new IncrementalChunkBuffer(capacity, increment, false);
```
adoroszlai, line 48: Can you please also change `TestChunkBuffer#runTestIncrementalChunkBuffer` to explicitly create `IncrementalChunkBuffer`?  Currently it uses this factory method, and so with this patch it really tests `ChunkBufferImplWithByteBuffer`.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -209,9 +224,16 @@ public IOException getIoException() {
   @Override
   public void write(int b) throws IOException {
     checkOpen();
-    byte[] buf = new byte[1];
-    buf[0] = (byte) b;
-    write(buf, 0, 1);
+    if (currentBufferRemaining == 0) {
+      allocateNewBuffer();
+    }
```
adoroszlai, line 229: For consistency, I think we should move the condition to `allocateNewBuffer` and rename it to `allocateNewBufferIfNeeded`.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -209,9 +224,16 @@ public IOException getIoException() {
...
+    currentBuffer.put((byte) b);
+    currentBufferRemaining--;
+    if (currentBufferRemaining == 0) {
+      writeChunk(currentBuffer);
+    }
```
adoroszlai, line 234: For consistency, I think we should move the condition to a new method `writeChunkIfNeeded`.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -481,7 +516,7 @@ private void handleFlush(boolean close)
     checkOpen();
     // flush the last chunk data residing on the currentBuffer
     if (totalDataFlushedLength < writtenDataLength) {
-      final ChunkBuffer currentBuffer = bufferPool.getCurrentBuffer();
+      currentBuffer = bufferPool.getCurrentBuffer();
```
adoroszlai, line 519: Should we also update `currentBufferRemaining`?

adoroszlai (commented):
Thanks @elek and @lokeshj1703 for working on this.  Is it intentionally created with target branch `apache:HDDS-4119` instead of `apache:master`?

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBuffer.java
elek: Wow, nice catch.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
elek: Not sure if I understood. Why is it more consistent to move this 3 lines to a separated method?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
elek: Yeah, I started to use the newly introduced method which is suggested by the first comment... (update current buffer + remaining)

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBuffer.java
bshashikant: IncrementalChunkBuffer was added to address cases where ozone client were running into OOM with keys less than chunk size , as without this, the smallest buffer which will be allocated will always be equal to the chunk size(4MB by default).

Please see https://issues.apache.org/jira/browse/HDDS-2331 for more details.
I would prefer to not remove this logic of incremental chunk buffer and may be hide it within an internal config.

@elek , how much of perf gain we will have of we still do incremental buffer allocation?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
adoroszlai: `checkOpen()` and `doFlushOrWatchIfNeeded()` calls are both concise.  I think moving the conditions for these two other calls would help make the level of abstraction in `write()` more similar.  But it's optional.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBuffer.java
elek: > @elek , how much of perf gain we will have of we still do incremental buffer allocation?

I can repeat the test to get exact numbers, but I couldn't get good performance without removing the incremental buffer. You can easily test it with the new unit test, if you write a lot of data with byte=1, it's still low.

> IncrementalChunkBuffer was added to address cases where ozone client were running into OOM with keys less than chunk size , as without this, the smallest buffer which will be allocated will always be equal to the chunk size(4MB by default).

I think it's a valid (and important question), but as far as I see it's safe to remove the `IncrementalByteBuffer`. As far as I see the situation is slightly different since HDDS-2331. I tried to test this patch with the commands from the HDDS-2331:

```
ozone freon rk --numOfThreads 1 --numOfVolumes 1 --numOfBuckets 1 --replicationType RATIS --factor ONE --keySize 1048576 --numOfKeys 5120 --bufferSize 65536
```

I couldn't reproduce the OOM.

Based on my understanding:

 1. We already have an increment by the `ByteBuffer` but size of the increment is 4MB (adding one more buffer when required)
 2. 4MB seems to be acceptable even with many clients in the same JVM, especially if we can have acceptable performance.

Let's say I have 100 Ozone clients (in the same JVM!!!) which write 1kb keys. I will have (4MB-1kb)  *100 overhead without the `IncrementalChunkBuffer` (as far as I understood). It's still <400MB in exchange for 30-100% performance gain. Sounds like a good deal.

But let me know if you see any problems here.

3. Let's say the 400MB overhead is unacceptable (or my calculation was wrong and the overhead is higher ;-) )

As far as I see the `BufferPool` is created per key. I think it would be possible to set the buffer size to `min(keySize, bufferSize)`. With this approach the first and only buffer of the BufferPool can have exactly the required size (which covers all the where the key size is < 4MB)

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
elek: Ok, I think I got the idea, but please double-check my commit as I am not sure ;-)

9ab01a7753b75009997cc33a338bbef282230687

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBuffer.java
adoroszlai: > the situation is slightly different since HDDS-2331

Note that default chunk size was 16MB at the time when HDDS-2331 was reported.  The benefit from IncrementalChunkBuffer is less now with 4MB default size.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
```diff
@@ -481,11 +514,9 @@ private void handleFlush(boolean close)
     checkOpen();
     // flush the last chunk data residing on the currentBuffer
     if (totalDataFlushedLength < writtenDataLength) {
-      final ChunkBuffer currentBuffer = bufferPool.getCurrentBuffer();
+      refreshCurrentBuffer(bufferPool);
       Preconditions.checkArgument(currentBuffer.position() > 0);
-      if (currentBuffer.hasRemaining()) {
-        writeChunk(currentBuffer);
```
adoroszlai, line 487: Note the condition is different here, might not be safe to replace with `writeChunkIfNeeded()`.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
adoroszlai: Thanks for the update.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBuffer.java
elek: I would suggest separating the two parts of the PR:

 1. reorganize the position calculation and allocation
 2. remove the usage of the Incremental buffer

While we can continue to searching for the safest method to do 2 (or do something instead of the removal), we can merge the first part where we already have an agreement.

elek:
See HDDS-4186 about the incremental chunk buffer.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChunkBuffer.java
elek: See #1374 about the 2nd.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockOutputStream.java
elek: Thanks the help. I fully reverted these lines in #23ba2d1 and build is green again.

elek:
Is there any more objections/comments? All the comments are addressed. Can we merge this?

bshashikant (approved):

elek:
Thanks the review @bshashikant and @adoroszlai 

I am merging it now...

PR #1339, opened 2020-08-19
HDDS-4131. Container report should update container key count and bytes used if they differ in SCM

sodonnel:

In HDDS-4037 it was noted that when blocks are deleted from closed containers, the bytesUsed and Key Count metrics on the SCM container are not updated correctly.

These stats should be updated via the container reports issued by the DNs to SCM periodically. However, in `AbstractContainerReportHandler#updateContainerStats`, the code assumes the values are always increasing and it will not update them if they are decreasing:

```
  private void updateContainerStats(final ContainerID containerId,
      throws ContainerNotFoundException {
    if (isHealthy(replicaProto::getState)) {
      final ContainerInfo containerInfo = containerManager
          .getContainer(containerId);

      if (containerInfo.getSequenceId() <
          replicaProto.getBlockCommitSequenceId()) {
        containerInfo.updateSequenceId(
            replicaProto.getBlockCommitSequenceId());
      }
      if (containerInfo.getUsedBytes() < replicaProto.getUsed()) {
        containerInfo.setUsedBytes(replicaProto.getUsed());
      }
      if (containerInfo.getNumberOfKeys() < replicaProto.getKeyCount()) {
        containerInfo.setNumberOfKeys(replicaProto.getKeyCount());
      }
    }
  }
```

In HDDS-4037 a change was made to the Replication Manager, so it updates the stats. However I don't believe that is the correct place to perform this check, and the issue is caused by the logic shared above.

In this Jira, I have removed the changes to Replication Manager in HDDS-4037 (but retained the other changes in that Jira), ensuring the problem statistics are only updated via the containers reports if they are different in SCM from what is reported.

https://issues.apache.org/jira/browse/HDDS-4131

## How was this patch tested?

Small change to existing unit test. Used it to reproduce the problem before making the changed.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/AbstractContainerReportHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/AbstractContainerReportHandler.java
@@ -20,6 +20,7 @@ package org.apache.hadoop.hdds.scm.container;
 
 import com.google.common.base.Preconditions;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.LifeCycleEvent;
 import org.apache.hadoop.hdds.protocol.proto
     .StorageContainerDatanodeProtocolProtos.ContainerReplicaProto;
@@ -28,6 +29,9 @@ import org.apache.hadoop.hdds.protocol.proto
 import org.slf4j.Logger;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
 import java.util.UUID;
 import java.util.function.Supplier;
 
@@ -76,7 +80,7 @@ public class AbstractContainerReportHandler {
     // Synchronized block should be replaced by container lock,
     // once we have introduced lock inside ContainerInfo.
     synchronized (containerManager.getContainer(containerId)) {
-      updateContainerStats(containerId, replicaProto);
+      updateContainerStats(datanodeDetails, containerId, replicaProto);
       updateContainerState(datanodeDetails, containerId, replicaProto);
       updateContainerReplica(datanodeDetails, containerId, replicaProto);
     }
@@ -90,7 +94,8 @@ public class AbstractContainerReportHandler {
    * @param replicaProto Container Replica information
    * @throws ContainerNotFoundException If the container is not present
    */
-  private void updateContainerStats(final ContainerID containerId,
+  private void updateContainerStats(final DatanodeDetails datanodeDetails,
+                                    final ContainerID containerId,
                                     final ContainerReplicaProto replicaProto)
       throws ContainerNotFoundException {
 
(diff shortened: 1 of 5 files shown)


ChenSammi:
Hi @sodonnel, since a typical container has 3 replicas and container report are asynchronously, we need a consensus on what's the container size is in SCM. Basically AbstractContainerReportHandler is not the perfect place to handle this because it doesn't have a global view while Replication Manager has.   
For OPEN container,  I think Math.min(replia1, replica2, replica3) is a safe way, and for CLOSED container, Math.max(replica1, replica2, replica3) is safer.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/AbstractContainerReportHandler.java
```diff
@@ -103,13 +108,44 @@ private void updateContainerStats(final ContainerID containerId,
         containerInfo.updateSequenceId(
             replicaProto.getBlockCommitSequenceId());
       }
-      if (containerInfo.getUsedBytes() != replicaProto.getUsed()) {
-        containerInfo.setUsedBytes(replicaProto.getUsed());
+      List<ContainerReplica> otherReplicas =
+          getOtherReplicas(containerId, datanodeDetails);
```
adoroszlai, line 112: I think we can skip filtering the replicas for "others", since the resulting count/bytes values will be the min/max of all replicas, including this one.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/AbstractContainerReportHandler.java
```diff
@@ -103,14 +108,44 @@ private void updateContainerStats(final ContainerID containerId,
         containerInfo.updateSequenceId(
             replicaProto.getBlockCommitSequenceId());
       }
+      List<ContainerReplica> otherReplicas =
+          getOtherReplicas(containerId, datanodeDetails);
```
adoroszlai, line 112: I think we can skip filtering the replicas for "others" (and omit the code for that), since the resulting count/bytes values are the min/max of all replicas, including this one, either way.  If we loop over all replicas, the iteration for the current replica will be a "no-op", not change the result.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/TestContainerReportHandler.java
```diff
@@ -483,9 +485,167 @@ public void testQuasiClosedToClosed()
...
+        containerId, state, dn.getUuidString(), bytesUsed, keyCount);
+
+    return new ContainerReportFromDatanode(dn, containerReport);
+  }
+
   private static ContainerReportsProto getContainerReportsProto(
       final ContainerID containerId, final ContainerReplicaProto.State state,
       final String originNodeId) {
+    return getContainerReportsProto(containerId, state, originNodeId,
+        100000000L, 2000000000L);
```
adoroszlai, line 643: Note: not a big deal, but it seems `usedBytes` and `keyCount` values are swapped compared to the original.

adoroszlai (commented):
Thanks @sodonnel for improving code responsibilities / fixing the original bug at the root cause.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/AbstractContainerReportHandler.java
sodonnel: I think we still do need to filter out the current replica. The reason is that at the point the stats are updated, the newly reported replica has not yet updated the previous values stored for it in SCM memory. This is due to the order of operations here, in AbstractContainerReportHandler:

```
  protected void processContainerReplica(final DatanodeDetails datanodeDetails,
      throws IOException {
    final ContainerID containerId = ContainerID
        .valueof(replicaProto.getContainerID());

    if (logger.isDebugEnabled()) {
      logger.debug("Processing replica of container {} from datanode {}",
          containerId, datanodeDetails);
    }
    // Synchronized block should be replaced by container lock,
    // once we have introduced lock inside ContainerInfo.
    synchronized (containerManager.getContainer(containerId)) {
      updateContainerStats(datanodeDetails, containerId, replicaProto);
      updateContainerState(datanodeDetails, containerId, replicaProto);
      updateContainerReplica(datanodeDetails, containerId, replicaProto);
    }
  }
```

The in-memory values for the reported replicas are not updated until the updateContainerReplica(...) call, but we updateContainerStats(...) before that.

For example, lets say we have an CLOSED container, so its the max of the values we use.

Existing replicas have values 4, 4, 5, so the value used is 5.

The new report changes the 5 to 4, so now we would have:

4, 4, 5 (stale value, will be updated soon), 4(new value of stale value).

If we don't filter out the stale replica, then the value will be wrong and left as 5 until the next container report is processed, when it would change to 4.

Does that make sense?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/TestContainerReportHandler.java
sodonnel: Ooops. That was not an intentional change. I will fix that just to keep things consistent with how they were before.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/AbstractContainerReportHandler.java
adoroszlai: Good point, I missed that.  Thanks for the explanation.

elek:
> In HDDS-4037 a change was made to the Replication Manager, so it updates the stats. However I don't believe that is the correct place to perform this check, and the issue is caused by the logic shared above.

Agree, -- based on my understanding -- it's better to do before replication manager:

 1. Replication manager can be more simple and easier to manage
 2. If we do the fix in the replication manager it's possible to have an inconsistent view between the container report and replication manager execution.

sodonnel:
@adoroszlai @ChenSammi Are you happy with this change at this stage? Can we commit it?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -398,8 +398,6 @@ public void updateDeleteTransactionId(Map<Long, Long> deleteTransactionMap)
               SCMException.ResultCodes.FAILED_TO_FIND_CONTAINER);
         }
         containerInfo.updateDeleteTransactionId(entry.getValue());
-        containerInfo.setNumberOfKeys(containerInfoInMem.getNumberOfKeys());
```
ChenSammi, line 401: Prefer to keep the KeyCount and UsedKeys persist action.

ChenSammi (commented):
Just one minor inline comment.  Good with the rest part.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
sodonnel: After looking at this area a bit more I understand why that is needed now. I have added those two lines back in.

sodonnel:
I think this change is good to commit now? @adoroszlai gave a thumbs up a few days back and I have addressed the only concern @ChenSammi raised.

I will commit tomorrow unless anyone objects before then.

ChenSammi:
+1.  Thanks @sodonnel for the contribution.

PR #1342, opened 2020-08-24
HDDS-4135: In ContainerStateManagerV2, modification of RocksDB should be consistent with that of memory state.

GlenGeng:

Fix a bug in https://issues.apache.org/jira/browse/HDDS-3895

In ContainerStateManagerV2, both disk state (column families in RocksDB) and memory state (container maps in memory) are protected by raft, and should keep their consistency upon each modification.

https://issues.apache.org/jira/browse/HDDS-4135

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerImpl.java
@@ -266,6 +266,8 @@ public final class ContainerStateManagerImpl
         containers.addContainer(container);
         pipelineManager.addContainerToPipeline(pipelineID, containerID);
       } catch (Exception ex) {
+        LOG.warn("Exception {} occurred while add Container {}.",
+            ex, containerInfo);
         containers.removeContainer(containerID);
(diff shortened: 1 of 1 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerImpl.java
```diff
@@ -334,8 +347,12 @@ ContainerInfo getMatchingContainer(final long size, String owner,
     throw new UnsupportedOperationException("Not yet implemented!");
   }
 
-  public void removeContainer(final HddsProtos.ContainerID id) {
-    containers.removeContainer(ContainerID.getFromProtobuf(id));
+  @Override
+  public void removeContainer(final HddsProtos.ContainerID id)
```
amaliujia, line 351: So there is no easy way to test such changes?

GlenGeng: we can verify the state of memory and RocksDB in Unit Test.

GlenGeng:
Close this PR, since @nandakumar131 would fix this issue in his future PR for container V2.

PR #1344, opened 2020-08-24
HDDS-4140. Auto-close /pending pull requests after 21 days of inactivity

elek:

Earlier we introduced a way to mark the inactive pull requests with "pending" label (with the help of /pending comment).

This pull requests introduce a new scheduled build which closes the "pending" pull requests after 21 days of inactivity.

IMPORTANT: Only the pull requests which are pending on the author will be closed.

We should NEVER close a pull requests which are waiting for the attention of a committer.

https://issues.apache.org/jira/browse/HDDS-4140

## How was this patch tested?

on the https://github.com/elek/hadoop-ozone/ fork with this PR: https://github.com/elek/hadoop-ozone/pull/13

### DIFF
--- a/.github/workflows/close-pending.yaml
+++ b/.github/workflows/close-pending.yaml
@@ -16,7 +16,7 @@ name: close-prs
 
 on:
   schedule:
-  - cron:  '*/5 * * * *'
+  - cron:  '0 0 * * *'
 
 jobs:
   close-pending:
--- a/.github/workflows/close-pending.yaml
+++ b/.github/workflows/close-pending.yaml
@@ -26,6 +26,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@master
       - name: Execute close-pending script
+        if: github.repository == 'apache/hadoop-ozone'
         run: ./.github/close-pending.sh
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


.github/workflows/close-pending.yaml
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+name: close-prs
+
+on:
+  schedule:
+  - cron:  '*/5 * * * *'
```
adoroszlai, line 19: I think it would be better to run this workflow once per day instead of every 5 minutes.  The list of pending PRs to be closed is filtered only by date (eg. 2020-08-03), not time, so most of the runs would be no-ops.

adoroszlai (commented):
Thanks @elek for implementing this.

.github/workflows/close-pending.yaml
elek: Absolutely, I started with one day but changed it to five minutes to show if it works. Changing back.

adoroszlai (approved):

elek:
Thanks the review @adoroszlai. Let me merge it and :fire: the :bomb: 

(I am planning to create a filter to review pending + closed and check it sometimes...)

PR #1346, opened 2020-08-25
HDDS-4115. CLI command to show current SCM leader and follower status.

amaliujia:

CLI command to show current SCM leader and follower status. E.g. `ozone admin scmha listratisstatus`

https://issues.apache.org/jira/browse/HDDS-4115

## How was this patch tested?

Unit Test

Command line output
```
[180.3.14.5:2333, 180.3.14.21:2333, 180.3.14.145:2333]
```

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
@@ -229,6 +229,4 @@ public interface StorageContainerLocationProtocol extends Closeable {
    * @return True if ReplicationManager is running, false otherwise.
    */
   boolean getReplicationManagerStatus() throws IOException;
-
-  List<String> getScmRatisStatus() throws IOException;
 }
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
@@ -230,4 +230,5 @@ public interface StorageContainerLocationProtocol extends Closeable {
    */
   boolean getReplicationManagerStatus() throws IOException;
 
+  List<String> getScmRatisStatus() throws IOException;
 }
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
@@ -229,4 +229,5 @@ public interface StorageContainerLocationProtocol extends Closeable {
    * @return True if ReplicationManager is running, false otherwise.
    */
   boolean getReplicationManagerStatus() throws IOException;
+
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
@@ -441,7 +441,7 @@ public final class StorageContainerLocationProtocolServerSideTranslatorPB
     return HddsProtos.GetScmInfoResponseProto.newBuilder()
         .setClusterId(scmInfo.getClusterId())
         .setScmId(scmInfo.getScmId())
-        .addAllPeerStatus(scmInfo.getRatisPeerStatus())
+        .addAllPeerRoles(scmInfo.getRatisPeerRoles())
         .build();
   }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
@@ -61,6 +61,8 @@ import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolPro
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.ScmContainerLocationRequest;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.ScmContainerLocationResponse;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.ScmContainerLocationResponse.Status;
+import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.ScmHARatisStatusRequestProto;
+import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.ScmHARatisStatusResponseProto;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.StartReplicationManagerRequestProto;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.StartReplicationManagerResponseProto;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.StopReplicationManagerRequestProto;
@@ -271,6 +273,14 @@ public final class StorageContainerLocationProtocolServerSideTranslatorPB
             .setGetSafeModeRuleStatusesResponse(getSafeModeRuleStatues(
                 request.getGetSafeModeRuleStatusesRequest()))
             .build();
+      case ScmHAStatus:
+        return
+            ScmContainerLocationResponse.newBuilder()
+            .setCmdType(request.getCmdType())
+            .setStatus(Status.OK)
+            .setGetScmHARatisStatusResponse(
+                getScmHARatisStatusResponseProto(request.getGetScmHARatisStatusRequest())
+            ).build();
       default:
         throw new IllegalArgumentException(
             "Unknown command type: " + request.getCmdType());
@@ -466,6 +476,12 @@ public final class StorageContainerLocationProtocolServerSideTranslatorPB
     }
     return GetSafeModeRuleStatusesResponseProto.newBuilder()
         .addAllSafeModeRuleStatusesProto(proto).build();
+  }
+
+  public ScmHARatisStatusResponseProto getScmHARatisStatusResponseProto(
+      ScmHARatisStatusRequestProto requestProto) throws IOException {
+    List<String> address = impl.getScmRatisStatus();
+    return ScmHARatisStatusResponseProto.newBuilder().addAllRaftPeerAddress(address).build();
   }
 
   public ForceExitSafeModeResponseProto forceExitSafeMode(
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
@@ -469,12 +469,6 @@ public final class StorageContainerLocationProtocolServerSideTranslatorPB
         .addAllSafeModeRuleStatusesProto(proto).build();
   }
 
-  // public ScmHARatisStatusResponseProto getScmHARatisStatusResponseProto(
-  //     ScmHARatisStatusRequestProto requestProto) throws IOException {
-  //   List<String> address = impl.getScmRatisStatus();
-  //   return ScmHARatisStatusResponseProto.newBuilder().addAllRaftPeerAddress(address).build();
-  // }
-
   public ForceExitSafeModeResponseProto forceExitSafeMode(
       ForceExitSafeModeRequestProto request)
       throws IOException {
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
@@ -61,8 +61,6 @@ import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolPro
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.ScmContainerLocationRequest;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.ScmContainerLocationResponse;
 import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.ScmContainerLocationResponse.Status;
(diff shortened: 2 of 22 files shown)


amaliujia:
R: @timmylicheng 

This is still WIP. Can you give some suggestions. Just want to make sure I am on the right track.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
```diff
@@ -560,6 +563,13 @@ public boolean getReplicationManagerStatus() {
     return scm.getReplicationManager().isRunning();
   }
 
+  @Override
+  public List<String> getScmRatisStatus() throws IOException {
+    return scm.getScmHAManager()
+        .getRatisServer().getRaftPeers()
+        .stream().map(peer -> peer.getAddress()).collect(Collectors.toList());
```
amaliujia, line 570: One thing I am not sure is what should be fetched from Ratis peers. Should I also include which one is leader and which one is follower?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
```diff
@@ -230,4 +230,5 @@ Pipeline createReplicationPipeline(HddsProtos.ReplicationType type,
    */
   boolean getReplicationManagerStatus() throws IOException;
 
+  List<String> getScmRatisStatus() throws IOException;
```
amaliujia, line 233: Does this API belongs to `StorageContainerLocationProtocol`?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestScmAdminHA.java
```diff
@@ -0,0 +1,66 @@
...
+      cluster.shutdown();
+    }
+  }
+
+  @Test
+  public void testGetLeaderStatus() {
+    InetSocketAddress address = cluster.getStorageContainerManager().getClientRpcAddress();
+    String hostPort = address.getHostName() + ":" + address.getPort();
+    String[] args = {"--scm", hostPort, "scmha", "listratisstatus"};
+    ozoneAdmin.execute(args);
```
amaliujia, line 64: This is actually not a unit test as it does not check what is returned. I will think about whether there is a simple way to test that API directly.

timmylicheng:
Could you attach your CLI command output?

amaliujia:
@timmylicheng 

The output is empty array. In the test setup, `scm.getScmHAManager().getRatisServer().getRaftPeers();` return empty list. My current guess is I haven't enabled SCM HA in the test setup.

I expect it will print something like `[127.0.1.1:1313, 127.0.1.1:1314, 127.0.1.1:1315]`. So I guess this JIRA requires a format of, e.g., `[leader: 127.0.1.1:1313, follower: 127.0.1.1:1314, follower: 127.0.1.1:1315]`?

So if my approach looks reasonable, I can continue working on this version to have a valid test somehow.

amaliujia:
Ok I think I can work to have an example output first.

amaliujia:
@timmylicheng 

I found the reason why in my current test empty array was printed. It is because the test is using `MockSCMHaManager`: https://github.com/apache/hadoop-ozone/blob/HDDS-2823/hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/MockSCMHAManager.java#L191

I manually add a peer locally and get the output print: `[localhost:7238]`, which is expected.


Any suggestion how should I proceed?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
```diff
@@ -271,6 +273,14 @@ public ScmContainerLocationResponse processRequest(
             .setGetSafeModeRuleStatusesResponse(getSafeModeRuleStatues(
                 request.getGetSafeModeRuleStatusesRequest()))
             .build();
+      case ScmHAStatus:
```
timmylicheng, line 276: GetSCMRatisRole will be better

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
timmylicheng: Could potentially throw NPE

timmylicheng: I can see this method be wrapper in ScmHAManager to show a list of scm hosts.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/GetScmRatisStatusSubcommand.java
```diff
@@ -0,0 +1,26 @@
+package org.apache.hadoop.ozone.admin.scm;
+
+import java.util.List;
+import java.util.concurrent.Callable;
+import org.apache.hadoop.hdds.cli.HddsVersionProvider;
+import org.apache.hadoop.hdds.scm.client.ScmClient;
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "listratisstatus",
```
timmylicheng, line 10: Name seems a bit verbose. Let me find a syntax to align with OM HA

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestScmAdminHA.java
```diff
@@ -0,0 +1,66 @@
+package org.apache.hadoop.ozone.shell;
```
timmylicheng, line 1: Apart from UT, we can have an acceptance test to test CLI. 
You may find examples here: https://github.com/apache/hadoop-ozone/pull/375

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
timmylicheng: Any way to merge it with GetSCMInfo?

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/GetScmRatisStatusSubcommand.java
adoroszlai: I think both this one and OM HA `getserviceroles` can be improved.

amaliujia: @adoroszlai 

What kind of improvement OM HA `getserviceroles` you are thinking of? I am happy to make a separate PR for that.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
amaliujia: Got it. Will move this functionality to ScmHAManager.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/GetScmRatisStatusSubcommand.java
adoroszlai: `status` or `roles` would probably be enough to indicate the goal of the subcommand, something like

```
ozone admin (om|scm) status
```

amaliujia: Thanks. I will adopt `ozone admin scm status` in this PR and I will send another PR for ``ozone admin om status`

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
amaliujia: Pushed a new commit to merge this logic into `GetScmInfo`

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestScmAdminHA.java
amaliujia: ack

amaliujia:
Addressed the following comments
1. Merge `getRatisStatus` with `GetScmInfo`
2. Adopt command syntax `ozone admin scm status`
3. added an acceptance test

@timmylicheng I am not sure how to test an acceptance test. Can you share a way to run it locally?

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/GetScmRatisStatusSubcommand.java
timmylicheng: ozone admin om getserviceroles -id=<<service id>>

This is what OM does. I have my +1 on ozone admin (om|scm) roles. Status is more like health check.

amaliujia: I am ok with both. 

@adoroszlai  what do you think?

adoroszlai: I think it depends on whether you want to keep this command specific to roles, or may extend the same command in the future with other status info.  Probably "roles" is better now, and "status" can be either a separate command or another alias later.

amaliujia: that makes sense. Indeed `status` sounds more like a health check and can carry much more information.

Consider `OM` is also actually getting `roles`. We can start from `roles`

amaliujia:
Uploaded one commit to 
1. use `ozone admin scm roles`
2. fix the acceptance test.

amaliujia:
Also R: @nandakumar131  can you please take a look?

amaliujia:
R @timmylicheng @nandakumar131 

I am thinking maybe we can first merge this PR and create a JIRA to track left work. Right now per feedback this command could print more information about Ratis peers, e.g. leader/follower roles, leader term, etc.

I took a look at how does OM HA does: https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java#L2535

Basically it seems such request will directly hit the leader OM, then get status will be much easier. Currently in SCM HA we haven't reached to the point with robust Ratis setup.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/GetScmRatisStatusSubcommand.java
```diff
@@ -0,0 +1,46 @@
...
+    versionProvider = HddsVersionProvider.class)
+public class GetScmRatisStatusSubcommand implements Callable<Void> {
+
+  @CommandLine.ParentCommand
+  private ScmAdmin parent;
+
+  @Override
+  public Void call() throws Exception {
+    ScmClient scmClient = parent.createScmClient();
+    List<String> status = scmClient.getScmRatisStatus();
```
timmylicheng, line 42: We probably want to align the naming here. Either we use roles or status for all external and internal interfaces.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/ScmAdmin.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.hadoop.hdds.scm.client.ScmClient;
+import org.apache.hadoop.ozone.admin.OzoneAdmin;
+import picocli.CommandLine;
+import picocli.CommandLine.Model.CommandSpec;
+import picocli.CommandLine.Spec;
+
+/**
+ * Subcommand for admin operations related to SCM.
+ */
+@CommandLine.Command(
```
timmylicheng, line 31: why do we need a separate scmAdmin? 
Shall we have like ozone admin getSCMRoles?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestScmAdminHA.java
```diff
@@ -0,0 +1,79 @@
...
+  private static String scmId;
+  private static MiniOzoneCluster cluster;
+
+  @BeforeClass
+  public static void init() throws Exception {
+    ozoneAdmin = new OzoneAdmin();
+    conf = new OzoneConfiguration();
+
+    // Init HA cluster
+    omServiceId = "om-service-test1";
```
timmylicheng, line 48: You mean SCM here?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/MockSCMHAManager.java
```diff
@@ -107,6 +108,14 @@ public void shutdown() throws IOException {
     ratisServer.stop();
   }
 
+  @Override
+  public List<String> getRatisStatus() {
+    return Arrays.asList(
```
timmylicheng, line 113: You can use the default SCM RatisServer port. Check ratisBindPort in SCMHAConfiguration

timmylicheng:
Thanks @amaliujia for the contribution. The patch looks good overall. Just a few comments inline.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/GetScmRatisStatusSubcommand.java
amaliujia: Good point. I have switched all `status` to `roles`

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestScmAdminHA.java
amaliujia: Actually I meant OM. It turns out that I cannot ignore `omServiceId` to start the cluster (there is a check for this service id).

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/MockSCMHAManager.java
amaliujia: Switched to use the default port. In the future I will think about how to better construct `MockSCMHAManager` for testing purpose (as the code is evolving we can keep updating `MockSCMHAManager`)

amaliujia:
@timmylicheng  comments addressed. Can you take another look?

bshashikant:
The CLI should be generic for both OM and SCM and probably extended to Datanodes as well.

amaliujia:
Re @bshashikant 

Agreed. Right now the command itself is unified (for both OM and SCM, we name this command as `roles`). Then we should unify the behavior fo both commands (and if there is DN command, that should be the same).

timmylicheng:
+1. Thanks for Rui's contribution.
Merging

PR #1348, opened 2020-08-25
Update version info in hadoop client dependency readme

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-4144


ChenSammi:
@elek @adoroszlai  could you help to review the patch?

elek (approved):
+1 LGTM

I am not sure if we must update this README for all the releases. It's just a helper section to check the newly introduced dependencies in case of a version upgrade.

For example when whe upgrade to hadoop 3.3, we can compare the new dependencies to the reference one and check if any new dependencies are introduced. (In fact we can remove this part from the README).

But there is no problem with updating it (1.0.0 is nicer here ;-) )

ChenSammi:
I think the most important part of the readme is tell user how to find the dependencies using "mvn dependency:tree ".  So we can position the left part(the command output) as an example,  then there is no need to update it during every release.

PR #1350, opened 2020-08-25
HDDS-4146. Show the ScmId and ClusterId in the scm web ui.

maobaolong:

Show the ScmId and ClusterId in the scm web ui, so that we can easy to identify the cluster.

HDDS-4146

## How was this patch tested?

Start scm web ui and check the homepage, a SCM Information table will appear.


xiaoyuyao:
Thanks @maobaolong  for the patch. LGTM, +1.

PR #1351, opened 2020-08-25
HDDS-4121. Implement OmMetadataMangerImpl#getExpiredOpenKeys.

errose28:

Create an implementation to the existing OmMetadataMangerImpl#getExpiredOpenKeys method stub as the first step towards completing parent Jira HDDS-4120. This method returns a list of key information for all keys in the OM open key table whose creation time is older than a time interval configurable with the existing ozone.open.key.expire.threshold setting. The existing default value for this setting (one day) was not changed.

HDDS-4121

## How was this patch tested?

The unit test TestOmMetadataManager#testGetExpiredOpenKeys was added, which creates a mixture of expired and unexpired open keys, and tests whether the expired open keys are read correctly. Functionality was added to TestOMRequestUtils to allow setting customized creation times for keys to simulate expiration.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
```diff
@@ -186,10 +186,11 @@ OmKeyLocationInfo allocateBlock(OmKeyArgs args, long clientID,
    * the key name and all its associated block IDs. A pending open key has
    * prefix #open# in OM DB.
    *
+   * @param count The maximum number of expired opne keys to return.
    * @return a list of {@link BlockGroup} representing keys and blocks.
    * @throws IOException
    */
-  List<BlockGroup> getExpiredOpenKeys() throws IOException;
+  List<BlockGroup> getExpiredOpenKeys(int count) throws IOException;
```
bharatviswa504, line 193: Here, we can collect just expired KeyNames.

bharatviswa504: And also I think we might need last expired key also, as in next iteration we need to start from that one.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
```diff
@@ -186,10 +186,11 @@ OmKeyLocationInfo allocateBlock(OmKeyArgs args, long clientID,
    * the key name and all its associated block IDs. A pending open key has
    * prefix #open# in OM DB.
    *
+   * @param count The maximum number of expired opne keys to return.
```
bharatviswa504, line 189: opne -> open

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
bharatviswa504: You can ignore my 2nd comment, as delete expiredKeys will be done in a single iteration, those expired keys will not be there in the openKeyTable.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManager.java
```diff
@@ -186,10 +186,11 @@ OmKeyLocationInfo allocateBlock(OmKeyArgs args, long clientID,
    * the key name and all its associated block IDs. A pending open key has
    * prefix #open# in OM DB.
```
errose28, line 187: Is this line in the original documentation still relevant? It looks like open keys are placed in the open key table unprefixed.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/TestOmMetadataManager.java
```diff
@@ -521,6 +527,80 @@ public void testListKeysWithFewDeleteEntriesInCache() throws Exception {
...
+    // Add unexpired keys to open key table.
+    for (int i = 0; i < numUnexpiredOpenKeys; i++) {
+      OmKeyInfo keyInfo = TestOMRequestUtils.createOmKeyInfo(volumeName,
+              bucketName, "unexpired" + i, HddsProtos.ReplicationType.RATIS,
+              HddsProtos.ReplicationFactor.ONE);
+
+      TestOMRequestUtils.addKeyToTable(true, false,
+              keyInfo, clientID, 0L, omMetadataManager);
+
+      String groupID = omMetadataManager.getOpenKey(volumeName, bucketName,
```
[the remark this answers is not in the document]
errose28: Yeah. Removed it and pushing again.

bharatviswa504:
Thank You @errose28 for the contribution.

PR #1352, opened 2020-08-25
HDDS-4147. Add OFS to FileSystem META-INF

smengcl:

With this change, `fs.ofs.impl` won't have to be explicitly set in `core-site.xml` when using OFS on a client.

This change is derived from https://github.com/apache/hadoop-ozone/pull/1088

https://issues.apache.org/jira/browse/HDDS-4147

## How was this patch tested?

Existing OFS tests shall pass even with `conf.set("fs.ofs.impl", ...)` removed.

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
+++ b/hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+CORE-SITE.XML_fs.defaultFS=ofs://om
+
 OZONE-SITE.XML_ozone.csi.owner=hadoop
 OZONE-SITE.XML_ozone.csi.socket=/tmp/csi.sock
 
--- a/hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
+++ b/hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
@@ -14,8 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-CORE-SITE.XML_fs.defaultFS=ofs://om
-
 OZONE-SITE.XML_ozone.csi.owner=hadoop
 OZONE-SITE.XML_ozone.csi.socket=/tmp/csi.sock
 
--- a/hadoop-ozone/dist/src/main/compose/ozone-om-ha/docker-config
+++ b/hadoop-ozone/dist/src/main/compose/ozone-om-ha/docker-config
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 CORE-SITE.XML_fs.defaultFS=ofs://omservice
-
 OZONE-SITE.XML_ozone.om.service.ids=omservice
 OZONE-SITE.XML_ozone.om.nodes.omservice=om1,om2,om3
 OZONE-SITE.XML_ozone.om.address.omservice.om1=om1
--- a/hadoop-ozone/dist/src/main/compose/ozone-om-ha/docker-config
+++ b/hadoop-ozone/dist/src/main/compose/ozone-om-ha/docker-config
@@ -14,8 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-CORE-SITE.XML_fs.o3fs.impl=org.apache.hadoop.fs.ozone.OzoneFileSystem
-CORE-SITE.XML_fs.defaultFS=o3fs://bucket1.volume1.omservice
+CORE-SITE.XML_fs.defaultFS=ofs://omservice
 OZONE-SITE.XML_ozone.om.service.ids=omservice
 OZONE-SITE.XML_ozone.om.nodes.omservice=om1,om2,om3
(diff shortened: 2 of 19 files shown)


hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+CORE-SITE.XML_fs.defaultFS=ofs://om
```
elek, line 17: Why is this line required to use META-INF based discovery?

hadoop-ozone/dist/src/main/compose/ozone-om-ha/docker-config
```diff
@@ -14,8 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-CORE-SITE.XML_fs.o3fs.impl=org.apache.hadoop.fs.ozone.OzoneFileSystem
-CORE-SITE.XML_fs.defaultFS=o3fs://bucket1.volume1.omservice
+CORE-SITE.XML_fs.defaultFS=ofs://omservice
```
elek, line 17: Why do we need this line to use META-INF services based discovery?

hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
elek: BTW why do we need defaultFS for the CSI test? How is it used.

elek (changes requested):
Thanks the patch @smengcl, looks good to me. Good simplification.

I think the patch contains some unnecessary lines from the original patch which are not related to the change to use META-INF/services discovery instead of configuration.

hadoop-ozone/dist/src/main/compose/ozone-om-ha/docker-config
adoroszlai: I think only MR tests (`ozone-mr` and `ozonesecure-mr`) use `defaultFS` when setting up MR environment:

https://github.com/apache/hadoop-ozone/blob/8102ac799816f417e096873b6f351fd531f7bfbf/hadoop-ozone/dist/src/main/smoketest/createmrenv.robot#L46-L47

Changing `defaultFS` to `ofs://` in these environments can help validate the META-INF changes.

Changing/adding `defaultFS` in the other ones is not needed IMO.

hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
smengcl: Hi @elek , in most automated robot tests `fs.defaultFS` isn't really used.
Mostly it improves QoL when user logs into the docker-compose cluster and manually checks for something (e.g. `docker-compose exec om /bin/bash` then `ozone fs -ls /`).

adoroszlai (approved):
Thanks @smengcl for updating the patch.  It's much cleaner now.  I have verified it with my changes for HDDS-3840.

hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
elek: > Mostly it improves QoL when user logs into the docker-compose cluster and manually checks for something (e.g. docker-compose exec om /bin/bash then ozone fs -ls /).

Good point. And I agree to support debug `ofs` is better choice as all the volumes are visible by default.

hadoop-ozone/dist/src/main/compose/ozone-om-ha/docker-config
elek: AFAIK these are just helper lines, not really the test of the file system. META-INF/services are already tested as test.sh is modified to do both o3fs/ofs tests.

I am fine to replace all the `defaultFS` just let's agree first what do we need to test. Do we need different MR test with both kind of `defaultFs`? 

Or is it enough to test o3fs/ofs low level functionality and MR can use just one of them?

(And you see it's a different discussion, that's the reason why I suggested to do it in a separated PR. But can be included if you think it's better.)

adoroszlai: > META-INF/services are already tested as test.sh is modified to do both o3fs/ofs tests.

Got it.

> Do we need different MR test with both kind of `defaultFs`?

I'm doing that in HDDS-3840.  I'll pass it via `-D` command-line option, as different value is needed for each test in the same cluster.

I guess I can retain the cluster-level config for convenience.

hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
smengcl: Thanks @elek . I will file another jira to add those trivial docker-config changes then.

smengcl: Filed https://issues.apache.org/jira/browse/HDDS-4161

smengcl:
Thanks @adoroszlai for verifying the changes.

@elek Is there anything left to be desired for this PR?

PR #1357, opened 2020-08-26
HDDS-4153. Increase default timeout in kubernetes tests

elek:

Kubernetes tests are timing out sometimes. (eg. here: https://github.com/elek/ozone-build-results/tree/master/2020/08/26/2562/kubernetes)

Based on the log, SCM couldn't move out from safe mode. It's either a real issue or github environment is slow sometimes.

To make it clear what is the problem I propose to increase the default timeout from 90 sec to 300 sec (5 min).

https://issues.apache.org/jira/browse/HDDS-4153

## How was this patch tested?

CI (did the same two days ago in https://github.com/elek/ozone-perf-env)

### DIFF
--- a/hadoop-ozone/dist/src/main/k8s/examples/testlib.sh
+++ b/hadoop-ozone/dist/src/main/k8s/examples/testlib.sh
@@ -17,14 +17,14 @@
 
 retry() {
    n=0
-   until [ $n -ge 30 ]
+   until [ $n -ge 100 ]
    do
       "$@" && break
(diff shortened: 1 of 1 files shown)


adoroszlai (approved):
Thanks @elek, I think this will help mitigate intermittent failures.  The check currently takes around 20 minutes for a complete build and 4 tests.  With this change in the worst case it would take 14 minutes more than earlier, which is still far from the longest check (around 1 hour).

PR #1358, opened 2020-08-27
HDDS-4143. Implement a version factory for OM Apply Transaction that …

avijayanhwx:
…uses the implementation based on layout version.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/utils/OzoneManagerRatisUtils.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/utils/OzoneManagerRatisUtils.java
@@ -25,39 +25,16 @@ import org.apache.hadoop.ozone.om.OmMetadataManagerImpl;
 import org.apache.hadoop.ozone.om.OzoneManager;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
 import org.apache.hadoop.ozone.om.ratis.OMTransactionInfo;
-import org.apache.hadoop.ozone.om.request.bucket.OMBucketCreateRequest;
-import org.apache.hadoop.ozone.om.request.bucket.OMBucketDeleteRequest;
-import org.apache.hadoop.ozone.om.request.bucket.OMBucketSetPropertyRequest;
 import org.apache.hadoop.ozone.om.request.OMClientRequest;
 import org.apache.hadoop.ozone.om.request.bucket.acl.OMBucketAddAclRequest;
 import org.apache.hadoop.ozone.om.request.bucket.acl.OMBucketRemoveAclRequest;
 import org.apache.hadoop.ozone.om.request.bucket.acl.OMBucketSetAclRequest;
-import org.apache.hadoop.ozone.om.request.file.OMDirectoryCreateRequest;
-import org.apache.hadoop.ozone.om.request.file.OMFileCreateRequest;
-import org.apache.hadoop.ozone.om.request.key.OMKeysDeleteRequest;
-import org.apache.hadoop.ozone.om.request.key.OMAllocateBlockRequest;
-import org.apache.hadoop.ozone.om.request.key.OMKeyCommitRequest;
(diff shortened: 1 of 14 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OmVersionFactory.java
```diff
@@ -0,0 +1,133 @@
...
+      String type, OMLayoutFeature layoutFeature,
+      Function<OMRequest, OMClientRequest> function) {
+    VersionFactoryKey key = new VersionFactoryKey.Builder()
+        .key(type).version(layoutFeature.layoutVersion()).build();
+    writeRequestFactory.register(key, function);
+  }
+
+  public OMClientRequest getClientRequest(OMRequest omRequest) {
+    Function<OMRequest, ? extends OMClientRequest> instance =
+        writeRequestFactory.getInstance(getVersionFactoryKey(omRequest));
```
linyiqun, line 118: When input omRequest type is not existed, we should return IllegalStateException error as original logic OzoneManagerRatisUtils#createClientRequest did.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/utils/OzoneManagerRatisUtils.java
```diff
@@ -84,86 +61,26 @@
...
   private OzoneManagerRatisUtils() {
   }
-  /**
-   * Create OMClientRequest which encapsulates the OMRequest.
-   * @param omRequest
-   * @return OMClientRequest
-   * @throws IOException
-   */
-  public static OMClientRequest createClientRequest(OMRequest omRequest) {
-    Type cmdType = omRequest.getCmdType();
```
linyiqun, line 94: Can we keep OzoneManagerRatisUtils#createClientRequest unchanged and invoke omVersionFactory#getClientRequest Here?
Now OzoneManagerRatisUtils defines the request/response functions, I prefer to keep this function there. This is just my personal taste for this change.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/upgrade/TestOmVersionFactory.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMRequest;
+import org.apache.hadoop.ozone.upgrade.VersionFactoryKey;
+import org.apache.hadoop.ozone.upgrade.VersionedInstanceSupplierFactory;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Test OmVersionFactory.
+ */
+public class TestOmVersionFactory {
```
linyiqun, line 39: Can we add a test case that using existed OM request and verify its layout version is set as expected?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
```diff
@@ -290,6 +290,11 @@ public void validateRequest(OMRequest omRequest) throws OMException {
       throw new OMException("ClientId is null",
           OMException.ResultCodes.INVALID_REQUEST);
     }
+
+    if (omRequest.getLayoutVersion() == null) {
```
linyiqun, line 294: We might add a comment to say why we should pass layout version in write type request. That will be good understanding.

linyiqun (commented):
Hi @avijayanhwx , thanks for attaching the PR.
Overall looks good, just some my minor comments.

avijayanhwx:
Thanks for the comments @linyiqun. I had closed the PR since I am changing the registration logic of new OM client requests, to an automatic registration based on Annotation. I will keep your comments in mind in my next PR.

PR #1361, opened 2020-08-27
HDDS-4155. Directory and filename can end up with same name in a path.

bharatviswa504:

Disallow creation/complete MPU when a directory already exists with same name.

https://issues.apache.org/jira/browse/HDDS-4155

## How was this patch tested?

Added integration tests.

Tested with S3 with this flag manually. Tests for this with enable will be added as part of HDDS-4154


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -158,6 +159,23 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       validateBucketAndVolume(omMetadataManager, volumeName, bucketName);
 
+      // Check by the time we commit this file, is there any directory
```
arp7, line 162: @bharatviswa504 can you please reword this comment? I couldn't understand what it is saying.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -158,6 +159,23 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+      // created if then fail it. This is being done for correctness.
+
+      // For now we are taking this approach similar to local file creation.
+      // TODO: In future this can be revisited to come up with an advanced
+      // approach.
+
+
+      if (ozoneManager.getEnableFileSystemPaths()) {
+        if (checkDirectoryAlreadyExists(volumeName, bucketName, keyName,
```
arp7, line 171: Does the directory name have a trailing `/` in RocksDB? If so, do we need to append it to keyname before calling `checkDirectoryAlreadyExists`?

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/endpoint/ObjectEndpoint.java
```diff
@@ -197,6 +198,18 @@ public Response put(
           .build();
     } catch (IOException ex) {
       LOG.error("Exception occurred in PutObject", ex);
+      if (ex instanceof  OMException) {
+        if (((OMException) ex).getResult() == ResultCodes.NOT_A_FILE) {
+          OS3Exception os3Exception = S3ErrorTable.newError(INVALID_REQUEST,
+              keyPath);
+          os3Exception.setErrorMessage("An error occurred (InvalidRequest) " +
```
arp7, line 205: Nice detailed explanation, thanks for adding this!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
bharatviswa504: >Does the directory name have a trailing / in RocksDB? 
Yes.
That is taken care of by the implementation of checkDirectoryAlreadyExists.

Where it uses getOzoneDirKey which appends "/"

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
bharatviswa504: done

bharatviswa504:
I have addressed your review comments.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSWithObjectStoreCreate.java
```diff
@@ -209,6 +215,94 @@ public void testObjectStoreCreateWithO3fs() throws Exception {
...
+    OzoneBucket ozoneBucket = ozoneVolume.getBucket(bucketName);
+
+    OzoneOutputStream ozoneOutputStream =
+        ozoneBucket.createKey("/a/b/c", 10);
+    byte[] b = new byte[10];
+    Arrays.fill(b, (byte)96);
+    ozoneOutputStream.write(b);
+
+    // Before close create directory with same name.
+    o3fs.mkdirs(new Path("/a/b/c"));
```
arp7, line 234: Do we have a test case that tests the reverse scenario? i.e. create the dir first, then try to create a file with the same name. Also are we testing this via the object store interface with the flag enabled?

arp7 (commented):
LGTM. One question on test coverage.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSWithObjectStoreCreate.java
bharatviswa504: >Do we have a test case that tests the reverse scenario? i.e. create the dir first, then try to create a file with the same >name
No such test is currently available. Will post shortly this test also.

>Also are we testing this via the object store interface with the flag enabled?
Yes test is with flag enabled.

bharatviswa504: Added the test.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSWithObjectStoreCreate.java
```diff
@@ -303,6 +304,12 @@ public void testMPUFailDuetoDirectoryCreationBeforeComplete()
 
   }
 
+  @Test(expected = FileAlreadyExistsException.class)
```
arp7, line 307: Thanks for the additional test. Can we also add a test case that exercises the REST API to create a file and dir with the same name, in both orders?

bharatviswa504: You mean use OzoneBucket or add a robot test using S3.
If robot test, we don't have test suite with this flag enabled, opened a jira 
https://issues.apache.org/jira/browse/HDDS-4154

bharatviswa504: Added a test

arp7 (approved):

bharatviswa504:

PR #1363, opened 2020-08-28

smengcl:
This PR is derived from https://github.com/apache/hadoop-ozone/pull/1088 as the previous one is split into multiple jiras and needs rebase. I'm hoping opening up a brand new PR for the original jira would make things simpler.

Use ClientProtocol directly in Adapter and FS.

## How was this patch tested?

Existing tests should pass.

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/ObjectStore.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/ObjectStore.java
@@ -81,7 +81,9 @@ public class ObjectStore {
     proxy = null;
   }
 
-  @VisibleForTesting
+  /**
+   * Returns the ClientProtocol of the ObjectStore.
+   */
   public ClientProtocol getClientProxy() {
     return proxy;
   }
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -107,7 +107,7 @@ public class TestRootedOzoneFileSystem {
   private static FileSystem fs;
   private static RootedOzoneFileSystem ofs;
   private static ObjectStore objectStore;
-  private static BasicRootedOzoneClientAdapterImpl adapter;
+  private static BasicRootedOzoneClientAdapterImpl impl;
   private static Trash trash;
 
   private static String volumeName;
@@ -147,7 +147,7 @@ public class TestRootedOzoneFileSystem {
     fs = FileSystem.get(conf);
     trash = new Trash(conf);
     ofs = (RootedOzoneFileSystem) fs;
-    adapter = (BasicRootedOzoneClientAdapterImpl) ofs.getAdapter();
+    impl = ofs.getImpl();
   }
 
   @AfterClass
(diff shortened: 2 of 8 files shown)


smengcl:
Note the previous 5 commits are directly from the previous PR. The last commit 37a441426a013a84ec1abcaf42d84e2196f61c13 is to address the new `getTrashRoots` impl.

I wanted to use proxy directly, only to find that I would need to build what ObjectStore has already done for me when listing volumes. So I kept ObjectStore as class global variable.

elek:
Thanks for the separation @smengcl, I think it's easier to discuss. 

I am not sure it's the same patch what I already commented or not. I had a comment in the previous PR, where the discussion is stopped:

>> I'm in favor of A. I'll attempt to remove the usage of OzoneClientAdapter in OFS altogether then.

> I am fine with that approach but let me add some comments to the latest patch.

>  The naming of BasicRootedOzoneFileSystem and BasicRootedOzoneFileSystemImpl is misleading. Usually the Impl postfix is used when the class implemented a well known interface. There is no such interface here. (It's more like the delegation design pattern not an implementation)

>  As a test: Can you please explain what are the differences between the two classes and the responsibilities?

>  If not, we don't need two classes. Just remove the Impl and remove the dedicated methods and directly call the proxy from the original methods of BasicRootedOzoneFileSystem.

>  Wouldn't it be more simple?

This patch seems to use `BasicRootedOzoneClientAdapterImpl` which is not an `Impl` (not an interface). Do you need a client adapter here (in the old code we need  an interface and an implementation for classpath separation but here this separation is removed.)? If yes, do you need `Impl` in the name (it doesn't implement anything)?

smengcl:
> Thanks for the separation @smengcl, I think it's easier to discuss.
> 
> I am not sure it's the same patch what I already commented or not. I had a comment in the previous PR, where the discussion is stopped:
> 
> > > I'm in favor of A. I'll attempt to remove the usage of OzoneClientAdapter in OFS altogether then.
> 
> > I am fine with that approach but let me add some comments to the latest patch.
> 
> > The naming of BasicRootedOzoneFileSystem and BasicRootedOzoneFileSystemImpl is misleading. Usually the Impl postfix is used when the class implemented a well known interface. There is no such interface here. (It's more like the delegation design pattern not an implementation)
> 
> > As a test: Can you please explain what are the differences between the two classes and the responsibilities?
> 
> > If not, we don't need two classes. Just remove the Impl and remove the dedicated methods and directly call the proxy from the original methods of BasicRootedOzoneFileSystem.
> 
> > Wouldn't it be more simple?
> 
> This patch seems to use `BasicRootedOzoneClientAdapterImpl` which is not an `Impl` (not an interface). Do you need a client adapter here (in the old code we need an interface and an implementation for classpath separation but here this separation is removed.)? If yes, do you need `Impl` in the name (it doesn't implement anything)?

I'm renaming `BasicRootedOzoneClientAdapterImpl` to `BasicRootedOzoneFileSystemHelper`. This way the name should make more sense?

elek:
> I'm thinking of renaming BasicRootedOzoneClientAdapterImpl to BasicRootedOzoneFileSystemHelper. This way the class name should make more sense?

Definitely better, IMHO. It's not clear why we need to move out some functions to a helper class (`"Can you please explain what are the differences between the two classes and the responsibilities?"`), but I can live with it, just to merge the patch earlier.

> The class is dealing with OzoneFSStorageStatistics, but not used anywhere. The o3fs counterpart is OzoneClientAdapterImpl, which is used in OzoneFileSystem.

This is only about the statistics. Independent to what type of classes do you have: 

 You need a `OzoneFSStorageStatistics storageStatistics` which is updated for each of the operations, but only for Hadoop3 (!!!).

For the old-school `o3fs` it's done by subclasses:

 * Default implementation is hadoop2 compatibility as `incdementCounter` does nothing
 * non-Basic implementation increments a real counter.

As far as I see we already have this logic for `ofs` as the imlementation of `RootedOzoneFileSystem` in `hadoop2` and `hadoop3` projects are different. (Later one update the statistics.)

If your helper class can be used from both project without problem, you don't need to create two helper classses for the two use-cases.

PR #1370, opened 2020-09-01
HDDS-4176. Fix failed UT: test2WayCommitForTimeoutException

runzhiwang:
**What's the problem ?**
image

**What's the reason of the failed ut ?**

When cluster create 9 nodes, the following code try to shutdown the follower of the pipeline. Actually, the pipeline only exist in 3 nodes, but the following code check 9 nodes, so the pipeline can not found in the other 6 nodes. 
```
    for (HddsDatanodeService dn : cluster.getHddsDatanodes()) {
      // shutdown the ratis follower
      if (ContainerTestHelper.isRatisFollower(dn, pipeline)) {
        cluster.shutdownHddsDatanode(dn.getDatanodeDetails());
        break;
      }
    }
```

https://issues.apache.org/jira/browse/HDDS-4176

## How was this patch tested?

Existed ut.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestWatchForCommit.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestWatchForCommit.java
@@ -298,10 +298,10 @@ public class TestWatchForCommit {
             xceiverClient.getPipeline()));
     reply.getResponse().get();
     Assert.assertEquals(3, ratisClient.getCommitInfoMap().size());
-    List<DatanodeDetails> datanodeDetails = pipeline.getNodes();
+    List<DatanodeDetails> nodesInPipeline = pipeline.getNodes();
     for (HddsDatanodeService dn : cluster.getHddsDatanodes()) {
       // shutdown the ratis follower
-      if (datanodeDetails.contains(dn.getDatanodeDetails())
+      if (nodesInPipeline.contains(dn.getDatanodeDetails())
           && ContainerTestHelper.isRatisFollower(dn, pipeline)) {
         cluster.shutdownHddsDatanode(dn.getDatanodeDetails());
         break;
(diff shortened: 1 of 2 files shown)


runzhiwang:

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestWatchForCommit.java
```diff
@@ -297,9 +298,11 @@ public void test2WayCommitForTimeoutException() throws Exception {
             xceiverClient.getPipeline()));
     reply.getResponse().get();
     Assert.assertEquals(3, ratisClient.getCommitInfoMap().size());
+    List<DatanodeDetails> datanodeDetails = pipeline.getNodes();
     for (HddsDatanodeService dn : cluster.getHddsDatanodes()) {
       // shutdown the ratis follower
-      if (ContainerTestHelper.isRatisFollower(dn, pipeline)) {
+      if (datanodeDetails.contains(dn.getDatanodeDetails())
+          && ContainerTestHelper.isRatisFollower(dn, pipeline)) {
```
amaliujia, line 305: This logic looks right to me.

Can you bump up the number of DN to make tests run against higher number DN: https://github.com/apache/hadoop-ozone/blob/34ee8311b0d0a37878fe1fd2e5d8c1b91aa8cc8f/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/Test2WayCommitInRatis.java#L109

Which seems will cover your change?

runzhiwang: @amaliujia Thanks for your review. TestWatchForCommit already startCluster with 9 nodes, maybe it is enough.
https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestWatchForCommit.java#L124

amaliujia: Oops. You are right. I was looking at a wrong file.

adoroszlai (approved):
Thanks @runzhiwang for fixing this long standing intermittent failure.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestWatchForCommit.java
```diff
@@ -297,9 +298,11 @@ public void test2WayCommitForTimeoutException() throws Exception {
             xceiverClient.getPipeline()));
     reply.getResponse().get();
     Assert.assertEquals(3, ratisClient.getCommitInfoMap().size());
+    List<DatanodeDetails> datanodeDetails = pipeline.getNodes();
```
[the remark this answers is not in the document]
runzhiwang: @adoroszlai Thanks for review. I have updated the patch.

adoroszlai:
Thanks @runzhiwang for the fix and @amaliujia for the review.

PR #1371, opened 2020-09-01
HDDS-2922. Balance ratis leader distribution in datanodes

runzhiwang:

**What's the problem ?**

When enable multi-raft, the leader distribution in datanodes is not balance. In my test, there are 72 datanodes, each datanode 
engage in 6 pipelines, so there are 144 pipelines. As the image shows, the leader number of the 4 datanodes is 0, 0, 4, 2, it's not balance. Because ratis leader not only accept client request, but also replicate log to 2 followers, and follower only replicate log from leader, so the leader's load is at least 3 times of follower. So we need to balance leader.

image

**How to improve ?**

With the guidance of @szetszwo , I implement [RATIS-967](https://issues.apache.org/jira/browse/RATIS-967), which not only support priority in leader election, but also support lower priority leader try to yield leadership to higher priority peer when  higher priority peer's log catch up, to address the higher priority leader lose the leadership.

So in ozone
1. assign the suggested leader with higher priority, and 2 followers with lower priority, then we can achieve leader distribution's balance.
2. record the suggested leader in Pipeline, when create new pipeline, choose 3 datanodes,  find pipelines on each datanode, calculate the suggested leader count on each datanode, then choose the datanode which has the minum suggested leader count as the leader.
3. to avoid we lose the suggested leader of pipeline in SCM when SCM restart, we store the suggested leader in pipeline table.

As the following image shows, there are 72 datanodes, each datanode engage in 6 pipelines, so there are 144 pipelines.
The leader count of each datanode is 2, there is no exception, we achieve the leader distribution's balance. 

image

https://issues.apache.org/jira/browse/HDDS-2922

## How was this patch tested?

add new ut.


runzhiwang:
@bshashikant @lokeshj1703 @mukul1987 @elek @xiaoyuyao Could you help review this patch ? Thank you very much.

amaliujia:
Thanks @runzhiwang. This is an awesome work! I will also try to help review this PR.

GlenGeng (commented):
Thanks for this fantastic feature! Just have some inline comments.

We can handle the suggestLeaderCount in a easier and accurate way by leveraging pipeline report.

runzhiwang:
@GlenGeng Thanks for review, I have updated the patch.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
```diff
@@ -711,10 +712,20 @@ private long calculatePipelineBytesWritten(HddsProtos.PipelineID pipelineID) {
 
   @Override
   public void addGroup(HddsProtos.PipelineID pipelineId,
-      Collection<DatanodeDetails> peers) throws IOException {
+      List<DatanodeDetails> peers) throws IOException {
+    List<Integer> priorityList =
```
xiaoyuyao, line 716: NIT: can we have a static default priority list here?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
```diff
@@ -98,8 +105,65 @@ private boolean exceedPipelineNumberLimit(ReplicationFactor factor) {
...
+    Map<DatanodeDetails, Integer> suggestedLeaderCount = new HashMap<>();
+    for (DatanodeDetails dn : dns) {
+      suggestedLeaderCount.put(dn, 0);
+
+      Set<PipelineID> pipelineIDSet = getNodeManager().getPipelines(dn);
+      for (PipelineID pipelineID : pipelineIDSet) {
+        try {
+          Pipeline pipeline = getPipelineStateManager().getPipeline(pipelineID);
+          if (!pipeline.isClosed()
+              && dn.getUuid().equals(pipeline.getSuggestedLeaderId())) {
```
xiaoyuyao, line 119: should we use getLeaderId() instead of getSuggestedLeaderId() here to reflect the actual leader count?

xiaoyuyao: And the method name can be changed to getLeaderCount() so that the suggest leader is determined by the actual leader count.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
```diff
@@ -98,8 +105,65 @@ private boolean exceedPipelineNumberLimit(ReplicationFactor factor) {
...
+    return suggestedLeader;
+  }
+
+  private List<Integer> getPriorityList(
+      List<DatanodeDetails> dns, DatanodeDetails suggestedLeader) {
+    List<Integer> priorityList = new ArrayList<>();
+
+    for (DatanodeDetails dn : dns) {
+      if (dn.getUuid().equals(suggestedLeader.getUuid())) {
+        priorityList.add(HIGH_PRIORITY);
```
xiaoyuyao, line 155: Does Ratis have guarantee to have the highpriority node act as leader?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestMiniOzoneCluster.java
```diff
@@ -102,7 +102,7 @@ public static void afterClass() {
     FileUtils.deleteQuietly(READ_TMP);
   }
 
-  @Test(timeout = 30000)
+  @Test(timeout = 60000)
```
xiaoyuyao, line 105: Does the new logic cause more timeout without increasing this?

xiaoyuyao (commented):
Thanks @runzhiwang  for adding this useful feature. It looks good to me overall. A few comments added inline.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
runzhiwang: Updated

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
runzhiwang: Yes, [RATIS-967](https://issues.apache.org/jira/browse/RATIS-967) guarantee the highpriority node act as leader

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestMiniOzoneCluster.java
runzhiwang: Yes, the highpriority node act as leader will increase the time of leader election,  but almost the increased time is within 10 seconds.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
runzhiwang: [RATIS-967](https://issues.apache.org/jira/browse/RATIS-967) guarantee the highpriority node act as leader when create pipeline. In a long running cluster, the leader maybe crash, then some follower will take the leadership, but when the old leader restart and catch up with current leader's log, the old leader can grab the leadership again by [RATIS-967](https://issues.apache.org/jira/browse/RATIS-967). 

So let me suppose the following case, there are 3 servers: s1, s2, s3, there are 2 pipelines now,  the first pipeline's leader is s1, the second pipeline's leader is s2, both the 2 leaders are suggested leader with high priority. Then s1 crash, suppose s3 will take the first pipline's leader.  Then s1 restart, but has not grab leadership of the first pipeline. If we use getLeaderId() instead of getSuggestedLeaderId() to reflect the actual leader count, when we create the 3 third pipeline, we find the leader number on s1, s2, s3 is 0, 1, 1, so we will select s1 as the suggest leader,  then s1 grab the leadership of the first pipeline, so the leader number on s1, s2, s3 will be 2, 1, 0, it's not balance.

runzhiwang:
@xiaoyuyao Thanks for review. I have updated the patch.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
xiaoyuyao: bq.  then s1 grab the leadership of the first pipeline by RATIS-967,

Does RATIS-967 always gives up its leader when s1 is back online even the current leader works fine? I think this is more specific to RATIS-967 wrt. how the priority is enforced. Any performance impact on the pipeline of forcing leader to be the original one? 

I'm thinking of instead of forcing leader of pipeline P1, P2, P3 like
S1   S2   S3
P1   P2
              P3

In the case of S1 temporarily down, why don't we keep P1 leader on S3 and create P3 with leader on S1, this gives more flexibility for higher level to choose leader?
S1   S2   S3
       P2
              P1
P3

Another situation I'm thinking of is writers on pipeline with slow leader(e.g., hardware slowness) may not be able to recover by leader change.

runzhiwang: @xiaoyuyao Good point, I also have thought this.

> Any performance impact on the pipeline of forcing leader to be the original one.

If there is performance problem, I can improve forcing leader change within 1 second. I already know how to improve it, but has not implemented it.

> Another situation I'm thinking of is writers on pipeline with slow leader(e.g., hardware slowness) may not be able to recover by leader change.

We can find slow leader by some metric, decrease the priority of the slow leader, select one faster datanode and increase it's priority, so the faster datanode will grab the leadership from the slow leader.


> In the case of S1 temporarily down, why don't we keep P1 leader on S3 and create P3 with leader on S1, this gives more flexibility for higher level to choose leader?

I want the cluster leader distribution as we planned, if the plan is not appropriate, we can adjust the plan by change priority.

If the leader distribution totally depends on hardware rather than plan, we maybe lost control of the leader distribution. Because the leaderId in scm was reported by datanode, it maybe a delayed leaderId. For example, datanode report:

S1  S2  S3
P1 P2

then P1's leader transfer to S3, but SCM has not received this report, SCM allocate P3's leader to S3, then

S1  S2  S3
     P2   P1
            P3.

It's not balance now.

xiaoyuyao: Good point, planed leader distribution should work with RATIS-967. How do we scale with this. Plan weight for each of node as a leader when the cluster has thousands of nodes can be difficult.

runzhiwang: > Plan weight for each of node as a leader when the cluster has thousands of nodes can be difficult.

If each node has similar hardware, i.e. CPU, memory, we just plan weight as now, assign each node with same leader number, it is cheap and reasonable.

I think the only case we need to consider is that some nodes' hardware is weaker than other nodes' obviously.  I think the weeker datanodes should engage in less pipeline than the stronger datanodes,  but ozone does not support this now.
If we can support this, the maxum leader number of each datanode should be less or equal to ((1/3) * the pipeline number it engaged in),  and we select the datanode as the leader which has lowest value of (leader number / pipeline number it engaged in) in 3 datanodes,  this is also cheap.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
```diff
@@ -98,8 +105,65 @@ private boolean exceedPipelineNumberLimit(ReplicationFactor factor) {
...
+              pipelineID, e);
+        }
+      }
+    }
+
+    return suggestedLeaderCount;
+  }
+
+  private DatanodeDetails getSuggestedLeader(List<DatanodeDetails> dns) {
+    Map<DatanodeDetails, Integer> suggestedLeaderCount =
```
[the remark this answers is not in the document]
runzhiwang: @bshashikant I agree. @xiaoyuyao What do you think of this suggestion?

runzhiwang:
@xiaoyuyao @bshashikant  I have updated the patch. Could you help review it again ? Thank you very much.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/protocol/commands/CreatePipelineCommand.java
```diff
@@ -48,16 +52,37 @@ public CreatePipelineCommand(final PipelineID pipelineID,
     this.factor = factor;
     this.type = type;
     this.nodelist = datanodeList;
+    if (datanodeList.size() ==
+        XceiverServerRatis.DEFAULT_PRIORITY_LIST.size()) {
```
GlenGeng, line 56: just forward to the Ctor with `priorityList `.

```
this(pipelineID, factor, type, datanodeList,
    new ArrayList<>(Collections.nCopies(datanodeList.size(), 0))
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManager.java
```diff
@@ -59,7 +59,8 @@ void addContainerToPipeline(PipelineID pipelineId, ContainerID containerID)
     pipelineStateMap.addContainerToPipeline(pipelineId, containerID);
   }
 
-  Pipeline getPipeline(PipelineID pipelineID) throws PipelineNotFoundException {
+  public Pipeline getPipeline(PipelineID pipelineID)
```
GlenGeng, line 62: revert unnecessary change.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
```diff
@@ -98,8 +115,28 @@ private boolean exceedPipelineNumberLimit(ReplicationFactor factor) {
...
+  @VisibleForTesting
+  public LeaderChoosePolicy getLeaderChoosePolicy() {
+    return leaderChoosePolicy;
+  }
+  private List<Integer> getPriorityList(
+      List<DatanodeDetails> dns, DatanodeDetails suggestedLeader) {
+    List<Integer> priorityList = new ArrayList<>();
+
+    for (DatanodeDetails dn : dns) {
+      if (dn.getUuid().equals(suggestedLeader.getUuid())) {
```
GlenGeng, line 127: why not use `DatanodeDetails.equals()` ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
```diff
@@ -121,18 +158,24 @@ public Pipeline create(ReplicationFactor factor) throws IOException {
       throw new IllegalStateException("Unknown factor: " + factor.name());
     }
 
+    DatanodeDetails suggestedLeader = leaderChoosePolicy.chooseLeader(
```
GlenGeng, line 161: make node manager and pipeline manager be member of leaderChoosePolicy, so that leaderChoosePolicy can has it own state, which will make future extension easier.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -619,9 +619,9 @@ private void finalizePipeline(PipelineID pipelineId) throws IOException {
    * @throws IOException
    */
   protected void destroyPipeline(Pipeline pipeline) throws IOException {
-    pipelineFactory.close(pipeline.getType(), pipeline);
     // remove the pipeline from the pipeline manager
     removePipeline(pipeline.getId());
+    pipelineFactory.close(pipeline.getType(), pipeline);
```
GlenGeng, line 624: why need this change ?  The sequence of sending SCMCommand and removing state may affect SCM HA.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/protocol/commands/CreatePipelineCommand.java
```diff
@@ -39,6 +42,7 @@
   private final ReplicationFactor factor;
   private final ReplicationType type;
   private final List<DatanodeDetails> nodelist;
+  private final List<Integer> priorityList;
```
GlenGeng, line 45: I consider that, should move `RatisPipelineProvider.getPriorityList()`, `HIGH_PRIORITY`, `LOW_PRIORITY` here, and replace priorityList in Ctor param as suggestedLeader.
We can minimize the existence of `priorityList`.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManager.java
runzhiwang: This is necessary, otherwise we can not call `getPipeline` in `MinLeaderCountChoosePolicy`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/RatisPipelineProvider.java
runzhiwang: LeaderChoosePolicy is an interface, define member in interface is not common.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
```diff
@@ -115,6 +116,9 @@
   private static final Logger LOG = LoggerFactory
       .getLogger(XceiverServerRatis.class);
   private static final AtomicLong CALL_ID_COUNTER = new AtomicLong();
+  public static final List<Integer> DEFAULT_PRIORITY_LIST =
```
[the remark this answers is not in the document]
runzhiwang: why always create a new list ? The size of DEFAULT_PRIORITY_LIST is 3, most case the size of datanode list is 3, their size are equal.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/protocol/commands/CreatePipelineCommand.java
runzhiwang: always new is a waste.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
runzhiwang: If did not change,  the order maybe: pipelineFactory.close(pipeline.getType(), pipeline) -> datanode close pipeline -> scm create new pipeline -> removePipeline(pipeline.getId()), then leader distribution will not be balance.
For example, on server S1, S3, S3, there are 3 pipelines: P1, P2, P3 with leader: S1, S2, S3.
S1 .. S2 .. S3
P1 .. P2 .. P3

If close P3, and order maybe: pipelineFactory.close(pipeline.getType(), P3) -> datanode close P3 -> scm create new pipeline P4 -> removePipeline(P3). When create new pipeline P4, because P3 has not been removed, P4 will choose S1 as the leader, then S1 has 2 leaders, but S3 has no leader.
S1 .. S2 .. S3
P1 .. P2
P4

bshashikant:
Can we also make the policy configurable? Also, one policy should also be defined for no priority at all incase, this turns out to be a performance killer.

runzhiwang:
@GlenGeng @bshashikant  Thanks for review. I have updated the patch.

`Can we also make the policy configurable? Also, one policy should also be defined for no priority at all incase, this turns out to be a performance killer.`

policy can be configured by `ozone.scm.pipeline.leader-choose.policy`. And define a policy for no priority named `RandomLeaderChoosePolicy`

bshashikant:
> @GlenGeng @bshashikant Thanks for review. I have updated the patch.
> 
> > Can we also make the policy configurable? Also, one policy should also be defined for no priority at all incase, this turns out to be a performance killer.
> 
> policy can be configured by `ozone.scm.pipeline.leader-choose.policy`. And define a policy for no priority named `RandomLeaderChoosePolicy`

@runzhiwang , please correct me if i am wrong. RandomLeaderChoosePolicy still chooses a datanode randomly and this is suggested to Ratis while creating the pipeline.
With NO_PRIORITY, i meant that, we should not have any recommendation for a leader at all (as what is currently). Usually in such cases, whoever starts the ratis leader election first initially, becomes the leader.

runzhiwang:
@bshashikant Thanks for suggestions. Actually, RandomLeaderChoosePolicy does not choose datanode, it return null in [chooseLeader](https://github.com/apache/hadoop-ozone/pull/1371/files#diff-b180475d031c8ef84f96e7d623f11506R40), then all the datanodes are assigned the [same priority](https://github.com/apache/hadoop-ozone/pull/1371/files#diff-a9f8deba9c990cb9b148008785c8cff2R60), as what is currently.  The name of RandomLeaderChoosePolicy seems confused, sorry for the misleading, do you have better name?

bshashikant:
> @bshashikant Thanks for suggestions. Actually, RandomLeaderChoosePolicy does not choose datanode, it return null in [chooseLeader](https://github.com/apache/hadoop-ozone/blob/f573b4a8b45f4463f0bcb95971a45789bda91d5c/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/leader/choose/algorithms/RandomLeaderChoosePolicy.java#L40), then all the datanodes are assigned the [same priority](https://github.com/apache/hadoop-ozone/blob/f573b4a8b45f4463f0bcb95971a45789bda91d5c/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/protocol/commands/CreatePipelineCommand.java#L60), as what is currently. The name of RandomLeaderChoosePolicy seems confused, sorry for the misleading, do you have better name?

i guess , this can be named as "DefaultLeaderChoosePolicy" and this should be made the default , until and unless we measure the performance with the minimumLeader election count policy and see the results.

runzhiwang:
> > @bshashikant Thanks for suggestions. Actually, RandomLeaderChoosePolicy does not choose datanode, it return null in [chooseLeader](https://github.com/apache/hadoop-ozone/blob/f573b4a8b45f4463f0bcb95971a45789bda91d5c/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/leader/choose/algorithms/RandomLeaderChoosePolicy.java#L40), then all the datanodes are assigned the [same priority](https://github.com/apache/hadoop-ozone/blob/f573b4a8b45f4463f0bcb95971a45789bda91d5c/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/protocol/commands/CreatePipelineCommand.java#L60), as what is currently. The name of RandomLeaderChoosePolicy seems confused, sorry for the misleading, do you have better name?
> 
> i guess , this can be named as "DefaultLeaderChoosePolicy" and this should be made the default , until and unless we measure the performance with the minimumLeader election count policy and see the results. What do you think?

@bshashikant I agree. I have updated the patch.

GlenGeng:
+1
Thanks for the work! LGTM

bshashikant (approved):

runzhiwang:
@xiaoyuyao Could you help merge this patch ? Thanks a lot.

ChenSammi:
LGTM + 1. 

Thanks @runzhiwang for the contribution and @bshashikant @xiaoyuyao @GlenGeng for the review.

runzhiwang:
@ChenSammi Thanks for merging it, @bshashikant @xiaoyuyao @GlenGeng Thanks for review.

PR #1372, opened 2020-09-01
HDDS-4166. Documentation index page redirects to the wrong address

cku328:

The server for the edge version of the documentation doesn't appear to use `index.html` as the default index page, but instead displays a list of directories. 

If the file name of the page is `index.html`, it doesn't add `index.html` to the `Permalink` parameters, which seems to be hugo's style. I found similar cases on the internet. (ref. [https://discourse.gohugo.io/t/ugly-urls-with-page-resources/10037](https://discourse.gohugo.io/t/ugly-urls-with-page-resources/10037))

The easiest solution is to check the permalink suffix and if it's not `.html` then print `index.html` at the end.

https://issues.apache.org/jira/browse/HDDS-4166

## How was this patch tested?

Tested locally using `hugo serve`

擷取

### DIFF
--- a/hadoop-hdds/docs/content/interface/O3fs.zh.md
+++ b/hadoop-hdds/docs/content/interface/O3fs.zh.md
@@ -29,7 +29,7 @@ summary: Hadoop 文件系统兼容使得任何使用类 HDFS 接口的应用无�
 
 Hadoop 的文件系统接口兼容可以让任意像 Ozone 这样的存储后端轻松地整合进 Hadoop 生态系统，Ozone 文件系统就是一个兼容 Hadoop 的文件系统。
 目前ozone支持两种协议: o3fs和ofs。两者最大的区别是o3fs只支持在单个bucket上操作，而ofs则支持跨所有volume和bucket的操作。关于两者在操作
-上的具体区别可以参考ofs.md中的"Differences from existing o3fs"。
+上的具体区别可以参考[OFS(英文页面)]({{< relref path="interface/ofs.md" lang="en" >}})中的"Differences from o3fs"。
 
 ## o3fs的配置及使用
 
@@ -115,7 +115,7 @@ hdfs dfs -ls o3fs://bucket.volume.om-host.example.com:6789/key
 注意：在这种情况下，`ozone.om.address` 配置中只有端口号会被用到，主机名是被忽略的。
 
 ## ofs的配置及使用
-这只是一个通用的介绍。了解更详细的用法，可以请参考ofs.md。
+这只是一个通用的介绍。了解更详细的用法，可以请参考[OFS(英文页面)]({{< relref path="interface/ofs.md" lang="en" >}})。
 
 请在 core-site.xml 中添加以下条目：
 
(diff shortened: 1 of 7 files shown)


adoroszlai (commented):
Thanks @cku328 for fixing this.

All _Home_ links in breadcrumbs (eg. in _Home > Features > Topology awareness_) are also affected, ie. `index.html` might need to be added to:

https://github.com/apache/hadoop-ozone/blob/13fe31b4927660bc2534656c6a7048b45d3c8051/hadoop-hdds/docs/themes/ozonedoc/layouts/_default/single.html#L33

FYI this can be reproduced locally by opening compiled docs from disk after `mvn package` without `hugo serve`, since there is no webserver in this case that would serve `index.html` as directory listing.

cku328:
Thanks @adoroszlai for your review and comments.
I re-checked the hyperlinks in the document related to "index.html".

I will update this patch to include the following changes:
1. Fixed ozone logo image not displaying properly and added a hyperlink to the index page.
2. Added "index.html" to the "Home" hyperlink in the breadcrumb.
3. Added "index.html" to the "Overview" hyperlink in the sidebar.

adoroszlai (approved):
Thanks @cku328 for updating the patch.  I couldn't find any remaining non-http links ending in `/`.

cku328:
Hi! @adoroszlai 
I found some broken hyperlinks and updated the patch to fix it.

It fixed the following items:
1. `Storage Container Manager` hyperlink in `/zh/concept/overview.html` page is not available.
2. `kerberos keytab` hyperlink in `/zh/security/secureozone.html`page is not available.
3. Replacing `ofs.md` with a hyperlink of OFS in `/zh/interface/o3fs.html`.

hadoop-hdds/docs/content/interface/O3fs.zh.md
```diff
@@ -115,7 +115,7 @@ hdfs dfs -ls o3fs://bucket.volume.om-host.example.com:6789/key
 注意：在这种情况下，`ozone.om.address` 配置中只有端口号会被用到，主机名是被忽略的。
 
 ## ofs的配置及使用
-这只是一个通用的介绍。了解更详细的用法，可以请参考ofs.md。
+这只是一个通用的介绍。了解更详细的用法，可以请参考[OFS(英文页面)]({{< relref path="interface/ofs.md" lang="en" >}})。
```
cxorm, line 118: nit
```suggestion
这只是一个通用的介绍。了解更详细的用法，请参考[OFS(英文页面)]({{< relref path="interface/ofs.md" lang="en" >}})。
```

cku328:
Thanks @cxorm for the review.
I've updated the patch based on your review.

hadoop-hdds/docs/content/interface/O3fs.zh.md
cku328: done

adoroszlai:
Thanks @cku328 for the improvement and @cxorm for the review.  I have confirmed that @cxorm's wording suggestion is applied, so merging now.

PR #1376, opened 2020-09-02
HDDS-3762. Intermittent failure in TestDeleteWithSlowFollower

adoroszlai:

Intermittent failure in `testDeleteKeyWithSlowFollower` seems to be caused by:

* `DeleteBlocksCommandHandler` increments `invocationCount` near the beginning of `handle()`, and only updates `deleteTransactionId` later
* `TestDeleteWithSlowFollower` waits for `invocationCount >= 1`, then asserts `deleteTransactionId` also increased

The test is fixed by changing the order in the handler: only increment `invocationCount` at the end, when `deleteTransactionId` is already updated.

I think `invocationCount` should be updated together with `totalTime` to provide (slightly more) correct average run time (`getAverageRunTime`).

https://issues.apache.org/jira/browse/HDDS-3762

## How was this patch tested?

Passed 50x:
https://github.com/adoroszlai/hadoop-ozone/runs/1059570465#step:5:4

Regular CI:
https://github.com/adoroszlai/hadoop-ozone/runs/1059560880


bshashikant (approved):

bshashikant:
Thanks @adoroszlai for the contribution.

adoroszlai:
Thanks @bshashikant for reviewing and committing it.

PR #1381, opened 2020-09-03
HDDS-4201. Improve the performance of OmKeyLocationInfoGroup

runzhiwang:

   ` locations.stream().collect(Collectors.groupingBy(OmKeyLocationInfo::getCreateVersion))` cost 2-3ms even though the locations size is zero.

https://issues.apache.org/jira/browse/HDDS-4201

## How was this patch tested?

no need to add test.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyLocationInfoGroup.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyLocationInfoGroup.java
@@ -36,8 +36,13 @@ public class OmKeyLocationInfoGroup {
   public OmKeyLocationInfoGroup(long version,
                                 List<OmKeyLocationInfo> locations) {
     this.version = version;
-    this.locationVersionMap = locations.stream()
-        .collect(Collectors.groupingBy(OmKeyLocationInfo::getCreateVersion));
+    locationVersionMap = new HashMap<>();
+    for (OmKeyLocationInfo info : locations) {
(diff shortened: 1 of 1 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmKeyLocationInfoGroup.java
```diff
@@ -36,8 +36,13 @@
...
                                 List<OmKeyLocationInfo> locations) {
     this.version = version;
-    this.locationVersionMap = locations.stream()
-        .collect(Collectors.groupingBy(OmKeyLocationInfo::getCreateVersion));
+    locationVersionMap = new HashMap<>();
+    for (OmKeyLocationInfo info : locations) {
+      if (!locationVersionMap.containsKey(info.getCreateVersion())) {
+        locationVersionMap.put(info.getCreateVersion(), new ArrayList<>());
+      }
+      locationVersionMap.get(info.getCreateVersion()).add(info);
```
adoroszlai, line 44: I think we can avoid two map lookups (and long autoboxing) by using `computeIfAbsent`:

```suggestion
      locationVersionMap
          .computeIfAbsent(info.getCreateVersion(), v -> new ArrayList<>())
          .add(info);
```

adoroszlai (commented):
Thanks @runzhiwang for working on this.

runzhiwang:
@adoroszlai Thanks for review. I have updated the patch.

adoroszlai (approved):

PR #1382, opened 2020-09-03
HDDS-4202. Upgrade ratis to 1.1.0-ea949f1-SNAPSHOT

runzhiwang:

1. Upgrade ratis to 1.1.0-ea949f1-SNAPSHOT to support the feature of HDDS-2922. Balance ratis leader distribution in datanodes.
2. We need to add `metrics-core` dependency in hadoop-hdds/common/pom.xml in ozone, because https://github.com/apache/incubator-ratis/commit/aab414f2af73809168657beb17419be87b0aa259 import MetricRegistry in ratis-client, if ozone does not add this dependency, hdfs o3fs will report NoClassDef

image

https://issues.apache.org/jira/browse/HDDS-4202

## How was this patch tested?

no need add test

### DIFF
--- a/hadoop-hdds/common/pom.xml
+++ b/hadoop-hdds/common/pom.xml
@@ -87,6 +87,11 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
     </dependency>
 
     <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-core</artifactId>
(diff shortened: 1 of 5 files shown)


elek (approved):
+1 Thanks the patch @runzhiwang 

Build is green, patch seems to be reasonable.

@mukul1987 is confirmed for me that we don't have big changes on Ratis side, risk is low.

PR #1384, opened 2020-09-03
HDDS-4193. Range used by S3 MultipartUpload copy-from-source should be inclusive.

elek:

S3 API provides a feature to copy a specific range from an existing key.

Based on the documentation, this range definitions is inclusive:

https://docs.aws.amazon.com/cli/latest/reference/s3api/upload-part-copy.html

> 
>     -copy-source-range (string)
> 
>     The range of bytes to copy from the source object. The range value must use the form bytes=first-last, where the first and last are the zero-based byte offsets to copy. For example, bytes=0-9 indicates that you want to copy the first 10 bytes of the source. You can copy a range only if the source object is greater than 5 MB.
> 

But as it's visible from our robot test, in our case we use exclusive range:

```
upload-part-copy ... --copy-source-range bytes=0-10485758
upload-part-copy ... --copy-source-range bytes=10485758-10485760
```

Based on this AWS documentation it will return with a (10485758 + 1) + 3 bytes long key, which is impossible if our original source key is just 10485760.

I think the right usage to get the original key is the following:

```shell
upload-part-copy ... --copy-source-range bytes=0-10485757
upload-part-copy ... --copy-source-range bytes=10485758-10485759
```

(Note, this bug is found with the script in HDDS-4194, which showed that AWS S3 is working in different way).

https://issues.apache.org/jira/browse/HDDS-4193

## How was this patch tested?

Unit test and acceptance tests are updated.

The truth of the specification is tested with https://issues.apache.org/jira/browse/HDDS-4194

### DIFF
--- a/hadoop-ozone/dist/src/main/smoketest/s3/MultipartUpload.robot
+++ b/hadoop-ozone/dist/src/main/smoketest/s3/MultipartUpload.robot
@@ -258,12 +258,12 @@ Test Multipart Upload Put With Copy and range
                         Should contain           ${result}    ${BUCKET}
                         Should contain           ${result}    UploadId
 
-    ${result} =         Execute AWSS3APICli      upload-part-copy --bucket ${BUCKET} --key copyrange/destination --upload-id ${uploadID} --part-number 1 --copy-source ${BUCKET}/copyrange/source --copy-source-range bytes=0-10485758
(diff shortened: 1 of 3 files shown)


ChenSammi:
+1.  LGTM.  Thanks @elek.

elek:
Thanks the review @ChenSammi. I am merging it now.

PR #1385, opened 2020-09-03
HDDS-3947: Sort DNs for client when the key is a file for #getFileStatus #listStatus APIs

rakeshadr:

Similar to OzoneManagerFS#lookupFile(OmKeyArgs args, String clientAddress) interface, "clientAddress" has been passed as an argument to OzoneManagerFS#listStatus(), OzoneManagerFS#getFileStatus() APIs.

https://issues.apache.org/jira/browse/HDDS-3947

## How was this patch tested?

Used existing TestKeyManagerImpl unit test cases.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -1732,17 +1732,36 @@ public class KeyManagerImpl implements KeyManager {
    * @param args Key args
    * @throws OMException if file does not exist
    *                     if bucket does not exist
+   *                     if volume does not exist
    * @throws IOException if there is error in the db
    *                     invalid arguments
    */
   public OzoneFileStatus getFileStatus(OmKeyArgs args) throws IOException {
+    Preconditions.checkNotNull(args, "Key args can not be null");
+    return getFileStatus(args, null);
+  }
+
+  /**
+   * OzoneFS api to get file status for an entry.
+   *
+   * @param args Key args
+   * @param clientAddress a hint to key manager, order the datanode in returned
+   *                      pipeline by distance between client and datanode.
+   * @throws OMException if file does not exist
+   *                     if bucket does not exist
+   *                     if volume does not exist
+   * @throws IOException if there is error in the db
+   *                     invalid arguments
+   */
+  public OzoneFileStatus getFileStatus(OmKeyArgs args, String clientAddress)
+          throws IOException {
     Preconditions.checkNotNull(args, "Key args can not be null");
     String volumeName = args.getVolumeName();
     String bucketName = args.getBucketName();
     String keyName = args.getKeyName();
 
     return getOzoneFileStatus(volumeName, bucketName, keyName,
-            args.getRefreshPipeline(), false, null);
+            args.getRefreshPipeline(), args.getSortDatanodes(), clientAddress);
   }
 
   private OzoneFileStatus getOzoneFileStatus(String volumeName,
@@ -1783,7 +1802,9 @@ public class KeyManagerImpl implements KeyManager {
 
       // if the key is a file then do refresh pipeline info in OM by asking SCM
       if (fileKeyInfo != null) {
-        refreshPipeline(fileKeyInfo);
+        if (refreshPipeline) {
+          refreshPipeline(fileKeyInfo);
+        }
         if (sortDatanodes) {
           sortDatanodeInPipeline(fileKeyInfo, clientAddress);
         }
@@ -2011,7 +2032,8 @@ public class KeyManagerImpl implements KeyManager {
    * @return list of file status
    */
   public List<OzoneFileStatus> listStatus(OmKeyArgs args, boolean recursive,
-      String startKey, long numEntries) throws IOException {
+      String startKey, long numEntries, String clientAddress)
+          throws IOException {
     Preconditions.checkNotNull(args, "Key args can not be null");
 
     List<OzoneFileStatus> fileStatusList = new ArrayList<>();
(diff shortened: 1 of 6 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/fs/OzoneManagerFS.java
```diff
@@ -49,6 +69,21 @@ OpenKeySession createFile(OmKeyArgs args, boolean isOverWrite,
...
+   *                      exists its status is included in the final list.
+   * @param numEntries    Number of entries to list from the start key
+   * @param clientAddress a hint to key manager, order the datanode in returned
+   *                      pipeline by distance between client and datanode.
+   * @return list of file status
+   * @throws IOException if file or bucket or volume does not exist
+   */
   List<OzoneFileStatus> listStatus(OmKeyArgs keyArgs, boolean recursive,
-      String startKey, long numEntries) throws IOException;
+      String startKey, long numEntries, String clientAddress)
```
adoroszlai, line 87: In my [comment](https://github.com/apache/hadoop-ozone/pull/1192#discussion_r481025587) on the previous PR:

> Should we also keep the two original methods?

I meant both `getFileStatus` and `listStatus`.  Sorry if it was unclear.

adoroszlai (commented):
Thanks @rakeshadr for updating the patch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1783,7 +1802,9 @@ private OzoneFileStatus getOzoneFileStatus(String volumeName,
 
       // if the key is a file then do refresh pipeline info in OM by asking SCM
       if (fileKeyInfo != null) {
-        refreshPipeline(fileKeyInfo);
+        if (refreshPipeline) {
```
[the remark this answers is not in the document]
rakeshadr: Thanks @adoroszlai again for the review help.

Yes, you are correct. Test passed locally without the `if (refreshPipeline)` condition.

Point-1) That means, the caller should call`refreshPipeline(OmKeyInfo value)` method without the `refreshPipeline` condition.

Point-2) Also, I'm planning to move the [refreshPipeline](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L2131) call outside BUCKET_LOCK and will be moving to [here](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L2147). Now, its a duplicate call added as part of HDDS-3658.

Does these two changes make sense to you?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/fs/OzoneManagerFS.java
rakeshadr: Sure, I will make the changes in my next commit.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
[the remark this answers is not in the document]
adoroszlai: Yes, both points make sense.  Thanks for noticing the duplicate call.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/fs/OzoneManagerFS.java
rakeshadr: Sorry for re-opening the discussion on this comment. On a second thought `listStatus` is only used in [one place in the source code](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java#L2910) and there won't be any other caller for `listStatus() without ClientAddress`. So, how about deleting the original method and can add it later if there is any need ?

adoroszlai: > listStatus is only used in one place in the source code

It is also used by several tests, which all pass `null` as address.

rakeshadr: OK, addressed in my new commit

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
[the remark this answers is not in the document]
rakeshadr: OK, Done!

rakeshadr:
Thanks @adoroszlai for the reviews. Uploaded new commit addressing your comments.

xiaoyuyao (approved):
Thanks @rakeshadr  for working on this. The patch LGTM, +1.

adoroszlai (commented):
Thanks @rakeshadr for updating the patch.

I have one question left: shouldn't `getFileStatus` call from `listStatus` pass `clientAddress`?

https://github.com/apache/hadoop-ozone/blob/088e40d2f79ab0e8c6021523fd4e2b8db23127af/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L2076

rakeshadr:
> Thanks @rakeshadr for updating the patch.
> 
> I have one question left: shouldn't `getFileStatus` call from `listStatus` pass `clientAddress`?
> 
> https://github.com/apache/hadoop-ozone/blob/088e40d2f79ab0e8c6021523fd4e2b8db23127af/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L2076

Good catch. Thanks @adoroszlai. 

Updated new commit. I've added clientAddress to the #getFileStatus call. Also, I've moved #getFileStatus call outside BUCKET_LOCK so that it will get the benefit of doing 'sort' and 'refreshPipeline' outside BUCKET_LOCK.

rakeshadr:
> Thanks @rakeshadr for working on this. The patch LGTM, +1.

Thanks a lot @xiaoyuyao for the reviews!

adoroszlai (approved):
Thanks @rakeshadr for improving this patch several times.

adoroszlai:
Thanks @rakeshadr for the contribution, and @xiaoyuyao for the review.

rakeshadr:
Thanks a lot @adoroszlai for review help and merging it.

PR #1388, opened 2020-09-03
HDDS-4204. upgrade docker environment does not work with KEEP_RUNNING=true

adoroszlai:

Ozone `upgrade` test fails if run with `KEEP_RUNNING=true`.  The variable is applied to both runs (pre- and post-upgrade), but pre-upgrade containers should be stopped regardless, since they will be replaced by the new ones.

Also remove leftover `debug` directory creation.

https://issues.apache.org/jira/browse/HDDS-4204

## How was this patch tested?

Tested locally:

```
$ cd hadoop-ozone/dist/target/ozone-1.1.0-SNAPSHOT/compose/upgrade
$ KEEP_RUNNING=true ./test.sh
...
Stopping upgrade_om_1    ...
...
Readdata :: Smoketest ozone cluster startup                           | PASS |
...
```

Regular CI:
https://github.com/adoroszlai/hadoop-ozone/runs/1067273739

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/upgrade/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/upgrade/test.sh
@@ -26,7 +26,6 @@ export COMPOSE_DIR
 export OZONE_VOLUME
 
 mkdir -p "${OZONE_VOLUME}"/{dn1,dn2,dn3,om,recon,s3g,scm}
-mkdir -p "${OZONE_VOLUME}/debug"
 
 if [[ -n "${OZONE_VOLUME_OWNER}" ]]; then
   current_user=$(whoami)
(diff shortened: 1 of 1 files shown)


hadoop-ozone/dist/src/main/compose/upgrade/test.sh
```diff
@@ -26,7 +26,6 @@ export COMPOSE_DIR
 export OZONE_VOLUME
 
 mkdir -p "${OZONE_VOLUME}"/{dn1,dn2,dn3,om,recon,s3g,scm}
-mkdir -p "${OZONE_VOLUME}/debug"
```
amaliujia, line 29: `/debug` is leftover means it is unused now?

adoroszlai: Yes.  I had added it for a `debug` container while trying to find out the cause of permission problems in GitHub Actions (it was working fine locally).  I [removed the container](https://github.com/apache/hadoop-ozone/pull/1142/commits/9dbbcbb7e1719a5bba94533e49674556d4b1e362) in the final version, but forgot about the directory.

elek (approved):
+1 thanks the fix

adoroszlai:
Thanks @amaliujia for the review, and @elek for reviewing and committing it.

PR #1389, opened 2020-09-03
HDDS-4206. Attempt pipeline creation more frequently in acceptance tests

adoroszlai:

Increase frequency of background pipeline creation from every 2 minutes to every 30 seconds.

Acceptance tests are frequently failing recently due to SCM not being able to exit safe mode for 3 minutes.  Decreasing this period in tests should result in faster startup and less idle wait time.

https://issues.apache.org/jira/browse/HDDS-4206

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/1067654337

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
+++ b/hadoop-ozone/dist/src/main/compose/ozone-csi/docker-config
@@ -20,6 +20,7 @@ OZONE-SITE.XML_ozone.csi.socket=/tmp/csi.sock
 OZONE-SITE.XML_ozone.om.address=om
 OZONE-SITE.XML_ozone.om.http-address=om:9874
 OZONE-SITE.XML_ozone.scm.container.size=1GB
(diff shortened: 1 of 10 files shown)


elek (commented):
Thanks to debug the problem @adoroszlai 

It looks good to me. But why don't we modify the default? If it's a generic problem (it's possible that Ozone couldn't be started in 2-3 minutes) we should fix it not just for the tests but everybody else.

Is there any risk to make the pipeline creation more frequent? (Even 10 sec?)

adoroszlai:
Thanks @elek for the review.

> It looks good to me. But why don't we modify the default? If it's a generic problem (it's possible that Ozone couldn't be started in 2-3 minutes) we should fix it not just for the tests but everybody else.

I'm not sure this is a problem for everybody, as timeout for "exit from safe mode" is test-specific.  Also, I guess the right value for this setting would depend on cluster size.

bshashikant (approved):

adoroszlai:
Thanks @bshashikant for the review.

PR #1390, opened 2020-09-03
HDDS-4196. Add an endpoint in Recon to query Prometheus

vivekratnavel:

- Add a metrics proxy endpoint to query metrics service providers like prometheus. 
  Example: `<recon_endpoint>/api/v1/metrics/query?query=ratis_leader_election_electionCount` will be proxied to `<prometheus_endpoint>/api/v1/query?query=ratis_leader_election_electionCount`
- Add a new config "ozone.recon.prometheus.http.endpoint" to configure Prometheus endpoint. The proxy endpoint (/api/v1/metrics) will respond with an error if Prometheus endpoint is not configured.
- Include `ratis_leader_election_electionCount` and `ratis_leader_election_lastLeaderElectionTime` metrics as part of the pipeline endpoint response if prometheus is configured.
**Note:** At this point `ratis_leader_election_lastLeaderElectionTime` is still not available and is tracked via https://issues.apache.org/jira/browse/RATIS-1049
- Add unit tests 

https://issues.apache.org/jira/browse/HDDS-4196

## How was this patch tested?

This patch was tested with unit tests and by verifying the Recon UI in a docker-compose cluster.
Screen Shot 2020-09-03 at 12 57 02 PM

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/MetricsServiceProviderFactory.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/MetricsServiceProviderFactory.java
@@ -1,0 +1,75 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.ozone.recon;
+
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
+import org.apache.hadoop.hdds.recon.ReconConfigKeys;
+import org.apache.hadoop.ozone.recon.spi.MetricsServiceProvider;
+import org.apache.hadoop.ozone.recon.spi.impl.PrometheusServiceProviderImpl;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * Factory class that is used to get the instance of configured Metrics Service
+ * Provider.
+ */
+@Singleton
+public class MetricsServiceProviderFactory {
+
+  private OzoneConfiguration configuration;
+  private ReconUtils reconUtils;
+
+  @Inject
+  public MetricsServiceProviderFactory(OzoneConfiguration configuration,
+                                       ReconUtils reconUtils) {
+    this.configuration = configuration;
+    this.reconUtils = reconUtils;
+  }
+
+  /**
+   * Returns the configured MetricsServiceProvider implementation (defaults
+   * to prometheus).
+   * If no metrics service providers are configured, returns null.
+   *
+   * @return MetricsServiceProvider instance that is configured.
+   */
+  public MetricsServiceProvider getMetricsServiceProvider() {
+    String prometheusEndpoint = getPrometheusEndpoint();
+    if (prometheusEndpoint != null && !prometheusEndpoint.isEmpty()) {
+      return new PrometheusServiceProviderImpl(configuration, reconUtils);
+    }
+    return null;
+  }
+
+  /**
+   * Returns the Prometheus endpoint if configured. Otherwise returns null.
+   *
+   * @return Prometheus endpoint if configured, null otherwise.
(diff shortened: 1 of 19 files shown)


hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/MetricsServiceProviderFactory.java
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Returns the configured MetricsServiceProvider implementation (defaults
+   * to prometheus).
+   * If no metrics service providers are configured, returns null.
+   *
+   * @return MetricsServiceProvider instance that is configured.
+   */
+  public MetricsServiceProvider getMetricsServiceProvider() {
+    String prometheusEndpoint = getPrometheusEndpoint();
+    if (prometheusEndpoint != null && !prometheusEndpoint.isEmpty()) {
```
avijayanhwx, line 55: Nit. can be StringUtils.isEmpty(prometheusEndpoint)
Nit. Let's add a log line saying we are choosing "Prometheus" as the configured Metrics provider here or inside the constructor of PrometheusServiceProviderImpl.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconConstants.java
```diff
@@ -52,6 +52,9 @@ private ReconConstants() {
   public static final String RECON_QUERY_BUCKET = "bucket";
   public static final String RECON_QUERY_FILE_SIZE = "fileSize";
 
+  public static final String PROMETHEUS_INSTANT_QUERY_API = "query";
```
avijayanhwx, line 55: Nit. These constants can go inside PrometheusServiceProviderImpl.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/MetricsProxyEndpoint.java
```diff
@@ -0,0 +1,117 @@
...
+          outputChannel.write(buffer);
+        }
+      } finally {
+        inputStream.close();
+      }
+    } else {
+      // Throw a Bad Gateway Error
+      httpServletResponse.sendError(HttpServletResponse.SC_BAD_GATEWAY,
+          "Metrics endpoint is not configured. Configure " +
+              PrometheusServiceProviderImpl.getEndpointConfigKey() + " and " +
```
avijayanhwx, line 113: Nit. getEndpointConfigKey() can be a method on the MetricsServiceProvider interface.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/metrics/Metric.java
```diff
@@ -0,0 +1,58 @@
...
+@InterfaceAudience.Private
+public final class Metric {
+
+  private Map<String, String> metadata;
+  private TreeMap<Double, Double> values;
+
+  public Metric(Map<String, String> metadata,
+                SortedMap<Double, Double> values) {
+    this.metadata = metadata;
+    this.values = (TreeMap<Double, Double>) values;
```
avijayanhwx, line 40: We should do values.addAll() or Collections. unmodifiableMap() here, since the caller may mutate the passed in reference.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/metrics/Metric.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Class for wrapping a metric response from
+ * {@link org.apache.hadoop.ozone.recon.spi.MetricsServiceProvider}.
+ */
+@InterfaceAudience.Private
+public final class Metric {
+
+  private Map<String, String> metadata;
+  private TreeMap<Double, Double> values;
```
avijayanhwx, line 35: Nit. Can be final fields.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/metrics/Metric.java
```diff
@@ -0,0 +1,58 @@
...
+                SortedMap<Double, Double> values) {
+    this.metadata = metadata;
+    this.values = (TreeMap<Double, Double>) values;
+  }
+
+  public Map<String, String> getMetadata() {
+    return metadata;
+  }
+
+  public void setMetadata(Map<String, String> metadata) {
```
avijayanhwx, line 47: Nit. Do we need setters in a Metric class?

avijayanhwx (commented):
Overall the approach LGTM @vivekratnavel. A few comments inline.

vivekratnavel:
@avijayanhwx Thanks for the review! Please take another look.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/MetricsProxyEndpoint.java
avijayanhwx: @vivekratnavel Makes sense, thanks!

avijayanhwx (approved):

PR #1393, opened 2020-09-04
HDDS-4143. Implement a factory for OM Requests that returns an instance based on layout version.

avijayanhwx:
- Implement Generic factory which stores different instances of Type 'T' sharded by a key & version. A single key can be associated with different versions of 'T'. This is to support a typical use case during upgrade to have multiple versions of a class / method / object and chose them based  on current layout version at runtime. Before finalizing, an older version is typically needed, and after finalize, a newer version is needed.
- Using the generic factory, we scan all the different OM "write" requests and associate them with versions.
- Persist current MLV to Ratis requests.
- Layout feature code refactoring. Added more comments and tests.

https://issues.apache.org/jira/browse/HDDS-4143

## How was this patch tested?
Unit tested.


avijayanhwx:
Marking as DRAFT until there are no major problems in CI.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/upgrade/TestOmRequestFactory.java
```diff
@@ -0,0 +1,116 @@
...
+      omRequestFactory.getRequestType(OMRequest.newBuilder()
+              .setCmdType(CreateKey)
+              .setClientId("c1")
+              .setLayoutVersion(LayoutVersion
+                  .newBuilder()
+                  .setVersion(CREATE_EC.layoutVersion())
+                  .build())
+              .build());
+      Assert.fail();
+    } catch (IllegalArgumentException ex) {
```
linyiqun, line 84: Can we do the verify for error message that contained in IllegalArgumentException? This will let us know why failure was happened.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/upgrade/TestOMVersionManager.java
```diff
@@ -54,11 +61,39 @@ public void testOMLayoutVersionManager() throws IOException {
...
+    try {
+      OMLayoutVersionManager.initialize(omStorage);
+      Assert.fail();
+    } catch (OMException ex) {
+      assertEquals(NOT_SUPPORTED_OPERATION, ex.getResult());
+    }
+  }
+
+  @Test
+  public void testOMLayoutVersionManagerReset() throws IOException {
```
linyiqun, line 79: Just curious about this scenario, when will we invoke OMLayoutVersionManager#resetLayoutVersionManager call.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
```diff
@@ -290,6 +292,11 @@ public void validateRequest(OMRequest omRequest) throws OMException {
       throw new OMException("ClientId is null",
           OMException.ResultCodes.INVALID_REQUEST);
     }
+
+    if (omRequest.getLayoutVersion() == null) {
```
linyiqun, line 296: Would be better to add a comment here to explain why we do layout version empty check on handling write request.

linyiqun (commented):
@avijayanhwx , the latest PR looks good, some minor comments from me.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/upgrade/TestOMVersionManager.java
avijayanhwx: @linyiqun That was done just for testing purposes. In the future, we may move away from "static" Version Manager class, and that time we can remove this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
avijayanhwx: Sure, will add.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/upgrade/TestOmRequestFactory.java
avijayanhwx: Will do.

avijayanhwx:
Thanks for the review @linyiqun. I will be addressing the comments in the next PR that I create.

PR #1394, opened 2020-09-04
HDDS-4205. Make coverage upload to codecov optional

elek:

Motivation is discussed on the Apache jira:

 1. We make codecov upload optional (accept failures)
 2. And execute it only for the master branch

https://issues.apache.org/jira/browse/HDDS-4205

## How was this patch tested?

As modification explicit says that it will be executed on non-apache repositories: it's not easy to test. But it's a small change and copied the same condition from the Sonar upload.


adoroszlai (approved):
Thanks @elek for making this change.

PR #1395, opened 2020-09-04
HDDS-4088. Adding Owner info for Authorizer plugin to honor owner access rights

xiaoyuyao:

adding isowner flag and volume owner info to acl check request context so that Ranger and 3rd party plugin can allow volume owner has full control over the volume

https://issues.apache.org/jira/browse/HDDS-4088

## How was this patch tested?
UT added.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -1667,8 +1667,12 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
     try {
       volumeArgs = metadataManager.getVolumeTable().get(dbVolumeKey);
     } catch (IOException ioe) {
-      throw new OMException("Volume " + volume + " is not found",
-          OMException.ResultCodes.VOLUME_NOT_FOUND);
+      if (ioe instanceof OMException) {
+        throw (OMException)ioe;
+      } else {
+        throw new OMException("getVolumeOwner for Volume " + volume + " failed",
+            ResultCodes.INTERNAL_ERROR);
+      }
     } finally {
       if (lockAcquired) {
         metadataManager.getLock().releaseReadLock(VOLUME_LOCK, volume);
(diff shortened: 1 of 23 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1642,25 +1644,42 @@ private boolean hasAcls(String userName, ResourceType resType,
...
+    return volOwnerName;
+  }
+
+  private String getVolumeOwner(String volume) throws OMException {
+    Boolean lockAcquired = metadataManager.getLock().acquireReadLock(
+        VOLUME_LOCK, volume);
+    String dbVolumeKey = metadataManager.getVolumeKey(volume);
+    OmVolumeArgs volumeArgs = null;
+    try {
+      volumeArgs = metadataManager.getVolumeTable().get(dbVolumeKey);
```
smengcl, line 1668: `getVolumeTable().get(vol)` seems to return `null` when volume not found rather than throwing.

the exception message in the following catch block needs to be changed.

smengcl (commented):
Thanks @xiaoyuyao  for the patch. Overall lgtm.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
xiaoyuyao: Thanks @smengcl  for the review. The null check of volumeArgs for volume not found is done outside the lock between Line 1677-1682.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1642,25 +1644,42 @@ private boolean hasAcls(String userName, ResourceType resType,
...
+    } catch (IOException ioe) {
+      throw new OMException("Volume " + volume + " is not found",
+          OMException.ResultCodes.VOLUME_NOT_FOUND);
```
smengcl, line 1671: I should tag here :) #tag

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
smengcl: Yes @xiaoyuyao . I do see the Line 1677-1682 check and message.

I meant Line 1670-1671 OMException, since `volumeArgs` won't throw if volume is not found, we should change that message. Or just wrap IOException inside OMException?

xiaoyuyao: Thanks for the clarification. I just pushed a change to refine the exception for this case as suggested.

smengcl (approved):
lgtm+1

xiaoyuyao:
Thanks @smengcl  for the review.

PR #1397, opened 2020-09-04
HDDS-4211. [OFS] Better owner and group display for listing Ozone volumes and buckets

smengcl:

Improve volumes' and buckets' owner and group display when listing root and in volumes in OFS.

1. Display short name instead of full Kerberos principal.
2. For volumes, get actual group of the owner (currently it is the volume admin name which is incorrect)
3. For buckets, display the owner and group of its parent volume.

https://issues.apache.org/jira/browse/HDDS-4211

## How was this patch tested?

Modified `testListStatusRootAndVolumeNonRecursive` to add the owner and group check for volumes and buckets listing.

### DIFF
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
@@ -663,12 +663,16 @@ public class BasicRootedOzoneClientAdapterImpl
     OFSPath ofsStartPath = new OFSPath(startPath);
     // list buckets in the volume
     OzoneVolume volume = objectStore.getVolume(volumeStr);
+    UserGroupInformation ugi =
+        UserGroupInformation.createRemoteUser(volume.getOwner());
+    String owner = ugi.getShortUserName();
+    String group = getGroupName(ugi);
     Iterator<? extends OzoneBucket> iter =
         volume.listBuckets(null, ofsStartPath.getBucketName());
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
```diff
@@ -972,24 +992,26 @@ private static FileStatusAdapter getFileStatusAdapterForVolume(
...
+              ozoneBucket.getName(), pathStr);
     }
     Path path = new Path(pathStr);
     return new FileStatusAdapter(0L, path, true, (short)0, 0L,
         ozoneBucket.getCreationTime().getEpochSecond() * 1000, 0L,
-        FsPermission.getDirDefault().toShort(),  // TODO: derive from ACLs later
-        // TODO: revisit owner and group
-        username, username, path, new BlockLocation[0]);
+        FsPermission.getDirDefault().toShort(),
+        // TODO: maybe derive owner and group from ACLs later
```
xiaoyuyao, line 1013: NIT: The comment is not accurate. The ACLs will not be able to provider owner and group information. We might add owner information as bucket metadata in future.

xiaoyuyao (approved):
Thanks @smengcl  for improving this. LGTM, +1.

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneClientAdapterImpl.java
smengcl: Thanks! Removed the comment.

PR #1398, opened 2020-09-04
HDDS-4210. ResolveBucket during checkAcls fails.

bharatviswa504:

When bucket links are levels, When ACLS are enabled, during `checkAcls` It will fail with NPE.
Finally in HA, this causes OM to be terminated.

https://issues.apache.org/jira/browse/HDDS-4210

## How was this patch tested?

Added UT.

The same test will fail without the fix.

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/test.sh
@@ -28,7 +28,7 @@ start_docker_env
 
 execute_robot_test scm kinit.robot
 
-execute_robot_test scm freon
+execute_robot_test scm freonexecute_robot_test scm basic/links.robot
 
 stop_docker_env
 
--- a/hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/test.sh
@@ -28,7 +28,9 @@ start_docker_env
 
 execute_robot_test scm kinit.robot
 
-execute_robot_test scm freonexecute_robot_test scm basic/links.robot
+execute_robot_test scm freon
+
+execute_robot_test scm basic/links.robot
 
 stop_docker_env
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAWithData.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAWithData.java
@@ -21,6 +21,7 @@ import org.apache.hadoop.hdds.client.ReplicationFactor;
 import org.apache.hadoop.hdds.client.ReplicationType;
 import org.apache.hadoop.ipc.RemoteException;
 import org.apache.hadoop.ozone.OzoneConfigKeys;
+import org.apache.hadoop.ozone.client.BucketArgs;
 import org.apache.hadoop.ozone.client.ObjectStore;
 import org.apache.hadoop.ozone.client.OzoneBucket;
 import org.apache.hadoop.ozone.client.OzoneKeyDetails;
@@ -44,6 +45,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 import java.util.UUID;
 
 import static org.apache.hadoop.ozone.MiniOzoneHAClusterImpl.NODE_FAILURE_TIMEOUT;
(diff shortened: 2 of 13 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAWithData.java
```diff
@@ -69,6 +70,38 @@ public void testAllOMNodesRunning() throws Exception {
     createKeyTest(true);
   }
 
+  @Test
+  public void testBucketLinkOps() throws Exception {
```
adoroszlai, line 74: We already have acceptance test for links, but currently it is only run in non-HA setup.  Instead of adding integration test, I think we should run the existing test in HA setup, too:

```
execute_robot_test scm basic/links.robot
```

could be added to `hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/test.sh` and maybe `hadoop-ozone/dist/src/main/compose/ozone-om-ha-s3/test.sh`.

This confirms that the bug and the fix.

However, I also see two of the test cases failing with the fix:

```
Cannot follow link without read access                                | FAIL |
ACL verified on source bucket                                         | FAIL |
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3523,6 +3533,47 @@ public ResolvedBucket resolveBucketLink(Pair<String, String> requested)
...
+   *   avoid infinite loops
+   * @return bucket location possibly updated with its actual volume and bucket
+   *   after following bucket links
+   * @throws IOException (most likely OMException) if ACL check fails, bucket is
+   *   not found, loop is detected in the links, etc.
+   */
+  private Pair<String, String> resolveBucketLink(
+      Pair<String, String> volumeAndBucket,
+      Set<Pair<String, String>> visited,
+      OMClientRequest omClientRequest) throws IOException {
```
adoroszlai, line 3552: Would it be possible to replace the original `resolveBucketLink` method with this one, instead of duplicating it?

adoroszlai (changes requested):
Thanks @bharatviswa504 for fixing this bug.  I have some suggestions, and I also see two unit test failures:

```
org.apache.hadoop.ozone.om.request.key.TestOMKeysRenameRequest
org.apache.hadoop.ozone.om.request.key.TestOMKeysDeleteRequest
```

https://github.com/apache/hadoop-ozone/pull/1398/checks?check_run_id=1073943761#step:4:4

xiaoyuyao (commented):
Thanks @bharatviswa504  for working on this. It LGTM overall, just a few comments added inline.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
bharatviswa504: I have taken this approach of passing OMClientRequest so that we don't need to touch all OMKey write Requests in passing this info.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAWithData.java
bharatviswa504: Thanks for the tip. Enabled docker-compose tests.

hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/test.sh
```diff
@@ -30,6 +30,8 @@ execute_robot_test scm kinit.robot
 
 execute_robot_test scm freon
 
+execute_robot_test scm basic/links.robot
```
amaliujia, line 33: Out of curiosity: why adding this robot test to both

` hadoop-ozone/dist/src/main/compose/ozone-om-ha-s3/test.sh` and
`hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/test.sh`.

Will it just run the same test twice on each CI?

bharatviswa504: Good question might be it is good to run only during security-enabled cluster.
But it has some tests which are bucket link features, so I thought it might be good to run in both secure and non-secure.

adoroszlai: I think it's useful to run the same test in different environments: secure and non-secure, non-HA and HA.  It increases coverage.  Also, non-secure versions provide some kind of baseline: if it works and secure one does not, we might know the problem is security-related.

If it takes too much time, we can tweak acceptance test splits (currently we have 3 concurrent runs in CI).

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3479,10 +3481,35 @@ public ResolvedBucket resolveBucketLink(OmKeyArgs args)
...
       throws IOException {
-    Pair<String, String> resolved =
-        resolveBucketLink(requested, new HashSet<>());
+
+    Pair<String, String> resolved;
+    if (isAclEnabled) {
+      resolved = resolveBucketLink(requested, new HashSet<>(),
+              ProtobufRpcEngine.Server.getRemoteUser(),
+              ProtobufRpcEngine.Server.getRemoteIp(),
+              ProtobufRpcEngine.Server.getRemoteIp().getHostName());
```
adoroszlai, line 3508: Nit: to avoid Sonar warning that _"static" base class members should not be accessed via derived types_.

```suggestion
              Server.getRemoteUser(),
              Server.getRemoteIp(),
              Server.getRemoteIp().getHostName());
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
adoroszlai: Thanks!

adoroszlai (approved):
Thanks @bharatviswa504 for updating the patch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
adoroszlai: Thanks!

hadoop-ozone/dist/src/main/compose/ozonesecure-om-ha/test.sh
amaliujia: Makes sense. Thanks for all your clarification!

bharatviswa504:
@xiaoyuyao Do you have any more comments?

bharatviswa504:
Review comments are addressed, and no further comments, proceeding with commit.
Thank You @adoroszlai and @xiaoyuyao  for the review.

PR #1400, opened 2020-09-04
HDDS-4141. Implement Finalize command in Ozone Manager client.

fapifta:
As part of the upgrade development, this addition is aiming to create the client side implementation of the finalizeUpgrade initiation and monitoring parts.
In order to test it the PR has some classes and some modifications that are related to the server side development, but are added here, to be able to test out the command implemented in FinalizeUpgradeSubCommand class.
The PR also contains the proto message additions for the two newly implemented requests, finalieUpgrade and finalizeUpgreadeProgress request and response pair, and a status message sent between the client and server with these requests.

The idea is the following:
only one client can initialize and monitor the finalization, so that we have a client ID with the requests, but in case of a client side failure the command allows a takeover from the original initiator to finish the monitoring.

The finalization is a write request, that potentially affects all table caches in OM, so CleanupTableInfo annotation was extended with a cleanupAll option. And the tests were added to accommodate that.
OmMetaDataManager was extended with an ALL_TABLES psf field, to have the full table list at the same place.
I was exploring the idea of using OmDbDefinition#getColumnFamilies here, but in order to use that we need to change the annotation in all the response classes to use DBColumnFamilyDefinition instead of String in cleanupTables field, as in the annotation it is impossible to point to a definition's getTableName method, and refactoring that should be done in a separate task, it affects too many files and code paths to include it here.
Also during toying with the idea it turned out that OmDbDefinition#getColumnFamilies returns S3_TABLE that is not there anymore in OM as I found, so probably we should remove it.

The server side handling is purely just toying with some ideas, and in a state where I got to a few problems/questions. At that point we have decided that the minimum viable solution is something like this:
- finalizeUpgrade will post one request to Ratis, and all OMs will start finalization and finish finalization as part of that request.
- monitoring will be supplied status from the request handling in the Leader and will emit finalization done when OM reached majority processing the finalization request.
- finalization itself will ensure to run the finalization of all the new LayoutFeature-s that are enabled and applied, and will update the MLV in all 3 OMs
A proposal that describes problems stemming from this approach and with a few possibilities to make this more robust will be added in a follow up JIRA later on as we finished the minimalistic first version, as ideas explore messaging between OMs outside Ratis, posting finalize requests for LayoutFeatures in a lazy manner one by one to ratis, just to mention a few from a couple possible approaches, and this for sure requires a lot of time and exploration and a consensus on how deep we want to go down this rabbit hole.

## How was this patch tested?

In ha and non-ha docker environments possible scenarios to run the command was tested by hand.


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
```diff
@@ -310,6 +311,69 @@ OmKeyLocationInfo allocateBlock(OmKeyArgs args, long clientID,
...
+   * progress, if any. The status returned is
+   * - FINALIZATION_IN_PROGRESS, and the messages since the last query if the
+   *    finalization is still running
+   * - FINALIZATION_DONE with a message list containing the messages since
+   *    the last query, if the finalization ended but the messages were not
+   *    yet emitted to the client.
+   * - ALREADY_FINALIZED with an empty message list otherwise
+   * - If finalization is not in progress, but software layout version and
+   *    metadata layout version are different, the method will throw an
+   *    {@link OMException} with a result code INVALID_REQUEST
```
linyiqun, line 358: INVALID_REQUEST is a coarse-grained result code, why not define more understandable code in OMException#ResultCodes for matching current scenario?

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -1003,6 +1009,36 @@ message DBUpdatesResponse {
...
 }
 
+
+message UpgradeFinalizationStatus {
+  enum Status {
+    ALREADY_FINALIZED = 1;
+    STARTING_FINALIZATION = 2;
+    FINALIZATION_IN_PROGRESS = 3;
+    FINALIZATION_DONE = 4;
+    FINALIZATION_REQUIRED = 5;
```
linyiqun, line 1019: Two comments for above status defined:
* Would be better rename FINALIZATION_REQUIRED to FINALIZATION_NOT_STARTED?
* Just from the status name, I don't see obvious difference between ALREADY_FINALIZED and FINALIZATION_DONE and I notice that ALREADY_FINALIZED is not set anywhere.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2613,6 +2615,73 @@ public ServiceInfoEx getServiceInfo() throws IOException {
...
+  private final List<String> finalizationMsgs = new ArrayList<>();
+  private UpgradeFinalizationStatus.Status finalizationStatus =
+      UpgradeFinalizationStatus.Status.FINALIZATION_REQUIRED;
+
+  @Override
+  public UpgradeFinalizationStatus finalizeUpgrade(String upgradeClientID)
+      throws IOException {
+    if (!finalizationStatus
+        .equals(UpgradeFinalizationStatus.Status.FINALIZATION_REQUIRED)){
+      throw new OMException("Finalization is not needed.", INVALID_REQUEST);
```
linyiqun, line 2627: As I commented, FINALIZATION_REQUIRED and INVALID_REQUEST could use a better more understandable name, : ).

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2613,6 +2615,73 @@ public ServiceInfoEx getServiceInfo() throws IOException {
...
+    }
+    finalizationStatus = UpgradeFinalizationStatus.Status.STARTING_FINALIZATION;
+    UpgradeFinalizationStatus status = UpgradeFinalizationStatus.newBuilder()
+        .setStatus(finalizationStatus)
+        .build();
+    LOG.info("finalizeUpgrade initiated by client: {}.", upgradeClientID);
```
linyiqun, line 2633: finalizeUpgrade -> FinalizeUpgrade

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2613,6 +2615,73 @@ public ServiceInfoEx getServiceInfo() throws IOException {
...
+        LOG.info("Finalization done.");
+        finalizationStatus = UpgradeFinalizationStatus.Status.FINALIZATION_DONE;
+      }, "Finalization-Thread").start();
+    }
+    return status;
+  }
+
+  @Override
+  public UpgradeFinalizationStatus queryUpgradeFinalizationProgress(
+      String upgradeClientID, boolean force
```
linyiqun, line 2671: force parameter is not actually used here, this parameter is related to the client takeover behavior.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/FinalizeUpgradeSubCommand.java
```diff
@@ -0,0 +1,221 @@
...
+        System.err.println("Invalid response from Ozone Manager.");
+        System.err.println(
+            "Current finalization status is: " + status.getStatus()
+        );
+        throw new IOException("Exiting...");
+      }
+    } catch (OMException e) {
+      handleInvalidRequestAfterInitiatingFinalization(e);
+    }
+    monitorAndWaitFinalization(client, upgradeClientID);
```
linyiqun, line 97: Sometimes users just want to query immediate upgrade status and don't want to wait and monitor finalization finished. Can we add one additionally option that we can skip monitorAndWaitFinalization?

linyiqun (commented):
Hi @fapifta , thanks for the work for this. 
Just some minor comments here. Please have a look when you get a chance.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
fapifta: There are a few different states here that we need to handle and signal.
- When Ozone Manager starts with an MLV < SLV that is the FINALIZATION_REQUIRED state, where we know that the software was updated and was not finalized yet.
- When Ozone Manager starts with an MLV = SLV that is the state when we are ALREADY_FINALIZED, this is not set so far, as the server side code is there in this patch to be able to demo and test the client functionality.
From ALREADY_FINALIZED, there should be no way out until the MLV = SLV is true, and that can only be changed normally during a restart if software is updated, and SLV is increasing. (MLV can be manipulated on the disk as well, but that does not considered as normal.)
- There is a chance that MLV >= SLV in which case Ozone Manager should not start until we do not support downgrade after finalization, and tbh it does not really makes sense to support downgrade after finalizing metadata changes, as that easily can become an impossible scenario especially during runtime.

From FINALIZATION_REQUIRED, there are one valid state change, STARTING_FINALIZATION, from which after sending the response, the OM should immediately switch forward to FINALIZATION_IN_PROGRESS.
When finalization is finished, then the state shoudl be FINALIZATION_DONE. One might argue this is the same as ALREADY_FINALIZED, but there is a slight difference here. In FINALIZATION_DONE state there might still be messages to be emitted to the monitoring client, while when ALREADY_FINALIZED, the progress request is also invalid.

During the implementation of HDDS-4172 this should be cleaned up much more, and when the last message is emitted then status should be switched to ALREADY_FINALIZED making subsequent progress monitoring requests to result in proper status, it was not really a consideration for the server side of things in this patch which is just there to see how the communication happens.

One more thing you may notice, there is no error status here. The reason is that at this stage, we plan to go with the simplest solution first, and commit the finalization work into the raft log in one step. Which means if it fails, it throws an exception, and MLV should not change as that is the final step. So when there is an error, next time the OM comes up, a new finalization attempt can be started.
There are a lot of ideas how to separate finalization into more steps, but there are certain problems that has to be solved there. I will also post a proposal about this later this week, describing why we choose the single transaction finalization approach first and how we can evolve it to be more system and developer friendly.

Also when both sides of the system is ready, I plan to add a detailed documentation on the different states to preserve this knowledge for the future as the last step, as there are still may experimenting is going on to explore how this is best done.

Let me know if this makes naming more clear and reasonable.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
fapifta: Yepp, this is an initial choice to use something that is already there, to test out the client, this should be changed with HDDS-4172 when the server side implementation is happening. I thought it is easier to add server side error message together with the real server side logic as there might be multiple values turning out to be required for clear error messages, but I can change this here if needed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
fapifta: As written to the previous comments, this part of the code will be changed with the server side logic implemented where we are still experimenting with a few things. But yes error messages has to be better, though states might remain as they are just handled fully as described in the comment for the CientProtocol.proto

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/FinalizeUpgradeSubCommand.java
fapifta: There is a plan to expose the current status via JMX(HDDS-4142), would that be a sufficient way to provide this functionality?
If JMX is not enough to expose a queryable state, this might come in as an improvement for the client side. I am happy to implement it here but I would get back to this after the core logic is working in a separate JIRA if you don't mind, let me know if that works for you, and I will file the JIRA.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
fapifta: In the PR that is forming for HDDS-4172 (the server side implementation), the parameter should be used to validate if we are getting the progress monitoring request from the same client, and error out if we do not.
But in the demo code, I haven't spent time on this part.
On the other hand as from the client side this has changed to be the -takeover parameter, I am renaming the parameter name to takeover here and in proto as well.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
fapifta: thank you for spotting this, I am fixing it in the follow up commit I am preparing shortly.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
linyiqun: Comments makes sense to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
linyiqun: Comments makes sense to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
linyiqun: I have seen that force has been renamed to -takeover. I will look into server side implementation in HDDS-4172 once you attach the new PR there.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/FinalizeUpgradeSubCommand.java
linyiqun: Yes, of course, we can address above suggestion later. That's just an improvement.

linyiqun (approved):
LGTM, +1.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/upgrade/OMFinalizeUpgradeProgressResponse.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
+
+import java.io.IOException;
+
+/**
+ * Response for finalizeUpgradeProgress request.
+ */
+// yepp this will not be a write request, adding a table here to the annotation
+// just to pass tests related to this annotation.
+@CleanupTableInfo(cleanupTables = { OmMetadataManagerImpl.USER_TABLE })
```
avijayanhwx, line 34: Can we confirm that this will not have a "side effect"? Is there a better way to do this, like passing an empty array and handling it as a special case in test?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMFinalizeUpgradeProgressRequest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+/**
+ * Handles finalizeUpgradeProgress request that serves to query the status
+ * of the async finalization progress.
+ */
+public class OMFinalizeUpgradeProgressRequest extends OMClientRequest {
```
avijayanhwx, line 40: As discussed offline, this can be a READ request. I am ok with doing it in the next patch (HDDS-4172).

sodonnel (approved):
The approach here looks good to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/upgrade/OMFinalizeUpgradeProgressResponse.java
fapifta: I already started to work on the server side, one of the first steps is to make this reguest a read request and delete this class, as read requests are handled in an other way. I can push that commit to this PR or it will be addressed in the PR coming for HDDS-4172.

avijayanhwx: @fapifta I am Ok with handling this in a follow up patch.

avijayanhwx (approved):
LGTM +1.

avijayanhwx:
Thanks for working on this @fapifta, and for the reviews @linyiqun , @sodonnel.

PR #1402, opened 2020-09-06
HDDS-4213. Log when a datanode has become dead in the DeadNodeHandler

maobaolong:

Print the dead datanode when DEAD message comming

HDDS-4213

## How was this patch tested?

Kill a datanode, after x minute past, watching the log of SCM, you can see the log `A dead datanode detected.`, and you can get the dead datanode information clearly.


sodonnel (approved):
+1 looks good to me. It is a good idea to print this information into the log.

PR #1405, opened 2020-09-07
HDDS-4143. Implement a factory for OM Requests that returns an instance based on layout version.

avijayanhwx:
- Implement Generic factory which stores different instances of Type 'T' sharded by a key & version. A single key can be associated with different versions of 'T'. This is to support a typical use case during upgrade to have multiple versions of a class / method / object and chose them based  on current layout version at runtime. Before finalizing, an older version is typically needed, and after finalize, a newer version is needed.
- Using the generic factory, we scan all the different OM "write" requests and associate them with versions.
- Persist current MLV to Ratis requests.
- Layout feature code refactoring. Added more comments and tests.

https://issues.apache.org/jira/browse/HDDS-4143

## How was this patch tested?
Unit tested.


avijayanhwx:
cc @linyiqun Addressed the review comments from you in this PR.

linyiqun:
@avijayanhwx , thanks for addressing the comments. LGTM now. 
Above build check is not show green, some steps are failed. Is that current change broken current ut? Can you have a check for this?

avijayanhwx:
> @avijayanhwx , thanks for addressing the comments. LGTM now.
> Following build check is not show all green, some steps are failed. Is that current change broken existing unit tests? Can you have a check for this?

Yes, my patch causes an s3g start failure which is causing all s3 related acceptance tests to fail. Looking into it.

avijayanhwx:
cc @fapifta  / @prashantpogde.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
```diff
@@ -206,4 +207,8 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
       return omClientResponse;
     }
   }
+
+  public static String getRequestType() {
```
bharatviswa504, line 211: Instead of a new method, we can use annotation.

avijayanhwx: @bharatviswa504 Yes, I pondered about adding annotation for this. Given that we already have annotations for Cleanup tables, and BelongsToLayoutVersion, I wanted to hold off on adding more annotations. If it is ok with you, after the first round of changes, we can change this to an annotation.

fapifta:
Hi @avijayanhwx, thank you for working on this.
Please find some comments below.

1. First and foremost I would like to comment on the LayoutVersionInstanceFactory.instances data structure as this data structure seems to be the most performance critical, because this will be used to look up the implementation class for every request. I may have an idea to make this a bit better, which I would like to discuss.
HashMap access is constant time however that constant time does include hashing.
TreeMap access is O(log(n)).

What we use here in the Map as key is the name of an enum, however this gets tricky when it comes to ACLs, and the Integer is the LayoutVersion what comes as well from an enum, however in OMLayoutVersion, we use a separate layoutVersion, it is still equals to the ordinal of the enum.

So if we change two things:
- OMLayoutVersion to use the enum's ordinal as the LayoutVersion
- ACL requests to be enumerated as separate request types for volume bucket and key ACL operations, instead of the dynamic string approach for the request type.

Then we can use an EnumMap<OzoneManagerProtocolProtos.Type, EnumMap<OMLayoutFeature, T>>, if we want to make the factory usable by other parts of the code, then we can supply OzoneManagerProtocolProtos.Type, and OMLayoutFeature as type parameters to the factory as needed.
With that we would have enummap of enummaps with guaranteed constant time access (array access basically), which is faster.

Of course this will result in some more memory consumption and a fair bit more complicated initialization as a tradeoff, because we would need to register class T to all LayoutVersions for every request type but even with fairly large request (1-200) and version counts(~1000), the required memory for these lookup tables are still in the few Mib range, if my estimation skills are not screwing me over, so it seems to be a fair tradeoff.

Having an enum that enumerates all the requests (ACLs for every type, and property for every type separately as an abstraction in OM, we can use that in many other ways possibly, and that also solves the getRequestType()->annotation conversion, as with the current request types, we can not introduce an annotation, as for example the "SetAcl.name() + "-" + ObjectType.KEY" expression does not qualify as a constant and can not be used as value inside an annotation.
What do you think?

2.
It would be nice to add a test for OMClientRequest, that ensures that it has and all of its implementations has a declared constructor with an OMRequest parameter, so that if for some reason this changes we can get a nice detailed notification about the problem with the change, and based on the test error, we can either change the OzoneManagerRatisUtils class to call the proper constructor, or consider the change in the constructor.

3.
I am unsure if we need to restrict BelongsToLayoutFeature to classes and DisallowedUntilLayoutVersion to methods, what is a reasoning behind such a restriction? As I see, the code does not understand the annotations in other places, and probably it is better to restrict, I am just curious if we thought about methods with features, or completely disallowed classes, in the future? So this is more like a question than a concern for now.

4.
TestOmRequestFactory does not have a closing empty line.

5.
I am not fully convinced wether it is a consistent and good design decision to do not have an interface to the LayoutVersionInstanceFactory, and leave types like OMLayoutVersionIstanceFactory lingering around in the API, are there any plans to generalize this to an interface -> abstract impl -> concrete impl structure as with LayoutVersionManager and use the interface only everywhere where we need to pass it in? Are there any reasons not to do so?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
fapifta: The problem currently with an annotation approach is the fact that we can not use expressions like "SetAcl.name() + "-" + ObjectType.KEY" in annotations, as that does not qualify as a constant.

avijayanhwx:
> Hi @avijayanhwx, thank you for working on this.
> Please find some comments below.
> 
> 
> First and foremost I would like to comment on the LayoutVersionInstanceFactory.instances data structure as this data structure seems to be the most performance critical, because this will be used to look up the implementation class for every request. I may have an idea to make this a bit better, which I would like to discuss.
> HashMap access is constant time however that constant time does include hashing.
> TreeMap access is O(log(n)).
> 
> What we use here in the Map as key is the name of an enum, however this gets tricky when it comes to ACLs, and the Integer is the LayoutVersion what comes as well from an enum, however in OMLayoutVersion, we use a separate layoutVersion, it is still equals to the ordinal of the enum.
> 
> So if we change two things:
> 
> * OMLayoutVersion to use the enum's ordinal as the LayoutVersion
> * ACL requests to be enumerated as separate request types for volume bucket and key ACL operations, instead of the dynamic string approach for the request type.
> 
> Then we can use an EnumMap<OzoneManagerProtocolProtos.Type, EnumMap<OMLayoutFeature, T>>, if we want to make the factory usable by other parts of the code, then we can supply OzoneManagerProtocolProtos.Type, and OMLayoutFeature as type parameters to the factory as needed.
> With that we would have enummap of enummaps with guaranteed constant time access (array access basically), which is faster.
> 
> Of course this will result in some more memory consumption and a fair bit more complicated initialization as a tradeoff, because we would need to register class T to all LayoutVersions for every request type but even with fairly large request (1-200) and version counts(~1000), the required memory for these lookup tables are still in the few Mib range, if my estimation skills are not screwing me over, so it seems to be a fair tradeoff.
> 
> Having an enum that enumerates all the requests (ACLs for every type, and property for every type separately as an abstraction in OM, we can use that in many other ways possibly, and that also solves the getRequestType()->annotation conversion, as with the current request types, we can not introduce an annotation, as for example the "SetAcl.name() + "-" + ObjectType.KEY" expression does not qualify as a constant and can not be used as value inside an annotation.
> What do you think?
> 
> 
> It would be nice to add a test for OMClientRequest, that ensures that it has and all of its implementations has a declared constructor with an OMRequest parameter, so that if for some reason this changes we can get a nice detailed notification about the problem with the change, and based on the test error, we can either change the OzoneManagerRatisUtils class to call the proper constructor, or consider the change in the constructor.
> 
> 
> I am unsure if we need to restrict BelongsToLayoutFeature to classes and DisallowedUntilLayoutVersion to methods, what is a reasoning behind such a restriction? As I see, the code does not understand the annotations in other places, and probably it is better to restrict, I am just curious if we thought about methods with features, or completely disallowed classes, in the future? So this is more like a question than a concern for now.
> 
> 
> TestOmRequestFactory does not have a closing empty line.
> 
> 
> I am not fully convinced wether it is a consistent and good design decision to do not have an interface to the LayoutVersionInstanceFactory, and leave types like OMLayoutVersionIstanceFactory lingering around in the API, are there any plans to generalize this to an interface -> abstract impl -> concrete impl structure as with LayoutVersionManager and use the interface only everywhere where we need to pass it in? Are there any reasons not to do so?

Thanks for the detailed code review @fapifta. Keeping in mind that we are still laying out the building blocks of the upgrade framework, and we plan to iteratively make it better, here are my responses.

1. Yes, I think this is a good catch. Although, we don't expect a proliferation of versions for a given requests type, we should avoid a O(logn) lookup on every write operation. For that, my proposal to move to a O(1) lookup cost is this. The factory will maintain ALL types >= MLV in a sorted data structure (maybe a priority queue). By doing that it guarantees O(1) lookup at all times, since we always would lookup the first element (top of the PQ). Multiple entries will be there ONLY during pre-finalized state. On finalization, we will be removing the entry one by one until we reach a single entry. On a regular OM (finalized), there will be a single request version associated with a request always. 

In the future, we could explore EnumMap that would bring more changes to the OM request layer.

2. Added.

3. I believe it is easy to extend the Target (method/class) of the annotations once we find some use cases for them.

4. Will fix it. 

5. The request factory is currently just used by OM, and hence I would like to maintain it in a separate interface than the vanilla LayoutVersionManager interface. However, I have extended the LayoutVersionManager interface into an OMLayoutVersionManager which adds OM request specific "READ" requests. Thus, we make sure that the getVersionManager() in the OzoneManager.java class gives out a READ only access to the system. Hopefully this makes better sense.

fapifta:
Hi @avijayanhwx, thank you for addressing my comments, the current solution seems to be good enough. As we discussed in a live conversation, I agree that introducing the EnumMap has consequences on how we handle the OM requests, and probably should go in an other patch, though I still believe we should work on this, and try to leave out hashing as well later.

So on the main points we got to a status where I can comfortably +1 the changeset. There is one small typo which should be fixed before we are commit it. (I add an inline comment for that.)

I have a few notes though on naming, and some possible extractions that can help the understandability of the code:
- In TestOmVersionManagerRequestFactory.java line 75, the test method might be named to something that implies we are checking for the existence of the getRequestType method as well, or if the method does not exists we should fail the test with a custom error message.
- In LayoutVersionInstanceFactory.java:
getInstances() method could return an UnModifiableMap instead of a copy, and we can use default visibility
we should name the long conditional expressions, in line 193, and 199
in the register method there is a large system of if statements from line 108, I think we can reorganise this to make it simpler to understand, if we return false first if that needs to be returned, then do a poll() if required, and then offer and return true otherwise. In this case we might extract and name the conditions and with that we might not need that much documentation either.

None of the above I think is absolutely necessary, look at them as suggestions for your consideration.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/LayoutVersionInstanceFactory.java
```diff
@@ -0,0 +1,246 @@
...
+      }
+
+      if ((vInstances.isEmpty() ||
+          vInstances.peek().version > lvm.getMetadataLayoutVersion())
+          && prevInstance != null) {
+        vInstances.offer(prevInstance);
+      }
+
+      if (vInstances.isEmpty()) {
+        LOG.info("Unregistering '{}' from factory since it has no entries." +
```
fapifta, line 206: I think we need a "," here instead of "+".

swagle (approved):
+1 LGTM latest version.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
bharatviswa504: If that is the problem we can define a static variable with direct value and do it :)

avijayanhwx: @bharatviswa504 As discussed offline, we can revisit in the next iteration.

fapifta:
Thank you for addressing the concerns, +1(non-binding) to commit the last patch.

avijayanhwx:
Thanks for the reviews. This is ongoing work on a branch, and the implementation can be revisited in case of any pitfall. Merging this now.

PR #1410, opened 2020-09-08
HDDS-3102. ozone getconf command should use the GenericCli parent class

amaliujia:

org.apache.hadoop.ozone.freon.OzoneGetCOnf implements a tool to print out current configuration values

With all the other CLI tools we already started to use picocli and the GenericCli parent class.

To provide better user experience we should migrate the tool to use GenericCli (+move it to the tools project + remove freon from the package name)

https://issues.apache.org/jira/browse/HDDS-3102

## How was this patch tested?

N/A


amaliujia:
@elek Any suggestion on how to test this change?

elek:
The easiest way IMHO to test cli is adding a robot test. 

A simple example is here:

https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/dist/src/main/smoketest/freon/freon.robot

But `Wait Until Keyword Succeeds      3min       10sec    ` is not required, just the `Should contain....`

you can easily test robot files with 
```
docker-compsoe exec scm bash
robot smoketest/basic/getconf.robot
```

And it can be added to one of the `test.sh` (eg. compose/ozone/test.sh) to be part of the CI tests.

amaliujia:
@elek 

Thanks for your help! I have added a robot test. Can you take a look please?

amaliujia:
note that because the `getconf.robot` is under `basic` directory, it will be picked up by `compose/ozone/test.sh`

amaliujia:
@elek friendly ping on this PR.

elek (approved):
+1, thanks the patch @amaliujia 

Tested locally and worked well.

PR #1411, opened 2020-09-09
HDDS-4557. [DESIGN] S3/Ozone Filesystem inter-op

elek:

A new design doc is included about S3/HCFS interoperability. Earlier it was discussed under https://issues.apache.org/jira/browse/HDDS-4097. 

But I created this PR as:

 1. I promised to do it to make it easier to include all the context specific comments
 2. Make it easier to follow the document specific changes / discussions

### DIFF
--- a/hadoop-hdds/docs/content/design/s3_hcfs.md
+++ b/hadoop-hdds/docs/content/design/s3_hcfs.md
@@ -64,7 +64,7 @@ This configuration is turned off by default, which means that S3 and HCFS couldn
 
 To solve the performance problems of the directory listing / rename, [HDDS-2939](https://issues.apache.org/jira/browse/HDDS-2939) is created, which propose to use a new prefix table to store the "directory" entries (=prefixes).
 
-[HDDS-4097](https://issues.apache.org/jira/browse/HDDS-4097) is created to normalize the key names based on file-system semantics if `ozone.om.enable.filesystem.paths` is enabled. But please note that `ozone.om.enable.filesystem.paths` should always be turned on if S3 and HCFS are both used which means that S3 and HCFS couldn't be used together with normalization.
+[HDDS-4097](https://issues.apache.org/jira/browse/HDDS-4097) is created to normalize the key names based on file-system semantics if `ozone.om.enable.filesystem.paths` is enabled. But please note that `ozone.om.enable.filesystem.paths` should always be turned on if S3 and HCFS are both used. It means that if both S3 and HCFS are used, normalization is forced, and S3 interface is not fully AWS S3 compatible. There is no option to use HCFS and S3 but with full AWS compatibility (and reduced HCFS compatibility). 
 
 # Goals
 
--- a/hadoop-hdds/docs/content/design/s3_hcfs.md
+++ b/hadoop-hdds/docs/content/design/s3_hcfs.md
@@ -2,7 +2,7 @@
 title: S3/Ozone Filesystem inter-op 
 summary: How to support both S3 and HCFS and the same time
 date: 2020-09-09
-jira: HDDS-4097
+jira: HDDS-4557
 status: accepted
 author: Marton Elek, 
 ---
@@ -28,7 +28,7 @@ Ozone is an object-store for Hadoop ecosystem which can be used from multiple in
  2. From S3 compatible applications (REST)
  3. From container orchestrator as mounted volume (CSI, alpha feature)
 
-As Ozone is an object store it stores key and values in a flat hierarchy which is enough to support S3 (2). But to support Hadoop Compatible File System (and CSI), Ozone should simulated file system hierarchy.
+As Ozone is an object store it stores key and values in a flat hierarchy which is enough to support S3 (2). But to support Hadoop Compatible File System (and CSI), Ozone should simulate file system hierarchy.
 
 There are multiple challenges when file system hierarchy is simulated by a flat namespace:
 
@@ -79,89 +79,85 @@ There are two main aspects of supporting both `ofs/o3fs` and `s3` together:
  1. `ofs/o3fs` require to create intermediate directory entries (for example `/a/b` for the key `/a/b/c`)
  2. Special file-system incompatible key names require special attention
 
-The second couldn't be done with compromise! For example if uploading a key with name '../../a.txt' can work with S3, but it's not possible by handled by `ofs/o3fs` as it would escape from the file system hierarchy.
+The second couldn't be done without compromise! For example if uploading a key with name `../../a.txt` can work with S3, but it's not possible by handled by `ofs/o3fs` as it would escape from the file system hierarchy.
 
 At high level, we have two options:
 
  1. We either support all key names (including non fs compatible key names), which means `ofs/o3fs` can provide only a partial view. Some invalid key names couldn't be visible. (But all keys can be uploaded from S3)
  2. Or we can normalize the key names to always be HCFS compatible (which would break the S3 compatibility for some specific key names. For example a file (!) uploaded with name `a/b/` might be visible as `/a/b` after the creation after the normalization required by the `ofs/o3fs`).
 
-HDDS-3955 introduced `ozone.om.enable.filesystem.paths`, with this setting we will have two possible usage pattern:
+As a result we can have three main options:
 
-| ozone.om.enable.filesystem.paths= | true | false
-|-|-|-|
-| create intermediate dirs | YES | NO |
-| normalize key names from `ofs/o3fs` | YES | NO
-| force to normalize key names of `s3` interface | YES | NO 
-| `s3` key `/a/b/c` available from `ofs/o3fs` | YES | NO
-| `s3` key `/a/b//c` available from `ofs/o3fs` | YES | NO
-| `s3` key `/a/b//c` available from `s3` | AWS S3 incompatibility | YES
+| option | prefix entries | normalization | S3 behavior | HCFS behavior | status |
+|-|-|-|-|-|-|
+| default | DISABLED | DISABLED | 100 % compatibility | FULLY UNSUPPORTED |  implemented, default |
+| ozone.om.enable.filesystem.paths=true | ENABLED | ENABLED | limited compatibility | FULLY SUPPORTED | implemented |
+| ozone.om.enable.intermediate.dirs=true | ENABLED | DISABLED | 100% compatibility | SUPPORTED but partial view. | NOT implemented |
 
-This proposal suggest to use a 3rd option where 100% AWS compatibility is guaranteed in exchange of a limited `ofs/o3fs` view:
+Let's talk all of these option in more details
 
-| ozone.om.intermediate.dir.generation= | true |
-|-|-|-|
-| create intermediate dirs | YES | 
-| normalize key names from `ofs/o3fs` | YES |
-| force to normalize key names of `s3` interface | **NO** |
-| `s3` key `/a/b/c` available from `ofs/o3fs` | YES | 
-| `s3` key `/a/b//c` available from `ofs/o3fs` | NO | 
-| `s3` key `/a/b//c` available from `s3` | *YES* (100% AWS compatibility) |
+## Pure S3 Object store without Hadoop support
 
-
-# Proposed solution
-
-In short: 
-
- **I propose to make it possible to configure **normalization** and **intermediate dir creation**, independent from each other**
- 
-It can be done in multiple ways. For the sake of simplicity, let's imagine two configuration option
-
-| configuration | behavior | 
+| function | status |
 |-|-|
-| `ozone.om.enable.filesystem.paths=true`  | Enable intermediate dir generation **AND** key name normalization 
-| `ozone.om.enable.intermediate.dirs=true` | Enable only the intermediate dir generation
+| normalization | DISABLED |
+| prefix entries | DISABLED |
+| S3 compatibility | 100% |
+| HCFS compatibility | DISABLED |
 
-## Priority 
+The first version is the most simple. Only the S3 API is supported but full S3 compatibility can be exacted. (For example `a//b` key can be uploaded and downloaded).
 
-Current Ozone `master` focuses to the implementation of `ozone.om.enable.filesystem.paths`. There is a discussion of the priority of the implementation of the 3rd option  (`ozone.om.enable.intermediate.dirs`) and the final implementation depends on the business requirement.
+With this option `o3fs/ofs` couldn't be supported as the prefix entries are not created for the keys / objects. 
 
-## S3/HCFS Interoperability
+**As of today this is the default behavior of Apache Ozone.**
 
-**In case of intermediate directory generation is enabled (with either of the configuraiton keys)**:
+## Hadoop Object store with limited S3 support
 
-When somebody creates a new key like `/a/b/c/d`, the same key should be visible from HCFS (`o3fs//` or `o3://`). `/a`, `/a/b` and `/a/b/c` should be visible as directories from HCFS.
+| function | status |
+|-|-|
+| normalization | ENABLED |
+| prefix entries | ENABLED |
+| S3 compatibility | 100% |
+| HCFS compatibility | DISABLED |
 
-S3 should list only the `/a/b/c/d` keys, (`/a`, `/a/b`, `/a/b/c` keys, created to help HCFS, **won't be visible** if the key is created from S3)
+This mode is introduced in HDDS-3955 with creating `ozone.om.enable.filesystem.paths`. This setting enabled both normalization and prefix entry creation. As normalization is forced, some of the behavior is S3 in-compatible. For example key uploaded with the name `/abs/d//f` can be read from the path `/abs/d/f`. Also some path which may be valid for S3 (like `../../a`) are disabled. 
 
-This can be done with persisting an extra flag with the implicit directory entries. These entries can be modified if they are explicit created.
+This mode is supported and implemented in Ozone. The implementation details can be found under [HDDS-4097](https://issues.apache.org/jira/browse/HDDS-4097)
 
-This flag should be added only for the keys which are created by S3. `ofs://` and `of3fs://`  create explicit directories all the time.
+## S3 object store with limited Hadoop support
 
-Advantages of this approach:
+| function | status |
+|-|-|
+| normalization | ENABLED |
+| prefix entries | DISABLED |
+| S3 compatibility | 100% |
+| HCFS compatibility | ENABLED but partial |
 
- 1. HCFS and S3 can work together
- 2. S3 behavior is closer to the original AWS s3 behavior (when `/a/b/c` key is created `/a/b` won't be visible)
+**This mode is not yet implemented**. Depends from the user feedback and requirements can be implemented as a hybrid mode.
 
-## Handling of the incompatible paths
+This approach makes the normalization and prefix creation independent from each other. A new configuration can be introduced to turn on *only* the prefix creation.
 
-As it's defined above the intermediate directory generation and normalization are two independent settings. (It's possible to choose only to create the intermediate directories).
+| configuration | behavior |
+|-|-|
+| `ozone.om.enable.filesystem.paths=true`  | Enable intermediate dir generation **AND** key name normalization |
+| `ozone.om.enable.intermediate.dirs=true` | Enable only the intermediate dir generation |
 
-**If normalization is choosen**: (`ozone.om.enable.filesystem.paths=true`), all the key names will be normalized to fs-compatible name. It may cause a conflict (error) if the normalized key is already exists (or exists as a file instead of directory)
+This mode can be 100% S3 compatibility (no forced normalization). But some of the path which are saved via S3 (like `a/b//////v`) couldn't be displayed via `ofs/o3fs`. This is the same behavior what S3A provides when incompatible path is found in S3.
 
-**Without normalization (`ozone.om.enable.intermediate.dirs=true`)**:
-
-Creating intermediate directories might not be possible if path contains illegal characters or can't be parsed as a file system path. **These keys will be invisible from HCFS** by default. They will be ignored during the normal file list.
-
-## Using Ozone in object-store only mode
-
-Creating intermediate directories can have some overhead (write amplification is increased if many keys are written with different prefixes as we need an entry for each prefixes). This write-amplification can be handled with the current implementation: based on the measurements RocksDB has no problems with billions of keys.
-
-If none of the mentioned configurations are enabled, the intermediate directories won't be created. But in this case, the consistent view of `ofs/o3fs` couldn't be guaranteed, so `ofs/o3fs` **should be disabled and throw an exception** (But Ozone can be used as a pure S3 replacement without using as a HCFS).
+With this approach `ofs/o3fs` can show only a partial view 
 
 # Problematic cases
 
 As described in the previous section there are some cases which couldn't be supported out-of-the-box due to the differences between the flat key-space and file-system hierarchy. These cases are collected here together with the information how existing tools (AWS console, AWS cli, AWS S3A Hadoop connector) behaves.
+
+
+| name | configuration | status |
+|-|-|-|
+| S3 mode | (default) | implemented, default |
+| Hadoop mode | `ozone.om.enable.filesystem.paths=true`  | implemented, can be configured |
+| Hybrid mode | `ozone.om.enable.intermediate.dirs=true` | **not implemented** (!) |
+
+Note: S3 functionality can be used together with Hadoop mode, but this is not 100% compatibility as some normalization are forced and validation is more strict.
 
 ## Empty directory path
 
@@ -175,12 +171,13 @@ Behavior:
  * *S3 web console*: empty dir is rendered as `____` to make it possible to navigate in
  * **aws s3 ls**: Prefix entry is visible as `PRE /`
  * S3A: Not visible
- 
+
 Proposed behavior:
 
- * `ozone.om.enable.intermediate.dirs=true`: `/y` is not accessible, `/a/b` directory doesn't contain this entry
- * `ozone.om.enable.filesystem.paths=true`: key stored as `/a/b/c`  
- 
+ * Hybrid mode: `/y` is not accessible, `/a/b` directory doesn't contain this entry
+ * Hadoop mode: key stored as `/a/b/c`  
+ * S3 mode: key is used as is, but not available for `o3fs`/`ofs`.
+
 ## Path with invalid characters (`..`,`.`) 
 
 Path segments might include parts which has file system semantics:
@@ -194,11 +191,12 @@ Behavior:
  * *S3 web console*: `.` and `..` are rendered as directories to make it possible to navigate in
  * **aws s3 ls**: Prefix entry is visible as `PRE ../` and `PRE ./`
  * S3A: Entries are not visible
- 
+
 Proposed behavior:
 
- * `ozone.om.enable.intermediate.dirs=true`: `e` and `f` are not visible
- * `ozone.om.enable.filesystem.paths=true`: key stored as `/a/e` and `a/b/f`  
+ * Hybrid mode: `e` and `f` are not visible from `ofs` and `o3fs`, raw key path is used as defined (including dots)
+ * Hadoop mode: key are stored as `/a/e` and `a/b/f`  
+ * S3 mode: keys are stored as is (`a/b/../e`/` a/b/./f`)
 
 ## Key and directory with the same name
 
@@ -213,11 +211,12 @@ Behavior:
  * *S3 web console*: both directory and file are rendered
  * **aws s3 ls**: prefix (`PRE h/`) and file (`h`) are both displayed
  * S3A: both entries are visible with the name `/a/b/h` but firt is a file (with size) second is a directory (with directory attributes)
- 
+
 Proposed behavior:
 
- * `ozone.om.enable.intermediate.dirs=true`: show both the file and the directory with the same name (similar to S3A)
- * `ozone.om.enable.filesystem.paths=true`: throwing exception when the second one is created  
+ * Hybrid mode: show both the file and the directory with the same name (similar to S3A)
+ * Hadoop mode: throwing exception when the second one is created  
+ * S3 mode: both keys are saved, reading from `ofs`/`o3fs` is not supported 
 
 ## Directory entry created with file content
 
@@ -231,11 +230,12 @@ Behavior:
  * *S3 web console*: rendered as directory (couldn't be downloaded)
  * **aws s3 ls**: showed as a prefix (`aws s3 ls s3://ozonetest/a/b`), but when the full path is used showed as a file without name (`aws s3 ls s3://ozonetest/a/b/i/`)
  * S3A: `./bin/hdfs dfs -ls s3a://ozonetest/a/b/` shows a directory `h`, `./bin/hdfs dfs -ls s3a://ozonetest/a/b/i` shows a file `i`
- 
+
 Proposed behavior:
 
- * `ozone.om.enable.intermediate.dirs=true`: possible but `i/` is hidden from o3fs/ofs
- * `ozone.om.enable.filesystem.paths=true`: key name is normalized to real key name
+ * Hybrid mode: possible but `i/` is hidden from o3fs/ofs
+ * Hadoop mode: key name is normalized to real key name
+ * Key is uploaded as is, without normalization. Using `ofs`/`o3fs` is not supported.
 
 ## Create key and explicit create parent dir
 
@@ -247,10 +247,10 @@ aws s3api put-object --bucket ozonetest --key e/f/
 Behavior:
 
  * S3 can support it without any problem
- 
+
 Proposed behavior:
 
-After the first command `/e/f/` and `/e/` entries created in the key space (as they are required by `ofs`/`o3fs`) but **with a specific flag** (explicit=false). 
+Hybrid mode: After the first command `/e/f/` and `/e/` entries created in the key space (as they are required by `ofs`/`o3fs`) but **with a specific flag** (explicit=false). 
 
 AWS S3 list-objects API should exclude those entries from the result (!).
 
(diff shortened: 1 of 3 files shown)


elek:
Opened as a **DRAFT** pull request as this is only a proposal. 

@arp7 and @bharatviswa504 still have concerns about the proposed approache.

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -67,45 +66,100 @@ To solve the performance problems of the directory listing / rename, [HDDS-2939]
...
+| ozone.om.enable.filesystem.paths= | true | false
+|-|-|-|
+| create itermediate dirs | YES | NO |
+| normalize key names from `ofs/o3fs` | YES | NO
+| force to normalize key names of `s3` interface | YES (1) | NO 
+| `s3` key `/a/b/c` available from `ofs/o3fs` | YES | NO
+| `s3` key `/a/b//c` available from `ofs/o3fs` | YES | NO
+| `s3` key `/a/b//c` available from `s3` | AWS S3 incompatibility | YES
+
+(1): Under implementation
```
arp7, line 98: I don't think that is true. Paths are normalized already on the S3 interface when writing new keys.

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -67,45 +66,100 @@ To solve the performance problems of the directory listing / rename, [HDDS-2939]
...
 
- * Out of the box Ozone should support both S3 and HCFS interfaces without any settings. (It's possible only for the regular path)
- * As 100% compatibility couldn't be achieved on both side we need a configuration to set the expectations in case of incompatible key names
- * Default behavior of `o3fs` and `ofs` should be as close to `s3a` as possible
+This proposal suggest to use a 3rd option where 100% AWS compatiblity is guaranteed in exchange of a limited `ofs/o3fs` view:
```
arp7, line 100: I completely disagree with this trade-off. The FS limited view is neither here nor there. You can insert keys via the S3 interface that are not visible via the FS view at all. To me this is the same as a corrupted filesystem. Marton, I liked your offline suggestion much better - disable FS access completely when operating in S3-compatible mode.

Taking this one step further, I have a different approach in mind. Let's make this a per-bucket setting. For buckets created via the S3 interface, by default the S3 semantics will be preserved 100% unless the global setting is enabled and FS access will not be allowed at all. For buckets created via FS interface, the FS semantics will always take precedence. If the global setting is enabled, then the value of the setting at the time of bucket creation is sampled and that takes effect for the lifetime of the bucket. Basically you can't change the behavior for a given bucket.

arp7:
I have an alternate proposal, idea left in a comment. cc @bharatviswa504 @elek

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+ls: `o3fs://bucket1.s3v/a/': No such file or directory
+```
+
+This problem is reported in [HDDS-3955](https://issues.apache.org/jira/browse/HDDS-3955), where a new configuration key is introduced (`ozone.om.enable.filesystem.paths`). If this is enabled, intermediate directories are created even if the object store interface is used.
+
+This configuration is turned off by default, which means that S3 and HCFS couldn't be used together.
+
+To solve the performance problems of the directory listing / rename, [HDDS-2939](https://issues.apache.org/jira/browse/HDDS-2939) is created, which propose to use a new prefix table to store the "directory" entries (=prefixes).
+
+[HDDS-4097](https://issues.apache.org/jira/browse/HDDS-4097) is created to normalize the key names based on file-system semantics if `ozone.om.enable.filesystem.paths` is enabled. But please note that `ozone.om.enable.filesystem.paths` should always be turned on if S3 and HCFS are both used which means that S3 and HCFS couldn't be used together with normalization.
```
bharatviswa504, line 67: S3 and HCFS couldn't be used together **without** normalization??

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+
+ * Out of the box Ozone should support both S3 and HCFS interfaces without any settings. (It's possible only for the regular, fs compatible key names)
+ * As 100% compatibility couldn't be achieved on both side we need a configuration to set the expectations for incompatible key names
+ * Default behavior of `o3fs` and `ofs` should be as close to `s3a` as possible (when s3 compatibilty is prefered)
+
+# Possible cases to support
+
+There are two main aspects of supporting both `ofs/o3fs` and `s3` together:
+
+ 1. `ofs/o3fs` require to create intermediate directory entries (for exapmle `/a/b` for the key `/b/c/c`)
```
bharatviswa504, line 79: exapmle -> example

hadoop-hdds/docs/content/design/s3_hcfs.md
bharatviswa504: Here AWS S3 incompatibility means, is it because we are showing normalized keys?

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+ 1. We either support all key names (including non fs compatible key names), which means `ofs/o3fs` can provide only a partial view
+ 2. Or we can normalize the key names to be fs compatible (which makes it possible to create inconsistent S3 keys)
+
+HDDS-3955 introduced `ozone.om.enable.filesystem.paths`, with this setting we will have two possible usage pattern:
+
+| ozone.om.enable.filesystem.paths= | true | false
+|-|-|-|
+| create itermediate dirs | YES | NO |
+| normalize key names from `ofs/o3fs` | YES | NO
+| force to normalize key names of `s3` interface | YES (1) | NO
```
bharatviswa504, line 93: There is no such force flag, if normalize is enable, we normalize from all interfaces.
As We have 4 interfaces.
1. CLI
2. S3
3. HCFS
4. Java Native Client.

So, this flag is not some thing special for S3.

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+| `s3` key `/a/b//c` available from `s3` | AWS S3 incompatibility | YES
+
+(1): Under implementation
+
+This proposal suggest to use a 3rd option where 100% AWS compatiblity is guaranteed in exchange of a limited `ofs/o3fs` view:
+
+| ozone.om.intermediate.dir.generation= | true |
+|-|-|-|
+| create itermediate dirs | YES | 
+| normalize key names from `ofs/o3fs` | YES |
```
bharatviswa504, line 105: There is no special normalize is needed, as we use Path Object, keyNames are normalized and sent to OM

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+
+In short: 
+
+ **I propose to make it possible to configure **normalization** and **intermediate dir creation**, independent from each other**
+ 
+It can be done in multiple ways. For the sake of simplicity, let's imagine two configuration option
+
+| configuration | behavior | 
+|-|-|
+| `ozone.om.enable.filesystem.paths=true`  | Enable intermediate dir generation **AND** key name normalization
```
bharatviswa504, line 122: so even if the other setting is false(ozone.om.enable.intermediate.dirs = false), we create intermediate directories?

bharatviswa504, line 122: Question: Is this config still OM config?

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+
+**In case of intermediate directory generation is enabled (with either of the configuraiton keys)**:
+
+When somebody creates a new key like `/a/b/c/d`, the same key should be visible from HCFS (`o3fs//` or `o3://`). `/a`, `/a/b` and `/a/b/c` should be visible as directories from HCFS.
+
+S3 should list only the `/a/b/c/d` keys, (`/a`, `/a/b`, `/a/b/c` keys, created to help HCFS, **won't be visible** if the key is created from S3)
+
+This can be done with persisting an extra flag with the implicit directory entries. These entries can be modified if they are explicit created.
+
+This flag should be added only for the keys which are created by S3. `ofs://` and `of3fs://`  create explicit directories all the time.
```
bharatviswa504, line 135: These entries can be modified if they are explicit created.
Can you explain a little more about this statement?

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+Behavior:
+ * *S3 web console*: empty dir is rendered as `____` to make it possible to navigate in
+ * **aws s3 ls**: Prefix entry is visible as `PRE /`
+ * S3A: Not visible
+ 
+Proposed behavior:
+
+ * `ozone.om.enable.intermediate.dirs=true`: `/y` is not accessible, `/a/b` directory doesn't contain this entry
```
bharatviswa504, line 177: `/y` is not accessible from FS, but will be accessible from S3 correct?

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+ * `ozone.om.enable.filesystem.paths=true`: key stored as `/a/b/c`
```
bharatviswa504, line 178: With this, visible from both, but in S3 it will be shown with normalized name?

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+
+Behavior:
+ * *S3 web console*: `.` and `..` are rendered as directories to make it possible to navigate in
+ * **aws s3 ls**: Prefix entry is visible as `PRE ../` and `PRE ./`
+ * S3A: Entries are not visible
+ 
+Proposed behavior:
+
+ * `ozone.om.enable.intermediate.dirs=true`: `e` and `f` are not visible
+ * `ozone.om.enable.filesystem.paths=true`: key stored as `/a/e` and `a/b/f`
```
bharatviswa504, line 197: Can we also add info on visibility and how it will be displayed also?

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+
+ * `ozone.om.enable.intermediate.dirs=true`: `e` and `f` are not visible
+ * `ozone.om.enable.filesystem.paths=true`: key stored as `/a/e` and `a/b/f`  
+
+## Key and directory with the same name
+
+It is possible to create directory and key with the same name in AWS:
+
+```
+aws s3api put-object --bucket ozonetest --key a/b/h --body README.md
```
bharatviswa504, line 204: aws s3api put-object --bucket ozonetest --key a/b/ --body README.md
Need one more put above

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+
+```
+hdfs dfs -mkdir -p s3a://b12345/d11/d12 # -> Success
+
+hdfs dfs -put /tmp/file1 s3a://b12345/d11/d12/file1 # -> fails with below error
+```
+
+Proposed behavior:
+
+ * `ozone.om.enable.intermediate.dirs=true`: should work without error
```
bharatviswa504, line 267: Will not work, as createDirectory is putObject with 0 byte.
Refer [link](https://issues.apache.org/jira/browse/HDDS-4209?focusedCommentId=17193033&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17193033)

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+ * `ozone.om.enable.filesystem.paths=true`: should work without error.
+
+This is an `ofs`/`o3fs` question not an S3. The directory created in the first step shouldn't block the creation of the file. This can be a **mandatory** normalization for `mkdir` directory creation. As it's an HCFS operation, s3 is not affected. Entries created from S3 can be visible from s3 without any problem.
```
bharatviswa504, line 270: It will block ofs/o3fs also, as it is not a directory in ozone, it is a file.

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,282 @@
...
+
+This problem is reported in HDDS-4209, thanks to @Bharat
+
+```
+hdfs dfs -mkdir -p s3a://b12345/d11/d12 -> Success
+
+hdfs dfs -put /tmp/file1 s3a://b12345/d11/d12/file1 
+```
+
+In this case first a `d11/d12/` key is created. The intermediate key creation logic in the second step should use it as a directory instead of throwing an exception.
```
bharatviswa504, line 282: d11/d12 is created without traiiling "/"

bharatviswa504 (commented):
Thank You @elek for the design document.

My understanding from this is the draft is as below. Let me know if I am missing something here.
Screen Shot 2020-09-09 at 10 59 11 AM

hadoop-hdds/docs/content/design/s3_hcfs.md
bharatviswa504: And do we need this **AND key name normalization** as anyway, we have another config to define this behavior, which can be handled from code, instead of giving some additional meaning to the config?

hadoop-hdds/docs/content/design/s3_hcfs.md
bharatviswa504: We also have another way right, an existing bucket can be exposed created via CLI to be exposed to S3, what semantics that bucket will get?

>For buckets created via FS interface, the FS semantics will always take precedence
Buckets creation is possible via only OFS, what about O3fs?

>If the global setting is enabled, then the value of the setting at the time of bucket creation is sampled and that takes >effect for the lifetime of the bucket.

A bucket created via Shell, when global flag (assuming ozone.om.enable.filesystem.paths=true), they will follow FS semantics and with slight S3 incompatibility.
So, a bucket created via Shell, when global flag (assuming ozone.om.enable.filesystem.paths=false), they will follow S3 semantics and with broken FS semantics or completely disallow.

Written from my understanding, as I have not got the complete context of the proposal.

arp7: bq. an existing bucket can be exposed created via CLI to be exposed to S3, what semantics that bucket will get?
If the bucket was created via FS interface, it will support FS semantics.

bq. Buckets creation is possible via only OFS, what about O3fs?
Good point, for buckets created via the Ozone shell, we could accept a command-line flag. The default can be filesystem because S3 buckets are traditionally created via the S3 API. You're right this needs some more discussion.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: > Here AWS S3 incompatibility means, is it because we are showing normalized keys?

Yes, keys are normalized. Content can be found under different key names.

I started to define the 100% compatibility here:

https://github.com/elek/hadoop-ozone/blob/s3-compat/hadoop-ozone/dist/src/main/smoketest/s3/s3-vs-filepath.robot

elek: > I don't think that is true. Paths are normalized already on the S3 interface when writing new keys.

But not for read, if I understood well. But happy to remove this line if it's confusing.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Different behavior on bucket level seems to be an interesting idea.

> For buckets created via FS interface, the FS semantics will always take precedence. 

How would you define the behavior of bucket is created from S3? 

I suppose in this case we should support 100% AWS S3 compatibility (without forced normalization).

But how would o3fs/ofs work in case of `s3` buckets:

 1. Partial view from ofs (incompatible keys are hidden)
 2. `ofs/o3fs` is disabled (exception), no intermediate directories are created.

elek: Also, one disadvantage: bucket level settings have increased complexity. It's harder to define the expected behavior for a specific path. Cluster level settings is easie, as there is one global behavior for the setup.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Yes, thanks. I also clarified this paragraph a little:

> But please note that `ozone.om.enable.filesystem.paths` should always be turned on if S3 and HCFS are both used. It means that if both S3 and HCFS are used, normalization is forced, and S3 interface is not fully AWS S3 compatible. There is no option to use HCFS and S3 but with full AWS compatibility (and reduced HCFS compatibility).

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Yes, agree. I tried to explain this behavior with the two lines to show that both are normalized. This is not because we have two flags.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Yes. I just tried to show here this out-of-the-box behavior. But agree. 

We have it for free and actually this couldn't be removed: if we use ofs/o3fs, the keys *from* ofs/o3fs should always be normalized.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: I am open to reorganize the configuration in any way. In this model the two configuration are independent. Existing config means CREATE_DIR+NORMALIZE, new config is just CREATE_DIR.

They are independent and not matrix. 

But we can switch back to the original version.

elek: I think so, but we can remove `om`. As it's not required to know from a config where is it used. It's an important Ozone config. But I am open to use any config name.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: If any key is created from s3 interface but it already exists but explicit=false flag, the flag should be changed to explicit=true, to make it visible from s3.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: yes. exactly

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Yes, exactly.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Sure, I added.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Why is it required? I think the current example defines the behavior when only s3 part is used.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: I tested it locally and found the trailing `/`. Are you sure?

elek:
> Thank You @elek for the design document.
> 
> My understanding from this is the draft is as below. Let me know if I am missing something here.
> Screen Shot 2020-09-09 at 10 59 11 AM

Correct. But this is not a matrix anymore. You should turn on either first or second of the configs, but not both.

hadoop-hdds/docs/content/design/s3_hcfs.md
bharatviswa504: Yes. Posted my comment in HDDS-4209. Refer for more info

bharatviswa504:
> > Thank You @elek for the design document.
> > My understanding from this is the draft is as below. Let me know if I am missing something here.
> > Screen Shot 2020-09-09 at 10 59 11 AM
> 
> Correct. But this is not a matrix anymore. You should turn on either first or second of the configs, but not both.

Not sure what is meant here, because we have 2 configs, now we can have 4 combinations according to proposal 3 are valid, 4th one is not.

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,280 @@
...
+
+ * Out of the box Ozone should support both S3 and HCFS interfaces without any settings. (It's possible only for the regular, fs compatible key names)
+ * As 100% compatibility couldn't be achieved on both side we need a configuration to set the expectations for incompatible key names
+ * Default behavior of `o3fs` and `ofs` should be as close to `s3a` as possible (when s3 compatibilty is prefered)
+
+# Possible cases to support
+
+There are two main aspects of supporting both `ofs/o3fs` and `s3` together:
+
+ 1. `ofs/o3fs` require to create intermediate directory entries (for example `/a/b` for the key `/b/c/c`)
```
xiaoyuyao, line 79: /b/c/c => /a/b/c

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,280 @@
...
+HDDS-3955 introduced `ozone.om.enable.filesystem.paths`, with this setting we will have two possible usage pattern:
+
+| ozone.om.enable.filesystem.paths= | true | false
+|-|-|-|
+| create itermediate dirs | YES | NO |
+| normalize key names from `ofs/o3fs` | YES | NO
+| force to normalize key names of `s3` interface | YES | NO 
+| `s3` key `/a/b/c` available from `ofs/o3fs` | YES | NO
+| `s3` key `/a/b//c` available from `ofs/o3fs` | YES | NO
+| `s3` key `/a/b//c` available from `s3` | AWS S3 incompatibility | YES
```
xiaoyuyao, line 96: Any pointer to S3 compatibility requirement? e.g., path handling, normalization, etc.

hadoop-hdds/docs/content/design/s3_hcfs.md
xiaoyuyao: bq. Existing config means CREATE_DIR+NORMALIZE, new config is just CREATE_DIR.

Are these the only two cases that are useful? do we need to support other combinations?

hadoop-hdds/docs/content/design/s3_hcfs.md
xiaoyuyao: bq. S3 should list only the `/a/b/c/d` keys, (`/a`, `/a/b`, `/a/b/c` keys, created to help HCFS, **won't be visible** if the key is created from S3)

what if user create additional files under intermediate dirs such as /a/b/d via FS interface, we still want to show them in this case for interop?

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Handling of the incompatible paths
+
+As it's defined above the intermediate directory generation and normalization are two independent settings. (It's possible to choose only to create the intermediate directories).
+
+**If normalization is choosen**: (`ozone.om.enable.filesystem.paths=true`), all the key names will be normalized to fs-compatible name. It may cause a conflict (error) if the normalized key is already exists (or exists as a file instead of directory)
+
+**Without normalization (`ozone.om.enable.intermediate.dirs=true`)**:
+
+Creating intermediate directories might not be possible if path contains illegal characters or can't be parsed as a file system path. **These keys will be invisible from HCFS** by default. They will be ignored during the normal file list.
```
xiaoyuyao, line 148: The illegal char from S3 path can be encoded into FS path except the /.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Good question, there are three useful cases IMHO:

 1. Do nothing (neither normalize, nor create dir) --> Ozone can be used only from S3, ofs/o3fs is inconsistent due to the missing intermediate directories (I suggest to throw exception in this case for ofs/o3fs)
 2. create dirs, don't normalize --> 100% AWS compatibility, partial view from ofs (invalid keys not visible from ofs. eg `/a/b/c////d`)
 3. create dirs, normalize --> reduced AWS s3 compatibility, full ofs view

The last option is: normalize but don't create dirs --> it doesn't make sense, as without creating intermediate dirs, ofs is not usable, therefore we don't need normalization.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: The compatibility story is only important when somebody uses `s3` interface. S3 gateway expected to be compatible with AWS S3.

But there is no rule limtations about showing something for S3G which is created from ofs/o3fs. From this point of view ofs/o3fs are external clients which can do anything, there couldn't be any excepttations in s3g.

 * Directories created by ofs/o3fs can be shown from s3 all the time
 * During key S3 creation intermediate directories should be created but (by default) hidden to S3
 * Unless somebody creates them with ofs interface. In those case they can be visible.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Thanks to explain it. I tested it with pure AWS S3 and `s3a`. The trailing `/` seems to be missing due to a bug in our normalization, and it's not and `s3a` behavior. I think we should fix the normalization.

elek:
> > > Thank You @elek for the design document.
> > > My understanding from this is the draft is as below. Let me know if I am missing something here.
> > > Screen Shot 2020-09-09 at 10 59 11 AM
> > 
> > 
> > Correct. But this is not a matrix anymore. You should turn on either first or second of the configs, but not both.
> 
> Not sure what is meant here, because we have 2 configs, now we can have 4 combinations according to proposal 3 are valid, 4th one is not.

Agree, but there are two ways to define this 3 options:

1st approach:

KEY1=true,KEY2=true --> option1
KEY1=false,KEY2=false --> option2
KEY1=true,KEY2=false --> option3
KEY1=false,KEY2=true --> invalid 

2nd approach:

KEY1=true --> option 1
KEY2=true --> option2
else --> option3

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Good catch, thanks, fixed.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: In general, AWS S3 can save any key without normalization, and the same content can be retrieved with the same key (and list shows the raw names).

`AWS S3 incompatibility`: means partial incompatibility. All the `s3` tools which depends on the original AWS S3 behavior can be failed. (For example if a S3 Fuse file system creates dirs with keys ending with `/` **AND** store real directory metadata on that specific key --> it will be broken).

I started to create robot tests for these cases which can be used to check this side of the compatibility:

https://github.com/elek/hadoop-ozone/commit/d21890bb0b0d08d0f8c631e9f24af4745b2f3aca

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: `..` also can be tricky.

But in general, I agree. We might be more permissive and show some of these elements even in `ofs/o3fs`. For example `a/b/c//d` might be possible to show as `/a/b/c/d`. But even if some of these are visible, we should accept that `ofs/o3fs` doesn't provide a *full* view when 100% AWS compatibility is requested.

elek:
We had an offline conversation with @bharatviswa504 @arp7

Got the feedback from Arpit: the 3rd option can be useful (we had disagreement how useful it is), but it was requested to fill the behavior in `Ozone filesystem path enabled` attached to the design doc.

I uploaded the updated version: `Ozone filesystem path enabled v3.xlsx` to the design doc.

@arp7 Would you be so kind to have a look and give me feedback.

elek:
All the questions answered here and didn't receive any new objection in the last month. I am planning to merge it if no more objections.

This document describes 3 possible compatibility options and we had some debate if the 3rd option (Full s3 compatibility but limited o3fs/ofs view) is important or not. (It was important for me but others said it's not a real use case).

As this document describes the expected behavior very well for case 1 and case 2 I prefer to commit it. I added one more paragraph (see #d24322ec68ca3c49181c201d1f41d18dd2b60f7f) which clarifies that the priority / implementation of the 3rd option depends on business need, and not included / and not under implementation on master. (If somebody needs it, we can show that we have a plan, and we can discuss the implementation).

I hope it's an acceptable consensus.

@arp7 @bharatviswa504 @mukul1987 @sidseth

arp7:
> All the questions answered here and didn't receive any new objection in the last month. I am planning to merge it if no more objections.

Please don't proceed without a binding +1.

For options 1 and 2 - how is this design doc different from the one attached to HDDS-4097?

elek:
> Please don't proceed without a binding +1.

This is a design doc, I would prefer to have a consensus not just one +1.

> For options 1 and 2 - how is this design doc different from the one attached to HDDS-4097?

Can you please help me with explaining the difference.  As far as I understood we agreed that it's the same. (It was not true with the first version but the handling of properties have been change since then).

arp7:
If options 1 and 2 are the same, why not just checkin the original design
doc and explain option 3 as a potential alternative?

elek:
> If options 1 and 2 are the same, why not just checkin the original design doc and explain option 3 as a potential alternative?

We can definitely check in the original design doc, too. I am +1 for that PR.

This design doc ... 
 1. ... explains options 1 and 2 in more details (I think it's better to have more information, description from different angles)
 2. ... introduces a new possible option in the same model (which is not an alternative but an additional possible feature which may not be required right now)

hadoop-hdds/docs/content/design/s3_hcfs.md
```diff
@@ -0,0 +1,284 @@
...
+
+To solve the performance problems of the directory listing / rename, [HDDS-2939](https://issues.apache.org/jira/browse/HDDS-2939) is created, which propose to use a new prefix table to store the "directory" entries (=prefixes).
+
+[HDDS-4097](https://issues.apache.org/jira/browse/HDDS-4097) is created to normalize the key names based on file-system semantics if `ozone.om.enable.filesystem.paths` is enabled. But please note that `ozone.om.enable.filesystem.paths` should always be turned on if S3 and HCFS are both used. It means that if both S3 and HCFS are used, normalization is forced, and S3 interface is not fully AWS S3 compatible. There is no option to use HCFS and S3 but with full AWS compatibility (and reduced HCFS compatibility). 
+
+# Goals
+
+ * Out of the box Ozone should support both S3 and HCFS interfaces without any settings. (It's possible only for the regular, fs compatible key names)
+ * As 100% compatibility couldn't be achieved on both side we need a configuration to set the expectations for incompatible key names
+ * Default behavior of `o3fs` and `ofs` should be as close to `s3a` as possible (when s3 compatibilty is prefered)
```
arp7, line 73: The tradeoff here is that in this mode the FS integrity can be affected by putting invalid keys via S3. E.g. i could put key names which have elements like `.` and `..`, or key names which try to escape the root of the filesystem. I think the tradeoffs should be called out else this doc is not representing the options correctly.

arp7:
I see multiple comments on the doc. Are all the discussion threads resolved?

elek:
> I see multiple comments on the doc. Are all the discussion threads resolved?

In which doc? As far as I know all the questions are answered. Please let me know If I missed something.

hadoop-hdds/docs/content/design/s3_hcfs.md
elek: Thanks the suggestion. I added a few more sentences to make it more clear. See: bd7792f commit (or the [rendered view](https://github.com/apache/ozone/commit/bd7792f05cbebe7b56738fd487a56d2ea5909361?short_path=8cd4ef7#diff-8cd4ef70bb6f221e7352b5f4e8e5a77233222403a88f02b2305ad37b88f56a69)).

elek:
@arp7 @bharatviswa504  Do you have any more comments? Can we merge it?

elek:
@arp7 @bharatviswa504 Do you have any more comments? Can we merge it?

@arp7 suggested offline merging this doc with other docs under HDDS-4097. I am not sure if it's required. All the docs under HDDS-4097 talks about the *implementation details* of one of the options described here (filesystem path enabled).

@bharatviswa504 Let me know if you have different opinion.

elek:
@bharatviswa504 Are you fine with merging this PR?

I think it's good to have a global picture even if the 3rd option is not a priority right now.

elek:
The doc is updated based on the agreements. It describes the three options, but the proposed new option is mentioned only as an option which can be implemented in case of real business demand:

image

@arp7 / @sidseth please let me know what do you think (This approach is already discussed with @mukul1987 and will be discussed with Bharat, soon).

elek:
Ping @arp7 @bharatviswa504 @mukul1987 @sidseth 

What is your suggestion about this PR?

bharatviswa504:
@elek 
I am fine with this, one concern I have is but once HDDS-2939 comes in, this doc might need changes, as it has old representation of key format, in V1 version the key is represented with objectID/name. So, do we want to hold off on this?

Because V1 representation is 100% FS as it will always create intermediate dir irrespective of the normalization flag.

@rakeshadr for any comments.

elek:
> I am fine with this, one concern I have is but once HDDS-2939 comes in, this doc might need changes, as it has old representation of key format, in V1 version the key is represented with objectID/name. So, do we want to hold off on this?

This is a pull requests to the master. I think we don't need to hold off, we can commit it as it represents the state of the master. 

But it's a good idea to modify it with HDDS-2939 and update in case of any changes on master (HDDS-2939 merge)

elek:
Closing this one due to lack of support.

I am tired to ping people for answers again and again and somewhat disappointed that design conversations are so hard....

:-(

Still I am strongly believe that the S3 / HCFS interoperability guarantees should be very well-defined and specified as part of the documentation (ASAP).

PR #1414, opened 2020-09-10
HDDS-4231. Background Service blocks on task results.

lokeshj1703:

Background service currently waits on the results of the tasks. The idea is to track the time it took for the task to execute and log if task takes more than configured timeout.
This does not require waiting on the task results and can be achieved by just comparing the execution time of a task with the timeout value.

https://issues.apache.org/jira/browse/HDDS-4231

## How was this patch tested?

This was one of the reasons for failure of TestBlockDeletion tracked in HDDS-3432. Please check https://github.com/apache/hadoop-ozone/pull/1121/commits/a3feb31ae94e2c1a127f3d84f08ccc4edfb0e0ac . HDDS-3432 is blocked on a ratis bug.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/BackgroundService.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/BackgroundService.java
@@ -118,8 +118,8 @@ public abstract class BackgroundService {
           } finally {
             long endTime = System.nanoTime();
             if (endTime - startTime > serviceTimeoutInNanos) {
-              LOG.warn("Background task execution took {}ns > {}ns(timeout)",
-                  endTime - startTime, serviceTimeoutInNanos);
+              LOG.warn("{} Background task execution took {}ns > {}ns(timeout)",
+                  serviceName, endTime - startTime, serviceTimeoutInNanos);
             }
           }
         }, exec);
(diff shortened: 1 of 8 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/utils/BackgroundService.java
```diff
@@ -62,11 +56,11 @@ public BackgroundService(String serviceName, long interval,
     this.interval = interval;
     this.unit = unit;
     this.serviceName = serviceName;
-    this.serviceTimeout = serviceTimeout;
+    this.serviceTimeoutInNanos = TimeDuration.valueOf(serviceTimeout, unit)
+            .toLong(TimeUnit.NANOSECONDS);
```
amaliujia, line 60: Usually long is not enough to contain NANOSECONDS but this variable seems just a service timeout, which should be a very small number so it should be fine.

amaliujia:
+1 tried to review this PR and it looks good to me.

lokeshj1703:
Pushed another commit to fix a unit test.

nandakumar131:
Overall the patch looks good to me.

A minor suggestion, feel free to ignore it.
Instead of having a generic reference in `BackgroundTaskQueue` (which actually represents the type of result returned by the task in the queue, not the item in the queue), the `BackgroundTask<T>` can be modified to `BackgroundTask<BackgroundTaskResult>`. This will be more explicit.

lokeshj1703:
@nandakumar131 Thanks for the suggestion! I have removed generic reference from both BackgroundTaskQueue and BackgroundTask. BackgroundTask also doesn't require a generic interface right now.

adoroszlai (approved):
Thanks @lokeshj1703 for the fix and sorry for late response.

adoroszlai:
Thanks @lokeshj1703 for the contribution and @amaliujia and @nandakumar131 for the review.

PR #1415, opened 2020-09-10
HDDS-4232. Use single thread for KeyDeletingService.

lokeshj1703:

KeyDeletingService scan the keys from a particular rocksdb table and sends deletion request to SCM. Every thread would scan the table and send deletion requests. This can lead to multiple deletion request for a particular block. There is currently no way to distribute the keys to be deleted amongst multiple threads.

HDDS-4232

## How was this patch tested?

This lead to failure in TestBlockDeletion. Please refer https://github.com/apache/hadoop-ozone/pull/1121/commits/49b610a9feb54899ae226d3b6f2e36c5f8cc784f

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyDeletingService.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyDeletingService.java
@@ -66,9 +66,6 @@ public class KeyDeletingService extends BackgroundService {
   private static final Logger LOG =
       LoggerFactory.getLogger(KeyDeletingService.class);
 
-  // The thread pool size for key deleting service.
-  private final static int KEY_DELETING_CORE_POOL_SIZE = 2;
-
   private final OzoneManager ozoneManager;
   private final ScmBlockLocationProtocol scmClient;
   private final KeyManager manager;
@@ -82,7 +79,7 @@ public class KeyDeletingService extends BackgroundService {
       KeyManager manager, long serviceInterval,
       long serviceTimeout, ConfigurationSource conf) {
     super("KeyDeletingService", serviceInterval, TimeUnit.MILLISECONDS,
-        KEY_DELETING_CORE_POOL_SIZE, serviceTimeout);
+        1, serviceTimeout);
     this.ozoneManager = ozoneManager;
     this.scmClient = scmClient;
     this.manager = manager;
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyDeletingService.java
```diff
@@ -66,9 +66,6 @@
   private static final Logger LOG =
       LoggerFactory.getLogger(KeyDeletingService.class);
 
-  // The thread pool size for key deleting service.
-  private final static int KEY_DELETING_CORE_POOL_SIZE = 2;
```
amaliujia, line 70: when just not change 2 to 1?

bharatviswa504:
Hi @lokeshj1703 
https://issues.apache.org/jira/browse/HDDS-4121 This Jira changes the behavior with this openKeyCleanup service does not call SCM directly, it moves the keys to deleteTable, and it will be picked by KeyDeletingService and it will be only the one to send delete keys request to SCM from OM.

bharatviswa504:
And also a question, these are expired Keys in the openKeyTable, how for the same block 2 threads can send the request before. And also I see there is no logic for deleteExpiredKey today, and openKeyCleanup is not yet integrated in OM.

lokeshj1703:
The PR requires change in 1414 to address the test failure.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyDeletingService.java
lokeshj1703: Addressed it in latest commit.

lokeshj1703:
> [HDDS-4120](https://issues.apache.org/jira/browse/HDDS-4120) This Jira changes the behavior of openKeyCleanup service, and does not call SCM directly, it moves the keys to deleteTable by making a write request to OM. And later it will be picked by KeyDeletingService and it will be only the one to send delete keys request to SCM from OM.

@bharatviswa504 Thanks for providing the info! In that case we do not need the changes for openKeyCleanupService. I have addressed it in latest commit.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyDeletingService.java
amaliujia: Thanks!

adoroszlai (approved):

adoroszlai:
Thanks @lokeshj1703 for the fix and @amaliujia and @bharatviswa504 for the review.

PR #1416, opened 2020-09-10
HDDS-4233. Interrupted exeception printed out from DatanodeStateMachine

elek:

Hide Interrupted exceptions from WARN level as they are expected.
 
https://issues.apache.org/jira/browse/HDDS-4233

## How was this patch tested?

CI. Simple change.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeStateMachine.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeStateMachine.java
@@ -230,7 +230,6 @@ public class DatanodeStateMachine implements Closeable {
         // Some one has sent interrupt signal, this could be because
         // 1. Trigger heartbeat immediately
         // 2. Shutdown has be initiated.
-        LOG.warn("Interrupt the execution.", e);
         Thread.currentThread().interrupt();
       } catch (Exception e) {
         LOG.error("Unable to finish the execution.", e);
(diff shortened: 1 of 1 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeStateMachine.java
```diff
@@ -242,7 +241,7 @@ private void start() throws IOException {
           try {
             Thread.sleep(nextHB.get() - now);
           } catch (InterruptedException e) {
-            LOG.warn("Interrupt the execution.", e);
+            //triggerHeartbeat is called during the sleep
```
amaliujia, line 244: Hmm I might misunderstand this: will then the InterruptedException silently be hidden? If so is this expected behavior?

elek: Good question. Based on my understanding the interrupt is intentional (when DN would like to send out a HB immediately) 

```java
 /**
   * Calling this will immediately trigger a heartbeat to the SCMs.
   * This heartbeat will also include all the reports which are ready to
   * be sent by datanode.
   */
  public void triggerHeartbeat() {
    if (stateMachineThread != null) {
      stateMachineThread.interrupt();
    }
  }
```

I think it's expected to interrupt the thread, therefore we need to hide the exception.

amaliujia: Got it! Thanks for your clarification.

adoroszlai (approved):

PR #1419, opened 2020-09-13
HDDS-3755 [DESIGN] Storage-class for Ozone

elek:

This is a design doc, which is moved from hackmd to make it easier to track the progress.

### DIFF
--- a/hadoop-hdds/docs/content/design/storage-class.md
+++ b/hadoop-hdds/docs/content/design/storage-class.md
@@ -4,7 +4,7 @@ summary: New abstraction to configure replication methods.
 date: 2020-06-08
 jira: HDDS-3755
 status: draft
-author: Marton Ele
+author: Marton Elek
 ---
 <!--
   Licensed under the Apache License, Version 2.0 (the "License");
@@ -19,10 +19,331 @@ author: Marton Ele
   See the License for the specific language governing permissions and
   limitations under the License. See accompanying LICENSE file.
 -->
+
+
 # Abstract
 
-Proposal suggest to introduce a new storage-class abstraction which can be used to define different replication strategies (factor, type, ...) for different bucket/keys.
+One of the fundamental abstraction of Ozone is the _Container_ which used as the unit of the replication.
 
-# Link
+Containers have to favors: _Open_ and _Closed_ containers: Open containers are replicated by Ratis and writable, Closed containers are replicated with data copy and read only.
 
-https://hackmd.io/4kxufJBOQNaKn7PKFK_6OQ?view
+In this document a new level of abstraction is proposed: the *storage class* which defines which type of containers should be used and what type of transitions are supported.
+
+# Goals / Use cases
+
+## [USER] Simplify user interface and improve usability
+
+Users can choose from an admin provided set of storage classes (for example `STANDARD`, `REDUCED`) instead of using implementation specific terms (`RATIS/THREE`, `RATIS/ONE`)
+
+Today the users should use implementation spefific terms when key is created:
+
+```
+ozone sh key put --replication=THREE --type=RATIS /vol1/bucket1/key1 source-file.txt
+```
+
+There are two problems here:
+
+ 1. User should use low-level, technical terms during the usage. User might not know what is `RATIS` and may not have enough information to decide the right replication scheme.
+
+ 2. The current keys are only for the *open* containers. There is no easy way to add configuration which can be used later during the lifecycle of containers/keys. (For example to support `Ratis/THREE` --> `Ratis/TWO`)
+
+With the storage-class abstraction the complexity of configuration can be moved to the admin side (with more flexibility). And user should choose only from the available storage-classes (or use the default one).
+
+Instead of the earlier CLI this document proposes to use an abstract storage-class parameter instead:
+
+```
+ozone sh key put --storage-class=STANDARD /vol1/bucket1/key1 source-file.txt
+```
+
+## [USER] Set a custom replication for a newly created bucket
+
+A user may want to set a custom replication for bucket at the time of creation. All keys in the bucket will respect the specified storage class (subject to storage and quota availability). E.g.
+
+```
+ozone sh bucket create --storage-class=INFREQUENT_ACCESS
+```
+
+
+Bucket-level default storage-class can be overridden for ay key, but will be used as default.
+
+
+## [USER] Fine grained replication control when using S3 API
+
+A user may want to set custom replication policies for any key **which uploaded via S3 API**. Storage-classes are already used by AWS S3 API. With first-class support of the same concept in Ozone users can choose from the predefined storage-classes (=replication rules) with using AWS API:
+
+
+```
+aws s3 cp --storage-class=REDUCED file1 s3://bucket/file1
+```
+
+
+## [USER] Set the replication for a specific prefix
+
+A user may want to set a custom replication for a specific key prefix. All keys matching that prefix will respect the specified storage class. This operation will not affect keys already in the prefix (question: consider supporting this with data movement?)
+
+```
+ozone sh prefix setClass --storage-class=REDUCED /vol1/bucket1/tmp
+```
+
+Prefix-level default storage-class can be overridden for ay key, but will be used as default.
+
+## [ADMIN/DEV] Support multiple replication schemes
+
+Today there are two replication schemes which are hard coded in the code. Storage-class abstraction extends this behavior to support any number of replication schemes.
+
+Keys (and containers) can be categorized by storage-class which determines the replication scheme.
+
+## [ADMIN/USER] Flexible administrations
+
+As it's mentioned above, today it's hard to configure the details of the replications for key/bucket level. The only thing what we can define is the replication type for open containers (RATIS/THREE or RATIS/ONE) which determines the later lifecycle of the keys/containers.
+
+Any specific replication configuration can be configured only on cluster level and not on key level.
+
+A storage-class can define all the parameters for the spefific containers/keys:
+
+As an example this could be a storage-class definitions:
+
+```
+name: STANDARD
+states:
+    - name: open
+      replicationType: RATIS
+      repliationFactor: THREE
+    - name: closed
+      replicationType: COPY
+      repliationFactor: TWO
+      rackPolicy: different
+      transitions:
+        - target: ec
+          trigger:
+             ratio: 90%
+             used: 30d
+    - name: ec
+      codec: Reed-Solomon
+      scheme:
+         data: 6
+         parity: 3
+```
+
+This defines a replication scheme where only two replicas are enough from closed containers, and container will be erasure encoded under the hood if the 90% of the content is not used in the last 30 days.
+
+Please note that:
+
+ * All the low-level details of the replication rules can be configured here by the administrators
+ * Configuration is not global and not cluster-level, one can have different configuration for different storage-classes (which means for different keys/containers)
+ * Users dont' need to face with these details as they can use the storage-class (or just use the pre-created buckets and use default storage-class) abstraction
+
+## [DEV] Give flexibility to the developers
+
+Storage-class abstraction provides an easy way to plug in newer replication schemes. New type of replications (like EC) can be supported easily as the system will be prepared to allocate different type of containers.
+
+## [ADMIN] Better upgrade support
+
+Let's imagine that a new type of Open container replication is introduced (`RATIS-STREAM/THREE` instead of `RATIS/THREE`). If storage-classes are stored with the keys and containers instead of the direct replication rules we can:
+
+ 1. Easily change the replicaiton method of existing buckets/keys
+ 2. Turn on experimental features for specific buckets
+
+
+## [ADMIN] Change the cluster-wide replication
+
+An admin may decide to set a custom policy for an entire cluster.
+
+```
+ozone sh prefix setClass --storage-class=EC_6_3 /
+```
+
+# Unsupported use cases
+
+The following use cases are specifically unsupported.
+
+## [USER] Change the replication policy for a pre-existing key
+
+Changing the replication policy for a pre-existing key will require data movement and reauthoring containers and hence it is unsupported.
+
+## [USER] Defining storage-classes using Hadoop Compatible File System interface
+
+It's not possible to defined storage-class (or any replication rule) with using *Hadoop Compatible File System* interface. However storage-class defined on bucket level (or prefix level) will be inherited, even if the keys are created view the `o3fs://` or `o3s://` interfaces
+
+# The storage-class as an abstraction
+
+The previos section explained some user facing property of the storage-class concept. This section explains the concept compared to the existing Ozone design.
+
+## Containers in more details
+
+Container is the unit of replication of Ozone. One Container can store multiple blocks (default container size is 5GB) and they are replicated together. Datanodes report only the replication state of the Containers back to the Storage Container Manager (SCM) which makes it possible to scale up to billions of objects.
+
+The identifier of a block (BlockId) containers ContainerId and LocalId (ID inside the container). ContainerId can be used to find the right Datanode which stores the data. LocalId can be used to find the data inside one container.
+
+Container type defines the following:
+
+ * How to write to the containers?
+ * How to read from the containers?
+ * How to recover / replicate data in case of error
+ * How to store the data on the Datanode (related to the *how to write* question?)
+
+THe current definition of *Ratis/THREE* is the following (simplified version):
+
+ * **How to write**: Call standard Datanode RPC API on *Leader*. Leader will replicate the data to the followers
+ * **How to read**: Read the data from the Leader (stale read can be possible long-term)
+ * **How to replicate / recover**
+    * Transient failures can be handled by new leader election
+    * Permanent degradation couldn't be handled. (Transition to Closed containers is required)
+
+The current definitions of the *Closed/THREE*:
+
+  * **How to write**: Closed containers are not writeable
+  * **How to read**: Read the data from any nodes (Simple RPC call to the DN)
+  * **How to replicate / recover**
+    * Datanodes provides a GRPC endpoint to publish containers as compressed package
+    * Replication Manager (SCM) can send commands to DN to replicate data FROM other Datanode
+    * Datanode downloads the compressed package and import it
+
+The definitions of *Closed/ONE*:
+  * **How to write**: Closed containers are not writeable
+  * **How to read**: Read the data from any nodes (Simple RPC call to the DN)
+  * **How to replicate / recover**: No recovery, sorry.
+
+
+## Storage-class
+
+Let's define the *storage-class* as set of used container **types and transitions** between them during the life cycle of the containers.
+
+The type of the Container can be defined with the implementation type (eg. Ratis, EC, Closed) and with additional parameters related to type (eg. replication type of Ratis, or EC algorithm for EC containers).
+
+Today's implementation of Ozone can be described with two storage-classes:
+
+![](https://i.imgur.com/02rmoFP.png)
+
+The definition of STANDARD Storage class:
+
+ * *First container type/parameters*: Ratis/THREE replicated containers
+ * *Transitions*: In case of any error or if the container is full, convert to closed containers
+ * *Second container type/parameters*: Closed/THREE container
+
+The definition of REDUCED Storage class:
+
+ * *First container type/parameters*: Ratis/ONE replicated containers
+ * *Transitions*: In case the container is full, convert to closed containers
+ * *Second container type/parameters*: Closed/ONE container
+
+This proposal suggests to introduce an abstraction and name the two possible scheme. Later we can define other storage classes as well. For example we can define (Ratis/THREE --> Closed/FIVE) storage class, or more specific containers can be used for Erasure Coding or Random Read/Write.
+
+With this approach the storage-class can be an adjustable abstraction to define the rules of replications. Some key properties of this approach:
+
+**Key properties of the storage-class abstraction**:
+
+ * **Storage-class can be defined by configuration**: Storage class is nothing more just the definition of rules to store / replicate containers. They can be configured in config files and changed any time.
+ * **Object creation requires storage class**: Right now we should defined *replication factor* and *replication type* during the key creation. They can be replaced with setting only the Storage class
+ * **Storage-class defined on key** (and container) **level**
+ * **Storage-class is property of the containers**: As the unit of replication in Ozone is container, one specific storage-class should be adjusted for each containers.
+ * **Changing definition/configuration of storage class** is allowed for most of the properties. The change will modify the behavior of the Replication Manager, and -- eventually -- the containers will be replaced in a different way.
+ * **Changing storage class of a key** is not possible wihout moving data between containers
+ * **Buckets and volumes can provide default storage-classes**, to be inherited if not set on the Key level
+
+*Note*: we already support storage class for S3 objects the only difference is that it would become an Ozone level abstraction and it would defined *all* the container types and transitions.
+
+# Storage-class inside Ozone
+
+First of all, we can configure different replication levels easily with this approach (eg. Ratis/THREE --> Closed/TWO). Ratis need quorum but we can have different replication number after closing containers.
+
+We can also define topology related transitions (eg. after closing, one replica should be copied to different rack) or storage specific constrains (Right now we have only one implementation of the storage: `KeyValueContainer` but we can implement more and storage class provides an easy abstraction to configure the required storage).
+
+Datanode also can provide different disk type for containers in a certain storage class (eg. SSD for fast access).
+
+## Transitions between storage classes
+
+Storage-class is a property of all the *containers*. If a container is tagged with `STANDARD` storage-class, it defines how the container should be replicated and what transitions should be done (in case of predefined conditions).
+
+For example a `STANDARD` container should be replicated with `RATIS/THREE`, and closed in case the container is full. In closed *state*, the container should be replicated with the standard closed container replication.
+
+When keys are created, blocks are assigned from the appropriate containers. **There is no way to change storage-class without moving data* (distcp).
+
+bucket.storage-class := STANDARD --> REDUCED
+
+container=REDUCED
+
+key1 (storage-class=REDUCED)
+key2 (straoge-class=REDUCED)
+
+REDUCED: (RATIS/ONE -- CLOSED/ONE) --> (RATIS/TWO -- CLOSED/TWO)
+
+## Multi-stage storage-class
+
+There is a very specific use case the "temperature" of data. Key (and containers) can become COLD or HOT over the the time. As we need transitions between the different **state** of the containers, the COLD data should be a new state of a container not a storage class.
+
+Therefore the STANDARD storage class can be modified to support three states: OPEN, WARM(=CLOSED), COLD. Transitino between WARM and CLOSED can be managed by SCM (if container is not used, SCM can request to enable erasure coding). Or manually (one specific container can be forced to Erasure Coding / COLD state. Only useful for administrators)
+
+![](https://i.imgur.com/3NWzlm5.png)
+
+## Storage class on the user interface
+
+Today, we need to define the replication factor (like `THREE`) and replication type (like `RATIS`) during the creation of the key. It's not possible to adjust possible replication rule for the key when it become part of a closed containers.
+
+With storage-class support the users don't need to udnerstand the details of the replication. Users can choose from the well-known or custom-defined storage-classes.
+
+# Implementation changes
+
+**Phase 1**: while storage class abstraction can provide a very generic framework for replication, in the first implementation It's proposed to keep it as simple as possible:
+
+ * Don't introduce any new custom container state
+ * Don't introduce any new transitions
+ * Don't introduce any new custom storage classes
+
+But we can start to use the abstraction for the existing implementation:
+
+ * We will implement two, hard-coded storage-classes (STANDARD, REDUCED) which can configure the behavior of existing replication schemes
+ * Ozone Client will be changed to support `--storage-class` instead of `--type` and `--factor`
+ * `ofs://` and `o3fs://` doesn't require any modification
+ * Storage class will be used as a custom String in RPC protocol (in storage-sensitive places it can be stored as an integer with an additional mapping)
+ * All the replication logic (PipelineManager/ReplicationManager) will work exactly as before. Storage-class will be resolved to the required replication config. Pipelines will have the same type as before (eg. Ratis/THREE)
+ * SCM allocateContainer logic should be changed. Right now it creates a predefined number of Open containers for each `owner` (high level applocation). It can be extended easily to create open containers for each storage-classes
+ * Having slightly more open containers doesn't make any difference (5-10 more open containers per disk vs. the 5000-10000 existing containers)
+
+# Backward compatibility
+
+ * In allocateBlock and intra-service communication we can replace replication type and replication factor with storage-class
+ * In client API replication factor and replication type will be optional
+ * Legacy type/factor definitions will be mapped to real storage classes under the hood (which defines the real replication scheme). For example `Ratis/THREE --> STANDARD`
+
+# Additional use cases
+
+In addition to the possible, replication related additional options there are two very specific use cases where we can use storage classes. Both requires more design discussion but here I quickly summarize some *possible directions* with the help of the storage class abstraction.
+
+This section demonstrates the power of the abstraction the exact implementation of these features can be (and probably will be slightly) different
+
+## Erasure coding
+
+To store cold data on less bits (less than the data * 3 replicas) we can encode the data and store some additional bits to survive replica loss. In a streaming situation (streaming write) it can be tricky as we need enough chunks to do the Reed-Solomon magic. With containers we are in a better position. After the first transition of the Open containers we can do EC encoding and that time we have enough data to encode.
+
+There are multiple options to do EC, one most simplest way is to encode Containers after the first transition:
+
+Storage class COLD:
+
+ * *First container type/parameters*: Ratis/THREE replicated containers
+ * *Transitions*: In case of any error or if the container is full, convert to closed containers
+ * *Second container type/parameters*: EC RS(6,2)
+
+With this storage class the containers can be converted to a specific EC container together with other containers (For example instead of 3 x C1, 3 x C2 and 3 x C3 containers can be converted to C1, C2, C3 + Parity 1 + Parity2).
+
+## Random-read write
+
+NFS / Fuse file system might require to support Random read/write which can be tricky as the closed containers are immutable. In case of changing any one byte in a block, the whole block should be re-created with the new data. It can have a lot of overhead especially in case of many small writes.
+
+But write is cheap with Ratis/THREE containers. Similar to any `ChunkWrite` and `PutBlock` we can implementa `UpdateChunk` call which modifies the current content of the chunk AND replicates the change with the help of Ratis.
+
+Let's imagine that we solved the resiliency of Ratis pipelines: In case of any Ratis error we can ask other Datanode to join to the Ratis ring *instead of* closing it. I know that it can be hard to implement, but if it is solved, we have an easy solution for random read/write.
+
+If it works, we can define the following storage-class:
+
+ * **Initial container** type/parameters: Ratis/THREE
+ * **Transitions**: NO. They will remain "Open" forever
+ * **Second container type**: NONE
+
+This would help the random read/write problem, with some limitations: only a few containers should be managed in this storage-class. It's not suitable for really Big data, but can be a very powerful addition to provide NFS/Fuse backend for containers / git db / SQL db.
+
+# References
+
+ * Storage class in S3: https://aws.amazon.com/s3/storage-classes/
+
+


hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+## [USER] Set a custom replication for a newly created bucket
+
+A user may want to set a custom replication for bucket at the time of creation. All keys in the bucket will respect the specified storage class (subject to storage and quota availability). E.g.
+
+```
+ozone sh bucket create --storage-class=INFREQUENT_ACCESS
+```
+
+
+Bucket-level default storage-class can be overridden for ay key, but will be used as default.
```
xiaoyuyao, line 67: ay=>any

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+```
+aws s3 cp --storage-class=REDUCED file1 s3://bucket/file1
+```
+
+
+## [USER] Set the replication for a specific prefix
+
+A user may want to set a custom replication for a specific key prefix. All keys matching that prefix will respect the specified storage class. This operation will not affect keys already in the prefix (question: consider supporting this with data movement?)
+
```
xiaoyuyao, line 83: not only movement but also existing keys with/without prefix storage class? do we need something like SPS?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+```
+ozone sh prefix setClass --storage-class=REDUCED /vol1/bucket1/tmp
+```
+
+Prefix-level default storage-class can be overridden for ay key, but will be used as default.
```
xiaoyuyao, line 88: Per key storage class is probably too much and we may end up lot of unfilled different containers on backend but the new request fail to find any match.

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+      replicationType: RATIS
+      repliationFactor: THREE
+    - name: closed
+      replicationType: COPY
+      repliationFactor: TWO
+      rackPolicy: different
+      transitions:
+        - target: ec
+          trigger:
+             ratio: 90%
```
xiaoyuyao, line 119: can you elaborate the ratio here? close criteria?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+# Unsupported use cases
+
+The following use cases are specifically unsupported.
+
+## [USER] Change the replication policy for a pre-existing key
+
+Changing the replication policy for a pre-existing key will require data movement and reauthoring containers and hence it is unsupported.
+
+## [USER] Defining storage-classes using Hadoop Compatible File System interface
```
xiaoyuyao, line 164: Hadoop apps use setReplication(), can we support that in storage class?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+With this approach the storage-class can be an adjustable abstraction to define the rules of replications. Some key properties of this approach:
+
+**Key properties of the storage-class abstraction**:
+
+ * **Storage-class can be defined by configuration**: Storage class is nothing more just the definition of rules to store / replicate containers. They can be configured in config files and changed any time.
+ * **Object creation requires storage class**: Right now we should defined *replication factor* and *replication type* during the key creation. They can be replaced with setting only the Storage class
+ * **Storage-class defined on key** (and container) **level**
+ * **Storage-class is property of the containers**: As the unit of replication in Ozone is container, one specific storage-class should be adjusted for each containers.
+ * **Changing definition/configuration of storage class** is allowed for most of the properties. The change will modify the behavior of the Replication Manager, and -- eventually -- the containers will be replaced in a different way.
```
xiaoyuyao, line 240: This seems only changable for Close->EC, which is taken care by the replication manager? Can you change from RATIS TO RATIS-STREAMING?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+![](https://i.imgur.com/3NWzlm5.png)
+
+## Storage class on the user interface
+
+Today, we need to define the replication factor (like `THREE`) and replication type (like `RATIS`) during the creation of the key. It's not possible to adjust possible replication rule for the key when it become part of a closed containers.
+
+With storage-class support the users don't need to udnerstand the details of the replication. Users can choose from the well-known or custom-defined storage-classes.
+
+# Implementation changes
```
xiaoyuyao, line 285: Can we make this compatible with Ozone GA client?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+ * *Transitions*: In case of any error or if the container is full, convert to closed containers
+ * *Second container type/parameters*: EC RS(6,2)
+
+With this storage class the containers can be converted to a specific EC container together with other containers (For example instead of 3 x C1, 3 x C2 and 3 x C3 containers can be converted to C1, C2, C3 + Parity 1 + Parity2).
+
+## Random-read write
+
+NFS / Fuse file system might require to support Random read/write which can be tricky as the closed containers are immutable. In case of changing any one byte in a block, the whole block should be re-created with the new data. It can have a lot of overhead especially in case of many small writes.
+
+But write is cheap with Ratis/THREE containers. Similar to any `ChunkWrite` and `PutBlock` we can implementa `UpdateChunk` call which modifies the current content of the chunk AND replicates the change with the help of Ratis.
```
xiaoyuyao, line 333: we can always create new chunks and link that to the block? the old chunk can be kept as old version or garbage collected like delete.

xiaoyuyao (commented):
LGTM. Thanks for writing this up. A few comments inline.

hadoop-hdds/docs/content/design/storage-class.md
elek: Good question. storage-class is assigned to containers and keys, so keys already have storage-classes, but might be the default. To start with I think we should enable to set default storage class only when the prefix/bucket is created.

hadoop-hdds/docs/content/design/storage-class.md
elek: > We may end up lot of unfilled different containers on backend but the new request fail to find any match.

The number of storage-classes are limited. Today we have two (REDUCED/STANDARD) So we couldn't have a lot of unfilled containers as we should have one open containers for all the storage-classes. I couldn't see any fragmentation as keys can be stored in the same container if they are in the same storage-class.

hadoop-hdds/docs/content/design/storage-class.md
elek: Good point. We can support it with mapping numbers to storage classes:

```
setReplication(1) --> StorageClass=REDUCED (can be STANALONE/ONE->CLOSED/ONE or different)
setReplication(3) --> StorageClass=STANDARD (can be RATIS/THREE or EC or anything else...)
```

hadoop-hdds/docs/content/design/storage-class.md
elek: I think the safe answer is now. For each of the replication models we can have definition about the possible migration. For example Closed/THREE can be migrated to Closed/<ANY>. But It might not be possible to change CLOSED/THREE to EC/5:2 (at least with the first iplementation)

hadoop-hdds/docs/content/design/storage-class.md
elek: Yes, this is also possible. This section shows the possibilities with the storage-class abstraction. With this abstraction we can define different behavior for the buckets which are used for random read write. The exact details can be changes

elek:
@umamaheswararao @arp7 Do you have any more comments / questions?

If, not, I would suggest to start to create smaller patches from #1208 and merge them one by one...

arp7:
Isn't this design doc a few steps away from coding? A ton of detail is missing around how SCM will manage multiple classes of pipelines and how replication manager will need to be modified.

> All the replication logic (PipelineManager/ReplicationManager) will work exactly as before. Storage-class will be resolved to the required replication config. Pipelines will have the same type as before (eg. Ratis/THREE)

Really? I seems impossible that we can introduce new storage classes without requiring changes to the replication logic.

elek:
>> Isn't this design doc a few steps away from coding? A ton of detail is missing around how SCM will manage multiple classes of pipelines and how replication manager will need to be modified.

Would you be so kind add a few questions which can cover what are the missing parts for you? It would be a big help.

We have a working POC which passes all the tests, but it is possible that some details are available only in the code. If you have any question I kindly ask to put it to here and I will answer them.

>> All the replication logic (PipelineManager/ReplicationManager) will work exactly as before. Storage-class will be resolved to the required replication config. Pipelines will have the same type as before (eg. Ratis/THREE)

> Really? It seems impossible that we can introduce new storage classes without requiring changes to the replication logic.

This paragraph is talking about the initial implementation where we wouldn't like to introduce new storage-classes, they are hard coded as Ratis/THREE and Ratis/ONE are hard-coded today.

Long-term it can be possible to introduce new storage-class without introducing new replication logic (for example Closed/TWO) or it can be possible to introduce totally new replication logic (like EC)

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+## Storage-class
+
+Let's define the *storage-class* as set of used container **types and transitions** between them during the life cycle of the containers.
+
+The type of the Container can be defined with the implementation type (eg. Ratis, EC, Closed) and with additional parameters related to type (eg. replication type of Ratis, or EC algorithm for EC containers).
+
+Today's implementation of Ozone can be described with two storage-classes:
+
+![](https://i.imgur.com/02rmoFP.png)
```
runzhiwang, line 216: Can we put the png in ozone rather than imgur ? If imgur crash or offline, we can not visit it.

elek: Good point. I added both the pictures to make it available offline. (#726843f)

elek:
I would like to move forward this design doc. 

There are now new (bit) concerns since 21st of September (image caching is addressed) and as it's an important abstraction both for EC and different replication levels I would prefer to achieve consensus shortly.

As far as I understood @arp7 still have concerns, and I am trying to organize an offline call with him to discuss all the open questions.

@arp7 Can you please help me to make the progress faster and write down your remaining question. It would be very great help. (Thank you in advance).

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+As it's mentioned above, today it's hard to configure the details of the replications for key/bucket level. The only thing what we can define is the replication type for open containers (RATIS/THREE or RATIS/ONE) which determines the later lifecycle of the keys/containers.
+
+Any specific replication configuration can be configured only on cluster level and not on key level.
+
+A storage-class can define all the parameters for the spefific containers/keys:
+
+As an example this could be a storage-class definitions:
+
+```
+name: STANDARD
```
arp7, line 107: I think this is way too flexible and describes too many features that this document does not deliver on. The example should be updated to reflect that.

If you really want to go this route, you need the define the grammar in detail, and also how various Ozone components should parse and operate on this grammar.

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+## [DEV] Give flexibility to the developers
+
+Storage-class abstraction provides an easy way to plug in newer replication schemes. New type of replications (like EC) can be supported easily as the system will be prepared to allocate different type of containers.
+
+## [ADMIN] Better upgrade support
+
+Let's imagine that a new type of Open container replication is introduced (`RATIS-STREAM/THREE` instead of `RATIS/THREE`). If storage-classes are stored with the keys and containers instead of the direct replication rules we can:
+
+ 1. Easily change the replicaiton method of existing buckets/keys
```
arp7, line 144: I think you need to clarify that you cannot change the replication method of individual objects. Instead you can only change the definition of a storage class which will result in cluster wide data movement!

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+The previos section explained some user facing property of the storage-class concept. This section explains the concept compared to the existing Ozone design.
+
+## Containers in more details
+
+Container is the unit of replication of Ozone. One Container can store multiple blocks (default container size is 5GB) and they are replicated together. Datanodes report only the replication state of the Containers back to the Storage Container Manager (SCM) which makes it possible to scale up to billions of objects.
+
+The identifier of a block (BlockId) containers ContainerId and LocalId (ID inside the container). ContainerId can be used to find the right Datanode which stores the data. LocalId can be used to find the data inside one container.
+
+Container type defines the following:
```
arp7, line 178: A container does not really define this today. There is a lot of replication specific logic encoded all over the system. If you want to introduce new replication types, all of that must be refactored and made generic behind interfaces. We are talking about months of refactoring/rewrite effort.

I think if EC is the immediate use case, it is easier to introduce a new EC-enabled write path.

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+ * How to write to the containers?
+ * How to read from the containers?
+ * How to recover / replicate data in case of error
+ * How to store the data on the Datanode (related to the *how to write* question?)
+
+THe current definition of *Ratis/THREE* is the following (simplified version):
+
+ * **How to write**: Call standard Datanode RPC API on *Leader*. Leader will replicate the data to the followers
```
arp7, line 187: This is a fraction of the logic. It also defines how recovery from failures must be done, how containers must be closed.

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+ * **Storage-class is property of the containers**: As the unit of replication in Ozone is container, one specific storage-class should be adjusted for each containers.
+ * **Changing definition/configuration of storage class** is allowed for most of the properties. The change will modify the behavior of the Replication Manager, and -- eventually -- the containers will be replaced in a different way.
+ * **Changing storage class of a key** is not possible wihout moving data between containers
+ * **Buckets and volumes can provide default storage-classes**, to be inherited if not set on the Key level
+
+*Note*: we already support storage class for S3 objects the only difference is that it would become an Ozone level abstraction and it would defined *all* the container types and transitions.
+
+# Storage-class inside Ozone
+
+First of all, we can configure different replication levels easily with this approach (eg. Ratis/THREE --> Closed/TWO). Ratis need quorum but we can have different replication number after closing containers.
```
arp7, line 248: I don't think that is so easy. A lot of the failure and recovery logic for the Ratis write pipeline assumes that we have three replicas.

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+ * We will implement two, hard-coded storage-classes (STANDARD, REDUCED) which can configure the behavior of existing replication schemes
+ * Ozone Client will be changed to support `--storage-class` instead of `--type` and `--factor`
+ * `ofs://` and `o3fs://` doesn't require any modification
+ * Storage class will be used as a custom String in RPC protocol (in storage-sensitive places it can be stored as an integer with an additional mapping)
+ * All the replication logic (PipelineManager/ReplicationManager) will work exactly as before. Storage-class will be resolved to the required replication config. Pipelines will have the same type as before (eg. Ratis/THREE)
+ * SCM allocateContainer logic should be changed. Right now it creates a predefined number of Open containers for each `owner` (high level application). It can be extended easily to create open containers for each storage-classes
+ * Having slightly more open containers doesn't make any difference (5-10 more open containers per disk vs. the 5000-10000 existing containers)
+
+# Backward compatibility
```
arp7, line 303: This section is too short. Since we have shipped Ozone GA, backwards compatibility is a must. So we should cover it in a lot more detail. E.g. when you say

> Legacy type/factor definitions will be mapped to real storage classes under the hood (which defines the real replication scheme).

What exactly does "under the hood" mean. What if a new client talks to an old server that doesn't support storage classes? What information should it send. And vice versa. Also how will upgrades be handled with OM HA?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+
+![Storage class transitions](storage-class-transition.png)
+
+## Storage class on the user interface
+
+Today, we need to define the replication factor (like `THREE`) and replication type (like `RATIS`) during the creation of the key. It's not possible to adjust possible replication rule for the key when it become part of a closed containers.
+
+With storage-class support the users don't need to udnerstand the details of the replication. Users can choose from the well-known or custom-defined storage-classes.
+
+# Implementation changes
```
arp7, line 285: We need a lot more details in this section.

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+ * Don't introduce any new custom storage classes
+
+But we can start to use the abstraction for the existing implementation:
+
+ * We will implement two, hard-coded storage-classes (STANDARD, REDUCED) which can configure the behavior of existing replication schemes
+ * Ozone Client will be changed to support `--storage-class` instead of `--type` and `--factor`
+ * `ofs://` and `o3fs://` doesn't require any modification
+ * Storage class will be used as a custom String in RPC protocol (in storage-sensitive places it can be stored as an integer with an additional mapping)
+ * All the replication logic (PipelineManager/ReplicationManager) will work exactly as before. Storage-class will be resolved to the required replication config. Pipelines will have the same type as before (eg. Ratis/THREE)
+ * SCM allocateContainer logic should be changed. Right now it creates a predefined number of Open containers for each `owner` (high level application). It can be extended easily to create open containers for each storage-classes
```
arp7, line 300: Please add more implementation details - how do you propose to extend the SCM? What parts of the SCM will change?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+ * Having slightly more open containers doesn't make any difference (5-10 more open containers per disk vs. the 5000-10000 existing containers)
+
+# Backward compatibility
+
+ * In allocateBlock and intra-service communication we can replace replication type and replication factor with storage-class
+ * In client API replication factor and replication type will be optional
+ * Legacy type/factor definitions will be mapped to real storage classes under the hood (which defines the real replication scheme). For example `Ratis/THREE --> STANDARD`
+
```
arp7, line 308: How will SCM replication monitor change?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+## Storage class on the user interface
+
+Today, we need to define the replication factor (like `THREE`) and replication type (like `RATIS`) during the creation of the key. It's not possible to adjust possible replication rule for the key when it become part of a closed containers.
+
+With storage-class support the users don't need to udnerstand the details of the replication. Users can choose from the well-known or custom-defined storage-classes.
+
+# Implementation changes
+
+**Phase 1**: while storage class abstraction can provide a very generic framework for replication, in the first implementation It's proposed to keep it as simple as possible:
```
arp7, line 287: In your phase1 (I think you call it PoC) patch, you have introduced a notion of storage class registry Can you explain it in detail here?

arp7 (commented):
Marton, I am adding a few detailed comments as you requested. I think you should update your design doc to describe in detail just the initial/phase 1 changes of capturing existing storage classes. The rest can be moved to future/addendum section, then it will be easier to review your patch.

hadoop-hdds/docs/content/design/storage-class.md
elek: >  It also defines how recovery from failures must be done

Agree, this is included in the line:

_How to recover / replicate data in case of error_

> how containers must be closed

Good point, and it's a very important. I think there are two possible way to think about it:

 * We have OPEN containers and CLOSED containers, and *replication mechanism* is the definition the transitions between them

 * We have one NORMAL type of container which two type of states (Open, Closed).

Technically both are the same (IMHO) just with different way to explain it. This document followed the first approach but I added more text to clarify it.

hadoop-hdds/docs/content/design/storage-class.md
elek: Good question, I added this information to the beginning of to the implementation section.

hadoop-hdds/docs/content/design/storage-class.md
elek: I wrote it down in more details. (Exactly as today but replication factor is coming from the storage-class configuration for each containers)

hadoop-hdds/docs/content/design/storage-class.md
elek: Thanks for the suggestion. I extended this part with different pagraph for OM, SCM, Client... changes (instead of using only one lines).

hadoop-hdds/docs/content/design/storage-class.md
elek: Thanks, Lastest version inculudes more information.

hadoop-hdds/docs/content/design/storage-class.md
elek: I added more information about this topic. 

TLDR; The change can be fully backward compatible as we can create a hard-coded mapping table to map existing factor/type to the initial storageClass definitions.

OM HA doesn't cause any problem as it works exactly as before just (if metadata layout version is upgraded) instead of factor/type the storage class information will be saved.

hadoop-hdds/docs/content/design/storage-class.md
elek: Agree, it's not easy. But might be possible in case of it's required by users. For the clarity I removed this paragraph.

hadoop-hdds/docs/content/design/storage-class.md
elek: I made it more clear. We have replication mechanism which uses different type of containers. Today we have one replication mechanism which uses open and closed containers and defines the transitions between them.

I agree that a new replication mechanism requires code changes.

But *choosing* and *configuring* the replication for specific containers / keys can be done easily with this abstraction.

hadoop-hdds/docs/content/design/storage-class.md
elek: Thanks, fair enough. I made it clear.

hadoop-hdds/docs/content/design/storage-class.md
elek: Sorry, if it was not clear. The intention was to show how different parts can be configured and not to provide a very flexible grammar (which may be suggested by this grammar fragment).

I re-wrote this part and added a table instead of the yaml which shows the point which can be customized.

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +19,331 @@ author: Marton Ele
...
+ozone sh prefix setClass --storage-class=EC_6_3 /
+```
+
+# Unsupported use cases
+
+The following use cases are specifically unsupported.
+
+## [USER] Change the replication policy for a pre-existing key
+
+Changing the replication policy for a pre-existing key will require data movement and reauthoring containers and hence it is unsupported.
```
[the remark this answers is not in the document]
elek: Line 48 is the following:

> . The current keys are only for the *open* containers. There is no easy way to add configuration which can be used later during the lifecycle of containers/keys. (For example to support `Ratis/THREE` --> `Ratis/TWO`)

Sorry if it was not clear, but it's not about the dynamic change, but configurability. To make the closed replication factor configurable with the current approach you may need new CLI parameters:

```
ozone sh key put --replication=THREE --closed-replication=TWO --type=RATIS /vol1/bucket1/key1 source-file.txt
```

With each of the new configurable options (which are per-key configurations) we should introduce new CLI parameters. Storage-class simplifies it.

This location talks about the changing of replication mechanism of one key. It's not easy beacuse it requires rewrite of the containers.

elek:
> I think if you really want to proceed with this idea, you should update your design doc to describe in detail just the initial/phase 1 changes of capturing existing storage classes. The rest can be moved to future/addendum section, then it will be easier to review your patch.

Thank you very much the comments, it was a great help. I updated the document based on your feedback and I believe it's more clear and understandable now.

1. It clearly talks only about the current plan. A few exceptions: future plans are moved to the last section to show some context but clearly stated that it's not part of the plan. HOT/COLD data is also explained, but marked as out of the scope. It is included because it was a specific question during the discussions.

2. Based on your feedback I found that the flexibility level is not very clear, therefore are made it more explicit everywhere. That's the biggest change in the commit. Instead of talking about storage-class which is the definition of transitions of states I talk about storage-class as a selection of a replication mechanism (which already has implementation) and the configuration of the replication (eg. set the factor for closed containers).

I believe this change makes it clear that our opinion is closer: I don't think that we need a very flexible way to combine and container states. (Earlier version of this document used more flexible context, but I learned from the feedback).

> I am not yet convinced that the effort to introduce custom new storage classes is worth it. The only immediate use case is EC which can be solved lot more easily with HDFS-like EC policies.

The other use case is the replication factor TWO, which was a request by our early adopters and tested by @maobaolong based on this concept.

elek:
Pushed the new version. Thanks again the feedback @arp7. Would you be so kind to take another look.

I also uploaded a PDF version to https://issues.apache.org/jira/browse/HDDS-3755, in case it's more readable for you...

elek:
@arp7 I think all of your concerns are addressed with the new version. It's the same concept but -- as you suggested -- scope is restricted to the current problem. 

Did you have any chance to check it? 

(I would prefer to finish this design process after 5 months.)

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +20,408 @@ author: Marton Ele
...
+NFS / Fuse file system might require to support Random read/write which can be tricky as the closed containers are immutable. In case of changing any one byte in a block, the whole block should be re-created with the new data. It can have a lot of overhead especially in case of many small writes.
+
+But write is cheap with Ratis/THREE containers. Similar to any `ChunkWrite` and `PutBlock` we can implement an `UpdateChunk` call which modifies the current content of the chunk AND replicates the change with the help of Ratis.
+
+Let's imagine that we solved the resiliency of Ratis pipelines: In case of any Ratis error we can ask other Datanode to join to the Ratis ring *instead of* closing it. I know that it can be hard to implement, but if it is solved, we have an easy solution for random read/write.
+
+If it works, we can define the following storage-class:
+
+ * **Initial container** type/parameters: Ratis/THREE
+ * **Transitions**: NO. They will remain "Open" forever
```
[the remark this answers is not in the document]
elek: No, we have no such formal proposal. Please see the beginning of this section:

> **Both requires more design discussion, and won't be implemented in the first phase,** but here I quickly summarize some *possible directions* with the help of the storage-class abstraction, to show the power and usefulness of the abstraction.

TLDR; It's an example, how storage-class can be useful in the future (in addition to the benefits discussed above)

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +20,408 @@ author: Marton Ele
...
+
+To store cold data on less bits (without the 3x replication overhead) we can encode the data with Erasure Coding and store some additional bits to survive replica loss. In a streaming situation (streaming write) it can be tricky as we need enough chunks to do the Reed-Solomon magic. With containers we are in a better position. After the first transition of the Open containers we can do EC encoding and that time we have enough data to encode.
+
+There are multiple options to do EC, one most simplest (but somewhat limited) way is to encode Containers after the first transition:
+
+Storage class COLD:
+
+ * *First container type/parameters*: Ratis/THREE replicated containers
+ * *Transitions*: In case of any error or if the container is full, convert to closed containers
+ * *Second container type/parameters*: EC RS(6,2)
```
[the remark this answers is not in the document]
elek: This is the same category as before, the beginning of the section defines it's a possible way to move forward.

This part is written to the container level erasure coding (not for the striped), which may or may not be implemented in addtional to the striped level. (I don't think we need it, but there were some opinions that striped EC can have some performance problems for Hive, where multiple disk read is required for small files instead of one disk read which can be cached by the OS).

But back to your question: let's say we have both container level async AND striped EC. Storage class is nothing more just an additional tag on the containers to classify which replication mechanism sholud be used. With this label we can classify containers easily and choose the right containers for a specific client request.

elek:
@arp7 Are you fine with the changes? 

I understand that you are very busy. I totally understand if you have no time to read it again. 

In this case, I will move it forward and present the final version at the community meeting and ask for merge / approval.

If you are planning to add your comments in the near future, just let me know and I will wait for it.

arp7:
I am -1 on this design. You have still not defined how you will abstract out all the behavior defined by storage classes.

hadoop-hdds/docs/content/design/storage-class.md
arp7: It is not sufficient to just define the replication factor. The replication/recovery behavior completely changes based on the replication scheme. E.g. for EC, we are talking about picking some coordinator during recovery. How do you abstract away behavior like that?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +20,408 @@ author: Marton Ele
...
+  
+**Protocol changes**: 
+
+  * `storageClass` field should be added to all the protocols where we use `factor/type` today. Old `factor/type` fields will be kept for compatibility reasons. 
+  * Storage-class will be represented as a custom String in the RPC protocol (in storage-sensitive places it can be stored as an integer with an additional mapping).
+  * There are two protocols which should be changed:
+   * *Key creation* and *Key information* related structures in `OmClientProtocol` should be updated (`KeyInfo`, `MultipartUploadInfo`, `KeyArgs`,... )
+   *  SCMAdmin protocol should be improved to include `storage-class` in `ContainerRequestProto`.
+
+**Ozone Client changes**: Ozone client classes (`OzoneBucket`, `OzoneKey`,...) should be modified to propagate the `storageClass` everywhere instead of `factor/type`. To make it backward compatible we can create legacy mapping table should be used.
```
arp7, line 341: This one sentence cannot capture all the client changes right? The client has to select different data paths depending on the replication scheme (Ratis vs EC). How will that be abstracted out?

hadoop-hdds/docs/content/design/storage-class.md
```diff
@@ -19,10 +20,408 @@ author: Marton Ele
...
+**Ozone Manager changes**: OM needs the following changes:
+
+  * `storageClass` should be stored for the keys instead of `factor/type`. THe *legacy mapping table* will be used to generate storageClass for old, persisted keys which has `factory/type` in the RocksDB but no storageClass. It can be added to the appropriate codec.
+  * `storageClass` should be propagated when SCM is called from OM for new container allocations.
+  * If `storageClass` is not defined for a key, the default should be used from the `bucket`/`volume`. If they are not specified the cluster-wide default will be used.
+
+**Storage Container Manager** changes:
+
+  * `ContainerInfo` will include the `storageClass` as a property
+  * Internal allocation request will contain the `storageClass` (old requests can be handled with the *legacy mapping table*)
```
arp7, line 356: How will SCM make decisions about allocating pipelines of different kinds?

arp7 (changes requested):
I added a couple of high-level comments about some things that are missing. It is hard to comment on this doc because it leaves out so much detail needed for a feature of this scope. Asking a reviewer to provide detailed feedback just pushes the work of thinking through these details to the reviewer.

This is a very ambitious feature. Storage policies, replication factor, and EC policies are all separate in HDFS and they were never unified under storage classes that you are proposing here.

The problem is that Ratis 3-way replication is so deeply embedded in the client, SCM and DNs that to abstract away details behind storage classes will be very hard. It is further complicated by the fact that objects from different buckets go to the same container. So you cannot change the storage policy of any one bucket, instead you can just change the definition of the policy overall which affects data globally. So there is no way for me to state that use replication factor 2 for a specific bucket.

What you could do is completely decouple EC from this proposal and just have this proposal be restricted to choosing the replication factor for closed containers, which I think covers your original goals.

hadoop-hdds/docs/content/design/storage-class.md
elek: Thanks the question. The implementation details of EC is intentionally removed from this doc as you suggested. This document focuses on the current replication schemes. But EC can be handed in the same way:

The current implementation of OzoneClient calls OM first to create a new key, and uses the returned pipeline information to initialize the right `XCeiverClient` (`XCeiverClientManager.getClient`)

With storage-class the logic is the same, the only required change is adding storage-class to the `CreateKey` request and use it instead of replication/factor. The storage-class will be used in the `AllocateScmBlock` call, and SCM will return with the right pipeline.

But this is an SCM change, client-side it's enough to have the mentioned  change.  

In fact if we implement it once, we don't need to modify the logic for EC, just implement a new `XCeiverClientSpi`. Without storage-class we need to modify the `CreateKey` (or at least adding a new replication type, which doesn't use the factor at all.)

hadoop-hdds/docs/content/design/storage-class.md
elek: We don't need to modify any logic in SCM. We already have allocation logic for Ratis/Standalone pipelines. When a new request comes in, we should choose from the pre-allocated pipelines OR create a new one, based on replication factor.

This logic is exactly the same with storage-class. Based on the proposed new storage-class field of the `AllocateScmBlock` can identify the required replication scheme (eg. Ratis) and parameters (eg. THREE) and choose the pipeline exactly the same way s today.

Obviously, EC requires new pipeline management, which is not covered in this document (as suggested).

But it shows the benefit of the storage-class: the list of configured storage-class provides a pre-defined list of the required pipelines. For example if two storage classes are defined asEC 6:3 and EC 3:2, SCM will have the information about the required pipelines to pre-allocate.

elek:
> It is not sufficient to just define the replication factor. The replication/recovery behavior completely changes based on the replication scheme. E.g. for EC, we are talking about picking some coordinator during recovery. How do you abstract away behavior like that?

I am not sure If I understand this question. Offline-recovery and co-ordinator election is an internal implementation detail of EC replication scheme. The only role of storage-class is to *select* and *configure* the required configuration scheme. If EC 3:2 is configured we will have EC 3:2 containers created which will be managed by SCM as EC 3:2 containers.

elek:
> I am -1 on this design. You have still not defined how you will abstract out all the behavior defined by storage classes.

Thank you the feedback @arp7. 

I answered all of your question, please let me know if you have anymore. I am also open to explain the full concept during a call if it's more convenient for you.

Based on your question I feel that you ask questions based on the original version, which followed a very generic approach. Based on the feedback I modified the proposal and the scope of this version very focused and limited. It's more like a technique to manage configuration without significant changes on SCM side.

Please let me know if you have any more concerns.

elek:
Hey, @arp7.

Not sure what is your suggestion here, can you please clarify (you added -1 and also started discussion to improve it).

 1. If you think it should be closed without further discussion, please close it 
 2. If you are interested about the answers to your questions please check them and let me know if you have any further questions.
 
 Thank you very much.

arp7:
> If you think it should be closed without further discussion, please close it
It is not the reviewer's task to close PRs they disagree with. As the PR author, you are free to do so yourself though.

elek:
> It is not the reviewer's task to close PRs they disagree with. As the PR author, you are free to do so yourself though.

Personally, I would prefer to not close this, but continue the discussion and solve the defined problems in some way to make it acceptable for anybody. I asked you only, because it's not clear what is your suggestion 1 or 2, if you close it I will accept it.  

If you don't close the PR I assume that you are fine to continue the discussion with you and/or others (if you have no time) to address all issues. In this case I will update the doc again and do a presentation on the community sync to find a way to go forward.

elek:
> If you don't close the PR I assume that you are fine to continue the discussion w

It seems that you have no more concerns. I am glad. I think the best way to move forward to combine the latest version (which is very lightweight) to the EC doc. We have exactly the same questions about EC: how they will be configured.

arp7:
@elek, I have a binding -1 above. A veto does not expire with time.

elek:
> @elek, I have a binding -1 above. A veto does not expire with time.

Thanks for the feedback. You gave a binding -1 vote for the previous version and also added your comments. All the comments are answered and addressed, you didn't review/comment the latest version which is a reworked version of the original version and concept itself is modified. 

As I wrote, we faced with the same questions during the EC discussion, therefore the questions should be answered in any way. Feel free to suggest any kind of solution. If you don't, I will do my best to answer the questions for the EC design, based on all the feedback and learning from this discussion.

And feel free to comment to EC design, too. It will be published very soon.

arp7:
Cosmetic changes don't count. You have failed to address fundamental questions about what storage classes mean with a container-based storage model like what Ozone uses.

elek:
Thanks for the answer @arp7.

I am afraid we have a different definition of *cosmetic change*. When a file with 349 lines (`wc hadoop-hdds/docs/content/design/storage-class.md`) is changed with 515 lines of patch (`curl https://github.com/apache/ozone/commit/f52cd5a4f617bb5984959ea928951eec12ae9ab9.patch | wc`) I don't think it's a cosmetic change. (even if the patch file contains the context of the changes).

You may miss, but the *cosmetic change* replaced the **definition** of the storage class compared to the previous version:

```
-Let's define the *storage-class* as set of used container **types and transitions** between them during the life cycle of the containers.
+Storage class is a selection of replication mechanism and definition of parameters.
```

> You have failed to address fundamental questions about what storage classes mean with a container-based storage model like what Ozone uses.

Thanks for the feedback. With all the respect I don't think that this sentence is completely true. At least not in a way, what I could use it to learn it from. From my point of this was answer in the document and I also tried to answer the questions in the comment (or at least get quetions).

But it seems that I am definitely failed to continue an effective discussion with you, wether it's my fault or not.

I am closing this PR now, and trying to answer the open questions (how to support multiple replication method) in the EC doc. Hope we can have a good discussion there.

PR #1421, opened 2020-09-14
HDDS-4174. Add current HDDS layout version to Datanode heartbeat/registration

prashantpogde:

Add current HDDS layout version to DataNode heartbeat/registration

https://issues.apache.org/jira/browse/HDDS-4174

## How was this patch tested?
Successful Build

### DIFF
--- a/hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
+++ b/hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
@@ -74,6 +74,15 @@ enum Status {
 }
 
 /**
+ * DataNode will advertise its current layout version through
+ * heartbeat as well as registration mechanism.
+ */
+message HDDSLayoutVersionProto {
+   required uint32 metadataLayoutVersion = 1;
+   required uint32 softwareLayoutVersion = 2;
+}
+
+/**
  * Request for version info of the software stack on the server.
  */
 message SCMVersionRequestProto {}
@@ -92,6 +101,7 @@ message SCMRegisterRequestProto {
   required NodeReportProto nodeReport = 2;
   required ContainerReportsProto containerReport = 3;
   required PipelineReportsProto pipelineReports = 4;
+  optional HDDSLayoutVersionProto dataNodeLayoutVersion = 5;
 }
 
 /**
@@ -126,6 +136,7 @@ message SCMHeartbeatRequestProto {
   optional ContainerActionsProto containerActions = 6;
   optional PipelineActionsProto pipelineActions = 7;
   optional PipelineReportsProto pipelineReports = 8;
+  optional HDDSLayoutVersionProto dataNodeLayoutVersion = 9;
 }
 
 /*
@@ -293,6 +304,7 @@ message SCMCommandProto {
     replicateContainerCommand = 5;
     createPipelineCommand = 6;
     closePipelineCommand = 7;
+    finalizeNewLayoutVersionCommand = 8;
   }
   // TODO: once we start using protoc 3.x, refactor this message using "oneof"
   required Type commandType = 1;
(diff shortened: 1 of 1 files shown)


hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
```diff
@@ -390,6 +404,14 @@ message ClosePipelineCommandProto {
   required int64 cmdId = 2;
 }
 
+/**
+ * This command asks the DataNode to finalize a new HDDS layout version.
+ */
+message FinalizeNewLayoutVersionCommandProto {
+  required bool finalizeNewHDDSLayoutVersion = 1 [default = false];
```
elek, line 411: What is the use case the command where this flag=false. Is it a no-op request?

elek (approved):
Thanks the patch @prashantpogde 

LGTM. Only one question inline.

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
```diff
@@ -92,6 +101,7 @@ message SCMRegisterRequestProto {
   required NodeReportProto nodeReport = 2;
   required ContainerReportsProto containerReport = 3;
   required PipelineReportsProto pipelineReports = 4;
+  optional HDDSLayoutVersionProto dataNodeLayoutVersion = 5;
```
linyiqun, line 104: Why not this be a required field? when Datanode firstly register on SCM, it should have its own HDDS layout version.

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
```diff
@@ -303,6 +315,8 @@ message SCMCommandProto {
   optional ReplicateContainerCommandProto replicateContainerCommandProto = 6;
   optional CreatePipelineCommandProto createPipelineCommandProto = 7;
   optional ClosePipelineCommandProto closePipelineCommandProto = 8;
+  optional FinalizeNewLayoutVersionCommandProto
+  finalizeNewLayoutVersionCommandProto = 9;
```
amaliujia, line 319: Why not use a single line: `optional FinalizeNewLayoutVersionCommandProto=9`?

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
prashantpogde: yes, that would be a no op request.

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
prashantpogde: Datanode need not pass the HDDSLaoutVersionProto information during registration.

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
prashantpogde: > 80 columns

amaliujia: Ah I see. Didn't know Ozone also enforce that in proto files (very common that message name is very long).

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
prashantpogde: All new fields added should be optional for backward compatibility.

avijayanhwx (approved):
Thanks for working on this @prashantpogde, and for the reviews @elek, @linyiqun & @amaliujia.

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
elek: Is there any use case for this noop request? Will it be used by somebody? Why do we need this field?

PR #1423, opened 2020-09-15
HDDS-4244. Container deleted wrong replica cause mis-replicated.

maobaolong:

Container deleted wrong replica cause mis-replicated.

HDDS-4244

## How was this patch tested?

- Related config file
  - ozone-site.xml
```xml
  <property>
    <name>ozone.scm.container.placement.impl</name>
    <value>org.apache.hadoop.hdds.scm.container.placement.algorithms.SCMContainerPlacementRackAware</value>
  </property>
  <property>
    <name>net.topology.node.switch.mapping.impl</name>
    <value>org.apache.hadoop.net.TableMapping</value>
  </property>
  <property>
    <name>net.topology.table.file.name</name>
    <value>/data/ozoneadmin/ozoneenv/ozone/etc/hadoop/network-config</value>
  </property>
```

  - network-config
```
192.168.1.100 /racks1
192.168.1.101 /racks1
192.168.1.102 /racks1
192.168.1.103 /racks2
192.168.1.104 /racks2
192.168.1.106 /racks2
```

First you should make the 3 replicas of the tested container on the same racks, like racks1.
secondly, change the config file update the request config.

2 ReplicationManager interval, you can use the following command to verify the container replicas are in the 2 racks.
```bash
ozone admin container info #xxx
```

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -670,8 +670,9 @@ public class ReplicationManager
           // First remove the replica we are working on from the set, and then
           // check if the set is now mis-replicated.
           replicaSet.remove(r);
-          boolean nowSatisfied = getPlacementStatus(replicaSet, replicationFactor)
-              .isPolicySatisfied();
+          boolean nowSatisfied =
+              getPlacementStatus(replicaSet, replicationFactor)
+                  .isPolicySatisfied();
           if (!satisfied || nowSatisfied) {
             // Remove the replica if the container was already unsatisfied
             // OR if losing this replica still keep satisfied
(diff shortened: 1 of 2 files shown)


ChenSammi:
@maobaolong , can we add a policy unsatisfied UT too?

maobaolong:
@ChenSammi Thank you for remind me, I add a `testOverReplicatedAndPolicyUnSatisfied` test case. PTAL.

ChenSammi:
LGTM +1.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -660,21 +660,23 @@ private void handleOverReplicatedContainer(final ContainerInfo container,
...
-          boolean nowMisRep = getPlacementStatus(replicaSet, replicationFactor)
-              .isPolicySatisfied();
-          if (misReplicated || !nowMisRep) {
-            // Remove the replica if the container was already mis-replicated
-            // OR if losing this replica does not make it become mis-replicated
+          ContainerPlacementStatus nowPS =
+              getPlacementStatus(replicaSet, replicationFactor);
+          if ((!ps.isPolicySatisfied()
+                && nowPS.actualPlacementCount() == ps.actualPlacementCount())
+              || (ps.isPolicySatisfied() && nowPS.isPolicySatisfied())) {
```
nandakumar131, line 676: Why do we need such a complex condition check?
Won't fixing the previous check solve the problem?
```if (misReplicated || !nowMisRep)``` to ```if (!misReplicated || nowMisRep)```

nandakumar131:
How exactly does the new test cases verify that the ReplicationManager is considering Rack Awareness and picks the correct replica for deletion? As far as I can understand, the new test cases are checking if the delete command is sent or not. The test cases don't verify any kind of Rack Placement. Am I missing something here?

PR #1426, opened 2020-09-15
HDDS-4247. Fixed log4j usage in some places

lamber-ken:

Fixed log4j usage in some places.

Examples, `{}` is redundant.
```
LOG.error("Error while scrubbing pipelines {}", e);
```

https://issues.apache.org/jira/projects/HDDS/issues/HDDS-4247

## How was this patch tested?

not need.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
@@ -600,7 +600,7 @@ public class OzoneDelegationTokenSecretManager
 
         }
       } catch (InterruptedException ie) {
-        LOG.error("ExpiredTokenRemover received {}", ie);
+        LOG.error("ExpiredTokenRemover received.", ie);
         Thread.currentThread().interrupt();
       } catch (Exception t) {
         LOG.error("ExpiredTokenRemover thread received unexpected exception",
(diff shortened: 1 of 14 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
```diff
@@ -600,7 +600,7 @@ public void run() {
 
         }
       } catch (InterruptedException ie) {
-        LOG.error("ExpiredTokenRemover received {}", ie);
+        LOG.error("ExpiredTokenRemover received.", ie);
```
adoroszlai, line 603:
```suggestion
        LOG.info("ExpiredTokenRemover was interrupted.", ie);
```

adoroszlai (commented):
Thanks @lamber-ken for fixing these placeholder issues.  LGTM, but I think the message in `ExpiredTokenRemover` is incomplete.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
lamber-ken: @adoroszlai done, thanks

adoroszlai (approved):

adoroszlai:
Thanks @lamber-ken for the contribution.

PR #1428, opened 2020-09-16
HDDS-4192: enable SCM Raft Group based on config ozone.scm.names

GlenGeng:

Say ozone.scm.names is `ip1,ip2,ip3`, scm with `ip1` identifies its RaftPeerId as `scm1`,  scm with `ip2` identifies its RaftPeerId as `scm2`, scm with `ip3` identifies its RaftPeerId as `scm3`. They will automatically become a raft group.

https://issues.apache.org/jira/browse/HDDS-4192

## How was this patch tested?

CI


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -59,18 +68,62 @@
...
+    List<String> scmHosts =
+        Arrays.stream(conf.getTrimmedStrings(ScmConfigKeys.OZONE_SCM_NAMES))
+            .map(scmName -> HddsUtils.getHostName(scmName).get())
+            .collect(Collectors.toList());
+
     this.address = haConf.getRatisBindAddress();
-    final RaftPeer localRaftPeer = new RaftPeer(raftPeerId, address);
+    InetAddress localHost = InetAddress.getLocalHost();
+
+    int selfIndex = -1;
```
timmylicheng, line 89: Could we move this self validation to an overall IP validation method so that we could throw in more validation once we are getting close to the ultimate HA config version?

timmylicheng:
LGTM. +1. Merging.

PR #1430, opened 2020-09-16
HDDS-4227. Implement a 'Prepare For Upgrade' step in OM that applies all committed Ratis transactions.

avijayanhwx:

Through HDDS-4143, we have a generic factory to handle multiple versions of apply transaction implementations based on layout version. Hence, this factory can be used to handle versioned requests across layout versions, whenever both the versions need to exist in the code (Let's say for HDDS-2939).

However, it has been noticed that the OM ratis requests are still undergoing lot of minor changes (HDDS-4006, HDDS-4007, HDDS-3903), and in these cases it will become hard to maintain 2 versions of the code just to support clean upgrades.

Hence, this patch adds a pre-upgrade utility (client API) that makes sure that an OM instance has no "un-applied" transactions in its Raft log. Invoking this client API before updating the software version makes sure that the upgrade starts with a clean state. Of course, this would be needed only in a HA setup. In a non HA setup, this can either be skipped, or when invoked will be a No-Op (Non Ratis) or cause no harm (Single node Ratis).

More details in JIRA.

https://issues.apache.org/jira/browse/HDDS-4227

## How was this patch tested?
Manually tested.
Unit/Acceptance tests pending.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisUpgradeUtils.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisUpgradeUtils.java
@@ -51,9 +51,11 @@ public final class RatisUpgradeUtils {
       StateMachine stateMachine,
       RaftGroup raftGroup,
       RaftServerProxy server,
-      long maxTimeToWaitSeconds)
+      long maxTimeToWaitSeconds,
+      long timeBetweenRetryInSeconds)
       throws InterruptedException, IOException {
 
+    long intervalTime = TimeUnit.SECONDS.toMillis(timeBetweenRetryInSeconds);
     long endTime = System.currentTimeMillis() +
         TimeUnit.SECONDS.toMillis(maxTimeToWaitSeconds);
     boolean success = false;
@@ -62,12 +64,13 @@ public final class RatisUpgradeUtils {
       if (success) {
         break;
       }
-      Thread.sleep(5000L);
+      Thread.sleep(intervalTime);
     }
 
     if (!success) {
-      throw new IOException(String.format("After waiting for %d seconds, OM " +
-          "has not applied  all the transactions.", maxTimeToWaitSeconds));
+      throw new IOException(String.format("After waiting for %d seconds, " +
+          "State Machine has not applied  all the transactions.",
+          maxTimeToWaitSeconds));
     }
 
     long snapshotIndex = stateMachine.takeSnapshot();
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisUpgradeUtils.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisUpgradeUtils.java
@@ -1,0 +1,93 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership.  The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.hadoop.hdds.ratis;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.ratis.protocol.RaftGroup;
+import org.apache.ratis.server.impl.RaftServerImpl;
+import org.apache.ratis.server.impl.RaftServerProxy;
+import org.apache.ratis.statemachine.StateMachine;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Ratis utility functions.
+ */
+public final class RatisUpgradeUtils {
+
+  private RatisUpgradeUtils() {
+  }
+
+  private static final Logger LOG =
+      LoggerFactory.getLogger(RatisUpgradeUtils.class);
+
+  /**
+   * Flush all committed transactions in a given Raft Server for a given group.
+   * @param stateMachine state machine to use
+   * @param raftGroup raft group
+   * @param server Raft server proxy instance.
+   * @param maxTimeToWaitSeconds Max time to wait before declaring failure.
+   * @throws InterruptedException when interrupted
+   * @throws IOException on error while waiting
+   */
+  public static void waitForAllTxnsApplied(
+      StateMachine stateMachine,
+      RaftGroup raftGroup,
+      RaftServerProxy server,
+      long maxTimeToWaitSeconds)
+      throws InterruptedException, IOException {
+
+    long endTime = System.currentTimeMillis() +
+        TimeUnit.SECONDS.toMillis(maxTimeToWaitSeconds);
+    boolean success = false;
+    while (System.currentTimeMillis() < endTime) {
+      success = checkIfAllTransactionsApplied(stateMachine, server, raftGroup);
+      if (success) {
+        break;
+      }
+      Thread.sleep(5000L);
+    }
+
+    if (!success) {
+      throw new IOException(String.format("After waiting for %d seconds, OM " +
+          "has not applied  all the transactions.", maxTimeToWaitSeconds));
+    }
+
+    long snapshotIndex = stateMachine.takeSnapshot();
+    if (snapshotIndex != stateMachine.getLastAppliedTermIndex().getIndex()) {
+      throw new IOException("Index from Snapshot does not match last applied " +
+          "Index");
+    }
+  }
+
+  private static boolean checkIfAllTransactionsApplied(
+      StateMachine stateMachine,
+      RaftServerProxy serverProxy,
+      RaftGroup raftGroup) throws IOException {
+    LOG.info("Checking for pending transactions to be applied.");
+    RaftServerImpl impl = serverProxy.getImpl(raftGroup.getGroupId());
+    long lastCommittedIndex = impl.getState().getLog().getLastCommittedIndex();
+    long appliedIndex = stateMachine.getLastAppliedTermIndex().getIndex();
+    LOG.info("lastCommittedIndex = {}, appliedIndex = {}",
+        lastCommittedIndex, appliedIndex);
+    return (lastCommittedIndex == appliedIndex);
+  }
+
+}
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -249,6 +249,5 @@ public final class OMConfigKeys {
 
   public static final long OZONE_OM_MAX_TIME_TO_WAIT_FLUSH_TXNS =
       TimeUnit.MINUTES.toSeconds(5);
-  public static final long OZONE_OM_FLUSH_TXNS_RETRY_INTERVAL_SECONDS =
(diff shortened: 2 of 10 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1179,15 +1229,22 @@ public void start() throws IOException {
       // Allow OM to start as Http Server failure is not fatal.
       LOG.error("OM HttpServer failed to start.", ex);
     }
-    omRpcServer.start();
-    isOmRpcServerRunning = true;
 
+    if (!prepareForUpgrade) {
+      omRpcServer.start();
+      isOmRpcServerRunning = true;
+    }
```
linyiqun, line 1236: During prepareForUpgrade, the RPC server is not stared. So we should also have the corresponding command to trigger to restart RPC server.

linyiqun (commented):
Only some minor comments above.
In additional, can you add the new unit test for PrepareForUpgrade scenario? Current unit test change cannot cover this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManagerStarter.java
```diff
@@ -98,6 +98,28 @@ public void initOm()
     }
   }
 
+
+  /**
+   * This function implements a sub-command to allow the OM to be
+   * "prepared for upgrade".
+   */
+  @CommandLine.Command(name = "--prepareForUpgrade",
+      aliases = {"--prepareForDowngrade", "--flushTransactions"},
```
linyiqun, line 107: Will prepareForUpgrade command send multiple OMs simultaneously here? Or we should trigger prepareForUpgrade command for each OM service.

fapifta:
Hi @avijayanhwx 

the initial changes look good to me thank you for sharing the wip state, and it seems to be a good direction so far.
Maybe one minor comment from me as well, are we sure we want to add --prepareForDowngrade as an option alias? It suggests that we might support downgrade, and I fear that might cause some misunderstandings.
I am unsure whether we can use the same functionality to get back to an older version in all scenarios either.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -994,6 +1005,45 @@ public static boolean omInit(OzoneConfiguration conf) throws IOException,
...
+        .getProperties().get(PURGE_UPTO_SNAPSHOT_INDEX_KEY);
+    if (!Boolean.parseBoolean(purgeConfig)) {
+      throw new IllegalStateException("Cannot prepare OM for Upgrade since  " +
+          "raft.server.log.purge.upto.snapshot.index is not true");
+    }
+
+    waitForAllTxnsApplied(omRatisServer.getOmStateMachine(),
+        omRatisServer.getRaftGroup(),
+        (RaftServerProxy) omRatisServer.getServer(),
+        TimeUnit.MINUTES.toSeconds(5));
```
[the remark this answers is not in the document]
fapifta: Are you sure we want to add a configuration for this one? I would argue we do not need one more configurable thing to this one at least.
prepareForUpgrade is a special startup type of OM, during which it applies all transactions that are in the raft log.
If 5 minutes is not enough to apply all transactions in the raft log, then the process will shut down and let the user know that some of the transactions were not applied, so that the user can start the process again as a last resort, to apply further transactions. If we assume that at least a few transactions are applied sooner or later the user can get to a state where everything is applied, and if none of the transactions can be applied within 5 minutes, that sounds like a serious problem anyways, independently from the upgrade.

Also in 5 minutes I would expect in all cases that the unapplied transactions can be applied, as the number of this kind of transactions should not be too much as far as I know, or if it is then the system is not healthy anyway.

Can you please elaborate, why would it be useful to make this configurable?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
fapifta: As we discussed this with @avijayanhwx during internal design discussions, after OM is started in prepareForUpgrade mode, it will tear down, when the last transaction is applied from the raft log, and a snapshot is taken in raft, so with that the OM reached a state when all transactions are applied and none needs to be applied after the next startup.

This is to ensure that all the transactions are applied with the code that was there when the transactions arrived in, so with that we can ensure consistency of the state of different OM instances.

After this is finished, and OM tear down from prepareForUpgrade, one will need a normal startup of OM to bring it up again, and at that time the RPC server will start properly.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManagerStarter.java
fapifta: This command should be issued when the OM is already stopped before the upgrade of software bits. This is a command that starts up the OM code in a special way, with that it can start up only the current local OM, as I understand.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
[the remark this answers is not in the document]
linyiqun: >Also in 5 minutes I would expect in all cases that the unapplied transactions can be applied, as the number of this kind of transactions should not be too much as far as I know, or if it is then the system is not healthy anyway.

I'm okay to let 5 mins as current threshold wait time, only one minor comment: can we define a variable for this time value rather hard-coded in method here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
linyiqun: Okay, enable RPC server via next startup makes sense to me.

hanishakoneru (commented):
Thanks @avijayanhwx for working on this and driving the upgrade feature. 
LGTM overall. A minor comment. Are you planning to add unit/acceptance test in this Jira?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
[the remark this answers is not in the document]
avijayanhwx: Thanks, will change it to a variable.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManagerStarter.java
avijayanhwx: +1 to @fapifta's reply.

avijayanhwx:
> Hi @avijayanhwx
> 
> the initial changes look good to me thank you for sharing the wip state, and it seems to be a good direction so far.
> Maybe one minor comment from me as well, are we sure we want to add --prepareForDowngrade as an option alias? It suggests that we might support downgrade, and I fear that might cause some misunderstandings.
> I am unsure whether we can use the same functionality to get back to an older version in all scenarios either.

Thanks for the review @fapifta. I agree "downgrades" may not be supportable from each any arbitrary state (post finalize). Since this is a specialized command, I see no harm in adding the alias. When the downgrade is not supported, starting the component in the older version will be flagged out anyway. And, this prepare step is a must do for any supported downgrade. In the future, we may be able to add a bit of validation on the prepareDowngrade step to check finalization state or some marker on disk.

avijayanhwx:
> Only some minor comments above.
> In additional, can you add the new unit test for PrepareForUpgrade scenario? Current unit test change cannot cover this.

Thanks for the review @linyiqun. I am planning to write tests in my next commit and bring this PR out of "Draft" status.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -994,6 +1005,45 @@ public static boolean omInit(OzoneConfiguration conf) throws IOException,
...
+      LOG.info("Ratis not enabled. Nothing to do.");
+      return true;
+    }
+
+    String purgeConfig = omRatisServer.getServer()
+        .getProperties().get(PURGE_UPTO_SNAPSHOT_INDEX_KEY);
+    if (!Boolean.parseBoolean(purgeConfig)) {
+      throw new IllegalStateException("Cannot prepare OM for Upgrade since  " +
+          "raft.server.log.purge.upto.snapshot.index is not true");
+    }
```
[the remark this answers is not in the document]
avijayanhwx: Maybe not. In the Ratis StateMachineUpdater, the takeSnapshot() method uses this config to  purge logs immediately after taking a state machine snapshot. Hence, I thought that it was good to have a check in place to make sure no one changes the config from within. I am ok with removing it for now, and handling snapshot + log purge in HDDS-4268 as a follow up patch.

avijayanhwx:
There are some follow up work items for Prepare for Upgrade flow. 

- Purging Logs after applying the last txn.
- Validating that an OM is "prepared for upgrade" when starting of in a newer version.
- Acceptance Tests

These work items will be handled in subsequent PRs.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
```diff
@@ -246,4 +246,9 @@ private OMConfigKeys() {
       "ozone.om.enable.filesystem.paths";
   public static final boolean OZONE_OM_ENABLE_FILESYSTEM_PATHS_DEFAULT =
       false;
+
+  public static final long OZONE_OM_MAX_TIME_TO_WAIT_FLUSH_TXNS =
+      TimeUnit.MINUTES.toSeconds(5);
+  public static final long OZONE_OM_FLUSH_TXNS_RETRY_INTERVAL_SECONDS =
+      TimeUnit.SECONDS.toMillis(5);
```
linyiqun, line 253: For OZONE_OM_FLUSH_TXNS_RETRY_INTERVAL_SECONDS , here should be 5. On RatisUpgradeUtils#waitForAllTxnsApplied, we already do the convert:
>long intervalTime = TimeUnit.SECONDS.toMillis(timeBetweenRetryInSeconds);

linyiqun (commented):
@avijayanhwx , thanks for addressing the comments! One comment below.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
avijayanhwx: Thank you, fixed!

linyiqun (approved):
The failed build check seems not related. +1.

fapifta:
Hi @avijayanhwx 

sorry for the long silence, the changes look good to me with the follow up items, +1.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisUpgradeUtils.java
```diff
@@ -0,0 +1,96 @@
...
+      RaftServerProxy server,
+      long maxTimeToWaitSeconds,
+      long timeBetweenRetryInSeconds)
+      throws InterruptedException, IOException {
+
+    long intervalTime = TimeUnit.SECONDS.toMillis(timeBetweenRetryInSeconds);
+    long endTime = System.currentTimeMillis() +
+        TimeUnit.SECONDS.toMillis(maxTimeToWaitSeconds);
+    boolean success = false;
+    while (System.currentTimeMillis() < endTime) {
```
swagle, line 62: Wouldn't this always be true? [ curr < curr + num ]

avijayanhwx: The 'curr' in the RHS was assigned before the while loop and hence does not change, the LHS 'curr' moves forward.

swagle (approved):
+1 LGTM

avijayanhwx:
Thank you for the reviews @linyiqun, @fapifta & @swagle. Merging this.

PR #1431, opened 2020-09-17
HDDS-4254. Bucket space: add usedBytes and update it when create and delete key.

captainzmc:

First, we increase usedBytes of Bucket. Later, we will judge whether the Bucket can be written based on this when we write the key.
Volume has implemented increase usedBytes when write, and this is based on HDDS-4053.

https://issues.apache.org/jira/browse/HDDS-4254

## How was this patch tested?

added UT

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMAllocateBlockResponse.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMAllocateBlockResponse.java
@@ -19,6 +19,7 @@
 package org.apache.hadoop.ozone.om.response.key;
 
 import org.apache.hadoop.ozone.om.OMMetadataManager;
+import org.apache.hadoop.ozone.om.helpers.OmBucketInfo;
 import org.apache.hadoop.ozone.om.helpers.OmKeyInfo;
 import org.apache.hadoop.ozone.om.helpers.OmVolumeArgs;
 import org.apache.hadoop.ozone.om.response.CleanupTableInfo;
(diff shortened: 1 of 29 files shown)


captainzmc:
Here are the freon test results:
Using 100 threads to write data to different buckets under the same volume, adding quota Usage has little impact on performance. (LongAdder disperses the concurrency of a single cell to each cell, improving concurrency efficiency compared to atomicLong). In my test, I used three virtual machines, each with a key size of 10K.
image

captainzmc:
hi @ChenSammi @xiaoyuyao, this PR is base on #1296. Can you help to review this?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMAllocateBlockResponse.java
```diff
@@ -72,5 +76,10 @@ public void addToDBBatch(OMMetadataManager omMetadataManager,
     omMetadataManager.getVolumeTable().putWithBatch(batchOperation,
         omMetadataManager.getVolumeKey(omVolumeArgs.getVolume()),
         omVolumeArgs);
+    // update bucket usedBytes.
+    omMetadataManager.getBucketTable().putWithBatch(batchOperation,
+        omMetadataManager.getBucketKey(omVolumeArgs.getVolume(),
+            omBucketInfo.getBucketName()),
+        omBucketInfo);
```
ChenSammi, line 83: the last two lines can be merged into one line.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeyCommitResponse.java
```diff
@@ -77,6 +80,11 @@ public void addToDBBatch(OMMetadataManager omMetadataManager,
     omMetadataManager.getVolumeTable().putWithBatch(batchOperation,
         omMetadataManager.getVolumeKey(omVolumeArgs.getVolume()),
         omVolumeArgs);
+    // update bucket usedBytes.
+    omMetadataManager.getBucketTable().putWithBatch(batchOperation,
+        omMetadataManager.getBucketKey(omVolumeArgs.getVolume(),
+            omBucketInfo.getBucketName()),
+        omBucketInfo);
```
ChenSammi, line 87: same as above.

captainzmc:
Thanks @ChenSammi for the review. Review issues has been fixed.

ChenSammi:
LGTM,  +1.

PR #1433, opened 2020-09-17
HDDS-4255. Remove unused Ant and Jdiff dependency versions

elek:

Versions of Ant and JDiff are not used in ozone project, but we have some version declaration (inherited from the Hadoo parent pom which was used as a base for the main pom.xml).

As the (unused) ANT version has security issues, I would remove them to avoid any confusion

https://issues.apache.org/jira/browse/HDDS-4255

## How was this patch tested?

Clean build + `rg org.apache.ant pom.xml`


lamber-ken:
retest

adoroszlai (approved):

adoroszlai:
Thanks @elek for the cleanup.

PR #1434, opened 2020-09-17
HDDS-3727. Volume space: check quotaUsageInBytes when write key.

captainzmc:

Volume has implemented increase usedBytes when write, and this PR is based on #1296.
In this PR we judge whether the Volume can be written when we write the key if the volume space quota enable.

https://issues.apache.org/jira/browse/HDDS-3727

## How was this patch tested?

UT added.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/exceptions/OMException.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/exceptions/OMException.java
@@ -229,7 +229,9 @@ public class OMException extends IOException {
 
     NOT_SUPPORTED_OPERATION,
 
-    PARTIAL_RENAME
+    PARTIAL_RENAME,
+
+    QUOTA_EXCEEDED
 
   }
 }
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -755,7 +755,7 @@ public abstract class TestOzoneRpcClientAbstract {
     try {
       OzoneOutputStream out = bucket.createKey(UUID.randomUUID().toString(),
           valueLength, STAND_ALONE, ONE, new HashMap<>());
-      for(int i=0; i <= blockSize/value.length(); i++) {
+      for (int i = 0; i <= blockSize / value.length(); i++) {
         out.write(value.getBytes());
       }
       out.close();
@@ -773,7 +773,7 @@ public abstract class TestOzoneRpcClientAbstract {
     try {
       OzoneOutputStream out = bucket.createKey(UUID.randomUUID().toString(),
           valueLength, STAND_ALONE, ONE, new HashMap<>());
-      for(int i=0; i <= (4 * blockSize)/value.length(); i++) {
+      for (int i = 0; i <= (4 * blockSize) / value.length(); i++) {
         out.write(value.getBytes());
       }
       out.close();
@@ -781,8 +781,7 @@ public abstract class TestOzoneRpcClientAbstract {
       countException++;
       GenericTestUtils.assertExceptionContains("QUOTA_EXCEEDED", ex);
     }
-    // AllocateBlock failed, volume usedBytes should be 4 * blockSize + 1 *
-    // blockSize = 5 * blockSize.
+    // AllocateBlock failed, volume usedBytes should be (4 + 1) * blockSize
     Assert.assertEquals(5 * blockSize,
         store.getVolume(volumeName).getUsedBytes());
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -749,7 +749,7 @@ public abstract class TestOzoneRpcClientAbstract {
     // Write failed, volume usedBytes should be 0
     Assert.assertEquals(0L, store.getVolume(volumeName).getUsedBytes());
 
-    // Write large key, test allocateBlock fails.
+    // Write a key(with two blocks), test allocateBlock fails.
     store.getVolume(volumeName).setQuota(
         OzoneQuota.parseQuota(blockSize + "Bytes", 100));
     try {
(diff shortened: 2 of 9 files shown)


captainzmc:
hi @ChenSammi @xiaoyuyao, this PR add judge whether we can be write if space quota enable. Can you help to review this?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/exceptions/OMException.java
```diff
@@ -229,7 +229,9 @@ public String toString() {
 
     NOT_SUPPORTED_OPERATION,
 
-    PARTIAL_RENAME
+    PARTIAL_RENAME,
+
+    QUOTA_CHECK_ERROR
```
ChenSammi, line 234: QUOTA_CHECK_ERROR  -> QUOTA_EXCEED

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -314,6 +314,8 @@ enum Status {
 
     PARTIAL_RENAME = 65;
 
+    QUOTA_CHECK_ERROR = 66;
```
ChenSammi, line 317: same as above.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -707,6 +707,62 @@ public void testPutKey() throws IOException {
     }
   }
 
+  @Test
+  public void testCheckUsedBytesQuota() throws IOException {
```
ChenSammi, line 711: Can we add used bytes check in each test case?

captainzmc:
Thanks for @ChenSammi ‘s review. The review issues have been fixed.  Could you help take another look?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -277,6 +277,16 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           .collect(Collectors.toList());
       omKeyInfo.appendNewBlocks(newLocationList, false);
 
+      omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
+      omBucketInfo = getBucketInfo(omMetadataManager, volumeName, bucketName);
+      // check volume quota
+      long preAllocatedSpace = newLocationList.size()
+          * ozoneManager.getScmBlockSize()
+          * omKeyInfo.getFactor().getNumber();
+      if (omVolumeArgs.getQuotaInBytes() > OzoneConsts.QUOTA_RESET) {
```
ChenSammi, line 286: Can we add a isQuotaInBytesSet check in OmVolumeArgs and use this function instead?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -227,6 +232,18 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+          OMException.ResultCodes.QUOTA_EXCEEDED.toString())) {
+        long keyAllocatedSpace = openKeyInfo.getLatestVersionLocations()
+            .getLocationListCount() * ozoneManager.getScmBlockSize()
+            * openKeyInfo.getFactor().getNumber();
+        // Update usedBytes atomically. ErrorOMResponse does not persist the DB,
+        // so we update the cache first. The next time another key is written,
+        // volume Args will be persisted to the DB.
+        // TODO: There is a delay in updating DB in this way, and if necessary
+        //  we can modify the ErrorOMResponse to avoid it.
+        omVolumeArgs.getUsedBytes().add(-keyAllocatedSpace);
```
ChenSammi, line 245: If A object has multiple blocks, and quota_exceeded exception is thrown out at the last block, shall we only deduct the block space size at this step?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -707,9 +707,71 @@ public void testPutKey() throws IOException {
...
+      }
+      out.close();
+    } catch (IOException ex) {
+      countException++;
+      GenericTestUtils.assertExceptionContains("QUOTA_EXCEEDED", ex);
+    }
+    // AllocateBlock failed, volume usedBytes should be 0
+    Assert.assertEquals(0L, store.getVolume(volumeName).getUsedBytes());
+
+    Assert.assertEquals(3, countException);
```
ChenSammi, line 769: Can we add a multiple block write case here，so the first few block will succeed， while the later block will fail due to quota exceed?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
```diff
@@ -533,6 +533,26 @@ protected FileEncryptionInfo getFileEncryptionInfo(KeyArgs keyArgs) {
...
+    long usedBytes = omVolumeArgs.getUsedBytes().sum();
+    long quotaInBytes = omVolumeArgs.getQuotaInBytes();
+    if (quotaInBytes - usedBytes < allocateSize) {
+      throw new OMException("The DiskSpace quota of volume:"
+          + omVolumeArgs.getVolume() + "exceeded: quotaInBytes: "
+          + quotaInBytes + " Bytes but diskspace consumed: " + (usedBytes
+          + allocateSize) + " Bytes.",
+          OMException.ResultCodes.QUOTA_EXCEEDED);
+    }
+
```
ChenSammi, line 553: unnessary empty line.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -239,8 +256,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
     auditLog(auditLogger, buildAuditMessage(OMAction.ALLOCATE_BLOCK, auditMap,
         exception, getOmRequest().getUserInfo()));
 
-
-
     return omClientResponse;
   }
+
```
ChenSammi, line 261: unnessary blank line.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
captainzmc: I'm going to put the if in checkVolumeQuotaInBytes, which might seem a little more compact.

captainzmc:
Thanks @ChenSammi for the review. Review issues has been fixed.

ChenSammi:
LGTM +1.

PR #1436, opened 2020-09-18
HDDS-4262. Use ClientID and CallID from Rpc Client to detect retry requests

bharatviswa504:

Use clientId and callID from the Client to uniquely identify the requests.
This will help in the case of retry requests.

(In this Jira, not changed default retry cache, the default is 1 minute using it same for now.)

https://issues.apache.org/jira/browse/HDDS-4262

## How was this patch tested?

Added one Integration test.
Able to get RetryCache and not able to acess methods like queryCache to check like whether retry(As this method is not public, for this we need to make a change in Ratis for that, so just tested this at OM level)


hanishakoneru:
Thanks Bharat. LGTM. +1.

bharatviswa504:

PR #1437, opened 2020-09-21
HDDS-4222: [OzoneFS optimization] Provide a mechanism for efficient path lookup

rakeshadr:

With the new file system HDDS-2939 like semantics design it requires multiple DB lookups to traverse the path component in top-down fashion. This patch provides a metadata caching service to cache directories for faster lookup. This patch is developed as an independent module and needs to be integrated once HDDS-2949 is pushed in.

## How was this patch tested?

Added unit test cases to verify cache init and other ops.

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2523,7 +2523,7 @@
   </property>
 
   <property>
-    <name>ozone.om.metadata.cache.directory</name>
+    <name>ozone.om.metadata.cache.directory.policy</name>
     <tag>OZONE, OM</tag>
     <value>DIR_LRU</value>
     <description>Configure the cache policy to provide caching of the
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2536,7 +2536,7 @@
   <property>
     <name>ozone.om.metadata.cache.directory.init.capacity</name>
     <tag>OZONE, OM</tag>
-    <value>DIR_LRU</value>
+    <value>100000</value>
     <description>Configure the minimum total size for directory cache.
     </description>
   </property>
@@ -2544,7 +2544,7 @@
   <property>
     <name>ozone.om.metadata.cache.directory.max.capacity</name>
     <tag>OZONE, OM</tag>
-    <value>DIR_LRU</value>
+    <value>5000000</value>
     <description>Specifies the maximum number of entries the cache may
       contain in the directory cache.
     </description>
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2521,4 +2521,32 @@
       filesystem semantics.
     </description>
   </property>
+
+  <property>
+    <name>ozone.om.metadata.cache.directory</name>
+    <tag>OZONE, OM</tag>
+    <value>DIR_LRU</value>
+    <description>Configure the cache policy to provide caching of the
+      directories for faster traversal of path components. Available values
+      are DIR_LRU and DIR_NOCACHE. As name says, DIR_LRU uses least recently
+      used eviction mechanism and DIR_NOCACHE has no caching.
+    </description>
+  </property>
+
+  <property>
+    <name>ozone.om.metadata.cache.directory.init.capacity</name>
+    <tag>OZONE, OM</tag>
+    <value>DIR_LRU</value>
+    <description>Configure the minimum total size for directory cache.
+    </description>
+  </property>
+
+  <property>
+    <name>ozone.om.metadata.cache.directory.max.capacity</name>
+    <tag>OZONE, OM</tag>
+    <value>DIR_LRU</value>
+    <description>Specifies the maximum number of entries the cache may
+      contain in the directory cache.
+    </description>
+  </property>
 </configuration>
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -246,4 +246,15 @@ public final class OMConfigKeys {
       "ozone.om.enable.filesystem.paths";
   public static final boolean OZONE_OM_ENABLE_FILESYSTEM_PATHS_DEFAULT =
       false;
+
+  public static final String OZONE_OM_CACHE_DIR_POLICY =
+          "ozone.om.metadata.cache.directory";
+  public static final String OZONE_OM_CACHE_DIR_DEFAULT = "DIR_LRU";
+  public static final String OZONE_OM_CACHE_DIR_INIT_CAPACITY =
+          "ozone.om.metadata.cache.directory.init.capacity";
+  public static final int OZONE_OM_CACHE_DIR_INIT_CAPACITY_DEFAULT = 100000;
+  public static final String OZONE_OM_CACHE_DIR_MAX_CAPACITY =
+          "ozone.om.metadata.cache.directory.max.capacity";
+  public static final long OZONE_OM_CACHE_DIR_MAX_CAPACITY_DEFAULT = 5000000;
+
 }
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -249,7 +249,7 @@ public final class OMConfigKeys {
 
   public static final String OZONE_OM_CACHE_DIR_POLICY =
           "ozone.om.metadata.cache.directory.policy";
-  public static final String OZONE_OM_CACHE_DIR_DEFAULT = "DIR_LRU";
+  public static final String OZONE_OM_CACHE_DIR_POLICY_DEFAULT = "DIR_LRU";
   public static final String OZONE_OM_CACHE_DIR_INIT_CAPACITY =
           "ozone.om.metadata.cache.directory.init.capacity";
   public static final int OZONE_OM_CACHE_DIR_INIT_CAPACITY_DEFAULT = 100000;
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -248,7 +248,7 @@ public final class OMConfigKeys {
       false;
 
   public static final String OZONE_OM_CACHE_DIR_POLICY =
-          "ozone.om.metadata.cache.directory";
+          "ozone.om.metadata.cache.directory.policy";
   public static final String OZONE_OM_CACHE_DIR_DEFAULT = "DIR_LRU";
   public static final String OZONE_OM_CACHE_DIR_INIT_CAPACITY =
           "ozone.om.metadata.cache.directory.init.capacity";
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/CacheEntity.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/CacheEntity.java
@@ -38,7 +38,7 @@ public enum CacheEntity {
 
   public static CacheEntity getEntity(String entityStr) {
     for (CacheEntity entity : CacheEntity.values()) {
-      if (entityStr.equalsIgnoreCase(entity.getName())) {
+      if (entity.getName().equalsIgnoreCase(entityStr)) {
         return entity;
       }
     }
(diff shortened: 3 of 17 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/CacheStore.java
```diff
@@ -0,0 +1,72 @@
...
+ * the cache.
+ * <p>
+ * TODO: can define specific 'CacheLoader' to handle the OM restart and
+ *       define cache loading strategies. It can be NullLoader, LazyLoader,
+ *       LevelLoader etc.
+ *
+ * @param <CACHEKEY>
+ * @param <CACHEVALUE>
+ */
+public interface CacheStore<CACHEKEY extends OMCacheKey,
```
linyiqun, line 32: Besides basical put/get/remove interface, we should also have other interface defined like cache hit/miss count for cache store in the future..

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/DirectoryLRUCacheStore.java
```diff
@@ -0,0 +1,87 @@
...
+            OMConfigKeys.OZONE_OM_CACHE_DIR_MAX_CAPACITY, maxSize);
+    mCache = CacheBuilder.newBuilder()
+            .initialCapacity(initSize)
+            .maximumSize(maxSize)
+            .build();
+  }
+
+  @Override
+  public void put(OMCacheKey key, OMCacheValue value) {
+    mCache.put(key, value);
```
linyiqun, line 65: Can we add the null check for key before putting it into cache?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/DirectoryLRUCacheStore.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @Override
+  public OMCacheValue get(OMCacheKey key) {
+    return mCache.getIfPresent(key);
+  }
+
+  @Override
+  public void remove(OMCacheKey key) {
```
linyiqun, line 74: Same comment like above.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/OMMetadataCacheFactory.java
```diff
@@ -0,0 +1,120 @@
...
+    switch (entity) {
+    case DIR:
+      OMMetadataCacheProvider provider = new OMDirectoryCacheProvider(config,
+              cachePolicy);
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("CacheStore initialized with {}:" + provider.getEntity());
+      }
+      return provider.getCache();
+    default:
+      return null;
```
linyiqun, line 66: How about throw error when cache store cannot be initialized with given cache policy? We would be better not let cache store as null returned. If cache store returned null, that means all operations from this cache store will be failed. Otherwise, we have to do empty check for each operation when cache store is used.

linyiqun (commented):
Hi @rakeshadr , some initial review comments below.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
```diff
@@ -246,4 +246,15 @@ private OMConfigKeys() {
       "ozone.om.enable.filesystem.paths";
   public static final boolean OZONE_OM_ENABLE_FILESYSTEM_PATHS_DEFAULT =
       false;
+
+  public static final String OZONE_OM_CACHE_DIR_POLICY =
+          "ozone.om.metadata.cache.directory";
+  public static final String OZONE_OM_CACHE_DIR_DEFAULT = "DIR_LRU";
```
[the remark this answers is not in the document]
rakeshadr: OK, will update.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/CacheEntity.java
```diff
@@ -0,0 +1,48 @@
...
+
+  private String entityName;
+
+  public String getName() {
+    return entityName;
+  }
+
+  public static CacheEntity getEntity(String entityStr) {
+    for (CacheEntity entity : CacheEntity.values()) {
+      if (entityStr.equalsIgnoreCase(entity.getName())) {
```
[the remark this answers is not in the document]
rakeshadr: OK, will update.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/CacheStore.java
rakeshadr: Good point. I thought of this and how about to do this via followup task.
For now, I will add a TODO item. Hope thats fine for you.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/DirectoryLRUCacheStore.java
rakeshadr: Sure, will update it

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/DirectoryLRUCacheStore.java
rakeshadr: OK

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/cache/OMMetadataCacheFactory.java
rakeshadr: OK, will add UTs to cover this case.

rakeshadr (commented):
> Hi @rakeshadr , some initial review comments below.
> In additional, one question from me: this is the first task of dir cache, there will be other further subtasks. But this part of work depends on [HDDS-2939](https://issues.apache.org/jira/browse/HDDS-2939) be completed. So that means [HDDS-4222](https://issues.apache.org/jira/browse/HDDS-4222) tasks cannot be merged immediately and be blocked for a long time? How do we plan to coordinate with [HDDS-2939](https://issues.apache.org/jira/browse/HDDS-2939) and [HDDS-4222](https://issues.apache.org/jira/browse/HDDS-4222) feature development works?

Good comment. yes, cache should be integrated eventually(case-by-case) to get full benefit. But HDDS-4222 can be upstreamed separately and not blocked, I feel.

Cache can be integrated once HDDS-2949 is finished. During dir creation, it checks the [DIR_EXISTS](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java#L185) and cache can be integrated into this call path. 

Later, while implementing File, lookups tasks, delete, rename etc will integrate into that areas.

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -2521,4 +2521,32 @@
       filesystem semantics.
     </description>
   </property>
+
+  <property>
+    <name>ozone.om.metadata.cache.directory</name>
```
linyiqun, line 2526: This name also needs to be updated, current unit test was broken by this
> TestOzoneConfigurationFields>TestConfigurationFieldsBase.testCompareConfigurationClassAgainstXml:493 class org.apache.hadoop.ozone.OzoneConfigKeys class org.apache.hadoop.hdds.scm.ScmConfigKeys class org.apache.hadoop.ozone.om.OMConfigKeys class org.apache.hadoop.hdds.HddsConfigKeys class org.apache.hadoop.ozone.recon.ReconServerConfigKeys class org.apache.hadoop.ozone.s3.S3GatewayConfigKeys class org.apache.hadoop.hdds.scm.server.SCMHTTPServerConfig has 1 variables missing in ozone-default.xml Entries:   ozone.om.metadata.cache.directory.policy expected:<0> but was:<1>
[ERROR]   TestOzoneConfigurationFields>TestConfigurationFieldsBase.testCompareXmlAgainstConfigurationClass

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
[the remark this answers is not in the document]
linyiqun: Looks like OZONE_OM_CACHE_DIR_DEFAULT was not updated to OZONE_OM_CACHE_DIR_POLICY_DEFAULT.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/cache/TestOMMetadataCache.java
```diff
@@ -0,0 +1,276 @@
...
+
+    omMetadataManager = new OmMetadataManagerImpl(conf);
+    CacheStore dirCacheStore =
+            omMetadataManager.getOMCacheManager().getDirCache();
+    Assert.assertEquals("CachePolicy Mismatches!", CachePolicy.DIR_NOCACHE,
+            dirCacheStore.getCachePolicy());
+  }
+
+  @Test
+  public void testInvalidCacheDirectoryPolicyConfigurationName() throws IOException {
```
linyiqun, line 258: Checkstyle issue of long line length
>[ERROR] src/test/java/org/apache/hadoop/ozone/om/cache/TestOMMetadataCache.java:[258] (sizes) LineLength: Line is longer than 80 characters (found 85).

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/cache/TestOMMetadataCache.java
```diff
@@ -0,0 +1,276 @@
...
+    Assert.assertNull("Unexpected Cache Value", dirCacheStore.get(dirC));
+    Assert.assertEquals("Incorrect cache size", 1, dirCacheStore.size());
+
+    // Step7. Make it empty
+    dirCacheStore.remove(dirB);
+    Assert.assertEquals("Incorrect cache size", 0, dirCacheStore.size());
+  }
+
+  @Test
+  public void testNullKeysAndValuesToLRUCacheDirectoryPolicy() throws IOException {
```
linyiqun, line 170: Checkstyle issue of long line length:
>[ERROR] src/test/java/org/apache/hadoop/ozone/om/cache/TestOMMetadataCache.java:[170] (sizes) LineLength: Line is longer than 80 characters (found 83).

linyiqun (commented):
@rakeshadr , thanks for addressing the comments!
Minor comments below.

linyiqun:
> > Hi @rakeshadr , some initial review comments below.
> > In additional, one question from me: this is the first task of dir cache, there will be other further subtasks. But this part of work depends on [HDDS-2939](https://issues.apache.org/jira/browse/HDDS-2939) be completed. So that means [HDDS-4222](https://issues.apache.org/jira/browse/HDDS-4222) tasks cannot be merged immediately and be blocked for a long time? How do we plan to coordinate with [HDDS-2939](https://issues.apache.org/jira/browse/HDDS-2939) and [HDDS-4222](https://issues.apache.org/jira/browse/HDDS-4222) feature development works?
> 
> Good comment. yes, cache should be integrated eventually(case-by-case) to get full benefit. But [HDDS-4222](https://issues.apache.org/jira/browse/HDDS-4222) can be upstreamed separately and not blocked, I feel.
> 
> Cache can be integrated once [HDDS-2949](https://issues.apache.org/jira/browse/HDDS-2949) is finished. During dir creation, it checks the [DIR_EXISTS](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java#L185) and cache can be integrated into this call path.
> 
> Later, while implementing File, lookups tasks, delete, rename etc will integrate into that areas.

Get it, sounds good to me.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/cache/TestOMMetadataCache.java
rakeshadr: Done!

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/cache/TestOMMetadataCache.java
rakeshadr: Done!

hadoop-hdds/common/src/main/resources/ozone-default.xml
rakeshadr: Fixed test failure

rakeshadr:
Thanks a lot @linyiqun for the useful review comments!

linyiqun (approved):
+1.

mukul1987:
RocksDB cache will provide caching of intermediate prefixes

bharatviswa504:
But I think if we depend on rocksdb cache, every time we need bytes to proto conversion for all intermediate paths.
I believe this is one of the reason, we want to go with our caching layer right?

bharatviswa504:
But I think if we depend on rocksdb cache, every time we need bytes to proto conversion for all intermediate paths.
I believe this is one of the reason, we want to go with our caching layer right?

bharatviswa504:
But I think if we depend on rocksdb cache, every time we need bytes to proto conversion for all intermediate paths.
I believe this is one of the reason, we want to go with our caching layer right?

PR #1438, opened 2020-09-21
HDDS-4263. ReplicatiomManager shouldn't retain one healthy replica per origin node Id.

maobaolong:

ReplicatiomManager shouldn't retain one healthy replica per origin node Id.

HDDS-4263

## How was this patch tested?

Make the replicas over replicated, and wait the replicationManager reduce it to the replicationFactor.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -24,12 +24,10 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


maobaolong:
Hi @nandakumar131 , thanks for the first review and suggestion, I agree to you that we should not remove the OriginDatanodeId handling blindly, now I add a condition for the uniqueReplicas collection, we don't have to collect the uniqueReplicas when the state of the container is closed. 

As the purpose os this PR is to make RATIS-3 CLOSED-2 work. I also submit two related PR to the `storage-class` branch, if you are interested in this feature, your can take a look at those PR.

- https://github.com/elek/hadoop-ozone/pull/20
- https://github.com/elek/hadoop-ozone/pull/21

nandakumar131 (approved):

nandakumar131:
Thanks @maobaolong for the contribution!

PR #1444, opened 2020-09-23
HDDS-4242. Copy PrefixInfo proto to new project hadoop-ozone/interface-storage

amaliujia:

As a following effort, try to move  `PrefixInfo` and other necessary classes to `interface-stroage`

https://issues.apache.org/jira/browse/HDDS-4242

## How was this patch tested?

Unit Test


amaliujia:
R @elek

hadoop-ozone/interface-storage/src/main/java/org/apache/hadoop/ozone/om/helpers/OzoneAclStorage.java
```diff
@@ -0,0 +1,63 @@
...
+ * This class is used by storage proto only.
+ *
+ * ACLs in Ozone follow this pattern.
+ * <ul>
+ * <li>user:name:rw
+ * <li>group:name:rw
+ * <li>world::rw
+ * </ul>
+ */
+final class OzoneAclStorage {
```
amaliujia, line 40: Have to duplicate this class because the original one was used by different places. That's why I make this class `package-private` to make it only useable by codec classes.

elek (approved):
+1, thanks the patch @amaliujia 

One comment: I am not sure if we need a separated util classes where we have static `toProtobuf` and `fromProtobuf` methods. As we already have an interface which has the responsibility of encode/decode classes (`Codec`) I would consider moving all encoding/decoding logic to the codec and re-use from there (if necessary). 

But this is just my opinion, and I can be convinced.

(As we expect more patches I am fine to merge it as is, but we can continue the discussion...)

amaliujia:
@elek your suggestion make senses as the new util classes are dedicated to be used by `interface-storage`. If it is even dedicated to be used by a `codec` class, it can be moved into the `codec` class.

How about let me try to remove util classes in the next PR? 

This PR is to have a working migration for single proto migration to build consensus around naming convention, how to deal with helpers/util classes. As this PR looks good overall, I will move ~3 proto in each future PR to accelerate migrate process while keep each PR be easier to maintain. 

I can address the util class comment in the next PR with more data points there (e.g. 3 more proto migration)

amaliujia:
@elek  friendly ping~ WDYT on the idea in my previous reply?

elek:
Sorry, I  was busy at apachecon last week. Sure, fine to do it later. it's just an opinion and fine to follow it in next PRs.

Let me merge it now.

PR #1446, opened 2020-09-25
HDDS-4274. Change the log level of the SCM Delete block to improve performance.

captainzmc:

There are scenarios in which the DELETE operation is very frequent. Now that info level logs in SCM are affecting performance, we should change this to Debug

https://issues.apache.org/jira/browse/HDDS-4274

## How was this patch tested?

 no need add ut.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMBlockProtocolServer.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMBlockProtocolServer.java
@@ -218,7 +218,7 @@ public class SCMBlockProtocolServer implements
   @Override
   public List<DeleteBlockGroupResult> deleteKeyBlocks(
       List<BlockGroup> keyBlocksInfoList) throws IOException {
-    LOG.info("SCM is informed by OM to delete {} blocks", keyBlocksInfoList
+    LOG.debug("SCM is informed by OM to delete {} blocks", keyBlocksInfoList
         .size());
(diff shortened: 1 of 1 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMBlockProtocolServer.java
```diff
@@ -218,7 +218,7 @@ public void join() throws InterruptedException {
   @Override
   public List<DeleteBlockGroupResult> deleteKeyBlocks(
       List<BlockGroup> keyBlocksInfoList) throws IOException {
-    LOG.info("SCM is informed by OM to delete {} blocks", keyBlocksInfoList
+    LOG.debug("SCM is informed by OM to delete {} blocks", keyBlocksInfoList
```
amaliujia, line 221: If changing to debug level, then better to add `if (LOG.isDebugEnabled())`?

captainzmc: Thanks for @amaliujia's  prompt, judgment has been added.

adoroszlai (approved):
Thanks @captainzmc for the improvement.  Do you have any measurement of its effect?

I think it's safe to change the log level, since audit log will include deletions as well as failures.

xiaoyuyao (approved):
LGTM, +1.

xiaoyuyao:
Thanks @captainzmc for the contribution and all for the reviews. PR has been merged.

PR #1454, opened 2020-09-29
HDDS-4285. Read is slow due to frequent calls to UGI.getCurrentUser() and getTokens()

adoroszlai:

Reduce the number of `getCurrentUser()` and `getTokens()` calls performed during some `ContainerProtocolCalls` operations.  This is achieved by getting the tokens once in `BlockInputStream` and `BlockOutputStream` initialization, and passing them to `getBlock`, `putBlock`, `readChunk`, `writeChunk` calls.

https://issues.apache.org/jira/browse/HDDS-4285

## How was this patch tested?

Verified that time required for the repro unit test is improved.

Without the patch:

```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 112.829 s - in org.apache.hadoop.ozone.client.io.TestKeyOutputStreamUnit
```

With the patch:

```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 30.293 s - in org.apache.hadoop.ozone.client.io.TestKeyOutputStreamUnit
```

Regular CI:
https://github.com/adoroszlai/hadoop-ozone/runs/1181129921


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/storage/ContainerProtocolCalls.java
```diff
@@ -84,7 +86,8 @@ private ContainerProtocolCalls() {
    * @throws IOException if there is an I/O error while performing the call
    */
   public static GetBlockResponseProto getBlock(XceiverClientSpi xceiverClient,
-      DatanodeBlockID datanodeBlockID) throws IOException {
+      DatanodeBlockID datanodeBlockID,
+      Collection<Token<? extends TokenIdentifier>> tokens) throws IOException {
```
linyiqun, line 90: Can we update the javadoc comment of this method as we new added param tokens here?
The same suggestion for methods we changed below:

* putBlockAsync
* readChunk
* writeChunkAsync

linyiqun (commented):
The fix looks good to me, @adoroszlai .
One minor comment below, : ).

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/storage/ContainerProtocolCalls.java
adoroszlai: Thanks @linyiqun for the suggestion.  Would

```java
@param tokens list of tokens the current user has, possibly including a token for this block
```

be OK, or do you have a better description for this param?

linyiqun: @adoroszlai , above description looks good to me. Thanks.

ChenSammi:
@adoroszlai  thanks for identify this performance issue.  I'm not quite familar with tokens, since it's a user token, suppose it has a much longer liveness than input and output stream object instance,  right?

adoroszlai:
> thanks for identify this performance issue

Credit goes to @elek who reported the problem, see HDDS-4285.

> I'm not quite familar with tokens

Neither am I, but we have some [doc](https://hadoop.apache.org/ozone/docs/1.0.0/security/secureozone.html#tokens-).

> since it's a user token, suppose it has a much longer liveness than input and output stream object instance, right?

It is a block token, specific to the block being written/read.  I think it's short-lived.

https://github.com/apache/hadoop-ozone/blob/d1ac423b3fe35c5d363b348108512440f08ee5f1/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/security/token/OzoneBlockTokenIdentifier.java#L41-L45

linyiqun:
>since it's a user token, suppose it has a much longer liveness than input and output stream object instance, right?
A good point that @ChenSammi mentioned, there is a setting to decide the expired time of block token and it's a long time as 1 day. This should be an enough time for client to read/write block data. Suppose there also should be some service to renew expired token.

```xml
  <property>
    <name>hdds.block.token.expiry.time</name>
    <value>1d</value>
    <tag>OZONE, HDDS, SECURITY, TOKEN</tag>
    <description>
      Default value for expiry time of block token. This
      setting supports multiple time unit suffixes as described in
      dfs.heartbeat.interval. If no suffix is specified, then milliseconds is
      assumed.
    </description>
  </property>
```

In additional, I'm +1 for current change. Thanks @adoroszlai  for addressing the comment.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockInputStream.java
```diff
@@ -193,10 +196,11 @@ public synchronized void initialize() throws IOException {
       if (token != null) {
         UserGroupInformation.getCurrentUser().addToken(token);
```
xiaoyuyao, line 197: Should we just pass down the token from the caller directly and bypass the UGI addToken/getTokens completely?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/storage/ContainerProtocolCalls.java
xiaoyuyao: We can directly encode the pass in token for the specific block id without additional token selection.

xiaoyuyao (commented):
Thanks @adoroszlai  for working on this. The change LGTM overall. Just have few comments added inline.

adoroszlai:
> Should we just pass down the token from the caller directly and bypass the UGI addToken/getTokens completely?
> We can directly encode the pass in token for the specific block id without additional token selection.

Thanks @xiaoyuyao for the comments.  I was also wondering about this, but I didn't have enough background info to decide.  If you feel comfortable with bypassing the UGI for the block token, I'm fine with making this simplification.

adoroszlai:
Thanks @xiaoyuyao for the suggestion to directly pass down the block token.  I have updated the patch accordingly.

ChenSammi:
@adoroszlai ， thanks for the explanation.  +1.

xiaoyuyao:
Thanks @adoroszlai for the update. LGTM, +1.

adoroszlai:
Thanks @xiaoyuyao for reviewing and committing it, @ChenSammi and @linyiqun for the review, and @elek for finding the issue.

PR #1456, opened 2020-09-30
HDDS-4172. Implement Finalize command in Ozone Manager server.

fapifta:

Server side implementation of the finalization logic is implemented in this pull request.
The initial idea in the client side implementation is to issue an initiating request, then monitor the process with client side polling.
This is reflected in the server side code, but due to complications, in the server side, the background finalization of layoutfeatures one by one, is postponed and tracked in HDDS-4286.
The problem with the backround finalization is that it has to be synced between the OMs, and we need to ensure that once requests are in the state machine for finalizing a layout feature, the requests processed by the master before the finalization requests, and after the finalization requests are in sync and in order.
This means that we need to post separate requests into the state machine internally on the leader OM, to have a flow, where the a specific request type or change to a request handling, once its activated, it is on all OMs activated at the same transaction. We do not really have such a logic for now, and it requires some further review. I will post a design doc on this one into HDDS-4286 in the near future about possible solutions.

With that, in this patch, the finalization of the features is happening inside the initiating UpgradeFinalizeRequest RPC call handling, and it has to finish on at least two OMs. So we process the finalization inside the statemachine in one batch.
After this, the client gets a STARTING_FINALIZATION status back, and after the 500ms delay on the client side, the client will grab the results.

https://issues.apache.org/jira/browse/HDDS-4172

## How was this patch tested?

Quick manual test for the very basic workflow so far.
JUnit tests to be added to the PR, but wanted to share to have reviews on the approach while I am working on the tests.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
```diff
@@ -0,0 +1,303 @@
...
+public class OMUpgradeFinalizer implements UpgradeFinalizer<OzoneManager> {
+
+  private Status status = ALREADY_FINALIZED;
+  private OMLayoutVersionManagerImpl versionManager;
+  private String clientID;
+
+  private Queue<String> msgs = new ConcurrentLinkedQueue<>();
+  private boolean isDone = false;
+
+  private static final OmUpgradeAction NOOP = a -> {};
```
avijayanhwx, line 54: Nit. Can this be moved to the interface for reuse in other implementations?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
```diff
@@ -0,0 +1,303 @@
...
+          + "The requestor is not the initiating client of the finalization,"
+          + " if you want to take over, and get unsent status messages, check"
+          + " -takeover option.", INVALID_REQUEST);
+    }
+  }
+
+
+
+
+  private class Worker implements Callable<Void> {
```
avijayanhwx, line 113: Nit. Class needs a Javadoc.

avijayanhwx (commented):
Code flow looks good to me. For my peace of mind, I will go through it one more time when tests are added. A couple of nits.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
```diff
@@ -0,0 +1,303 @@
...
+
+    @Override
+    public Void call() throws OMException {
+      try {
+        emitStartingMsg();
+
+        for (OMLayoutFeature f : versionManager.unfinalizedFeatures()) {
+          finalizeFeature(f);
+          updateLayoutVersionInVersionFile(f);
+          versionManager.finalized(f);
```
linyiqun, line 128: Should we do the safe layout check here? We should ensure the feature layout is increasing between OMLayoutFeature, otherwise versionManager#finalized will throw IllegalArgumentException error.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
```diff
@@ -65,53 +64,66 @@ protected void initializeFeatures(LayoutFeature[] lfs) {
...
+    isInitialized = false;
+  }
+
+  public void finalized(T layoutFeature) {
+    if (layoutFeature.layoutVersion() > metadataLayoutVersion) {
+      metadataLayoutVersion = layoutFeature.layoutVersion();
+    } else {
+      throw new IllegalArgumentException(
+          "Finalize attempt on a layoutFeature which has already been"
+              + "finalized. Software Layout version: " + softwareLayoutVersion
```
linyiqun, line 81: Nit: Missing a whitespace before 'finalized'.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/StorageInfo.java
```diff
@@ -150,6 +179,16 @@ private void verifyClusterId()
     }
   }
 
+  private void verifyUpgradingToLayoutVersion()
+      throws InconsistentStorageStateException {
+    int upgradeMark = getUpgradingToLayoutVersion();
+    if (upgradeMark != INVALID_LAYOUT_VERSION) {
+      throw new InconsistentStorageStateException("Ozone Manager died during"
+          + "a LayoutFeature upgrade.");
```
linyiqun, line 187: Nit: Missing a whitespace before 'a LayoutFeature upgrade'.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
```diff
@@ -0,0 +1,303 @@
...
+      }
+    }
+
+    private void putFinalizationMarkIntoVersionFile(OMLayoutFeature feature)
+        throws OMException {
+      try {
+        emitUpgradeToLayoutVersionPersistingMsg(feature.name());
+
+        setUpgradeToLayoutVersionInStorage(feature.layoutVersion());
+        persistStorage();
```
linyiqun, line 175: Same comment for persistStorage like above.

linyiqun (commented):
Hi @fapifta , few review comments from me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
```diff
@@ -0,0 +1,303 @@
...
+    new Worker(om).call();
+    return STARTING_MSG;
+  }
+
+  @Override
+  public StatusAndMessages reportStatus(
+      String upgradeClientID, boolean takeover
+  ) throws IOException {
+    if (takeover) {
+      clientID = upgradeClientID;
```
linyiqun, line 90: Here clientID should be thread-safe to update, can we make clientID as a volatile variable?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
fapifta: I moved it to OmUpgradeAction, I believe that is the best place. Thank you for catching this one.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
fapifta: Added doc to the Worker class as well.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
fapifta: fixed

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/StorageInfo.java
fapifta: fixed

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    private void updateLayoutVersionInVersionFile(OMLayoutFeature feature)
+        throws OMException {
+      int prevLayoutVersion = currentStoredLayoutVersion();
+
+      updateStorageLayoutVersion(feature.layoutVersion());
+      try {
+        persistStorage();
+      } catch (IOException e) {
```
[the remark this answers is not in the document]
fapifta: The main things we want to ensure here:
1. at any point in time there can be a failure point, so we would better persist where we are at before and after every major step, that is what we can not loose, as all the in-memory states can be lost at any point in time.
2. The finalization action that we do for LayoutFeatures does the modifications required by the LayoutFeature inside our internal structures, so we need to know when we have started one such a step but we have not finished it due to a catastrophic failure, as in that case we are potentially in an inconsistent state, and OM should not proceed with anything until the state is examined and we are free to move further.
3. If the finalization action is a NOOP, we skip this part, and just update the LayoutVersion and persist the update
4. After a LayoutFeature is finalized we need to update and persist the LayoutVersion, as if the next finalize action ends in a catastrophic failure, we need to know where to catch up from.

So In this particular case, we persisting the LayoutFeature's LayoutVersion to Storage to ensure point 4, so we never redo the finalization of a LayoutFeature, as redoing it might lead to an issue as the start state of a second finalization would be unexpected after it finished once.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
fapifta: From the comment for the previous persisting related comment point 2 is relevant here, we need to know if we have started the finalization of a LayoutFeature but never finished it, and we need to know it even if there is a catastrophic failure, and OM has been restarted due to that failure, and persisting this information to disk seems to be the way to fullfil this requirement.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
fapifta: AbstractLayoutVersionManager gives the insurance for us here, as that has a map for the features with the feature's LayoutVersion as a key, and unfinalizedFeatures() works based off of that map.
So the internal structure, and the overall design should ensure that there can be only one LayoutFeature associated with one LayoutVersion.
It would be beneficial to work based on the ordinal of the OMLayoutFeature enum, but this is not the case for now, I think that is an improvement we can add there, so that a wrongly defined enum member can not hide an other LayoutFeature, or we can ensure that with a test. I filed HDDS-4303 to track this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
fapifta: Good point, though making clientID volatile, does not really help to solve a race condition here I believe, so I propose to synchronize the whole reportStatus method.
The reasons is that, if there is the original client that polls together with a client that does a takeover, then both can step on each others toes, as if for the original client we are already processing the message queue when an other client takes over, then returning status might change as well, and also there might be missing messages in the responses sent to the clients.

Fortunately we don't have to worry about isDone and messages at least, as even if isDone is being updated late, and the final message is polled from the msgs queue with an IN_PROGRES state, the client in the next cycle gets a DONE state without messages, and that is acceptable.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
fapifta: Ok, it turned out that if OMUpgradeAction has just one constants, that qualifies as a bad practice, as if an interface does contain just constants, that can have some disadvantages according to seemingly more clever ppl than me. :)

I moved it back to OMUpgradeFinalizer, as due to the logic we need to preserve the OMUpgradeAction-ness of the NOOP to work well there, as we want to push OzoneManager instances to the executeAction, and the type system is against having a simple UpgradeAction as a NOOP. We can thin about adding this to OMLayoutFeature instead, but as we do not use it elsewhere at the moment, and because we use Optional and we provide just a null in case the finalize action is a noop, we do not need it elsewhere as it seems...

An other approach can be an orElse(null), but with that we can even leave out the Optional in the OMLayoutFeature and LayoutFeature implementations/definitions, and instead checking if it is a NOOP, we can check for null.

What do you think?

fapifta:
Guys thank you for the reviews, hopefully I was able to address all the concerns mentioned so far, let me know if you see anything more, and please share your opinion about the syncronization, and about the NOOP, as that seems to be remaining open questions.

I will add unit level tests hopefully by Monday and then will move out the PR from draft.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
linyiqun: To synchronize the whole reportStatus method looks good to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
linyiqun: Okay, let's do the future track on HDDS-4303

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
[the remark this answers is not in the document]
linyiqun: Comment makes sense to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
linyiqun: Comment makes sense to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
avijayanhwx: There is a test that covers this check (in a way) --> TestOMVersionManager#testOMLayoutFeatureCatalog. Maybe, we can build on that.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
```diff
@@ -0,0 +1,328 @@
...
+
+    private int currentStoredLayoutVersion() {
+      return ozoneManager.getOmStorage().getLayoutVersion();
+    }
+
+    private void updateStorageLayoutVersion(int version) {
+      ozoneManager.getOmStorage().setLayoutVersion(version);
+    }
+
+    private void persistStorage() throws IOException {
```
avijayanhwx, line 237: Can this be moved to the individual layout version update methods above, so that the caller does not need to invoke the persist explicitly every time? As a caller, if I do **storage**.updateStorageLayoutVersion(), I can expect it to be persisted, right? I don't have a strong opinion about this, just a suggestion.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMFinalizeUpgradeRequest.java
```diff
@@ -63,11 +64,20 @@ public OMClientResponse validateAndUpdateCache(
 
       String upgradeClientID = request.getUpgradeClientId();
 
-      UpgradeFinalizationStatus status =
+      StatusAndMessages omStatus =
```
avijayanhwx, line 67: Line **54** has a debug log line left.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
```diff
@@ -0,0 +1,328 @@
...
+// Until that is done, monitoring is not really doing anything meaningful
+// but this is a tradoff we can take for the first iteration either if needed,
+// as the finalization of the first few features should not take that long.
+// Follow up JIRA is in HDDS-4286
+//    String threadName = "OzoneManager-Upgrade-Finalizer";
+//    ExecutorService executor =
+//        Executors.newSingleThreadExecutor(r -> new Thread(threadName));
+//    executor.submit(new Worker(om));
+    new Worker(om).call();
+    return STARTING_MSG;
```
avijayanhwx, line 77: Just an observation. By the time we return 'STARTING_MSG' at 77, we are already done with Finalization right? Is this return message left as it is for a future refactoring through HDDS-3286 where async finalization will be done?

avijayanhwx (approved):
Thanks for working on this @fapifta. 

LGTM +1.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
fapifta: yes, I did not wanted to change the overall flow of messages and statusas that I came up with, but technically at this point we already done in the current implementation.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/upgrade/OMUpgradeFinalizer.java
fapifta: Storage and OMStorage chose to have an implementation for setters where the setter does not persist the change to the disk, these are the two that I have checked, and decided to remain consistent with this behaviour. This is the sole reason for controlling this from outside the Storage class implementation. Here it is extracted into a method to make it easier to follow the flow, and to hide the implementation detail of how to persist the storage from the reader when he reads the backbone of the internal logic of the methods with a higher layer in the abstractions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMFinalizeUpgradeRequest.java
fapifta: Nice catch, it seems I forgot to remove this one. I am pushing the deletion of this line.

PR #1458, opened 2020-09-30
HDDS-3728. Bucket space: check quotaUsageInBytes when write key and allocate block.

captainzmc:

Bucket has implemented increase usedBytes when write, and this PR is based on #1431.
In this PR we judge whether the Bucket can be written when we write the key if the Bucket space quota enable.

https://issues.apache.org/jira/browse/HDDS-3728

## How was this patch tested?

ut added

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
@@ -279,11 +279,12 @@ public class OMFileCreateRequest extends OMKeyRequest {
 
       omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
       omBucketInfo = getBucketInfo(omMetadataManager, volumeName, bucketName);
-      // check volume quota
+      // check volume and bucket quota
       long preAllocatedSpace = newLocationList.size()
           * ozoneManager.getScmBlockSize()
           * omKeyInfo.getFactor().getNumber();
       checkVolumeQuotaInBytes(omVolumeArgs, preAllocatedSpace);
+      checkBucketQuotaInBytes(omBucketInfo, preAllocatedSpace);
 
       // Add to cache entry can be done outside of lock for this openKey.
       // Even if bucket gets deleted, when commitKey we shall identify if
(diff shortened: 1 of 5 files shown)


captainzmc:
Hi @ChenSammi , this PR add judge whether we can be write if bucket space quota enable. Can you help to review this?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -218,6 +219,8 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
       // update usedBytes atomically.
       omVolumeArgs.getUsedBytes().add(preAllocatedSpace);
       omBucketInfo.getUsedBytes().add(preAllocatedSpace);
+      long vol = omVolumeArgs.getUsedBytes().sum();
+      long buk = omBucketInfo.getUsedBytes().sum();
```
ChenSammi, line 223: Are these two used somewhere?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -279,11 +279,12 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
       omBucketInfo = getBucketInfo(omMetadataManager, volumeName, bucketName);
-      // check volume quota
+      // check volume and bucket quota
       long preAllocatedSpace = newLocationList.size()
           * ozoneManager.getScmBlockSize()
           * omKeyInfo.getFactor().getNumber();
       checkVolumeQuotaInBytes(omVolumeArgs, preAllocatedSpace);
+      checkBucketQuotaInBytes(omBucketInfo, preAllocatedSpace);
```
ChenSammi, line 287: Better check bucket quota before volume quota check.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
captainzmc: Redundancy code,  will delete this.

ChenSammi:
LGTM +1.

PR #1459, opened 2020-09-30
HDDS-4297. Allow multiple transactions per container to be sent for deletion by SCM.

lokeshj1703:

Currently SCM Block Deleting Service allows only one transaction per container to be sent for deletion to the datanode. This can slow down deletion if there are multiple delete transactions for a container.

https://issues.apache.org/jira/browse/HDDS-4297

## How was this patch tested?

UT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DatanodeDeletedBlockTransactions.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DatanodeDeletedBlockTransactions.java
@@ -82,30 +82,20 @@ public class DatanodeDeletedBlockTransactions {
   }
 
   private boolean addTransactionToDN(UUID dnID, DeletedBlocksTransaction tx) {
-    if (transactions.containsKey(dnID)) {
-      List<DeletedBlocksTransaction> txs = transactions.get(dnID);
-      if (txs != null && txs.size() < maximumAllowedTXNum) {
-        boolean hasContained = false;
-        for (DeletedBlocksTransaction t : txs) {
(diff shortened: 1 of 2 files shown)


linyiqun:
@lokeshj1703 , seems the original purpose of letting only one transaction per container to delete on datanode is to make container block deletion more balanced. So the containers will have the similar available space for SCM to choose to do block allocation.

But current existed implement is not friendly for the scenario that our deleted blocks focus on the same (or very few)containers in the datanode. And current PR change is helpful on this scenario. I think container block allocation should be random, not sure if this corner case will happened.

lokeshj1703:
@linyiqun currently we allow block deletion for only closed containers. For open containers deletion needs to be done via ratis pipeline which is not currently implemented. So there is no block allocation advantage right now. I do not think there is a huge window where containers are open and are simultaneously receiving block deletion requests in bulk.
Further it is better to accumulate transactions of a container if possible. There is an overhead to handling a block deletion request for datanode. It needs to load rockdb instance and then add those transactions to the db. It is better for datanode to process multiple such transactions at once.
I tested the PR and found that it gives better deletion speed for deleting a milllion keys. 
Without PR - SCM sends total 180 blocks for deletion in a minute. This number does not increase even after increasing configs. Datanode deletion speed was very slow. I monitored a datanode disk space and it deleted less than a GB even after 5 minutes.
With PR - SCM sends 6000 blocks for deletion. Datanode was deleting 1 GB of data every minute.

bshashikant (approved):

linyiqun:
Good to know the testing result. +1 for this change. Thanks @lokeshj1703 .

linyiqun (approved):

lokeshj1703:
@bshashikant @linyiqun Thanks for reviewing the PR! I have committed it to master branch.

PR #1460, opened 2020-09-30
HDDS-4298. Use an interface in Ozone client instead of XceiverClientManager

elek:

`XceiverClientManager` is used everywhere in the ozone client (Key/Block Input/OutputStream) to get a client when required.

To make it easier to create genesis/real unit tests, it would be better to use a generic interface instead of `XceiverClientManager` which can be replaced easily with a mock implementation.

https://issues.apache.org/jira/browse/HDDS-4298

## How was this patch tested?

Usability: Created a real unit test (mocked-read branch on my fork) for HDDS-4285. 

Correctness: CI 

(this is a very simple change, I just introduced a new interface and Input/Ouptut streams use the interface instead of the implementation)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestReadRetries.java
```diff
@@ -147,7 +145,8 @@ public void testPutKeyAndGetKeyThreeNodes()
             ReplicationFactor.THREE, new HashMap<>());
     KeyOutputStream groupOutputStream =
         (KeyOutputStream) out.getOutputStream();
-    XceiverClientManager manager = groupOutputStream.getXceiverClientManager();
+    XceiverClientManager manager =
+        (XceiverClientManager) groupOutputStream.getXceiverClientFactory();
```
adoroszlai, line 149: Why cast instead of changing the variable to `XceiverClientFactory`?

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.hadoop.hdds.scm.pipeline.Pipeline;
+
+import org.apache.ratis.thirdparty.com.google.protobuf.ByteString;
+
+/**
+ * Interface to provide XceiverClient when needed.
+ */
+public interface XceiverClientFactory {
+
+  Function<ByteBuffer, ByteString> byteBufferToByteStringConversion();
```
adoroszlai, line 33: I don't think this method should be part of the interface.  The implementation is a convenience method to avoid passing configuration down the `RpcClient -> KeyOutputStream.Builder -> KeyOutputStream -> BlockOutputStreamEntryPool` chain.

https://github.com/apache/hadoop-ozone/blob/7216e3cb651dffc8c7a86d18729fd5f5cffa24b4/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientManager.java#L281-L283

I'd rather get rid of this smell as part of this refactoring.

adoroszlai (commented):
Thanks @elek for this refactoring.  Mostly LGTM.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestReadRetries.java
elek: Both can work, but I feel it more safe as it doesn't require changing method signatures in `KeyOutputStream`. Even if `getXceiverClientFactory` is marked as `@VisibleForTesting` I feel it safer to return only the interface. Unit test can have cast, it's expected to have access to the internal representations.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientFactory.java
elek: Totally agree, I also felt like this, but I was lazy to hunt it down. Let me try to remove it.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestReadRetries.java
adoroszlai: I meant: is there anything against changing to

```java
    XceiverClientFactory manager =
        groupOutputStream.getXceiverClientFactory();
```

Sorry if it was unclear.

elek: Aghh, got it. Obviously: you are right, I just didn't notice the usage of the `manager`. Pushed the improved version.

adoroszlai (approved):

bshashikant (approved):

PR #1463, opened 2020-09-30
HDDS-4292. Ozone Client not working with Hadoop Version < 3.2

bharatviswa504:

The ozone client is not working with Hadoop Version < 3.2.
The issue is due to not properly initialization of the proxy object.

https://issues.apache.org/jira/browse/HDDS-4292
## How was this patch tested?

Tested the fix on a cluster with Hadoop version 3.0

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/ha/OMFailoverProxyProvider.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/ha/OMFailoverProxyProvider.java
@@ -141,15 +141,14 @@ public class OMFailoverProxyProvider implements
 
         if (omProxyInfo.getAddress() != null) {
 
-          ProxyInfo<OzoneManagerProtocolPB> proxyInfo =
-              new ProxyInfo(null, omProxyInfo.toString());
 
           // For a non-HA OM setup, nodeId might be null. If so, we assign it
           // a dummy value
           if (nodeId == null) {
             nodeId = OzoneConsts.OM_NODE_ID_DUMMY;
           }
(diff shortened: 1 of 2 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/ha/OMFailoverProxyProvider.java
```diff
@@ -141,15 +141,14 @@ protected void loadOMClientConfigs(ConfigurationSource config, String omSvcId)
...
-          ProxyInfo<OzoneManagerProtocolPB> proxyInfo =
-              new ProxyInfo(null, omProxyInfo.toString());
 
           // For a non-HA OM setup, nodeId might be null. If so, we assign it
           // a dummy value
           if (nodeId == null) {
             nodeId = OzoneConsts.OM_NODE_ID_DUMMY;
           }
-          omProxies.put(nodeId, proxyInfo);
+          // ProxyInfo will be set during first time call to server.
```
vivekratnavel, line 150: This should be 
`ProxyInfo will not be set`?

bharatviswa504: It will be set when we make a first-time call to the server.
getProxy() is called when first-time submitRequest is called.

vivekratnavel: Got it, thatnks!

vivekratnavel:
+1 LGTM

bharatviswa504:
Thank You @vivekratnavel for the review.

PR #1465, opened 2020-10-01
HDDS-4300. Removed unneeded class DatanodeAdminNodeDetails

sodonnel:

DatanodeAdminNodeDetails was added earlier in the decommission branch, to track metrics and, the decommission state and maintenance end time.

After enhancing NodeStatus to old the Maintenance Expiry time, this class is no longer needed and it also duplicates information which is stored in other existing places.

This change removes it and then metrics etc can be added later in a different way.

https://issues.apache.org/jira/browse/HDDS-4300

## How was this patch tested?

Existing tests


elek (approved):
+1 LGTM

Thanks to move this branch forward.

PR #1467, opened 2020-10-01
HDDS-4268. Prepare for Upgrade step should purge the log after waiting for the last txn to be applied.

avijayanhwx:
This is a follow up task from HDDS-4227 in which the prepare upgrade/downgrade task should purge the Raft log immediately after waiting for the last txn to be applied. This is to make sure that we dont "apply" transactions in different versions of the code across the quorum. A lagging follower will use a Ratis snapshot to bootstrap itself on restart.

https://issues.apache.org/jira/browse/HDDS-4268

## How was this patch tested?
Manually tested.
Added unit/integration test.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -432,7 +443,14 @@ private RaftProperties newRaftProperties(ConfigurationSource conf) {
     final int logPurgeGap = conf.getInt(
         OMConfigKeys.OZONE_OM_RATIS_LOG_PURGE_GAP,
         OMConfigKeys.OZONE_OM_RATIS_LOG_PURGE_GAP_DEFAULT);
-    RaftServerConfigKeys.Log.setPurgeGap(properties, logPurgeGap);
+
+    if (forUpgrade) {
+      LOG.info("Setting purge gap to 1.");
+      RaftServerConfigKeys.Log.setPurgeGap(properties, 1);
+    } else {
+      LOG.info("Setting purge gap to 1000.");
```
linyiqun, line 451: Nit: purge gap should not hard-coded as 1000 in this log, can you use variable logPurgeGap instead of?

linyiqun (commented):
PR almost looks great. Will take a look again after unit test added.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
avijayanhwx: Thanks, used it while testing it manually :) Forgot to remove it.

avijayanhwx:
cc @fapifta / @prashantpogde

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/ratis/RatisUpgradeUtils.java
```diff
@@ -78,14 +82,26 @@ public static void waitForAllTxnsApplied(
       throw new IOException("Index from Snapshot does not match last applied " +
           "Index");
     }
+
+    if (purgeLogsAfter) {
+      CompletableFuture<Long> purgeFuture =
+          impl.getState().getLog().purge(snapshotIndex);
+      try {
+        Long purgeIndex = purgeFuture.get();
```
linyiqun, line 90: I prefer to have a timeout limitation here, it will wait forever if it hasn't finished the purge behavior.
Can we reuse maxTimeToWaitSeconds as the purge timeout and catch timeout exception?
And then add a corresponding test case for this corner case.

linyiqun (commented):
@avijayanhwx , only one minor comment below.

avijayanhwx:
/pending Looking into an edge case where Ratis log index can go over applied index, thereby causing purge not to succeed.

PR #1468, opened 2020-10-01
HDDS-4301. SCM CA certificate does not encode KeyUsage extension properly

xiaoyuyao:

Fix the encoding of KeyUsage in the CSR to ensure it can be used in FIPS environment. 

https://issues.apache.org/jira/browse/HDDS-4301

## How was this patch tested?

FIPS enabled secure ozone cluster and verify that the CA certificate can be properly generated and used.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificates/utils/CertificateSignRequest.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificates/utils/CertificateSignRequest.java
@@ -265,7 +265,7 @@ public final class CertificateSignRequest {
       }
       KeyUsage keyUsage = new KeyUsage(keyUsageFlag);
       return new Extension(Extension.keyUsage, true,
-          new DEROctetString(keyUsage));
+          keyUsage.getEncoded());
     }
 
     private Optional<Extension> getSubjectAltNameExtension() throws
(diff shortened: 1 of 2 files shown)


elek (commented):
Thanks to fix this @xiaoyuyao

I don't fully understand it (yet), and I have no __FIPS certificate suite__ off-the-shelf, so I am asking to learn and to make it possible to merge it ;-)

 1. Is it a backward compatible change? If I understood well both the old method and new method is good enough for validation, but the new version is more standard.

 2. There is a `new DEROctetString(keyUsage)` call in  `SelfSignedCertificate` which is very similar to the one which is replaced in `CertificateSigningRequest`.  

Do we need to replace it? Is it always better to use `keyUsage.getEncoded()` instead of `DEROctetString`?

xiaoyuyao:
bq. Is it a backward compatible change? If I understood well both the old method and new method is good enough for validation, but the new version is more standard.

Yes. The new version is the standard way to encode the keyUsage. 

bq. There is a new DEROctetString(keyUsage) call in SelfSignedCertificate which is very similar to the one which is replaced in CertificateSigningRequest.
 There are 4 "new DEROctetString" but only two of them are for keyUsage. This PR is to fix the encoding of the keyUsage. I don't see find issue of using DEROctetString for for constraints and subject alternate names extension.

arp7 (approved):
LGTM

xiaoyuyao:
Thanks @elek and @arp7 for the review.

PR #1472, opened 2020-10-04
HDDS-4306. Ozone checkstyle rule can't be imported to IntelliJ.

xiaoyuyao:

CheckStyle: Move LineLength Check parent from TreeWalker to Checker, otherwise fail to import to latest IntelliJ

https://issues.apache.org/jira/browse/HDDS-4306

## How was this patch tested?

Import to IntelliJ checkstyle plugin and verify there is no syntax error.


xiaoyuyao:
Thanks @adoroszlai for the review. The suggested change LGTM, I will make change in the next commit. One thing I notice that you suggest use LineLengthCheck instead of LineLength module. I think it is typo based on the document [here|https://checkstyle.sourceforge.io/config_sizes.html#LineLength]

<module name="LineLength">
  <property name="fileExtensions" value="java"/>
</module>

adoroszlai:
> One thing I notice that you suggest use LineLengthCheck instead of LineLength module. I think it is typo

Today I learned: both forms work fine.

However, you are correct, I didn't intend to propose changing the module name.  Even if both are OK, it's unnecessary for the fix.

adoroszlai (approved):
Thanks @xiaoyuyao for updating the patch.

PR #1475, opened 2020-10-06
HDDS-4310: Ozone getconf broke the compatibility

umamaheswararao:

https://issues.apache.org/jira/browse/HDDS-4310

## How was this patch tested?

Added test cases.


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/conf/PrintConfKeyCommandHandler.java
```diff
@@ -42,8 +43,9 @@ public Void call() throws Exception {
     String value = tool.getConf().getTrimmed(confKey);
     if (value != null) {
       tool.printOut(value);
+    } else {
+      tool.printError("Configuration " + confKey + " is missing.");
```
adoroszlai, line 47: Nice find.

adoroszlai (approved):
Thanks @umamaheswararao for fixing this.

PR #1478, opened 2020-10-06
Fix inconsistency recon config keys starting with recon and not ozone

frischHWC:
Fix recon configs inconsistent

https://issues.apache.org/jira/browse/HDDS-4309

## How was this patch tested?

Build up project, launch dockers, and new default parameters were present but not old ones anymore.

On recon container:
```
> ozone getconf confKey recon.om.socket.timeout
Configuration recon.om.socket.timeout is missing.
> ozone getconf confKey ozone.recon.om.socket.timeout
5s
```

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServerConfigKeys.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServerConfigKeys.java
@@ -63,28 +63,46 @@ public final class  ReconServerConfigKeys {
   public static final String OZONE_RECON_OM_SOCKET_TIMEOUT =
       "ozone.recon.om.socket.timeout";
   public static final String OZONE_RECON_OM_SOCKET_TIMEOUT_DEFAULT = "5s";
+  @Deprecated
+  public static final String RECON_OM_SOCKET_TIMEOUT =
+      "recon.om.socket.timeout";
 
   public static final String OZONE_RECON_OM_CONNECTION_TIMEOUT =
       "ozone.recon.om.connection.timeout";
   public static final String OZONE_RECON_OM_CONNECTION_TIMEOUT_DEFAULT = "5s";
+  @Deprecated
+  public static final String RECON_OM_CONNECTION_TIMEOUT =
+      "recon.om.connection.timeout";
 
   public static final String OZONE_RECON_OM_CONNECTION_REQUEST_TIMEOUT =
       "ozone.recon.om.connection.request.timeout";
-
-  public static final String OZONE_RECON_OM_CONNECTION_REQUEST_TIMEOUT_DEFAULT = "5s";
+  public static final String OZONE_RECON_OM_CONNECTION_REQUEST_TIMEOUT_DEFAULT =
+      "5s";
+  @Deprecated
+  public static final String RECON_OM_CONNECTION_REQUEST_TIMEOUT =
+      "recon.om.connection.request.timeout";
 
   public static final String OZONE_RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY =
       "ozone.recon.om.snapshot.task.initial.delay";
   public static final String
       OZONE_RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY_DEFAULT = "1m";
+  @Deprecated
+  public static final String RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY =
+      "recon.om.snapshot.task.initial.delay";
 
   public static final String OZONE_RECON_OM_SNAPSHOT_TASK_INTERVAL_DELAY =
       "ozone.recon.om.snapshot.task.interval.delay";
   public static final String OZONE_RECON_OM_SNAPSHOT_TASK_INTERVAL_DEFAULT
       = "10m";
+  @Deprecated
+  public static final String RECON_OM_SNAPSHOT_TASK_INTERVAL_DELAY =
+      "recon.om.snapshot.task.interval.delay";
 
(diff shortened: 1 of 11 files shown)


vivekratnavel (approved):
+1 LGTM

vivekratnavel:
@frischHWC Thanks for working on this! The patch looks good to me except the failing checkstyles. Please fix those longer lines and I can merge it.

```
hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServerConfigKeys.java
 74: Line is longer than 80 characters (found 86).
hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/OzoneManagerServiceProviderImpl.java
 128: Line is longer than 80 characters (found 82).
hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/PrometheusServiceProviderImpl.java
 68: Line is longer than 80 characters (found 84).
hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/recon/TestReconWithOzoneManager.java
 99: Line is longer than 80 characters (found 82).
hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestOzoneConfigurationFields.java
 72: Line is longer than 80 characters (found 82).
```

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServerConfigKeys.java
```diff
@@ -60,31 +60,31 @@
...
+  public static final String OZONE_RECON_OM_CONNECTION_REQUEST_TIMEOUT =
+      "ozone.recon.om.connection.request.timeout";
 
-  public static final String RECON_OM_CONNECTION_REQUEST_TIMEOUT_DEFAULT = "5s";
+  public static final String OZONE_RECON_OM_CONNECTION_REQUEST_TIMEOUT_DEFAULT = "5s";
 
-  public static final String RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY =
-      "recon.om.snapshot.task.initial.delay";
+  public static final String OZONE_RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY =
+      "ozone.recon.om.snapshot.task.initial.delay";
```
adoroszlai, line 77:
```suggestion
  public static final String OZONE_RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY =
      "ozone.recon.om.snapshot.task.initial.delay";
  @Deprecated
  public static final String RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY = 
      "recon.om.snapshot.task.initial.delay";
```

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/OzoneManagerServiceProviderImpl.java
```diff
@@ -205,12 +205,12 @@ public void start() {
     }
     reconTaskController.start();
     long initialDelay = configuration.getTimeDuration(
-        RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY,
-        RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY_DEFAULT,
+        OZONE_RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY,
+        OZONE_RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY_DEFAULT,
```
adoroszlai, line 209:
```suggestion
        OZONE_RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY,
        configuration.get(
            ReconServerConfigKeys.RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY,
            OZONE_RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY_DEFAULT),
```

adoroszlai (changes requested):
Thanks @frischHWC for working on this.

The new key names look good, but if we simply change them, existing clusters would not respect after upgrade any custom value previously configured for these properties.  Administrators would have to manually update their configs.

So I think we should keep the constants for the old names, and use them as fallback when looking up the config.  For one of them (`RECON_OM_SNAPSHOT_TASK_INITIAL_DELAY`) I have suggested two changes, which can be applied to all other old config keys.

Note: commit message should include Jira ID (and match Jira title), eg. `HDDS-4309. Fix inconsistent Recon config keys that start with "recon.om."`.  Whoever merges this, please pay attention to prepend Jira ID in commit message.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/ReconServerConfigKeys.java
frischHWC: You are right, I just added these deprecated configurations in a new commit!

adoroszlai:
Thanks @frischHWC for updating the patch.  Sorry, I forgot `TestOzoneConfigurationFields`, a unit test that checks whether properties in `ozone-default.xml` and the `*ConfigKeys` classes are in sync.  We need to enumerate the deprecated constants in `addPropertiesNotInXml`, since these are no longer present in the XML.

https://github.com/apache/hadoop-ozone/blob/f9b1ca45fdeaaa63b9355c5c54fafc96ddf4597e/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestOzoneConfigurationFields.java#L60-L76

adoroszlai (approved):

vivekratnavel:
@adoroszlai Thanks for the review and @frischHWC thanks for the contribution.

PR #1480, opened 2020-10-06
HDDS-4315. Use Epoch to generate unique ObjectIDs

hanishakoneru:

In a non-Ratis OM, the transaction index used to generate ObjectID is reset on OM restart. This can lead to duplicate ObjectIDs when the OM is restarted. ObjectIDs should be unique. 
For HDDS-2939 and NFS are some of the features which depend on ObjectIds being unique.

~~This Jira aims to introduce an epoch number in OM which is incremented on OM restarts. The epoch is persisted on disk. This epoch will be used to set the first 16 bits of the objectID to ensure that objectIDs are unique even after OM restart.
The highest epoch number is reserved for transactions coming through ratis. This will take care of the scenario where OM ratis is enabled on an existing cluster.~~

To ensure that objectIDs are unique across restarts in non-ratis OM cluster, the transaction index should be updated in DB on every flush to DB. This can be done in a similar fashion to what is being done for ratis enabled cluster today. TransactionInfo table is updated with transaction index as part of every batch write operation to DB.

Also, and epoch number is introduced to ensure that objectIDs do not clash with older clusters in which this fix does not exist. From the 64 bits of ObjectID (long variable), 2 bits are reserved for epoch and 8 bits for recursive directory creation, if required. The most significant 2 bits of objectIDs is set to epoch. For clusters before HDDS-4315 there is no epoch as such. But it can be safely assumed that the most significant 2 bits of the objectID will be 00 (as it unlikely to reach trxn index > 2^62 in an existing cluster). From HDDS-4315 onwards, the Epoch for non-ratis OM clusters will be binary 01 (= decimal 1) and for ratis enabled OM cluster will be binary 10 (= decimal 2).

https://issues.apache.org/jira/browse/HDDS-4315

## How was this patch tested?

Added unit tests.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
@@ -85,7 +85,8 @@ public final class OmUtils {
   // reserved for creating S3G volume on OM start {@link
   // OzoneManager#addS3GVolumeToDB()}.
   public static final long EPOCH_ID_SHIFT = 62; // 64 - 2
-  public static final long MAX_TRXN_ID = (long) (Math.pow(2, 54) - 2);
+  public static final long REVERSE_EPOCH_ID_SHIFT = 2; // 64 - EPOCH_ID_SHIFT
+  public static final long MAX_TRXN_ID = (long) ((1 << 54) - 2);
   public static final int EPOCH_WHEN_RATIS_NOT_ENABLED = 1;
   public static final int EPOCH_WHEN_RATIS_ENABLED = 2;
 
@@ -544,15 +545,15 @@ public final class OmUtils {
    * Get the valid base object id given the transaction id.
    * @param epoch a 2 bit epoch number. The 2 most significant bits of the
    *              object will be set to this epoch.
-   * @param id of the transaction. This value cannot exceed 2^54 - 1 as
+   * @param txId of the transaction. This value cannot exceed 2^54 - 1 as
    *           out of the 64 bits for a long, 2 are reserved for the epoch
    *           and 8 for recursive directory creation.
    * @return base object id allocated against the transaction
    */
-  public static long getObjectIdFromTxId(long epoch, long id) {
-    Preconditions.checkArgument(id <= MAX_TRXN_ID, "TransactionID " +
+  public static long getObjectIdFromTxId(long epoch, long txId) {
+    Preconditions.checkArgument(txId <= MAX_TRXN_ID, "TransactionID " +
         "exceeds max limit of " + MAX_TRXN_ID);
-    return addEpochToObjectId(epoch, id);
+    return addEpochToTxId(epoch, txId);
   }
 
   /**
@@ -561,8 +562,8 @@ public final class OmUtils {
    * when OM is started first time to add S3G volume. In call other cases,
    * getObjectIdFromTxId() should be called to append epoch to objectID.
    */
-  public static long addEpochToObjectId(long epoch, long id) {
-    long lsb54 = id << TRANSACTION_ID_SHIFT;
+  public static long addEpochToTxId(long epoch, long txId) {
+    long lsb54 = txId << TRANSACTION_ID_SHIFT;
     long msb2 = epoch << EPOCH_ID_SHIFT;
 
     return msb2 | lsb54;
@@ -574,7 +575,8 @@ public final class OmUtils {
    */
   @VisibleForTesting
   public static long getTxIdFromObjectId(long objectId) {
-    return ((Long.MAX_VALUE >> 2) & objectId) >> 8;
+    return ((Long.MAX_VALUE >> REVERSE_EPOCH_ID_SHIFT) & objectId)
+        >> TRANSACTION_ID_SHIFT;
   }
 
   /**
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
@@ -17,6 +17,8 @@
 
 package org.apache.hadoop.ozone;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.protobuf.ServiceException;
 import java.io.File;
 import java.io.IOException;
@@ -75,6 +77,17 @@ public final class OmUtils {
   public static final Logger LOG = LoggerFactory.getLogger(OmUtils.class);
   private static final SecureRandom SRAND = new SecureRandom();
   private static byte[] randomBytes = new byte[32];
+
+  private static final long TRANSACTION_ID_SHIFT = 8;
+  // from the 64 bits of ObjectID (long variable), 2 bits are reserved for
+  // epoch and 8 bits for recursive directory creation, if required. This
+  // leaves 54 bits for the transaction ID. Also, the last transaction ID is
+  // reserved for creating S3G volume on OM start {@link
+  // OzoneManager#addS3GVolumeToDB()}.
+  public static final long EPOCH_ID_SHIFT = 62; // 64 - 2
+  public static final long MAX_TRXN_ID = (long) (Math.pow(2, 54) - 2);
+  public static final int EPOCH_WHEN_RATIS_NOT_ENABLED = 1;
+  public static final int EPOCH_WHEN_RATIS_ENABLED = 2;
 
   private OmUtils() {
   }
@@ -360,8 +373,6 @@ public final class OmUtils {
     }
   }
 
-
-
   /**
    * If a OM conf is only set with key suffixed with OM Node ID, return the
    * set value.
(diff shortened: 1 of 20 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -523,6 +531,32 @@ public static OmKeyInfo prepareKeyForRecover(OmKeyInfo keyInfo,
...
+  /**
+   * Get the valid base object id given the transaction id.
+   * @param id of the transaction. This value cannot exceed 2^40 - 1 as
+   *           out of the 64 bits for a long, 16 are reserved for the epoch
+   *           and 8 for recursive directory creation.
+   * @return base object id allocated against the transaction
+   */
+  public static long getObjectIdFromTxId(long epoch, long id) {
+    Preconditions.checkArgument(id <= MAX_TRXN_ID, "TransactionID " +
+        "exceeds max limit of " + MAX_TRXN_ID);
```
linyiqun, line 543: I am thinking for this extreme case, user cannot write object anymore when TransactionID exceeds MAX_TRXN_ID, right? So what can we do for this, have to setup a new Ozone cluster to use?

linyiqun (commented):
@hanishakoneru , I'm +1 for this proposal. But I have one thought below.

prashantpogde:
General comment on using the epoch id that increments with every OM restart. This can get tricky.
 If OM goes in crash restart loop then we have just 2^16 increments available which is 65K attempts. if it takes 1 secs for OM to comeback online we have 65 K secs worth epoch number or 20 hours of crash looping. This is very pessimistic view, it may take several seconds for OM to restart but it does show how
- 16 bit space can be insufficient for this scheme.
- epoch need not be dependent on restart based increment. if it increments based on both of the following conditions
    A) OM restart +
    B) some object gets created after epoch id is incremented
   then epoch may last longer. But even then 16 bit looks insufficient. What  if OM creates one object and restarts.

hanishakoneru:
Thank you @linyiqun and @prashantpogde for the reviews. 

Agree that setting aside 16 bits for epoch doesn't work for both the epoch as well as the transaction ids. 16 bits would not be enough to cover restarts and 40 bits might not be enough for transaction ids. 
The new proposal is to have only 2 bits set aside for epoch. For non-Ratis OM, the transactionIndex will be saved in DB with every sync operation. When OM is restarted, this transactionIndex will be read from DB so that new transactions do not have clashing indices.
The epoch would let us distinguish objects created before and after this upgrade. This would help if someone needs to fix the duplicate objectIDs in existing clusters. 

Thank you @bharatviswa504 and @prashantpogde for the offline discussion.

hanishakoneru:
@prashantpogde, @linyiqun, @bharatviswa504, I have updated the PR with the discussed approach. Please review when you get a chance. Thanks.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -524,6 +535,48 @@ public static OmKeyInfo prepareKeyForRecover(OmKeyInfo keyInfo,
...
+    return addEpochToObjectId(epoch, id);
+  }
+
+  /**
+   * Note - This function should not be called directly. It is directly called
+   * only from OzoneManager#addS3GVolumeToDB() which is a one time operation
+   * when OM is started first time to add S3G volume. In call other cases,
+   * getObjectIdFromTxId() should be called to append epoch to objectID.
+   */
+  public static long addEpochToObjectId(long epoch, long id) {
```
linyiqun, line 564: Can we rename id -> trxnId

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -394,6 +403,8 @@ private OzoneManager(OzoneConfiguration conf) throws IOException,
         OMConfigKeys.OZONE_OM_RATIS_ENABLE_KEY,
         OMConfigKeys.OZONE_OM_RATIS_ENABLE_DEFAULT);
 
+    omEpoch = OmUtils.getOMEpoch(isRatisEnabled);
```
linyiqun, line 406: We can reuse metadataManager#getOmEpoch to set epoch value.

linyiqun (commented):
@hanishakoneru , the new implementation looks good to me. Only minor comments from me below.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -76,6 +78,17 @@
...
   private static byte[] randomBytes = new byte[32];
 
+  private static final long TRANSACTION_ID_SHIFT = 8;
+  // from the 64 bits of ObjectID (long variable), 2 bits are reserved for
+  // epoch and 8 bits for recursive directory creation, if required. This
+  // leaves 54 bits for the transaction ID. Also, the last transaction ID is
+  // reserved for creating S3G volume on OM start {@link
+  // OzoneManager#addS3GVolumeToDB()}.
+  public static final long EPOCH_ID_SHIFT = 62; // 64 - 2
+  public static final long MAX_TRXN_ID = (long) (Math.pow(2, 54) - 2);
```
prashantpogde, line 88: can we use  1 << 54  instead of Math.pow ?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -76,6 +78,17 @@
...
+  public static final int EPOCH_WHEN_RATIS_NOT_ENABLED = 1;
```
prashantpogde, line 89: Don't we want these values to be 0 and 1 instead of 1 & 2 ?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -524,6 +535,48 @@ public static OmKeyInfo prepareKeyForRecover(OmKeyInfo keyInfo,
...
+   *              object will be set to this epoch.
+   * @param id of the transaction. This value cannot exceed 2^54 - 1 as
+   *           out of the 64 bits for a long, 2 are reserved for the epoch
+   *           and 8 for recursive directory creation.
+   * @return base object id allocated against the transaction
+   */
+  public static long getObjectIdFromTxId(long epoch, long id) {
+    Preconditions.checkArgument(id <= MAX_TRXN_ID, "TransactionID " +
+        "exceeds max limit of " + MAX_TRXN_ID);
+    return addEpochToObjectId(epoch, id);
```
prashantpogde, line 555: nit : s/addEpochToObjectId /addEpochToTxnId since your definition is ObjectId = EpochId+TxnId

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
```diff
@@ -524,6 +535,48 @@ public static OmKeyInfo prepareKeyForRecover(OmKeyInfo keyInfo,
...
+    return msb2 | lsb54;
+  }
+
+  /**
+   * Given an objectId, unset the 2 most significant bits to get the
+   * corresponding transaction index.
+   */
+  @VisibleForTesting
+  public static long getTxIdFromObjectId(long objectId) {
+    return ((Long.MAX_VALUE >> 2) & objectId) >> 8;
```
prashantpogde, line 577: consider EpochIdShift and TRANSACTION_ID_SHIFT here instead of hardcoded value.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -306,6 +306,15 @@
   private final Map<String, RatisDropwizardExports> ratisMetricsMap =
       new ConcurrentHashMap<>();
 
+  // Epoch is used to generate the objectIDs. The most significant 2 bits of
+  // objectIDs is set to this epoch. For clusters before HDDS-4315 there is
+  // no epoch as such. But it can be safely assumed that the most significant
+  // 2 bits of the objectID will be 00. From HDDS-4315 onwards, the Epoch for
+  // non-ratis OM clusters will be binary 01 (= decimal 1)  and for ratis
```
prashantpogde, line 313: why differentiate between epoch before this change and non-ratis OM cluster ? both can be 0 ?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
hanishakoneru: Wanted to avoid 0 as we can assume that currently it is 0. This would give us a way to separate out objectIds created before this fix. If ever, these non-unique objectIds need to be fixed, it would be easy to identify them.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
hanishakoneru: It would help if we ever wanted to update the non-unique objectIds to maintain uniqueness throughout.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
hanishakoneru: Done.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
hanishakoneru: Done.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
hanishakoneru: Done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
hanishakoneru: Done.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OmUtils.java
hanishakoneru: Done.

hanishakoneru:
Thank you @linyiqun and @prashantpogde for the reviews. I have addressed the comments. Please take a look.

linyiqun (approved):
+1. Thanks for addressing the comments, @hanishakoneru .

prashantpogde (commented):
+1 LGTM

prashantpogde (approved):

hanishakoneru:
Thanks @linyiqun and @prashantpogde for the reviews.

PR #1484, opened 2020-10-07
HDDS-4322. Add integration tests for Decommission and resolve issues detected by the tests.

sodonnel:

Add a series of integration tests to prove decommission works, and that decommission can survive a restart of SCM.

As part of adding these tests, some issues were discover that were fixed in the process of debugging the tests.

https://issues.apache.org/jira/browse/HDDS-4322

## How was this patch tested?

New integration tests added

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
@@ -476,6 +476,9 @@ public class DatanodeDetails extends NodeImpl implements
       this.setupTime = details.getSetupTime();
       this.revision = details.getRevision();
       this.buildDate = details.getBuildDate();
+      this.persistedOpState = details.getPersistedOpState();
+      this.persistedOpStateExpiryEpochSec =
+          details.getPersistedOpStateExpiryEpochSec();
       return this;
     }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/DatanodeAdminMonitorImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/DatanodeAdminMonitorImpl.java
@@ -294,6 +294,9 @@ public class DatanodeAdminMonitorImpl implements DatanodeAdminMonitor {
             "in containerManager", cid, dn);
       }
     }
+    LOG.info("{} has {} sufficientlyReplicated, {} underReplicated and {} " +
+        "unhealthy containers",
+        dn, sufficientlyReplicated, underReplicated, unhealthy);
     return underReplicated == 0 && unhealthy == 0;
   }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NewNodeHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NewNodeHandler.java
@@ -20,9 +20,14 @@ package org.apache.hadoop.hdds.scm.node;
 
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
+import org.apache.hadoop.hdds.scm.container.ContainerReportHandler;
+import org.apache.hadoop.hdds.scm.node.states.NodeNotFoundException;
 import org.apache.hadoop.hdds.scm.pipeline.PipelineManager;
 import org.apache.hadoop.hdds.server.events.EventHandler;
 import org.apache.hadoop.hdds.server.events.EventPublisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Handles New Node event.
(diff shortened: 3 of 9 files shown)


elek (commented):
Thank you to create this patch @sodonnel 

> some issues were discover that were fixed in the process of debugging the tests.

Can you please share more details. It would help me to understand the changes.

(Based on the comment on `newNodeStatus`, it can be related to the case when DN report back a specific state which requires running monitor. )

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/DatanodeAdminMonitorImpl.java
```diff
@@ -294,6 +294,9 @@ private boolean checkContainersReplicatedOnNode(DatanodeDetails dn)
             "in containerManager", cid, dn);
       }
     }
+    LOG.info("{} has {} sufficientlyReplicated, {} underReplicated and {} " +
```
elek, line 297: Do we need it on info level? What is the expected frequency of this message?

I can create a decom insight point to make it easier to see this message, if it's required for debugging...

sodonnel: That message would be printed for each decommissioned or maintenance node for each run of the replication monitor. The default now is 30 seconds, but that is probably too short, and we will need to change that default to something longer, maybe a minute or two.

I think it is useful to have this information in the logs from a supportability perspective. It lets us see the progress of decommission, whether a node appears to be stuck, etc, and a few messages per minute only when nodes are decommissioning is not excessively noisy.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
```diff
@@ -476,6 +476,9 @@ public Builder setDatanodeDetails(DatanodeDetails details) {
       this.setupTime = details.getSetupTime();
       this.revision = details.getRevision();
       this.buildDate = details.getBuildDate();
+      this.persistedOpState = details.getPersistedOpState();
+      this.persistedOpStateExpiryEpochSec =
+          details.getPersistedOpStateExpiryEpochSec();
       return this;
```
sodonnel, line 482: These fields had been missed from the DatanodeDetails builder object, so when the DN reported back its "persisted state" the DN was always IN_SERVICE. Adding this change fixed that problem and allowed the state to be returned correctly.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NewNodeHandler.java
```diff
@@ -20,27 +20,49 @@
...
+      try {
+        decommissionManager.continueAdminForNode(datanodeDetails);
+      } catch (NodeNotFoundException e) {
+        // Should not happen, as the node has just registered to call this event
+        // handler.
+        LOG.warn("NodeNotFound when adding the node to the decommissionManager",
+            e);
+      }
+    }
   }
```
sodonnel, line 67: If a DN is DECOMMISSIONING, and then SCM is restarted, SCM will forget all about the decommission nodes. Then the nodes will re-register with SCM and report they are DECOMMISSIONING. If the node is DECOMMISSIONING rather than DECOMMISSIONED, we need to get it back into the decommission workflow. This NewNodeHandler is invoked for a new registration of a Node, so this change continues the decom process.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
```diff
@@ -239,11 +239,32 @@ private void initializeStateMachines() {
...
+          dnOpState, dn.getPersistedOpStateExpiryEpochSec());
+      return new NodeStatus(dnOpState, nodeHealthSM.getInitialState(),
+          dn.getPersistedOpStateExpiryEpochSec());
+    } else {
+      return new NodeStatus(
+          NodeOperationalState.IN_SERVICE, nodeHealthSM.getInitialState());
+    }
+  }
+
   /**
```
sodonnel, line 268: This code was lifted out of SCMNodeManager as it fits better here. The NodeStatus is now set as part of registration and before any events are triggered (eg PipelineCreation). Before there was a race condition, where the NodeStatus was IN_SERVICE and pipeline creation could be triggered. Then the state was quickly changed to DECOMMISSIONED etc.

sodonnel:
Thanks for taking a look @elek 

> Can you please share more details. It would help me to understand the changes.

I added some comments inline with each change that explains the reasoning behind them. There were two main problems:

1) The state reported by the DN was not reflected in DatanodeDetails seen in SCM. This was due to missing fields in the builder object and resulted in decommission not being able to survive a SCM restart.

2) After an SCM restart when a node reported as DECOMMISSIONING or ENTERING_MAINTENANCE, it failed to make any further progress and was stuck in that state. It needed to be put back into the DatanodeAdminMonitor to finish out the process.

3) There was a small race condition, which may not have caused any problems but was better to fix, in the node status getting set and then updated on registration.

Each of the problems was discovered via one of the new tests and then corrected to get the test to pass.

swagle (approved):
+1, since earlier review comments were addressed.

PR #1485, opened 2020-10-08
HDDS-4325. Incompatible return codes from Ozone getconf -confKey

adoroszlai:

Make `ozone getconf -confKey` return error (255) instead of success (0) for missing key, to be compatible with Ozone 1.0.

https://issues.apache.org/jira/browse/HDDS-4325

## How was this patch tested?

Manually verified that output and exit codes with patch:

```
$ ozone getconf -confKey asdf; echo $?
Configuration asdf is missing.
255
$ ozone getconf -confKey ozone.om.address; echo $?
om
0
```

match 1.0.0:

```
$ ozone getconf -confKey asdf; echo $?
Configuration asdf is missing.
255
$ ozone getconf -confKey ozone.om.address; echo $?
om
0
```

Added check for exit code in acceptance test case for missing key.  Also added test case for existing key.

https://github.com/adoroszlai/hadoop-ozone/actions/runs/295406415

### DIFF
--- a/hadoop-ozone/dist/src/main/smoketest/basic/getconf.robot
+++ b/hadoop-ozone/dist/src/main/smoketest/basic/getconf.robot
@@ -14,20 +14,33 @@
 # limitations under the License.
 
 *** Settings ***
-Documentation       Smoketest ozone cluster startup
-Library             OperatingSystem
-Resource            ../commonlib.robot
+Documentation       Test 'ozone getconf' command
+Resource            ../lib/os.robot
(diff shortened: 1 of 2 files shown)


umamaheswararao:
Thanks @adoroszlai for working on this. The changes looks good to me.
+1

umamaheswararao (approved):

adoroszlai:
Thanks @umamaheswararao for reviewing and merging it.

PR #1486, opened 2020-10-09
HDDS-4296. SCM changes to process Layout Info in heartbeat request/response

prashantpogde:

SCM changes to process Layout Info in heartbeat request/response

https://issues.apache.org/jira/browse/HDDS-4296

## How was this patch tested?

UT. I will fix any CI failure.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
@@ -18,6 +18,10 @@
 
 package org.apache.hadoop.ozone.upgrade;
 
+import static org.apache.hadoop.ozone.upgrade.UpgradeFinalizer.Status.ALREADY_FINALIZED;
+import static org.apache.hadoop.ozone.upgrade.UpgradeFinalizer.Status.FINALIZATION_DONE;
+import static org.apache.hadoop.ozone.upgrade.UpgradeFinalizer.Status.FINALIZATION_REQUIRED;
+
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -39,6 +43,8 @@ public abstract class AbstractLayoutVersionManager<T extends LayoutFeature>
   protected TreeMap<Integer, T> features = new TreeMap<>();
   protected Map<String, T> featureMap = new HashMap<>();
   protected volatile boolean isInitialized = false;
+  protected volatile UpgradeFinalizer.Status currentUpgradeState =
+      FINALIZATION_REQUIRED;
 
   protected void init(int version, T[] lfs) throws IOException {
 
@@ -52,8 +58,14 @@ public abstract class AbstractLayoutVersionManager<T extends LayoutFeature>
             String.format("Cannot initialize VersionManager. Metadata " +
                     "layout version (%d) > software layout version (%d)",
                 metadataLayoutVersion, softwareLayoutVersion));
+      } else if (metadataLayoutVersion == softwareLayoutVersion) {
+        currentUpgradeState = ALREADY_FINALIZED;
       }
     }
+  }
+
+  public UpgradeFinalizer.Status getUpgradeState() {
+    return currentUpgradeState;
   }
 
   private void initializeFeatures(T[] lfs) {
@@ -71,6 +83,7 @@ public abstract class AbstractLayoutVersionManager<T extends LayoutFeature>
     featureMap.clear();
     features.clear();
     isInitialized = false;
+    currentUpgradeState = ALREADY_FINALIZED;
   }
 
   public void finalized(T layoutFeature) {
@@ -90,6 +103,10 @@ public abstract class AbstractLayoutVersionManager<T extends LayoutFeature>
           msgStart + "Software Layout version: " + softwareLayoutVersion
               + " Feature Layout version: " + layoutFeature.layoutVersion());
     }
+  }
+
+  public void completeFinalization() {
+    currentUpgradeState = FINALIZATION_DONE;
   }
 
   private boolean softwareIsBehindMetaData() {
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
@@ -49,6 +49,7 @@ import org.apache.hadoop.hdds.scm.VersionInfo;
 import org.apache.hadoop.hdds.scm.container.ContainerID;
 import org.apache.hadoop.hdds.scm.container.placement.metrics.SCMNodeMetric;
 import org.apache.hadoop.hdds.scm.container.placement.metrics.SCMNodeStat;
+import org.apache.hadoop.hdds.scm.events.SCMEvents;
 import org.apache.hadoop.hdds.scm.net.NetworkTopology;
 import org.apache.hadoop.hdds.scm.node.states.NodeAlreadyExistsException;
 import org.apache.hadoop.hdds.scm.node.states.NodeNotFoundException;
@@ -65,8 +66,10 @@ import org.apache.hadoop.net.TableMapping;
 import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.protocol.VersionResponse;
 import org.apache.hadoop.ozone.protocol.commands.CommandForDatanode;
+import org.apache.hadoop.ozone.protocol.commands.FinalizeNewLayoutVersionCommand;
 import org.apache.hadoop.ozone.protocol.commands.RegisteredCommand;
 import org.apache.hadoop.ozone.protocol.commands.SCMCommand;
+import org.apache.hadoop.ozone.upgrade.UpgradeFinalizer;
 import org.apache.hadoop.util.ReflectionUtils;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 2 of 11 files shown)


avijayanhwx:
cc @fapifta / @sodonnel Please review.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
```diff
@@ -107,6 +121,10 @@ public void doFinalize(Object param) {
     }
   }
 
+  public void completeFinalize(Object param) {
+    currentUpgradeState = Finalized;
```
linyiqun, line 125: param not used here.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/OzoneUpgradeState.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.apache.hadoop.ozone.upgrade;
+
+/**
+ * Generic UpgradeState or Ozone.
+ */
+public enum OzoneUpgradeState {
+
+  InValid(0, "Invalid State"),
```
linyiqun, line 26: Can we use uppercase letter here, like INVALID, PRE_FINALIZE.. FINALIZED.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/LayoutVersionReportHandler.java
```diff
@@ -0,0 +1,54 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Handles Node Layout Version Reports from DataNode.
+ */
+public class LayoutVersionReportHandler implements
+    EventHandler<LayoutReportFromDatanode> {
+
+  private static final Logger LOGGER = LoggerFactory
+      .getLogger(NodeReportHandler.class);
```
linyiqun, line 36: Wrong log instance name used.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -400,6 +405,66 @@ public void processNodeReport(DatanodeDetails datanodeDetails,
...
+        // this state but it can not be made part of any Pipeline.
+
+        // Also send Finalize command to the data node. Its OK to
+        // send Finalize command multiple times.
+        scmNodeEventPublisher.fireEvent(SCMEvents.DATANODE_COMMAND,
+            new CommandForDatanode<>(datanodeDetails.getUuid(),
+                new FinalizeNewLayoutVersionCommand(true,
+                    LayoutVersionProto.newBuilder()
+                        .setSoftwareLayoutVersion(dnSlv)
+                        .setMetadataLayoutVersion(dnSlv).build())));
```
linyiqun, line 463: Here the dnSlv is returned in finalize command, is that right? should be scmSlv/scmMlv?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -400,6 +405,66 @@ public void processNodeReport(DatanodeDetails datanodeDetails,
...
+          scmLayoutVersionManager.getUpgradeState();
+      int dnSlv = layoutVersionReport.getSoftwareLayoutVersion();
+      int dnMlv = layoutVersionReport.getMetadataLayoutVersion();
+
+      // If the data node slv is > scm slv => log error condition
+      if (dnSlv > scmSlv) {
+        LOG.error("Rogue data node {} in the cluster. " +
+                "DataNode SoftwareLayoutVersion = {}, SCM " +
+                "SoftwareLayoutVersion = {}",
+            datanodeDetails.getHostName(), dnSlv, scmSlv);
```
linyiqun, line 440: Can we put this node into exclude list temporarily? I mean SCM should take some action when it finds some unexpected version DNs.

linyiqun (commented):
Hi @prashantpogde , some initial review comments from me. 
It would be better if you can additionally add corresponding unit test for this PR change.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/AbstractLayoutVersionManager.java
prashantpogde: yup. eventually I would remove it if not used.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/OzoneUpgradeState.java
prashantpogde: yup

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/LayoutVersionReportHandler.java
prashantpogde: yup.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
prashantpogde: yes, its asking datanode to move to dnSlv because thats the software version datanode is running. Its possible that its lower than the SCM slv but dnSlv is the correct version for that particular datanode,.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
prashantpogde: Not sure what the temporary would be. Eventually the plan is that rogue data-node would be highlighted with the Recon as well and admins would be able to take some action on such nodes. 
For normal cases, Finalize would be a no-op for datanode.

prashantpogde: I will upload the changes together with all review comments addressed.

avijayanhwx (commented):
Thanks for working on this @prashantpogde. A couple of general comments.

- @fapifta has also introduced states a component can be in, with respect to Upgrade. We should discuss and use just one set of states.

- I am not convinced why we need another report for Layout Version handling. Layout version is like node metadata, and should be sent in every heartbeat. Also, I am not sure if we can handle the LayoutVersion report through an event handler since it is 'async' and there can be delay in handling.

prashantpogde:
Updates new set of changes after taking care of all review comments. Please take a look.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -400,6 +405,66 @@ public void processNodeReport(DatanodeDetails datanodeDetails,
...
+    if (layoutVersionReport != null) {
+      int scmSlv = scmLayoutVersionManager.getSoftwareLayoutVersion();
+      int scmMlv = scmLayoutVersionManager.getMetadataLayoutVersion();
+      UpgradeFinalizer.Status scmUpgradeState =
+          scmLayoutVersionManager.getUpgradeState();
+      int dnSlv = layoutVersionReport.getSoftwareLayoutVersion();
+      int dnMlv = layoutVersionReport.getMetadataLayoutVersion();
+
+      // If the data node slv is > scm slv => log error condition
+      if (dnSlv > scmSlv) {
```
avijayanhwx, line 436: Do we also need to add the condition (dnMLV > scmMLV) here (and at register step)?

avijayanhwx (commented):
@prashantpogde Patch generally looks good to me. Can we add a unit test for the layout version check during heartbeat processing?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
prashantpogde: - yes, we need the additional check here. The nodes that are already registered they will send the heartbeat when they are booted.
- I will add an end to end test case for this once the workflow is complete.

avijayanhwx: @prashantpogde End to End test cases make sense for integration or acceptance tests. Is it possible to add a unit test just for the SCMNodeManager.processLayoutVersion method? Since it has a few condition checks (path flows), it may be a good candidate for a unit test based on mocks.

prashantpogde: yup

avijayanhwx (approved):
LGTM +1.

@prashantpogde Are you planning to comment out the other 2 unused variable findbug failures as well? 
Can we trigger an empty commit to get acceptance tests(unsecure) to pass?

prashantpogde:
Failure in TestOzoneManagerHAMetadataOnly is unrelated with the changes.

avijayanhwx:
Thank you for the patch @prashantpogde and the review @linyiqun.

PR #1487, opened 2020-10-12
HDDS-4318. Disable single node pipeline creation by default in Ozone.

aryangupta1998:

Currently, single node pipeline creation is ON by default in ozone, though its not used by default in Ozone write path. It would be good to disable this by turning off the config "ozone.scm.pipeline.creation.auto.factor.one" by default. It may lead to some unit test failures and for those tests , this config needs to b explicitly set to true.

https://issues.apache.org/jira/browse/HDDS-4318

## How was this patch tested?

Tested manually


elek:
Thanks to work on this @aryangupta1998. 

> Currently, single node pipeline creation is ON by default in ozone, though its not used by default in Ozone write path.

Can you please help me to understand the details? Is the standalone GRPC endpoint endpoint used for RATIS/ONE write?

One note: it has a little risk when we modify the default, but all of tests use the old version. The new default won't be tested...

PR #1489, opened 2020-10-13
HDDS-4308. Fix issue with quota update

captainzmc:

Currently volumeArgs using getCacheValue and put the same object in doubleBuffer, this might cause issue.

Let's take the below scenario:

InitialVolumeArgs quotaBytes -> 10000
1. T1 -> Update VolumeArgs, and subtracting 1000 and put this updated volumeArgs to DoubleBuffer.
2. T2-> Update VolumeArgs, and subtracting 2000 and has not still updated to double buffer.

Now at the end of flushing these transactions, our DB should have 7000 as bytes used.

Now T1 is picked by double Buffer and when it commits, and as it uses cached Object put into doubleBuffer, it flushes to DB with the updated value from T2(As it is a cache object) and update DB with bytesUsed as 7000.

And now OM has restarted, and only DB has transactions till T1. (We get this info from TransactionInfo Table(https://issues.apache.org/jira/browse/HDDS-3685)

Now T2 is again replayed, as it is not committed to DB, now DB will be again subtracted with 2000, and now DB will have 5000.

But after T2, the value should be 7000, so we have DB in an incorrect state.

Issue here:
1. As we use a cached object and put the same cached object into double buffer this can cause this kind of issue.

https://issues.apache.org/jira/browse/HDDS-4308

## How was this patch tested?

Use the existing UT

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
@@ -20,7 +20,6 @@ package org.apache.hadoop.ozone.om.helpers;
 
 import java.util.ArrayList;
 import java.util.BitSet;
-import java.util.concurrent.atomic.LongAdder;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -80,7 +79,7 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
 
   private final String sourceBucket;
 
-  private final LongAdder usedBytes = new LongAdder();
+  private long usedBytes;
 
   private long quotaInBytes;
   private long quotaInCounts;
@@ -132,7 +131,7 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
     this.bekInfo = bekInfo;
     this.sourceVolume = sourceVolume;
     this.sourceBucket = sourceBucket;
-    this.usedBytes.add(usedBytes);
+    this.usedBytes = usedBytes;
     this.quotaInBytes = quotaInBytes;
     this.quotaInCounts = quotaInCounts;
   }
@@ -241,9 +240,14 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
   }
 
 
-  public LongAdder getUsedBytes() {
+  public long getUsedBytes() {
     return usedBytes;
   }
+
+  public void incrUsedBytes(long bytes) {
+    this.usedBytes = this.usedBytes + bytes;
+  }
+
   public long getQuotaInBytes() {
     return quotaInBytes;
   }
@@ -324,7 +328,7 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
         .setSourceBucket(sourceBucket)
         .setAcls(acls)
         .addAllMetadata(metadata)
-        .setUsedBytes(usedBytes.sum())
+        .setUsedBytes(usedBytes)
         .setQuotaInBytes(quotaInBytes)
         .setQuotaInCounts(quotaInCounts);
   }
@@ -489,7 +493,7 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
         .setModificationTime(modificationTime)
         .setObjectID(objectID)
         .setUpdateID(updateID)
-        .setUsedBytes(usedBytes.sum())
+        .setUsedBytes(usedBytes)
         .addAllMetadata(KeyValueUtil.toProtobuf(metadata))
         .setQuotaInBytes(quotaInBytes)
         .setQuotaInCounts(quotaInCounts);
@@ -557,7 +561,7 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
         ", isVersionEnabled='" + isVersionEnabled + "'" +
         ", storageType='" + storageType + "'" +
         ", creationTime='" + creationTime + "'" +
-        ", usedBytes='" + usedBytes.sum() + "'" +
+        ", usedBytes='" + usedBytes + "'" +
         ", quotaInBytes='" + quotaInBytes + "'" +
         ", quotaInCounts='" + quotaInCounts + '\'' +
         sourceInfo +
@@ -582,7 +586,7 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
         storageType == that.storageType &&
         objectID == that.objectID &&
         updateID == that.updateID &&
-        usedBytes.sum() == that.usedBytes.sum() &&
+        usedBytes == that.usedBytes &&
         Objects.equals(sourceVolume, that.sourceVolume) &&
         Objects.equals(sourceBucket, that.sourceBucket) &&
         Objects.equals(metadata, that.metadata) &&
@@ -609,7 +613,7 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
         ", objectID=" + objectID +
         ", updateID=" + updateID +
         ", metadata=" + metadata +
-        ", usedBytes=" + usedBytes.sum() +
+        ", usedBytes=" + usedBytes +
         ", quotaInBytes=" + quotaInBytes +
         ", quotaInCounts=" + quotaInCounts +
         '}';
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
@@ -245,7 +245,7 @@ public final class OmBucketInfo extends WithObjectID implements Auditable {
   }
 
   public void incrUsedBytes(long bytes) {
-    this.usedBytes = this.usedBytes + bytes;
+    this.usedBytes += bytes;
   }
 
   public long getQuotaInBytes() {
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmVolumeArgs.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmVolumeArgs.java
@@ -23,7 +23,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.concurrent.atomic.LongAdder;
 
 import org.apache.hadoop.ozone.OzoneAcl;
 import org.apache.hadoop.ozone.OzoneConsts;
@@ -47,7 +46,6 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
   private long quotaInBytes;
   private long quotaInCounts;
   private final OmOzoneAclMap aclMap;
-  private final LongAdder usedBytes = new LongAdder();
 
   /**
    * Private constructor, constructed via builder.
@@ -57,10 +55,9 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
    * @param quotaInBytes - Volume Quota in bytes.
    * @param quotaInCounts - Volume Quota in counts.
    * @param metadata - metadata map for custom key/value data.
-   * @param usedBytes - Volume Quota Usage in bytes.
    * @param aclMap - User to access rights map.
    * @param creationTime - Volume creation time.
-   * @param  objectID - ID of this object.
+   * @param objectID - ID of this object.
    * @param updateID - A sequence number that denotes the last update on this
    * object. This is a monotonically increasing number.
    */
@@ -68,15 +65,14 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
       "builder."})
   private OmVolumeArgs(String adminName, String ownerName, String volume,
       long quotaInBytes, long quotaInCounts, Map<String, String> metadata,
-      long usedBytes, OmOzoneAclMap aclMap, long creationTime,
-      long modificationTime, long objectID, long updateID) {
+      OmOzoneAclMap aclMap, long creationTime, long modificationTime,
+      long objectID, long updateID) {
     this.adminName = adminName;
     this.ownerName = ownerName;
     this.volume = volume;
     this.quotaInBytes = quotaInBytes;
     this.quotaInCounts = quotaInCounts;
     this.metadata = metadata;
-    this.usedBytes.add(usedBytes);
     this.aclMap = aclMap;
     this.creationTime = creationTime;
     this.modificationTime = modificationTime;
@@ -177,10 +173,6 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
     return aclMap;
   }
 
-  public LongAdder getUsedBytes() {
-    return usedBytes;
-  }
-
   /**
    * Returns new builder class that builds a OmVolumeArgs.
    *
@@ -204,8 +196,6 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
         String.valueOf(this.quotaInCounts));
     auditMap.put(OzoneConsts.OBJECT_ID, String.valueOf(this.getObjectID()));
     auditMap.put(OzoneConsts.UPDATE_ID, String.valueOf(this.getUpdateID()));
-    auditMap.put(OzoneConsts.USED_BYTES,
-        String.valueOf(this.usedBytes));
     return auditMap;
   }
 
@@ -241,7 +231,6 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
     private OmOzoneAclMap aclMap;
     private long objectID;
     private long updateID;
-    private long usedBytes;
 
     /**
      * Sets the Object ID for this Object.
@@ -319,11 +308,6 @@ public final class OmVolumeArgs extends WithObjectID implements Auditable {
       return this;
     }
 
-    public Builder setUsedBytes(long quotaUsage) {
-      this.usedBytes = quotaUsage;
-      return this;
-    }
-
     public Builder addOzoneAcls(OzoneAclInfo acl) throws IOException {
       aclMap.addAcl(acl);
       return this;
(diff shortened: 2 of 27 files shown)


captainzmc:
Hi @bharatviswa504, Please help to review this PR.

captainzmc:
Hi @linyiqun,I modified the implementation based on the latest comments. Can you help to review this PR.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
```diff
@@ -597,27 +596,40 @@ protected boolean checkDirectoryAlreadyExists(String volumeName,
...
+  /**
+   * Return volume info that updated usageBytes for the specified volume. And
+   * check Volume usageBytes quota.
+   * @param omMetadataManager
+   * @param volume
+   * @param updateUsage
+   * @return OmVolumeArgs
+   * @throws IOException
+   */
+  protected static synchronized OmVolumeArgs syncCheckAndUpdateUsage(
```
linyiqun, line 624: I don't prefer to add new methods here, this makes current PR not clear to understand.
@captainzmc , can you make the minor adjustment for getVolumeInfo as I suggested in JIRA HDDS-4308.

captainzmc: Thanks for @linyiqun's review.
Modifying getVolumeInfo to synchronized and get copyObject would not suffice. There is only one instance of volumeArgs in memory, and we need to update volumeArgs atomic after getVolumeInfo. Then get the value of the copyObject.
So, I made a modification based on your suggestion, added update volumeArgs's usedBytes to the getVolumeInfo method.

bharatviswa504: I have not understood one part, how without volume lock this will help here.  

Because other threads can be updating volumeArgs when this update is happening/ other threads read omVolumeArgs.

bharatviswa504: Do we need LongAdder still?

captainzmc: > I have not understood one part, how without volume lock this will help here.
> 
> Because other threads can be updating volumeArgs when this update is happening/ other threads read omVolumeArgs.

This is a possible case. Can lead to update usageBytes inaccurate. 
Suppose that ThreadA update VolumeArgs using OMVolumeSetOwnerRequest, it gets VolumeArgs from DB if usageBytes is 1000 at this point. When ThreadB writes the new key using OMKeyCreateRequest, he changes usageBytes to 1100.
There is no Volume lock at this point, so ThreadB may finish executing first and ThreadA last. The Final usageBytes for VolumeArgs would be 1000. This should be a problem.
Maybe we still can't avoid using Volume lock. Any suggestions here? @linyiqun

captainzmc: > Do we need LongAdder still?
LongAdder will no longer be necessary if there are no concurrent updates to usedBytes.

captainzmc: > Do we need LongAdder still?

LongAdder will no longer be necessary if there are no concurrent updates to usedBytes.

linyiqun: >Thank for @bharatviswa504's feedback. This is a possible case. Can lead to update usageBytes inaccurate.
Suppose that ThreadA update VolumeArgs using OMVolumeSetOwnerRequest, it gets VolumeArgs from DB if usageBytes is 1000 at this point. When ThreadB writes the new key using OMKeyCreateRequest, he changes usageBytes to 1100.

Good catch, @captainzmc . Seems we still use volume lock to help us do the volume quota update control. We should acquire volume lock before get volume info (copy volume info) and release volume lock after add the OMClientResponse into double buffer cache.

captainzmc:
Hi @bharatviswa504 @linyiqun, This PR has been modified. Can you help to review this PR.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmVolumeArgs.java
```diff
@@ -177,10 +176,14 @@ public OmOzoneAclMap getAclMap() {
     return aclMap;
   }
 
-  public LongAdder getUsedBytes() {
+  public long getUsedBytes() {
     return usedBytes;
   }
 
+  public void setUsedBytes(long usedBytes) {
+    this.usedBytes = this.usedBytes + usedBytes;
```
linyiqun, line 184: Method name setUsedBytes seems confused, can we rename to incrUsedBytes(long bytes)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -300,8 +302,26 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+      }
+
+      acquireVolumeLock = omMetadataManager.getLock().acquireWriteLock(
+          VOLUME_LOCK, volumeName);
+      omVolumeArgs.setUsedBytes(preAllocatedSpace);
+      OmVolumeArgs copyVolumeArgs = omVolumeArgs.copyObject();
+      if (acquireVolumeLock) {
+        omMetadataManager.getLock().releaseWriteLock(VOLUME_LOCK, volumeName);
+        acquireVolumeLock = false;
+      }
```
linyiqun, line 322: We should not release volume lock until add omClientResponse into double buffer. So that we can make sure stale OmVolumeArgs in other OM request won't be earlier processed than this request. Here we should ensure that the OmVolumeArgs is updated in order.

I prefer to add volume lock for the whole method.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -300,8 +302,26 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-      omVolumeArgs.getUsedBytes().add(preAllocatedSpace);
       omBucketInfo.getUsedBytes().add(preAllocatedSpace);
+      OmBucketInfo copyBucketInfo = omBucketInfo.copyObject();
+      // We cannot acquire VOLUME_LOCK while holding BUCKET_LOCK. So
+      // release BUCKET_LOCK first.
+      if (acquiredLock) {
+        omMetadataManager.getLock().releaseWriteLock(BUCKET_LOCK, volumeName,
+            bucketName);
+        acquiredLock = false;
+      }
```
linyiqun, line 313: The similar concurrent update issue for bucket usedBytes will happens if we release bucket lock here and then try to acquire bucket lock again.
Seems we should both lock the volume and bucket lock for the whole method.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCommitPartRequest.java
```diff
@@ -222,16 +218,34 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+          volumeName, bucketName);
 
       omResponse.setCommitMultiPartUploadResponse(
           MultipartCommitUploadPartResponse.newBuilder()
               .setPartName(partName));
       omClientResponse = new S3MultipartUploadCommitPartResponse(
           omResponse.build(), multipartKey, openKey,
           multipartKeyInfo, oldPartKeyInfo, omKeyInfo,
-          ozoneManager.isRatisEnabled(), omVolumeArgs, omBucketInfo);
+          ozoneManager.isRatisEnabled(), copyVolumeArgs, copyBucketInfo);
```
linyiqun, line 248: We should also update below lines to use copyVolumeArgs,  copyBucketInfo.
```java
    } catch (IOException ex) {
      result = Result.FAILURE;
      exception = ex;
      omClientResponse = new S3MultipartUploadCommitPartResponse(
          createErrorOMResponse(omResponse, exception), multipartKey, openKey,
          multipartKeyInfo, oldPartKeyInfo, omKeyInfo,
          ozoneManager.isRatisEnabled(), omVolumeArgs, omBucketInfo);
    } finally {
    ...
    }
```

linyiqun (commented):
Hi @captainzmc , please have a look for the latest comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -206,8 +206,8 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           IAccessAuthorizer.ACLType.CREATE, OzoneObj.ResourceType.KEY);
 
       // acquire lock
-      acquiredLock = omMetadataManager.getLock().acquireWriteLock(BUCKET_LOCK,
-          volumeName, bucketName);
+      acquireVolumeLock = omMetadataManager.getLock().acquireWriteLock(
+          VOLUME_LOCK, volumeName);
```
linyiqun, line 211: >The similar concurrent update issue for bucket usedBytes will happens if we release bucket lock here and then try to acquire bucket lock again.
As mentioned above, we should both lock the volume and bucket lock for the whole method.

@captainzmc , can you address above comment, bucket lock is also needed since we also do the bucket used bytes update. Can you update this in all below places?

Other change looks good to me.

captainzmc: Thanks  for @linyiqun's feedback. A new commit has been committed to fix this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -206,6 +208,8 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           IAccessAuthorizer.ACLType.CREATE, OzoneObj.ResourceType.KEY);
 
       // acquire lock
+      acquireVolumeLock = omMetadataManager.getLock().acquireWriteLock(
+          VOLUME_LOCK, volumeName);
       acquiredLock = omMetadataManager.getLock().acquireWriteLock(BUCKET_LOCK,
```
linyiqun, line 213: Nit: Can we rename all variable name of acquiredLock to acquireBucketLock?

linyiqun (commented):
Thanks for addressing the comments, @captainzmc ! Leave one minor comment.
@bharatviswa504 , does current fix way make sense to you?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
captainzmc: Thanks @linyiqun.  Changed this with new commit.

linyiqun (approved):
LGTM +1. 
Let's wait for others to have a look for latest change as well, : ).

bharatviswa504 (commented):
If we use volume lock for key operations we serialize all bucket ops across volume which have major perf impact.

So some general questions on Quota:
1. Can we skip updating volume bytes used during key operations, when it is required just calculated bytes used from all bucket info. As with volume lock, it affects key operations.
2. During bucket creation why can't we make check if buckets created exceed volume quota set if so, fail bucket creation. In this way, during key operations we don't need to check volume bytesUsed, just bucket bytesUsed will be enough.
3. Is there a way to disable quota feature? My question is so when upgraded during key creation we by default bytesUsed. And when not required, we don't need to update bytesUsed.

captainzmc:
Thanks for @bharatviswa504's optimization Suggestions.

> 3. Is there a way to disable quota feature? My question is so when upgraded during key creation we by default bytesUsed. And when not required, we don't need to update bytesUsed. (I see that we have check during checkQuota, but not during update bytesUsed)

This bytesUsed should not be removed. If quota is not enabled when a bucket is created, then quota is enabled after a certain amount of data is written. At this point we need to know how much bytesUsed was before quota was enabled so that we can correctly update it.
And in addition to quota, displaying bytesUsed on buckets also gives the user a more intuitive view of the current usage of bucket data. This is similar to bytesUsed in the current container.

> 1. Can we skip updating volume bytes used during key operations, when it is required just calculated bytes used from all bucket info. As with volume lock, it affects key operations.
> 2. During bucket creation why can't we make check if buckets created exceed volume quota set if so, fail bucket creation. In this way, during key operations we don't need to check volume bytesUsed, just bucket bytesUsed will be enough.

Agree, I think this is a better way to avoid the current use of Volume lock.

linyiqun:
Also +1 with no volume lock approach and aggregate the bucket bytes used instead. 

>This bytesUsed should be updated all the time. If quota is not enabled when a bucket is created, then quota is enabled after a certain amount of data is written. At this point we need to know how much bytesUsed was before quota was enabled so that we can correctly update it.

Agreed, bytes used still needed but don't need to do any quota check if we have a way to disable quota.

captainzmc:
Hi @bharatviswa504 @linyiqun. Based on what we discussed yesterday, I revised the PR. Can you take another look?  In addition, [all UT has been approved on my personal branch](https://github.com/captainzmc/hadoop-ozone/runs/1358813259).

runzhiwang:
We no longer need to update volume usedBytes in Response. You can also delete the logic for updating volume usedBytes in Response.

runzhiwang:
Now that we no longer need the usebyte of the volume; You need to synchronize the contents of quote.md.

runzhiwang:
Overall LGTM. Just a couple of minor suggestions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeysDeleteRequest.java
```diff
@@ -171,15 +171,15 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
         quotaReleased += sumBlockLengths(omKeyInfo);
       }
       // update usedBytes atomically.
```
runzhiwang, line 173: we do not need this comment any more.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmBucketInfo.java
```diff
@@ -241,9 +240,14 @@ public String getSourceBucket() {
   }
 
 
-  public LongAdder getUsedBytes() {
+  public long getUsedBytes() {
     return usedBytes;
   }
+
+  public void incrUsedBytes(long bytes) {
+    this.usedBytes = this.usedBytes + bytes;
```
runzhiwang, line 248: this.usedBytes += bytes;

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -167,7 +167,6 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
         }
       }
 
-
```
runzhiwang, line 170: revert this line

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequest.java
```diff
@@ -158,7 +158,8 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       omClientResponse = new OMKeyDeleteResponse(omResponse
           .setDeleteKeyResponse(DeleteKeyResponse.newBuilder()).build(),
-          omKeyInfo, ozoneManager.isRatisEnabled(), omVolumeArgs, omBucketInfo);
+          omKeyInfo, ozoneManager.isRatisEnabled(), omVolumeArgs,
+          copyBucketInfo);
```
runzhiwang, line 162: copyBucketInfo -> omBucketInfo.copyObject()

captainzmc:
Thanks for @runzhiwang‘s review. The issues has been fixed.

runzhiwang (approved):
LGTM.

runzhiwang:
@captainzmc Thanks the patch. @bharatviswa504 @linyiqun Thanks for review. I have merged the patch.

linyiqun:
@captainzmc  and @runzhiwang , I noticed that one also removed below function:
```java
  protected void checkVolumeQuotaInBytes(OmVolumeArgs omVolumeArgs,
      long allocateSize) throws IOException {
    ...
  }
```
So now where we will check the quota usage of the volume? There should be one place to do the sum of volume buckets usage and then do the quote check.

captainzmc:
> @captainzmc and @runzhiwang , I noticed that we also removed below function:
> 
> ```java
>   protected void checkVolumeQuotaInBytes(OmVolumeArgs omVolumeArgs,
>       long allocateSize) throws IOException {
>     ...
>   }
> ```
> 
> So now where we will check the quota usage of the volume? There should be one place to do the sum of volume buckets usage and then do the quota check.
Thanks for @linyiqun's attention. For now we no longer need to check the Quota of Volume. Because we have ensured that all bucket quota and do not exceed volume quota when we set bucket and volume quota. Therefore, to write a key under a bucket, we simply check bucket quota. Volume's quota will naturally not exceed as long as the bucket's quota checks pass.

linyiqun:
Sounds good, thanks @captainzmc  for the explanation.

PR #1490, opened 2020-10-13
HDDS-4269. Ozone DataNode thinks a volume is failed if an unexpected file is in the HDDS root directory.

flirmnave:

Adding a log makes it easy to track the problem.

https://issues.apache.org/jira/browse/HDDS-4269

## How was this patch tested?
Test on local.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/HddsVolumeUtil.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/HddsVolumeUtil.java
@@ -213,7 +213,9 @@ public final class HddsVolumeUtil {
       // The hdds root dir should always have 2 files. One is Version file
       // and other is SCM directory.
       logger.error("The hdds root dir {} should always have 2 files. " +
-              "One is Version file and other is SCM directory.",
+              "One is Version file and other is SCM directory. " +
+              "Please remove any other extra files from the directory " +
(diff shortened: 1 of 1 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/HddsVolumeUtil.java
```diff
@@ -212,6 +212,9 @@ public static boolean checkVolume(HddsVolume hddsVolume, String scmId, String
     } else {
       // The hdds root dir should always have 2 files. One is Version file
       // and other is SCM directory.
+      logger.error("The hdds root dir {} should always have 2 files. " +
+              "One is Version file and other is SCM directory.",
```
jojochuang, line 216: Let's add a more actionable message: "Please remove any other extra files from the directory so that DataNode startup can proceed."

flirmnave: Thanks @jojochuang  for reviewing and committing it.

jojochuang (approved):
Thanks! +1 pending precommit.

cxorm:
Thanks @flirmnave for the work, +1.
And thanks @jojochuang for the report and review.
I have committed it.

flirmnave:
Thanks @cxorm for reviewing and committing it, and @jojochuang for the review.

PR #1492, opened 2020-10-13
HDDS-4314. OM Layout Version Manager init throws silent CNF error in integration tests.

avijayanhwx:
Fix class scanner initialization such that it does not scan test packages.

https://issues.apache.org/jira/browse/HDDS-4314

## How was this patch tested?
Ran integration test without and with the fix.


swagle (approved):
+1 LGTM

PR #1493, opened 2020-10-13
HDDS-4327. Potential resource leakage using BatchOperation.

bharatviswa504:

Potential resource leakage using BatchOperation.

Use try enclosed resource/close the batch once after it usage is completed.

https://issues.apache.org/jira/browse/HDDS-4327

## How was this patch tested?

Existing tests.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
@@ -41,6 +41,7 @@ import org.apache.hadoop.hdds.scm.exceptions.SCMException;
 import org.apache.hadoop.hdds.scm.pipeline.Pipeline;
 import org.apache.hadoop.hdds.scm.pipeline.PipelineManager;
 import org.apache.hadoop.hdds.scm.pipeline.PipelineNotFoundException;
+import org.apache.hadoop.hdds.utils.db.BatchOperation;
 import org.apache.hadoop.hdds.utils.db.BatchOperationHandler;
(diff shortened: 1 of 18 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
```diff
@@ -397,33 +399,29 @@ public void updateDeleteTransactionId(Map<Long, Long> deleteTransactionMap)
...
       return;
     }
-    org.apache.hadoop.hdds.utils.db.BatchOperation batchOperation =
-        batchHandler.initBatchOperation();
-    lock.lock();
-    try {
-      for (Map.Entry<Long, Long> entry : deleteTransactionMap.entrySet()) {
+    try(org.apache.hadoop.hdds.utils.db.BatchOperation batchOperation =
+        batchHandler.initBatchOperation()) {
+      lock.lock();
```
hanishakoneru, line 404: Shouldn't lock() be called outside the try block? If there is a chance that the initBatchOperation() can fail, then lock.unlock() would throw IllegalMonitorStateException as the lock is not held by it.

hanishakoneru (commented):
Thanks @bharatviswa504 for fixing this. 
Have one question (posted inline). LGTM otherwise.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/SCMContainerManager.java
bharatviswa504: Good catch. Fixed in the latest commit.

hanishakoneru:
Thanks @bharatviswa504. +1 for merge.

PR #1494, opened 2020-10-13
HDDS-4330. Bootstrap new OM node

hanishakoneru:

In a ratis enabled OM cluster, add support to bootstrap a new OM node and add it to OM ratis ring. 

First step would be to update the ozone-site.xml with the configs (nodeId, address, ports etc.) for the new OM. 
The new node(s) should be started in BOOTSTRAP mode using the following command. This command will also initialize the OM. Hence, no need to run om init command before this command.
`ozone om --bootstrap`
When this command is run, it starts the OM in bootstrap mode. The new OM starts up its Ratis server and sends a request to existing OMs to perform a SetConfiguration operation. After the SetConfiguration operation is successful, the new OM is bootstrapped and part of the OM Ratis ring.

https://issues.apache.org/jira/browse/HDDS-4330

## How was this patch tested?

Added unit tests

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OMInterServiceProtocolPB.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OMInterServiceProtocolPB.java
@@ -34,7 +34,6 @@ import org.apache.hadoop.ozone.security.OzoneDelegationTokenSelector;
     protocolVersion = 1)
 @KerberosInfo(
     serverPrincipal = OMConfigKeys.OZONE_OM_KERBEROS_PRINCIPAL_KEY)
-@TokenInfo(OzoneDelegationTokenSelector.class)
 @InterfaceAudience.Private
 public interface OMInterServiceProtocolPB
     extends OzoneManagerInterService.BlockingInterface {
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OMInterServiceProtocolPB.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OMInterServiceProtocolPB.java
@@ -23,8 +23,6 @@ import org.apache.hadoop.ipc.ProtocolInfo;
 import org.apache.hadoop.ozone.protocol.proto
     .OzoneManagerInterServiceProtocolProtos.OzoneManagerInterService;
 import org.apache.hadoop.security.KerberosInfo;
-import org.apache.hadoop.security.token.TokenInfo;
-import org.apache.hadoop.ozone.security.OzoneDelegationTokenSelector;
 
 /**
  * Protocol used for communication between OMs.
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OMInterServiceProtocolPB.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OMInterServiceProtocolPB.java
@@ -1,0 +1,41 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.ozone.om.protocolPB;
+
+import org.apache.hadoop.hdds.annotation.InterfaceAudience;
+import org.apache.hadoop.ozone.om.OMConfigKeys;
+import org.apache.hadoop.ipc.ProtocolInfo;
+import org.apache.hadoop.ozone.protocol.proto
+    .OzoneManagerInterServiceProtocolProtos.OzoneManagerInterService;
+import org.apache.hadoop.security.KerberosInfo;
+import org.apache.hadoop.security.token.TokenInfo;
+import org.apache.hadoop.ozone.security.OzoneDelegationTokenSelector;
+
+/**
+ * Protocol used for communication between OMs.
+ */
+@ProtocolInfo(protocolName =
+    "org.apache.hadoop.ozone.om.protocol.OMInterServiceProtocol",
+    protocolVersion = 1)
+@KerberosInfo(
+    serverPrincipal = OMConfigKeys.OZONE_OM_KERBEROS_PRINCIPAL_KEY)
+@TokenInfo(OzoneDelegationTokenSelector.class)
+@InterfaceAudience.Private
+public interface OMInterServiceProtocolPB
+    extends OzoneManagerInterService.BlockingInterface {
+}
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
@@ -73,8 +73,8 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
   private int waitForClusterToBeReadyTimeout = 120000; // 2 min
 
   private static final Random RANDOM = new Random();
-  private static final int RATIS_RPC_TIMEOUT = 1000; // 10 second
-  public static final int NODE_FAILURE_TIMEOUT = 2000; // 20 seconds
+  private static final int RATIS_RPC_TIMEOUT = 1000; // 1 second
+  private static final int NODE_FAILURE_TIMEOUT = 2000; // 2 seconds
 
   /**
    * Creates a new MiniOzoneCluster.
@@ -365,7 +365,7 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
         numOfActiveOMs = numOfOMs;
       }
 
-      // If num of OMs it not set, set it to 1.
+      // If num of SCMs it not set, set it to 1.
       if (numOfSCMs == 0) {
         numOfSCMs = 1;
       }
@@ -703,12 +703,6 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
             omNodeId, basePort);
 
         om = bootstrapNewOM(omNodeId);
-
-        // Get the CertClient from an existing OM and set for new OM
-        if (omhaService.getServiceByIndex(0).getCertificateClient() != null) {
-          om.setCertClient(
-              omhaService.getServiceByIndex(0).getCertificateClient());
-        }
 
         LOG.info("Bootstrapped OzoneManager {} RPC server at {}", omNodeId,
             om.getOmRpcServerAddr());
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
@@ -68,11 +68,13 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
   private final OMHAService omhaService;
   private final SCMHAService scmhaService;
 
+  private final String clusterMetaPath;
+
   private int waitForClusterToBeReadyTimeout = 120000; // 2 min
 
   private static final Random RANDOM = new Random();
-  private static final int RATIS_RPC_TIMEOUT = 1000; // 1 second
-  public static final int NODE_FAILURE_TIMEOUT = 2000; // 2 seconds
+  private static final int RATIS_RPC_TIMEOUT = 1000; // 10 second
+  public static final int NODE_FAILURE_TIMEOUT = 2000; // 20 seconds
 
   /**
    * Creates a new MiniOzoneCluster.
@@ -89,12 +91,14 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
       List<HddsDatanodeService> hddsDatanodes,
       String omServiceId,
       String scmServiceId,
+      String clusterPath,
       ReconServer reconServer) {
     super(conf, hddsDatanodes, reconServer);
     omhaService =
         new OMHAService(activeOMList, inactiveOMList, omServiceId);
     scmhaService =
         new SCMHAService(activeSCMList, inactiveSCMList, scmServiceId);
+    this.clusterMetaPath = clusterPath;
   }
 
   /**
@@ -107,9 +111,10 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
       List<StorageContainerManager> scmList,
       List<HddsDatanodeService> hddsDatanodes,
       String omServiceId,
-      String scmServiceId) {
+      String scmServiceId,
+      String clusterPath) {
     this(conf, omList, null, scmList, null, hddsDatanodes,
-        omServiceId, scmServiceId, null);
+        omServiceId, scmServiceId, clusterPath, null);
   }
 
   @Override
@@ -181,7 +186,7 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
    */
   public OzoneManager getOMLeader() {
     OzoneManager res = null;
-    for (OzoneManager ozoneManager : this.omhaService.getServices()) {
+    for (OzoneManager ozoneManager : this.omhaService.getActiveServices()) {
       if (ozoneManager.isLeaderReady()) {
         if (res != null) {
           // Found more than one leader
@@ -239,7 +244,7 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
     LOG.info("Shutting down StorageContainerManager " + scm.getScmId());
 
     scm.stop();
-    scmhaService.removeInstance(scm);
+    scmhaService.deactivate(scm);
   }
 
   public void restartStorageContainerManager(
@@ -251,7 +256,7 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
     shutdownStorageContainerManager(scm);
     scm.join();
     scm = TestUtils.getScmSimple(scmConf);
-    scmhaService.addInstance(scm);
+    scmhaService.activate(scm);
     scm.start();
     if (waitForSCM) {
       waitForClusterToBeReady();
@@ -305,13 +310,17 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
   }
 
   public void stopOzoneManager(int index) {
-    omhaService.getServices().get(index).stop();
-    omhaService.getServices().get(index).join();
+    OzoneManager om = omhaService.getServices().get(index);
+    om.stop();
+    om.join();
+    omhaService.deactivate(om);
   }
 
   public void stopOzoneManager(String omNodeId) {
-    omhaService.getServiceById(omNodeId).stop();
-    omhaService.getServiceById(omNodeId).join();
+    OzoneManager om = omhaService.getServiceById(omNodeId);
+    om.stop();
+    om.join();
+    omhaService.deactivate(om);
   }
 
   /**
@@ -356,7 +365,12 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
         numOfActiveOMs = numOfOMs;
       }
 
-      // If num of ActiveOMs is not set, set it to numOfOMs.
+      // If num of OMs it not set, set it to 1.
+      if (numOfSCMs == 0) {
+        numOfSCMs = 1;
+      }
+
+      // If num of ActiveSCMs is not set, set it to numOfSCMs.
       if (numOfActiveSCMs == ACTIVE_SCMS_NOT_SET) {
         numOfActiveSCMs = numOfSCMs;
       }
@@ -383,7 +397,7 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
 
       MiniOzoneHAClusterImpl cluster = new MiniOzoneHAClusterImpl(conf,
           activeOMs, inactiveOMs, activeSCMs, inactiveSCMs,
-          hddsDatanodes, omServiceId, scmServiceId, reconServer);
+          hddsDatanodes, omServiceId, scmServiceId, path, reconServer);
 
       if (startDataNodes) {
         cluster.startHddsDatanodes();
@@ -430,7 +444,7 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
       List<OzoneManager> omList = Lists.newArrayList();
 
       int retryCount = 0;
-      int basePort = 10000;
+      int basePort;
 
       while (true) {
         try {
@@ -466,7 +480,7 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
             if (i <= numOfActiveOMs) {
               om.start();
               activeOMs.add(om);
-              LOG.info("Started OzoneManager RPC server at {}",
+              LOG.info("Started OzoneManager {} RPC server at {}", nodeId,
                   om.getOmRpcServerAddr());
             } else {
               inactiveOMs.add(om);
@@ -642,13 +656,14 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
       conf.set(OMConfigKeys.OZONE_OM_INTERNAL_SERVICE_ID, omServiceId);
       String omNodesKey = ConfUtils.addKeySuffixes(
           OMConfigKeys.OZONE_OM_NODES_KEY, omServiceId);
-      StringBuilder omNodesKeyValue = new StringBuilder();
+      List<String> omNodeIds = new ArrayList<>();
 
       int port = basePort;
 
       for (int i = 1; i <= numOfOMs; i++, port+=6) {
         String omNodeId = OM_NODE_ID_PREFIX + i;
-        omNodesKeyValue.append(",").append(omNodeId);
+        omNodeIds.add(omNodeId);
+
         String omAddrKey = ConfUtils.addKeySuffixes(
             OMConfigKeys.OZONE_OM_ADDRESS_KEY, omServiceId, omNodeId);
         String omHttpAddrKey = ConfUtils.addKeySuffixes(
@@ -664,8 +679,161 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
         conf.setInt(omRatisPortKey, port + 4);
       }
 
-      conf.set(omNodesKey, omNodesKeyValue.substring(1));
+      conf.set(omNodesKey, String.join(",", omNodeIds));
     }
+  }
+
+  /**
+   * Bootstrap new OM and add to existing OM HA service ring.
+   * @return new OM nodeId
+   */
+  public void bootstrapOzoneManager(String omNodeId) throws Exception {
+
+    int basePort;
+    int retryCount = 0;
+
+    OzoneManager om = null;
+
+    long leaderSnapshotIndex = getOMLeader().getRatisSnapshotIndex();
+
+    while (true) {
+      try {
+        basePort = 10000 + RANDOM.nextInt(1000) * 4;
+        OzoneConfiguration newConf = addNewOMToConfig(getOMServiceId(),
+            omNodeId, basePort);
+
+        om = bootstrapNewOM(omNodeId);
+
+        // Get the CertClient from an existing OM and set for new OM
+        if (omhaService.getServiceByIndex(0).getCertificateClient() != null) {
+          om.setCertClient(
+              omhaService.getServiceByIndex(0).getCertificateClient());
+        }
+
+        LOG.info("Bootstrapped OzoneManager {} RPC server at {}", omNodeId,
+            om.getOmRpcServerAddr());
+
+        // Add new OMs to cluster's in memory map and update existing OMs conf.
+        setConf(newConf);
+
+        omhaService.addInstance(om, true);
+        break;
+      } catch (IOException e) {
+        // Existing OM config could have been updated with new conf. Reset it.
+        for (OzoneManager existingOM : omhaService.getServices()) {
+          existingOM.setConfiguration(getConf());
+        }
+        if (e instanceof BindException ||
+            e.getCause() instanceof BindException) {
+          if (om != null) {
+            om.stop();
+            om.join();
+            LOG.info("Stopping OzoneManager server at {}",
+                om.getOmRpcServerAddr());
+          }
+          ++retryCount;
+          LOG.info("MiniOzoneHACluster port conflicts, retried {} times",
+              retryCount);
+        } else {
+          throw e;
+        }
+      }
+    }
+
+    waitForBootstrappedNodeToBeReady(om, leaderSnapshotIndex);
+    waitForConfigUpdateOnAllOMs(omNodeId);
+  }
+
+  /**
+   * Set the configs for new OMs.
+   */
+  private OzoneConfiguration addNewOMToConfig(String omServiceId,
+      String omNodeId, int basePort) {
+    OzoneConfiguration newConf = getConf();
+    String omNodesKey = ConfUtils.addKeySuffixes(
+        OMConfigKeys.OZONE_OM_NODES_KEY, omServiceId);
+    StringBuilder omNodesKeyValue = new StringBuilder();
+    omNodesKeyValue.append(newConf.get(omNodesKey))
+        .append(",").append(omNodeId);
+
+    String omAddrKey = ConfUtils.addKeySuffixes(
+        OMConfigKeys.OZONE_OM_ADDRESS_KEY, omServiceId, omNodeId);
+    String omHttpAddrKey = ConfUtils.addKeySuffixes(
+        OMConfigKeys.OZONE_OM_HTTP_ADDRESS_KEY, omServiceId, omNodeId);
+    String omHttpsAddrKey = ConfUtils.addKeySuffixes(
+        OMConfigKeys.OZONE_OM_HTTPS_ADDRESS_KEY, omServiceId, omNodeId);
+    String omRatisPortKey = ConfUtils.addKeySuffixes(
+        OMConfigKeys.OZONE_OM_RATIS_PORT_KEY, omServiceId, omNodeId);
+
+    newConf.set(omAddrKey, "127.0.0.1:" + basePort);
+    newConf.set(omHttpAddrKey, "127.0.0.1:" + (basePort + 2));
+    newConf.set(omHttpsAddrKey, "127.0.0.1:" + (basePort + 3));
+    newConf.setInt(omRatisPortKey, basePort + 4);
+
+    newConf.set(omNodesKey, omNodesKeyValue.toString());
+
+    return newConf;
+  }
+
+  /**
+   * Start a new OM in Bootstrap mode. Configs for the new OM must already be
+   * set.
+   */
+  private OzoneManager bootstrapNewOM(String nodeId)
+      throws IOException, AuthenticationException {
+    OzoneConfiguration config = new OzoneConfiguration(getConf());
+    config.set(OMConfigKeys.OZONE_OM_NODE_ID_KEY, nodeId);
+    // Set the OM rpc and http(s) address to null so that the cluster picks
+    // up the address set with service ID and node ID
+    config.set(OMConfigKeys.OZONE_OM_ADDRESS_KEY, "");
+    config.set(OMConfigKeys.OZONE_OM_HTTP_ADDRESS_KEY, "");
+    config.set(OMConfigKeys.OZONE_OM_HTTPS_ADDRESS_KEY, "");
+
+    // Set metadata/DB dir base path
+    String metaDirPath = clusterMetaPath + "/" + nodeId;
+    config.set(OZONE_METADATA_DIRS, metaDirPath);
+
+    // Update existing OMs config
+    for (OzoneManager existingOM : omhaService.getServices()) {
+      existingOM.setConfiguration(config);
+    }
+
+    OzoneManager.omInit(config);
+    OzoneManager om = OzoneManager.createOm(config,
+        OzoneManager.StartupOption.BOOTSTRAP);
+    om.start();
+    return om;
+
+  }
+
+  /**
+   * Wait for AddOM command to execute on all OMs.
+   */
+  private void waitForBootstrappedNodeToBeReady(OzoneManager newOM,
+      long leaderSnapshotIndex) throws Exception {
+    // Wait for bootstrapped nodes to catch up with others
+    GenericTestUtils.waitFor(() -> {
+      try {
+        if (newOM.getRatisSnapshotIndex() >= leaderSnapshotIndex) {
+          return true;
+        }
+      } catch (IOException e) {
+        return false;
+      }
+      return false;
+    }, 1000, waitForClusterToBeReadyTimeout);
+  }
+
+  private void waitForConfigUpdateOnAllOMs(String newOMNodeId)
+      throws Exception {
+    GenericTestUtils.waitFor(() -> {
+      for (OzoneManager om : omhaService.getActiveServices()) {
+        if (!om.doesPeerExist(newOMNodeId)) {
+          return false;
+        }
+      }
+      return true;
+    }, 1000, waitForClusterToBeReadyTimeout);
   }
 
   /**
@@ -683,11 +851,15 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
     private List<Type> activeServices;
     private List<Type> inactiveServices;
 
+    // Function to extract the Id from service
+    Function<Type, String> serviceIdProvider;
+
     MiniOzoneHAService(String name, List<Type> activeList,
                        List<Type> inactiveList, String serviceId,
                        Function<Type, String> idProvider) {
       this.serviceName = name;
       this.serviceMap = Maps.newHashMap();
+      this.serviceIdProvider = idProvider;
       if (activeList != null) {
         for (Type service : activeList) {
           this.serviceMap.put(idProvider.apply(service), service);
@@ -717,12 +889,30 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
       return services;
     }
 
+    public List<Type> getActiveServices() {
+      return activeServices;
+    }
+
     public boolean removeInstance(Type t) {
       return services.remove(t);
     }
 
-    public boolean addInstance(Type t) {
-      return services.add(t);
+    public void addInstance(Type t, boolean isActive) {
+      services.add(t);
+      serviceMap.put(serviceIdProvider.apply(t), t);
+      if (isActive) {
+        activeServices.add(t);
+      }
+    }
+
+    public void activate(Type t) {
+      activeServices.add(t);
+      inactiveServices.remove(t);
+    }
+
+    public void deactivate(Type t) {
+      activeServices.remove(t);
+      inactiveServices.add(t);
     }
 
     public boolean isServiceActive(String id) {
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
@@ -34,6 +34,7 @@ import org.apache.hadoop.ozone.client.OzoneClientFactory;
 import org.apache.hadoop.ozone.ha.ConfUtils;
 import org.apache.hadoop.ozone.om.OMConfigKeys;
 import org.apache.hadoop.ozone.om.OzoneManager;
+import org.apache.hadoop.ozone.om.ratis.OzoneManagerRatisServer;
 import org.apache.hadoop.ozone.recon.ReconServer;
 import org.apache.hadoop.security.authentication.client.AuthenticationException;
 import org.apache.ozone.test.GenericTestUtils;
@@ -814,9 +815,17 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
 
   private void waitForConfigUpdateOnAllOMs(String newOMNodeId)
       throws Exception {
+    OzoneManagerRatisServer newOMRatisServer =
+        omhaService.getServiceById(newOMNodeId).getOmRatisServer();
     GenericTestUtils.waitFor(() -> {
+      // Each existing active OM should contain the new OM in its peerList.
+      // Also, the new OM should contain each existing active OM in it's
+      // RatisServer peerList.
       for (OzoneManager om : omhaService.getActiveServices()) {
         if (!om.doesPeerExist(newOMNodeId)) {
+          return false;
+        }
+        if (!newOMRatisServer.doesPeerExist(om.getOMNodeId())) {
           return false;
         }
       }
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
@@ -725,12 +725,6 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
         }
         if (e instanceof BindException ||
             e.getCause() instanceof BindException) {
-          if (om != null) {
-            om.stop();
-            om.join();
-            LOG.info("Stopping OzoneManager server at {}",
-                om.getOmRpcServerAddr());
-          }
           ++retryCount;
           LOG.info("MiniOzoneHACluster port conflicts, retried {} times",
               retryCount);
@@ -852,7 +846,7 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
     private List<Type> inactiveServices;
 
     // Function to extract the Id from service
-    Function<Type, String> serviceIdProvider;
+    private Function<Type, String> serviceIdProvider;
 
     MiniOzoneHAService(String name, List<Type> activeList,
                        List<Type> inactiveList, String serviceId,
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
@@ -815,14 +815,17 @@ public class MiniOzoneHAClusterImpl extends MiniOzoneClusterImpl {
 
   private void waitForConfigUpdateOnAllOMs(String newOMNodeId)
       throws Exception {
-    OzoneManagerRatisServer newOMRatisServer =
-        omhaService.getServiceById(newOMNodeId).getOmRatisServer();
+    OzoneManager newOMNode = omhaService.getServiceById(newOMNodeId);
+    OzoneManagerRatisServer newOMRatisServer = newOMNode.getOmRatisServer();
     GenericTestUtils.waitFor(() -> {
       // Each existing active OM should contain the new OM in its peerList.
-      // Also, the new OM should contain each existing active OM in it's
-      // RatisServer peerList.
+      // Also, the new OM should contain each existing active OM in it's OM
+      // peer list and RatisServer peerList.
       for (OzoneManager om : omhaService.getActiveServices()) {
         if (!om.doesPeerExist(newOMNodeId)) {
+          return false;
+        }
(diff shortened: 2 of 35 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OzoneIllegalArgumentException.java
```diff
@@ -37,4 +37,8 @@
   public OzoneIllegalArgumentException(final String message) {
```
swagle, line 37: General question why are specializing IllegalArgumentException if not other state/behavior added to it :)

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
```diff
@@ -442,7 +457,169 @@ private void initHAConfig(int basePort) throws IOException {
...
+      conf.set(omNodesKey, String.join(",", omNodeIds));
+    }
+  }
+
+  /**
+   * Bootstrap new OMs and add them to existing OM ring.
+   * @param numNewOMs number of new OMs to bootstrap
+   * @return list of new OMs nodeIds
+   */
+  public List<String> bootstrapOzoneManagers(int numNewOMs) throws IOException,
```
swagle, line 469: Why is the number of newOMs an argument? I was expecting this to read based on the new HA configuration

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -137,6 +141,59 @@ public OMResponse submitRequest(OMRequest omRequest) throws ServiceException {
...
+      InetSocketAddress newOMRatisAddr = new InetSocketAddress(
+          newOMNode.getHostName(), newOMNode.getRatisPort());
+
+      newRaftPeerList.add(RaftPeer.newBuilder()
+          .setId(newOMRaftPeerId)
+          .setAddress(newOMRatisAddr)
+          .build());
+    }
+
+    if (LOG.isDebugEnabled()) {
```
fapifta, line 171: nit: old log4j habit, we should not need this check as parameters are evaluated only if debug is enabled with this format. same is true for line 384

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -164,6 +167,27 @@ public void notifyTermIndexUpdated(long currentTerm, long index) {
...
+      RaftConfigurationProto newRaftConfiguration) {
+    List<RaftPeerProto> newPeers = newRaftConfiguration.getPeersList();
+    LOG.info("Received Configuration change notification from Ratis. New Peer" +
+        " list:\n{}", newPeers);
+    for (RaftProtos.RaftPeerProto raftPeerProto : newPeers) {
+      String omNodeId = RaftPeerId.valueOf(raftPeerProto.getId()).toString();
+      if (!ozoneManager.doesPeerExist(omNodeId)) {
+        LOG.info("Adding new OM {} to the cluster.", omNodeId);
+        ozoneManager.addNewOMNode(omNodeId);
+      } else if(LOG.isDebugEnabled()) {
```
fapifta, line 184: nit: again old log4j habit

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/AddOMSubcommand.java
```diff
@@ -0,0 +1,174 @@
...
+
+  @Override
+  public Void call() throws Exception {
+    boolean success = false;
+    try {
+      ozoneManagerClient =  parent.createOmClient(omServiceId);
+
+      List<OMNodeInfo> newOMNodeInfos = getNewNodeInfos();
+
+      if (!newOMNodeInfos.isEmpty()) {
```
bharatviswa504, line 88: Here first we need to check for not null, as in few code paths getNewNodeInfos returns null.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -1203,11 +1203,41 @@ message UpdateGetS3SecretRequest {
...
+    optional string serviceId = 7;
+}
+
+enum BootstrapErrorCode {
+    RATIS_NOT_ENABLED = 1;
+    LEADER_UNDETERMINED = 2;
+    RATIS_BOOTSTRAP_ERROR = 3;
+}
+
+message BootstrapRequest {
```
bharatviswa504, line 1222: Can we call this AddOMRequest, as this is an admin command submitted to add new OM's to the cluster.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -1203,11 +1203,41 @@ message UpdateGetS3SecretRequest {
     required string awsSecret = 2;
 }
 
+message OMNodeInfo {
+    required string nodeId = 1;
+    required string hostAddress = 2;
+    required uint32 rpcPort = 3;
```
bharatviswa504, line 1209: We are not using rpcPort, httpAddr, httpsAddr, serviceId these can be removed from this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -388,6 +412,15 @@ public String toStateMachineLogEntryString(StateMachineLogEntryProto proto) {
     return OMRatisHelper.smProtoToString(proto);
   }
 
+  @Override
+  public void close() throws IOException {
+    // OM should be shutdown as the StateMachine has shutdown.
```
bharatviswa504, line 417: Question:
I see this is called from Ratis, when any StateMachine has hit exception.
In this case, do we want to terminate OM, is this handled to handle any failure of notifyConfigurationChanged where we are throwing OzoneIllegalArgumentException?

bharatviswa504 (commented):
Can we also open a new doc jIra to document details to how to convert single node ratis to 3 node ratis
And also how we can add new nodes to Ratis ring.

Because for a single node to 3 node, end user should use
serviceid as omServiceIdDefault
And nodeID as storage version file omID

bharatviswa504:
Thank You @hanishakoneru for the PR.
PR also needs a rebase.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
hanishakoneru: Giving numNewOMs are parameter to test different scenarios - adding OMs one by one and adding two OMs together. This is the function which generates the new HA configuration for the new nodes.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -2707,6 +2728,57 @@ public ServiceInfoEx getServiceInfo() throws IOException {
...
+
+  /**
+   * Add a new OM Node to the HA cluster.
+   */
+  public void addNewOMNode(String newOMNodeId) {
+    OMNodeDetails newOMNodeDetails = OMNodeDetails.getOMNodeDetailsFromConf(
+        getConfiguration(), getOMServiceId(), newOMNodeId);
+    if (newOMNodeDetails == null) {
+      // Load new configuration object to read in new peer information
+      setConfiguration(new OzoneConfiguration());
```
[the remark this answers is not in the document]
hanishakoneru: This is run on all OMs (old OMs also). OzoneManager#addNewOMNode() is called when Ratis sends notifyConfigurationChanged(). And since the old OMs are not necessarily restarted, we need to reload the configuration.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
hanishakoneru: If the StateMachine has terminated for any reason, then the OM should also be terminated. 
This is not just for notifyConfigurationChanged exception. If the RaftServer shutsdown for reason, we want the OM also to shutdown. Though the OM would not service write requests any more, the shutdown would not be visible to user as OM is still running.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OzoneIllegalArgumentException.java
hanishakoneru: Good point. Removed this from current patch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
hanishakoneru: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
hanishakoneru: Done

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OmTransport.java
```diff
@@ -33,6 +35,9 @@
    */
   OMResponse submitRequest(OMRequest payload) throws IOException;
 
+  BootstrapResponse bootstrapRequest(BootstrapRequest bootstrapRequest)
```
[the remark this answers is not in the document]
hanishakoneru: Done

hadoop-ozone/ozonefs-hadoop2/src/main/java/org/apache/hadoop/fs/ozone/Hadoop27RpcTransport.java
```diff
@@ -71,6 +73,16 @@ public OMResponse submitRequest(OMRequest payload) throws IOException {
     }
   }
 
+  @Override
+  public BootstrapResponse bootstrapRequest(BootstrapRequest bootstrapRequest)
+      throws IOException {
+    try {
+      return proxy.bootstrap(NULL_RPC_CONTROLLER, bootstrapRequest);
+    } catch (ServiceException e) {
+      throw new IOException("Service exception during the bootstrap", e);
```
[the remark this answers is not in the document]
hanishakoneru: Yes, Hadoop27OMTransport does not support it. Replicated the behavior as is in submitRequest.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -192,6 +195,51 @@ public void close() throws IOException {
...
+
+    return new ServiceInfoEx(
+        resp.getServiceInfoList().stream()
+            .map(ServiceInfo::getFromProtobuf)
+            .collect(Collectors.toList()),
+        resp.getCaCertificate());
+  }
+
+  @Override
+  public boolean bootstrap(List<OMNodeInfo> omNodeInfos) throws IOException {
```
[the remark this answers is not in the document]
hanishakoneru: Done. Using OMNodeDetails instead.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -137,6 +141,59 @@ public OMResponse submitRequest(OMRequest omRequest) throws ServiceException {
...
+   * Add a new OM to the Ratis ring.
+   * @return true if the new OM is successfully added to the ratis ring,
+   * false otherwise.
+   */
+  public boolean bootstrapNewOMs(List<OMNodeDetails> newOMNodes) {
+    StringBuilder newOMNodeIdsBuilder = new StringBuilder();
+    newOMNodes.stream().forEach(newOMNode ->
+        newOMNodeIdsBuilder.append(", ").append(newOMNode.getOMNodeId()));
+    String newOMNodeIds = newOMNodeIdsBuilder.toString().substring(2);
+    LOG.info("Bootstrapping new OM(s): {}", newOMNodeIds);
```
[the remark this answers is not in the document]
hanishakoneru: Done. Not propagating the unused data anymore.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -317,8 +377,18 @@ public static OzoneManagerRatisServer newOMRatisServer(
         .build();
 
     List<RaftPeer> raftPeers = new ArrayList<>();
-    // Add this Ratis server to the Ratis ring
-    raftPeers.add(localRaftPeer);
+
+    // If the OM is started in bootstrap mode, do not add it to the ratis ring.
+    // It will be added later using SetConfiguration from the leader OM.
+    if (isBootstrapping) {
```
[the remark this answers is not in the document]
hanishakoneru: It could possibly cause a different ring to form with the new OMs.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/AddOMSubcommand.java
hanishakoneru: Done.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
hanishakoneru: Done.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
hanishakoneru: Done.

hanishakoneru:
Thanks @swagle, @fapifta and @bharatviswa504 for the reviews. I have addressed the review comments in the new commit (or answered inline).

bshashikant:
@hanishakoneru , ca you please rebase??

ChenSammi:
Thanks @hanishakoneru for working on this.  We have a plan to enable OM HA in our production cluster.  This feature is definitely we need.

elek:
@hanishakoneru Are you planning to update this?

hanishakoneru:
> @hanishakoneru Are you planning to update this?

@elek , yes I am working on it currently. Will upload a new patch in a day or two.

hanishakoneru:
Updated the patch to remove the extra "addOM" command. Now when a new OM bootstraps, it sends an RPC request to existing OM ring to add itself to the ring.

hanishakoneru:
@bharatviswa504 can you please take a look when you get a chance. Thanks.

bshashikant:
@hanishakoneru , can you rebase this? Sorry for getting late to this..

@bharatviswa504 , can you help reviewing this as well?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerBootstrap.java
```diff
@@ -0,0 +1,212 @@
...
+        .getObjectStore();
+
+    // Perform some transactions
+    objectStore.createVolume(VOLUME_NAME);
+    OzoneVolume volume = objectStore.getVolume(VOLUME_NAME);
+    volume.createBucket(BUCKET_NAME);
+    OzoneBucket bucket = volume.getBucket(BUCKET_NAME);
+    createKey(bucket);
+
+    lastTransactionIndex = cluster.getOMLeader().getRatisSnapshotIndex();
```
bharatviswa504, line 102: We commit to DB once double buffer flushes. Do you think we need some wait or read from StateMachine and getLastAppliedIndex?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
```diff
@@ -652,10 +667,158 @@ private void initOMHAConfig(int basePort) throws IOException {
...
+      try {
+        basePort = 10000 + RANDOM.nextInt(1000) * 4;
+        OzoneConfiguration newConf = addNewOMToConfig(getOMServiceId(),
+            omNodeId, basePort);
+
+        om = bootstrapNewOM(omNodeId);
+
+        // Get the CertClient from an existing OM and set for new OM
+        if (omhaService.getServiceByIndex(0).getCertificateClient() != null) {
+          om.setCertClient(
```
bharatviswa504, line 697: Question:
Not understood the reason to do this.
As our MiniOzoneHACluster is only used in non-secure, why certClient is being set here?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerBootstrap.java
```diff
@@ -0,0 +1,212 @@
...
+    GenericTestUtils.waitFor(new Supplier<Boolean>() {
+      @Override
+      public Boolean get() {
+        try {
+          return newOM.getRatisSnapshotIndex() >= lastTransactionIndex;
+        } catch (IOException e) {
+          return false;
+        }
+      }
+    }, 100, 10000);
```
bharatviswa504, line 127: Increase time duration here to avoid flakiness in CI runs.

And also as method name says check OM Exists in peerList, can we check from ratis server peer info also?

bharatviswa504 (commented):
Overall LGTM.
Few comments/questions inline.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
```diff
@@ -65,11 +65,13 @@
...
   private final SCMHAService scmhaService;
 
+  private final String clusterMetaPath;
+
   private int waitForClusterToBeReadyTimeout = 120000; // 2 min
 
   private static final Random RANDOM = new Random();
-  private static final int RATIS_RPC_TIMEOUT = 1000; // 1 second
-  public static final int NODE_FAILURE_TIMEOUT = 2000; // 2 seconds
+  private static final int RATIS_RPC_TIMEOUT = 1000; // 10 second
```
bharatviswa504, line 73: Question: Here comments changed, but actual values not changed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -166,6 +167,29 @@ public void notifyTermIndexUpdated(long currentTerm, long index) {
...
+      RaftProtos.RaftConfigurationProto newRaftConfiguration) {
+    List<RaftProtos.RaftPeerProto> newPeers =
+        newRaftConfiguration.getPeersList();
+    LOG.info("Received Configuration change notification from Ratis. New Peer" +
+        " list:\n{}", newPeers);
+    for (RaftProtos.RaftPeerProto raftPeerProto : newPeers) {
+      String omNodeId = RaftPeerId.valueOf(raftPeerProto.getId()).toString();
+      if (!ozoneManager.doesPeerExist(omNodeId)) {
+        LOG.info("Adding new OM {} to the Peer list.", omNodeId);
+        ozoneManager.addOMNodeToPeers(omNodeId);
```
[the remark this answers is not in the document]
bharatviswa504: Looks like this is taken care of. With this       // Load new configuration object to read in new peer information
      setConfiguration(new OzoneConfiguration());

So, here it reloads config from disk again? Have you tried on a docker/cluster to verify this. (From reading code looks like config load has taken care) Just curious.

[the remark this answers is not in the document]
hanishakoneru: Yes, tried it out on a docker cluster and the config is reloaded from disk.
> The only thing is when they add a new node, the users should change the config on all nodes.
Yes that is a requirement to ensure that if the cluster is restarted, the old nodes have information about the new node.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerBootstrap.java
hanishakoneru: Done

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerBootstrap.java
hanishakoneru: Done. Getting SM's last applied index.

hanishakoneru:
Thanks @bharatviswa504 for the review. I have addressed your comments.

hanishakoneru:
@xiaoyuyao can you please take a look from security perspective? I have added KerberosInfo to the new Protocol classes. 
But how can we ensure that the request is coming from another OM?

vivekratnavel:
@hanishakoneru I believe we cannot restrict access until we set the "ozone.om.security.admin.protocol.acl" to authorized user principals such as OM. In production clusters, "hadoop.security.authorization" should be set to "true" and "ozone.om.security.admin.protocol.acl" should have a list of all the admin users.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -671,6 +672,21 @@ public RaftServerStatus checkLeaderStatus() {
...
+  public List<String> getCurrentPeersFromRaftConf() {
+    try {
+      Collection<RaftPeer> currentPeers =
+          server.getDivision(raftGroupId).getRaftConf().getCurrentPeers();
+      List<String> currentPeerList = new ArrayList<>();
+      currentPeers.forEach(e -> currentPeerList.add(e.getId().toString()));
+      return currentPeerList;
+    } catch (IOException e) {
+      // In this case we return not a leader.
+      LOG.error("Failed to get RaftServer information. ", e);
```
bharatviswa504, line 685: Question: Why in exception case logging and return null, can't we throw an exception?
And also can we add JavaDoc for this?

bharatviswa504 (commented):
Thank You @hanishakoneru for the update.
I have resolved all comments which are fixed, except not taken care of. PTAL.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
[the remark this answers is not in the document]
bharatviswa504: We can revisit this if we want to support with out change in config.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -166,6 +167,29 @@ public void notifyTermIndexUpdated(long currentTerm, long index) {
...
 
+  /**
+   * Called to notify state machine about configuration changes.
+   * Configurations changes include addition of newly bootstrapped OM.
+   */
+  @Override
+  public void notifyConfigurationChanged(long term, long index,
+      RaftProtos.RaftConfigurationProto newRaftConfiguration) {
+    List<RaftProtos.RaftPeerProto> newPeers =
+        newRaftConfiguration.getPeersList();
```
bharatviswa504, line 178: Looks like we need to call computeAndUpdateLastAppliedIndex. As applyLogToStateMachine in StateMachineUpdater when configEntry, we call 
      stateMachine.event().notifyConfigurationChanged(next.getTerm(), next.getIndex(), next.getConfigurationEntry());

As in OM StateMachine, we update lastAppliedIndex using computeAndUpdateLastAppliedIndex.

bharatviswa504: Sorry reading more looks we should be fine, as for any no state machine log entry we call notifyTermIndexUpdated.

    if (!next.hasStateMachineLogEntry()) {
      stateMachine.event().notifyTermIndexUpdated(next.getTerm(), next.getIndex());
    }

Can you also confirm once?

hanishakoneru: Correct. If the log entry is a configurationEntry or a metadataEntry, then RaftServerImpl calls notifyTermIndexUpdated on StateMachine.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
hanishakoneru: Done.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OMInterServiceProtocolPB.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Protocol used for communication between OMs.
+ */
+@ProtocolInfo(protocolName =
+    "org.apache.hadoop.ozone.om.protocol.OMInterServiceProtocol",
+    protocolVersion = 1)
+@KerberosInfo(
+    serverPrincipal = OMConfigKeys.OZONE_OM_KERBEROS_PRINCIPAL_KEY)
+@TokenInfo(OzoneDelegationTokenSelector.class)
```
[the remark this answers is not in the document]
hanishakoneru: I am not sure. 
@xiaoyuyao can you please help with this?

hanishakoneru:
@bharatviswa504, I updated the patch. Can you please take another look.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OMInterServiceProtocolPB.java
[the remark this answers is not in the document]
hanishakoneru: Thank you @xiaoyuyao for the offline discussion. 
@bharatviswa504, you are right. We do not need to add TokenInfo for this protocol. Since this is inter OM service, OMs can use their Kerberos principal to authenticate. We do not need to provide Token authorization also.

bharatviswa504:
@hanishakoneru 
I think this PR has an issue or not sure any more steps needed to test it on the cluster.
I have updated ozone-ha to use om2,om3 start use bootstrap it does not succeed, because when OM1 starts it adds OM2,OM3 as its peers, and there will be no leader in the cluster.


To try this scenario, you can update ozone-ha docker-compose.yaml to use below change

 ```
 om2:
    <<: *common-config
    environment:
      WAITFOR: scm3:9894
      <<: *replication
    ports:
      - 9874
      - 9862
    hostname: om2
    command: ["ozone","om", "--bootstrap"]
  om3:
    <<: *common-config
    environment:
      WAITFOR: scm3:9894
      <<: *replication
    ports:
      - 9874
      - 9862
    hostname: om3
    command: ["ozone","om", "--bootstrap"]
```

**Error snippet:**

**om1 leader election failing**
```
om1_1       | 2021-06-11 06:49:00,434 [om1@group-D66704EFC61C-LeaderElection63] INFO impl.LeaderElection: om1@group-D66704EFC61C-LeaderElection63: ELECTION REJECTED received 0 response(s) and 2 exception(s):
om1_1       | 2021-06-11 06:49:00,435 [om1@group-D66704EFC61C-LeaderElection63] INFO impl.LeaderElection:   Exception 0: java.util.concurrent.ExecutionException: org.apache.ratis.thirdparty.io.grpc.StatusRuntimeException: INTERNAL: om3@group-D66704EFC61C is not in [RUNNING]: current state is STARTING
om1_1       | 2021-06-11 06:49:00,435 [om1@group-D66704EFC61C-LeaderElection63] INFO impl.LeaderElection:   Exception 1: java.util.concurrent.ExecutionException: org.apache.ratis.thirdparty.io.grpc.StatusRuntimeException: INTERNAL: om2@group-D66704EFC61C is not in [RUNNING]: current state is STARTING
om1_1       | 2021-06-11 06:49:00,435 [om1@group-D66704EFC61C-LeaderElection63] INFO impl.LeaderElection: om1@group-D66704EFC61C-LeaderElection63 ELECTION round 0: result REJECTED
Similar on om3.
```

```
om2_1       | 2021-06-11 06:49:02,604 [Listener at om2/9862] INFO retry.RetryInvocationHandler: com.google.protobuf.ServiceException: org.apache.hadoop.ipc.RemoteException(org.apache.hadoop.ozone.om.exceptions.OMNotLeaderException): OM:om1 is not the leader. Could not determine the leader node.
om2_1       | 	at org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerRatisUtils.createNotLeaderException(OzoneManagerRatisUtils.java:408)
om2_1       | 	at org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerRatisUtils.checkLeaderStatus(OzoneManagerRatisUtils.java:395)
om2_1       | 	at org.apache.hadoop.ozone.protocolPB.OMInterServiceProtocolServerSideImpl.checkLeaderStatus(OMInterServiceProtocolServerSideImpl.java:87)
om2_1       | 	at org.apache.hadoop.ozone.protocolPB.OMInterServiceProtocolServerSideImpl.bootstrap(OMInterServiceProtocolServerSideImpl.java:63)
om2_1       | 	at org.apache.hadoop.ozone.protocol.proto.OzoneManagerInterServiceProtocolProtos$OzoneManagerInterService$2.callBlockingMethod(OzoneManagerInterServiceProtocolProtos.java:1675)
om2_1       | 	at org.apache.hadoop.ipc.ProtobufRpcEngine$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine.java:528)
om2_1       | 	at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1086)
om2_1       | 	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1029)
om2_1       | 	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:957)
om2_1       | 	at java.base/java.security.AccessController.doPrivileged(Native Method)
om2_1       | 	at java.base/javax.security.auth.Subject.doAs(Subject.java:423)
om2_1       | 	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1762)
om2_1       | 	at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2957)
om2_1       | , while invoking $Proxy44.bootstrap over nodeId=om1,nodeAddress=om1:9862 after 486 failover attempts. Trying to failover immediately.
om2_1       | 2021-06-11 06:49:02,606 [Listener at om2/9862] INFO retry.RetryInvocationHandler: com.google.protobuf.ServiceException: org.apache.hadoop.ipc.RemoteException(org.apache.hadoop.ozone.om.exceptions.OMNotLeaderException): OM:om2 is not the leader. Could not determine the leader node.
om2_1       | 	at org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerRatisUtils.createNotLeaderException(OzoneManagerRatisUtils.java:408)
om2_1       | 	at org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerRatisUtils.checkLeaderStatus(OzoneManagerRatisUtils.java:395)
om2_1       | 	at org.apache.hadoop.ozone.protocolPB.OMInterServiceProtocolServerSideImpl.checkLeaderStatus(OMInterServiceProtocolServerSideImpl.java:87)
om2_1       | 	at org.apache.hadoop.ozone.protocolPB.OMInterServiceProtocolServerSideImpl.bootstrap(OMInterServiceProtocolServerSideImpl.java:63)
om2_1       | 	at org.apache.hadoop.ozone.protocol.proto.OzoneManagerInterServiceProtocolProtos$OzoneManagerInterService$2.callBlockingMethod(OzoneManagerInterServiceProtocolProtos.java:1675)
om2_1       | 	at org.apache.hadoop.ipc.ProtobufRpcEngine$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine.java:528)
om2_1       | 	at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1086)
om2_1       | 	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1029)
om2_1       | 	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:957)
om2_1       | 	at java.base/java.security.AccessController.doPrivileged(Native Method)
om2_1       | 	at java.base/javax.security.auth.Subject.doAs(Subject.java:423)
om2_1       | 	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1762)
om2_1       | 	at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2957)
om2_1       | , while invoking $Proxy44.bootstrap over nodeId=om2,nodeAddress=om2:9862 after 487 failover attempts. Trying to failover immediately.
om2_1       | 2021-06-11 06:49:02,608 [Listener at om2/9862] INFO retry.RetryInvocationHandler: com.google.protobuf.ServiceException: org.apache.hadoop.ipc.RemoteException(org.apache.hadoop.ozone.om.exceptions.OMNotLeaderException): OM:om3 is not the leader. Could not determine the leader node.
om2_1       | 	at org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerRatisUtils.createNotLeaderException(OzoneManagerRatisUtils.java:408)
om2_1       | 	at org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerRatisUtils.checkLeaderStatus(OzoneManagerRatisUtils.java:395)
om2_1       | 	at org.apache.hadoop.ozone.protocolPB.OMInterServiceProtocolServerSideImpl.checkLeaderStatus(OMInterServiceProtocolServerSideImpl.java:87)
om2_1       | 	at org.apache.hadoop.ozone.protocolPB.OMInterServiceProtocolServerSideImpl.bootstrap(OMInterServiceProtocolServerSideImpl.java:63)
om2_1       | 	at org.apache.hadoop.ozone.protocol.proto.OzoneManagerInterServiceProtocolProtos$OzoneManagerInterService$2.callBlockingMethod(OzoneManagerInterServiceProtocolProtos.java:1675)
om2_1       | 	at org.apache.hadoop.ipc.ProtobufRpcEngine$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine.java:528)
om2_1       | 	at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1086)
om2_1       | 	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1029)
om2_1       | 	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:957)
om2_1       | 	at java.base/java.security.AccessController.doPrivileged(Native Method)
om2_1       | 	at java.base/javax.security.auth.Subject.doAs(Subject.java:423)
om2_1       | 	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1762)
om2_1       | 	at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2957)
om2_1       | , while invoking $Proxy44.bootstrap over nodeId=om3,nodeAddress=om3:9862 after 488 failover attempts. Trying to failover after sleeping for 2000ms.
```

**Comment by bharatviswa504** [comment]
@hanishakoneru 
Thanks for the update. 

**Few questions,**


1. To convert non-HA to HA cluster?  (Because for this on a single node we need to enable ratis and restart? And also serviceID is not configured, we use omServiceIdDefault  (If we don't specify we use this). And also we need to download snapshot outside ratis, because there are some transactions that happened outside of ratis? As even though we enable ratis, new transactions will be in ratis log, and new OM is bootstrapped it only gets these new logs (If not purged, but it does not get DB, so that is why we need to download snapshot outside ratis)

2. And also a Ratis enabled cluster to HA cluster, I think this can work, if the config has HA style no issues, else if not HA style again they need to use omServiceIdDefault. (No need of download snapshot out side ratis, as this cluster from begining it is ratis enabled)


Can we have a follow-up Jira to give have some documentation on how to handle the above cases?

**Comment by hanishakoneru** [comment]
> To try this scenario, you can update ozone-ha docker-compose.yaml to use below change

This scenario will not work because OM1 was started with all 3 peer information. So it formed a ring assuming there are 3 peers in the system. But since only om1 is active, the Leader Election will never finish.

To test bootstrap, we would have to start non-HA or HA regularly and then update the confs to include a new OM and start the new OM in bootstrap mode.

> To convert non-HA to HA cluster? (Because for this on a single node we need to enable ratis and restart? And also serviceID is not configured, we use omServiceIdDefault (If we don't specify we use this). And also we need to download snapshot outside ratis, because there are some transactions that happened outside of ratis? As even though we enable ratis, new transactions will be in ratis log, and new OM is bootstrapped it only gets these new logs (If not purged, but it does not get DB, so that is why we need to download snapshot outside ratis)

Yes. I will open a new Jira to address this scenario. We could transfer a checkpoint to a bootstrapping OM before the setConf request (but only when there are non-ratis transactions in RocksDB?).

> And also a Ratis enabled cluster to HA cluster, I think this can work, if the config has HA style no issues, else if not HA style again they need to use omServiceIdDefault. (No need of download snapshot out side ratis, as this cluster from begining it is ratis enabled)
Can we have a follow-up Jira to give have some documentation on how to handle the above cases?

Sure. I will open a follow up Jira for this.

**Comment by bharatviswa504** [comment]
> > To try this scenario, you can update ozone-ha docker-compose.yaml to use below change
> 
> This scenario will not work because OM1 was started with all 3 peer information. So it formed a ring assuming there are 3 peers in the system. But since only om1 is active, the Leader Election will never finish.
> 
> To test bootstrap, we would have to start non-HA or HA regularly and then update the confs to include a new OM and start the new OM in bootstrap mode.
> 

IMHO, this looks a little confusing. Because let's say I am the customer who is running ozone, and as I know we donot support rolling upgrade and I need this feature, as I want to convert my non-HA (ratis enabled cluster to HA enabled cluster) generally I will update the config on all the nodes to add 2 more node details, and restart all 3 nodes.  (So, from code point of view it is an upgrade/fresh install should not matter if we can support this).  

Now with this way If I want to work upgrade bits. Do not update config start OM's, and then only update config when adding new OM's. (this will be tricky in docker/Kubernetes world also, as now I need to update config only after the start of older OM's)

(Once this feature comes in, this feature will help to add new OMs with out any restart, after the next release of this feature availability)


I think from a code perspective we can support this. (We can introduce another flag like ozone.om.bootstrap.the startup, if you want to separate out normal start where all OM"s started together with later OM's adding to ring) if this flag is enabled, we start each OM with peers empty, and they will learn about other OM's in-ring with setConf call/addOM.

**Comment by hanishakoneru** [comment]
> IMHO, this looks a little confusing. Because let's say I am the customer who is running ozone, and as I know we donot support rolling upgrade and I need this feature, as I want to convert my non-HA (ratis enabled cluster) to HA enabled cluster. Generally I will update the config on all the nodes to add 2 more node details, and restart all 3 nodes. (Because in general, Ozone will have the same config on all nodes) (So, from a code point of view it is an upgrade/fresh install should not matter if we can support this).

I don't completely agree with this. You are saying we should be able to skip the bootstrap step if all the OM's are started with new config. But what would happen if the old OM (let's say om1) did not come up during this restart. The new OMs (om2 and om3) started up and formed a ring and started servicing requests. This will lead to data loss and an inconsistent system.

> I think from a code perspective we can support this. (We can introduce another flag like ozone.om.bootstrap.the startup, if you want to separate out normal start where all OM"s started together with later OM's adding to ring) if this flag is enabled, we start each OM with peers empty, and they will learn about other OM's in-ring with setConf call/addOM. (We cannot remove the old way start up). On a side note, this is the approach followed in SCM HA.

This approach would also require that one OM be set as primary (similar to the primordial SCM node). Subsequently, it would mean we need to redo the way OM HA works currently. I am not opposed to this approach but it would be a big change and I think not worth the effort (seeing how we have to keep it compatible too).

**Comment by bharatviswa504** [comment]
> > IMHO, this looks a little confusing. Because let's say I am the customer who is running ozone, and as I know we donot support rolling upgrade and I need this feature, as I want to convert my non-HA (ratis enabled cluster) to HA enabled cluster. Generally I will update the config on all the nodes to add 2 more node details, and restart all 3 nodes. (Because in general, Ozone will have the same config on all nodes) (So, from a code point of view it is an upgrade/fresh install should not matter if we can support this).
> 
> I don't completely agree with this. You are saying we should be able to skip the bootstrap step if all the OM's are started with new config. But what would happen if the old OM (let's say om1) did not come up during this restart. The new OMs (om2 and om3) started up and formed a ring and started servicing requests. This will lead to data loss and an inconsistent system.
> 

Sorry, Let me be more clear. What I mean is the old OM which was there before the upgrade will do a normal start, and the other 2 OM's will do bootstrap way of start. So, in my view, this scenario should be the same as let's say one OM I start with normal, and the other 2 OM's with bootstrap on a freshly installed cluster. Because config update needs to happen on all nodes, so updating config after the node starts normally is confusing, if it happens before will it work on an upgraded cluster? This kind of scenario (when the config is updated with newly added nodes when start), will not work only on freshly installed clusters? like this is the [scenario](https://github.com/apache/ozone/pull/1494#issuecomment-859326322)?

> I think from a code perspective we can support this. (We can introduce another flag like ozone.om.bootstrap.the startup, if you want to separate out normal start where all OM"s started together with later OM's adding to ring) if this flag is enabled, we start each OM with peers empty, and they will learn about other OM's in-ring with setConf call/addOM. (We cannot remove the old way start up). On a side note, this is the approach followed in SCM HA.
> 
> This approach would also require that one OM be set as primary (similar to the primordial SCM node). Subsequently, it would mean we need to redo the way OM HA works currently. I am not opposed to this approach but it would be a big change and I think not worth the effort (seeing how we have to keep it compatible too).

This is one way to handle this, as we cannot remove the old way of starting up OM's. So, if I understandand for the normal start we shall follow the existing way of start-up, and when new OM needs to be added then only we shall use bootstrap way?

**Comment by hanishakoneru** [comment]
> Questions:
> 1. On an upgraded cluster, config updated before the start and old OM start normally and new OM's started with bootstrap will it work?
> 2. The same scenario config updated before and one OM started normally, and the other two bootstrap ways on the fresh install cluster will not work, as discussed in the above comments?

Both these scenarios will not work. But I think it is reasonable to ask users to first upgrade (without any config changes) and the follow the steps to convert non-HA to HA.

> This is one way to handle this, as we cannot remove the old way of starting up OM's. So, if I understandand for the normal start we shall follow the existing way of start-up, and when new OM needs to be added then only we shall use bootstrap way?

Correct.

**Comment by bharatviswa504** [comment]
> > Questions:
> > 
> > 1. On an upgraded cluster, config updated before the start and old OM start normally and new OM's started with bootstrap will it work?
> > 2. The same scenario config updated before and one OM started normally, and the other two bootstrap ways on the fresh install cluster will not work, as discussed in the above comments?

I am okay with it if this is the way we want to pursue it. Now I got more clarity on the current PR deployment approach and limitations. We can revisit the decision if there is a requirement to update the config before the upgrade.

File: hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/utils/OzoneManagerRatisUtils.java [hunk 23]
```diff
@@ -340,4 +359,70 @@ public static boolean isBucketFSOptimized() {
...
+  public static String getOMRatisDirectory(ConfigurationSource conf) {
+    String storageDir = conf.get(OMConfigKeys.OZONE_OM_RATIS_STORAGE_DIR);
+
+    if (Strings.isNullOrEmpty(storageDir)) {
+      storageDir = ServerUtils.getDefaultRatisDirectory(conf);
+    }
+    return storageDir;
+  }
+
+  public static String getOMRatisSnapshotDirectory(ConfigurationSource conf) {
```
bharatviswa504, line 374: Add javadocs to public methods

bharatviswa504 (commented):
Overall LGTM.
Few minor comments/questions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -104,9 +104,127 @@
...
+      LOG.debug("OM started in Bootstrap mode and hence will not be added " +
+          "to Ratis group during startup.");
+    } else {
+      // On regular startup, add current OM to Ratis ring
+      raftPeers.add(localRaftPeer);
+    }
+
+    for (OMNodeDetails peerInfo : peerNodes) {
+      String peerNodeId = peerInfo.getNodeId();
+      RaftPeerId raftPeerId = RaftPeerId.valueOf(peerNodeId);
```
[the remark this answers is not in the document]
hanishakoneru: The existing peers can either be added to the RaftServer (of Bootstrapping node) during initialization or they can be added later when the bootstrapping node gets the setConf log entry from leader. AFAIK, either way should work the same.
Now that you mention it, if we do not add the peers during initialization, then we can update the config once for with all new OM nodes. This would be useful when we want to bootstrap multiple OMs.

Will update the PR to use the 2nd approach. Thanks.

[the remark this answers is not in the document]
hanishakoneru: Updated the patch. Please take a look. Thanks.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1339,7 +1339,38 @@ public void bootstrap(OMNodeDetails newOMNode) throws IOException {
...
 
   /**
-   * Add a new OM Node to the HA cluster. This call comes from OMRatisServer
+   * When OMStateMachine receives a configuration change update, it calls
+   * this function to update the peers list, if required.
+   */
+  public void updatePeerList(List<String> omNodeIds) {
+    List<String> ratisServerPeerIdsList = omRatisServer.getPeerIds();
+    for (String omNodeId : omNodeIds) {
+      if (getOMNodeId().equals(omNodeId)) {
```
bharatviswa504, line 1348: In normal OM start node, ratisServer peer list has local + peer.

```
 if (!isBootstrapping) {
      // On regular startup, add all OMs to Ratis ring
      raftPeers.add(localRaftPeer);
```
In bootstrap case, local node is not added to ratis server peer list.

bharatviswa504: While reading more, I have observed that before this change, we have not added local OM node to peer node  list before(raftPeers). Should we have the same behavior?

bharatviswa504 (commented):
Thank You @hanishakoneru for the update.
Overall LGTM. Few comments, once addressed good to commit.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
hanishakoneru: > In bootstrap case, local node is not added to ratis server peer list.

Yes. This is because the local node is not part of the ratis ring yet. When the SetConfiguration is executed, this node would be added to the peer list.

> While reading more, I have observed that before this change, we have not added local OM node to peer node list before(raftPeers). Should we have the same behavior?. So, should we remove adding local node to peer list?

Before this change also the local OM node was added to the peer list in OMRatisServer. Are you referring to any other class?

bharatviswa504:
```
newOMRatisServer
    for (OMNodeDetails peerInfo : peerNodes) {
      String peerNodeId = peerInfo.getNodeId();
      RaftPeerId raftPeerId = RaftPeerId.valueOf(peerNodeId);
      RaftPeer raftPeer;
      if (peerInfo.isHostUnresolved()) {
        raftPeer = RaftPeer.newBuilder()
            .setId(raftPeerId)
            .setAddress(peerInfo.getRatisHostPortStr())
            .build();
      } else {
        InetSocketAddress peerRatisAddr = new InetSocketAddress(
            peerInfo.getInetAddress(), peerInfo.getRatisPort());
        raftPeer = RaftPeer.newBuilder()
            .setId(raftPeerId)
            .setAddress(peerRatisAddr)
            .build();
      }

      // Add other OM nodes belonging to the same OM service to the Ratis ring
      raftPeers.add(raftPeer);
    }

    return new OzoneManagerRatisServer(ozoneConf, omProtocol, omServiceId,
        localRaftPeerId, ratisAddr, raftPeers, secConfig, certClient);

Looks like previously we don't save raftPeers in OzoneManagerRatisServer.

But now with this PR, on old node all nodes peer+local will be in peer list. 

if (!isBootstrapping) { 
// On regular startup, add all OMs to Ratis ring 
raftPeers.add(localRaftPeer);

Where as for bootstrap due to this skip, we don't add

if (getOMNodeId().equals(omNodeId)) {
continue;
}
Due to this on bootstrap node peerlist does not include local node

There is a difference in raftPeers list in bootstrap and old node.

File: hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java [hunk 26]
```diff
@@ -273,106 +483,6 @@ private OMResponse processReply(OMRequest omRequest, RaftClientReply reply)
...
-        omNodeDetails.getInetAddress(), omNodeDetails.getRatisPort());
-
-    RaftPeer localRaftPeer = RaftPeer.newBuilder()
-        .setId(localRaftPeerId)
-        .setAddress(ratisAddr)
-        .build();
-
-    List<RaftPeer> raftPeers = new ArrayList<>();
-    // Add this Ratis server to the Ratis ring
-    raftPeers.add(localRaftPeer);
```
hanishakoneru, line 348: @bharatviswa504, in old code too, local OM node is added to raftPeers list.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
hanishakoneru: > There is a difference in raftPeers list in bootstrap and old node.

Correct. On a bootstrapping node, we do not add the local node to raftPeer list. This is done later when a setConfiguration request is executed and it calls OMRatisServer#addRaftPeer().

```
  /**
   * Add given node to list of RaftPeers.
   */
  public void addRaftPeer(OMNodeDetails omNodeDetails) {
    InetSocketAddress newOMRatisAddr = new InetSocketAddress(
        omNodeDetails.getHostAddress(), omNodeDetails.getRatisPort());

    raftPeers.add(RaftPeer.newBuilder()
        .setId(RaftPeerId.valueOf(omNodeDetails.getNodeId()))
        .setAddress(newOMRatisAddr)
        .build());

    LOG.info("Added OM {} to Ratis Peers list.", omNodeDetails.getNodeId());
  }
```

As for the old code, the local node was added to raftPeers there too (OzoneManagerRatisServer Line#348 in old code).

bharatviswa504: This is done later when a setConfiguration request is executed and it calls OMRatisServer#addRaftPeer().

Yes, but notifyConfigurationChanged calls updatePeerList

And in updatePeerList we have 

    for (String omNodeId : omNodeIds) {
```
      if (getOMNodeId().equals(omNodeId)) {
        continue;
      }
```

Which skips calling addRaftPeer, when it is matching with localNode ID

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
bharatviswa504: Yes, ur correct. Thanks some how i missed this part.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
hanishakoneru: Thanks Bharat. I get it now. Should not skip the checks for local node. I have updated the patch. Thanks for catching this.

bharatviswa504 (commented):
+1 LGTM, pending CI.
Thank You @hanishakoneru for the update.

hanishakoneru:
Thank you @bharatviswa504. I had to make a minor change to fix the unit tests. Can you please take a look at the last commit.

bharatviswa504 (approved):
+1 LGTM

bharatviswa504:
Thank You @hanishakoneru for the work and for patiently addressing questions/review comments

hanishakoneru:
Thank you @bharatviswa504 for diligently reviewing all the iterations.

PR #1495, opened 2020-10-14
HDDS-4343: ReplicationManager.handleOverReplicatedContainer() does not handle unhealthyReplicas properly.

GlenGeng:
What changes were proposed in this pull request?

ReplicationManager.handleOverReplicatedContainer() does not handle unhealthyReplicas properly

```
      // If there are unhealthy replicas, then we should remove them even if it
      // makes the container violate the placement policy, as excess unhealthy
      // containers are not really useful. It will be corrected later as a
      // mis-replicated container will be seen as under-replicated.
      for (ContainerReplica r : unhealthyReplicas) {
        if (excess > 0) {
          sendDeleteCommand(container, r.getDatanodeDetails(), true);
          excess -= 1;
        }
        break;
      }
      // After removing all unhealthy replicas, if the container is still over
      // replicated then we need to check if it is already mis-replicated.
      // If it is, we do no harm by removing excess replicas. However, if it is
      // not mis-replicated, then we can only remove replicas if they don't
      // make the container become mis-replicated.
```

From the comment, it wants to remove all unhealthy replicas until excess reach 0 ? It should be

```
      for (ContainerReplica r : unhealthyReplicas) {
        if (excess > 0) {
          sendDeleteCommand(container, r.getDatanodeDetails(), true);
          excess -= 1;
        } else {
          break;
        }
      }
```

https://issues.apache.org/jira/browse/HDDS-4343#

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
@@ -746,8 +746,9 @@ public class ReplicationManager
         if (excess > 0) {
           sendDeleteCommand(container, r.getDatanodeDetails(), true);
           excess -= 1;
+        } else {
+          break;
         }
-        break;
       }
       // After removing all unhealthy replicas, if the container is still over
       // replicated then we need to check if it is already mis-replicated.


GlenGeng:
@sodonnel  Hi Stephen, Please take a look at this PR. Thanks!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
```diff
@@ -746,8 +746,9 @@ private void handleOverReplicatedContainer(final ContainerInfo container,
         if (excess > 0) {
           sendDeleteCommand(container, r.getDatanodeDetails(), true);
           excess -= 1;
+        } else {
+          break;
         }
-        break;
```
linyiqun, line 750: Looking into this PR change, I am thinking if we still need to do excess check before sending delete command for unhealthy containers. As we want to remove all unhealthy container replicas, we can just send delete command and decrease the excess count by the way.

I prefer simplified the logic like below:
```java
      for (ContainerReplica r : unhealthyReplicas) {
          sendDeleteCommand(container, r.getDatanodeDetails(), true);
          excess -= 1;
      }
```

sodonnel: @linyiqun What if there is an excess of containers, but all containers are somehow unhealthy? We don't want RM to remove all copies of them, as that could result in a lost container. That is why we limit the number of unhealthy containers removed to the excess number.

sodonnel:
@GlenGeng Thanks for highlighting this problem and well done for finding it. The change you have made looks good to me, +1 from my side.

sodonnel (approved):

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ReplicationManager.java
linyiqun: Okay, get it. Thanks @sodonnel.

linyiqun (approved):

PR #1498, opened 2020-10-14
HDDS-4339. Allow AWSSignatureProcessor init when aws signature is absent.

timmylicheng:
Skip aws signature retrieval in AWSSignatureProcessor. It was causing NPE when signature is absent in header.

https://issues.apache.org/jira/browse/HDDS-4339

## How was this patch tested?
UT

### DIFF
--- a/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSSignatureProcessor.java
+++ b/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSSignatureProcessor.java
@@ -129,7 +129,6 @@ public class AWSSignatureProcessor implements SignatureProcessor {
 
 
   private void parse() throws Exception {
-    Preconditions.checkNotNull(v4Header);
 
     StringBuilder strToSign = new StringBuilder();
     // According to AWS sigv4 documentation, authorization header should be
@@ -175,7 +174,6 @@ public class AWSSignatureProcessor implements SignatureProcessor {
 
   @VisibleForTesting
   protected String buildCanonicalRequest() throws OS3Exception {
-    Preconditions.checkNotNull(v4Header);
 
     Iterable<String> parts = split("/", uri);
     List<String> encParts = new ArrayList<>();
--- a/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSSignatureProcessor.java
+++ b/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSSignatureProcessor.java
@@ -42,6 +42,7 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import com.google.common.base.Preconditions;
 import org.apache.hadoop.ozone.s3.exception.OS3Exception;
 import org.apache.hadoop.ozone.s3.header.AuthorizationHeaderV2;
 import org.apache.hadoop.ozone.s3.header.AuthorizationHeaderV4;
@@ -108,23 +109,29 @@ public class AWSSignatureProcessor implements SignatureProcessor {
 
     this.method = context.getMethod();
     String authHeader = headers.get(AUTHORIZATION_HEADER);
-    String[] split = authHeader.split(" ");
-    if (split[0].equals(AuthorizationHeaderV2.IDENTIFIER)) {
-      if (v2Header == null) {
-        v2Header = new AuthorizationHeaderV2(authHeader);
+    if (authHeader != null) {
+      String[] split = authHeader.split(" ");
+      if (split[0].equals(AuthorizationHeaderV2.IDENTIFIER)) {
+        if (v2Header == null) {
+          v2Header = new AuthorizationHeaderV2(authHeader);
+        }
+      } else {
+        if (v4Header == null) {
+          v4Header = new AuthorizationHeaderV4(authHeader);
+        }
+        parse();
       }
-    } else {
-      if (v4Header == null) {
-        v4Header = new AuthorizationHeaderV4(authHeader);
-      }
-      parse();
+    } else { // no auth header
+      v4Header = null;
+      v2Header = null;
     }
   }
 
 
-  public void parse() throws Exception {
-    StringBuilder strToSign = new StringBuilder();
+  private void parse() throws Exception {
+    Preconditions.checkNotNull(v4Header);
 
+    StringBuilder strToSign = new StringBuilder();
     // According to AWS sigv4 documentation, authorization header should be
     // in following format.
     // Authorization: algorithm Credential=access key ID/credential scope,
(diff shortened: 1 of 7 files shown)


hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneClientProducer.java
```diff
@@ -116,6 +116,13 @@ private OzoneClient getClient(OzoneConfiguration config) throws IOException {
     }
   }
 
+  // ONLY validate aws access id when needed.
+  private void validateAccessId(String awsAccessId) throws Exception {
+    if (awsAccessId == null || awsAccessId.equals("")) {
+      throw S3_AUTHINFO_CREATION_ERROR;
```
bharatviswa504, line 122: https://github.com/apache/hadoop-ozone/pull/1110#issuecomment-648405855

When S3 misses the auth header error returned is AuthorizationHeaderMalformedHeader.

Question: After this fix, will it show this error when the auth header is missing or it will be just logged in S3G logs?

timmylicheng: @bharatviswa504 Right now, it catching all exceptions and just print out logs. Do we want to throw OS3Exception?

bharatviswa504: Yes, When I tried AWS S3 end point it shows AuthorizationHeaderMalformed to user.
IMHO S3G needs to mimic AWS S3 behavior, so S3 tools can handle S3G.

timmylicheng: @bharatviswa504 I agree. But it will involve discussion over our internal errors mapping to s3 error codes, which could be a larger discussion. This patch is to have a hot fix for AWSSignatureProcessor init process to avoid NPE. Shall I create a separate JIRA to track handling s3g errors?

bharatviswa504: fine with me if we want to fix the error in a new Jira.
Here we need to map this missing auth header to AuthorizationHeaderMalformed. We should mimic AWS S3 behavior like any other error like BucketNotFound etc.,

bharatviswa504: During initial S3G development, we have followed this error list, and also for API we have tried with S3 to see the behavior and also documentation for that API.

https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList

See if this helps here.

timmylicheng: Open https://issues.apache.org/jira/browse/HDDS-4361 to track proper s3 errors back to bad requests.

timmylicheng: @bharatviswa504 I think a correct handling for ozone internal errors and s3 native errors is very important and would involve more in-depth changes between Ozone client and s3g. HDDS-4361 will be an umbrella JIRA to track the corrections we wanna make.

timmylicheng:
This fix is to keep AWSSignatureProcessor out of NPE so that when OzoneClientProducer could be instantiated even when auth is missing in header. After that OzoneClientProducer can return proper s3 auth errors as part of response, which we would track in HDDS-4361. @ChenSammi @bharatviswa504

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneClientProducer.java
```diff
@@ -80,9 +81,15 @@ private OzoneClient getClient(OzoneConfiguration config) throws IOException {
       UserGroupInformation remoteUser =
```
bharatviswa504, line 81: This will throw NPE when awsAccessId is null, do we need to move validateAccessId before remoteUser creation.

bharatviswa504 (commented):
I have one comment, rest LGTM.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneClientProducer.java
timmylicheng: Sure. Updated.

timmylicheng:
@bharatviswa504 Thanks for the review! I've updated the PR accordingly. Please take another look.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSSignatureProcessor.java
```diff
@@ -108,23 +109,29 @@ public void init()
...
+      v4Header = null;
+      v2Header = null;
     }
   }
 
 
-  public void parse() throws Exception {
-    StringBuilder strToSign = new StringBuilder();
+  private void parse() throws Exception {
+    Preconditions.checkNotNull(v4Header);
```
bharatviswa504, line 132: Minor NIT: do we still need this precondition check.
Sorry missed it earlier?

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSSignatureProcessor.java
```diff
@@ -167,7 +174,9 @@ public void parse() throws Exception {
   }
 
   @VisibleForTesting
-  public String buildCanonicalRequest() throws OS3Exception {
+  protected String buildCanonicalRequest() throws OS3Exception {
+    Preconditions.checkNotNull(v4Header);
```
bharatviswa504, line 178: Minor NIT: do we still need this precondition check.
Same here.

bharatviswa504 (commented):
Thank You @timmylicheng for the update.
Just minor comments, sorry missed them earlier.

xiaoyuyao (approved):
LGTM, +1.

bharatviswa504:
@xiaoyuyao 
I have addressed test failures, during that figured out that it needs a few more changes.
Updated the patch accordingly.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSSignatureProcessor.java
timmylicheng: I think so. Otherwise, v4header is still likely to throw NPE...

bharatviswa504:
Thank You @xiaoyuyao for the review and @timmylicheng for the initial PR.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/AWSSignatureProcessor.java
bharatviswa504: sorry missed this, parse we called only when authHeader!=null.

But anyway, even without this it will fail with NPE, and now this code has moved to try catch and it is properly handled and this error is propagated to OzoneClientProducer. So, the pending part is propagating error from OzoneClientProducer to S3 clients.

PR #1500, opened 2020-10-14
HDDS-4324. DatanodeAdminMonitor no longers needs maintenance end time to be passed

sodonnel:

An earlier change moved the maintenance endtime into the NodeStatus object. However when adding a node to the decommission monitor the end time must still be passed. This value is never used.

This Jira will remove the endInHours field from the interface:

```
public interface DatanodeAdminMonitor extends Runnable {

  void startMonitoring(DatanodeDetails dn, int endInHours);
  void stopMonitoring(DatanodeDetails dn);
}
```

https://issues.apache.org/jira/browse/HDDS-4324

## How was this patch tested?

Existing tests


swagle (approved):
+1 LGTM

PR #1502, opened 2020-10-16
HDDS-4354. Avoid spamming of logs because of deleted transactions.

mukul1987:

Avoid spamming of logs because of deleted transactions.

https://issues.apache.org/jira/browse/HDDS-4354

## How was this patch tested?
MiniOzoneChaosCluster

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lease/LeaseManager.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lease/LeaseManager.java
@@ -235,7 +235,7 @@ public class LeaseManager<T> {
           }
         } catch (InterruptedException e) {
           // This means a new lease is added to activeLeases.
-          LOG.error("Execution was interrupted ", e);
+          LOG.debug("Execution was interrupted ", e);
           Thread.currentThread().interrupt();
         }
       }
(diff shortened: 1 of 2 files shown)


ayushtkn (approved):

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
```diff
@@ -130,9 +130,7 @@ public void incrementCount(List<Long> txIDs) throws IOException {
         DeletedBlocksTransaction block =
             scmMetadataStore.getDeletedBlocksTXTable().get(txID);
         if (block == null) {
-          // Should we make this an error ? How can we not find the deleted
-          // TXID?
-          LOG.warn("Deleted TXID {} not found.", txID);
+          LOG.debug("Deleted TXID {} not found.", txID);
```
lokeshj1703, line 133: NIT: We can include the log inside check for debug log enabled.
```suggestion
          if (LOG.isDebugEnabled()) {
            LOG.debug("Deleted TXID {} not found.", txID);
          }
```

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lease/LeaseManager.java
```diff
@@ -235,7 +235,7 @@ public void run() {
           }
         } catch (InterruptedException e) {
           // This means a new lease is added to activeLeases.
-          LOG.error("Execution was interrupted ", e);
+          LOG.debug("Execution was interrupted ", e);
```
lokeshj1703, line 238: This should not be related to block deletion. What exception are we seeing here?

lokeshj1703 (commented):
@mukul1987  Thanks for working on this! I have minor comments inline.

elek:
/pending Pending on the minor comments from @lokeshj1703

PR #1503, opened 2020-10-16
HDDS-4332: ListFileStatus - do lookup in directory and file tables

rakeshadr:

https://issues.apache.org/jira/browse/HDDS-4332

This task is to perform look up of the user given key path in the directory, file and openFile tables.
OzoneFileSystem APIs:
 - GetFileStatus
 - listStatus 

## How was this patch tested?

Added UTs - TestHadoopDirTreeGeneratorV1, TestOzoneFileInterfacesV1, TestOzoneFileSystemV1...

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
@@ -122,6 +122,7 @@ public class TestOzoneFileInterfaces {
 
   private OMMetrics omMetrics;
 
+  @SuppressWarnings("checkstyle:VisibilityModifier")
   protected boolean enableFileSystemPaths;
 
   public TestOzoneFileInterfaces(boolean setDefaultFs,
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
@@ -56,6 +56,8 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.RandomStringUtils;
 import static org.apache.hadoop.fs.ozone.Constants.OZONE_DEFAULT_USER;
+
+import org.jetbrains.annotations.NotNull;
 import org.junit.After;
 import org.junit.Assert;
 
@@ -120,7 +122,7 @@ public class TestOzoneFileInterfaces {
 
   private OMMetrics omMetrics;
 
-  private boolean enableFileSystemPaths;
+  protected boolean enableFileSystemPaths;
 
   public TestOzoneFileInterfaces(boolean setDefaultFs,
       boolean useAbsolutePath, boolean enabledFileSystemPaths) {
@@ -135,9 +137,8 @@ public class TestOzoneFileInterfaces {
     volumeName = RandomStringUtils.randomAlphabetic(10).toLowerCase();
     bucketName = RandomStringUtils.randomAlphabetic(10).toLowerCase();
 
-    OzoneConfiguration conf = new OzoneConfiguration();
-    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS,
-        enableFileSystemPaths);
+    OzoneConfiguration conf = getOzoneConfiguration();
+
     cluster = MiniOzoneCluster.newBuilder(conf)
         .setNumDatanodes(3)
         .build();
@@ -160,6 +161,14 @@ public class TestOzoneFileInterfaces {
     o3fs = (OzoneFileSystem) fs;
     statistics = (OzoneFSStorageStatistics) o3fs.getOzoneFSOpsCountStatistics();
     omMetrics = cluster.getOzoneManager().getMetrics();
+  }
+
+  @NotNull
+  protected OzoneConfiguration getOzoneConfiguration() {
+    OzoneConfiguration conf = new OzoneConfiguration();
+    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS,
+        enableFileSystemPaths);
+    return conf;
   }
 
   @After
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -64,6 +64,7 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import org.apache.hadoop.test.LambdaTestUtils;
+import org.jetbrains.annotations.NotNull;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Rule;
@@ -100,17 +101,18 @@ public class TestOzoneFileSystem {
   private static final Logger LOG =
       LoggerFactory.getLogger(TestOzoneFileSystem.class);
 
-  private boolean enabledFileSystemPaths;
+  protected boolean enabledFileSystemPaths;
 
-  private MiniOzoneCluster cluster;
-  private FileSystem fs;
-  private OzoneFileSystem o3fs;
-  private String volumeName;
-  private String bucketName;
-  private int rootItemCount;
-  private Trash trash;
+  protected MiniOzoneCluster cluster;
+  protected FileSystem fs;
+  protected OzoneFileSystem o3fs;
+  protected String volumeName;
+  protected String bucketName;
+  protected int rootItemCount;
+  protected Trash trash;
 
   public void testCreateFileShouldCheckExistenceOfDirWithSameName()
+
       throws Exception {
     /*
      * Op 1. create file -> /d1/d2/d3/d4/key2
@@ -153,6 +155,28 @@ public class TestOzoneFileSystem {
     try (FSDataOutputStream outputStream2 = fs.create(file3, false)) {
       fail("Should throw FileAlreadyExistsException");
     } catch (FileAlreadyExistsException fae) {
+      // ignore as its expected
+    }
+
+    // Directory
+    FileStatus fileStatus = fs.getFileStatus(parent);
+    assertEquals("FileStatus did not return the directory",
+            "/d1/d2/d3/d4", fileStatus.getPath().toUri().getPath());
+    assertTrue("FileStatus did not return the directory",
+            fileStatus.isDirectory());
+
+    // invalid sub directory
+    try{
+      fs.getFileStatus(new Path("/d1/d2/d3/d4/key3/invalid"));
+      fail("Should throw FileNotFoundException");
+    } catch (FileNotFoundException fnfe) {
+      // ignore as its expected
+    }
+    // invalid file name
+    try{
+      fs.getFileStatus(new Path("/d1/d2/d3/d4/invalidkey"));
+      fail("Should throw FileNotFoundException");
+    } catch (FileNotFoundException fnfe) {
       // ignore as its expected
     }
 
@@ -249,12 +273,10 @@ public class TestOzoneFileSystem {
     }
   }
 
-  private void setupOzoneFileSystem()
+  protected void setupOzoneFileSystem()
       throws IOException, TimeoutException, InterruptedException {
-    OzoneConfiguration conf = new OzoneConfiguration();
-    conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
-    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS,
-        enabledFileSystemPaths);
+    OzoneConfiguration conf = getOzoneConfig();
+    //conf.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION, "V1");
     cluster = MiniOzoneCluster.newBuilder(conf)
         .setNumDatanodes(3)
         .build();
@@ -276,7 +298,16 @@ public class TestOzoneFileSystem {
     trash = new Trash(conf);
   }
 
-  private void testOzoneFsServiceLoader() throws IOException {
+  @NotNull
+  protected OzoneConfiguration getOzoneConfig() {
+    OzoneConfiguration conf = new OzoneConfiguration();
+    conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
+    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS,
+            enabledFileSystemPaths);
+    return conf;
+  }
+
+  protected void testOzoneFsServiceLoader() throws IOException {
     assertEquals(
         FileSystem.getFileSystemClass(OzoneConsts.OZONE_URI_SCHEME, null),
         OzoneFileSystem.class);
@@ -441,7 +472,7 @@ public class TestOzoneFileSystem {
 
   }
 
-  private void testListStatus() throws Exception {
+  protected void testListStatus() throws Exception {
     Path parent = new Path("/testListStatus");
     Path file1 = new Path(parent, "key1");
     Path file2 = new Path(parent, "key2");
@@ -489,7 +520,7 @@ public class TestOzoneFileSystem {
   /**
    * Tests listStatus operation on root directory.
    */
-  private void testListStatusOnRoot() throws Exception {
+  protected void testListStatusOnRoot() throws Exception {
     Path root = new Path("/");
     Path dir1 = new Path(root, "dir1");
     Path dir12 = new Path(dir1, "dir12");
@@ -516,7 +547,7 @@ public class TestOzoneFileSystem {
   /**
    * Tests listStatus operation on root directory.
    */
-  private void testListStatusOnLargeDirectory() throws Exception {
+  protected void testListStatusOnLargeDirectory() throws Exception {
     Path root = new Path("/");
     Set<String> paths = new TreeSet<>();
     int numDirs = 5111;
@@ -540,7 +571,7 @@ public class TestOzoneFileSystem {
   /**
    * Tests listStatus on a path with subdirs.
    */
-  private void testListStatusOnSubDirs() throws Exception {
+  protected void testListStatusOnSubDirs() throws Exception {
     // Create the following key structure
     //      /dir1/dir11/dir111
     //      /dir1/dir12
@@ -653,7 +684,7 @@ public class TestOzoneFileSystem {
     GenericTestUtils.assertExceptionContains("KEY_NOT_FOUND", ex);
   }
 
-  private void testGetDirectoryModificationTime()
+  protected void testGetDirectoryModificationTime()
       throws IOException, InterruptedException {
     Path mdir1 = new Path("/mdir1");
     Path mdir11 = new Path(mdir1, "mdir11");
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -101,14 +101,22 @@ public class TestOzoneFileSystem {
   private static final Logger LOG =
       LoggerFactory.getLogger(TestOzoneFileSystem.class);
 
+  @SuppressWarnings("checkstyle:VisibilityModifier")
   protected boolean enabledFileSystemPaths;
 
+  @SuppressWarnings("checkstyle:VisibilityModifier")
   protected MiniOzoneCluster cluster;
+  @SuppressWarnings("checkstyle:VisibilityModifier")
   protected FileSystem fs;
+  @SuppressWarnings("checkstyle:VisibilityModifier")
   protected OzoneFileSystem o3fs;
+  @SuppressWarnings("checkstyle:VisibilityModifier")
   protected String volumeName;
+  @SuppressWarnings("checkstyle:VisibilityModifier")
   protected String bucketName;
+  @SuppressWarnings("checkstyle:VisibilityModifier")
   protected int rootItemCount;
+  @SuppressWarnings("checkstyle:VisibilityModifier")
   protected Trash trash;
 
   public void testCreateFileShouldCheckExistenceOfDirWithSameName()
@@ -276,7 +284,6 @@ public class TestOzoneFileSystem {
   protected void setupOzoneFileSystem()
       throws IOException, TimeoutException, InterruptedException {
     OzoneConfiguration conf = getOzoneConfig();
-    //conf.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION, "V1");
     cluster = MiniOzoneCluster.newBuilder(conf)
         .setNumDatanodes(3)
         .build();
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -2288,7 +2288,6 @@ public class KeyManagerImpl implements KeyManager {
     if (numEntries <= 0) {
       return fileStatusFinalList;
     }
-
     String volumeName = args.getVolumeName();
     String bucketName = args.getBucketName();
     String keyName = args.getKeyName();
@@ -2296,11 +2295,9 @@ public class KeyManagerImpl implements KeyManager {
     String seekDirInDB;
     long prefixKeyInDB;
     String prefixPath = keyName;
-
     int countEntries = 0;
 
     // TODO: recursive flag=true will be handled in HDDS-4360 jira.
-
     metadataManager.getLock().acquireReadLock(BUCKET_LOCK, volumeName,
             bucketName);
     try {
@@ -2312,8 +2309,8 @@ public class KeyManagerImpl implements KeyManager {
 
         // Not required to search in DeletedTable because all the deleted
         // keys will be marked directly in dirTable or in keyTable by
-        // breaking the pointer to its sub-dirs. So, there is no issue of
-        // inconsistency.
+        // breaking the pointer to its sub-dirs and sub-files. So, there is no
+        // issue of inconsistency.
 
         /*
          * keyName is a directory.
@@ -2345,8 +2342,7 @@ public class KeyManagerImpl implements KeyManager {
         /*
          * startKey will be used in iterator seek and sets the beginning point
          * for key traversal.
-         *
-         * key name will be used as parentID where the user has requested to
+         * keyName will be used as parentID where the user has requested to
          * list the keys from.
          *
          * When recursive flag=false, parentID won't change between two pages.
@@ -2357,9 +2353,8 @@ public class KeyManagerImpl implements KeyManager {
          * the startKey value.
          */
 
-        // Check startKey is an immediate child of keyName.
-        // For example, keyName=/a/ and expected startKey=/a/b.
-        // startKey can't be /xyz/b.
+        // Check startKey is an immediate child of keyName. For example,
+        // keyName=/a/ and expected startKey=/a/b. startKey can't be /xyz/b.
         if (!OzoneFSUtils.isImmediateChild(keyName, startKey)) {
           if (LOG.isDebugEnabled()) {
             LOG.debug("StartKey {} is not an immediate child of keyName {}. " +
@@ -2377,32 +2372,30 @@ public class KeyManagerImpl implements KeyManager {
             seekDirInDB = metadataManager.getOzonePathKey(prefixKeyInDB,
                     fileStatusInfo.getKeyInfo().getFileName());
 
-            // Order of seek -> (1) Seek dirs in dirTable. In OM, always the
-            // order of search is, first seek into fileTable and then dirTable.
-            // So, its not required to search again into the fileTable.
+            // Order of seek -> (1) Seek dirs only in dirTable. In OM, always
+            // the order of search is, first seek into fileTable and then
+            // dirTable. So, its not required to search again in the fileTable.
 
             // Seek the given key in dirTable.
             getDirectories(fileStatusList, seekDirInDB, prefixPath,
                     prefixKeyInDB, startKey, countEntries, numEntries,
                     volumeName, bucketName, recursive);
-
           } else {
             seekFileInDB = metadataManager.getOzonePathKey(prefixKeyInDB,
                     fileStatusInfo.getKeyInfo().getFileName());
             // begins from the first sub-dir under the parent dir
             seekDirInDB = metadataManager.getOzonePathKey(prefixKeyInDB, "");
 
-            // Seek the given key in key table.
+            // 1. Seek the given key in key table.
             countEntries = getFilesFromDirectory(fileStatusList, seekFileInDB,
                     prefixPath, prefixKeyInDB, startKey, countEntries,
                     numEntries);
-            // Seek the given key in dir table.
+            // 2. Seek the given key in dir table.
             getDirectories(fileStatusList, seekDirInDB, prefixPath,
                     prefixKeyInDB, startKey, countEntries, numEntries,
                     volumeName, bucketName, recursive);
           }
         } else {
-          // No key exists for the given startKey.
           // TODO: HDDS-4364: startKey can be a non-existed key
           if (LOG.isDebugEnabled()) {
             LOG.debug("StartKey {} is a non-existed key and returning empty " +
@@ -2415,8 +2408,6 @@ public class KeyManagerImpl implements KeyManager {
       metadataManager.getLock().releaseReadLock(BUCKET_LOCK, volumeName,
               bucketName);
     }
-
-    // Convert results in cacheKeyMap to List
     for (OzoneFileStatus fileStatus : fileStatusList) {
       if (fileStatus.isFile()) {
         // refreshPipeline flag check has been removed as part of
@@ -2578,6 +2569,7 @@ public class KeyManagerImpl implements KeyManager {
     return countEntries;
   }
 
+  @SuppressWarnings("parameternumber")
   private int addKeyInfoToFileStatusList(Set<OzoneFileStatus> fileStatusList,
       long prefixKeyInDB, String seekKeyInDB, String startKey,
       int countEntries, String cacheKey, OmKeyInfo cacheOmKeyInfo,
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -1773,7 +1773,7 @@ public class KeyManagerImpl implements KeyManager {
 
     if (OzoneManagerRatisUtils.isOmLayoutVersionV1()) {
       return getOzoneFileStatusV1(volumeName, bucketName, keyName,
-              args.getSortDatanodes(), clientAddress, true);
+              args.getSortDatanodes(), clientAddress, false);
     }
     return getOzoneFileStatus(volumeName, bucketName, keyName,
             args.getRefreshPipeline(), args.getSortDatanodes(), clientAddress);
@@ -1842,7 +1842,7 @@ public class KeyManagerImpl implements KeyManager {
 
   private OzoneFileStatus getOzoneFileStatusV1(String volumeName,
       String bucketName, String keyName, boolean sortDatanodes,
-      String clientAddress, boolean skipNonExistsKeyCheck) throws IOException {
+      String clientAddress, boolean skipFileNotFoundError) throws IOException {
     OzoneFileStatus fileStatus = null;
     metadataManager.getLock().acquireReadLock(BUCKET_LOCK, volumeName,
             bucketName);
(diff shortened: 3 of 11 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+      metadataManager.getLock().releaseReadLock(BUCKET_LOCK, volumeName,
+              bucketName);
+    }
+
+    int countEntries = 0;
+    // Convert results in cacheKeyMap to List
+    for (Map.Entry<String, OzoneFileStatus> entry : cacheKeyMap.entrySet()) {
+      // No need to check if a key is deleted or not here, this is handled
+      // when adding entries to cacheKeyMap from DB.
+      if (args.getRefreshPipeline()) {
```
linyiqun, line 2387: We also don't need args.getRefreshPipeline() here due to HDDS-3658 change.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+          // for recursive list all the entries
+          prefixPath = OMFileRequest.getAbsolutePath(prefixPath,
+                  dirInfo.getName());
+          OmKeyInfo omKeyInfo = OMFileRequest.getOmKeyInfo(volumeName,
+                  bucketName, dirInfo, prefixPath);
+          cacheKeyMap.put(entryInDb,
+                  new OzoneFileStatus(omKeyInfo, 0, true));
+          ++countEntries;
+          // files from this directory
+          seekKeyInDB = dirInfo.getObjectID() + OZONE_URI_DELIMITER;
```
linyiqun, line 2362: prefixKeyInDB should also updated to new dir objectID dirInfo.getObjectID()

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -583,4 +588,112 @@ public static OmKeyInfo getOmKeyInfoFromFileTable(boolean openFileTable,
...
+    return null;
+  }
+
+  /**
+   * Prepare OmKeyInfo from OmDirectoryInfo.
+   *
+   * @param volumeName volume name
+   * @param bucketName bucket name
+   * @param dirInfo    directory info
+   * @param keyName    usewr given key name
```
linyiqun, line 659: Typo: usewr -> user

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+
+    List<OzoneFileStatus> fileStatusList = new ArrayList<>();
+    if (numEntries <= 0) {
+      return fileStatusList;
+    }
+
+    String volumeName = args.getVolumeName();
+    String bucketName = args.getBucketName();
+    String keyName = args.getKeyName();
+    // A map sorted by OmKey to combine results from TableCache and DB.
```
linyiqun, line 2288: Update the comment here, actually no results get from TableCache.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+        // startKey will be used in iterator seek and sets the beginning point
+        // for key traversal.
+        // key name will be used as parent ID where the user has requested to
+        // list the keys from.
+        OzoneFileStatus fileStatusInfo = getOzoneFileStatusV1(volumeName,
+                bucketName, startKey, false, null);
+        if (fileStatusInfo != null) {
+          prefixKeyInDB = fileStatusInfo.getKeyInfo().getParentObjectID();
+          seekKeyInDB = prefixKeyInDB + OZONE_URI_DELIMITER
+                  + fileStatusInfo.getKeyInfo().getFileName();
```
linyiqun, line 2327: We should reset prefixPath after we find the ozone dir info to list.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+                  + OZONE_URI_DELIMITER;
+          prefixKeyInDB = fileStatus.getKeyInfo().getObjectID();
+        } else {
+          String bucketKey = metadataManager.getBucketKey(volumeName,
+                  bucketName);
+          OmBucketInfo omBucketInfo =
+                  metadataManager.getBucketTable().get(bucketKey);
+          seekKeyInDB = omBucketInfo.getObjectID()
+                  + OZONE_URI_DELIMITER;
+          prefixKeyInDB = omBucketInfo.getObjectID();
```
linyiqun, line 2315: Should reset prefixPath to bucketKey I think.

linyiqun (commented):
@rakeshadr , thanks for working on this. Initial review comments from me.
Please have a look for this when you get a chance.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+    Preconditions.checkNotNull(args, "Key args can not be null");
+
+    List<OzoneFileStatus> fileStatusList = new ArrayList<>();
+    if (numEntries <= 0) {
+      return fileStatusList;
+    }
+
+    String volumeName = args.getVolumeName();
+    String bucketName = args.getBucketName();
+    String keyName = args.getKeyName();
```
linyiqun, line 2287: Additionally, startKey needs to be normalized when FileSystemPaths enabled, related JIRA: HDDS-4117

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+          prefixKeyInDB = fileStatusInfo.getKeyInfo().getParentObjectID();
+          seekKeyInDB = prefixKeyInDB + OZONE_URI_DELIMITER
+                  + fileStatusInfo.getKeyInfo().getFileName();
+        }
+      }
+
+      // Not required to search in TableCache because all the deleted keys
+      // are marked directly in directory table or in key table by breaking
+      // the pointer to its sub-dirs. So, there is no issue of inconsistency.
+      int countEntries = 0;
```
linyiqun, line 2334: Currently, CreateFile/CreateDir of V1 version also add entry into double buffer cache first, then flush to db. So seems we still need to combine the result from table cache and db.

For the further DeleteFile/DeleteDir API implement, we may still follow this cache way. Not fully understand why here we say ' deleted keys are marked directly in directory table'. Current write type request always operate the cache first, then flush to db I think.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -583,4 +588,112 @@ public static OmKeyInfo getOmKeyInfoFromFileTable(boolean openFileTable,
...
+    builder.addAllMetadata(dirInfo.getMetadata());
+    builder.setVolumeName(volumeName);
+    builder.setBucketName(bucketName);
+    builder.setCreationTime(dirInfo.getCreationTime());
+    builder.setModificationTime(dirInfo.getModificationTime());
+    builder.setObjectID(dirInfo.getObjectID());
+    builder.setUpdateID(dirInfo.getUpdateID());
+    builder.setFileName(dirInfo.getName());
+    builder.setReplicationType(HddsProtos.ReplicationType.RATIS);
+    builder.setReplicationFactor(HddsProtos.ReplicationFactor.ONE);
```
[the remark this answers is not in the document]
rakeshadr: I've followed existing logic during directory creation : [OMDirectoryCreateRequest.java#L351](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMDirectoryCreateRequest.java#L351).

OzoneConfigKeys.OZONE_REPLICATION_TYPE_DEFAULT is string object but 
OmKeyInfo.Builder().setReplicationType(HddsProtos.ReplicationType.RATIS) requires `enum ReplicationType`. Changing code requires a conversion logic.

Also, OZONE_REPLICATION_DEFAULT is THREE but directory has a value ONE.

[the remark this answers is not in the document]
linyiqun: Okay, let us keep current logic here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: "/vol/bucket/" is the root directory and the first level directories are under this. Thats the reason I left, `prefixPath` as empty. PrefixPath is used to build the absolute key path name which will be added to the OMKeyInfo#keyName. Existing code base don't have `volume/bucket name` in the key path name.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Good catch. With recursive flag false, prefix path will always be same as user given keyname. But with recursive flag true, it can change between two pages, if the second page is iterating on the sub-dirs of the given user keyname.

I have raised separate jira HDDS-4360 to handle the recursive logic. IMHO that would be helpful to do more focussed development.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Thanks a lot for pointing out this. Since the startKey value is internally taken from the OMKeyInfo.keyName, I thought the path would have already been in the normalized form. I saw startKey normalization in PR #1451, I would like to hear from @bharatviswa504 about the plans and follow the same logic here as well.

linyiqun: Seems the keyName in OmKeyArgs already be normalized. But startKey is not included in that, it's an independent parameter passed in here. So the PR #1451 did the fix for this.

rakeshadr: Noted and will add TODO in V1 code. I will watch #1451 progress and will incorporate that logic in V1 once that PR is merged. Hope that make sense to you?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
linyiqun: Yes, the logic for recursive makes this more complex.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
linyiqun: Makes sense to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
linyiqun: Thanks for correct me. I am thinking more for this, here the case is that we cannot find the expected key info due to a non-exist key name, right. So maybe we should just return the empty list. Because we give a wrong key name.

rakeshadr: Yes, this condition is for the root directory listing. `fileSystem.listStatus(new Path("o3fs://bucket1.vol1/"))`
For a wrong or non-existing keyname, today it is throwing exception. I have referred code base [KeyManagerImpl.java#L1823](https://github.com/apache/hadoop-ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L1823)

linyiqun: Okay, get it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1831,6 +1838,62 @@ private OzoneFileStatus getOzoneFileStatus(String volumeName,
...
+      // Check if this is the root of the filesystem.
+      if (keyName.length() == 0) {
+        validateBucket(volumeName, bucketName);
+        return new OzoneFileStatus();
+      }
+
+      fileStatus = OMFileRequest.getOMKeyInfoIfExists(metadataManager,
+              volumeName, bucketName, keyName, scmBlockSize);
+
+      // Check if the key is a directory.
```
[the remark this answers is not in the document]
rakeshadr: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+      if (Strings.isNullOrEmpty(startKey)) {
+        OzoneFileStatus fileStatus = getFileStatus(args, clientAddress);
+        if (fileStatus.isFile()) {
+          return Collections.singletonList(fileStatus);
+        }
+
+        // keyName is a directory
+        if (fileStatus.getKeyInfo() != null) {
+          seekKeyInDB = fileStatus.getKeyInfo().getObjectID()
+                  + OZONE_URI_DELIMITER;
```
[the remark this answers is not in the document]
rakeshadr: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+          prefixKeyInDB = omBucketInfo.getObjectID();
+        }
+      } else {
+        // startKey will be used in iterator seek and sets the beginning point
+        // for key traversal.
+        // key name will be used as parent ID where the user has requested to
+        // list the keys from.
+        OzoneFileStatus fileStatusInfo = getOzoneFileStatusV1(volumeName,
+                bucketName, startKey, false, null);
```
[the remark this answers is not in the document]
rakeshadr: Good point. I kept this as an open point now and not addressed in my latest commit. Need to do some more testing to understand the existing behavior.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+
+      if (!isImmediateChild(keyInfo.getParentObjectID(), prefixKeyInDB)) {
+        break;
+      }
+
+      keyInfo.setFileName(keyInfo.getKeyName());
+      String fullKeyPath = OMFileRequest.getAbsolutePath(prefixKeyPath,
+              keyInfo.getKeyName());
+      keyInfo.setKeyName(fullKeyPath);
+      cacheKeyMap.put(entryInDb, new OzoneFileStatus(keyInfo, 0, false));
```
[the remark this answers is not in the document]
rakeshadr: Done

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -249,12 +273,10 @@ public void tearDown() {
...
 
-  private void setupOzoneFileSystem()
+  protected void setupOzoneFileSystem()
       throws IOException, TimeoutException, InterruptedException {
-    OzoneConfiguration conf = new OzoneConfiguration();
-    conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
-    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS,
-        enabledFileSystemPaths);
+    OzoneConfiguration conf = getOzoneConfig();
+    //conf.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION, "V1");
```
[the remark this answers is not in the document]
rakeshadr: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
rakeshadr: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Thanks @linyiqun for pointing out this case. I have implemented File & Dir Table cache logic in latest commit. As the plan is to update the Dir & FileTable directly during delete api, ideally there is no need of considering deletedTableCache. Will revisit this logic once we implement the delete api.

I have commented in [HDDS-4358 jira](https://issues.apache.org/jira/browse/HDDS-4358?focusedCommentId=17218477&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17218477)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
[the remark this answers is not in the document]
rakeshadr: Raised HDDS-4364 jira to handle this case.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2276,318 @@ private void listStatusFindKeyInTableCache(
...
+            // order of search is, first seek into fileTable and then dirTable.
+            // So, its not required to search again into the fileTable.
+
+            // Seek the given key in dirTable.
+            getDirectories(recursive, startKey, numEntries,
+                    fileStatusList, volumeName, bucketName, seekDirInDB,
+                    prefixKeyInDB, prefixPath, countEntries);
+
+          } else {
+            seekFileInDB = metadataManager.getOzonePathKey(prefixKeyInDB,
```
linyiqun, line 2377: For this branch, we use startKey to get the key info and it's a file in key table. So we will not have the same key string as startKey in dir table. Here I think we should just return one  single file
```java
Collections.singletonList(fileStatusInfo);
```

linyiqun (commented):
Thanks for updating the PR, @rakeshadr . One further review comment below.

In additional, current test change not fully cover the test for listStatusV1. Example, current OzoneFS test change doesn't address the case for listStatus with other startKey specified.
I see the test class TestKeyManagerImpl does the good coverage for listStatus call, can we add a test unit like that or make a minor refactor based on that?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
linyiqun: Okay, makes sense to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: As we have to perform seek in two tables to finish the listing, I'm maintaining a seek order 
1) Finish listing all the files from FileTable, then

rakeshadr:
> Thanks for updating the PR, @rakeshadr . One further review comment below.
> 
> In additional, current test change not fully cover the test for listStatusV1. Example, current OzoneFS test change doesn't address the case for listStatus with other startKey specified.
> I see the test class TestKeyManagerImpl does the good coverage for listStatus call, can we add a test unit like that or make a minor refactor based on that?

Thanks @linyiqun . Thats really a good point. I also looked at keyManagerImpl UT and it requires some refactoring effort to modify the createDir and other logic. I will work on and update you.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2276,318 @@ private void listStatusFindKeyInTableCache(
...
+                    fileStatusInfo.getKeyInfo().getFileName());
+            // begins from the first sub-dir under the parent dir
+            seekDirInDB = metadataManager.getOzonePathKey(prefixKeyInDB, "");
+
+            // Seek the given key in key table.
+            countEntries = getFilesFromDirectory(fileStatusList, seekFileInDB,
+                    prefixPath, prefixKeyInDB, startKey, countEntries,
+                    numEntries);
+            // Seek the given key in dir table.
+            getDirectories(recursive, startKey, numEntries, fileStatusList,
```
bharatviswa504, line 2387: Minor NIT: Can we maintain similar order for parameters for getFilesFromDirectories and getDirectories it will be easy to follow while reading code, as it has many parameters.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2276,318 @@ private void listStatusFindKeyInTableCache(
...
+      metadataManager.getLock().releaseReadLock(BUCKET_LOCK, volumeName,
+              bucketName);
+    }
+
+    // Convert results in cacheKeyMap to List
+    for (OzoneFileStatus fileStatus : fileStatusList) {
+      // refreshPipeline flag check has been removed as part of
+      // https://issues.apache.org/jira/browse/HDDS-3658.
+      // Please refer this jira for more details.
+      refreshPipeline(fileStatus.getKeyInfo());
```
bharatviswa504, line 2407: Can we avoid the call to refreshPipeline for directories and also for sortDatanodeInPipeline.

```
if (fileStatus.isFile()) {
refreshPipeline(fileStatus.getKeyInfo());
if (args.getSortDatanodes()) {
        sortDatanodeInPipeline(fileStatus.getKeyInfo(), clientAddress);
      }
    }
}
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2276,318 @@ private void listStatusFindKeyInTableCache(
...
+
+      // No need to check if a key is deleted or not here, this is handled
+      // when adding entries to cacheKeyMap from DB.
+      if (args.getSortDatanodes()) {
+        sortDatanodeInPipeline(fileStatus.getKeyInfo(), clientAddress);
+      }
+    }
+    fileStatusFinalList.addAll(fileStatusList);
```
bharatviswa504, line 2415: When startKey is passed, we seek to that and to construct keyName we used keyArgs keyName as keyName and construct the paths.

But let's say

KeyArgs keyName /a/b
startkey is /f/d

Then we have used parentObjectID and seek and then return all keys in directory /f, but we construct keyNames with prefix as "/a/b".

Do you think we need to have a check when startKey is passed is the parentObjectID is same when recursive is false. To avoid this kind of corner case scenarios?

bharatviswa504 (commented):
Thank You @rakeshadr for the contribution.
I have reviewed code changes, have few comments, pending test changes review.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
linyiqun: Hi @rakeshadr , I agreed with the case you mentioned. That's the case when there is no entry key name exactly 
 equals to startKey. And then we need to lookup the file table, dir table.

 Can you check below corner case I am thinking here:

FileTable 
/a/a1
/a/a1_2

DirTable
/a/a12

Here we pass startKey /a/a1 and find the target file entry which name is completely matched, then we still need to list FileTable, DirTable and get /a/a1_2,  /a/a12 entry?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -583,4 +588,112 @@ public static OmKeyInfo getOmKeyInfoFromFileTable(boolean openFileTable,
...
+   * @param omMetadataMgr metadata manager
+   * @param volumeName    volume name
+   * @param bucketName    bucket name
+   * @param keyName       key name
+   * @param scmBlockSize  scm block size
+   * @return OzoneFileStatus
+   * @throws IOException DB failure
+   */
+  @Nullable
+  public static OzoneFileStatus getOMKeyInfoIfExists(
```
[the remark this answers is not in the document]
rakeshadr: yes, it should under bucket lock. The caller KeyManagerImpl#getOzoneFileStatusV1() has acquireReadLock and hope thats enough?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2276,318 @@ private void listStatusFindKeyInTableCache(
...
+
+      cacheOmKeyInfo.setFileName(cacheOmKeyInfo.getKeyName());
+      String fullKeyPath = OMFileRequest.getAbsolutePath(prefixKeyPath,
+              cacheOmKeyInfo.getKeyName());
+      cacheOmKeyInfo.setKeyName(fullKeyPath);
+
+      if (StringUtils.isBlank(startKey)) {
+        // cacheOmKeyInfo is null if an entry is deleted in cache
+        if (cacheKey.startsWith(seekKeyInDB) &&
+                cacheKey.compareTo(seekKeyInDB) >= 0) {
```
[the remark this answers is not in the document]
rakeshadr: Sorry, I failed to understand the "second check". Are you asking about `cacheKey.compareTo(seekKeyInDB)` this?
If yes, this is required for listing files/dirs under the keyname.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Thanks @linyiqun for the detailed example case.

yes, it will first finish fetching the FileTable entries and then do fetching the DirTable entries of keyname `/a`.

For the startKey `/a/a1 `, it will return list [`/a/a1, /a/a1_2, /a/a12`]. 

In BasicOzoneFileSystem client it will skip the first entry and prepare subList without duplicates `startKey`.

Please refer [BasicOzoneFileSystem.java#L591](https://github.com/apache/ozone/blob/master/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java#L591) for the client side sublist logic to filter out startKey from the return list.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: As I'm implementing only recursive=false cases in this patch, it will return list with only immediate files/dirs. In your example, the keyname is ` /a/b`, then it will list only upto  `/a/b/f`. It won't have ` /a/b/f/d` entry in the result list.

KeyArgs keyName `/a/b`
startkey will be in the fashion `/a/b/f`. It will be full path, right?.

I believe the case what you mentioned will occur when I'm doing recursive=true in jira HDDS-4360, am I correct?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2276,318 @@ private void listStatusFindKeyInTableCache(
...
+
+    return countEntries;
+  }
+
+  private int getFilesFromDirectory(Set<OzoneFileStatus> fileStatusList,
+      String seekKeyInDB, String prefixKeyPath, long prefixKeyInDB,
+      String startKey, int countEntries, long numEntries) throws IOException {
+
+    Table<String, OmKeyInfo> keyTable = metadataManager.getKeyTable();
+    listStatusFindFilesInTableCache(fileStatusList, keyTable, prefixKeyInDB,
```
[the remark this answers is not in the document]
rakeshadr: Done!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2276,318 @@ private void listStatusFindKeyInTableCache(
...
+
+  @SuppressWarnings("parameternumber")
+  protected int getDirectories(boolean recursive, String startKey,
+      long numEntries, Set<OzoneFileStatus> fileStatusList,
+      String volumeName, String bucketName, String seekDirInDB,
+      long prefixKeyInDB, String prefixPath,
+      int countEntries) throws IOException {
+
+    Table dirTable = metadataManager.getDirectoryTable();
+    listStatusFindDirsInTableCache(fileStatusList, dirTable, prefixKeyInDB,
```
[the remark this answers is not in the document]
rakeshadr: Done!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Done!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Done!

rakeshadr:
> Thank You @rakeshadr for the contribution.
> I have reviewed code changes, have few comments, pending test changes review.

Thanks a lot @bharatviswa504 for the review comments. I've updated the PR. Appreciate if you could provide clarifications on few of the comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2205,6 +2272,167 @@ private void listStatusFindKeyInTableCache(
...
+        // for key traversal.
+        // key name will be used as parent ID where the user has requested to
+        // list the keys from.
+        OzoneFileStatus fileStatusInfo = getOzoneFileStatusV1(volumeName,
+                bucketName, startKey, false, null);
+        if (fileStatusInfo != null) {
+          prefixKeyInDB = fileStatusInfo.getKeyInfo().getParentObjectID();
+          seekKeyInDB = prefixKeyInDB + OZONE_URI_DELIMITER
+                  + fileStatusInfo.getKeyInfo().getFileName();
+        }
```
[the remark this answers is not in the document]
rakeshadr: Done!

rakeshadr:
> Thanks for updating the PR, @rakeshadr . One further review comment below.
> 
> In additional, current test change not fully cover the test for listStatusV1. Example, current OzoneFS test change doesn't address the case for listStatus with other startKey specified.
> I see the test class TestKeyManagerImpl does the good coverage for listStatus call, can we add a test unit like that or make a minor refactor based on that?

I've created HDDS-4412 jira to refactor TestKeyManagerImpl unit testing. Today, it uses stale APIs and IMHO will refactor this in a separate task to make this patch comparatively smaller:-)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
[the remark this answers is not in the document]
rakeshadr: OK, based on the offline discussions I will remove the `cacheKey.compareTo(seekKeyInDB) >= 0 `check. Adding discussion details to make the scenario clear to all.
```
'seekKeyInDB' will have two type of values.
      1. "1024/"   -> startKey is null or empty
      2. "1024/b"  -> startKey exists
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Based on offline discussions, added a case to safe check to verify `startKey` is an immediate child of `keyName`.

Following are the valid parentKey formats:
`    parentKey="" or parentKey="/" or parentKey="/a" or parentKey="a"`
Following are the valid startKey formats, which is called as childKey:
`    childKey="/" or childKey="/a/b" or childKey="a/b"`

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
linyiqun: Get it, thanks for the confirmation, @rakeshadr .

rakeshadr:
Thanks @linyiqun for the detailed reviews and your quick responses really helped a lot. Appreciate if you could provide your overall vote/confirmation about the patch.

linyiqun:
> I've created HDDS-4412 jira to refactor TestKeyManagerImpl unit testing. Today, it uses stale APIs and IMHO will refactor this in a separate task to make this patch comparatively smaller:-)

Let's track that on HDDS-4412,

> Thanks @linyiqun for the detailed reviews and your quick responses really helped a lot. Appreciate if you could provide your overall vote/confirmation about the patch.

+1. Give my approval,  : ).

linyiqun (approved):

bharatviswa504 (commented):
HDDS-4391 using Paths object has performance disadvantage, so can we see we can use Hadoop FS Path, as Paths use FileSystem and then initialize UnixPath. 

We can look into this in further Jira.

One minor comment, take care of that in test opened Jira for refactoring.

bharatviswa504 (approved):

bharatviswa504:
Thank You @rakeshadr for the contribution and @linyiqun for the review.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfaces.java
```diff
@@ -162,6 +164,14 @@ public void init() throws Exception {
     omMetrics = cluster.getOzoneManager().getMetrics();
   }
 
+  @NotNull
```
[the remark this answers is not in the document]
rakeshadr: Sure, will take care along with refactoring jira.

rakeshadr:
> [HDDS-4391](https://issues.apache.org/jira/browse/HDDS-4391) using Paths object has performance disadvantage, so can we see we can use Hadoop FS Path, as Paths use FileSystem and then initialize UnixPath.
> 
> We can look into this in further Jira.
> 
> One minor comment, take care of that in test opened Jira for refactoring.

Noted these points and will take care in the following jira tasks.

Thanks a lot @bharatviswa504, @linyiqun for the continuous support in reviews and committing the changes!

PR #1504, opened 2020-10-19
HDDS-4239. [Design] Ozone support truncate operation

runzhiwang:

This is a design doc, which is moved from google doc to make it easier to track the progress.

https://issues.apache.org/jira/browse/HDDS-4239.


maobaolong (approved):
LGTM, We have discuss with @runzhiwang , if there are not any objection idea, I will merge this tomorrow.

captainzmc (approved):
LGTM, +1

linyiqun:
Hi @maobaolong , 
Before do the final merge, can we bring this design in ozone dev mailing list and let more people can look into this?
I don't think other people aware this now.

runzhiwang:
@linyiqun Thanks for suggestions, I have send this design in ozone dev mailing list more than one month.
image

maobaolong:
@linyiqun Thank you for remind me, but this design doc has been sent to ozone-dev when 2020-9-21, and we have discussed this design in the ozone community sync meeting, and no objection for this design, we need to merge this design doc before further coding work, now this PR block our schedule for more than a month, to respect your suggestion, and double check to make sure there are no objection opinion, we can discuss this design doc in the next ozone community sync meeting, everyone are welcome.

The meeting will start at every Friday 12:15 PM (time zone +8, Beijing).

linyiqun:
Okay, maybe I missed that.

maobaolong:
@linyiqun Would you like to join tomorrow's Ozone sync meeting?

runzhiwang:
@xiaoyuyao  Would you like to review this PR again ? So that this PR can be merged before next week.  Thanks a lot.

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+  limitations under the License. See accompanying LICENSE file.
+-->
+
+
+# Abstract
+
+This is a proposal to introduce truncate operation for Ozone, which will allow removing data from the tail of a file.
+ 
+The use cases for truncate include
+1. Transaction, journal handling for external systems. Transactions are appended to a   Ozone file while they succeed. When a transaction fails it is aborted and the file is truncated, which rolls it back to the previous successful transaction.
```
GlenGeng, line 29: Append transactions to a file during pre-write, truncate the file if transaction is rollbacked.

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+# Implementation
+
+## Timing diagram
+
+This timing diagram omit the delete block between SCM and Datanode.
+![avatar](image/truncate.png)
+
+## OM
+
+When OM receives `truncate(key, newLength)`,in the `keyTable`, OM deletes the blocks which are fully truncated, and updates the block length which is partially truncated.
```
GlenGeng, line 45: `truncate(key, truncatedLength)`

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+
+// A partially truncated block, including the newLength
+message PartialTruncateBlock {
+    required BlockID truncateBlock = 1;
+    required uint64 newLength = 2;
+}
+```
+
+## SCM
+When scm receives `TruncateScmKeyRequestProto`, for the `deleteBlocks`, scm has already implemented the code.
```
GlenGeng, line 90:
```
  List<DeleteBlockGroupResult>
      deleteKeyBlocks(List<BlockGroup> keyBlocksInfoList) throws IOException;
```
A truncate key request can be divided into several delete blocks request and a single truncate key request.

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+     required int64 txID = 1;
+     required int64 containerID = 2;
+     repeated PartialTruncateBlock blocks = 3;
+     // the retry time of sending truncating command to datanode.
+     required int32 count = 4;
+}
+```
+
+## Datanode
+When Datanode receives `TruncateBlocksCommand`, in the RocksDB, Datanode deletes the chunks which are fully truncated, and updates the chunk length and checksum which is partially truncated. Datanode puts `<TRUNCATING_blockID, newLength>` in RocksDB, and returns succ to SCM.
```
GlenGeng, line 117: In the master branch, container's rocksdb has been divided into 3 different column families.

hadoop-hdds/docs/content/design/truncate.md
xiaoyuyao: I feel newLength is more consistent with HCFS spec

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+## Timing diagram
+
+This timing diagram omit the delete block between SCM and Datanode.
+![avatar](image/truncate.png)
+
+## OM
+
+When OM receives `truncate(key, newLength)`,in the `keyTable`, OM deletes the blocks which are fully truncated, and updates the block length which is partially truncated.  
+
+OM stores `<key, RepeatedTruncateOmKeyInfo>` in truncateTable, and returns succ to client.  The `TruncateKey` in `RepeatedTruncateOmKeyInfo` represents one truncate key operation, so the list of `TruncateKey` allows us to store a list of truncate operations related to one key.
```
xiaoyuyao, line 47: what's the *key state* after return succ to client but actual truncate is pending? Does it allow truncate? This could lead to data loss is there is no control between pending truncate and append.

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+ */
+message TruncateScmKeyRequestProto {
+     repeated TruncateKey truncateKeys = 1;
+}
+
+/**
+ * A truncated key and its blocks needed to be deleted and the  block needed to be partially truncated.
+ */
+message TruncateKey {
+     required string key = 1;
```
xiaoyuyao, line 77: Is this the full key name including /volume/bucket/key? This seems redundant as it has been normalized in RepeatedTruncateOmKeyInfo

hadoop-hdds/docs/content/design/truncate.md
xiaoyuyao: +1

hadoop-hdds/docs/content/design/truncate.md
xiaoyuyao, line 117: "Datanode puts `<TRUNCATING_blockID, newLength>` in RocksDB, and returns succ to SCM." This should be placed before the text about the async chunk delete.

xiaoyuyao (commented):
LGTM, a few more comments inline.

hadoop-hdds/docs/content/design/truncate.md
runzhiwang: newLength is aim to keep consistence

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+    required uint64 newLength = 2;
+}
+```
+
+## SCM
+When scm receives `TruncateScmKeyRequestProto`, for the `deleteBlocks`, scm has already implemented the code.
+
+For the `partialTruncateBlocks`, we process it as a transaction like `deleteBlocks`, store `<transactionId, TruncatedBlocksTransaction>` in `truncatedBlocksTable`, and return succ to OM.  
+
+We abstract the code related to delete block transaction, so that truncate and delete blocks can share the abstract code.
```
GlenGeng, line 94: SCM will hold on the delete blocks request for open container. Will the same logic be used on truncate blocks ?

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+     // the retry time of sending truncating command to datanode.
+     required int32 count = 4;
+}
+```
+
+## Datanode
+When Datanode receives `TruncateBlocksCommand`, in the RocksDB, Datanode deletes the chunks which are fully truncated, and updates the chunk length and checksum which is partially truncated. Datanode puts `<TRUNCATING_blockID, newLength>` in RocksDB, and returns succ to SCM.
+
+Datanode starts a background thread to process `<TRUNCATING_blockID, newLength>`.
+If it is `FilePerBlock`, we use `FileChannel.truncate` to truncate file to newLength directly.  If it is `FilePerChunk`, we delete the files of the fully truncated chunks, and use `FileChannel.truncate` to process the partially truncated file.  Then, in RocksDB, Datanode delete `<TRUNCATING_blockid, newLength>`, and put `<TRUNCATED_blockid, newLength>`.
```
GlenGeng, line 120: I consider the key in RocksDB should be `localId` instead of `blockId`.

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+
+Both 2 cases are acceptable, HDFS also does not guarantee the behavior of the read client.
+
+### Truncate When Truncate
+
+When multi clients truncate at the same time, the BUCKET_LOCK can avoid inconsistency.
+
+### Truncate when Append
+
+Append mean ozone client can append some content to the tail of some key. So if Client1 truncates the file but datanode has not truncated blockN -> Client2 appends some content to the tail of blockN -> Datanode truncates blockN, then error happens. To avoid this, OmKeyLocationInfo add a new flag, i.e. toBeTruncated, to mark the block need to be truncated in the future. When Client2 appends, Client2 finds blockN with toBeTruncated flag, Client2 will allocate a new block, and append the content to the new block.
```
GlenGeng, line 142: After toBeTruncated is set, it is hard to reset this flag in architecture of current Ozone

bshashikant (commented):
How about supporting truncate only at block level granularity i.e, If a key needs to be truncated, change the lengths of the partial block to be truncated in OM and all the other blocks which need to be deleted fully can just be deleted . These blocks can then by picked by KeyDeleting service in OM and will get deleted by Delete Workflow in ozone?

runzhiwang:
> How about supporting truncate only at block level granularity i.e, If a key needs to be truncated, change the lengths of the partial block to be truncated in OM and all the other blocks which need to be deleted fully can just be deleted . These blocks can then by picked by KeyDeleting service in OM and will get deleted by Delete Workflow in ozone?

@bshashikant  +1, thanks the suggestion. With this suggestion, for the partially truncate block, we only need to change the block length in OM, and do not need to change SCM and Datanode. the work become easy. The only cons is we waste maxum 128MB (size of block) of disk of each truncate operation, because we do not process the partially truncate block. But because truncate happen sparsely, so it's not a problem.  If we want to process the partially truncate block to save disk capacity, we can still do it in future, it's compatible with current design. @GlenGeng @xiaoyuyao  What do you think ?

GlenGeng:
Yeah, this is quite a straightforward solution. 
+1

hadoop-hdds/docs/content/design/truncate.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Related New Class
+```
+// include a list of truncate operation
+public RepeatedTruncateOmKeyInfo {
+    private final String volumeName;
+    private final String bucketName;
+    // name of key client specified
+    private String keyName;
+    public List<TruncateKey> truncateKeyList;
```
linyiqun, line 59: Can rename TruncateKey to TruncateBlock? It lets me a little confused for the name of this.

hadoop-hdds/docs/content/design/truncate.md
linyiqun, line 142: When appending content on an existed key, this key should be not in a closed state. So the truncate won't happen based on above assumption. Is this still a valid scenery of  truncate when Append?

linyiqun (commented):
Hi @runzhiwang , I left some comments below.

In additional, one another question:
As we know ReplicationManager will also do the replication for the closed containers(blocks), but now truncate operation can update the content for this state containers. So how do we  do the consistency control between this?

runzhiwang:
> In additional, one another question:
> As we know ReplicationManager will also do the replication for the closed containers(blocks), but now truncate operation can update the content for this state containers. So how do we do the consistency control between this? If we are doing truncate operation for one key while this key block container is in replication, there is a large possibility this truncate behavior will be failed.

@linyiqun Thanks for review. In this design, we only delete the fully truncated block and change the length of partially truncated block in OM, and delete the fully truncated block file in datanode. In OM, HA response for consistency. In datanode, delete block has already implemented for delete key, we can use this logic to keep consistency.

hadoop-hdds/docs/content/design/truncate.md
runzhiwang: @linyiqun After close key, we can first truncate the key, and then we can append the key. So it could happen.

kerneltime:
@runzhiwang @ChenSammi I will close this PR if it is not being worked on.

kerneltime:
Closing this PR for now, we can reopen it when we want to.

PR #1513, opened 2020-10-22
HDDS-4366. SCM deletion service should delete configured number of blocks every interval.

lokeshj1703:

SCM service currently uses datanode's configuration to determine the number of blocks to delete every interval. It should have its own congifuration for maximum number of blocks to delete in every interval.
Further it currently scans the entire DB to fetch block deletion transactions. This can be avoided with this approach. With this patch service would always fetch configured number of blocks from the db.

https://issues.apache.org/jira/browse/HDDS-4366

## How was this patch tested?

Existing UT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
@@ -200,7 +200,7 @@ public class DeletedBlockLogImpl
             LOG.debug("Transaction txId={} commit by dnId={} for containerID={}"
                     + " failed. Corresponding entry not found.", txID, dnID,
                 containerId);
-            return;
+            continue;
           }
 
           dnsWithCommittedTxn.add(dnID);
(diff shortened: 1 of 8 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
```diff
@@ -127,62 +112,65 @@ public int getPriority() {
...
+          }
+
+          containerManager.updateDeleteTransactionId(containerIdToMaxTxnId);
+          LOG.info("Totally added {} blocks to be deleted for"
+                  + " {} datanodes, task elapsed time: {}ms",
+              transactions.getBlocksDeleted(),
+              transactions.getDatanodeTransactionMap().size(),
+              Time.monotonicNow() - startTime);
         } catch (IOException e) {
           // We may tolerant a number of failures for sometime
```
bshashikant, line 164: tolerant ---> tolerate

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
```diff
@@ -323,31 +322,54 @@ public void addTransactions(Map<Long, List<Long>> containerBlocksMap)
   public void close() throws IOException {
   }
 
+  private void getTransaction(DeletedBlocksTransaction tx,
```
bshashikant, line 325: getTransaction -> getTransactions

bshashikant (approved):
looks good with minor cosmetic changes.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
lokeshj1703: The function is used to fetch a single transaction so avoided s at end here.

lokeshj1703:
@bshashikant Thanks for reviewing the PR! I have added a commit which addresses the comments.

bshashikant (approved):

lokeshj1703:
@bshashikant Thanks for the review! I have committed the PR to master branch.

PR #1514, opened 2020-10-22
HDDS-4191 Add failover proxy for SCM container location.

timmylicheng:
Add failover proxy for SCM container client

https://issues.apache.org/jira/browse/HDDS-4191

## How was this patch tested?
UT

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
@@ -22,11 +22,10 @@ import com.google.common.annotations.VisibleForTesting;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.scm.ScmConfigKeys;
-import org.apache.hadoop.hdds.scm.protocolPB.ScmBlockLocationProtocolPB;
+import org.apache.hadoop.hdds.scm.protocolPB.StorageContainerLocationProtocolPB;
 import org.apache.hadoop.hdds.utils.LegacyHadoopConfigurationSource;
 import org.apache.hadoop.io.retry.FailoverProxyProvider;
 import org.apache.hadoop.io.retry.RetryPolicy;
-import org.apache.hadoop.io.retry.RetryPolicy.RetryAction;
 import org.apache.hadoop.ipc.ProtobufRpcEngine;
 import org.apache.hadoop.ipc.RPC;
 import org.apache.hadoop.net.NetUtils;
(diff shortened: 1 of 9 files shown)


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
```diff
@@ -0,0 +1,284 @@
...
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_NAMES;
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_SERVICE_IDS_KEY;
+
+/**
+ * Failover proxy provider for SCM container location.
+ */
+public class SCMContainerLocationFailoverProxyProvider implements
+    FailoverProxyProvider<StorageContainerLocationProtocolPB>, Closeable {
+  public static final Logger LOG =
+      LoggerFactory.getLogger(StorageContainerLocationProtocolPB.class);
```
xiaoyuyao, line 58: LOG name should be SCMContainerLocationFailoverProxyProvider.class.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
```diff
@@ -0,0 +1,284 @@
...
+  private final int maxRetryCount;
+  private final long retryInterval;
+
+  public static final String SCM_DUMMY_NODEID_PREFIX = "scm";
+
+  public SCMContainerLocationFailoverProxyProvider(ConfigurationSource conf) {
+    this.conf = conf;
+    this.scmVersion = RPC.getProtocolVersion(
+        StorageContainerLocationProtocolPB.class);
+    this.scmServiceId = conf.getTrimmed(OZONE_SCM_SERVICE_IDS_KEY);
```
xiaoyuyao, line 82: Can we move line 82 into loadConfigs()?

timmylicheng: Wanted to keep it as final variable so that value is only set in constructor. What do you think?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
timmylicheng: Updated.

elek:
Is this change compatible with the hadoop2.x client? 

With OM we created OMTransport with two implementation: 

 * one for hadoop3 which used the failover proxy
 * one for hadoop2 which is without failover proxy

I don't have the full context, but we may need a similar method here, too.

timmylicheng:
> Is this change compatible with the hadoop2.x client?
> 
> With OM we created OMTransport with two implementation:
> 
> * one for hadoop3 which used the failover proxy
> * one for hadoop2 which is without failover proxy
> 
> I don't have the full context, but we may need a similar method here, too.

@elek Hey Marton. I think the failover proxy here is a little different from OM's. Here we are retrying with IP addresses specified in SCM HA configurations to see who should be the current leader SCM to handle requests. It doesn't seem to involve hadoop version.

elek:
Thanks for the answer @timmylicheng 

As far as I remember the `FailoverProxyProvider` caused the problem which is refactored between hadoop2 and hadoop3. 

But thinking it again it shouldn't be a problem at all. I assume the client will use only OM calls as before, and we can use any hadoop3.x classes on server side. So my question doesn't have any sense ;-)

ChenSammi:
LGTM + 1.

PR #1515, opened 2020-10-22
HDDS-4373. [Design] Ozone support append operation 

captainzmc:

This is a design doc, which is moved from google doc to make it easier to track the progress.

https://issues.apache.org/jira/browse/HDDS-4373

## How was this patch tested?

You can see all content here at: https://github.com/captainzmc/hadoop-ozone/blob/add-append-doc/hadoop-hdds/docs/content/design/append.md


hadoop-hdds/docs/content/design/append.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Solution：
+When Append occurs, determine if the container for the last block is closed. If it's closed, we create a new block. if it's open we append data to the last block. This can avoid creating new blocks as much as possible.
+                                                                                                                                                                              
+### Request process：
+![avatar](doc-image/append.png)
+
+ 1. Client executes append key operation to OM
+
+ 2. OM checks if the key is in appendTable; if so, the key is being called by another client append. we cannot append this key at this point. If not, add the key to appendTable.
```
arp7, line 44: Why not have the last append win?

hadoop-hdds/docs/content/design/append.md
```diff
@@ -0,0 +1,87 @@
...
+
+ 3. Check whether the last block of the key belongs to a closed container, if so, apply to SCM allocate a new block, if not, use the current block directly.
```
arp7, line 46: Blocks must be immutable, we should never modify the contents of a block.

hadoop-hdds/docs/content/design/append.md
captainzmc: Thanks  for @arp7's review. This similar to implementing a lock for the current Append key. If thread A is in the append key(it get the append lock for the key), thread B cannot append.

hadoop-hdds/docs/content/design/append.md
captainzmc: Thanks  for @arp7's review. Does changing an existing block cause any other problems?
As far as I know, we can't change the block under the closed container because we can't try to reopen the container.However, the block under the container that is not closed can be modified. 
The purpose of this is to reduce the number of key blocks. If a key Append is particularly frequent, each append will generate a new block. And then finally this key is going to produce a bunch of little blocks. Too many little blocks will make the DB larger and also have an impact on read performance.
The current [truncate design](https://github.com/apache/ozone/pull/1504/files) also needs to modify the existing block.

captainzmc: Thanks  for @arp7's review. Does changing an existing block cause any other problems?
As far as I know, we can't change the block under the closed container because we can't try to reopen the container.However, the block under the container that is not closed can be modified. 
The purpose of this is to reduce the number of key blocks. If a key Append is particularly frequent, each append will generate a new block. And then finally this key is going to produce a bunch of little blocks. Too many little blocks will make the DB larger and also have an impact on read performance.
The current [ozone truncate design](https://github.com/apache/ozone/pull/1504/files) also needs to modify the existing block.

runzhiwang (approved):

hadoop-hdds/docs/content/design/append.md
linyiqun: appendTable  is not really needed, why not reuse openkey file to indicating one key that is being appended.

captainzmc: Thanks for @linyiqun's  advice. The original solution was to use the existing openkeyTable. However, considering the possibility of later implementation changes, we use openkeyTable here.

captainzmc:
Maybe it's not time to talk about Append yet, so turn this PR off for now. JIRA will remain in place until we have a real need for it.

PR #1517, opened 2020-10-23
HDDS-4375. OM changes the block length when receives truncate request

runzhiwang:

1. When OM receives truncate(key, newLength),in the keyTable, OM deletes the blocks which are fully truncated, and updates the block length which is partially truncated, then return success to client.

2. Ozone client read key according to the block length got from OM, which has already been implemented. 

3. So, with this PR, Ozone client can read the truncated key with newLength, when the truncate operation returns success.

https://issues.apache.org/jira/browse/HDDS-4375

## How was this patch tested?

new UT and IT

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
@@ -134,6 +134,7 @@ import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.SetAclR
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.SetAclResponse;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.SetBucketPropertyRequest;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.SetVolumePropertyRequest;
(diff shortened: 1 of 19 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -775,6 +776,30 @@ public void deleteKeys(OmDeleteKeys deleteKeys) throws IOException {
 
   }
 
+  /**
+   * Truncates an existing key.
+   *
+   * @param args the args of the key.
+   * @throws IOException
```
maobaolong, line 783: Would you please finish this java doc, there are many same issue in this PR.

For example `if there is an I/O error while performing the call`

maobaolong:
@runzhiwang Thank you for your contribution, It's looks good overall, left a minor comments.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestOMRequestUtils.java
```diff
@@ -73,11 +73,19 @@
  * Helper class to test OMClientRequest classes.
  */
 public final class TestOMRequestUtils {
+  private static final long DATA_SIZE = 1000L;
 
   private TestOMRequestUtils() {
     //Do nothing
   }
```
captainzmc, line 80: Useless methods can be deleted

captainzmc (commented):
LGTM overall. Just few minor comments inline.

PR #1518, opened 2020-10-23
HDDS-4386: Each EndpointStateMachine uses its own thread pool to talk with SCM/Recon

GlenGeng:

In Tencent production environment, after start Recon for a while, we got warnings that all DNs become stale/dead at SCM side. After kill recon, all DNs become healthy in a very short time.

**The root cause is:**
1. EndpointStateMachine for SCM and that for Recon share the thread pool created by DatanodeStateMachine, which is a fixed size thread pool:
```
executorService = Executors.newFixedThreadPool(
    getEndPointTaskThreadPoolSize(),
    new ThreadFactoryBuilder()
        .setNameFormat("Datanode State Machine Task Thread - %d").build());
```

```
private int getEndPointTaskThreadPoolSize() {
  // TODO(runzhiwang): current only support one recon, if support multiple
  //  recon in future reconServerCount should be the real number of recon
  int reconServerCount = 1;
  int totalServerCount = reconServerCount;

  try {
    totalServerCount += HddsUtils.getSCMAddresses(conf).size();
  } catch (Exception e) {
    LOG.error("Fail to get scm addresses", e);
  }

  return totalServerCount;
}
```
meanwhile, current Recon has some performance issue, after running for hours, it became slower and slower, and crashed due to OOM. 
2. The communication between DN and Recon will soon exhaust all the threads in DatanodeStateMachine.executorService, there will be no available threads for DN to talk SCM. 
3. all DNs become stale/dead at SCM side.
 
**The fix is quite straightforward:**
Each EndpointStateMachine uses its own thread pool to talk with SCM/Recon, a slow Recon won't interfere the communication between DN and SCM, or vice versa.
 
**P.S.**
The first edition for` DatanodeStateMachine.executorService` is a cached thread pool, if there exists a slow SCM/Recon, more and more threads will be created, and DN will OOM eventually, due to tens of thousands of threads are created.

https://issues.apache.org/jira/browse/HDDS-4386

CI

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/EndpointStateMachine.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/EndpointStateMachine.java
@@ -16,6 +16,7 @@
  */
 package org.apache.hadoop.ozone.container.common.statemachine;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.ozone.protocol.VersionResponse;
 import org.apache.hadoop.ozone.protocolPB
@@ -27,6 +28,8 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.time.ZonedDateTime;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
@@ -51,6 +54,7 @@ public class EndpointStateMachine
   private VersionResponse version;
   private ZonedDateTime lastSuccessfulHeartbeat;
   private boolean isPassive;
+  private final ExecutorService executorService;
 
   /**
    * Constructs RPC Endpoints.
@@ -66,6 +70,11 @@ public class EndpointStateMachine
     state = EndPointStates.getInitState();
     lock = new ReentrantLock();
     this.conf = conf;
+    executorService = Executors.newSingleThreadExecutor(
+        new ThreadFactoryBuilder()
+            .setNameFormat("EndpointStateMachine task thread for "
+                + this.address + " - %d ")
+            .build());
   }
 
   /**
@@ -127,6 +136,13 @@ public class EndpointStateMachine
   public EndPointStates setState(EndPointStates epState) {
     this.state = epState;
     return this.state;
+  }
+
+  /**
+   * Returns the endpoint specific ExecutorService.
+   */
+  public ExecutorService getExecutorService() {
+    return executorService;
   }
 
   /**
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
@@ -140,7 +140,13 @@ public class RunningDatanodeState implements DatanodeState {
     for (EndpointStateMachine endpoint : connectionManager.getValues()) {
       Callable<EndPointStates> endpointTask = getEndPointTask(endpoint);
       if (endpointTask != null) {
-        ecs.submit(endpointTask);
+        // Just do a timely wait. A slow EndpointStateMachine won't occupy
+        // the thread in executor from DatanodeStateMachine for a long time,
+        // so that it won't affect the communication between datanode and
+        // other EndpointStateMachine.
+        ecs.submit(() -> endpoint.getExecutorService()
+            .submit(endpointTask)
+            .get(context.getHeartbeatFrequency(), TimeUnit.MILLISECONDS));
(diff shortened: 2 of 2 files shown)


avijayanhwx:
@GlenGeng Thanks for fixing this issue.

Is there a JIRA created for the Recon slowness issue?

Also, irrespective of whether there is a slow SCM or Recon, the DN should not submit a request to an endpoint if there is a pending request to that endpoint. Is that not true currently? Even if there are 2 threads in the same thread pool (1 for SCM and 1 for Recon), then a slow Recon should not affect SCM reports.

GlenGeng:
@avijayanhwx Thanks for looking at this PR. 
@ChenSammi Could you please create a Jira for slow Recon ?

> Also, irrespective of whether there is a slow SCM or Recon, the DN should not submit a request to an endpoint if there is a pending request to that endpoint. Is that not true currently?

It is not true.

`DatanodeStateMachine.stateMachineThread` calls `context.execute(executorService, heartbeatFrequency, TimeUnit.MILLISECONDS);` every heartbeatFrequency. 

The job is
```
DatanodeState<DatanodeStateMachine.DatanodeStates> task = getTask();
task.execute(service);
DatanodeStateMachine.DatanodeStates newState = task.await(time, unit);
```
where the task is either `InitDatanodeState` or `RunningDatanodeState`, depending on state of Datanode state machine:
```
  public DatanodeState<DatanodeStateMachine.DatanodeStates> getTask() {
    switch (this.state) {
    case INIT:
      return new InitDatanodeState(this.conf, parent.getConnectionManager(),
          this);
    case RUNNING:
      return new RunningDatanodeState(this.conf, parent.getConnectionManager(),
          this);
    case SHUTDOWN:
      return null;
    default:
      throw new IllegalArgumentException("Not Implemented yet.");
    }
  }
```

Say it is `RunningDatanodeState`, it will do a endpointTask for each endpoint, which does not consider whether there is an outstanding request for that endpoint.
```
  public void execute(ExecutorService executor) {
    ecs = new ExecutorCompletionService<>(executor);
    for (EndpointStateMachine endpoint : connectionManager.getValues()) {
      Callable<EndPointStates> endpointTask = getEndPointTask(endpoint);
      if (endpointTask != null) {
        ecs.submit(endpointTask);
      } else {
        // This can happen if a task is taking more time than the timeOut
        // specified for the task in await, and when it is completed the task
        // has set the state to Shutdown, we may see the state as shutdown
        // here. So, we need to Shutdown DatanodeStateMachine.
        LOG.error("State is Shutdown in RunningDatanodeState");
        context.setState(DatanodeStateMachine.DatanodeStates.SHUTDOWN);
      }
    }
  }
```

and the 'await()' will do a timely wait, so that it will not block the next heartbeat task of DatanodeStateMachine
```
  public DatanodeStateMachine.DatanodeStates
      await(long duration, TimeUnit timeUnit)
      throws InterruptedException {

    ....
    while (returned < count && timeLeft > 0) {
      Future<EndPointStates> result =
          ecs.poll(timeLeft, TimeUnit.MILLISECONDS);
      if (result != null) {
        results.add(result);
        returned++;
      }
      timeLeft = durationMS - (Time.monotonicNow() - startTime);
    }
    return computeNextContainerState(results);
  }
```

Consider `HeartbeatEndpointTask`, its `call()` just acquire the lock and do a heartbeat RPC:
```
  public EndpointStateMachine.EndPointStates call() throws Exception {
    rpcEndpoint.lock();

    try {
      ...
      SCMHeartbeatResponseProto reponse = rpcEndpoint.getEndPoint()
          .sendHeartbeat(request);
      processResponse(reponse, datanodeDetailsProto);
      ...
    } catch (IOException ex) {
      ...
    } finally {
      rpcEndpoint.unlock();
    }
    return rpcEndpoint.getState();
  }
```

Here is the problem:
Say there is a slow Recon, two threads in thread pool. During the first heartbeat, one thread will acquire the lock and pending on the RPC with Recon, during the next heartbeat, the idle thread, after it finishes the job with SCM, it will finally blocked on acquiring endpoint lock of Recon, during the third heartbeat, there will no available threads in the thread pool.

runzhiwang (approved):
LGTM

avijayanhwx (approved):
LGTM +1

avijayanhwx:
> @GlenGeng Thanks for fixing this issue.
> 
> Is there a JIRA created for the Recon slowness issue?
> 
> Also, irrespective of whether there is a slow SCM or Recon, the DN should not submit a request to an endpoint if there is a pending request to that endpoint. Is that not true currently? Even if there are 2 threads in the same thread pool (1 for SCM and 1 for Recon), then a slow Recon should not affect SCM reports.

Thanks for the explanation @GlenGeng. IMO, it is good to have Recon and SCM use different thread pools.

GlenGeng:
@avijayanhwx We've create two Jiras for the Recon issue, HDDS-4385 and HDDS-4355. I've cc you in them.

BTW, in our production cluster, since starting up recon will make the cluster dead, we can not debug the issue until next release with this fix. I guess we might have to wait for a while QAQ.

avijayanhwx:
@GlenGeng  / @ChenSammi How many containers do you have on your prod cluster? I have encountered a similar issue in another cluster with million containers. The problem is explained in HDDS-4403.

GlenGeng:
@avijayanhwx  @ChenSammi 
Our cluster has about 2 millions containers. Our phenomenon is very similar to the explanations in HDDS-4403.

PR #1519, opened 2020-10-23
HDDS-4388. Make writeStateMachineTimeout retry count proportional to node failure timeout

bshashikant:
Currently, in ratis "writeStateMachinecall" gets retried indefinitely in event of a timeout. In case, where disks are slow/overloaded or number of chunk writer threads are not available for a period of 10s, writeStateMachine call times out in 10s. In cases like these, the same write chunk keeps on getting retried causing the same chunk of data to be overwritten. The idea here is to abort the request once the node failure timeout reaches.

https://issues.apache.org/jira/browse/HDDS-4388

## How was this patch tested?
Verified by checking the config value in tests.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
@@ -163,6 +163,9 @@ public final class XceiverServerRatis implements XceiverServerSpi {
     this.containerController = containerController;
     this.raftPeerId = RatisHelper.toRaftPeerId(dd);
     chunkExecutors = createChunkExecutors(conf);
+    nodeFailureTimeoutMs =
+            conf.getObject(DatanodeRatisServerConfig.class)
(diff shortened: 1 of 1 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
```diff
@@ -200,6 +200,9 @@ private RaftProperties newRaftProperties() {
     TimeUnit timeUnit;
     long duration;
 
+    // set the node failure timeout
+    setNodeFailureTimeout(properties);
```
lokeshj1703, line 204: `setNodeFailureTimeout` initializes a field. I think we can initialize the field in the constructor itself and remove this function.

lokeshj1703 (commented):
@bshashikant Thanks for working on this! The changes look good to me. I have a minor comment - unrelated to PR. +1 o.w.

lokeshj1703 (approved):
@bshashikant Thanks for updating the PR! The changes look good to me. +1.

lokeshj1703:
@bshashikant Thanks for the contribution! I have merged the PR to master branch.

PR #1526, opened 2020-10-28
HDDS-4399. Safe mode rule for piplelines should only consider open pipelines.

bshashikant:
Currently, for safe mode we consider all pipelines existing in DB for safe mode exit criteria. It ma happen that, SCM has the pipelines craeted , but none of the participants datanodes ever created these datanodes. In such cases, SCM fails to come out of safemode as these pipelines are never reported back to SCM.

The idea here is to consider pipelines which are marked open during SCM startup.
https://issues.apache.org/jira/browse/HDDS-4399

## How was this patch tested?
Added unit tests

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/safemode/HealthyPipelineSafeModeRule.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/safemode/HealthyPipelineSafeModeRule.java
@@ -78,7 +78,7 @@ public class HealthyPipelineSafeModeRule extends SafeModeExitRule<Pipeline> {
     // We want to wait for RATIS THREE factor write pipelines
     int pipelineCount = pipelineManager.getPipelines(
         HddsProtos.ReplicationType.RATIS, HddsProtos.ReplicationFactor.THREE,
-        Pipeline.PipelineState.ALLOCATED).size();
+        Pipeline.PipelineState.OPEN).size();
 
     // This value will be zero when pipeline count is 0.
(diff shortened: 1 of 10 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/safemode/HealthyPipelineSafeModeRule.java
```diff
@@ -118,7 +118,7 @@ protected void process(Pipeline pipeline) {
     Preconditions.checkNotNull(pipeline);
     if (pipeline.getType() == HddsProtos.ReplicationType.RATIS &&
         pipeline.getFactor() == HddsProtos.ReplicationFactor.THREE &&
-        pipeline.isHealthy() &&
+      //  pipeline.isHealthy() &&
```
bharatviswa504, line 121: do we want to remove this?

bharatviswa504: I understand now why we can remove this, we trigger open Pipeline event when healthy only. Can we remove the code comment completely.

bharatviswa504 (commented):
Unrelated file pnpm-lock.yaml need to be removed.
I have one minor comment, overall patch LGTM.
Can you look into failed CI tests?

bshashikant:
@bharatviswa504 @nandakumar131 can you please review?

bharatviswa504 (approved):
+1 LGTM

bharatviswa504:
Thank You @bshashikant for the contribution.

PR #1531, opened 2020-10-29
HDDS-4405. Proxy failover is logging with out trying all OMS.

bharatviswa504:

Update ProxyInfo with specific OM information which will help in skipping the retry log during the first time fail overs, tilll it finds leader OM.
Skip Retry INFO logging on the first failover from a proxy is broken. This fixes the behavior which is broken by HDDS-4292. For more info refer to this https://github.com/apache/ozone/pull/1531#issuecomment-718857999 link.

https://issues.apache.org/jira/browse/HDDS-4405

## How was this patch tested?

Tested it on the deployed ozone cluster.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/ha/OMFailoverProxyProvider.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/ha/OMFailoverProxyProvider.java
@@ -214,7 +214,7 @@ public class OMFailoverProxyProvider implements
     try {
       OzoneManagerProtocolPB proxy = createOMProxy(address);
       // Create proxyInfo here, to make it work with all Hadoop versions.
-      proxyInfo = new ProxyInfo<>(proxy, omProxyInfos.toString());
+      proxyInfo = new ProxyInfo<>(proxy, omProxyInfo.toString());
       omProxies.put(nodeId, proxyInfo);
     } catch (IOException ioe) {
       LOG.error("{} Failed to create RPC proxy to OM at {}",


umamaheswararao:
Thanks @bharatviswa504  for fixing it.
+1 LGTM. Could you please check the failures checks above once? Thanks

adoroszlai (commented):

I think there are 2 problems and I'm confused about which one is being fixed here:

1. Failover exception is shown in client unnecessarily (after trying 2 OMs, before trying last one).
2. Failover exception message shows list of all OMs (`submitRequest over nodeId=...`) instead of the single OM it contacted.

The PR description and issue title suggest that it is about problem 1, but the patch seems to fix problem 2 instead.

The failover exception is shown if client contacts OMs in "follower1, follower2, leader" order.  Which it does occasionally because client ignores "suggested leader" info and tries all OMs in order (as reported in HDDS-3936).

Steps to consistently reproduce problem 1:

1. start OM HA cluster, create volume and bucket
   ```
   ozone freon ockg -n1 -t1
   ```
2. note which OM is leader
   ```
   ozone admin om roles
   ```
3. override `ozone.om.nodes.<service>` in client, moving leader OM name to last position, eg.:
   ```
   ozone fs -Dozone.om.nodes.id1=om3,om1,om2 -ls o3fs://bucket1.vol1.id1/
   ```

bharatviswa504:
Thank You @umamaheswararao for the review.

hanishakoneru:
@bharatviswa504, can you elaborate on the issue and the fix in the description please. It is not clear why the issue occurred and what the fix is. Thanks.

bharatviswa504:
> I think there are 2 problems and I'm confused about which one is being fixed here:
> 
> Failover exception is shown in client unnecessarily (after trying 2 OMs, before trying last one).
> Failover exception message shows list of all OMs (submitRequest over nodeId=...) instead of the single OM it contacted.
> The PR description and issue title suggest that it is about problem 1, but the patch seems to fix problem 2 instead.
> 
> The failover exception is shown if client contacts OMs in "follower1, follower2, leader" order. Which it does occasionally because client ignores "suggested leader" info and tries all OMs in order (as reported in HDDS-3936).

Here the issue is each proxyInfo object should have information about their specific nodeID and IPAddress. During fixing of the Bug HDDS-4292 it has updated thee proxyInfo for each object with all proxyInfo(map of proxy info to String).

The fix for skipping log is done in HADOOP-17116 which went in 3.3.1 release and Apache Ozone depends on 3.2.1, to fix logging issue in Ozone we need a new release of Hadoop with HADOOP-17116. That is the reason we are still observing this in Apache Ozone.

Internally at Cloudera, we have backported HADOOP-17116, and HDDS-4292 broke this.

Having proxy info with each OM information, instead of all OM information in proxy info will fix the logging once we have the fix HADOOP-17116.





```
boolean info = true;
  |   | 397 | // If this is the first failover to this proxy, skip logging at INFO level
  |   | 398 | if (!failedAtLeastOnce.contains(proxyDescriptor.getProxyInfo().toString()))
  |   | 399 | {
  |   | 400 | failedAtLeastOnce.add(proxyDescriptor.getProxyInfo().toString());
  |   | 401 |  
  |   | 402 | // If successful calls were made to this proxy, log info even for first
  |   | 403 | // failover
  |   | 404 | info = hasSuccessfulCall \|\| asyncCallHandler.hasSuccessfulCall();
  |   | 405 | if (!info && !LOG.isDebugEnabled()) {
  |   | 406 | return;
  |   | 407 | }
399 | } | 408
```

So having each proxy info, and if the call fails, it will not be there in Map, we add to map and if there is no successfulCall wereturn with out logging.

adoroszlai:
> The fix for skipping log is done in HADOOP-17116

Thanks @bharatviswa504 for the detailed explanation.

umamaheswararao (approved):

hanishakoneru:
Thanks Bharat. +1.

bharatviswa504:
Thank You @umamaheswararao @adoroszlai and @hanishakoneru for the review.

PR #1532, opened 2020-10-29
HDDS-4406. Run MiniOzoneChaosCluster via github actions.

mukul1987:

## How was this patch tested?

### DIFF
--- a/.github/workflows/chaos.yml
+++ b/.github/workflows/chaos.yml
@@ -1,0 +1,59 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


elek (changes requested):
Thanks @mukul1987 the change. Overall it looks good to me.

One missing point might be the collection of available logs as artifacts.

To make it easier I would suggest following the existing convention and create a `hadoop-ozone/dev-support/checks/chaos.sh` which executes the chaos test and collects all require results under the "$REPORT_DIR". 

It would help us to be independent of github actions and execute chaos test in any other (external or internal) environments.

.github/workflows/chaos.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  build:
+    name: compile
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        java: [ 8, 11 ]
+      fail-fast: false
+    steps:
+      - name: Setup link to SSD
```
elek, line 30: Just curious: are you sure that root is not on SSD. I know that this secondary partition has more space, but didn't know that it has different hardware (SSD vs HDD).

adoroszlai: This comes from `post-commit.yml`, where I added this name based on my understanding of the [GitHub-hosted  runner specification](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources).  It says "14 GB of SSD disk space", which matches `/mnt`, but not the other mount, which has more total space but less free of it.

PR #1535, opened 2020-10-30
HDDS-4346.Ozone specific Trash Policy

sadanand48:
This defines an ozone specific TrashPolicy implementation. This current patch is almost a copy of [TrashPolicyDefault ](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/TrashPolicyDefault.java)but will be modified in further jira's to support a multithreaded Trash emptier.

https://issues.apache.org/jira/browse/HDDS-4346

## How was this patch tested?
Unit Test

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -27,6 +27,8 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathIsNotEmptyDirectoryException;
 import org.apache.hadoop.fs.Trash;
+import org.apache.hadoop.fs.TrashPolicy;
+import org.apache.hadoop.fs.TrashPolicyDefault;
 import org.apache.hadoop.fs.contract.ContractTestUtils;
 import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
@@ -41,6 +43,7 @@ import org.apache.hadoop.ozone.client.OzoneVolume;
 import org.apache.hadoop.ozone.client.VolumeArgs;
 import org.apache.hadoop.ozone.client.protocol.ClientProtocol;
 import org.apache.hadoop.ozone.om.OMConfigKeys;
+import org.apache.hadoop.ozone.om.TrashPolicyOzone;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer.ACLIdentityType;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer.ACLType;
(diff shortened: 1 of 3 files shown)


bshashikant (commented):
Can't we change this to OzoneTrashPolicyDefault which extends TrashPolicyDefault and override only those functions which needed to be changed specifically for ozone.

Copying the same code doesn't make sense here.

mukul1987 (commented):
Can we also make change in OzoneManager to call this new TrashEmptier ?

sadanand48:
> Can we also make change in OzoneManager to call this new TrashEmptier ?

Done.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -1184,6 +1189,10 @@ public void testTrash() throws Exception {
     try (FSDataOutputStream stream = fs.create(path)) {
       stream.write(1);
     }
+    Assert.assertTrue(trash.getConf().getClass(
+        "fs.trash.classname", TrashPolicyDefault.class).
```
mukul1987, line 1193: This should be 

trash.getConf().getClass(
        "fs.trash.classname", TrashPolicyOzone.class, TrashPolicy.class) ?

mukul1987: Referring to the following lines of code

https://github.com/apache/hadoop/blob/6f5a36c13c4d2eb24e207aa99a9781c0be11660b/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/TestTrash.java#L643-L642

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -0,0 +1,234 @@
...
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.TrashPolicyDefault;
+import org.apache.hadoop.fs.permission.FsAction;
+import org.apache.hadoop.fs.permission.FsPermission;
+import org.apache.hadoop.fs.FileAlreadyExistsException;
+import org.apache.hadoop.util.Time;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** TrashPolicy for Ozone Specific Trash Operations.Through this implementation
```
mukul1987, line 40: The first word should start from the next line.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -27,6 +27,8 @@
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathIsNotEmptyDirectoryException;
 import org.apache.hadoop.fs.Trash;
+import org.apache.hadoop.fs.TrashPolicy;
+import org.apache.hadoop.fs.TrashPolicyDefault;
```
mukul1987, line 31: Is this import needed ?

mukul1987 (changes requested):

prashantpogde (approved):
LGTM, once other comments are taken care of.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
sadanand48: done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
sadanand48: done

mukul1987 (approved):
+1, the changes looks good to me.

PR #1544, opened 2020-11-02
HDDS-4337. Implement RocksDB options cache for new datanode DB utilities.

errose28:
## What changes were proposed in this pull request

Add a cache of `ColumnFamilyOptions` objects to `AbstractDatanodeStore`, so that each `ConfigurationSource` object used to construct a container will result in the same RocksDB `ColumnFamilyOptions` object being used. Since the same configuration instance is used for all containers on a datanode, this results in all containers on a datanode using the same  `ColumnFamilyOptions`, and by extension the same RocksDB metadata cache. A configuration parameter is added for adjusting the size of this cache. Each RocksDB instance will have its own `DBOptions`, since this requires the use of RocksDB `Statistics` objects which are not thread safe and cannot be shared.

The `DBStoreBuilder` class was refactored to make configuration easier. It is now possible to use the same configurable column family options for all column families, and it is easier to reason about which configurations will be applied. It is no longer an error to register a table twice with the builder, the second call will simply override the first, which is usually the expected behavior in setters for builders. There should be no other functionality changes to this class.

HDDS-4337

### Automated Testing

- `DBStoreBuilder` unit tests were updated.
- A unit test was added to `TestKeyValueContainer` to test that `ColumnFamilyOptions` objects are shared between containers with the same configuration.

### Manual Testing

Manual testing was also done with a docker cluster to verify that the same cache is used for all containers on a datanode. Steps:

1. Start an ozone docker cluster and enter a datanode.
2. Write 100 keys: `ozone freon ockg -n 100`
3. Close the pipeline those keys were written to.
    - `ozone admin pipeline list` to get the pipeline IDs.
    - `ozone admin pipeline close <id>` to close the pipeline.
4. Write 100 more keys.
5. Get the address of the cache object from the RocksDB log for each of the two containers:
`grep -A5 "block_cache:" /data/hdds/hdds/*/current/containerDir0/1/metadata/1-dn-container.db/LOG`
and
`grep -A5 "block_cache:" /data/hdds/hdds/*/current/containerDir0/2/metadata/2-dn-container.db/LOG`

- Make sure that the address listed after `block_cache:` is the same for both containers.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -343,10 +343,10 @@ public final class OzoneConfigKeys {
   public static final double
       HDDS_DATANODE_STORAGE_UTILIZATION_CRITICAL_THRESHOLD_DEFAULT = 0.95;
 
+  public static final String HDDS_DATANODE_METADATA_ROCKSDB_CACHE_SIZE =
+      "hdds.datanode.metadata.rocksdb.cache.size";
   public static final String
-      HDDS_DATANODE_METADATA_CACHE_SIZE = "hdds.datanode.metadata.cache.size";
-  public static final String
-      HDDS_DATANODE_CACHE_SIZE_DEFAULT = "64MB";
+      HDDS_DATANODE_METADATA_ROCKSDB_CACHE_SIZE_DEFAULT = "64MB";
 
   public static final String OZONE_SECURITY_ENABLED_KEY =
       "ozone.security.enabled";
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -343,6 +343,11 @@ public final class OzoneConfigKeys {
   public static final double
       HDDS_DATANODE_STORAGE_UTILIZATION_CRITICAL_THRESHOLD_DEFAULT = 0.95;
 
+  public static final String
+      HDDS_DATANODE_METADATA_CACHE_SIZE = "hdds.datanode.metadata.cache.size";
+  public static final String
+      HDDS_DATANODE_CACHE_SIZE_DEFAULT = "64MB";
+
   public static final String OZONE_SECURITY_ENABLED_KEY =
       "ozone.security.enabled";
   public static final boolean OZONE_SECURITY_ENABLED_DEFAULT = false;
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1279,6 +1279,16 @@
   </property>
 
   <property>
+    <name>hdds.datanode.metadata.cache.size</name>
+    <value>64MB</value>
+    <tag>OZONE, DATANODE, MANAGEMENT</tag>
+    <description>
+        Size of the shared block metadata cache on each datanode. All
+        containers on a datanode will share this cache.
+    </description>
+  </property>
+
+  <property>
     <name>hdds.command.status.report.interval</name>
     <value>30s</value>
     <tag>OZONE, DATANODE, MANAGEMENT</tag>
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1279,12 +1279,12 @@
   </property>
 
   <property>
-    <name>hdds.datanode.metadata.cache.size</name>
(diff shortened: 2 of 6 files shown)


sodonnel:
If I understand this correctly, in AbstractDatanodeStore.start(), it pulls the default CF options from a static variable where they are setup on first use. This creates a single `LRUCache(CACHE_SIZE)` which is passed into the the DBStoreBuilder for all container RocksDBs. This means they should share the LRU cache.

Its a little hard to follow the logic to ensure this cache is definitely being reused. Could you create a docker cluster, use Freon to create some containers and then check the RocksDB log files to ensure the same cache instance is being used - There is an example of what to look for to prove this in https://issues.apache.org/jira/browse/HDDS-4246.

I also think the 8MB shared cache is probably too small. Could we add a configuration for it, and make the default higher, maybe 64MB for now? What do you think?

errose28:
Thanks, @sodonnel , I will add a setting for cache size and test this in a docker cluster. Looking at the rocks docs it appears those LRUCache objects can't be accidentally modified by callers, so I could also expose it as public and add a unit test to check that the same cache is used.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
```diff
@@ -343,6 +343,11 @@
   public static final double
       HDDS_DATANODE_STORAGE_UTILIZATION_CRITICAL_THRESHOLD_DEFAULT = 0.95;
 
+  public static final String
+      HDDS_DATANODE_METADATA_CACHE_SIZE = "hdds.datanode.metadata.cache.size";
```
sodonnel, line 347: I think a better name for this parameter might be:

"hdds.datanode.container.rocksdb.cache.size"

That more clearly says "this is the cache size for the container rocksDBs". What do you think?

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
```diff
@@ -343,6 +343,11 @@
...
+  public static final String
+      HDDS_DATANODE_CACHE_SIZE_DEFAULT = "64MB";
```
sodonnel, line 349: The usual pattern here is to have a pair of variables name like:

<PARAMETER>_KEY
<PARAMETER>_DEFAULT

I think it would be good to change this variable to HDDS_DATANODE_METADATA_CACHE_SIZE_DEFAULT.

I see some other parameters in this file don't end in "_KEY", so I guess it is not essential to rename HDDS_DATANODE_METADATA_CACHE_SIZE to "HDDS_DATANODE_METADATA_CACHE_SIZE_KEY".

If we change the parameter name to "hdds.datanode.container.rocksdb.cache.size" based on the above comment, maybe we should also rename the variables to HDDS_DATANODE_CONTAINER_ROCKSDB_CACHE_SIZE ?

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
errose28: My initial thought was that *metadata* tells what the cache is used for, which seemed more useful than *rocksdb*, which tells how the cache is implemented. But if this config is being changed, knowing that it is for a RocksDB instance might be useful. Maybe we could include both parts? For example, the existing config *ozone.metastore.rocksdb.statistics* contains both pieces of information.

I think putting *container* in the name makes the config sound like it is for one container, whereas just having *datanode* makes it more apparent that this is the size of one shared cache for the whole DN.

These would give a new name like *hdds.datanode.rocksdb.metadata.cache.size* or *hdds.datanode.metadata.store.rocksdb.cache.size* which are informative but fairly long. Since ozone-default.xml has a description, this may not be necessary.

I don't really have a strong opinion either way. If you think *hdds.datanode.container.rocksdb.cache.size* the best name I am fine with that.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
errose28: Yes, this was a typo. I forgot to update the name of the default variable when I updated the name. I will fix it once we decide on a name for the config.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
sodonnel: How about `hdds.datanode.metadata.rocksdb.cache.size`? I just dropped `store` from your suggestion.

errose28: Sounds good to me, I'll update the name.

sodonnel (approved):
These changes look good to me. I am +1 on them, so I will go ahead and commit.

PR #1552, opened 2020-11-03
HDDS-4404. Datanode can go OOM when a Recon or SCM Server is very slow in processing reports

smengcl:

`HeartbeatEndpointTask#call` should not resend reports to passive endpoints (Recon)

https://issues.apache.org/jira/browse/HDDS-4404

## How was this patch tested?

No new test added. Existing ones should do.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/endpoint/HeartbeatEndpointTask.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/endpoint/HeartbeatEndpointTask.java
@@ -147,8 +147,12 @@ public class HeartbeatEndpointTask
       rpcEndpoint.setLastSuccessfulHeartbeat(ZonedDateTime.now());
       rpcEndpoint.zeroMissedCount();
     } catch (IOException ex) {
-      // put back the reports which failed to be sent
-      putBackReports(requestBuilder);
+      // don't resend reports to recon as it could be down for days
+      // DN is expected to work fine without recon and not go OOM
+      if (!rpcEndpoint.isPassive()) {
+        // put back the reports which failed to be sent
+        putBackReports(requestBuilder);
+      }
 
       rpcEndpoint.logIfNeeded(ex);
     } finally {


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/endpoint/HeartbeatEndpointTask.java
```diff
@@ -147,8 +147,12 @@ public void setDatanodeDetailsProto(DatanodeDetailsProto
       rpcEndpoint.setLastSuccessfulHeartbeat(ZonedDateTime.now());
       rpcEndpoint.zeroMissedCount();
     } catch (IOException ex) {
-      // put back the reports which failed to be sent
-      putBackReports(requestBuilder);
+      // don't resend reports to recon as it could be down for days
+      // DN is expected to work fine without recon and not go OOM
```
GlenGeng, line 151: Why don't put back container actions and pipeline actions for SCM, or we have a legacy reason?

smengcl: @GlenGeng This change shouldn't affect SCM at all. Since SCM should be active (`rpcEndpoint.isPassive() == false`).

smengcl: Also I believe container actions and pipeline actions **are** put back on exception. cmiiw.

cc @nandakumar131  @lokeshj1703

GlenGeng: This is a concise fix. I am fine with it.
The question about container/pipeline Action won't be a blocker of merging this fix.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/endpoint/HeartbeatEndpointTask.java
```diff
@@ -147,8 +147,12 @@ public void setDatanodeDetailsProto(DatanodeDetailsProto
...
+      if (!rpcEndpoint.isPassive()) {
+        // put back the reports which failed to be sent
+        putBackReports(requestBuilder);
+      }
```
linyiqun, line 155: For the case of Recon is down , current change can fix this.
But for SCM is down, it could still lead OOM error I think since we put back reports again and again.
A better way I am thinking for this: if we can check from thrown exception to see if SCM/Recon is out of service, if yes, no need to put back report anymore.

smengcl: Thanks for the comment @linyiqun .

Yes. If SCM is down for a very long time, DN can still OOM due to the same reason as Recon. So this is, at best, a short-term fix.

> A better way I am thinking for this: if we can check from thrown exception to see if SCM/Recon is out of service, if yes, no need to put back report anymore.

@nandakumar131 mentioned in the [jira comment](https://issues.apache.org/jira/browse/HDDS-4404?focusedCommentId=17228389&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17228389) that `putBackReports()` serves the purpose of informing SCM of block deletion. In this case we don't want to just toss those reports away when SCM is down. Checking the exact type of exception (e.g. if it is a connection issue) is a good idea though.

I agree in the long run the fix could be to only queue the latest and necessary report.

linyiqun: As the quick fix for current issue, I'm okay for this change.

smengcl:
Closing this PR as the proper fix is being done in https://github.com/apache/ozone/pull/1601

PR #1553, opened 2020-11-03
HDDS-4431. Improve output message for key/bucket list command.

errose28:

Currently when a user which doesn't have permission to list on a bucket, they get the following message to stderr:

```
$ ozone sh key list o3://ozone1/volume1/bucket1
PERMISSION_DENIED org.apache.hadoop.ozone.om.exceptions.OMException: User  <user> doesn't have LIST permission to access bucket
```

The `org.apache.hadoop.ozone.om.exceptions.OMException:` should not be displayed in the error message to make it consistent with other PERMISSION_DENIED messages we get with other commands where we don't have this Exception class shown.
 
### Cause

The `Shell#printError` applies special formatting to `OMException` instances in order to render them as error messages to the user. The `OMException`s generated by ACL failure in list commands, however, are thrown from an `Iterator` interface implementation. This requires the exceptions to be unchecked to not break the interface spec, so methods like `ObjectStore#getNextListOfVolumes` chain any `IOExceptions` (including `OMException`) to a `RuntimeException` before throwing. When `Shell#printError` received this, it determined the exception was not an instance of `OMException`, and used the printing format for all other exception types, which displays the exception name with the message. `Shell#printError` has been updated to check both the derived class and cause of the `Throwable` it receives to determine whether it should convert it to the user error message format or not. 

HDDS-4431

## How was this patch tested?

Robot acceptance test updated with the expected output for list command on ACL failure.

### DIFF
--- a/hadoop-ozone/dist/src/main/smoketest/security/ozone-secure-fs.robot
+++ b/hadoop-ozone/dist/src/main/smoketest/security/ozone-secure-fs.robot
@@ -118,7 +118,7 @@ Test native authorizer
     Execute         kdestroy
     Run Keyword     Kinit test user     testuser2    testuser2.keytab
     ${result} =     Execute And Ignore Error         ozone sh bucket list /${volume3}/
-                    Should contain      ${result}    PERMISSION_DENIED org.apache.hadoop.ozone.om.exceptions.OMException: User testuser2/scm@EXAMPLE.COM doesn't have LIST permission to access volume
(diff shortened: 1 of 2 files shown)


adoroszlai (approved):
Thanks @errose28 for this improvement.

ayushtkn (approved):
LGTM

ayushtkn:
Thanx @errose28 for the contribution and @adoroszlai for the review!!!

PR #1554, opened 2020-11-04
HDDS-4420. Missing a container shouldn't log message at ERROR level

wycccccc:

According to the instructions I replace error to warn.

https://issues.apache.org/jira/browse/HDDS-4420

## How was this patch tested?

Ran UTs pass.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/IncrementalContainerReportHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/IncrementalContainerReportHandler.java
@@ -71,8 +71,10 @@ public class IncrementalContainerReportHandler extends
     for (ContainerReplicaProto replicaProto :
         report.getReport().getReportList()) {
       try {
+
         final ContainerID id = ContainerID.valueof(
             replicaProto.getContainerID());
+
         if (!replicaProto.getState().equals(
             ContainerReplicaProto.State.DELETED)) {
           nodeManager.addContainer(dd, id);
@@ -87,7 +89,7 @@ public class IncrementalContainerReportHandler extends
             report.getDatanodeDetails(), ex);
       } catch (IOException e) {
         success = false;
-        LOG.error("Exception while processing ICR for container {}",
+        LOG.warn("Exception while processing ICR for container {}",
             replicaProto.getContainerID(), e);
       }
     }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/IncrementalContainerReportHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/IncrementalContainerReportHandler.java
@@ -71,10 +71,8 @@ public class IncrementalContainerReportHandler extends
     for (ContainerReplicaProto replicaProto :
         report.getReport().getReportList()) {
       try {
-
         final ContainerID id = ContainerID.valueof(
             replicaProto.getContainerID());
-
         if (!replicaProto.getState().equals(
             ContainerReplicaProto.State.DELETED)) {
           nodeManager.addContainer(dd, id);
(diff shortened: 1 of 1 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/IncrementalContainerReportHandler.java
```diff
@@ -71,8 +71,10 @@ public void onMessage(final IncrementalContainerReportFromDatanode report,
     for (ContainerReplicaProto replicaProto :
         report.getReport().getReportList()) {
       try {
+
         final ContainerID id = ContainerID.valueof(
             replicaProto.getContainerID());
+
```
ayushtkn, line 77: Avoid the empty line changes here and above

ayushtkn (changes requested):

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/IncrementalContainerReportHandler.java
```diff
@@ -87,7 +89,7 @@ public void onMessage(final IncrementalContainerReportFromDatanode report,
             report.getDatanodeDetails(), ex);
       } catch (IOException e) {
         success = false;
-        LOG.error("Exception while processing ICR for container {}",
+        LOG.warn("Exception while processing ICR for container {}",
```
ayushtkn, line 92: I think rather than changing here, we should catch a separate `ContainerReplicaNotFoundException ` and there reduce the log level, suppressing for all IOE might not be a good choice.
@jojochuang does that make sense?

ayushtkn: @wycccccc can you address this as well

wycccccc: @ayushtkn I think what you said makes sense，let me try it.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/IncrementalContainerReportHandler.java
```diff
@@ -80,11 +80,16 @@ public void onMessage(final IncrementalContainerReportFromDatanode report,
         processContainerReplica(dd, replicaProto, publisher);
       } catch (ContainerNotFoundException e) {
         success = false;
-        LOG.warn("Container {} not found!", replicaProto.getContainerID());
+        LOG.warn("Container {} not found!",
+            replicaProto.getContainerID());
```
ayushtkn, line 84: @wycccccc This is just a formatting change. It isn't required here. Please avoid changing it, Apart from this changes Looks Good

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/IncrementalContainerReportHandler.java
```diff
@@ -80,11 +80,16 @@ public void onMessage(final IncrementalContainerReportFromDatanode report,
...
       } catch (NodeNotFoundException ex) {
         success = false;
         LOG.error("Received ICR from unknown datanode {}",
             report.getDatanodeDetails(), ex);
+      } catch (ContainerReplicaNotFoundException e){
+        success = false;
+        LOG.warn("Container {} replica not found!",
+            replicaProto.getContainerID());
```
cku328, line 92: Based on the CI logs generated by the previous commit (#5756e86), the CheckStyle tool only reported that this line didn't meet the character length limit (<=80). I think it is only necessary to adjust this part.
image

cku328 (commented):
I think it's a good fix that only reduces the log level of ContainerReplicaNotFoundException, not the whole IOException.

ayushtkn (approved):
+1
Can you share your jira id or assign HDDS-4420, if you are already added as contributor.
If not drop a comment at HDDS-4420, I will add you as contributor

wycccccc:
@ayushtkn I have already replied in JIRA, thank you.

ayushtkn:
Thanx have assigned it you.
If no comments from @jojochuang or @cku328 
Will merge this by tomorrow.

jojochuang (approved):
LGTM thanks all!

cku328 (approved):
Looks good to me. 👍 
Thanks @ayushtkn & @jojochuang  for the review.

ayushtkn:
Merged PR. Thanx Everyone!!!

PR #1555, opened 2020-11-06
HDDS-4347.Make Ozone specific Trash remover multi threaded

sadanand48:
This change makes the Trash Emptier multithreaded. i.e the main emptier thread starts a thread pool executor and executes tasks (checkpointing and deletion) present in the task queue. the parallelism is obtained at the  bucket level as fs.trashRoots for a root OFS URI returns a list of bucket level trashroots. 
https://issues.apache.org/jira/browse/HDDS-4347

## How was this patch tested?
Added unit tests and tested manually

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -55,7 +55,6 @@ import org.apache.hadoop.test.GenericTestUtils;
 import org.apache.commons.io.IOUtils;
 
 import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_TRASH_INTERVAL_KEY;
-import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_TRASH_CHECKPOINT_INTERVAL_KEY;
 import static org.apache.hadoop.fs.FileSystem.TRASH_PREFIX;
 import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_FS_ITERATE_BATCH_SIZE;
 import static org.apache.hadoop.ozone.OzoneConsts.OZONE_URI_DELIMITER;
@@ -258,6 +257,7 @@ public class TestOzoneFileSystem {
     OzoneConfiguration conf = new OzoneConfiguration();
     conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS,
         enabledFileSystemPaths);
+    conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
     cluster = MiniOzoneCluster.newBuilder(conf)
         .setNumDatanodes(3)
         .build();
@@ -275,8 +275,6 @@ public class TestOzoneFileSystem {
     conf.set(CommonConfigurationKeysPublic.FS_DEFAULT_NAME_KEY, rootPath);
     // Set the number of keys to be processed during batch operate.
     conf.setInt(OZONE_FS_ITERATE_BATCH_SIZE, 5);
-    conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
-    conf.setInt(FS_TRASH_CHECKPOINT_INTERVAL_KEY, 1);
     fs = FileSystem.get(conf);
     trash = new Trash(conf);
   }
@@ -796,8 +794,6 @@ public class TestOzoneFileSystem {
         "fs.trash.classname", TrashPolicy.class).
         isAssignableFrom(TrashPolicyOzone.class));
     Assert.assertEquals(trash.getConf().getInt(FS_TRASH_INTERVAL_KEY, 0), 1);
-    Assert.assertEquals(trash.getConf().getInt(FS_TRASH_CHECKPOINT_INTERVAL_KEY,
-        0), 1);
     // Call moveToTrash. We can't call protected fs.rename() directly
     trash.moveToTrash(path);
 
@@ -811,7 +807,7 @@ public class TestOzoneFileSystem {
     // Wait until the TrashEmptier purges the key
     GenericTestUtils.waitFor(()-> {
       try {
-        return !fs.exists(trashPath);
+        return !o3fs.exists(trashPath);
       } catch (IOException e) {
         LOG.error("Delete from Trash Failed");
         Assert.fail("Delete from Trash Failed");
@@ -825,7 +821,7 @@ public class TestOzoneFileSystem {
     // wait for deletion of checkpoint dir
     GenericTestUtils.waitFor(()-> {
       try {
-        return fs.listStatus(userTrash).length==0;
+        return o3fs.listStatus(userTrash).length==0;
       } catch (IOException e) {
         LOG.error("Delete from Trash Failed");
         Assert.fail("Delete from Trash Failed");
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -28,7 +28,16 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.TimeoutException;
 
-import org.apache.hadoop.fs.*;
+import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileAlreadyExistsException;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.InvalidPathException;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.Trash;
+import org.apache.hadoop.fs.TrashPolicy;
 import org.apache.hadoop.fs.contract.ContractTestUtils;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.ozone.MiniOzoneCluster;
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -28,15 +28,7 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.TimeoutException;
 
-import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
-import org.apache.hadoop.fs.FSDataInputStream;
-import org.apache.hadoop.fs.FSDataOutputStream;
-import org.apache.hadoop.fs.FileAlreadyExistsException;
-import org.apache.hadoop.fs.FileStatus;
-import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.InvalidPathException;
-import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.fs.Trash;
+import org.apache.hadoop.fs.*;
 import org.apache.hadoop.fs.contract.ContractTestUtils;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.ozone.MiniOzoneCluster;
@@ -45,6 +37,7 @@ import org.apache.hadoop.ozone.TestDataUtil;
 import org.apache.hadoop.ozone.client.OzoneBucket;
 import org.apache.hadoop.ozone.client.OzoneKeyDetails;
 import org.apache.hadoop.ozone.om.OMConfigKeys;
+import org.apache.hadoop.ozone.om.TrashPolicyOzone;
 import org.apache.hadoop.ozone.om.helpers.OmKeyArgs;
 import org.apache.hadoop.ozone.om.helpers.OpenKeySession;
 import org.apache.hadoop.security.UserGroupInformation;
@@ -53,6 +46,7 @@ import org.apache.hadoop.test.GenericTestUtils;
 import org.apache.commons.io.IOUtils;
 
 import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_TRASH_INTERVAL_KEY;
+import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_TRASH_CHECKPOINT_INTERVAL_KEY;
 import static org.apache.hadoop.fs.FileSystem.TRASH_PREFIX;
 import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_FS_ITERATE_BATCH_SIZE;
 import static org.apache.hadoop.ozone.OzoneConsts.OZONE_URI_DELIMITER;
@@ -239,6 +233,7 @@ public class TestOzoneFileSystem {
     testDeleteRoot();
 
     testRecursiveDelete();
+    testTrash();
   }
 
   @After
@@ -252,7 +247,6 @@ public class TestOzoneFileSystem {
   private void setupOzoneFileSystem()
       throws IOException, TimeoutException, InterruptedException {
     OzoneConfiguration conf = new OzoneConfiguration();
-    conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
     conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS,
         enabledFileSystemPaths);
     cluster = MiniOzoneCluster.newBuilder(conf)
@@ -272,6 +266,8 @@ public class TestOzoneFileSystem {
     conf.set(CommonConfigurationKeysPublic.FS_DEFAULT_NAME_KEY, rootPath);
     // Set the number of keys to be processed during batch operate.
     conf.setInt(OZONE_FS_ITERATE_BATCH_SIZE, 5);
+    conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
+    conf.setInt(FS_TRASH_CHECKPOINT_INTERVAL_KEY, 1);
     fs = FileSystem.get(conf);
     trash = new Trash(conf);
   }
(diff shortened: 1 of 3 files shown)


linyiqun:
@sadanand48 , current unit test cases only cover the basic functionality of trash delete behavior. But here we introduce many new setting, like trash interval, emptier interval. Can we add more test case to cover this?

In additional, will Ozone trash feature also support S3 semantic path later?
Like we accidentally delete one key /vol/bucket/key (not a ozfs path), then we can recover this key from a trash path.

I see currently we implement this only for FS semantic path.

sadanand48:
> @sadanand48 , current unit test cases only cover the basic functionality of trash delete behavior. But here we introduce many new setting, like trash interval, emptier interval. Can we add more test case to cover this?
> 
> In additional, will Ozone trash feature also support S3 semantic path later?
> Like we accidentally delete one key /vol/bucket/key (not a ozfs path), then we can recover this key from a trash path.
> 
> I see currently we implement this only for FS semantic path.

Thanks for the comment @linyiqun . Yes I will add more unit tests to this. About the trash support for the object store I don't think we'll be supporting atleast for now since s3 too doesn't support trash.

linyiqun:
>About the trash support for the object store I don't think we'll be supporting atleast for now since s3 too doesn't support trash.

Okay, get it,

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -0,0 +1,288 @@
...
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Collection;
+import java.util.Date;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.*;
```
bshashikant, line 37: Please remove the * imports

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -0,0 +1,288 @@
...
+    private long ceiling(long time, long interval) {
+      return floor(time, interval) + interval;
+    }
+    private long floor(long time, long interval) {
+      return (time / interval) * interval;
+    }
+
+  }
+
+  private void createCheckpoint(Path trashRoot, Date date) throws IOException {
```
bshashikant, line 207: Why do we need to copy these functions? Is it doing anything special with respect to ozone
/

sadanand48: Cant override these methods from TrashPolicyDefault since they have private scope in TrashPolicyDefault.so had to copy.

bshashikant: If you want to use the same functionality as already existing, there is no need to override.

bshashikant: Understood.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -206,19 +210,19 @@ private void checkInvalidPath(Path path) throws Exception {
...
-  @Test(timeout = 300_000)
+  @Test(timeout = 540_000)
   public void testFileSystem() throws Exception {
     setupOzoneFileSystem();
 
     testOzoneFsServiceLoader();
     o3fs = (OzoneFileSystem) fs;
 
-    testCreateFileShouldCheckExistenceOfDirWithSameName();
+  /*  testCreateFileShouldCheckExistenceOfDirWithSameName();
```
prashantpogde, line 220: why are these tests commented out ?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1228,17 +1238,61 @@ public void restart() throws IOException {
       // Allow OM to start as Http Server failure is not fatal.
       LOG.error("OM HttpServer failed to start.", ex);
     }
-
     omRpcServer.start();
+
     isOmRpcServerRunning = true;
 
+    if (isLeader()) {
```
prashantpogde, line 1245: this check should be inside run method in TrashPolicyOzone. I also do not see a problem if it was run on all OM nodes.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+
+    private long ceiling(long time, long interval) {
+      return floor(time, interval) + interval;
+    }
+    private long floor(long time, long interval) {
+      return (time / interval) * interval;
+    }
+
```
linyiqun, line 208: Can you add a shutdown method for executor thread pool? And we can shutdown executor thread firstly, then interrupted the emptier thread. This will be a better way to smoothly shutdown emptier thread

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -0,0 +1,292 @@
...
+          }
+        } catch (Exception e) {
+          LOG.warn("RuntimeException during Trash.Emptier.run(): ", e);
+        }
+      }
+      try {
+        fs.close();
+      } catch(IOException e) {
+        LOG.warn("Trash cannot close FileSystem: ", e);
+      }
```
linyiqun, line 198: Can we shutdown thread pool operation in a finally block? 

```java
        try {
             while (true) {
                 ....
             }
         } finally {
            shutdown executor pool.
         }
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -0,0 +1,292 @@
...
+              if (!trashRoot.isDirectory()) {
+                continue;
+              }
+              Runnable task = ()->{
+                try {
+                  TrashPolicyOzone trash = new TrashPolicyOzone(fs, conf);
+                  trash.deleteCheckpoint(trashRoot.getPath(), false);
+                  trash.createCheckpoint(trashRoot.getPath(),
+                          new Date(Time.now()));
+                } catch (IOException e) {
```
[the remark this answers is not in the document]
sadanand48: done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
sadanand48: done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -115,20 +131,24 @@ public void run() {
...
-                TrashPolicyOzone trash = new TrashPolicyOzone(fs, conf);
-                trash.deleteCheckpoint(trashRoot.getPath(), false);
-                trash.createCheckpoint(trashRoot.getPath(), new Date(now));
-              } catch (IOException e) {
-                LOG.warn("Trash caught: "+e+". Skipping " +
-                    trashRoot.getPath() + ".");
-              }
+              Runnable task = ()->{
+                try {
+                  TrashPolicyOzone trash = new TrashPolicyOzone(fs, conf);
```
mukul1987, line 143: This should be outside the runnable function.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -115,20 +131,24 @@ public void run() {
           now = Time.now();
           if (now >= end) {
             Collection<FileStatus> trashRoots;
-            trashRoots = fs.getTrashRoots(true);      // list all trash dirs
-
-            for (FileStatus trashRoot : trashRoots) {   // dump each trash
+            trashRoots = fs.getTrashRoots(true); // list all trash dirs
+            LOG.info("TrashrootSize: " + trashRoots.size());
+            for (FileStatus trashRoot : trashRoots) {  // dump each trash
+              LOG.info("Trashroot:" + trashRoot.getPath().toString());
```
mukul1987, line 137: Same as above.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -115,20 +131,24 @@ public void run() {
           now = Time.now();
           if (now >= end) {
             Collection<FileStatus> trashRoots;
-            trashRoots = fs.getTrashRoots(true);      // list all trash dirs
-
-            for (FileStatus trashRoot : trashRoots) {   // dump each trash
+            trashRoots = fs.getTrashRoots(true); // list all trash dirs
+            LOG.info("TrashrootSize: " + trashRoots.size());
```
mukul1987, line 135: Please change the LOG.info to LOG.debug

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -139,9 +159,12 @@ public void run() {
         fs.close();
       } catch(IOException e) {
         LOG.warn("Trash cannot close FileSystem: ", e);
+      } finally {
+        executor.shutdown();
```
mukul1987, line 163: We also need to do executor.awaitTermination as well here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -49,6 +53,8 @@
 
   private static final Path CURRENT = new Path("Current");
 
+  private final static int TRASH_EMPTIER_CORE_POOL_SIZE = 5;
```
mukul1987, line 56: Please change this to great from a config.

mukul1987 (changes requested):

prashantpogde (approved):

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -776,4 +784,59 @@ public void testRenameToTrashEnabled() throws Exception {
...
+
+  public void testTrash() throws Exception {
+    String testKeyName = "testKey2";
+    Path path = new Path(OZONE_URI_DELIMITER, testKeyName);
+    ContractTestUtils.touch(fs, path);
+    Assert.assertTrue(trash.getConf().getClass(
+        "fs.trash.classname", TrashPolicy.class).
+        isAssignableFrom(TrashPolicyOzone.class));
+    Assert.assertEquals(trash.getConf().getInt("fs.trash.interval", 0), 1);
+    Assert.assertEquals(trash.getConf().getInt("fs.trash.checkpoint.interval",
```
mukul1987, line 801: Lets use CommonConfigurationKeysPublic#FS_TRASH_CHECKPOINT_INTERVAL_KEY here

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -776,4 +784,59 @@ public void testRenameToTrashEnabled() throws Exception {
...
+   */
+
+  public void testTrash() throws Exception {
+    String testKeyName = "testKey2";
+    Path path = new Path(OZONE_URI_DELIMITER, testKeyName);
+    ContractTestUtils.touch(fs, path);
+    Assert.assertTrue(trash.getConf().getClass(
+        "fs.trash.classname", TrashPolicy.class).
+        isAssignableFrom(TrashPolicyOzone.class));
+    Assert.assertEquals(trash.getConf().getInt("fs.trash.interval", 0), 1);
```
mukul1987, line 800: Lets use CommonConfigurationKeysPublic#FS_TRASH_INTERVAL_KEY here

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1275,7 +1275,7 @@ private void startTrashEmptier(Configuration conf) throws IOException {
 
     // configuration for the FS instance that  points to a root OFS uri.
     // This will ensure that it will cover all volumes and buckets
-    Configuration fsconf = new Configuration();
+    Configuration fsconf = new OzoneConfiguration();
```
mukul1987, line 1278: I think this is intentional. @elek @bharatviswa504 can you please confirm this line ?

mukul1987 (changes requested):

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
sadanand48: done

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
sadanand48: done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
bharatviswa504: From my understanding, we need to use OzoneConfiguration, as OzoneConfiguration activates ozone-default.xml

  public static void activate() {
    // adds the default resources
    Configuration.addDefaultResource("hdfs-default.xml");
    Configuration.addDefaultResource("hdfs-site.xml");
    Configuration.addDefaultResource("ozone-default.xml");
  }

By using Configuration, defaults of Ozone will not be loaded. And also as we have for each config, we have defaults defined in Java code that should also be fine.

Using OzoneConfiguration is better IMHO.

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -1907,6 +1907,15 @@
     </description>
   </property>
 
+  <property>
```
elek, line 1910: Can you please use Java, annotation based configuration?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -115,20 +128,23 @@ public void run() {
           now = Time.now();
           if (now >= end) {
             Collection<FileStatus> trashRoots;
-            trashRoots = fs.getTrashRoots(true);      // list all trash dirs
-
-            for (FileStatus trashRoot : trashRoots) {   // dump each trash
+            trashRoots = fs.getTrashRoots(true); // list all trash dirs
```
elek, line 131: I know it's committed in the previous commit, but can you please help me to understand why do we need FileSystem in OM side? As far as I understood the design doc has more cons against this approach.

elek (changes requested):
Thanks for the work @sadanand48 

I have a minor suggestion about config handling.

I also feel that there is no clear agreement about the followed approach (at least I didn't find the sign in the public records of the community, but I have seen a question similar to mine: why do we need FileSystem in OM side.). 

I suggested in HDDS-3367 to answer to the thread on the mailing list with clarifying the chosen direction, but didn't get any answer yet. It can be useful to have everybody informed and up-to-date.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
mukul1987: thx @bharatviswa504

mukul1987 (approved):

hadoop-hdds/common/src/main/resources/ozone-default.xml
mukul1987: This is done in the latest changeset

mukul1987:
> Thanks for the work @sadanand48
> 
> I have a minor suggestion about config handling.
> 
> I also feel that there is no clear agreement about the followed approach (at least I didn't find the sign in the public records of the community, but I have seen a question similar to mine: why do we need FileSystem in OM side.).
> 
> I suggested in [HDDS-3367](https://issues.apache.org/jira/browse/HDDS-3367) to answer to the thread on the mailing list with clarifying the chosen direction, but didn't get any answer yet. It can be useful to have everybody informed and up-to-date.

The question has been answered on HDDS-3367. The current patch looks good to me. Resolving this comment for now.

mukul1987:
@sadanand48 , the test failures seems related, TestOzoneFileSystem#testTrash is failing in the latest run.

mukul1987:
All the checks are green. merging this change.

elek:
>>  I suggested in HDDS-3367 to answer to the thread on the mailing list with clarifying the chosen direction, but didn't get any answer yet. It can be useful to have everybody informed and up-to-date.

> The question has been answered on HDDS-3367. The current patch looks good to me. Resolving this comment for now.

I might miss but couldn't find the update on the mailing list.

Would you be so kind to update this thread:

https://lists.apache.org/thread.html/r9f6c11546a7a4792aadf6ea4c1b1bb4b7680eeb5406162e155d3c58b%40%3Cdev.ozone.apache.org%3E

PR #1559, opened 2020-11-07
HDDS-4435. Fix can not insert the second entry when hive on ozone

runzhiwang:
**What's the problem ?**
When use hive-2.3.7 on ozone, I can not insert the second entry to table, and exception thrown from ozone.
image

**What's the reason ?**
The following [hive code](https://github.com/apache/hive/blob/rel/release-2.3.7/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L3046) will call ozone's `rename(src, dst)` method, but if the last file of src and dst is same, ozone will throw exception.
image

**How to fix ?**
Then I tried hive on hdfs, I find hdfs's rename in this case will not throw exception, it just [return false](https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSDirRenameOp.java#L186), then hive will change the dst path at `destFilePath =  new Path(destDirPath, name + ("_copy_" + counter) + (!type.isEmpty() ? "." + type : ""));`. So we should keep ozone and hdfs's behavior the same, so that hive can run on ozone.

https://issues.apache.org/jira/browse/HDDS-4435

## How was this patch tested?

Use hive-2.3.7 on ozone,  insert the second entry to table successfully.

image

### DIFF
--- a/hadoop-ozone/integration-test/src/test/resources/contract/ozone.xml
+++ b/hadoop-ozone/integration-test/src/test/resources/contract/ozone.xml
@@ -52,6 +52,11 @@
     </property>
 
     <property>
+        <name>fs.contract.rename-returns-false-if-dest-exists</name>
+        <value>true</value>
+    </property>
+
+    <property>
         <name>fs.contract.rename-remove-dest-if-empty-dir</name>
         <value>false</value>
     </property>
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -25,7 +25,6 @@ import org.apache.hadoop.fs.BlockLocation;
 import org.apache.hadoop.fs.CreateFlag;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
-import org.apache.hadoop.fs.FileAlreadyExistsException;
 import org.apache.hadoop.fs.FileChecksum;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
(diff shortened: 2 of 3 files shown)


ayushtkn:
This problem is there in `BasicOzoneFileSystem`, we can correct it there as well.

runzhiwang:
@ayushtkn Thanks for review. I have updated the patch.

maobaolong (approved):
+1. Thanks for this improve, we should keep consistent behavior with hdfs

ayushtkn (approved):
+1

maobaolong:
As there are no objection comment, I will merge this PR tomorrow.
@runzhiwang Thanks for your contribution.
@ayushtkn Thanks for your review.

adoroszlai (changes requested):
Can you please change Ozone's definition for contract tests, ie. add

```
    <property>
        <name>fs.contract.rename-returns-false-if-dest-exists</name>
        <value>true</value>
    </property>
```

to hadoop-ozone/integration-test/src/test/resources/contract/ozone.xml?

I have reported [HADOOP-17365](https://issues.apache.org/jira/browse/HADOOP-17365) because I think the test is too loose.  It should have failed when changing the implementation without updating the contract definition accordingly.

runzhiwang:
@adoroszlai Thanks for suggestions. I have updated the patch. Could you help review it again ?

adoroszlai (approved):

adoroszlai:
> ## How was this patch tested?
> No need to add new ut.

@runzhiwang: This part of the PR template is not specific to unit tests.  If you tested the change with Hive to confirm that the problem is fixed, this is the place to explain it (ideally with steps).  Thanks.

runzhiwang:
> > ## How was this patch tested?
> > No need to add new ut.
> 
> @runzhiwang: This part of the PR template is not specific to unit tests. If you tested the change with Hive to confirm that the problem is fixed, this is the place to explain it (ideally with steps). Thanks.

@adoroszlai Thanks, I have updated it.

PR #1560, opened 2020-11-07
HDDS-4436.TestReplicationManager#testUnderReplicatedQuasiClosedContainerWithUnhealthyReplica fails intermittently.

ayushtkn:

Test Fix

https://issues.apache.org/jira/browse/HDDS-4436

## How was this patch tested?

Ran test multiple times locally.


adoroszlai (approved):
Thanks @ayushtkn for improving this.  Test [passed 200x](https://github.com/adoroszlai/hadoop-ozone/runs/1375517680#step:4:4).  I think this should be OK, although I have no data about failure rate before the fix.

ayushtkn:
Thanx @adoroszlai for the review!!!

PR #1563, opened 2020-11-10
HDDS-4445. Owner info is not passed to authorizer for BUCKET/KEY crea…

xiaoyuyao:
skip passing volume owner info only for volume creation(admin only) or root access. For bucket/key create, we should populate the volume owner to the authorizer. 

https://issues.apache.org/jira/browse/HDDS-4445

## How was this patch tested?
New acceptance test added for native authorizer. 
Tested with Ranger authorizer in real cluster.


arp7 (approved):
+1

xiaoyuyao:
Thanks @arp7 for the Review. I've merged the PR to master.

PR #1569, opened 2020-11-10
HDDS-4448. Duplicate refreshPipeline in listStatus

adoroszlai:

Currently `KeyManagerImpl#listStatus` issues duplicate `refreshPipeline` for each file.  HDDS-3824 moved `refreshPipeline` outside the bucket lock.  But HDDS-3658 added it back, while keeping the one outside the lock, probably as a result of merge conflict resolution.

This PR removes the `refreshPipeline` call which is made while holding the bucket lock.  It also converts the remaining one to batched (single call with list, instead of N times with single OM key info).

https://issues.apache.org/jira/browse/HDDS-4448

## How was this patch tested?

Added unit test to verify that a single batched `refreshPipeline` call is made from `listStatus`.

https://github.com/adoroszlai/hadoop-ozone/runs/1381172579

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -2192,9 +2192,6 @@ public class KeyManagerImpl implements KeyManager {
         // No need to check if a key is deleted or not here, this is handled
         // when adding entries to cacheKeyMap from DB.
         OzoneFileStatus fileStatus = entry.getValue();
-        if (fileStatus.isFile()) {
-          refreshPipeline(fileStatus.getKeyInfo());
-        }
         fileStatusList.add(fileStatus);
         countEntries++;
         if (countEntries >= numEntries) {
@@ -2209,14 +2206,20 @@ public class KeyManagerImpl implements KeyManager {
           bucketName);
     }
 
-    for (OzoneFileStatus fileStatus : fileStatusList) {
-      if (args.getRefreshPipeline()) {
-        refreshPipeline(fileStatus.getKeyInfo());
+    if (args.getRefreshPipeline()) {
+      List<OmKeyInfo> keyInfoList = new ArrayList<>(fileStatusList.size());
+      for (OzoneFileStatus fileStatus : fileStatusList) {
+        keyInfoList.add(fileStatus.getKeyInfo());
       }
-      if (args.getSortDatanodes()) {
+      refreshPipeline(keyInfoList);
+    }
+
+    if (args.getSortDatanodes()) {
+      for (OzoneFileStatus fileStatus : fileStatusList) {
         sortDatanodeInPipeline(fileStatus.getKeyInfo(), clientAddress);
       }
     }
+
     return fileStatusList;
   }
 
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
@@ -2188,10 +2188,9 @@ public class KeyManagerImpl implements KeyManager {
 
       countEntries = 0;
       // Convert results in cacheKeyMap to List
-      for (Map.Entry<String, OzoneFileStatus> entry : cacheKeyMap.entrySet()) {
+      for (OzoneFileStatus fileStatus : cacheKeyMap.values()) {
         // No need to check if a key is deleted or not here, this is handled
         // when adding entries to cacheKeyMap from DB.
-        OzoneFileStatus fileStatus = entry.getValue();
         fileStatusList.add(fileStatus);
         countEntries++;
         if (countEntries >= numEntries) {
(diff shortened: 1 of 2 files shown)


linyiqun (approved):
Good catch, +1.
@adoroszlai , I notice one thing that if we can completely remove field OmKeyArgs#refreshPipeline after HDDS-3658 was merged.

In some place, this option doesn't really make sense now.
Like KeyManagerImpl#getOzoneFileStatus:
```java
  private OzoneFileStatus getOzoneFileStatus(String volumeName,
                                             String bucketName,
                                             String keyName,
                                             boolean refreshPipeline,
                                             boolean sortDatanodes,
                                             String clientAddress)
      throws IOException {
    OmKeyInfo fileKeyInfo = null;
    metadataManager.getLock().acquireReadLock(BUCKET_LOCK, volumeName,
        bucketName);
    ...

      // if the key is a file then do refresh pipeline info in OM by asking SCM
      if (fileKeyInfo != null) {
        // refreshPipeline flag check has been removed as part of
        // https://issues.apache.org/jira/browse/HDDS-3658.
        // Please refer this jira for more details.
        refreshPipeline(fileKeyInfo);
        if (sortDatanodes) {
          sortDatanodeInPipeline(fileKeyInfo, clientAddress);
        }
        return new OzoneFileStatus(fileKeyInfo, scmBlockSize, false);
      }
    }
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2209,14 +2205,18 @@ private void listStatusFindKeyInTableCache(
...
-        refreshPipeline(fileStatus.getKeyInfo());
-      }
-      if (args.getSortDatanodes()) {
+      keyInfoList.add(fileStatus.getKeyInfo());
+    }
+    refreshPipeline(keyInfoList);
+
+    if (args.getSortDatanodes()) {
+      for (OzoneFileStatus fileStatus : fileStatusList) {
         sortDatanodeInPipeline(fileStatus.getKeyInfo(), clientAddress);
```
bharatviswa504, line 2216: This call still makes number of entries in listStatus call to put nodes in sorted order in pipeline when ozone.network.topology.aware.read is set. (Default false, so this will not be a problem, but we have still the problem of making number of entries in listStatus calls to SCM.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2209,14 +2205,18 @@ private void listStatusFindKeyInTableCache(
...
 
+    List<OmKeyInfo> keyInfoList = new ArrayList<>(fileStatusList.size());
     for (OzoneFileStatus fileStatus : fileStatusList) {
-      if (args.getRefreshPipeline()) {
-        refreshPipeline(fileStatus.getKeyInfo());
-      }
-      if (args.getSortDatanodes()) {
+      keyInfoList.add(fileStatus.getKeyInfo());
+    }
+    refreshPipeline(keyInfoList);
```
bharatviswa504, line 2212: Just a question listStatus is limited by batch size.
But we are making a single Rpc Call to SCM for all the containers we got from this list, is this fine here? (My question is if it has a large number of containerIDs due to larger key sizes, will it have an issue in Rpc Response size.

bharatviswa504 (commented):
Overall LGTM, I have few comments/questions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
adoroszlai: `listKeys()` is also limited by batch size and already does the same single refresh call:

https://github.com/apache/ozone/blob/d0aa34c4afae21538c6c6225f029c1d1c4c4bafd/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L931-L933

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
adoroszlai: Thanks for highlighting this further possible improvement.  I'll file a separate Jira issue for this, as I think it's out of scope here.

adoroszlai: https://issues.apache.org/jira/browse/HDDS-4473

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
bharatviswa504: My question is keyList is limited by batch size, but if Keys are huge size, then it might have huge containerIDs, and then calling SCM with all those containerIDS to fetch ContainerWithPipeline, will it cause an issue.

Then this might be a problem for listKeys also, but do you see this as a problem?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
bharatviswa504: Fine with me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
adoroszlai: > Then this might be a problem for listKeys also, but do you see this as a problem?

I don't see this as an immediate problem.

As far as I understand, batch size for `listKeys` and `listStatus` is a convenience for the client, not a safety guarantuee for the server.  If RPC response size is a problem when performing `getContainerWithPipelineBatch` call for multiple keys, then the same can be triggered by the client simply by increasing batch size (if there are enough keys in the bucket).

bharatviswa504: Ya agreed, not only that thinking more if with normall batch size it is causing issue for containerID's, then it will be same for OMResponse for listKeys also, as it adds KeyInfo and Pipeline.

bharatviswa504 (approved):
+1 LGTM

adoroszlai:
Thanks @linyiqun for the review, and @bharatviswa504 for review and merge.

PR #1570, opened 2020-11-11
HDDS-4443. Recon: Using Mysql database throws exception and fails startup

vivekratnavel:

Fix length of the primary key in one of the SQL tables of Recon and set auto-commit flag to default to true.

https://issues.apache.org/jira/browse/HDDS-4443

## How was this patch tested?

Tested the changes in a real cluster by replacing recon jar and using mariadb connector jar and jdbc URL configs for recon.

### DIFF
--- a/hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/ReconSqlDbConfig.java
+++ b/hadoop-ozone/recon-codegen/src/main/java/org/hadoop/ozone/recon/codegen/ReconSqlDbConfig.java
@@ -97,7 +97,7 @@ public class ReconSqlDbConfig {
 
   @Config(key = "auto.commit",
       type = ConfigType.BOOLEAN,
-      defaultValue = "false",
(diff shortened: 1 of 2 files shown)


swagle (approved):

avijayanhwx:
Thanks for working on this @vivekratnavel. 

Is it possible to add Mysql based embedded unit tests to org.apache.hadoop.ozone.recon.persistence.TestReconWithDifferentSqlDBs?  Maybe something like https://stackoverflow.com/questions/6734410/is-there-a-way-to-run-mysql-in-memory-for-junit-test-cases. I am OK with adding it through a different JIRA.

vivekratnavel:
Thanks for the reviews @swagle and @avijayanhwx! 

Good suggestion @avijayanhwx. I will try the embedded MySQL unit test in a different jira - https://issues.apache.org/jira/browse/HDDS-4457

avijayanhwx (approved):

PR #1571, opened 2020-11-11
HDDS-4441. Add metrics for ACL related operations.

ayushtkn:

Add acl related metrics in OM

https://issues.apache.org/jira/browse/HDDS-4441

## How was this patch tested?

Added  UT

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOmMetrics.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOmMetrics.java
@@ -16,6 +16,9 @@
  */
 package org.apache.hadoop.ozone.om;
 
+import static org.apache.hadoop.ozone.OzoneAcl.AclScope.ACCESS;
+import static org.apache.hadoop.ozone.security.acl.OzoneObj.ResourceType.VOLUME;
+import static org.apache.hadoop.ozone.security.acl.OzoneObj.StoreType.OZONE;
 import static org.apache.hadoop.test.MetricsAsserts.assertCounter;
 import static org.apache.hadoop.test.MetricsAsserts.getMetrics;
 import static org.mockito.Matchers.any;
@@ -24,6 +27,7 @@ import static org.mockito.Matchers.anyLong;
 
 import java.io.IOException;
 import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.hadoop.hdds.client.BlockID;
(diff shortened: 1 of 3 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOmMetrics.java
```diff
@@ -318,6 +326,42 @@ public void testKeyOps() throws IOException {
...
+      assertCounter("NumAddAcl", 1L, omMetrics);
+
+      // Test setAcl
+      ozoneManager.setAcl(volObj, acls);
+      omMetrics = getMetrics("OMMetrics");
+      assertCounter("NumSetAcl", 1L, omMetrics);
+
+      // Test removeAcl
+      ozoneManager.removeAcl(volObj, acls.get(0));
+      omMetrics = getMetrics("OMMetrics");
```
jojochuang, line 357: I wonder if it is really necessary to call this again and again. Doesn't look needed.

ayushtkn: Well, I also thought the same, it should be required only once, But my test itself fails without that. If you observe the tests above. `testVolumeOps`, `testBucketOps` and `testKeyOps` they too are also calling ``getMetrics("OMMetrics");`` before asserting.
The `getMetrics` seems to fetch the values and keep it. it doesn't seems to be updating.

jojochuang: ok thx for the clarification

jojochuang (approved):

ayushtkn:
Thanx @jojochuang for the review!!!

cxorm:
Thanks @ayushtkn for the contribution.

IMHO the `setAcl`, `removeAcl` and `addAcl` are write-operation, the metrics should be updated in `OMRequest` instead of `OzoneManager`. 
(For example, when we set acl of some volume, we should update the metric in `OMVolumeSetAclRequest`)

Could you be so kind as to file a jira to update it ? (Feel free to correct me if I miss something, thanks)

ayushtkn:
Thanx @cxorm I will raise one jira for that. But I didn't actually catch what to do.
`createDirectory`, `createFile` are also write calls, for them the metrics are also updated in OM. 
I couldn't find any metrics instance itself in `OMVolumeSetAclRequest` and why we can't update the metrics in OM?

cxorm:
> `createDirectory`, `createFile` are also write calls, for them the metrics are also updated in OM.
> I couldn't find any metrics instance itself in `OMVolumeSetAclRequest` and why we can't update the metrics in OM?

Yeah it's a good question, 
for OMHA,  actually the write-operation of OzoneManager is processed by `OMRequest` instead of `OzoneManager`.
(The write calls in `OzoneManager` is deprecated, but the code is not removed.) 

> And the metrics being updated is `OMMetrics`, So, that should track statistics of OM only?

Right, it is used to track statistics of OM (But the write calls of OM is processed by OMRequest.)

ayushtkn:
Ohh, Means for HA, the code flow is separate. I thought `OzoneManager` like the `NamenodeRpcServer`. It should be mentioned somewhere it is only for Non-HA :-(
Well, Thanx @cxorm for the explanation. I have tried to handle this for HA, in an Addendum PR - https://github.com/apache/ozone/pull/1584

Please have a look

bharatviswa504:
For HA/non-HA new code is being used for write requests.

cxorm:
Thanks @bharatviswa504  , and yes I mean non-HA/ HA is using new code.

PR #1574, opened 2020-11-11
HDDS-4453. Replicate closed container from random selected datanode

elek:

Che-Wei Chang reported a very particular case related to the data replication:

When a container is downloaded successfully but the content is invalid, the error the retry mechanism doesn't work very well. Downloader reports the download successful therefore it won't be reported as failure, but import will fail again and again.

To make it more resilient we can select datanode randomly,

https://issues.apache.org/jira/browse/HDDS-4453

## How was this patch tested?

Unit test is added.


sodonnel (approved):
Thanks for working on this. The changes LGTM. +1.

nandakumar131 (approved):

PR #1580, opened 2020-11-12
HDDS-1880. Decommission and Maintenance mode in Ozone

sodonnel:

This change adds the decommission and maintenance feature for Ozone, allowing datanodes to be taken offline in a controlled fashion.

There will be a vote thread on the mailing list to discuss and agree about merging this change.

https://issues.apache.org/jira/browse/HDDS-1880

## How was this patch tested?

New and existing tests added.


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/StartDatanodeAdminHandler.java
```diff
@@ -0,0 +1,68 @@
...
+                        EventPublisher publisher) {
+    Set<PipelineID> pipelineIds =
+        nodeManager.getPipelines(datanodeDetails);
+    LOG.info("Admin start on datanode {}. Finalizing its pipelines {}",
+        datanodeDetails, pipelineIds);
+    for (PipelineID pipelineID : pipelineIds) {
+      try {
+        Pipeline pipeline = pipelineManager.getPipeline(pipelineID);
+        pipelineManager.finalizeAndDestroyPipeline(pipeline, false);
+      } catch (IOException e) {
```
linyiqun, line 62: I see the pipeline info will be removed (DestroyPipeline)in SCM. Client reads container block data depending on this pipeline info I think.
So when multiple nodes be in decommissioned, client won't read the data temporarily because of no valid container pipeline returned, right?

Not sure if below steps will be a better way:
1)finalize dn pipeline
2)wait for dn decommission finished
3)destroy dn pipeline

linyiqun (commented):
Hi @sodonnel , there is one major concern below that I also mentioned in discussion mail thread.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/StartDatanodeAdminHandler.java
linyiqun: Revisit this logic again, SCM will reconstruct pipeline based on ContainerReplica once it finds its pipeline be destroyed.

Please ignore my previous comment, : ).

sodonnel:
It seems like the test failures were caused by HDDS-4453 and fixed in HDDS-4429. I have rebased to include HDDS-4429. Once I get a few full CI runs without failure we should be good to merge the branch.

elek:
# 20202f6  is failed with TestOzoneClientRetriesOnException. The same error is visible on master in https://github.com/apache/ozone/runs/1475898756?check_suite_focus=true and https://github.com/apache/ozone/runs/1478924771?check_suite_focus=true

#cc5eebe is failed only with coverage calculations

sodonnel:
@elek The failures now seem to be for a different reason each time. The issues caused by HDDS-4453 have not reappeared since merging HDDS-4429. Latest run seems to have hit a problem in the acceptance / smoketests which have been stable till now.

I will try to get another green run and then we can decide if its stable enough to merge.

sodonnel:
This PR has been merged manually via the CLI to preserve the history. Therefore I am closing this PR.

PR #1582, opened 2020-11-12
HDDS-4235. Ozone client FS path validation is not present in OFS.

ayushtkn:

Added path name validation in OFS

https://issues.apache.org/jira/browse/HDDS-4235

## How was this patch tested?

Added UT

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -1230,15 +1230,7 @@ public class TestRootedOzoneFileSystem {
   }
 
   private void checkInvalidPath(Path path) throws Exception {
-    FSDataOutputStream outputStream = null;
-    try {
-      LambdaTestUtils.intercept(InvalidPathException.class, "Invalid path Name",
-          () -> fs.create(path, false));
-    } finally {
-      if (outputStream != null) {
-        outputStream.close();
-      }
-    }
+    LambdaTestUtils.intercept(InvalidPathException.class, "Invalid path Name",
+        () -> fs.create(path, false));
   }
-
 }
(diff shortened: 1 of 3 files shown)


hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -840,8 +842,12 @@ public String pathToKey(Path path) {
     if (!path.isAbsolute()) {
       path = new Path(workingDir, path);
     }
+    String key = path.toUri().getPath();
+    if (!OzoneFSUtils.isValidName(path.toUri().getPath())) {
```
adoroszlai, line 846: Avoid unnecessary duplicate conversion:

```suggestion
    if (!OzoneFSUtils.isValidName(key)) {
```

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -817,7 +817,7 @@ public String pathToKey(Path path) {
     if (OzoneFSUtils.isValidName(key)) {
       key = path.toUri().getPath();
     } else {
```
adoroszlai, line 819: Not introduced in this patch, but would be nice to remove duplicate conversion:

```suggestion
    if (!OzoneFSUtils.isValidName(key)) {
```

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -1210,4 +1212,33 @@ public void testTrash() throws Exception {
...
+    checkInvalidPath(file1);
+  }
+
+  private void checkInvalidPath(Path path) throws Exception {
+    FSDataOutputStream outputStream = null;
+    try {
+      LambdaTestUtils.intercept(InvalidPathException.class, "Invalid path Name",
+          () -> fs.create(path, false));
+    } finally {
+      if (outputStream != null) {
```
adoroszlai, line 1238: Seems like `outputStream` is never updated to non-`null` value, hence unnecessary.

adoroszlai (changes requested):
Thanks @ayushtkn for taking care of this.

ayushtkn:
Thanx @adoroszlai for the review, have addressed the above comments

adoroszlai (approved):
Thanks @ayushtkn for updating the patch.

ayushtkn:
Thanx @adoroszlai for the review!!!

PR #1584, opened 2020-11-12
HDDS-4441. Add metrics for ACL related operations.(Addendum for HA).

ayushtkn:

Acl metrics in case of HA

https://issues.apache.org/jira/browse/HDDS-4441

## How was this patch tested?

Added UT

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/acl/OMBucketAddAclRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/acl/OMBucketAddAclRequest.java
@@ -24,6 +24,7 @@ import java.util.List;
 import com.google.common.collect.Lists;
 import org.apache.hadoop.ozone.om.OMMetrics;
 import org.apache.hadoop.ozone.om.OzoneManager;
+import org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerDoubleBufferHelper;
 import org.apache.hadoop.ozone.om.request.util.OmResponseUtil;
 import org.apache.hadoop.ozone.util.BooleanBiFunction;
 import org.apache.hadoop.util.Time;
@@ -130,5 +131,12 @@ public class OMBucketAddAclRequest extends OMBucketAclRequest {
     }
   }
 
+  @Override
+  public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
+      long trxnLogIndex, OzoneManagerDoubleBufferHelper omDoubleBufferHelper) {
+    ozoneManager.getMetrics().incNumAddAcl();
+    return super.validateAndUpdateCache(ozoneManager, trxnLogIndex,
+        omDoubleBufferHelper);
+  }
 }
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOmMetrics.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOmMetrics.java
@@ -17,6 +17,7 @@
 package org.apache.hadoop.ozone.om;
 
 import static org.apache.hadoop.ozone.OzoneAcl.AclScope.ACCESS;
+import static org.apache.hadoop.ozone.security.acl.OzoneObj.ResourceType.BUCKET;
 import static org.apache.hadoop.ozone.security.acl.OzoneObj.ResourceType.VOLUME;
 import static org.apache.hadoop.ozone.security.acl.OzoneObj.StoreType.OZONE;
 import static org.apache.hadoop.test.MetricsAsserts.assertCounter;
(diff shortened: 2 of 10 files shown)


jojochuang:
This is beyond my comfort zone. @bharatviswa504  would you like to review this since it touches OM HA?

bharatviswa504 (approved):
+1 LGTM.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/acl/OMBucketAddAclRequest.java
```diff
@@ -130,5 +131,12 @@ void onComplete(boolean operationResult, IOException exception,
     }
   }
 
+  @Override
+  public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
+      long trxnLogIndex, OzoneManagerDoubleBufferHelper omDoubleBufferHelper) {
+    ozoneManager.getMetrics().incNumAddAcl();
```
cxorm, line 137: How about moving this line after true `operationResult` branch of `onComplete()` at line-120 ?

cxorm (commented):
Thanks @ayushtkn for the work, overall looks good to me except one question.

How do you think if we move the acl-metric code to true branch of `onComplete()` of every requests ? (Feel free to correct me if I miss something, thanks)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/acl/OMBucketAddAclRequest.java
ayushtkn: Thanx @cxorm for the review. Metrics should actually track all requests, not just success, Secondly, in `onComplete()`, there doesn't seems to be any instance of `ozoneManager` available, and I quickly had a check over, couple of other files, they too were incrementing metrics in `validateAndUpdateCache`, I think we can be in sync with them.

Let me know, if you are ok going ahead this way.

cxorm: > Thanx @cxorm for the review. In `onComplete()`, there doesn't seems to be any instance of `ozoneManager` available, and I quickly had a check over, couple of other files, they too were incrementing metrics in `validateAndUpdateCache`, I think we can be in sync with them.

Thanks @ayushtkn for the comment.

IMHO the `onComplete()` is called by super class `validateAndUpdateCache` (for example, in this part its super class is `OMBucketAclRequest`), and it has instance of `ozoneManager`.

An operation scenario is that when we call `ozoneManager.getMetrics().incNumAddAcl()` before `super.validateAndUpdateCache()` and failure occurs in `validateAndUpdateCache()`, the `incNumAddAcl()` shouldn't take effect, but it did.

ayushtkn: Thanx @cxorm 
Are you talking about this `onComplete`?
https://github.com/apache/ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/acl/OMVolumeAddAclRequest.java#L114

If not can you point me to that. It doesn't have `ozonemanager`

>An operation scenario is that when we call ozoneManager.getMetrics().incNumAddAcl() before super.validateAndUpdateCache() and failure occurs in validateAndUpdateCache(), the incNumAddAcl() shouldn't take effect, but it did.

Aren't these metrics suppose to track the number of calls? not just success?

`validateAndUpdateCache ` too has a metrics getting incremented, `    omMetrics.incNumVolumeUpdates();`, if a failure occurs post that, it isn't bothered.

ayushtkn: A reference from HDFS namenode.
https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/NameNodeRpcServer.java#L1960

Here too, the metrics is incremented before and the execution is done post that, even if the call fails , the metrics shall still track that call.

bharatviswa504: In Ozone, we increment these metrics represents number of operations, (Not the number of successful operations)
IMHO, this approach matches with all other requests, and LGTM.

cxorm: Thank you @bharatviswa504 for the clarification (and sorry for my previous view.) 
I'm +1 for the patch.

cxorm (approved):
Thanks @ayushtkn for the work and discussion, and @bharatviswa504 for the clarification and review.

I'm merging it.

PR #1585, opened 2020-11-12
HDDS-4458. Fix Max Transaction ID value in OM.

bharatviswa504:

Right now Max Transaction ID value is coming as 4194302, whereas it should be 18014398509481982.

Due to this after 4194302, all transactions in OM will fail.

https://issues.apache.org/jira/browse/HDDS-4458

## How was this patch tested?

Added a test.


hanishakoneru (approved):

hanishakoneru:
LGTM. Thanks for catching this Bharat. 
+1 pending CI.

captainzmc (approved):
LGTM +1. Thanks for Bharat’s patch.

PR #1587, opened 2020-11-13
HDDS-4393. Addressing test failures after master merge.

nandakumar131:

Addressing test failures after master merge.

HDDS-4393

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/ScmAdmin.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/ScmAdmin.java
@@ -20,6 +20,8 @@ package org.apache.hadoop.ozone.admin.scm;
 import org.apache.hadoop.hdds.cli.GenericCli;
 import org.apache.hadoop.hdds.cli.HddsVersionProvider;
 import org.apache.hadoop.hdds.cli.OzoneAdmin;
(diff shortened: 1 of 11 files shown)


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/ScmAdmin.java
```diff
@@ -20,6 +20,8 @@
 import org.apache.hadoop.hdds.cli.GenericCli;
 import org.apache.hadoop.hdds.cli.HddsVersionProvider;
 import org.apache.hadoop.hdds.cli.OzoneAdmin;
+import org.apache.hadoop.hdds.cli.SubcommandWithParent;
+import org.kohsuke.MetaInfServices;
```
mukul1987, line 24: This is a new dependency, Can you please add detail about this ?

mukul1987 (changes requested):

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/scm/ScmAdmin.java
nandakumar131: It's not a new dependency, it was added as part of HDDS-4046.

xiaoyuyao (approved):
LGTM, +1

mukul1987 (approved):
+1

PR #1588, opened 2020-11-13
HDDS-4461. Reuse compiled binaries in acceptance test

adoroszlai:

Save Ozone binaries from `compile` check, and use these in `acceptance` and `kubernetes` checks, instead of each check performing its own full build.  Total execution time is similar, but the dependent checks are started later, so we save GitHub Actions cycles.

https://issues.apache.org/jira/browse/HDDS-4461

## How was this patch tested?

Tested in fork:

```
Artifact ozone-bin has been successfully uploaded!
```

https://github.com/adoroszlai/hadoop-ozone/runs/1395732068#step:8:52

```
Artifact ozone-bin was downloaded to /home/runner/work/hadoop-ozone/hadoop-ozone
```

https://github.com/adoroszlai/hadoop-ozone/runs/1395772736#step:4:11

Also verified coverage is the same (79%) as with the previous commit (760c1e8f9ecc5daa5cd49c95ca946a415976088e).

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -195,7 +195,9 @@ jobs:
         with:
           name: ozone-bin
       - name: Untar binaries
-        run: tar xzvf hadoop-ozone*.tar.gz -C /mnt/ozone
+        run: |
+          mkdir -p /mnt/ozone/hadoop-ozone/dist/target
+          tar xzvf hadoop-ozone*.tar.gz -C /mnt/ozone/hadoop-ozone/dist/target
       - name: Install robotframework
         run: sudo pip install robotframework
       - name: Execute tests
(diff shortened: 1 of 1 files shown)


elek:
Great idea @adoroszlai, thanks the patch.

What do you think about using `-Pdist` instead of the custom tar execution (`-Pdist` supposed to create the tar file).

With this approach we will test exactly the same tar file which will be released...

adoroszlai:
> What do you think about using `-Pdist` instead of the custom tar execution (`-Pdist` supposed to create the tar file).
> 
> With this approach we will test exactly the same tar file which will be released...

Testing part of the release process is a good idea, thanks.  Updated the patch with this.

adoroszlai:
> With this modification we not only share the artifact between jobs but also archive them for 90 days.

Good point.  Updated to keep it only for one day.

elek:
> Good point. Updated to keep it only for one day.

TIL it's possible to define retention per upload. Thanks.

elek (approved):

adoroszlai:

PR #1589, opened 2020-11-13
HDDS-4462. Add --frozen-lockfile to pnpm install to prevent ozone-recon-web/pnpm-lock.yaml from being updated automatically

smengcl:
Trivial fix.
~~Update `ozone-recon-web/pnpm-lock.yaml` as Git is reporting not clean after a maven build.~~

Add --frozen-lockfile to pnpm install to prevent ozone-recon-web/pnpm-lock.yaml from being updated automatically.

### DIFF
--- a/hadoop-ozone/recon/pom.xml
+++ b/hadoop-ozone/recon/pom.xml
@@ -118,7 +118,7 @@
               <goal>npx</goal>
             </goals>
             <configuration>
-              <arguments>pnpm install</arguments>
+              <arguments>pnpm install --frozen-lockfile</arguments>
             </configuration>
           </execution>
           <execution>
--- a/hadoop-ozone/recon/src/main/resources/webapps/recon/ozone-recon-web/pnpm-lock.yaml
+++ b/hadoop-ozone/recon/src/main/resources/webapps/recon/ozone-recon-web/pnpm-lock.yaml
@@ -42,7 +42,7 @@ devDependencies:
   json-server: 0.15.1
   npm-run-all: 4.1.5
   xo: 0.30.0
-lockfileVersion: 5.1
+lockfileVersion: 5.2
 packages:
   /3d-view/2.0.0:
     dependencies:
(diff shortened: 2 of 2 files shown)


vivekratnavel:
This shouldn't be happening. I am curious to know which maven build command did you use?

smengcl:
> This shouldn't be happening. I am curious to know which maven build command did you use?

`mvn clean install -DskipTests -Dmaven.javadoc.skip`

I found this issue on both my Mac and in a Ubuntu VM.

vivekratnavel:
Thanks, @smengcl! I am unable to reproduce this lock file update on my machine.

Can you please try `<arguments>pnpm install --frozen-lockfile</arguments>` in https://github.com/apache/ozone/blob/master/hadoop-ozone/recon/pom.xml#L121 ? You can then run the same mvn command inside `hadoop-ozone/recon` directory to build only recon and quickly test. Before the test, don't forget to checkout ozone-recon-web/pnpm-lock.yaml if there are any local changes. This time it should not update the lock file and your git tree should be clean. Please let me know the results. Thanks!

smengcl:
> Thanks, @smengcl! I am unable to reproduce this lock file update on my machine.
> 
> Can you please try `<arguments>pnpm install --frozen-lockfile</arguments>` in https://github.com/apache/ozone/blob/master/hadoop-ozone/recon/pom.xml#L121 ? You can then run the same mvn command inside `hadoop-ozone/recon` directory to build only recon and quickly test. Before the test, don't forget to checkout ozone-recon-web/pnpm-lock.yaml if there are any local changes. This time it should not update the lock file and your git tree should be clean. Please let me know the results. Thanks!

I tested that adding `--frozen-lockfile` to the argument works. `ozone-recon-web/pnpm-lock.yaml` is clean after the same mvn command.

I will change this jira to adding `--frozen-lockfile` instead.

vivekratnavel:
Thanks @smengcl! Patch looks good to me. +1

vivekratnavel (approved):

PR #1593, opened 2020-11-16
HDDS-4444. Remove KeyManagerImpl#refreshPipeline because it is the same as refresh().

flirmnave:

Remove KeyManagerImpl#refreshPipeline() and replace with KeyManagerImpl#refresh().

https://issues.apache.org/jira/browse/HDDS-4444

## How was this patch tested?

Unit test passed.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestKeyManagerImpl.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestKeyManagerImpl.java
@@ -1181,7 +1181,7 @@ public class TestKeyManagerImpl {
       KeyManagerImpl keyManagerImpl =
           new KeyManagerImpl(ozoneManager, scmClientMock, conf, "om1");
 
-      keyManagerImpl.refreshPipeline(omKeyInfo);
+      keyManagerImpl.refresh(omKeyInfo);
 
       verify(sclProtocolMock, times(1))
           .getContainerWithPipelineBatch(containerIDs);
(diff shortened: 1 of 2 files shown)


bharatviswa504 (commented):
Can we remove refresh, and keep refreshPipeline instead, which is clear.

And also right now OmKeyInfo does not have pipelineInfo after HDDS-3658, so it is basically fetchPipeline, I am fine with leaving it as refresh also, just thought of posting here.

ChenSammi:
Hi @flirmnave,  thanks for working on this. 

The key point backed in HDDS-3658 is new API "refresh" is added in the KeyManager interface which is called from other modules.   It's good to remove the duplicate functions and keep the code clean.

flirmnave:
Thanks @bharatviswa504 @ChenSammi for comment and review.

Based on @ChenSammi comment, 
KeyManagerImpl#refresh() is a API which called from other modules,
so I remove KeyManagerImpl#refreshPipeline() and keep KeyManagerImpl#refresh() instead.

jojochuang (approved):
LGTM
@bharatviswa504  let's get this merged before HDDS-4537 / PR-1660?

bharatviswa504:
@flirmnave Then we can rename in the KeyManager to refreshPipeline, right?

bharatviswa504 (approved):

bharatviswa504:
> LGTM
> @bharatviswa504 let's get this merged before [HDDS-4537](https://issues.apache.org/jira/browse/HDDS-4537) / PR-1660?

Sure. I am fine with keeping refresh method.

flirmnave:
Thanks @bharatviswa504 @ChenSammi @jojochuang for review and comment.

PR #1598, opened 2020-11-17
HDDS-4478. Large deletedKeyset slows down OM via listStatus.

fapifta:

See details about the issue found in the JIRA, the change proposed is fairly simple, as the key table is missing from the CleanupTableInfo annotation values, which prevents cache eviction in the OMDoubleBuffer logic.

https://issues.apache.org/jira/browse/HDDS-4478

## How was this patch tested?

The JUnit test I wrote tests this particular scenario with the given request. I am posting a follow up JIRA for a most comprehensive test that covers probably all the request - response pairs somehow. Or at least as an outcome, helps us detect these kind of problems easier and earlier.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponse.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponse.java
@@ -20,18 +20,24 @@ package org.apache.hadoop.ozone.om.response.file;
 
 import javax.annotation.Nonnull;
 
+import org.apache.hadoop.ozone.om.OmMetadataManagerImpl;
 import org.apache.hadoop.ozone.om.helpers.OmBucketInfo;
 import org.apache.hadoop.ozone.om.helpers.OmKeyInfo;
 import org.apache.hadoop.ozone.om.helpers.OmVolumeArgs;
+import org.apache.hadoop.ozone.om.response.CleanupTableInfo;
 import org.apache.hadoop.ozone.om.response.key.OMKeyCreateResponse;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos
     .OMResponse;
 
 import java.util.List;
 
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.KEY_TABLE;
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.OPEN_KEY_TABLE;
+
 /**
  * Response for crate file request.
  */
+@CleanupTableInfo(cleanupTables = {KEY_TABLE, OPEN_KEY_TABLE})
 public class OMFileCreateResponse extends OMKeyCreateResponse {
 
   public OMFileCreateResponse(@Nonnull OMResponse omResponse,
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponse.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponse.java
@@ -20,7 +20,6 @@ package org.apache.hadoop.ozone.om.response.file;
 
 import javax.annotation.Nonnull;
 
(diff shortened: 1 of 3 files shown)


fapifta:
I am also planning to install this patch on the cluster I ran into the problem, and check if the problem is fixed there as well.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponse.java
```diff
@@ -23,15 +23,20 @@
...
 
 import java.util.List;
 
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.KEY_TABLE;
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.OPEN_KEY_TABLE;
+
 /**
  * Response for crate file request.
  */
+@CleanupTableInfo(cleanupTables = {KEY_TABLE, OPEN_KEY_TABLE})
```
bharatviswa504, line 39: KEY_TABLE is needed for KeyCreate also, as when ozone.om.enable.filesystem.paths is true.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/TestCleanupTableInfo.java
```diff
@@ -66,4 +105,188 @@ public void checkAnnotationAndTableName() throws Exception {
...
+
+
+    request.validateAndUpdateCache(om, 1, dbh);
+
+
+    CleanupTableInfo ann =
+        OMFileCreateResponse.class.getAnnotation(CleanupTableInfo.class);
+    List<String> cleanup = Arrays.asList(ann.cleanupTables());
+    for (String tableName : omMetaMgr.listTableNames()) {
+      if (!cleanup.contains(tableName)) {
```
bharatviswa504, line 147: Here we have checked only tables which are not part of FileCreateResponse cleanupTable annotation.
Do we want to check tables which are affected also.

Just a question, not got what these lines are testing? (Is it just to see any tables which are not affected have same size in Cache) But how this is verifying fix, not sure if i am missing something basic here.

bharatviswa504 (commented):
Good catch @fapifta 
Over all LGTM. Few comments inline.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponse.java
fapifta: Hmm... yes, you are right, key create I have missed for this one...
Let me add that along with a test tomorrow.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/TestCleanupTableInfo.java
fapifta: Yes, the test does what you have summarized.
The basic idea is the following:
The issue is that we have certain epochs that are pushing entries to unexpected table caches. Unexpected in a way that eviction for those epochs on particular tables is not called when the DoubleBuffer flushes, because of the missing table name in the annotation.
I think it is sufficient to check whether we have added any unexpected cache entries to any other table's cache during applyTransaction. I might be wrong on this one, or there might be an easier way though.

fapifta:
Thank you for the review @bharatviswa504 please find my answers inline, and expect the addition of key create case tomorrow.
While working on that one, I will try to come up with a more general solution, if the overall idea of how to test this seems suitable.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/TestCleanupTableInfo.java
bharatviswa504: Ya, it makes sense to me. It is basically checking all other table cache entries should be what it has an initial value, which should not be changed.

bharatviswa504 (approved):
+1 LGTM

bharatviswa504 (commented):
Thank You @fapifta for the contribution.

PR #1600, opened 2020-11-18
HDDS-4475.Extend DatanodeChunkGenerator to write all on all pipelines…

sadanand48:
Currently, DatanodeChunkGenerator takes a single pipeline as a parameter. This will allow passing a list of pipelines as comma-separated by their pipeline ids and the load will be generated on the dns of the provided pipelines.

https://issues.apache.org/jira/browse/HDDS-4475

## How was this patch tested?
Tested on docker.

### DIFF
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/DatanodeChunkGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/DatanodeChunkGenerator.java
@@ -85,7 +85,9 @@ public class DatanodeChunkGenerator extends BaseFreonGenerator implements
   private String pipelineIds;
 
   @Option(names = {"-d", "--datanodes"},
-      description = "Datanodes to use. ",
+      description = "Datanodes to use." +
+          " Test will write to all the existing pipelines " +
+          "which this datanode is member of.",
       defaultValue = "")
   private String datanodes;
 
@@ -133,8 +135,6 @@ public class DatanodeChunkGenerator extends BaseFreonGenerator implements
             .acquireClient(firstPipeline);
         xceiverClients = new ArrayList<>();
         xceiverClients.add(xceiverClientSpi);
-        LOG.info("Using pipeline {}", firstPipeline.getId());
-        runTest();
       } else {
         xceiverClients = new ArrayList<>();
         pipelines = new HashSet<>();
@@ -152,11 +152,10 @@ public class DatanodeChunkGenerator extends BaseFreonGenerator implements
         }
         if (pipelines.isEmpty()){
           throw new IllegalArgumentException(
-              "Coudln't find the any/the selected pipeline");
-        } else {
-          runTest();
+              "Couldn't find the any/the selected pipeline");
         }
       }
+      runTest();
     } finally {
       for (XceiverClientSpi xceiverClientSpi : xceiverClients) {
         if (xceiverClientSpi != null) {
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/DatanodeChunkGenerator.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/DatanodeChunkGenerator.java
@@ -18,9 +18,11 @@ package org.apache.hadoop.ozone.freon;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.Set;
 import java.util.List;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Arrays;
 import java.util.concurrent.Callable;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


bshashikant:
@sadanand48 , as discuseed offline. Can you make the write chunk tests to run concurrently rather than sequentially?

Also, can we also add support for multiple datanodes along with pipeline Ids?

bshashikant:
@elek , can you plz have a look at this?

elek (changes requested):
> This will allow passing a list of pipelines as comma-separated by their pipeline ids and the load will be generated on the dns of the provided pipelines.

Thanks for the patch @sadanand48. I don't fully understand this line. Can you please explain the goals in more details?

I have one guess: the goal is to define the pipeline with the help of a datanode host name.

Today we have two options:

 1. If pipelineId is defined, that will be used
 2. If pipline is not defined, we check the SCM and will use the first Ratis/THREE open pipeline

In both cases the first datanode in the pipeline will be used to save chunk.

If I understood well, this patch would like to offer a third option: if the datanode is set, the list of the pipelines will be further restricted to the pipelines where the specified datanode is a member.

If this is the goal, it should be possible with adjusting the filter conditions:

```java
        temp = pipelinesFromSCM.stream()
              .filter(p -> p.getFactor() == ReplicationFactor.THREE)
              .filter(p -> datanodeHosts.size() ==0 || pipelineContainsDatanode(p, datanodeHosts))
              .findFirst()
              .orElseThrow(() -> new IllegalArgumentException(
                  "Pipeline ID is NOT defined, and no pipeline " +
                      "has been found with factor=THREE"));
```

(see the second filter)

I agree with @bshashikant, we shouldn't remove the logic of the parallel executions for this specific case.

(But please let me know if I misunderstood something)

sadanand48:
Thanks @elek  for thr comments . The current tool takes only 1 pipeline as parameter and generates chunks on this pipeline.If any arguments not provided, it will by default select the RATIS THREE pipeline and write to it.
The goal here is to 
1. Pass multiple pipelines at once and generate write chunk requests on them .
2. Pass datanode/datanodes and select those pipelines of which these datanodes are part of and write to them. I will use the second filter for that like you suggested.
The default behaviour remains the same.

elek:
Thanks to explain it, now I got it. 

In that case you should have additional modification to use the current loop:

You should either select the first one OR the pipelines assigned to the datanode.

Change `private XceiverClientSpi xceiverClientSpi` to `private List<XceiverClientSpi> clients` to have clients to all pipelines.

And in `writeChunk` select a random one instead of using the single one. 

(You should also initialize the full list of clients and close them...) 

Also: how would you like to guarantee that if the selected datanode is the leader of bot pipelines?

elek:
(You can ping me offline if it was not clear, need help or if my understanding is till not right ;-) )

elek:
Thanks the update @sadanand48 

This part seems to be suspicious:

```
 179   │   private void writeOnPipeline(OzoneConfiguration ozoneConf,
 180   │       List<Pipeline> pipelines) throws IOException {
 181   │     LOG.info("Inside write Pipeline and pipeline size" + pipelines.size());
 182   │     xceiverClients = new ArrayList<>();
 183   │     for (Pipeline p: pipelines){
 184   │       init();
 185   │       LOG.info("run test on pipeline" + p.getId().toString());
 186   │       XceiverClientSpi clientSpi = xceiverClientManager.acquireClient(p);
 187   │       xceiverClients.add(clientSpi);
 188   │       runTest(clientSpi);
 189   │     }
 190   │   }
 ```

It has multiple `init()` calls which reset all the counters. `runTest` should be called only once IMHO.

File: hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/DatanodeChunkGenerator.java [hunk 1]
```diff
@@ -75,70 +82,120 @@
...
+               .collect(Collectors.toList());
+          pipelines.addAll(selectedPipelines);
+        }
+        for (Pipeline p:pipelines){
+          LOG.info("Writing to pipeline: " + p.getId());
+          xceiverClients.add(xceiverClientManager.acquireClient(p));
+        }
+        if (pipelines.isEmpty()){
+          throw new IllegalArgumentException(
+              "Coudln't find the any/the selected pipeline");
```
elek, line 155:
```suggestion
              "Couldn't find the any/the selected pipeline");
```

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/DatanodeChunkGenerator.java
```diff
@@ -75,70 +82,120 @@
...
+              .filter(p -> p.getFactor() == ReplicationFactor.THREE)
+              .findFirst()
+              .orElseThrow(() -> new IllegalArgumentException(
+                  "Pipeline ID is NOT defined, and no pipeline " +
+                      "has been found with factor=THREE"));
+        XceiverClientSpi xceiverClientSpi = xceiverClientManager
+            .acquireClient(firstPipeline);
+        xceiverClients = new ArrayList<>();
+        xceiverClients.add(xceiverClientSpi);
+        LOG.info("Using pipeline {}", firstPipeline.getId());
```
elek, line 136:
```suggestion
```

You don't need this line as you log the same information in the loop bellow.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/DatanodeChunkGenerator.java
```diff
@@ -75,70 +82,120 @@
       description = "Pipeline to use. By default the first RATIS/THREE "
           + "pipeline will be used.",
       defaultValue = "")
-  private String pipelineId;
+  private String pipelineIds;
 
-  private XceiverClientSpi xceiverClientSpi;
+  @Option(names = {"-d", "--datanodes"},
+      description = "Datanodes to use. ",
```
elek, line 88:
```suggestion
      description = "Datanodes to use. Test will write to all the existing pipelines which this datanode is member of.",
```

elek (commented):
+1, thanks the update @sadanand48 

I have 3 minor/typo comments, but I like this approach

elek (approved):

elek:
Merging it. Thanks for the continuous improvement @sadanand48

PR #1604, opened 2020-11-19
HDDS-4356. SCM is flooded with useless "Deleting blocks" messages.

lokeshj1703:

SCM deletes 1000 blocks max at a time, and the "Deleting blocks" message repeats for 1000 times. The PR changes the log level to DEBUG.

https://issues.apache.org/jira/browse/HDDS-4356

## How was this patch tested?

Existing UT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
@@ -129,9 +129,12 @@ public class DeletedBlockLogImpl
         DeletedBlocksTransaction block =
             scmMetadataStore.getDeletedBlocksTXTable().get(txID);
         if (block == null) {
-          // Should we make this an error ? How can we not find the deleted
-          // TXID?
-          LOG.warn("Deleted TXID {} not found.", txID);
+          if (LOG.isDebugEnabled()) {
+            // This can occur due to race condition between retry and old
+            // service task where old task removes the transaction and the new
+            // task is resending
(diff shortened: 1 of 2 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
```diff
@@ -129,9 +129,12 @@ public void incrementCount(List<Long> txIDs) throws IOException {
         DeletedBlocksTransaction block =
             scmMetadataStore.getDeletedBlocksTXTable().get(txID);
         if (block == null) {
-          // Should we make this an error ? How can we not find the deleted
-          // TXID?
-          LOG.warn("Deleted TXID {} not found.", txID);
```
xiaoyuyao, line 134: Let's keep this rare error case as WARN. Other changes LGTM.

lokeshj1703: @xiaoyuyao I can see this warning message being printed a lot of times during deletion. It can easily fill up SCM log during deletion. How about we change the log level and investigate the reason in a separate jira?
Since SCM is processing thousand of transactions and it resends these transactions to a particular datanode. Then this log can be printed thousands of times.

xiaoyuyao: Agree with the analysis. Let's investigate the cause in a separate JIRA. +1 for merging the PR as-is.

lokeshj1703:
@xiaoyuyao Thanks for the review! I have committed the PR to master branch.

PR #1605, opened 2020-11-19
HDDS-4488. Open RocksDB read only when loading containers at Datanode startup

sodonnel:

When a datanode is started, it must read some metadata from all the Containers. Part of that metadata is stored in RocksDB, so the startup process involves opening each rocksDB and closing it again.

Testing on a dense node, with 45 high performance spinning disks and 200K containers, I saw about 75ms on average to open each RockDB. Further testing demonstrated that if we open RockDB read only, the average open time is about 35ms.

At startup time, the DBs are only read, and never written, so opening read only is fine. HDDS-4427 already ensures these opened DBs are not cached.

This change adds an "openReadOnly" flag to RDBStore and the related classes to pass it down the stack, allowing the DB to be opened read only.

https://issues.apache.org/jira/browse/HDDS-4488

## How was this patch tested?

Currently existing unit tests.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/ContainerCache.java
@@ -157,7 +157,7 @@ public final class ContainerCache extends LRUMap {
       try {
         long start = Time.monotonicNow();
         DatanodeStore store = BlockUtils.getUncachedDatanodeStore(containerID,
-            containerDBPath, schemaVersion, conf);
+            containerDBPath, schemaVersion, conf, false);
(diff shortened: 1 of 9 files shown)


runzhiwang:
@sodonnel Thanks the patch. Do you test the time cost when open RockDB with read only ?

sodonnel:
@runzhiwang We have a cluster running with some dense nodes. I put some experimental changes into the code and tested the startup time before and after. With the original code, the startup time was about 500 seconds.

With HDDS-4427 plus opening RocksDB readonly (this change) the startup time dropped to about 250 seconds.

Timing the average RocksDB open time - read-write it was 75ms and read only it was about 35ms. 

These results were consistent across several test runs.

runzhiwang:
@sodonnel Thanks. Got it.

runzhiwang:
@sodonnel Thanks the patch. I have merged it.

PR #1607, opened 2020-11-20
HDDS-4358: Delete : make delete an atomic operation

rakeshadr:
This task handles directory and file deletions. This won't cleanup the garbage and orphans which will be implemented in another jira.

https://issues.apache.org/jira/browse/HDDS-4358

## How was this patch tested?

TestOzoneFileSystemV1 suite

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemV1.java
@@ -18,7 +18,6 @@
 
 package org.apache.hadoop.fs.ozone;
 
-import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.LocatedFileStatus;
@@ -26,15 +25,8 @@ import org.apache.hadoop.fs.RemoteIterator;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.contract.ContractTestUtils;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
-import org.apache.hadoop.hdds.utils.db.Table;
-import org.apache.hadoop.hdds.utils.db.TableIterator;
-import org.apache.hadoop.hdds.utils.db.cache.CacheKey;
-import org.apache.hadoop.hdds.utils.db.cache.CacheValue;
 import org.apache.hadoop.ozone.om.OMConfigKeys;
-import org.apache.hadoop.ozone.om.OMMetadataManager;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
-import org.apache.hadoop.ozone.om.helpers.OmDirectoryInfo;
-import org.apache.hadoop.ozone.om.helpers.OmKeyInfo;
 import org.jetbrains.annotations.NotNull;
 import org.junit.Assert;
 import org.junit.Rule;
@@ -48,11 +40,11 @@ import org.slf4j.LoggerFactory;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.Map;
 
 import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_TRASH_INTERVAL_KEY;
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
 
 /**
  * Ozone file system tests that are not covered by contract tests,
@@ -309,78 +301,6 @@ public class TestOzoneFileSystemV1 extends TestOzoneFileSystem {
   }
 
   /**
-   * Case-5) If new destin '/dst/source' exists then throws exception.
-   * If destination is a directory then rename source as sub-path of it.
-   * <p>
-   * For example: rename /a to /b will lead to /b/a. This new path should
-   * not exist.
-   */
-  protected void testRenameToNewSubDirShouldNotExist() throws Exception {
-    // Case-5.a) Rename directory from /a to /b.
-    // created /a
-    final Path aSourcePath = new Path(fs.getUri().toString() + "/a");
-    fs.mkdirs(aSourcePath);
-
-    // created /b
-    final Path bDestinPath = new Path(fs.getUri().toString() + "/b");
-    fs.mkdirs(bDestinPath);
-
-    // Add a sub-directory '/b/a' to '/b'. This is to verify that rename
-    // throws exception as new destin /b/a already exists.
-    final Path baPath = new Path(fs.getUri().toString() + "/b/a");
-    fs.mkdirs(baPath);
-
-    try {
-      fs.rename(aSourcePath, bDestinPath);
-      Assert.fail("Should fail as new destination dir exists!");
-    } catch (OMException ome) {
-      // expected as new sub-path /b/a already exists.
-      assertEquals(ome.getResult(), OMException.ResultCodes.KEY_ALREADY_EXISTS);
-    }
-
-    // Case-5.b) Rename file from /a/b/c/file1 to /a.
-    // Should be failed since /a/file1 exists.
-    final Path abcPath = new Path(fs.getUri().toString() + "/a/b/c");
-    fs.mkdirs(abcPath);
-    Path abcFile1 = new Path(abcPath, "/file1");
-    ContractTestUtils.touch(fs, abcFile1);
-
-    final Path aFile1 = new Path(fs.getUri().toString() + "/a/file1");
-    ContractTestUtils.touch(fs, aFile1);
-
-    final Path aDestinPath = new Path(fs.getUri().toString() + "/a");
-
-    try {
-      fs.rename(abcFile1, aDestinPath);
-      Assert.fail("Should fail as new destination file exists!");
-    } catch (OMException ome) {
-      // expected as new sub-path /b/a already exists.
-      assertEquals(ome.getResult(), OMException.ResultCodes.KEY_ALREADY_EXISTS);
-    }
-  }
-
-  /**
-   * Case-6) Rename directory to an existed file, should be failed.
-   */
-  protected void testRenameDirToFile() throws Exception {
-    final String root = "/root";
-    Path rootPath = new Path(fs.getUri().toString() + root);
-    fs.mkdirs(rootPath);
-
-    Path file1Destin = new Path(fs.getUri().toString() + root + "/file1");
-    ContractTestUtils.touch(fs, file1Destin);
-    Path abcRootPath = new Path(fs.getUri().toString() + "/a/b/c");
-    fs.mkdirs(abcRootPath);
-    try {
-      fs.rename(abcRootPath, file1Destin);
-      Assert.fail("key already exists /root_dir/file1");
-    } catch (OMException ome) {
-      // expected
-      assertEquals(ome.getResult(), OMException.ResultCodes.KEY_ALREADY_EXISTS);
-    }
-  }
-
-  /**
    * Fails if the (a) parent of dst does not exist or (b) parent is a file.
    */
   protected void testRenameDestinationParentDoesntExist() throws Exception {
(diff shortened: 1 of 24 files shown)


hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -498,6 +498,14 @@ public boolean delete(Path f, boolean recursive) throws IOException {
     incrementCounter(Statistic.INVOCATION_DELETE, 1);
     statistics.incrementWriteOps(1);
     LOG.debug("Delete path {} - recursive {}", f, recursive);
+
+    String layOutVersion = adapter.getBucketLayoutVersion();
+    if (layOutVersion != null &&
+            OMConfigKeys.OZONE_OM_LAYOUT_VERSION_V1.equals(layOutVersion)) {
+      String key = pathToKey(f);
+      return adapter.deleteObject(key);
```
linyiqun, line 506: HI @rakeshadr , current PR change hasn't implemented recursively delete, right?

rakeshadr: Thanks a lot @linyiqun for the reviews.

This patch is removing the given keyPath from the DirTable(if its a dir) or from the KeyTable(if its a file). Presently, it is not checking whether the dir is empty or not then handle it based on recursive flag.

_**Need to implement: Point-1**_: I've to pass the 'recursive' flag to the OM then throw OMException("Directory is not empty") if `recursive==true && keyPath#isNotEmptyDir()`. This I will do in current PR on next commit.

_**Need to implement: Point-2**_: Like I mentioned in the description, once the given keyPath is removed it requires cleanup of sub-dirs/files for a non-empty parent. Since the parent doesn't exists the traversal logic will not be able to find its children. For example: '/a/b/c/d/e/file1'. Now user deletes '/b' , then the key entry '<parentID>/b=<objectID>' will be removed on the user call path. Assume parentID=1023 and objectID=1024. Deletion of '1023/b=1024' key hides all its children and become orphan. '1024/c'. Now, these garbage has to cleaned up async way and can be done via HDDS-4495.

_**Need to implement: Point-3**_: Also, for the above point-2, it has to modify the existing KeyDeletingService to delete all the file-blocks. I will analyse these flow during HDDS-4495 implementation then will handle it based on the complexity.

IMHO, Point-2 & Point-3 can be done separately as its not affecting the functionality of recursive or non-recursive deletion. Does this make sense to you?

linyiqun: @rakeshadr , sounds good to me. Please go head for this, : ).

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequestV1.java
```diff
@@ -0,0 +1,195 @@
...
+        // Update table cache.
+        omMetadataManager.getKeyTable().addCacheEntry(
+                new CacheKey<>(ozonePathKey),
+                new CacheValue<>(Optional.absent(), trxnLogIndex));
+      }
+
+      omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
+      omBucketInfo = getBucketInfo(omMetadataManager, volumeName, bucketName);
+
+      long quotaReleased = sumBlockLengths(omKeyInfo);
```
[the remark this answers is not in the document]
bharatviswa504: There is a discussion going on to remove bytesUsed from VolumeArgs once HDDS-4308, we might need to revisit once after that went in.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeyDeleteResponseV1.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.annotation.Nonnull;
+import java.io.IOException;
+
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.DELETED_TABLE;
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.KEY_TABLE;
+
+/**
+ * Response for DeleteKey request.
+ */
+@CleanupTableInfo(cleanupTables = {KEY_TABLE, DELETED_TABLE})
```
bharatviswa504, line 39: KEY_TABLE  -> FILE_TABLE

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemV1.java
```diff
@@ -475,6 +475,15 @@ public void testFileSystem() throws Exception {
...
+    tableCleanup();
+
+    testDeleteRoot();
+    tableCleanup();
+
+    testRecursiveDelete();
+    tableCleanup();
   }
 
   /**
```
bharatviswa504, line 489: Can we rename tableCleanup to deleteRootDir or something meaningful?
And also update Javadoc now delete is supported for V1 with this patch.

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -498,6 +498,14 @@ public boolean delete(Path f, boolean recursive) throws IOException {
...
+    }
+
```
bharatviswa504, line 508: For V1, we missed the root check delete at the client end.


For old buckets, this check is there.
```
      if (f.isRoot()) {
        return false;
      }
```

bharatviswa504 (commented):
Thank You @rakeshadr for the PR.
Few comments inline.

bharatviswa504:
And also as discussed offline from S3G interface
Recursive=false should be default for key/directory deletion and the directory key can only be deleted when the directory is empty. (Currently, when fs path enabled directory deletion says KeynotFound, this will fix that behavior)

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
rakeshadr: Yes, agreed. I will make this way:-)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequestV1.java
```diff
@@ -0,0 +1,195 @@
...
+
+      // Validate bucket and volume exists or not.
+      validateBucketAndVolume(omMetadataManager, volumeName, bucketName);
+
+      OzoneFileStatus keyStatus =
+              OMFileRequest.getOMKeyInfoIfExists(omMetadataManager, volumeName,
+                      bucketName, keyName, 0);
+
+      if (keyStatus == null) {
+        throw new OMException("Key not found", KEY_NOT_FOUND);
```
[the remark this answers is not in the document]
rakeshadr: Sure, will add it

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequestV1.java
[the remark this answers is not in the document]
linyiqun: HDDS-4308 was resolved , volume used bytes is not used. Please have a look for this RP fix #1489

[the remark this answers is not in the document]
rakeshadr: Thanks for pointing out this. It requires more wider changes. I've raised HDDS-4321 to revisit and correct all the places in the branch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeyDeleteResponseV1.java
rakeshadr: Done in latest commit.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemV1.java
rakeshadr: Done in latest commit.

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
rakeshadr: Done in latest commit.

rakeshadr:
@bharatviswa504  updated patch by addressing your comments. Please let me know your feedback. Thanks!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -759,4 +762,93 @@ public static long getToKeyNameParentId(String volumeName,
...
+    String seekDirInDB = metaMgr.getOzonePathKey(omKeyInfo.getObjectID(), "");
+    TableIterator<String, ? extends Table.KeyValue<String, OmDirectoryInfo>>
+            iterator = dirTable.iterator();
+
+    iterator.seek(seekDirInDB);
+
+    while (iterator.hasNext()) {
+      OmDirectoryInfo dirInfo = iterator.value().getValue();
+      return isImmediateChild(dirInfo.getParentObjectID(),
+              omKeyInfo.getObjectID());
```
linyiqun, line 810: Here should use 'if (iterator.hasNext())' not 'while (iterator.hasNext())'. We just check if the next path is immediate child.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -759,4 +762,93 @@ public static long getToKeyNameParentId(String volumeName,
...
+    TableIterator<String, ? extends Table.KeyValue<String, OmKeyInfo>>
+            iterator = fileTable.iterator();
+
+    iterator.seek(seekFileInDB);
+
+    while (iterator.hasNext()) {
+      OmKeyInfo fileInfo = iterator.value().getValue();
+      return isImmediateChild(fileInfo.getParentObjectID(),
+              omKeyInfo.getObjectID()); // found a sub path file
+    }
```
linyiqun, line 847: The same comment for above place.

linyiqun (commented):
The latest PR almost looks good. Please have a look for my minor comments below, @rakeshadr .

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemV1.java
```diff
@@ -308,78 +300,6 @@ protected void testRenameDirToItsOwnSubDir() throws Exception {
...
-    Path file1Destin = new Path(fs.getUri().toString() + root + "/file1");
-    ContractTestUtils.touch(fs, file1Destin);
-    Path abcRootPath = new Path(fs.getUri().toString() + "/a/b/c");
-    fs.mkdirs(abcRootPath);
-    try {
-      fs.rename(abcRootPath, file1Destin);
-      Assert.fail("key already exists /root_dir/file1");
-    } catch (OMException ome) {
-      // expected
-      assertEquals(ome.getResult(), OMException.ResultCodes.KEY_ALREADY_EXISTS);
```
linyiqun, line 379: As I see we catch the OMException.ResultCodes.KEY_ALREADY_EXISTS error case, can we just make a minor change to adapt this? These test cases can be kept here.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
rakeshadr: Thanks again @linyiqun for the review help. I will change it.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemV1.java
rakeshadr: IIUC, you are suggesting to retain this test. Actually,  I haven't removed this test instead I avoided overriding  #testRenameDirToFile in TestOzoneFileSystemV1. Since TestOzoneFileSystemV1 extends TestOzoneFileSystem, it will run from the parent class. Does this fine for you?

linyiqun: Okay, have seen that test cases.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
rakeshadr: Done!

linyiqun (approved):
+1. Thanks for addressing the comments, @rakeshadr .

rakeshadr:
Thanks @linyiqun  and @bharatviswa504  for the useful review comments. @bharatviswa504

PR #1608, opened 2020-11-21
HDDS-4481. With HA OM can send deletion blocks to SCM multiple times.

bharatviswa504:

OM sends multiple times the same batch of block deletes to SCM, as keys will not be deleted from delete table, even though they are submitted for the purge.

**Reason for this issue:**
1. When KeyDelete Service submits a request KeyPurge from KeyDelete Service, as this is submitted internally from OMServer the callID and ClientID will be passed the same, the subsequent requests are considered as retry, and it will not be processed until ratis clears its retry cache.

https://issues.apache.org/jira/browse/HDDS-4481

## How was this patch tested?

Added a test and also tested the fix on the cluster, now KeyDeletion service is working as expected.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
@@ -34,6 +34,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
+import com.google.common.base.Preconditions;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.conf.StorageUnit;
 import org.apache.hadoop.hdds.server.ServerUtils;
@@ -85,6 +86,9 @@ import org.apache.ratis.util.StringUtils;
 import org.apache.ratis.util.TimeDuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import static org.apache.hadoop.ipc.RpcConstants.DUMMY_CLIENT_ID;
+import static org.apache.hadoop.ipc.RpcConstants.INVALID_CALL_ID;
 
 /**
  * Creates a Ratis server endpoint for OM.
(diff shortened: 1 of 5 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -126,15 +126,32 @@ private static long nextCallId() {
   public OMResponse submitRequest(OMRequest omRequest) throws ServiceException {
     RaftClientRequest raftClientRequest =
         createWriteRaftClientRequest(omRequest);
```
lokeshj1703, line 128: `createWriteRaftClientRequest` was creating a request with the same callId. Is that the bug here?
If that is the case we would also need to make changes in this function so that callId is different for every request.

lokeshj1703 (commented):
@bharatviswa504 Thanks for working on this! The changes look good to me. I have one comment inline.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
bharatviswa504: When called from Client, Server.getCallID() will not return constant value RpcConstants.INVALID_RETRY_COUNT.
Where as when keydeletingService is called it is not coming thru RpcClient, directly calling server API submitRequest, so CallID will be same for each call.

So it is fine use createWriteRaftClientRequest for normal flow of requests, and the values ClientID and CallID can be fetched from Server.getClientID and Server.getCallID().

lokeshj1703: Can we add an assertion to check that Server.getCallID() is not returning RpcConstants.INVALID_CALL_ID and Server.getClientId() is not returning RpcConstants.DUMMY_CLIENT_ID?

bharatviswa504: Done.

lokeshj1703 (approved):
@bharatviswa504 Thanks for updating the PR! The changes look good to me. +1. Pending CI.

bharatviswa504:
Thank You @lokeshj1703 for the review.

PR #1609, opened 2020-11-22
HDDS-4492. CLI flag --quota should default to 'spaceQuota' to preserve backward compatibility.

ayushtkn:

Added a hidden option --quota, to work same as --space-quota, if --space-quota isn't specified. Made hidden so as to prevent new users from using it.

https://issues.apache.org/jira/browse/HDDS-4492

## How was this patch tested?

Added UT and on docker cluster ::

```
Usage: ozone sh volume create [-hV] [--count-quota=<quotaInCounts>]
                              [--space-quota=<quotaInBytes>] [-u=<ownerName>]
                              <value>
Creates a volume for the specified user
      <value>              URI of the volume.
                           Ozone URI could start with o3:// or without prefix.
                             URI may contain the host/serviceId  and port of
                             the OM server. Both are optional. If they are not
                             specified it will be identified from the config
                             files.
      --count-quota=<quotaInCounts>
                           For volume this parameter represents the number of
                             buckets, and for buckets represents the number of
                             keys (eg. 5)
  -h, --help               Show this help message and exit.
      --space-quota=<quotaInBytes>
                           The maximum space quota can be used (eg. 1GB)
  -u, --user=<ownerName>   Owner of the volume
  -V, --version            Print version information and exit.
bash-4.2$ ozone sh volume create vol1 --quota 10BYTES
bash-4.2$ ozone sh volume info vol1
{
  "metadata" : { },
  "name" : "vol1",
  "admin" : "hadoop",
  "owner" : "hadoop",
  "quotaInBytes" : 10,
  "quotaInCounts" : -1,
  "creationTime" : "2020-11-22T17:42:57.206Z",
  "modificationTime" : "2020-11-22T17:43:32.772081Z",
  "acls" : [ {
    "type" : "USER",
    "name" : "hadoop",
    "aclScope" : "ACCESS",
    "aclList" : [ "ALL" ]
  }, {
    "type" : "GROUP",
    "name" : "users",
    "aclScope" : "ACCESS",
    "aclList" : [ "ALL" ]
  } ],
  "usedBytes" : 0
}
bash-4.2$ 


Usage: ozone sh bucket create [-ghV] [--count-quota=<quotaInCounts>]
                              [-k=<bekName>] [--space-quota=<quotaInBytes>]
                              <value>
creates a bucket in a given volume
      <value>         URI of the volume/bucket.
                      Ozone URI could start with o3:// or without prefix. URI
                        may contain the host/serviceId  and port of the OM
                        server. Both are optional. If they are not specified it
                        will be identified from the config files.
      --count-quota=<quotaInCounts>
                      For volume this parameter represents the number of
                        buckets, and for buckets represents the number of keys
                        (eg. 5)
  -g, --enforcegdpr   if true, indicates GDPR enforced bucket,
                        false/unspecified indicates otherwise
  -h, --help          Show this help message and exit.
  -k, --bucketkey=<bekName>
                      bucket encryption key name
      --space-quota=<quotaInBytes>
                      The maximum space quota can be used (eg. 1GB)
  -V, --version       Print version information and exit.
bash-4.2$ ozone sh bucket create vol1/buck1 --quota 5BYTES
bash-4.2$ ozone sh bucket info vol1/buck1               
{
  "metadata" : { },
  "volumeName" : "vol1",
  "name" : "buck1",
  "storageType" : "DISK",
  "versioning" : false,
  "usedBytes" : 0,
  "creationTime" : "2020-11-22T17:45:14.969Z",
  "modificationTime" : "2020-11-22T17:45:14.969Z",
  "encryptionKeyName" : null,
  "sourceVolume" : null,
  "sourceBucket" : null,
  "quotaInBytes" : 5,
  "quotaInCounts" : -1
}
bash-4.2$
```

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
@@ -20,6 +20,7 @@ package org.apache.hadoop.ozone.shell;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.io.PrintStream;
 import java.util.Arrays;
 import java.util.List;
@@ -36,6 +37,7 @@ import org.apache.hadoop.ozone.MiniOzoneCluster;
 import org.apache.hadoop.ozone.MiniOzoneHAClusterImpl;
 import org.apache.hadoop.ozone.OmUtils;
 import org.apache.hadoop.ozone.OzoneConsts;
+import org.apache.hadoop.ozone.client.ObjectStore;
 import org.apache.hadoop.ozone.om.OMConfigKeys;
 import org.apache.hadoop.ozone.om.OzoneManager;
 import org.apache.hadoop.test.GenericTestUtils;
@@ -48,6 +50,8 @@ import static org.apache.hadoop.ozone.OzoneConsts.OZONE_OFS_URI_SCHEME;
 import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Assert;
+
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 import org.junit.Before;
 import org.junit.BeforeClass;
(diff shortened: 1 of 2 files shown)


adoroszlai (commented):
Thanks @ayushtkn for working on this.  Backward compatibility is addressed by this change.

Have you considered adding `--quota` as an alias for `--space-quota`, instead of as a hidden option?  In addition to requiring less code (both for option definition and in accessor method), it also prevents usage of both options in the same command.  With the current patch one can specify eg. `--space-quota 10GB --quota 1GB` and the latter is ignored.

ayushtkn:
Thanx @adoroszlai for the review.
>Have you considered adding --quota as an alias for --space-quota

I tried that, But for commands there was an option to provide an alias but for `@CommandLine.Option` I couldn't find any. One way was to add it in the names array but the value will be visible. 
Is there some way to add an alias for options as well, so that it isn't visible.
or do you suggest we add it as part of name array only like this -
` @CommandLine.Option(names = {"--space-quota","--quota"},`

adoroszlai:
> or do you suggest we add it as part of name array only like this -
> ` @CommandLine.Option(names = {"--space-quota","--quota"},`

Yes, sorry for the confusion about aliases.  Command aliases are also visible I think.

> One way was to add it in the names array but the value will be visible.
> Is there some way to add an alias for options as well, so that it isn't visible.

I don't think it's a problem that `--quota` would appear in help output as an alternative to `--space-quota`.  It's not a secret. ;)

ayushtkn:
Thanx @adoroszlai I have changed the code to add `--quota` to the names array.

adoroszlai (approved):
Thanks @ayushtkn for updating the patch.

avijayanhwx (approved):
Thanks for working on this @ayushtkn. LGTM +1.

avijayanhwx:
Thanks for the review @adoroszlai!

PR #1611, opened 2020-11-23
HDDS-4179. Implement post-finalize SCM logic.

prashantpogde:

Implement post-finalize SCM logic to allow nodes of only new version to participate in pipelines.

https://issues.apache.org/jira/browse/HDDS-4179

## How was this patch tested?

build UT. I will fix failures in CI.


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -1166,4 +1172,107 @@ public String getClusterId() {
...
+  // This function can block indefinitely till the conditions are met to safely
+  // finalize Upgrade.
+
+  public void preFinalizeUpgrade() throws IOException {
+    /**
+     * Ask pipeline manager to not create any new pipelines. Pipeline
+     * creation will remain frozen until postFinalizeUpgrade().
+     */
+    pipelineManager.freezePipelineCreation();
+
```
linyiqun, line 1244: Can we also reset SCMSafeModeManager#setInSafeMode to true? In some scenarios, we alreadt use safemode status check to reject some create type operations for new containers/pipelines.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/exceptions/SCMException.java
```diff
@@ -124,6 +124,7 @@ public ResultCodes getResult() {
     FAILED_TO_ALLOCATE_ENOUGH_BLOCKS,
     INTERNAL_ERROR,
     FAILED_TO_INIT_PIPELINE_CHOOSE_POLICY,
-    FAILED_TO_INIT_LEADER_CHOOSE_POLICY
+    FAILED_TO_INIT_LEADER_CHOOSE_POLICY,
+    FAILED_TO_FINALIZE_UPGRADE
```
avijayanhwx, line 128: I don't see any usages for this.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -268,6 +275,12 @@ public Pipeline createPipeline(ReplicationType type,
     }
     lock.writeLock().lock();
     try {
+      if (freezePipelineCreation.get()) {
+        LOG.debug("Pipeline creation is frozen while an upgrade is in " +
```
avijayanhwx, line 279: Do we want to throw an exception here? The background pipeline creator will keep trying to create pipelines, and the logs may be filled with exceptions. Is there a way to do this more cleaner? Maybe something like

_Stop pipeline creator --> Add flag in pipeline manager to stop createPipeline (for manual pipeline creation) -> Destroy existing pipelines_

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -1166,4 +1172,107 @@ public String getClusterId() {
   public HDDSLayoutVersionManager getLayoutVersionManager() {
     return scmLayoutVersionManager;
   }
+
+  private void waitForAllContainersToClose() {
```
avijayanhwx, line 1176: This may become a costly operation during finalize. If we have destroyed all pipelines, the containers are automatically closed. So, we maybe able to skip this.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
```diff
@@ -0,0 +1,271 @@
...
+
+  public BasicUpgradeFinalizer(V versionManager) {
+    this.versionManager = versionManager;
+  }
+
+  /*
+   * This method must be overriden by the component implementing the
+   * finalization logic.
+   */
+  public StatusAndMessages finalize(String upgradeClientID, T id)
```
avijayanhwx, line 62: What happens it this is invoked while the SCM (or OM) is being finalized? If it restarts, we may be OK with losing state, but how about the case when SCM is running & currently finalizing, and that time someone invokes finalize again?

avijayanhwx (commented):
Thanks for working on this @prashantpogde. A few comments inline.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
prashantpogde: That is another route but we will need stop SCM from coming out of safe mode until upgrade is complete. Also safe mode is good to enter when the SCM is starts. Here the SCM was already running and I just needed a way to pause pipeline creation for an upgrade window.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/exceptions/SCMException.java
prashantpogde: yes, left over code when I added UpgradeException. I will remove this.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
prashantpogde: OZONE_SCM_PIPELINE_CREATION_INTERVAL_DEFAULT is 2 minutes. I thought that upgrade finalization should be few minutes. Its possible that we may have few messages in the log to indicate that pipeline creation is frozen due to upgrade. But I can add a "pause" for pipeline creator as suggested.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
prashantpogde: Yes, but shouldnt we wait for all containers to close before we finalize the upgrade ? Because as part of closing the containers wemay excercise older layout version of the code otherwirse.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
linyiqun: Okay, makes sense to me.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
linyiqun: Agree the point that @avijayanhwx  mentioned. With the number of containers/pipeline becomes more large, this operation cal be very costly. can we just add a flag switch to avoid write type ops for container/pipelines, like freezePipelineCreation we added above.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
prashantpogde: yes, there is a. clientId based check. But I will add additional safeguards to prevent this.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
```diff
@@ -289,7 +292,7 @@ private void initialiseState2EventMap() {
    */
   private void initializeStateMachine() {
     stateMachine.addTransition(
-        NodeState.HEALTHY_READONLY, NodeState.HEALTHY,
+        NodeState.HEALTHY_READONLY, HEALTHY,
         NodeLifeCycleEvent.LAYOUT_MATCH);
```
fapifta, line 296: Nit: as we statically import HEALTHY_READONLY, we can leave out the qualification of NodeState here, as we do in forceNodesToHealthyReadOnly

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
```diff
@@ -0,0 +1,300 @@
...
+          "finalization, if you want to take over, and get unsent status " +
+          "messages, check -takeover option.", INVALID_REQUEST);
+    }
+  }
+
+  protected void finalizeFeature(LayoutFeature feature, Storage config)
+      throws UpgradeException {
+    Optional<? extends UpgradeAction> action = feature.onFinalizeAction();
+
+    if (action.isPresent()) {
```
fapifta, line 131: Aren't we want to do a noop when the action is not present?

fapifta:
I am thinking about the refactor of the UpgradeFinalizer.
In the BasicFinalizer, there is a finalizeFeature, that is called from the SCMUpgradeFinalizer, while the OMUpgradeFinalizer has its own finalizeFeature method.

What if we generalize this one to be the one in BasicUpgradeFinalizer along with the worker, or why don't we?
Roughly this is what I would try:
- BasicUpgradeFinalizer defines the Worker, and we encapsulate the component into a wrapper behind an interface and use that in the worker.
- the component's wrapper should have the two hooks, preFinalizeUpgrade and postFinalizeUpgrade
- if we add a getStorage method we can avoid the problem that comes from the different Storage getters as well
- we can add any other required stuff into the wrapper, and cut off the upgrade finalization logic from the component's implementation without cutting the UpgradeActions, as there we can still provide the component from the wrapper.

So at the end of the day we may get to one generic UpgradeFinalizer, that uses a let's say UpgradingComponent interface for which the implementations may be OzoneManagerComponent, and StorageContainerManagerComponent.

I started to think about this because of the similarities in the two concrete UpgradeFinalizer, and the two pretty much different approach on how we handled the Optional UpgradeAction :)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
prashantpogde: yup, will change.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
prashantpogde: why do we want to do noop ? action is not present we do not have to do anything.

prashantpogde:
>
> I am thinking about the refactor of the UpgradeFinalizer.
> In the BasicFinalizer, there is a finalizeFeature, that is called from the SCMUpgradeFinalizer, while the OMUpgradeFinalizer has its own finalizeFeature method.
....
....
> I started to think about this because of the similarities in the two concrete UpgradeFinalizer, and the two pretty much different approach on how we handled the Optional UpgradeAction :

Earlier I was keeping worker also in the Basic UpgradeFinalizer as you suggested but it seemed Om finalization didnt need pre/post hook. So left it this way. We can also discuss this in a call.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
fapifta: Sorry for this, but my wording was not correct as it seems.
Let me put it this way:
Here if action is present in the optional, we emit a noop message and return, I believe we want to emit a noop message and return if action.isPresent yields is false, so we miss a negation in the condition.

fapifta:
Even though the wrapper for OM does nothing in the pre/post hook, it might worth to write the code once and just give an empty implementation for the hooks. Sure let's discuss this one during one of our next calls.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
prashantpogde: Agreed and removed the "waitForAllContainersToClose()" call. Leaving the function around for now just in case we need it in future.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
prashantpogde: yes, I see now :)

prashantpogde:
TestOmBlockVersioning failure is unrelated. Its passing on my local setup.

fapifta:
Thanks @prashantpogde for working on this one, and addressing the review comments.
The changes look good to me, as we discussed the proposed refactoring is something we might get back to but we would defer it for now.

+1 from my side.

avijayanhwx (approved):
Thanks for addressing the comments @prashantpogde.

avijayanhwx:
CI failure is on OM side, while the changes are on HDDS side. Unable to repro this on my local setup.

PR #1613, opened 2020-11-23
HDDS-4480. Implement OM Prepare Request/Response.

errose28:

This is the first Jira to begin implementing prepare functionality for OM non rolling upgrade/downgrades. See HDDS-4470 for the design these Jiras aim to accomplish.

This PR adds the new OM request that will be used to prepare the OM for upgrade. In the applyTransaction step for this request:
1. Wait for double buffer flush to finish.
2. Take a snapshot.
3. purge the log.
The transaction log index of this request will be returned in the response to the caller.

This PR exposed a bug in the OM table cache cleanup, where the s3 table was still listed in the `OMDBDefinition` (although not used), and not listed in the `OmMetadataManager`. This resulted in an error when using the `@CleanupTableInfo(cleanupAll = true)` annotation on an `OMResponse`, because the `OzoneManagerDoubleBuffer#cleanupCache` method would get the s3 table from the `OMDBDefinition`, but fail to retrieve it from the `OmMetadataManager`. This issue has been fixed by removing the unused s3 table, and a unit test has been added to catch the issue.

Before the design of HDDS-4470 was finalized, it was initially proposed to prepare the OM using a CLI command, rather than a Ratis request, and some code to implement this was added. Since this approach will no longer be used, that code has either been removed by this PR, adapted to fit this PR, or commented out with a TODO marker so it can be adapted in later PRs for the prepare feature.

According to HDDS-4470, correctness of the prepare feature is dependent on a gate in `OzoneManagerStateMachine#preAppend`, and a marker file written to disk. These are not implemented in this PR, but TODO notes referencing them have been left that will be picked up by later Jira issues.

HDDS-4480

## How was this patch tested?

Integration tests added.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,108 @@
...
+      RaftServerProxy server = (RaftServerProxy) omRatisServer.getServer();
+      RaftServerImpl serverImpl =
+          server.getImpl(omRatisServer.getRaftGroup().getGroupId());
+
+      // Wait for outstanding double buffer entries to flush to disk,
+      // so they will not be purged from the log before being persisted to
+      // the DB.
+      RatisUpgradeUtils.waitForAllTxnsApplied(omStateMachine, serverImpl,
+          DOUBLE_BUFFER_FLUSH_TIMEOUT_SECONDS,
+          DOUBLE_BUFFER_FLUSH_CHECK_SECONDS);
```
linyiqun, line 83: Transaction entry added in double buffer is async flushed to db. Shall we pass transactionLogIndex to double check and make sure previous txns be applied.

errose28: Yes Yiqun, you are right. This was an early draft of the feature. Please see the new code and let me know if this resolves your concern.

avijayanhwx (commented):
Initial approach looks good to me. We can refine this as we make progress on the larger picture. Minor comments inline.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -0,0 +1,288 @@
...
+        waitAndCheckPrepared(om, prepareIndex);
+      }
+    }
+
+    // Restart the downed OM and wait for it to catch up.
+    // Since prepare was the last Ratis transaction, it should have all data
+    // it missed once it receives the prepare transaction.
+    cluster.restartOzoneManager(downedOM, true);
+    // Wait for other OMs to catch this one up on transactions.
+    LambdaTestUtils.await(3000, 1000,
```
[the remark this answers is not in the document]
errose28: Sure, done.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -0,0 +1,288 @@
...
+    OzoneManager downedOM = cluster.getOzoneManager(shutdownOMIndex);
+    Assert.assertFalse(downedOM.isRunning());
+
+    // Write keys with the remaining OMs up.
+    for (int i = 51; i <= 100; i++) {
+      String keyName = keyPrefix + i;
+      writeTestData(store, volumeName, bucketName, keyName);
+      writtenKeys.add(keyName);
+    }
+
```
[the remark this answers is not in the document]
errose28: Added this check on all OMs in both the tests that have transactions (even though one is commented out and does not run).

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,196 @@
...
+    boolean success = false;
+
+    while (!success && System.currentTimeMillis() < endTime) {
+      // If no transactions have been persisted to the DB, transaction info
+      // will be null, not zero, causing a null pointer exception within
+      // ozoneManager#getRatisSnaphotIndex.
+      // Get the transaction directly instead.
+      OMTransactionInfo txnInfo = ozoneManager.getMetadataManager()
+          .getTransactionInfoTable().get(TRANSACTION_INFO_KEY);
+      if (txnInfo == null) {
```
linyiqun, line 142: I prefer to get txn Info form StateMachine as well. StateMachine will be updated after new transaction flused to db. We could get latest StateMachine applied index here, related logic: [OzoneManagerDoubleBuffer.java#L340](https://github.com/apache/ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java#L340)

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -74,6 +74,7 @@ enum Type {
   DBUpdates = 53;
   FinalizeUpgrade = 54;
   FinalizeUpgradeProgress = 55;
+  Prepare = 56;
```
linyiqun, line 77: Can we rename to PrepareUpgrade which can be more readable? In the design doc, it also called PrepareUpgrade request.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -145,6 +146,7 @@ message OMRequest {
   optional DBUpdatesRequest                  dbUpdatesRequest              = 53;
   optional FinalizeUpgradeRequest           finalizeUpgradeRequest         = 54;
   optional FinalizeUpgradeProgressRequest   finalizeUpgradeProgressRequest = 55;
+  optional PrepareRequest                   prepareRequest                 = 56;
```
linyiqun, line 149: Can we rename to PrepareUpgradeRequest/PrepareUpgradeResponse?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,196 @@
...
+
+      OzoneManagerRatisServer omRatisServer = ozoneManager.getOmRatisServer();
+      RaftServerProxy server = (RaftServerProxy) omRatisServer.getServer();
+      RaftServerImpl serverImpl =
+          server.getImpl(omRatisServer.getRaftGroup().getGroupId());
+
+      takeSnapshotAndPurgeLogs(serverImpl);
+
+      // TODO: Create marker file with txn index.
+
```
linyiqun, line 112: Not a comment for this PR, another TODO thing I am thinking: After this prepare request be executed, OM should  reject subsequent requests immediately, how do we plan to do this, exit OM?

errose28: Subsequent requests will be rejected by the leader using a gate in the preAppend step. This will will be done in another Jira. See HDDS-4470 for the design document and flow diagram of the overall prepare approach. This also documents how the upgrade startup flag or cancel prepare request (to be implemented in following Jiras) can be used to take the OM out of prepare mode.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
errose28: The design doc and flow diagram are a bit inconsistent with usage of PrepareRequest or PrepareUpgradeRequest. We were initially planning on calling it PrepareUpgradeRequest during the design phase, but later decided that the command is not actually specific to upgrades. It will also be used for downgrades, and it is really a more general operation to clear out the OM logs, without any upgrade/downgrade specific actions in it. We opted to keep the name generic to reflect this.

@avijayanhwx let me know if I missed anything here, and your current thoughts on the prepare vs. prepareUpgrade naming.

avijayanhwx: +1 to @errose28's comment. We can keep it "prepare" as a generic operation that can be used during upgrade, downgrade as well as creating checkpoint for supporting "rollbacks" in the future.

linyiqun: Okay, get it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
linyiqun: Okay, will have a look for this. Thanks, @errose28 .

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
linyiqun: Ignored this, see below comments about this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,196 @@
...
+ * in the log is applied to the database in the old version of the code in one
+ * OM, and the new version of the code on another OM.
+ */
+public class OMPrepareRequest extends OMClientRequest {
+  private static final Logger LOG =
+      LoggerFactory.getLogger(OMPrepareRequest.class);
+
+  // Allow double buffer this many seconds to flush all transactions before
+  // returning an error to the caller.
+  private static final Duration DOUBLE_BUFFER_FLUSH_TIMEOUT =
```
swagle, line 61: Minor nit: Why isn't this externally configurable? Since we throw an IOException after waiting. For non-rolling upgrades probably unnecessary hence a minor nit.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,196 @@
...
+
+      // Wait for outstanding double buffer entries to flush to disk,
+      // so they will not be purged from the log before being persisted to
+      // the DB.
+      // Since the response for this request was added to the double buffer
+      // already, once this index reaches the state machine, we know all
+      // transactions have been flushed.
+      waitForDoubleBufferFlush(ozoneManager, transactionLogIndex);
+
+      OzoneManagerRatisServer omRatisServer = ozoneManager.getOmRatisServer();
```
swagle, line 104: Minor nit: unused local references

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,196 @@
...
+   * @param impl RaftServerImpl instance
+   * @throws IOException on Error.
+   */
+  public static long takeSnapshotAndPurgeLogs(RaftServerImpl impl)
+      throws IOException {
+
+    StateMachine stateMachine = impl.getStateMachine();
+    long snapshotIndex = stateMachine.takeSnapshot();
+    if (snapshotIndex != stateMachine.getLastAppliedTermIndex().getIndex()) {
+      throw new IOException("Index from Snapshot does not match last applied " +
```
swagle, line 169: Can we add a comment here: What is the recovery step from this? If the snapshot index does not match will subsequent prepare ever succeed? cc: @avijayanhwx

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
errose28: I can make this an external config if we think it is necessary. @avijayanhwx any thoughts here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
errose28: Which reference is unused, `omRatisServer`? It is used in the following two lines to construct the `RaftServerImpl`.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
errose28: If this fails, prepare will fail and an error response will be sent back to the client. I will add a comment to address this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
swagle: server and serverImpl do not seem to be used, only a code readability comment, it makes the method more concise

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
avijayanhwx: We can add this as a passed in client side param, rather than an OM config. Something like 
`ozone admin om --prepare --wait-for-flush=2m`

This will be added in subsequent JIRAs.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
errose28: server is used to obtain serverImpl, which is passed to takeSnapshotAndPurgeLogs. I could collapse it, not sure if its more readable. It would look like this:
``` 
takeSnapshotAndPurgeLogs(((RaftServerProxy) omRatisServer.getServer())
          .getImpl(omRatisServer.getRaftGroup().getGroupId()));
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
errose28: Actually @avijayanhwx suggested offline we should probably make this a client side param later. Something like `ozone om prepare --waitFor=1m`.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,204 @@
...
+  public static long takeSnapshotAndPurgeLogs(RaftServerImpl impl)
+      throws IOException {
+
+    StateMachine stateMachine = impl.getStateMachine();
+    long snapshotIndex = stateMachine.takeSnapshot();
+
+    // If the snapshot indices from Ratis and the state machine do not match,
+    // the exception is propagated, resulting in an error response to the
+    // client. They can retry the prepare request.
+    if (snapshotIndex != stateMachine.getLastAppliedTermIndex().getIndex()) {
```
bharatviswa504, line 176: I believe there is a chance of snapshotIndex != lastAppliedTermIndex.

Because once doubleBuffer flush completes, it calls updateLastAppliedIndex from DoubleBuffer flush thread, to update lastAppliedIndex, but if doubleBuffer flush thread has not completed updateLastAppliedIndex updating, then this might not be equal.

There is very little chance to happen, as there is takeSnapshot which flush to DB, as updateLastAppliedIndex is inmemory map update.

So, to be safer side check lastAppliedTermIndex is the same as the TransactionInfo table lastAppliedIndex in the waitForDoubleBufferFlush method.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -0,0 +1,326 @@
...
+  /**
+   * Writes data to the cluster.
+   * Shuts down one OM.
+   * Writes more data to the cluster.
+   * Submits prepare as ratis request.
+   * Checks that two live OMs are prepared.
+   * Revives the third OM
+   * Checks that third OM received all transactions and is prepared.
+   * @throws Exception
+   */
```
bharatviswa504, line 154: Is this failing test that needs more changes?

bharatviswa504 (commented):
Approach LGTM, few comments/questions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,204 @@
...
+   * @param impl RaftServerImpl instance
+   * @throws IOException on Error.
+   */
+  public static long takeSnapshotAndPurgeLogs(RaftServerImpl impl)
+      throws IOException {
+
+    StateMachine stateMachine = impl.getStateMachine();
+    long snapshotIndex = stateMachine.takeSnapshot();
+
+    // If the snapshot indices from Ratis and the state machine do not match,
```
bharatviswa504, line 173: Offline discussion with @avijayanhwx this will be revisited in next jiras, if another PrepareUpgrade request is let in, we cleanup logIndexes greater than snapshotIndex and also syncWithSnapshot does not close when end logIndex is greater than snapshotIndex, so that might be left out.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
swagle: Since typecasting is involved, lets leave this the way it is. It is more of a personal choice. Since the reference is utilized only once, I think it make sense to eliminate the reference, probably an interesting discussion for some other time :) Marking this as resolved.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,204 @@
...
+  // Time between checks to see if double buffer finished flushing.
+  private static final Duration DOUBLE_BUFFER_FLUSH_CHECK_INTERVAL =
+      Duration.of(1, ChronoUnit.SECONDS);
+
+  public OMPrepareRequest(OMRequest omRequest) {
+    super(omRequest);
+  }
+
+  @Override
+  public OMClientResponse validateAndUpdateCache(
```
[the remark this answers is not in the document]
avijayanhwx: Yes, in follow up JIRAs, the 'Prepared' OM state maintenance and the gate for stopping other requests will be added.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
errose28: The comment I put above this check is actually incorrect. This check is redundant since it is checking Ratis's in memory snapshot index against itself. It does not touch the transaction info table in the DB. I will remove this check.

> check lastAppliedTermIndex is the same as the TransactionInfo table lastAppliedIndex in the waitForDoubleBufferFlush method

I believe the check we have in the waitForDoubleBufferFlush method provides the same guarantee. Let me know if there is any added benefit to putting this check there as well.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,204 @@
...
+      // already, once this index reaches the state machine, we know all
+      // transactions have been flushed.
+      waitForDoubleBufferFlush(ozoneManager, transactionLogIndex);
+
+      OzoneManagerRatisServer omRatisServer = ozoneManager.getOmRatisServer();
+      RaftServerProxy server = (RaftServerProxy) omRatisServer.getServer();
+      RaftServerImpl serverImpl =
+          server.getImpl(omRatisServer.getRaftGroup().getGroupId());
+
+      takeSnapshotAndPurgeLogs(((RaftServerProxy) omRatisServer.getServer())
```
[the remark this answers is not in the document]
errose28: Will do

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,204 @@
...
+
+    while (!success && System.currentTimeMillis() < endTime) {
+      // If no transactions have been persisted to the DB, transaction info
+      // will be null, not zero, causing a null pointer exception within
+      // ozoneManager#getRatisSnaphotIndex.
+      // Get the transaction directly instead.
+      OMTransactionInfo txnInfo = ozoneManager.getMetadataManager()
+          .getTransactionInfoTable().get(TRANSACTION_INFO_KEY);
+      if (txnInfo == null) {
+        success = (txnLogIndex == 0);
```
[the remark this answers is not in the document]
errose28: This left over from an earlier implementation where the prepare request was not part of the snapshot and there could be no transactions in the table. Now that we add the response to the double buffer before calling this method, this extra logic can be removed and we can just check the transaction info table against our log index.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/upgrade/OMPrepareResponse.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.hadoop.ozone.om.response.CleanupTableInfo;
+import org.apache.hadoop.ozone.om.response.OMClientResponse;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
+
+import java.io.IOException;
+
+/**
+ * Response for prepare request.
+ */
+@CleanupTableInfo(cleanupAll = true)
```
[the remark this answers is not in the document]
errose28: Yes, we can actually just set this to false.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
errose28: I could not get this test to work without client side changes. Once those are implemented by @avijayanhwx, we will use them to implement more robust integration tests with failures. This is left here for reference to come back to.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -0,0 +1,204 @@
...
+    // client. They can retry the prepare request.
+    if (snapshotIndex != stateMachine.getLastAppliedTermIndex().getIndex()) {
+      throw new IOException("Index from Snapshot does not match last applied " +
+          "Index");
+    }
+
+    RaftLog raftLog = impl.getState().getLog();
+    // In order to get rid of all logs, make sure we also account for
+    // intermediate Ratis entries that do not pertain to OM.
+    long lastIndex = Math.max(snapshotIndex,
```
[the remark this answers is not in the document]
errose28: This was originally done defensively against meta transactions that Ratis may put in the log, which are not related to OM activities. @avijayanhwx and I are unable to reproduce earlier errors we thought we saw related to this, so we will remove this max statement and just use snapshotIndex. If an error surfaces later under more robust prepare and upgrade testing, we will re-address this issue, but for now we can remove it to avoid unintended side effects this might cause.

errose28:
Thanks for the comments @bharatviswa504. Everything should be addressed in the latest version of the code. Please let me know if you have further questions/concerns.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
errose28: I understand the issue now, sorry for the confusion about the original code. It looks like this did surface on the most recent CI integration test run. I believe we should be waiting for both the lastAppliedIndex (Ratis's view) and the transaction index in the DB to update to the prepare request's index. There is no guarantee whether one or the other is updated first.

bharatviswa504 (commented):
I have a few more comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -166,36 +166,29 @@ private static void waitForDoubleBufferFlush(
...
+    RaftLog raftLog = impl.getState().getLog();
+    long raftLogIndex = raftLog.getLastEntryTermIndex().getIndex();
 
-    // If the snapshot indices from Ratis and the state machine do not match,
-    // the exception is propagated, resulting in an error response to the
-    // client. They can retry the prepare request.
-    if (snapshotIndex != stateMachine.getLastAppliedTermIndex().getIndex()) {
-      throw new IOException("Index from Snapshot does not match last applied " +
-          "Index");
+    if (snapshotIndex != raftLogIndex) {
```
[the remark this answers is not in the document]
avijayanhwx: @bharatviswa504 That will be the next work item in the prepare OM feature.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -137,16 +137,21 @@ private static void waitForDoubleBufferFlush(
...
-      // Get the transaction directly instead.
-      OMTransactionInfo txnInfo = ozoneManager.getMetadataManager()
+      // Get the transaction directly instead to handle the case when it is
+      // null.
+      OMTransactionInfo dbTxnInfo = ozoneManager.getMetadataManager()
           .getTransactionInfoTable().get(TRANSACTION_INFO_KEY);
-      if (txnInfo == null) {
-        success = (txnLogIndex == 0);
+      if (dbTxnInfo == null) {
+        // If there are no transactions in the DB, we are prepared to log
```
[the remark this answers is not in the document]
bharatviswa504: Ohh misread the condition indexToWaitFor==0, indexTowaitFor will never be zero, so in case when doubleBuffer flush fails to flush this condition is not met. Is my understanding correct?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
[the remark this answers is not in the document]
bharatviswa504: Makes sense, thanks for the clear explanation.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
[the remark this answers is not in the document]
bharatviswa504: Or simply here, we can have if dbTxnInfo success=false right?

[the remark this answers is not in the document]
errose28: Right, the condition should never be met. Because of that, we could just set `success = false` when the dbTxnInfo is null, but since we have to do the null check anyways, I figured it was better to add a line to handle the case correctly, even if it should never arise.

bharatviswa504 (approved):

bharatviswa504 (approved):
+1 LGTM

avijayanhwx (approved):
Thanks for working on this @errose28. This is good for the first version of the Prepare request. In subsequent patches, edge cases can be tested and fixed.

avijayanhwx:
Thanks for the review @linyiqun, @swagle & @bharatviswa504.

PR #1614, opened 2020-11-24
HDDS-4502. [OFS] Listing volumes under root should return all volumes whenever possible

smengcl:
https://issues.apache.org/jira/browse/HDDS-4502

Currently `listStatusRoot()` only lists volumes owned(ACL disabled) or accessible(ACL enabled) by current user.

**This prevents HttpFS from listing all volumes under OFS root.**

Since we probably can't provide an argument to allow passing in parameter `--all` due to conforming to HCFS, we should default to list ALL volumes whenever this is possible.

Note: `ozone.om.volume.listall.allowed` is an OM side argument that controls whether any user can list all volumes on an Ozone cluster, it defaults to `true`.

## TBD

- [ ] Check how OFS list root would behave when OM has `ozone.om.volume.listall.allowed` = `false` when ACL is enabled.

## Testing

Pending CI. May need to add new UT to check whether the change breaks any existing UTs.


smengcl:
When ACL is enabled, it should check for accessible directories (instead of owner name). To test with Ranger.

PR #1615, opened 2020-11-24
HDDS-4451.Handle start & stop of Trash Emptier thread when node becom…

sadanand48:
…es leader/follower

Add handling for when to start/stop thread when OM switches from follower to leader or vice-versa.

https://issues.apache.org/jira/browse/HDDS-4451

## How was this patch tested?
Manually.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -57,8 +57,6 @@ import org.apache.hadoop.crypto.key.KeyProviderCryptoExtension;
 import org.apache.hadoop.fs.CommonConfigurationKeys;
 import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
 import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.Trash;
-import org.apache.hadoop.fs.TrashPolicy;
 import org.apache.hadoop.hdds.HddsConfigKeys;
 import org.apache.hadoop.hdds.HddsUtils;
 import org.apache.hadoop.hdds.annotation.InterfaceAudience;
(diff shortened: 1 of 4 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -98,7 +105,8 @@ public Runnable getEmptier() throws IOException {
 
     @Override
     public void run() {
-      if (emptierInterval == 0) {
+      // if this is not the leader node,don't run the emptier
+      if (emptierInterval == 0 || !om.isLeader()) {
```
mukul1987, line 109: I think this should happen after the thread.sleep at line 117. As for most of the cases, when an OM restart the trash emptier it should check this status and skip doing any Trash processing.

mukul1987: Also lets see if we can add a unit test for this change. Please check with @bharatviswa504 on the test

mukul1987 (changes requested):

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashPolicyOzone.java
```diff
@@ -107,6 +136,9 @@ public void run() {
         end = ceiling(now, emptierInterval);
         try {                                     // sleep for interval
           Thread.sleep(end - now);
+          if (!om.isLeader()){
+            return;
```
mukul1987, line 140: This should be continue here ?

mukul1987 (changes requested):

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1402,10 +1394,7 @@ public void stop() {
       }
       // TODO:Also stop this thread if an OM switches from leader to follower.
```
mukul1987, line 1395: Lets remove this TODO's

mukul1987 (approved):

mukul1987 (approved):

PR #1618, opened 2020-11-24
HDDS-3363. Intermittent failure in testContainerImportExport

adoroszlai:

Container DBs are compacted during export for replication.  This opens the DB, but does not close it, only decrements the reference count.  So RocksDB keeps running and may modify files while they are being archived.

This change makes sure the DB is closed after compaction to avoid concurrent writes.

https://issues.apache.org/jira/browse/HDDS-3363

## How was this patch tested?

Ran unit test 100x: [failed 6% without the fix](https://github.com/adoroszlai/hadoop-ozone/runs/1447259621), but [0% with it](https://github.com/adoroszlai/hadoop-ozone/runs/1447198023).

Regular CI: https://github.com/adoroszlai/hadoop-ozone/actions/runs/380843788


sodonnel (approved):
The changes LGTM. +1.

bshashikant (approved):

adoroszlai:
Thanks @sodonnel and @bshashikant for the review.

PR #1619, opened 2020-11-24
HDDS-4503. Provide info on block size via FileSystem

adoroszlai:

Provide information about Ozone's configured block size via Hadoop-compatible file system implementations by overriding `getDefaultBlockSize()` from `FileSystem`.

Ozone does not make new files visible until they are closed.  This change allows Impala to query the actual configured block size and use this info during file write, instead of relying on the default block size, which can be overridden in config.

https://issues.apache.org/jira/browse/HDDS-4503

## How was this patch tested?

Added unit test.

### DIFF
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -21,6 +21,7 @@ package org.apache.hadoop.fs.ozone;
 import com.google.common.base.Preconditions;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.conf.StorageUnit;
 import org.apache.hadoop.fs.BlockLocation;
 import org.apache.hadoop.fs.CreateFlag;
 import org.apache.hadoop.fs.FSDataInputStream;
@@ -71,6 +72,8 @@ import static org.apache.hadoop.fs.ozone.Constants.OZONE_DEFAULT_USER;
 import static org.apache.hadoop.fs.ozone.Constants.OZONE_USER_DIR;
 import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_FS_ITERATE_BATCH_SIZE;
 import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_FS_ITERATE_BATCH_SIZE_DEFAULT;
+import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_SCM_BLOCK_SIZE;
+import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_SCM_BLOCK_SIZE_DEFAULT;
 import static org.apache.hadoop.ozone.OzoneConsts.OZONE_URI_DELIMITER;
 import static org.apache.hadoop.ozone.OzoneConsts.OZONE_URI_SCHEME;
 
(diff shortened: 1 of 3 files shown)


hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicRootedOzoneFileSystem.java
```diff
@@ -723,6 +726,12 @@ public boolean mkdirs(Path f, FsPermission permission) throws IOException {
     return mkdir(f);
   }
 
+  @Override
+  public long getDefaultBlockSize() {
+    return (long) getConf().getStorageSize(
+        OZONE_SCM_BLOCK_SIZE, OZONE_SCM_BLOCK_SIZE_DEFAULT, StorageUnit.BYTES);
```
ayushtkn, line 732: Is this block size used by client while writing? In create and createNonRecursive it tends to ignore the specified block size. 
So, if it isn't correct, Isn't it same as not implementing it? `FileSystem` will anyway return some value, that can be tweaked as well by `fs.local.block.size`

adoroszlai: > In create and createNonRecursive it tends to ignore the specified block size.

Ozone also ignores it and uses uniform block size on server side (configurable via this `OZONE_SCM_BLOCK_SIZE` property).

> can be tweaked as well by `fs.local.block.size`

You are right, we could achieve the same result by forcing this config to the same value as SCM block size.  I think it's simpler and cleaner to override the method than to keep the config in sync.

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -698,6 +701,12 @@ public boolean mkdirs(Path f, FsPermission permission) throws IOException {
     return mkdir(f);
   }
 
+  @Override
+  public long getDefaultBlockSize() {
+    return (long) getConf().getStorageSize(
+        OZONE_SCM_BLOCK_SIZE, OZONE_SCM_BLOCK_SIZE_DEFAULT, StorageUnit.BYTES);
```
arp7, line 707: What if the client who wrote the block and the client who is querying have a different configuration?

adoroszlai: This is intended to be used by the client that creates the file.  Once the file is closed, the actual block size can be queried via `getBlockSize(Path)`.  Currently this method returns 32MB, regardless of SCM config:

https://github.com/apache/hadoop/blob/a6df05bf5e24d04852a35b096c44e79f843f4776/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L2521-L2522

adoroszlai: Thanks @arp7 for taking a look.  Does the above answer your quetion, or would you like further clarification?

arp7: Yes, thanks. Sorry i missed responding.

arp7 (approved):
+1

adoroszlai:
Thanks @arp7 for reviewing and committing it.  Thanks @ayushtkn for the review.

PR #1622, opened 2020-11-24
HDDS-4501. Reload OM State fail should terminate OM for any exceptions.

bharatviswa504:

During reload state for any exception terminate OM, and also cleanup DB backup folder.

https://issues.apache.org/jira/browse/HDDS-4501

## How was this patch tested?

Existing tests, this issue is idenitified from one of our customer issues, due to this follower OM went into bad state.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -3324,8 +3324,16 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
       omRatisServer.getOmStateMachine().unpause(lastAppliedIndex, term);
       LOG.info("Reloaded OM state with Term: {} and Index: {}", term,
           lastAppliedIndex);
-    } catch (IOException ex) {
+    } catch (Exception ex) {
       String errorMsg = "Failed to reload OM state and instantiate services.";
+      // Delete the backup DB if exists and then terminate OM.
+      try {
+        if (dbBackup != null) {
+          FileUtils.deleteFully(dbBackup);
+        }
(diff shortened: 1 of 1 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3324,8 +3324,16 @@ TermIndex installCheckpoint(String leaderId, Path checkpointLocation,
...
+    } catch (Exception ex) {
       String errorMsg = "Failed to reload OM state and instantiate services.";
+      // Delete the backup DB if exists and then terminate OM.
+      try {
+        if (dbBackup != null) {
+          FileUtils.deleteFully(dbBackup);
+        }
+      } catch (Exception e) {
+        LOG.error("Failed to delete the backup of the original DB {}", dbBackup);
+      }
```
hanishakoneru, line 3337: Bharat, in this scenario where the OM state was not reloaded with new checkpoint, it would be better to keep the backup also in place. Just in case the OM needs to be manually reversed back to old state.

bharatviswa504: Ya thought of it, but my thought process here is when OM restarts it will get a new install Snapshot and try again. Because when OM it starts, anyway it cannot use this backup path, until some one manual replacing.

But I think to be on safer side, will remove this code for now.

bharatviswa504: Done.

hanishakoneru (approved):
Thanks @bharatviswa504 for finding and fixing this.
+1 pending CI.

bharatviswa504:

PR #1623, opened 2020-11-25
HDDS-4508. Add total container count to SCMContainerMetrics

sky76093016:

Add total container count to SCMContainerMetrics.

https://issues.apache.org/jira/browse/HDDS-4508

## How was this patch tested?

Tests passed.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMContainerMetrics.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMContainerMetrics.java
@@ -67,9 +67,9 @@ public class SCMContainerMetrics implements MetricsSource {
   public void getMetrics(MetricsCollector collector, boolean all) {
     Map<String, Integer> stateCount = scmmxBean.getContainerStateCount();
 
-    int TotalContainers = 0;
-    for(HddsProtos.LifeCycleState state : HddsProtos.LifeCycleState.values()){
-      TotalContainers = TotalContainers + stateCount.get(state.toString());
+    int totalContainers = 0;
+    for (HddsProtos.LifeCycleState state : HddsProtos.LifeCycleState.values()) {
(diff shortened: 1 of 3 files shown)


xiaoyuyao (approved):
LGTM, +1.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMContainerMetrics.java
```diff
@@ -84,6 +84,14 @@ public void getMetrics(MetricsCollector collector, boolean all) {
...
         .addGauge(Interns.info("DeletedContainers",
             "Number of containers in deleted state"),
+            stateCount.get(DELETED.toString()))
+        .addGauge(Interns.info("TotalContainers",
+            "Number of all containers"),
+          stateCount.get(OPEN.toString())+
+            stateCount.get(CLOSING.toString())+
+            stateCount.get(QUASI_CLOSED.toString())+
+            stateCount.get(CLOSED.toString())+
+            stateCount.get(DELETING.toString())+
```
adoroszlai, line 94: Nit: Can this be converted to a loop over all states summing individual values?  That way if a new state is introduced, no further change in counting totals will be required.

adoroszlai (changes requested):
Thanks @sky76093016 for working on this.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMContainerMetrics.java
sky76093016: Thanks @adoroszlai  for the suggestion,I tried to modify it.

adoroszlai (approved):
Thanks @sky76093016 for updating the patch.

adoroszlai:
Thanks @sky76093016 for the contribution, @xiaoyuyao for the review.

PR #1624, opened 2020-11-25
HDDS-4491. Deprecate raft.server.rpcslowness.timeout config key

wycccccc:
Add old config key `rpcslowness.timeout` to handle deprecated configs

https://issues.apache.org/jira/browse/HDDS-4491

## How was this patch tested?
Added unit test.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
@@ -157,7 +157,7 @@ public final class OzoneQuota {
     }
 
     String uppercase = quotaInBytes.toUpperCase()
-       .replaceAll("\\s+", "");
+        .replaceAll("\\s+", "");
     String size = "";
     long nSize = 0;
     Units currUnit = Units.MB;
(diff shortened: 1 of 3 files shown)


adoroszlai (changes requested):
Thanks @wycccccc for working on this.  Unfortunately the patch is not enough to handle deprecated configs.

1. `rpc.slowness.timeout` is only part of the config key name.  It is defined in https://github.com/apache/ozone/blob/a4cd12c8567245699ba5c4aea6d1865d8c299017/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/conf/DatanodeRatisServerConfig.java#L91 and should be used together with its prefix, which comes from https://github.com/apache/ozone/blob/a4cd12c8567245699ba5c4aea6d1865d8c299017/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/conf/DatanodeRatisServerConfig.java#L34-L35
2. The `ConfigurationProvider` changed in this patch is for Recon, but the config key belongs to Datanode, another component of Ozone.  The `DeprecationDelta` should be added somewhere in `org.apache.hadoop.hdds.conf` in `hadoop-hdds`.  (Sidenote: I think `DatanodeRatis*Config` classes belong to `hadoop-hdds/container-service`, not `hadoop-hdds/framework`.)

wycccccc:
Thank @adoroszlai for your suggestion, I will try to modify it correctly.

adoroszlai (commented):
Thanks @wycccccc for updating the patch.  The change looks good to achieve the desired goal.  I think cohesion might be improved by adding the deprecation in `RaftServerConfigKeys` instead of `OzoneConfiguration`, but it may have downsides, too, so I'm OK with the current version.

Can you please fix [checkstyle violations](https://github.com/apache/ozone/pull/1624/checks?check_run_id=1457661867#step:4:4)?

wycccccc:
I have fixed checkstyle violations.Thanks @adoroszlai for your suggestion, it helped me a lot.

adoroszlai (approved):

adoroszlai:
Thanks @wycccccc for the contribution.

PR #1629, opened 2020-11-26
HDDS-4520. [DOC]Add initialization configuration to ozone java API doc.

captainzmc:

Currently, the usage is described in Ozone API Doc, not how to initialize the configuration. Without this section, some users cannot run the sample code based on Doc. So we need to add to that.

https://issues.apache.org/jira/browse/HDDS-4520

## How was this patch tested?

update doc needn't UT

### DIFF
--- a/hadoop-hdds/docs/content/interface/JavaApi.md
+++ b/hadoop-hdds/docs/content/interface/JavaApi.md
@@ -35,19 +35,19 @@ Before creating an ozone client, you need to initialize the configuration by cre
 OzoneConfiguration conf = new OzoneConfiguration();
 {{< /highlight >}}
 
-If HA not enabled, just set the OM address (you need to change the host in the example to the address of your OM):
+If HA not enabled, just set the OM address (Need to change the correct host):
 
 {{< highlight java >}}
 conf.set("ozone.om.address", "ozone.master.host:9862");
 {{< /highlight >}}
 
-If HA enabled, all OM addresses need to be set (you need to change the host in the example to the address of your OM)：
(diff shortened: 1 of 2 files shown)


adoroszlai (changes requested):
Thanks @captainzmc for proposing improvement to the documentation.

I think config in files separate from regular code is preferred for maintainability (ie. `ozone-site.xml`, which is picked up from the classpath).  Instead of an example that hard-codes addresses in the code, I think we should provide an overview page about the XML-based config.

Config specific to OM HA is described in _Features > High Availability_, while required config for regular OM is listed in _Architecture > Ozone Manager_.

captainzmc:
> Thanks @captainzmc for proposing improvement to the documentation.
> 
> I think config in files separate from regular code is preferred for maintainability (ie. `ozone-site.xml`, which is picked up from the classpath). Instead of an example that hard-codes addresses in the code, I think we should provide an overview page about the XML-based config.
> 
> Config specific to OM HA is described in _Features > High Availability_, while required config for regular OM is listed in _Architecture > Ozone Manager_.

Agree, when I personally run the Ozone API, I usually put configuration files in my environment (such as Ozone -site.xml) under the Classpath resource, so that no configuration is required in the code. I only ask people to write the OM configuration to their code if they don't want to use the configuration file.

Perhaps we can emphasize here that we prefer to use a configuration file to load the configuration of the client. we can also  lists the necessary configuration items in ozone-site.xml on the client.

captainzmc:
Let's close this PR first and then we'll provide an overview page about the based xml config.

PR #1631, opened 2020-11-26
HDDS-4185. Disable IncrementalByteBuffer by default in Ozone Client

elek:

Disable (by default) the incremental byte buffer usage due to performance reasons and allocate full buffer in each steps.

In the previous PR we agreed with @bshashikant to use a configuration variable instead of removing it. 

https://github.com/apache/ozone/pull/1374#issuecomment-696169868

https://issues.apache.org/jira/browse/HDDS-4185

## How was this patch tested?

Tested with local docker cluster and with CI.

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/OzoneClientConfig.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/OzoneClientConfig.java
@@ -55,7 +55,7 @@ public class OzoneClientConfig {
   private int streamBufferSize = 4 * 1024 * 1024;
 
   @Config(key = "stream.buffer.increment",
-      defaultValue = "0",
+      defaultValue = "0B",
       type = ConfigType.SIZE,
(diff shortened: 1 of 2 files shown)


hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/OzoneClientConfig.java
```diff
@@ -54,6 +54,18 @@
       tags = ConfigTag.CLIENT)
   private int streamBufferSize = 4 * 1024 * 1024;
 
+  @Config(key = "stream.buffer.increment",
+      defaultValue = "0",
```
adoroszlai, line 58:
```suggestion
      defaultValue = "0B",
```

(Unfortunately "0" is not valid for `StorageSize`).

elek:
@bshashikant we already discussed it in #1374, but can you please confirm if you are fine with this approach?

bshashikant (approved):

PR #1634, opened 2020-11-30
HDDS-4505. Increase default value for SCM heartbeat timeout to 5s.

lokeshj1703:

On increasing the block deletion limit for SCM from 10000 blocks per interval to 100000 blocks, the heartbeats were timing out in the datanodes. Current rpc timeout configured for SCM heartbeat is 1 second. The jira proposes to increase the rpc timeout to 5 seconds.
With 5 second configuration timeouts were not seen by the datanodes.

https://issues.apache.org/jira/browse/HDDS-4505

## How was this patch tested?

Existing UT.


adoroszlai (commented):
Thanks @lokeshj1703 for working on this.  *Default* block deletion limit is still 10000, so I don't see why *default* heartbeat timeout should be increased.  It can be configured by admins who change the limit manually.  Maybe the description for block deletion limit config can mention how these settings are related.

https://github.com/apache/ozone/blob/19dd94d594cfab0308b778dba367ecc62ed0f47e/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfig.java#L80-L82

nandakumar131 (approved):

PR #1636, opened 2020-11-30
HDDS-4496. Separate client and server2server GRPC services of datanode

elek:

Today both the closed-container-replication service and datanode client service are exposed on the same datanode port which makes it impossible to use different network configuration (like mutual TLS).

In this patch I propose to separate the two services and use two network ports.

https://issues.apache.org/jira/browse/HDDS-4496

## How was this patch tested?

Tested the replication with local docker-compose cluster (adding keys, closing container, waiting for replication)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/ReplicationServer.java
```diff
@@ -0,0 +1,147 @@
...
+    } catch (InterruptedException ex) {
+      LOG.warn("{} couldn't be stopped gracefully", getClass().getSimpleName());
+    }
+  }
+
+  public int getPort() {
+    return port;
+  }
+
+  @ConfigGroup(prefix = "datanode.replication")
```
xiaoyuyao, line 129: To be consistent with existing replication related configurations, should we move this Configuration key as part of DatanodeConfiguration.java class under the prefix as hdds.datanode.replication.*?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/ReplicationServer.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.ratis.thirdparty.io.grpc.ServerInterceptors;
+import org.apache.ratis.thirdparty.io.grpc.netty.GrpcSslContexts;
+import org.apache.ratis.thirdparty.io.grpc.netty.NettyServerBuilder;
+import org.apache.ratis.thirdparty.io.netty.handler.ssl.ClientAuth;
+import org.apache.ratis.thirdparty.io.netty.handler.ssl.SslContextBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Separated network server for server2server container replication.
```
xiaoyuyao, line 44: NIT: separate line 44 from the *.

xiaoyuyao (commented):
Thanks @elek for working on this. The PR LGTM overall. A few comments added inline.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/ReplicationServer.java
```diff
@@ -0,0 +1,147 @@
...
+
+    server = nettyServerBuilder.build();
+  }
+
+  public void start() throws IOException {
+    server.start();
+
+    port = server.getPort();
+
+    if (port == 0) {
```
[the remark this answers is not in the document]
elek: That's my fault. This is in the wrong order. here the `server.getPort()` should never be 0. `port` includes the final port number which is used in  `datanodeDetails.setPort(Name.REPLICATION, replicationServer.getPort());`. 

But it was not clear as I used the wrong order. port check should be done before setting the variable. Fixed with:

```
    if (port == 0) {
      LOG.info("{} is started using port {}", getClass().getSimpleName(), server.getPort());
    }

    port = server.getPort();
```

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/ReplicationServer.java
elek: YES. Thanks the suggestion.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/ReplicationServer.java
elek: Sorry, I don't get it. Seems to be a normal Javadoc for me. Must be a typo but I don't see it. (can be the reason why I wrote it ;-) )

xiaoyuyao (commented):
LGTM, +1 pending CI

elek:
CI is green. Merging it now. Thanks for the review @xiaoyuyao

PR #1637, opened 2020-11-30
HDDS-4526. Remove false-positive error logs from LeaseManager

elek:

There is an error log which can be seen frequently in the log:

```
2020-11-30 15:35:58,539 [CommandWatcher-LeaseManager#LeaseMonitor] ERROR lease.LeaseManager (LeaseManager.java:run(238)) - Execution was interrupted 
java.lang.InterruptedException: sleep interrupted
	at java.lang.Thread.sleep(Native Method)
	at org.apache.hadoop.ozone.lease.LeaseManager$LeaseMonitor.run(LeaseManager.java:234)
```

This log is introduced by HDDS-2561 to make Sonar happy, but it was not required as LeaseManager use the thread interrupt intentionally.

For a proper fix we can keep the logging on WARN level AND replace thread interrupts with thread notify/wait.

https://issues.apache.org/jira/browse/HDDS-4526

## How was this patch tested?

Unit test. `TestLeaseManager` has good coverage.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lease/LeaseManager.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/lease/LeaseManager.java
@@ -17,9 +17,6 @@
 
 package org.apache.hadoop.ozone.lease;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.util.List;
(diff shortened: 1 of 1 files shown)


adoroszlai (approved):
LGTM.

adoroszlai:
Thanks @elek for fixing this.

PR #1638, opened 2020-11-30
HDDS-4484. Use RaftServerImpl isLeader instead of periodic leader update logic in OM and isLeaderReady for read/write requests

bharatviswa504:

Use RatisServer isLeader and isLeaderReady API from RaftServerImpl Class.

And also for reading requests, the check should be isLeaderReady before serving read requests.

https://issues.apache.org/jira/browse/HDDS-4484

## How was this patch tested?

Existing tests in TestOzoneManagerHA*.

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1644,17 +1644,6 @@
   </property>
 
   <property>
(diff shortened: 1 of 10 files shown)


bharatviswa504:
Using direct RaftServerImpl API's for now until RATIS-1181 is completed.

szetszwo (approved):
+1 the change looks good.

bharatviswa504:
Thank You @szetszwo for the review.

PR #1644, opened 2020-12-01
HDDS-4535. Use fixed thread pool for closed container replication

elek:

Number of threads for closed container replications can be adjusted by the settings  `hdds.datanode.replication.streams.limit`. But this number is ignored today due to the misuse of `ThreadPoolExecutor`:

```java
new ThreadPoolExecutor(
        0, poolSize, 60, TimeUnit.SECONDS,
        new LinkedBlockingQueue<>(),
        new ThreadFactoryBuilder().setDaemon(true)
            .setNameFormat("ContainerReplicationThread-%d")
            .build())
```

Here the minimal number of threads is 0 and the maximum number of the threads is the configured value.  Threads in the thread pool supposed to be scaled up, but it doesn't.

[From the JDK docs](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html#ThreadPoolExecutor(int,%20int,%20long,%20java.util.concurrent.TimeUnit,%20java.util.concurrent.BlockingQueue)):

> A ThreadPoolExecutor will automatically adjust the pool size (see getPoolSize()) according to the bounds set by corePoolSize (see getCorePoolSize()) and maximumPoolSize (see getMaximumPoolSize()). When a new task is submitted in method execute(java.lang.Runnable), [...] [AND]  If there are more than corePoolSize but less than maximumPoolSize threads running, a new thread will be created only if the queue is full.

So if queue is not full (and `LinkedBlockgingQueue` is unbounded by default) the threads will never be created.

For a quick fix we can switch to use static thread pool instead of dynamic and always keep the required number of threads.

https://issues.apache.org/jira/browse/HDDS-4535

## How was this patch tested?

A new utility is committed as a unit test (`ReplicationSupervisorScheduling`). It helps to test the scheduling (and thread numbers can be checked by jstack) but it's not executed as part of the unit test (because it's slow and un-predictable) as it doesn't start with `Test*`

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/ReplicationSupervisor.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/ReplicationSupervisor.java
@@ -25,11 +25,11 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.hadoop.ozone.container.common.impl.ContainerSet;
 import org.apache.hadoop.ozone.container.replication.ReplicationTask.Status;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


runzhiwang:
> So if queue is not full (and LinkedBlockgingQueue is unbounded by default) the threads will never be created.

@elek Please correct me if I am wrong. If the queue is not full, it means we do not have too many tasks, then we do not need to create new thread, so maybe dynamic thread pool is also appropriate ?

elek:
> @elek Please correct me if I am wrong. If the queue is not full, it means we do not have too many tasks, then we do not need to create new thread, so maybe dynamic thread pool is also appropriate ?

Let's say we have 5 replication tasks scheduled. I would prefer to have 5 threads to work on it. But the current queue is unbounded, it's never full:

image

With the current code, we will have on thread all the time as we have unlimited space in the queue: Java doesn't panic.

An alternative approach is to use a bounded queue (`new LinkedBlockingQueue(10)`) but in this case the scheduling can be blocked (if SCM sends requests to replicate 100 containers, the scheduling will be blocked after the first 10 is added to the queue) which cause other problems on the RPC side.

For me it was not intuitive (threads are not scaled up unless the queue is full), but this seems to be the normal behavior of `ThreadPool`

adoroszlai (approved):

adoroszlai:
Thanks @elek for the fix, @runzhiwang for the review.

PR #1645, opened 2020-12-01
HDDS-4463. Verify that no node can start when Software Layout Version < Metadata Layout Version.

errose28:

Add tests for OzoneManager and StorageContainerManager classes that will verify an Exception is thrown on startup when the metadata layout version on disk is larger than the software version of the component. A similar test will be added for the datanode after the DN version file logic is refactored in a later upgrade Jira.

HDDS-4463

## How was this patch tested?

Added tests pass.


hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
```diff
@@ -43,7 +43,7 @@
   public static final String STORAGE_ID = "storageID";
   public static final String DATANODE_UUID = "datanodeUuid";
   public static final String CLUSTER_ID = "clusterID";
-  public static final String LAYOUTVERSION = "layOutVersion";
+  public static final String LAYOUTVERSION = "layoutVersion";
```
avijayanhwx, line 46: Changing this may break backward compatibility since the value of this constant is written down to and read from the Version file. We can keep this as it is.

avijayanhwx (commented):
Thanks for working on this @errose28. Mostly looks good to me. One comment inline.

avijayanhwx (approved):

avijayanhwx:
@errose28 The CI failure looks related. Please take a look.

avijayanhwx:
Test failure seems unrelated.  I ran it multiple times on local setup with the patch, but could not reproduce it. Since this is a test only change to feature branch, overriding CI failure.

avijayanhwx:
Thanks for the contribution @errose28.

PR #1648, opened 2020-12-02
HDDS-4449. Rename config ozone.datanode.pipeline.limit

wycccccc:

Renamed `ozone.datanode.pipeline.limit` a suitable name.

https://issues.apache.org/jira/browse/HDDS-4449

## How was this patch tested?

no test.


ayushtkn:
Is directly changing the conf fine? Don't we need to deprecate the existing one? To maintain compatibility, or is that not the case here

wycccccc:
Thanks for reminding @ayushtkn ，you are right. I will add deprecation to avoid compatibility issues.

adoroszlai:
/pending I will add deprecation to avoid compatibility issues

jojochuang:
Not that familiar with Ozone code but can we try using this kind of Hadoop configuration APIs?

https://github.com/apache/hadoop/blob/b69ac575a1a02d39c64c6cf998ec2ef4eb5918cd/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/HdfsConfiguration.java#L70
```
    Configuration.addDeprecations(new DeprecationDelta[]{
    new DeprecationDelta("dfs.backup.address",
    DeprecatedKeys.DFS_NAMENODE_BACKUP_ADDRESS_KEY),
```

wycccccc:
Thanks for your suggestion @jojochuang , I found a similar method in ozone.

wycccccc:
Thanks for your help @adoroszlai. I have corrected this, I am not sure if it is appropriate to change it here, do you think it is feasible.

adoroszlai:
Thanks @wycccccc for adding the deprecation.  It reminded me to merge your other similar PR.  Now there is a conflict in this PR.  Sorry for the inconvenience.  Since `TestOzoneConfigurationFields` failure is caused by the changes here, a minor update is needed anyway: old config key name is leftover in the XML.

https://github.com/apache/ozone/blob/d02659c802a55366ed5d42176f83fd3917f3526c/hadoop-hdds/common/src/main/resources/ozone-default.xml#L779

wycccccc:
Thanks @adoroszlai review, I have removed old config key and pass`TestOzoneConfigurationFields`.

adoroszlai (approved):
Thanks @wycccccc for updating the patch.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/conf/OzoneConfiguration.java
```diff
@@ -293,8 +294,11 @@ public Properties getAllPropertiesByTag(String tag) {
     return configMap;
   }
 
+
```
ayushtkn, line 297: nit: Remove the new line change

ayushtkn (approved):
Thanx @wycccccc for the contribution.
Changes LGTM

ayushtkn:
Thanx @wycccccc , @adoroszlai and @jojochuang

PR #1649, opened 2020-12-02
HDDS-4538: Workaround on HDDS-2823, hard code scmUuid and clusterID

GlenGeng:

A typical SCM group consists of 3 SCMs. If each of the them has its own scmUuid and clusterID, it will cause chaos in Datanode side.
For now, we just hard code the clusterID and scmUuid on HDDS-2823, so that we can setup a SCM raft group.
We need figure out a decent solution when merging back HDDS-2823 back to master. 

HDDS-4538

## How was this patch tested?

CI


GlenGeng:
@ChenSammi @nandakumar131  Please take a look at this PR. Thanks!

nandakumar131 (approved):

PR #1651, opened 2020-12-02
HDDS-4178. SCM Finalize client command implementation

prashantpogde:

SCM Finalize client command implementation

https://issues.apache.org/jira/browse/HDDS-4178

## How was this patch tested?

UT. I will fix CI failures.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
@@ -223,10 +223,7 @@ public class BasicUpgradeFinalizer<T, V extends AbstractLayoutVersionManager>
   }
 
   protected void emitStartingMsg() {
-    String msg = "Finalization started.\n";
-    msg += "Existing pipelines and containers will be closed during Upgrade.\n";
-    msg += "New pipelines creation will remain frozen until Upgrade is " +
-        "finalized";
+    String msg = "Finalization started.";
     logAndEmit(msg);
   }
 
@@ -266,7 +263,7 @@ public class BasicUpgradeFinalizer<T, V extends AbstractLayoutVersionManager>
     logAndEmit(msg);
   }
 
-  private void logAndEmit(String msg) {
+  protected void logAndEmit(String msg) {
     LOG.info(msg);
     msgs.offer(msg);
   }
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
@@ -218,14 +218,15 @@ public class BasicUpgradeFinalizer<T, V extends AbstractLayoutVersionManager>
 
   protected void emitNOOPMsg(String feature) {
     String msg = "No finalization work defined for feature: " + feature + ".";
-    String msg2 = "Skipping.";
 
     logAndEmit(msg);
-    logAndEmit(msg2);
   }
 
   protected void emitStartingMsg() {
-    String msg = "Finalization started.";
+    String msg = "Finalization started.\n";
+    msg += "Existing pipelines and containers will be closed during Upgrade.\n";
+    msg += "New pipelines creation will remain frozen until Upgrade is " +
+        "finalized";
     logAndEmit(msg);
   }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
@@ -214,7 +214,7 @@ public final class StorageContainerManager extends ServiceRuntimeInfoImpl
   private PipelineChoosePolicy pipelineChoosePolicy;
 
   private HDDSLayoutVersionManager scmLayoutVersionManager;
-  private UpgradeFinalizer upgradeFinalizer;
+  private UpgradeFinalizer<StorageContainerManager> upgradeFinalizer;
 
   /**
    * Creates a new StorageContainerManager. Configuration will be
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
@@ -90,6 +90,7 @@ import org.apache.hadoop.hdds.scm.pipeline.PipelineReportHandler;
 import org.apache.hadoop.hdds.scm.pipeline.SCMPipelineManager;
 import org.apache.hadoop.hdds.scm.pipeline.choose.algorithms.PipelineChoosePolicyFactory;
 import org.apache.hadoop.hdds.scm.safemode.SCMSafeModeManager;
+import org.apache.hadoop.hdds.scm.server.upgrade.SCMUpgradeFinalizer;
 import org.apache.hadoop.hdds.security.exception.SCMSecurityException;
 import org.apache.hadoop.hdds.security.x509.SecurityConfig;
 import org.apache.hadoop.hdds.security.x509.certificate.authority.CertificateServer;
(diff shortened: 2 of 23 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -211,6 +214,7 @@
   private PipelineChoosePolicy pipelineChoosePolicy;
 
   private HDDSLayoutVersionManager scmLayoutVersionManager;
+  private UpgradeFinalizer upgradeFinalizer;
```
avijayanhwx, line 217: Nit. Can be UpgradeFinalizer\<StorageContainerManager\>.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/ScmSubcommand.java
```diff
@@ -17,21 +17,25 @@
...
 
 import java.io.IOException;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 
 /**
  * Base class for admin commands that connect via SCM client.
  */
-public abstract class ScmSubcommand implements Callable<Void> {
+public abstract class ScmSubcommand extends FinalizeUpgradeBaseCommand
```
avijayanhwx, line 31: Why do we need ScmSubCommand to extend FinalizeUpgradeBaseCommand? Shouldn't the inheritance be the other way?

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/upgrade/FinalizeScmUpgradeSubcommand.java
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Handler of Finalize SCM command.
+ */
+@CommandLine.Command(
+    name = "FinalizeScm",
+    description = "Finalize SCM Upgrade",
+    mixinStandardHelpOptions = true,
+    versionProvider = HddsVersionProvider.class)
+
+public class FinalizeScmUpgradeSubcommand extends ScmSubcommand {
```
avijayanhwx, line 48: With this patch I see the CLI different for OM and SCM.

For SCM, it is something like **_ozone admin upgrade FinalizeSCM_**
For OM, it is **_ozone admin om finalizeupgrade_**

Can we make sure it is common across OM and SCM? I believe the following are OK -> 
_ozone admin om/scm finalizeupgrade
ozone admin upgrade finalize-om-upgrade/finalize-scm-upgrade_

In Ozone, the convention is to have CLI operations should be lower case with "-" between words as needed.

avijayanhwx (commented):
Thanks for working on this @prashantpogde. A few comments inline.

I tried out the patch on a docker based cluster and I have attached the screenshot. A few observations there : 
- It is not clear what is meant by "Skipping" in the sys out. If it means there is no upgrade action for a layout feature, that is not needed since the previous line conveys that. 
- Also, there is no log line for pre-finalize actions like freezing the pipeline creation and closing the pipelines down. Can we add that?

Screen Shot 2020-12-04 at 1 22 03 PM

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
prashantpogde: yup. Made the changes as suggested.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/ScmSubcommand.java
prashantpogde: I changed the structure as discussed over call so this is not applicable anymore.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/upgrade/FinalizeScmUpgradeSubcommand.java
prashantpogde: yes! I changed  it to

ozone admin om/scm finalizeupgrade

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
```diff
@@ -218,14 +218,15 @@ private void persistStorage(Storage config) throws IOException {
...
-    String msg2 = "Skipping.";
 
     logAndEmit(msg);
-    logAndEmit(msg2);
   }
 
   protected void emitStartingMsg() {
-    String msg = "Finalization started.";
+    String msg = "Finalization started.\n";
+    msg += "Existing pipelines and containers will be closed during Upgrade.\n";
```
avijayanhwx, line 227: If we add here, then OM finalization will also print out the message. Can we print out the message after the actual work is done (freeze pipeline creation, pipelines are closed etc)?

avijayanhwx (commented):
Thanks for addressing the comments @prashantpogde. One minor comment inline.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
prashantpogde: yup, I thought I was making change in ScmFinalizer. Fixed now.

prashantpogde:
acceptance failure is unrelated.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/upgrade/FinalizeScmUpgradeSubcommand.java
```diff
@@ -0,0 +1,153 @@
...
+    Future<?> monitor =
+        exec.submit(new UpgradeMonitor(client, upgradeClientID, force));
+    try {
+      monitor.get();
+      emitFinishedMsg("Storage Container Manager");
+    } catch (CancellationException |InterruptedException e) {
+      emitCancellationMsg("Storage Container Manager");
+    } catch (ExecutionException e) {
+      emitGeneralErrorMsg();
+      throw e;
```
fapifta, line 104: I think, here we should just throw the cause of the ExecutionException wrapped into an IOException here.
I believe it is a better approach, as with that other subcommands are not forced to declare to throw two types of exceptions.

However as an alternative I am absolutely fine with declaring execute to throw an Exception, and do not mark specific Exceptions to be thrown.

fapifta (commented):
Hi @prashantpogde
great work, thank you for working on this one.
The changes looks good to me.

Please find one small thing inline.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/upgrade/FinalizeScmUpgradeSubcommand.java
prashantpogde: yup, I will make the change.

prashantpogde:
thanks @fapifta @avijayanhwx for the review.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/upgrade/FinalizeScmUpgradeSubcommand.java
```diff
@@ -0,0 +1,152 @@
...
+    Future<?> monitor =
+        exec.submit(new UpgradeMonitor(client, upgradeClientID, force));
+    try {
+      monitor.get();
+      emitFinishedMsg("Storage Container Manager");
+    } catch (CancellationException |InterruptedException e) {
+      emitCancellationMsg("Storage Container Manager");
+    } catch (ExecutionException e) {
+      emitGeneralErrorMsg();
+      throw new IOException(e.getMessage());
```
fapifta, line 103: Thank you for making this change, one more thing still regarding this part, I think it is more beneficial to do not swallow the stack trace of this exception and having it reported in the output, so can we provide e.getCause() instead of e.getMessage() here to the wrapping IOException?

prashantpogde: Done.

fapifta (approved):
+1 from my side if ?CI is ok, thank you for addressing the comments I made.

avijayanhwx (approved):
Thanks for addressing the changes @prashantpogde.

avijayanhwx:
Merging this since CI failure is unrelated.

PR #1652, opened 2020-12-03
HDDS-4542. Need throw exception to trigger FailoverProxyProvider of SCM client to work

GlenGeng:

ScmBlockLocationProtocolServerSideTranslatorPB and StorageContainerLocationProtocolServerSideTranslatorPB need throw exception to trigger FailoverProxyProvider of SCM client to work.

https://issues.apache.org/jira/browse/HDDS-4542

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
@@ -133,9 +133,7 @@ public final class StorageContainerLocationProtocolServerSideTranslatorPB
   public ScmContainerLocationResponse submitRequest(RpcController controller,
       ScmContainerLocationRequest request) throws ServiceException {
     if (!isLeader()) {
-      return ScmContainerLocationResponse.newBuilder()
-              .setCmdType(request.getCmdType()).setTraceID(request.getTraceID())
-              .setSuccess(false).setStatus(Status.SCM_NOT_LEADER).build();
+      throw new ServiceException(new IOException("SCM IS NOT LEADER"));
     }
     return dispatcher
         .processRequest(request, this::processRequest, request.getCmdType(),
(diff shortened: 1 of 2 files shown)


GlenGeng:
@ChenSammi @timmylicheng @nandakumar131  Please take a look at this PR. This is one of the hot fix found during the phase 1.5 test.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
```diff
@@ -133,9 +133,7 @@ private boolean isLeader() throws ServiceException {
   public ScmContainerLocationResponse submitRequest(RpcController controller,
       ScmContainerLocationRequest request) throws ServiceException {
     if (!isLeader()) {
-      return ScmContainerLocationResponse.newBuilder()
-              .setCmdType(request.getCmdType()).setTraceID(request.getTraceID())
-              .setSuccess(false).setStatus(Status.SCM_NOT_LEADER).build();
+      throw new ServiceException(new IOException("SCM IS NOT LEADER"));
```
timmylicheng, line 136: Shall we expose scm names or ipadress for the current SCM?

timmylicheng:
Thanks @GlenGeng 's contribution. Overall LGTM. Just left a comment inline.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
GlenGeng: @timmylicheng  Thanks for looking at this patch !

Good point !
As far as I know, the failover proxy provider will print out the IP address, along with the exception.

We need revisit this code in future, when the failover proxy provider should support leaderHint. I've file a jira to track this task, https://issues.apache.org/jira/browse/HDDS-4545

ChenSammi:
+1

PR #1653, opened 2020-12-03
HDDS-4543. Failed to list keys when there are two bucket names with s…

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-4543

This task fixes the bug that bucket listKeys operation returns empty list,  when there are two buckets with same prefix names, such as "hadoop" and "hadoop-ozone".


bharatviswa504 (commented):
+1 LGTM.
Good catch.
Thank You @ChenSammi for the contribution.

ChenSammi:
Thanks @bharatviswa504 for the code review.

PR #1656, opened 2020-12-04
HDDS-4514. AllocateBlock : lookup and update open file table for the given path

rakeshadr:

This task is to use open file table for the allocate block operations. This has been identified as part of Hive TPCDS benchmark test.

https://issues.apache.org/jira/browse/HDDS-4514

## How was this patch tested?
Added UT cases. Also, verified by running hive tpcds test

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -64,6 +64,7 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import org.apache.hadoop.test.LambdaTestUtils;
+import org.jetbrains.annotations.NotNull;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Rule;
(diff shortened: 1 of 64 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -822,26 +849,26 @@ protected void testRenameDirToFile() throws Exception {
...
     fs.mkdirs(abcRootPath);
-    try {
-      fs.rename(abcRootPath, file1Destin);
-      Assert.fail("key already exists /root_dir/file1");
-    } catch (FileAlreadyExistsException faee) {
-      // expected
-    }
+    Assert.assertFalse("key already exists /root_dir/file1",
+            fs.rename(abcRootPath, file1Destin));
   }
```
linyiqun, line 854: Above rename related change already contained in PR #1607.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/key/TestOMAllocateBlockResponseV1.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMResponse;
+import org.apache.hadoop.util.Time;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Tests OMAllocateBlockResponse layout version V1.
+ */
+public class TestOMAllocateBlockResponseV1
+        extends TestOMAllocateBlockResponse {
+
```
linyiqun, line 36: Do we need to override getOzoneConfiguration method in this class?

```java
  protected OzoneConfiguration getOzoneConfiguration() {
    OzoneConfiguration config = super.getOzoneConfiguration();
    config.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION, "V1");
    return config;
  }
```

linyiqun (commented):
The change looks good to me. Below some minor comments from me.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
rakeshadr: @linyiqun  yes, you are right. Since that is in progress, I just added to make clean build report. We can push #1607 push/upstream it first and then I will rebase this PR on top of that. Hope that would be fine.

bharatviswa504:
@rakeshadr can you rebase this PR, it has multiple changes looks like it. 
Can you post a PR related to this Jira, which will make it easier to review?

rakeshadr:
Current PR has multiple changes and looks like stale. I am closing this.
@linyiqun @bharatviswa504 , I've created new PR# https://github.com/apache/ozone/pull/1679 on top of the branch. Kindly review the same. Thanks!

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/key/TestOMAllocateBlockResponseV1.java
rakeshadr: Thanks @linyiqun , configuring V1 version will be required to pick new tables, otherwise all the db entries will go to old table with new format. 

In new PR https://github.com/apache/ozone/pull/1679, I've updated existing response test cases as well to follow same pattern. Please review it again. Thanks!

PR #1659, opened 2020-12-04
HDDS-4329. Expose Ratis retry config cache in OM.

bharatviswa504:

Expose Ratis retry cache. Followed new config approach and also used same config name as RatisServer config.
And also with this Jira if any config is prefixed with "ozone.om.ha" and after that users try to use same the RatisServer config it will be used by OM RaftServer. (This allows if any configs need to be tuned and in OM we don't have a config, this capability can help in such scenarios)

Used 5 minutes as default value. As with default leader election time out and with retry 15 count, this value should be more than enough to catch retry requests and return from the cache.

https://issues.apache.org/jira/browse/HDDS-4329

## How was this patch tested?

Added a test

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAMetadataOnly.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAMetadataOnly.java
@@ -422,6 +422,24 @@ public class TestOzoneManagerHAMetadataOnly extends TestOzoneManagerHA {
     Assert.assertFalse(logCapturer.getOutput().contains("created volume:"
         + volumeName));
 
+    //Sleep for little above seconds to get cache clear.
+    Thread.sleep(35000);
+
+    raftClientReply =
+        raftServer.submitClientRequest(new RaftClientRequest(clientId,
+            raftServer.getId(), ozoneManagerRatisServer.getRaftGroup()
+            .getGroupId(), callId, Message.valueOf(
+            OMRatisHelper.convertRequestToByteString(omRequest)),
+            RaftClientRequest.writeRequestType(), null));
+
+    Assert.assertTrue(raftClientReply.isSuccess());
+
+    // As second time with same client id and call id, this request should
+    // be executed ratis server as we are sending this request after cache
+    // expiry duration.
+    Assert.assertTrue(logCapturer.getOutput().contains(
+        "Volume creation failed"));
+
   }
 
   private void validateVolumesList(String userName,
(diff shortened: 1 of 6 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAMetadataOnly.java
```diff
@@ -422,6 +422,24 @@ public void testOMRetryCache() throws Exception {
     Assert.assertFalse(logCapturer.getOutput().contains("created volume:"
         + volumeName));
 
+    //Sleep for little above seconds to get cache clear.
+    Thread.sleep(65000);
```
lokeshj1703, line 426: Can we use a smaller cache duration?

lokeshj1703 (commented):
@bharatviswa504 Thanks for working on this! The changes look good to me. 
What is the total retry duration for a om client request? I mean what is the retry policy used right now. The cache duration should be >= time for which a request can be retried.

szetszwo (commented):
The change looks good other than @lokeshj1703 's comment.  Thanks.

bharatviswa504:
> @bharatviswa504 Thanks for working on this! The changes look good to me.
> What is the total retry duration for a om client request? I mean what is the retry policy used right now. The cache duration should be >= time for which a request can be retried.

We have 15 retries currently, and if same om is contacted again after failure it doubles the duration. 
And the double of the time is not for the first round of failovers until we find leader.

And for any other errors, we don't have sleep between retry.

So basically first 3 to find leader OM no time wait. And next, we have detected leader then if it fails over to a new OM, we don't wait. So at best case it would be less than a minute (as for any n/w error we immediately failover). And if we continuously get LeaderNotReady, we try to contact same OM after first round, then it will be (2+4+8+16+ .... +2^12) = 8190 duration, this is the worst time. If OM has taken so long to become Ready after failover and finally in last iteration it became ready.

So, in the worst case, we wait for 150 minutes. But setting at 150 mts, will be taking too much heap on wellbehaved OM. As in most cases, if OM's are up/down the total time if the same OM is not contacted will be less than a minute.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAMetadataOnly.java
bharatviswa504: Done.

lokeshj1703 (approved):
@bharatviswa504 Thanks for updating the PR! The changes look good to me. +1.
I think we should limit the total retry time for a client request. Can we do that as well as part of separate jira?

bharatviswa504:
Thank You @lokeshj1703 and @szetszwo  for the review

lokeshj1703:
Sorry! Missed the update. I think we are good with 132 seconds of total retry duration.

PR #1663, opened 2020-12-05
HDDS-4494. Provide Guice module to inject configuration to annotated fields

elek:

hadoop-hdds/config project provides a light-weight annotation based configuration interface. It supports to create an object with injection:

```
ReplicationConfig replicationConfig := ozoneConfig.getObject(replicationConfig); 
```

However, it seems to be hard to inject configuration to the service itself as usually we inject a lot of other dependencies to the constructor, not just the configuration.

One possible solution is using a Guice module. Guice is already used by recon, and this patch adds some optional modules to inject configuration variables to any service / instance if required.

https://issues.apache.org/jira/browse/HDDS-4494

## How was this patch tested?

New utility is not yet used by production code, but covered with a new unit test.

### DIFF
--- a/hadoop-hdds/config/pom.xml
+++ b/hadoop-hdds/config/pom.xml
@@ -38,11 +38,16 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <artifactId>slf4j-api</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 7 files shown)


elek:
Two months and 3 days without any answer.

It seems that nobody is interested to check or comment this PR.

I am closing it now.

PR #1664, opened 2020-12-06
HDDS-4540. Add a new OM admin operation to submit the OMPrepareRequest.

avijayanhwx:
- Introduce a new OM client operation to "prepare" the OM quorum.
- As a first pass, the client will just submit the request (HDDS-4480) and print out the response (Txn ID)
- In a follow up JIRA, the subsequent steps to probe every individual OM for preparation completeness will be added.

https://issues.apache.org/jira/browse/HDDS-4540

## How was this patch tested?
Manually tested.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
@@ -587,4 +587,14 @@ public interface OzoneManagerProtocol
     return false;
   }
 
+  /**
+   *
+   * @param flushWaitTimeout
+   * @param flushCheckInterval
+   * @return
+   */
+  default long prepareOzoneManager(long flushWaitTimeout,
+                                   long flushCheckInterval) throws IOException {
+    return -1;
+  }
 }
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
@@ -589,12 +589,17 @@ public interface OzoneManagerProtocol
 
   /**
    *
-   * @param flushWaitTimeout
-   * @param flushCheckInterval
+   * @param txnApplyWaitTimeoutSeconds Max time in SECONDS to wait for all
+   *                                   transactions before the prepare request
+   *                                   to be applied to the OM DB.
+   * @param txnApplyCheckIntervalSeconds Time in SECONDS to wait between
+   *                                     successive checks for all transactions
+   *                                     to be applied to the OM DB.
    * @return
    */
-  default long prepareOzoneManager(long flushWaitTimeout,
-                                   long flushCheckInterval) throws IOException {
+  default long prepareOzoneManager(
+      long txnApplyWaitTimeoutSeconds, long txnApplyCheckIntervalSeconds)
+      throws IOException {
     return -1;
   }
 }
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
@@ -1548,21 +1548,22 @@ public final class OzoneManagerProtocolClientSideTranslatorPB
   }
 
   @Override
-  public long prepareOzoneManager(long flushWaitTimeout,
-                                  long flushCheckInterval) throws IOException {
-    Preconditions.checkArgument(flushWaitTimeout > 0,
-        "flushWaitTimeout has to be > zero");
+  public long prepareOzoneManager(
+      long txnApplyWaitTimeoutSeconds, long txnApplyCheckIntervalSeconds)
+      throws IOException {
+    Preconditions.checkArgument(txnApplyWaitTimeoutSeconds > 0,
+        "txnApplyWaitTimeoutSeconds has to be > zero");
 
-    Preconditions.checkArgument(flushCheckInterval > 0 &&
-            flushCheckInterval < flushWaitTimeout / 2,
-        "flushCheckInterval has to be > zero and < half of " +
-            "flushWaitTimeout to make sense.");
+    Preconditions.checkArgument(txnApplyCheckIntervalSeconds > 0 &&
+            txnApplyCheckIntervalSeconds < txnApplyWaitTimeoutSeconds / 2,
+        "txnApplyCheckIntervalSeconds has to be > zero and < half "
+            + "of txnApplyWaitTimeoutSeconds to make sense.");
 
     PrepareRequest prepareRequest =
         PrepareRequest.newBuilder().setArgs(
             PrepareRequestArgs.newBuilder()
-                .setFlushWaitCheckInterval(flushCheckInterval)
-                .setFlushWaitTimeOut(flushWaitTimeout)
+                .setTxnApplyWaitTimeoutSeconds(txnApplyWaitTimeoutSeconds)
+                .setTxnApplyCheckIntervalSeconds(txnApplyCheckIntervalSeconds)
                 .build()).build();
 
     OMRequest omRequest = createOMRequest(Type.Prepare)
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
@@ -1547,6 +1547,32 @@ public final class OzoneManagerProtocolClientSideTranslatorPB
     return recoverResponse.getResponse();
   }
 
+  @Override
+  public long prepareOzoneManager(long flushWaitTimeout,
+                                  long flushCheckInterval) throws IOException {
+    Preconditions.checkArgument(flushWaitTimeout > 0,
+        "flushWaitTimeout has to be > zero");
+
+    Preconditions.checkArgument(flushCheckInterval > 0 &&
+            flushCheckInterval < flushWaitTimeout / 2,
+        "flushCheckInterval has to be > zero and < half of " +
+            "flushWaitTimeout to make sense.");
+
+    PrepareRequest prepareRequest =
+        PrepareRequest.newBuilder().setArgs(
+            PrepareRequestArgs.newBuilder()
+                .setFlushWaitCheckInterval(flushCheckInterval)
+                .setFlushWaitTimeOut(flushWaitTimeout)
+                .build()).build();
+
+    OMRequest omRequest = createOMRequest(Type.Prepare)
+        .setPrepareRequest(prepareRequest).build();
+
+    PrepareResponse prepareResponse =
+        handleError(submitRequest(omRequest)).getPrepareResponse();
+    return prepareResponse.getTxnID();
+  }
+
   @VisibleForTesting
   public OmTransport getTransport() {
     return transport;
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
@@ -24,7 +24,6 @@ import java.io.File;
 import java.nio.file.Paths;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Random;
 import java.util.Set;
 import java.util.UUID;
 
@@ -53,6 +52,8 @@ public class TestOzoneManagerPrepare extends TestOzoneManagerHA {
 
   private final String keyPrefix = "key";
   private final int timeoutMillis = 30000;
+  private final static Long PREPARE_FLUSH_WAIT_TIMEOUT_SECONDS = 300L;
+  private final static Long PREPARE_FLUSH_INTERVAL_SECONDS = 5L;
 
   /**
    * Calls prepare on all OMs when they have no transaction information.
@@ -64,7 +65,8 @@ public class TestOzoneManagerPrepare extends TestOzoneManagerHA {
     ClientProtocol ozClient = OzoneClientFactory.getRpcClient(getConf())
         .getObjectStore().getClientProxy();
     long prepareRequestLogIndex =
-        ozClient.getOzoneManagerClient().prepareOzoneManager(300, 5);
+        ozClient.getOzoneManagerClient().prepareOzoneManager(
+            PREPARE_FLUSH_WAIT_TIMEOUT_SECONDS, PREPARE_FLUSH_INTERVAL_SECONDS);
 
     // Prepare response processing is included in the snapshot,
     // giving index of 1.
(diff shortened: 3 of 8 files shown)


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/PrepareSubCommand.java
```diff
@@ -0,0 +1,62 @@
...
+    versionProvider = HddsVersionProvider.class
+)
+public class PrepareSubCommand implements Callable<Void> {
+
+  @CommandLine.ParentCommand
+  private OMAdmin parent;
+
+  @CommandLine.Option(
+      names = {"-id", "--service-id"},
+      description = "Ozone Manager Service ID"
```
linyiqun, line 43: Shall we need to add 'required = true' for this option?

linyiqun (commented):
Minor comments from me:

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
```diff
@@ -587,4 +587,14 @@ default boolean recoverTrash(String volumeName, String bucketName,
     return false;
   }
 
+  /**
+   *
+   * @param flushWaitTimeout
+   * @param flushCheckInterval
+   * @return
+   */
+  default long prepareOzoneManager(long flushWaitTimeout,
```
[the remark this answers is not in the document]
avijayanhwx: Thanks, done!

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/PrepareSubCommand.java
avijayanhwx: Yes, I have added it.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -152,11 +131,10 @@ public void testPrepareWithTransactions() throws Exception {
    * Checks that third OM received all transactions and is prepared.
    * @throws Exception
    */
-  // TODO: Fix this test so it passes.
-  //   @Test
+//  @Test
   public void testPrepareDownedOM() throws Exception {
     // Index of the OM that will be shut down during this test.
-    final int shutdownOMIndex = 2;
+    final int shutdownOMIndex = new Random().nextInt(3);
```
errose28, line 137: nit: Is there any benefit to using a random index over a fixed one? When observing the test run from logs, we now need to search the messages to determine which OM was taken down on this run (also distinguish the deliberate takedown from a crash or JVM pause induced leader change), instead of having that info beforehand.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -217,15 +188,11 @@ public void testPrepareDownedOM() throws Exception {
     // Since prepare was the last Ratis transaction, it should have all data
     // it missed once it receives the prepare transaction.
     cluster.restartOzoneManager(downedOM, true);
-    // Wait for other OMs to catch this one up on transactions.
-    LambdaTestUtils.await(timeoutMillis, 1000,
-        () -> downedOM.getRatisSnapshotIndex() == prepareIndex);
-    checkPrepared(downedOM, prepareIndex);
+    LambdaTestUtils.await(timeoutMillis, 2000,
+        () -> checkPrepared(downedOM, prepareIndex));
```
errose28, line 192: Can we replace this with a `waitAndCheckPrepared` call? Then we know that the logs have also been removed as well.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -266,61 +233,23 @@ private void writeTestData(ObjectStore store, String volumeName,
...
     // Log files are deleted after the snapshot is taken,
     // So once log files have been deleted, OM should be prepared.
     LambdaTestUtils.await(timeoutMillis, 1000,
         () -> !logFilesPresentInRatisPeer(om));
-    checkPrepared(om, prepareRequestLogIndex);
+    Assert.assertTrue(checkPrepared(om, prepareRequestLogIndex));
   }
 
-  private void checkPrepared(OzoneManager om, long prepareRequestLogIndex)
+  private boolean checkPrepared(OzoneManager om, long prepareRequestLogIndex)
```
errose28, line 245: Do we still need this method? Since we are no longer immediately checking the leader and waiting on all 3 OMs, can we just put these lines in `waitAndCheckPrepared`? This method on its own is kind of misleading, since it no longer checks that the logs have been removed, and therefore doesn't do a full check for preparedness.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -1547,6 +1547,32 @@ public boolean recoverTrash(String volumeName, String bucketName,
     return recoverResponse.getResponse();
   }
 
+  @Override
+  public long prepareOzoneManager(long flushWaitTimeout,
+                                  long flushCheckInterval) throws IOException {
+    Preconditions.checkArgument(flushWaitTimeout > 0,
+        "flushWaitTimeout has to be > zero");
```
errose28, line 1554: Should we add this check in `PrepareSubCommand` as well to give the user a specific error message if they pass a bad value? I'm not sure how the client handles these precondition generated exceptions and presents them to the user.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -1547,6 +1547,32 @@ public boolean recoverTrash(String volumeName, String bucketName,
...
+
+    Preconditions.checkArgument(flushCheckInterval > 0 &&
+            flushCheckInterval < flushWaitTimeout / 2,
+        "flushCheckInterval has to be > zero and < half of " +
+            "flushWaitTimeout to make sense.");
```
errose28, line 1559: If the user passes in a value for flush wait timeout that is less than the hardcoded flush check interval, they will see this message. This is confusing to them since it indicates the problem is flush check interval, which they have no knowledge of. They must also guess as to what the flush check interval value is to make sure their flush wait timeout is two times that when they try again.

Similar to above, a min value check on the user passed flush wait timeout to make sure it is large enough (or automatically set flush check interval based on the passed value) might be good to add in `PrepareSubCommand`.

errose28 (commented):
Thanks for working on this @avijayanhwx. This will be very useful in further testing of the prepare feature. I left some comments inline.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
avijayanhwx: Given that we have already made sure that there are no logs present in the functional OMs, I believe that it is sufficient to check just the prepare request apply marker in the downed OM.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
avijayanhwx: Due to my last comment, this still has one usage.

avijayanhwx (commented):
Thanks for the review @errose28. I have taken up most of your suggestions.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
avijayanhwx: Changed this to match the command's user parameter name.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
avijayanhwx: Reverted this change.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/PrepareSubCommand.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Handler of ozone admin om finalizeUpgrade command.
+ */
+@CommandLine.Command(
+    name = "prepare",
+    description = "Prepares Ozone Manager for upgrade/downgrade, by applying " +
+        "all pending transactions, taking a Ratis snapshot at the last txn " +
+        "and purging all logs on each OM instance. The returned txn id " +
+        "corresponds to the last txn in the quorum in which the snapshot is " +
+        "taken.",
```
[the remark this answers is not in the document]
avijayanhwx: I have used the full word in the description, and marked the parameter as "hidden".

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -162,9 +167,7 @@ private static void waitForLogIndex(long indexToWaitFor,
             && (ratisTxnIndex >= indexToWaitFor);
       }
 
-      if (!success) {
-        Thread.sleep(DOUBLE_BUFFER_FLUSH_CHECK_INTERVAL.toMillis());
-      }
+      Thread.sleep(flushCheckInterval.toMillis());
```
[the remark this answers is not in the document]
avijayanhwx: Good catch, this refactor was due to a merge mistake.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/PrepareSubCommand.java
```diff
@@ -0,0 +1,67 @@
...
+  @CommandLine.Option(
+      names = {"-id", "--service-id"},
+      description = "Ozone Manager Service ID",
+      required = true
+  )
+  private String omServiceId;
+
+  @CommandLine.Option(
+      names = {"-ft", "--flush-wait-timeout"},
+      description = "Max time to wait for OM Double Buffer flush in seconds.",
```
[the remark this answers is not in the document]
avijayanhwx: Refactored this to transaction-apply-wait-timeout, and marked this field as hidden. Since I am expecting a global "prepare-timeout" parameter for the client which includes the wait for each OM to be prepared, I have used a more applicable parameter name here. But, I agree this is an internal detail and hence I have moved it to a hidden field.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/PrepareSubCommand.java
```diff
@@ -0,0 +1,67 @@
...
+      defaultValue = "300"
+  )
+  private long flushWaitTime;
+
+  @Override
+  public Void call() throws Exception {
+    OzoneManagerProtocol client = parent.createOmClient(omServiceId);
+    long prepareTxnId = client.prepareOzoneManager(flushWaitTime, 5);
```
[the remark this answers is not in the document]
avijayanhwx: Added this as a hidden parameter as well.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -1547,6 +1547,32 @@ public boolean recoverTrash(String volumeName, String bucketName,
     return recoverResponse.getResponse();
   }
 
+  @Override
+  public long prepareOzoneManager(long flushWaitTimeout,
+                                  long flushCheckInterval) throws IOException {
```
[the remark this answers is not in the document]
avijayanhwx: Done.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -1079,7 +1079,12 @@ message FinalizeUpgradeProgressResponse {
 }
 
 message PrepareRequest {
+    required PrepareRequestArgs args = 1;
+}
 
+message PrepareRequestArgs {
+    optional uint64 flushWaitTimeOut = 1 [default = 300];
+    optional uint64 flushWaitCheckInterval = 2 [default = 5];
```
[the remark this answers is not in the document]
avijayanhwx: Done.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -112,26 +100,17 @@ public void testPrepareWithTransactions() throws Exception {
 
     // Make sure all OMs have logs from writing data, so we can check that
     // they are purged after prepare.
-    for (OzoneManager om: cluster.getOzoneManagersList()) {
-      LambdaTestUtils.await(timeoutMillis, 1000,
-          () -> logFilesPresentInRatisPeer(om));
-    }
+    cluster.getOzoneManagersList().forEach(om ->
+        Assert.assertTrue(logFilesPresentInRatisPeer(om)));
```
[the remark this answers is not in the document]
avijayanhwx: Fixed it.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -112,26 +100,17 @@ public void testPrepareWithTransactions() throws Exception {
...
 
-    OzoneManager leader = cluster.getOMLeader();
-    OMResponse omResponse = submitPrepareRequest(leader.getOmRatisServer());
-    // Get the log index of the prepare request.
+    OzoneManagerProtocol ozoneManagerClient =
+        ozClient.getObjectStore().getClientProxy().getOzoneManagerClient();
     long prepareRequestLogIndex =
-        omResponse.getPrepareResponse().getTxnID();
+        ozoneManagerClient.prepareOzoneManager(300, 5);
```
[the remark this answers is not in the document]
avijayanhwx: Done.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
errose28: Since the timeout command is now hidden this is probably okay.

errose28:
Thanks @avijayanhwx LGTM +1

swagle (approved):
+1 LGTM

avijayanhwx:
Thanks for the reviews @linyiqun, @errose28 & @swagle. I am merging this to unblock the efforts in HDDS-4569. If there are further review comments, they can be addressed in the following patches.

PR #1665, opened 2020-12-07
HDDS-4554. numKey metrics goes negative after intermediate directory deletion.

mukul1987:
During Ozone Key, File, and Directory creations, intermediate directories are not counted towards the number of keys in the system. When a directory is deleted, however, we traverse each of the keys and delete it, this results in underflow of the numKey metric.

https://issues.apache.org/jira/browse/HDDS-4554

Added a new unit test.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemMetrics.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemMetrics.java
@@ -30,7 +30,10 @@ import org.apache.hadoop.ozone.TestDataUtil;
 import org.apache.hadoop.ozone.client.OzoneBucket;
 import org.apache.hadoop.ozone.client.io.OzoneOutputStream;
 import org.apache.hadoop.ozone.om.OMConfigKeys;
-import org.junit.*;
+import org.junit.Rule;
+import org.junit.BeforeClass;
+import org.junit.AfterClass;
+import org.junit.Test;
 import org.junit.rules.Timeout;
 
 import java.io.IOException;
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemMetrics.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemMetrics.java
@@ -34,6 +34,7 @@ import org.junit.Rule;
 import org.junit.BeforeClass;
 import org.junit.AfterClass;
 import org.junit.Test;
+import org.junit.Assert;
 import org.junit.rules.Timeout;
 
 import java.io.IOException;
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemMetrics.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemMetrics.java
@@ -110,17 +110,20 @@ public class TestOzoneFileSystemMetrics {
   }
 
   private void testOzoneFileCommit(TestOps op) throws Exception {
-    long numKeysBeforeCreate = cluster.getOzoneManager().getMetrics().getNumKeys();
+    long numKeysBeforeCreate = cluster
+        .getOzoneManager().getMetrics().getNumKeys();
 
     int fileLen = 30 * 1024 * 1024;
     byte[] data = string2Bytes(RandomStringUtils.randomAlphanumeric(fileLen));
 
     Path parentDir = new Path("/" + RandomStringUtils.randomAlphanumeric(5));
-    Path filePath = new Path(parentDir, RandomStringUtils.randomAlphanumeric(5));
+    Path filePath = new Path(parentDir,
+        RandomStringUtils.randomAlphanumeric(5));
 
     switch (op) {
     case Key:
-      try (OzoneOutputStream stream = bucket.createKey(filePath.toString(), fileLen)) {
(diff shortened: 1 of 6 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -336,6 +338,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
     switch (result) {
     case SUCCESS:
+      omMetrics.incNumKeys(numMissingParents);
```
xiaoyuyao, line 341: We need to add incNumCreateFileFails for the failure case below?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemMetrics.java
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
+import org.apache.hadoop.ozone.MiniOzoneCluster;
+import org.apache.hadoop.ozone.OzoneConsts;
+import org.apache.hadoop.ozone.TestDataUtil;
+import org.apache.hadoop.ozone.client.OzoneBucket;
+import org.apache.hadoop.ozone.client.io.OzoneOutputStream;
+import org.apache.hadoop.ozone.om.OMConfigKeys;
+import org.junit.*;
```
xiaoyuyao, line 33: NIT: can we expand the wildcard import?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
xiaoyuyao, line 341: Should this be numMissingParents +1 ?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
```diff
@@ -345,6 +346,7 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
     switch (result) {
     case SUCCESS:
+      omMetrics.incNumKeys(numMissingParents);
```
xiaoyuyao, line 349: Same as above.

xiaoyuyao (commented):
LGTM, overall. A few minor questions inline...

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
bharatviswa504: This is already taken care 
```
    } catch (IOException ex) {
      result = Result.FAILURE;
      exception = ex;
      omMetrics.incNumCreateFileFails();
```

Is this what you are looking for?

bharatviswa504 (commented):
LGTM, once @xiaoyuyao comments are addressed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
mukul1987: The metric being incremented here is because of the missing directories. The key/file value is incremented because of metric increment in OmKeyCommitRequest

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
mukul1987: Same as the previous discussion

ayushtkn:
Thanx Mukul for the fix.
 `TestOMDirectoryCreateRequest.testCreateDirectoryOMMetric` looks related, I think you need to increment the value there in the test to include the missing directories seems related, Apart changes looks reasonable

>The key/file value is incremented because of metric increment in OmKeyCommitRequest

Thnax for this I too spent around an hour trying to figure out the logic behind this yesterday. :-)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
xiaoyuyao: yes. Thanks @bharatviswa504

xiaoyuyao: Thanks for the detail. Makes sense to me. Can we add some comments on this?

mukul1987:
@ayushtkn @xiaoyuyao @bharatviswa504 I have addressed the review comments for the patch. Please have a look at the patch.

ayushtkn (approved):

mukul1987:
Thanks for the reviews @ayushtkn @xiaoyuyao and @bharatviswa504. I have merged this change.

poorna243:
@mukul1987   In ozone recon version 1.2.0 . Recon dash board in overview shows key value but in Diskusage for the path which has keys always shows zero . Is it a bug fixed in later release or is there any configuration setting which we didn't add. Please let me know asap.

PR #1667, opened 2020-12-07
HDDS-4525. Replace Hadoop variables and functions in Ozone shell scripts with Ozone-specific ones

adoroszlai:

1. Replace shell functions and variables that were originally copied from Hadoop.
   * Deprecate `HADOOP_*` variables, but use their values, unless the corresponding `OZONE_` variable is also defined (which indicates "new code").
   * ~`HADOOP_CONF_DIR` is replaced by `OZONE_CONFIG_DIR` (instead of `OZONE_CONF_DIR`) to workaround a behavior `envtoconf` that cannot handle variables named `*_CONF_*`.~ (Root cause is fixed in HDDS-4601 instead.)
2. Drop unmaintained Windows scripts
3. Drop empty `mapreduce`, `yarn` etc. directories in final artifact, they are no longer necessary.
4. Fix wrong accumulation of return codes in `start-ozone.sh`: it was incremented for the first and third commands and unconditionally set for the second one, so the result of the first command was lost.  It should be set for the first one instead.

https://issues.apache.org/jira/browse/HDDS-4525

## How was this patch tested?

Added:
 * Bats test for the function that deprecates a variable
 * acceptance test for compatibility with `HDFS_OM_OPTS` etc. variables
 * acceptance tests for `ozone classpath` and `ozone envvars` commands

Changed MR acceptance tests to globally define both `HADOOP_CLASSPATH` and `OZONE_CLASSPATH`.  This confirms that shaded Ozone FS jar being present in `HADOOP_CLASSPATH` does not break Ozone startup.  Previously `HADOOP_CLASSPATH` had to be limited to the Hadoop containers.

Regular CI:
https://github.com/adoroszlai/hadoop-ozone/actions/runs/405676505

Manually tested `start-ozone.sh` and `stop-ozone.sh` in `ozonescripts` environment (see HDDS-4556 for automating the smoketest).


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/DBConfigFromFile.java
```diff
@@ -65,7 +65,7 @@ public static File getConfigLocation() throws IOException {
 
     if (StringUtil.isBlank(path)) {
       LOG.debug("Unable to find the configuration directory. "
-          + "Please make sure that HADOOP_CONF_DIR is setup correctly.");
+          + "Please make sure that " + CONFIG_DIR + " is setup correctly.");
```
smengcl, line 68: Should be `OZONE_CONFIG_DIR`?

fapifta:
Hi @adoroszlai, thank you for working on this change, please find a few comments from me inline.

I have a concern about OZONE_CONFIG_DIR, we brake a convention here, as not just Hadoop, but HBase, or Hive as well for example uses the COMPONENT_CONF_DIR to specify the config dir.
Is it possible to instead changing to config and breaking the convention fix the envtoconf behaviour, I see that OZONE_CONFIG_DIR is added to docker-config files and that's how envtoconf came into the picture, previously we did not needed HADOOP_CONF_DIR there, we do we need OZONE_CONFIG_DIR in the docker-config files now?

hadoop-hdds/common/src/main/conf/ozone-env.sh
```diff
@@ -0,0 +1,280 @@
...
+
+# If OZONE_USE_CLIENT_CLASSLOADER is set, OZONE_CLASSPATH and
+# OZONE_USER_CLASSPATH_FIRST are ignored.
+# export OZONE_USE_CLIENT_CLASSLOADER=true
+
+###
+# Options for remote shell connectivity
+###
+
+# There are some optional components of hadoop that allow for
```
fapifta, line 125: start-dfs.sh is mentioned two times here, can you please rephrase this comment, and the next which mentions it to point to start-ozone.sh, and to mention Ozone roles?

hadoop-hdds/common/src/main/conf/ozone-env.sh
```diff
@@ -0,0 +1,280 @@
...
+
+# Default log4j setting for daemons spawned explicitly by
+# --daemon option of hadoop, hdfs, mapred and yarn command.
+# Java property: hadoop.root.logger
+# export OZONE_DAEMON_ROOT_LOGGER=INFO,RFA
+
+# Default log level and output location for security-related messages.
+# You will almost certainly want to change this on a per-daemon basis via
+# the Java property (i.e., -Dhadoop.security.logger=foo). (Note that the
+# defaults for the NN and 2NN override this by default.)
```
fapifta, line 184: We should refer to Ozone roles here.

hadoop-hdds/common/src/main/conf/ozone-env.sh
```diff
@@ -0,0 +1,280 @@
...
+# .. and then use it as per the b option under the namenode.
+
+###
+# Secure/privileged execution
+###
+
+#
+# Out of the box, Ozone uses jsvc from Apache Commons to launch daemons
+# on privileged ports.  This functionality can be replaced by providing
+# custom functions.  See hadoop-functions.sh for more information.
```
fapifta, line 213: We should refer ozone-functions.sh here.

hadoop-hdds/common/src/main/conf/ozone-env.sh
```diff
@@ -0,0 +1,280 @@
...
+
+###
+# HDFS StorageContainerManager specific parameters
+###
+# Specify the JVM options to be used when starting the HDFS Storage Container Manager.
+# These options will be appended to the options specified as OZONE_OPTS
+# and therefore may override any similar flags set in OZONE_OPTS
+#
+# export OZONE_SCM_OPTS=""
+
```
fapifta, line 264: It is maybe out of scope for this PR, but shouldn't we have OZONE_RECON_OPTS, and OZONE_S3GW_OPTS similarly?

hadoop-hdds/common/src/main/conf/ozone-env.sh
```diff
@@ -0,0 +1,280 @@
...
+# via this special env var:
+# export OZONE_ENABLE_BUILD_PATHS="true"
+
+#
+# To prevent accidents, shell commands be (superficially) locked
+# to only allow certain users to execute certain subcommands.
+# It uses the format of (command)_(subcommand)_USER.
+#
+# For example, to limit who can execute the namenode command,
+# export HDFS_NAMENODE_USER=hdfs
```
fapifta, line 280: Does this remain here on purpose?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/DBConfigFromFile.java
```diff
@@ -65,7 +65,7 @@ public static File getConfigLocation() throws IOException {
 
     if (StringUtil.isBlank(path)) {
```
fapifta, line 66: We duplicate this if statement here, first to write a debug level log message second to return null, can we pull the two together, and also change the log message here, and state the reason why we fail by saying something like:
CONFIG_DIR + " variable is empty, please make sure it is setup correctly!"

hadoop-ozone/dist/src/main/compose/ozone-csi/docker-compose.yaml
```diff
@@ -24,7 +24,7 @@ services:
     env_file:
       - docker-config
     environment:
-      HADOOP_OPTS: ${HADOOP_OPTS}
+      OZONE_OPTS:
```
fapifta, line 27: For all these replacements, shouldn't we still provide ${OZONE_OPTS} to the environment as we did with HADOOP_OPTS before? We specify the coverage related options via this variable in test_all.sh for example.

hadoop-ozone/dist/src/main/smoketest/compatibility/om.robot
```diff
@@ -0,0 +1,27 @@
...
+
+*** Settings ***
+Documentation       Test om compatibility
+Library             BuiltIn
+Resource            ../lib/os.robot
+Test Timeout        5 minutes
+
+*** Test Cases ***
+Picks up command line options
+    Pass Execution If    '%{HDFS_OM_OPTS}' == ''    Command-line option required for process check
```
fapifta, line 24: Shouldn't we rename this envvar also to OZONE_OM_OPTS?

hadoop-ozone/dist/src/main/smoketest/compatibility/scm.robot
```diff
@@ -0,0 +1,27 @@
...
+
+*** Settings ***
+Documentation       Test scm compatibility
+Library             BuiltIn
+Resource            ../lib/os.robot
+Test Timeout        5 minutes
+
+*** Test Cases ***
+Picks up command line options
+    Pass Execution If    '%{HDFS_STORAGECONTAINERMANAGER_OPTS}' == ''    Command-line option required for process check
```
fapifta, line 24: Should we rename this envvar also to OZONE_SCM_OPTS or OZONE_STORAGECONTAINERMANAGER_OPTS?

hadoop-ozone/dist/src/shell/hdds/workers.sh
```diff
@@ -20,40 +20,43 @@
...
 fi
 
-HADOOP_LIBEXEC_DIR="${HADOOP_LIBEXEC_DIR:-$HADOOP_DEFAULT_LIBEXEC_DIR}"
-# shellcheck disable=SC2034
-HADOOP_NEW_CONFIG=true
-if [[ -f "${HADOOP_LIBEXEC_DIR}/hadoop-config.sh" ]]; then
-  . "${HADOOP_LIBEXEC_DIR}/hadoop-config.sh"
-else
-  echo "ERROR: Cannot execute ${HADOOP_LIBEXEC_DIR}/hadoop-config.sh." 2>&1
+if ! ozone_bootstrap; then
```
fapifta, line 50: This seems to be a strange one here...
So in line 45 if ozone_bootstrap is not declared, we error out because ozone-functions.sh could not be loaded.

As I understand here we check for the exit status of ozone_bootstrap function, and if it is false we exit because we can not find ozone-config.sh. Why we need this second check? As I see the ozone_bootstrap function is not doing anything that should fail, but maybe my eye slipped through something.

This same we do in hadoop-ozone/dist/src/shell/ozone/ozone, in hadoop-ozone/dist/src/shell/ozone/start-ozone.sh and in hadoop-ozone/dist/src/shell/ozone/stop-ozone.sh files as well

hadoop-ozone/dist/src/shell/ozone/start-ozone.sh
```diff
@@ -64,68 +63,53 @@ if [[ $# -ge 1 ]]; then
...
 echo "Starting datanodes"
-hadoop_uservar_su hdfs datanode "${HADOOP_HDFS_HOME}/bin/ozone" \
+ozone_uservar_su hdfs datanode "${OZONE_HOME}/bin/ozone" \
     --workers \
-    --config "${HADOOP_CONF_DIR}" \
+    --config "${OZONE_CONFIG_DIR}" \
     --daemon start \
     datanode ${dataStartOpt}
-(( HADOOP_JUMBO_RETCOUNTER=HADOOP_JUMBO_RETCOUNTER + $? ))
+OZONE_JUMBO_RETCOUNTER=$?
```
fapifta, line 85: Is this changed intentionally from accumulation to a set? Probably, but I wanted to be sure, and even so if this is the first place where we can safely set if this runs in just a pure context, this variable may have an initial value in some workflow which makes it worth to preserve the external value even at the first assignment?

adoroszlai: Yes, this is intentional, please see item 4 in PR description:

> it was incremented for the first and third commands and unconditionally set for the second one, so the result of the first command was lost.  It should be set for the first one instead.

hadoop-ozone/dist/src/main/smoketest/compatibility/om.robot
adoroszlai: No, these tests intentionaly use old names to verify compatibility with old scripts.  Some of the variables (eg. `HDFS_OM_OPTS` and `HDFS_STORAGECONTAINERMANAGER_OPTS`) were already deprecated before this change, but I changed the function that deprecates them.

https://github.com/apache/ozone/blob/dfd2aaf7fff4815e906e7201cee1739ad9776d97/hadoop-ozone/dist/src/shell/ozone/ozone#L160

hadoop-ozone/dist/src/main/smoketest/compatibility/scm.robot
adoroszlai: Same as [above](https://github.com/apache/ozone/pull/1667#discussion_r543078959).

hadoop-ozone/dist/src/main/compose/ozone-csi/docker-compose.yaml
adoroszlai: > Environment variables with only a key are resolved to their values on the machine Compose is running on

https://docs.docker.com/compose/compose-file/#environment

So repeating the variable name is unnecessary.

hadoop-hdds/common/src/main/conf/ozone-env.sh
adoroszlai: Thanks.

hadoop-hdds/common/src/main/conf/ozone-env.sh
adoroszlai: Thanks.

hadoop-hdds/common/src/main/conf/ozone-env.sh
adoroszlai: Thanks.

hadoop-hdds/common/src/main/conf/ozone-env.sh
adoroszlai: Agree, out of scope. ;)

hadoop-hdds/common/src/main/conf/ozone-env.sh
adoroszlai: No, thanks.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/DBConfigFromFile.java
adoroszlai: Not introduced by this change, but thanks for pointing out, let's fix it anyway.

adoroszlai:
> Is it possible to instead changing to config and breaking the convention fix the envtoconf behaviour, I see that OZONE_CONFIG_DIR is added to docker-config files and that's how envtoconf came into the picture, previously we did not needed HADOOP_CONF_DIR there, we do we need OZONE_CONFIG_DIR in the docker-config files now?

`HADOOP_CONF_DIR` is already defined in the docker images we use, hence it's not needed in `docker-config`, eg.:

```
$ docker run -it --rm apache/ozone-runner:20200625-1 env | grep HADOOP
HADOOP_LOG_DIR=/var/log/hadoop
HADOOP_CONF_DIR=/etc/hadoop
```

Also, `HADOOP_CONF_DIR` is an explicitly defined exception in `envtoconf`:

https://github.com/apache/ozone/blob/09579756b0756fcaf1b22b860c1ee1eb927e82c2/hadoop-ozone/dist/src/main/dockerlibexec/envtoconf.py#L39

We could fix this script, but it would only apply to `ozone-runner` containers, since it uses mounted Ozone binaries.  Both `ozone` and `hadoop` containers (used in `upgrade`, `ozone-mr`, etc. tests) come with `envtoconf.py` baked in.

```
$ docker run -it --rm apache/hadoop:3 grep CONF_DIR /opt/envtoconf.py
    self.excluded_envs = ['HADOOP_CONF_DIR']
$ docker run -it --rm apache/ozone:1.0.0 grep CONF_DIR /opt/hadoop/libexec/envtoconf.py
    self.excluded_envs = ['HADOOP_CONF_DIR']
```

So we have 3 options:

1. proper fix: update `envtoconf` in both Ozone and Hadoop and build new Docker images (`apache/ozone-runner`, `apache/hadoop`, `apache/ozone`, the latter two for multiple versions)
2. hack: modify all `docker-compose.yaml` files to mount the fixed version of `envtoconf`
3. workaround: use different variable name

fapifta:
Hi Attila, thank you for addressing review comments of mine, and going further and find way more cases than I did!

Regarding the CONF_DIR environment variables, I would say we should fix it properly, and I understand that updating the docker images would be a pain, and a bit of a tedious project, and this task is already tedious.

On the other hand, we learned that the envtoconf.py may change over time, as requirements are changing, I don't think that option two is hacky. It is hacky in a way that we overwrite a file backed into the docker images, but on the other hand, I would argue that the file should not be in the docker image itself, as even if this does not happen often, the file changes from time to time, so we should preserve the possibility to change it easily.
What do you think?

Ofc, a properly done option two would require to modify all the images to leave out envtoconf.py, and to modify all the docker-compose file to mount the file into the image, which is again even more tedious.
What if as part of this JIRA you add the mount to the docker compose files, and we create a follow up jira to remove the file from containers?
In return I can offer to volunteer for doing the docker image update, though I might need some help with it, but I really don't want to break this convention ;)

fapifta:
Hi @adoroszlai, thank you for addressing my concern and revert back to OZONE_CONF_DIR.

Let me ask as I did not have the time to go through and understand fully the transformation.py usage and how that effects the environment processing that is problematic with env vars with names like *CONF*, and to understand what is happening exactly as the consequence of your change?
Now I am guessing, but based on your change envtoconf.py is using transformation.py and with adding it to the docker-compose.yaml file, we effectively overriding the original transformation.py that is backed into the image(?).
The change itself in transformation.py is to go through the items instead of the collection means that we are transforming something different in those cases and we do it similarly as the one case you mentioned working well before this change also, I am not sure though, how this solves the problem.
Can you please give me some pointers where should I see what makes the difference or can you please point out the difference for me to understand a bit more easily what is happening?

adoroszlai:
Thanks @fapifta for taking another look, and also your initial review and nudging me to fix it properly.

> Now I am guessing, but based on your change envtoconf.py is using transformation.py and with adding it to the docker-compose.yaml file, we effectively overriding the original transformation.py that is backed into the image(?).

Exactly.  This change is necessary only for `apache/ozone` and `apache/hadoop` images, since `apache/ozone-runner` requires Ozone binaries to be mounted anyway.  We can remove these mounts if/when the images are rebuilt with fixed `envtoconf`.

Answered the other part of your question about `transformation.py` in the separate PR where this is being fixed (should be merged before this one).

fapifta (approved):
After discussing what will happen during env to conf conversion with the *CONF* environment variables, now it is clear how the changes solve that problem. (Files might get generated, from the environment variables, but those files are not used by anything, so it is fine to left those there as garbage.

I went through the changes again, I haven't spot anything problematic, I am +1 to commit it.

smengcl (commented):
Thanks @adoroszlai for working on this. Overall looks pretty good to me.

Some of my inline comment might be outdated (posted a few days ago before Pifta's review and your new. changes) -- will update those soon.

Do we have a `HADOOP_ROOT_LOGGER` alternative here?

Also do we take in `HADOOP_OPTS` when `OZONE_OPTS` is not set? IIRC some CM configs may still append to `HADOOP_OPTS` at the moment.

Also I think it would be a good idea to put a small table in wiki or somewhere to show the Ozone equivalents of Hadoop environment variables. Should be very helpful for admins adapting Ozone from HDFS. To start:

| Hadoop / HDFS   | Ozone          | Note                                                         |
| --------------- | -------------- | ------------------------------------------------------------ |
| HADOOP_CONF_DIR | OZONE_CONF_DIR | When OZONE_CONF_DIR is not set, HADOOP_CONF_DIR will be used; if set, HADOOP_CONF_DIR is ignored. |
| HADOOP_OPTS     | OZONE_OPTS     |                                                              |

hadoop-ozone/dist/dev-support/bin/dist-layout-stitching
```diff
@@ -103,10 +93,8 @@ run cp -r "${ROOT}/hadoop-ozone/dist/src/main/dockerlibexec/." "libexec/"
 run cp "${ROOT}/hadoop-ozone/dist/src/shell/ozone/ozone" "bin/"
 
 
-run cp "${ROOT}/hadoop-ozone/dist/src/shell/hdds/hadoop-config.sh" "libexec/"
-run cp "${ROOT}/hadoop-ozone/dist/src/shell/hdds/hadoop-config.cmd" "libexec/"
-run cp "${ROOT}/hadoop-ozone/dist/src/shell/hdds/hadoop-functions.sh" "libexec/"
```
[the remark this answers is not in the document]
smengcl: Yes we have `OZONE_OPTS`, but not sure if `HADOOP_OPTS` will be picked up when `OZONE_OPTS` right now. Probably not?

hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop27/docker-config
```diff
@@ -18,4 +18,7 @@ CORE-SITE.xml_fs.AbstractFileSystem.o3fs.impl=org.apache.hadoop.fs.ozone.OzFs
 CORE-SITE.xml_fs.AbstractFileSystem.ofs.impl=org.apache.hadoop.fs.ozone.RootedOzFs
 MAPRED-SITE.XML_mapreduce.application.classpath=/opt/hadoop/share/hadoop/mapreduce/*:/opt/hadoop/share/hadoop/mapreduce/lib/*:/opt/ozone/share/ozone/lib/hadoop-ozone-filesystem-hadoop2-@project.version@.jar
 
+HADOOP_CLASSPATH=/opt/ozone/share/ozone/lib/hadoop-ozone-filesystem-hadoop2-@project.version@.jar
+OZONE_CLASSPATH=
```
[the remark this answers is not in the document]
adoroszlai: > setting it to empty on purpose

Exactly: without it `HADOOP_CLASSPATH` containing OzoneFS jar would be picked up, which is bad for Ozone's health.  Actually, this is the primary motivation for this entire change.  From description of HDDS-4525:

> severe problem happens if we would like to access Ozone filesystem both via `ozone` and `hadoop` commands.  The latter needs shaded Ozone FS JAR in `HADOOP_CLASSPATH`.  The same `HADOOP_CLASSPATH` results in `ClassNotFound` for `ozone`.

hadoop-ozone/dist/src/main/smoketest/cli/classpath.robot
```diff
@@ -0,0 +1,46 @@
...
+*** Settings ***
+Documentation       Test ozone classpath command
+Library             BuiltIn
+Resource            ../lib/os.robot
+Resource            ../ozone-lib/shell.robot
+Test Timeout        5 minutes
+Suite Setup         Find Jars Dir
+
+*** Test Cases ***
+Ignores HADOOP_CLASSPATH if OZONE_CLASSPATH is set
```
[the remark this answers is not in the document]
adoroszlai: Yes.  There is a compatibility test for this.  We define some old `OPTS`:

https://github.com/apache/ozone/blob/a70def928eca460dcf5abb3b0dfe87f01cec29e5/hadoop-ozone/dist/src/main/compose/compatibility/docker-config#L31-L34

and then check if these are passed to the java process that `ozone om` command starts:

https://github.com/apache/ozone/blob/e54c4390bcc8374970baa120b6c6d9b5a375ff1d/hadoop-ozone/dist/src/main/smoketest/compatibility/om.robot#L23-L27

(Similar test exists for other components.)

adoroszlai:
Thanks @smengcl for the review.

> Do we have a `HADOOP_ROOT_LOGGER` alternative here?

Yes, `OZONE_ROOT_LOGGER`.  Note that the java property is still `hadoop.root.logger` (since I didn't find a way to deprecate these).

hadoop-ozone/dist/src/main/smoketest/cli/classpath.robot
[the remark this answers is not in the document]
smengcl: awesome. thanks!

smengcl (approved):
+1. Pending CI

smengcl:
Thanks @adoroszlai for rechecking. Looks like we got a good run. Will merge this shortly.

adoroszlai:
Thanks @fapifta and @smengcl for the reviews, and @smengcl for merging it.

PR #1669, opened 2020-12-07
HDDS-4460. Add metrics for closed container replication

elek:

Today it's hard to understand what's going on and what is the performance of the replication.

https://issues.apache.org/jira/browse/HDDS-4460

## How was this patch tested?

 1. Started docker compose cluster
 2. Generated keys
 3. Removed one datanode
 4. Restored one datanode
 5. Waited until data is replicated
 6. Checked the metrics


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/MeasuredReplicator.java
```diff
@@ -0,0 +1,109 @@
...
+    DefaultMetricsSystem.instance()
+        .register(NAME, "Closed container replication", this);
+  }
+
+  @Override
+  public void replicate(ReplicationTask task) {
+    long start = System.currentTimeMillis();
+
+    long msInQueue =
+        Instant.now().getNano() - task.getQueued().getNano() / 1_000_000;
```
jojochuang, line 67: did you forget a pair of brackets here? Like 
```
(Instant.now().getNano() - task.getQueued().getNano()) / 1_000_000;
```

jojochuang (dismissed):
the rest looks good

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/MeasuredReplicator.java
elek: OMG.

Fixed, thanks to save me ;-)

adoroszlai (approved):
Some ideas for future improvement:

* bytes transferred for failed tasks
* time spent on failed attempts
* failure reason (download or import - unless these steps are merged)

adoroszlai:
Thanks @elek for the patch and @jojochuang for the review.

PR #1670, opened 2020-12-07
HDDS-4403. Update the container replica history to the Recon DB lazily instead of for every report

smengcl:

1. Container replica history DB should be updated less frequently. -- Pre-patch it is updating the SQL DB for every new report, which becomes a bottleneck when scaling with # of containers.
2. Switch the underlying DB for container replica history report from SQL DB to RocksDB.
3. DB (container replica history table) now internally stores Datanode UUID instead of hostname. (The Web UI will still display hostnames.)

https://issues.apache.org/jira/browse/HDDS-4403

## How was this patch tested?

1. Existing test `TestContainerEndpoint#testGetReplicaHistoryForContainer` modded and passed.
2. Adding new UTs and fixing existing ones.


hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/codec/ContainerReplicaHistoryListCodec.java
```diff
@@ -0,0 +1,86 @@
...
+
+  // UUID takes 2 long to store. Each timestamp takes 1 long to store.
+  static final int SIZE_PER_ENTRY = 4 * Long.BYTES;
+  private final Codec<Long> lc = new LongCodec();
+
+  @Override
+  public byte[] toPersistedFormat(ContainerReplicaHistoryList obj)
+      throws IOException {
+
+    List<ContainerReplicaHistory> lst = obj.getList();
```
avijayanhwx, line 49: Based on 16 bytes per container-DN combination, it comes to around 16KB for a degenerate case where a container has been in 1000 Datanodes through its lifecycle. Hence, I think we should be OK on the replica history growth over time.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ContainerReplicaHistory.java
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.ozone.recon.scm;
+
+import java.util.UUID;
+
+/**
+ * A ContainerReplica timestamp class that tracks first and last seen time.
```
avijayanhwx, line 24: Can we add some documentation here on on the ContainerEndpoint API class that Recon can only track first and last seen time of the container replica on a DN. It cannot track intermediate states where a replica can move out a DN and then come back in the future again to the same DN. All Recon guarantees is first and last seen time, and not the whole interval in between the 2 times.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
```diff
@@ -171,23 +182,93 @@ public void addNewContainer(long containerId,
...
+    }
+  }
+
+  @VisibleForTesting
   public ContainerSchemaManager getContainerSchemaManager() {
     return containerSchemaManager;
   }
+
+  @VisibleForTesting
+  public Map<Long, Map<UUID, ContainerReplicaHistory>> getReplicaHistoryMap() {
```
avijayanhwx, line 262: I don't see any usages for this method.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
```diff
@@ -65,13 +73,16 @@
...
-      BatchOperationHandler batchHandler,
+      DBStore batchHandler,
       PipelineManager pipelineManager,
       StorageContainerServiceProvider scm,
       ContainerSchemaManager containerSchemaManager) throws IOException {
     super(conf, containerStore, batchHandler, pipelineManager);
     this.scmClient = scm;
     this.containerSchemaManager = containerSchemaManager;
+    this.replicaHistoryMap = new ConcurrentHashMap<>();
+    containerSchemaManager.setReplicaHistoryMap(replicaHistoryMap);
```
avijayanhwx, line 84: Is there any reason why we cannot maintain the ScmDB and replica history map only in this class and move the related methods (getContainerHistory, upsertContainerHistory, flushDb) from the ContainerSchemaManager to this class? The ContainerEndpoint class already has an instance of the Recon container manager. After this change, maybe we can also rename ContainerSchemaManager to ContainerHealthSchemaManager?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/spi/impl/ContainerDBServiceProviderImpl.java
```diff
@@ -184,6 +214,34 @@ public long getKeyCountForContainer(Long containerID) throws IOException {
...
 
+  /**
+   * Get the container replica history of the given containerID.
+   *
+   * @param containerID the given containerId.
+   * @return A map of ContainerReplicaWithTimestamp of the given containerID.
+   * @throws IOException
+   */
+  @Override
+  public Map<UUID, ContainerReplicaHistory> getContainerReplicaHistoryMap(
```
avijayanhwx, line 225: Nit. Method name can just be _getContainerReplicaHistory_.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/persistence/ContainerSchemaManager.java
```diff
@@ -113,40 +142,121 @@ public void insertUnhealthyContainerRecords(List<UnhealthyContainers> recs) {
...
+    if (replicaHistoryMap == null) {
+      return;
+    }
+    synchronized (replicaHistoryMap) {
+      try {
+        for (Map.Entry<Long, Map<UUID, ContainerReplicaHistory>> entry :
+            replicaHistoryMap.entrySet()) {
+          final long containerId = entry.getKey();
+          final Map<UUID, ContainerReplicaHistory> map = entry.getValue();
+          dbServiceProvider.storeContainerReplicaHistoryMap(containerId, map);
```
avijayanhwx, line 251: Can we support a batch interface that uses the RocksDB BatchOperation? That will help with faster flushing.

avijayanhwx (commented):
Thanks for working on this @smengcl. This is an involved patch with a lot of changes. A few comments inline.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
smengcl: It is referenced only in `TestReconContainerManager#testUpdateContainerReplica`

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/codec/ContainerReplicaHistoryListCodec.java
smengcl: Each entry costs 32 bytes (4*8). Should be fine anyway.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/persistence/ContainerSchemaManager.java
smengcl: Good idea. I'll try

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
smengcl: done!

avijayanhwx (approved):

smengcl:
Thanks @avijayanhwx for reviewing this. Will merge shortly.

PR #1672, opened 2020-12-07
HDDS-4558. Support Ozone block token with access mode check.

xiaoyuyao:

Adding read/write access mode check for ozone block token. 

https://issues.apache.org/jira/browse/HDDS-4558

## How was this patch tested?

adding unit tests and existing acceptance test.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/token/BlockTokenVerifier.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/token/BlockTokenVerifier.java
@@ -21,6 +21,7 @@ package org.apache.hadoop.hdds.security.token;
 import com.google.common.base.Strings;
 import org.apache.hadoop.hdds.HddsUtils;
 import org.apache.hadoop.hdds.protocol.datanode.proto.ContainerProtos;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.security.exception.SCMSecurityException;
 import org.apache.hadoop.hdds.security.x509.SecurityConfig;
 import org.apache.hadoop.hdds.security.x509.certificate.client.CertificateClient;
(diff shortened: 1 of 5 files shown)


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/token/BlockTokenVerifier.java
```diff
@@ -118,7 +125,19 @@ public void verify(String user, String tokenStr,
...
+        throw new BlockTokenException("Block token with " + id
+            + " doesn't have READ permission");
+      }
+    } else if (cmd == WriteChunk || cmd == PutBlock || cmd == PutSmallFile) {
+      if (!tokenId.getAccessModes().contains(
+          HddsProtos.BlockTokenSecretProto.AccessModeProto.WRITE)) {
+        throw new BlockTokenException("Block token with " + id
+            + " doesn't have WRITE permission");
+      }
+    }
```
elek, line 140: NIT: may be safer to throw an exception from a third, generic  `}else {`. Today there is no other option but if somebody implements a new message type, it will be forced to add condition for the new message.

elek (approved):
+1 Thanks the very quick fix @xiaoyuyao 

Didn't test it with real cluster (seems to be complicated as I need to extract the block token somehow), but based on the code it looks to me.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/token/BlockTokenVerifier.java
xiaoyuyao: good point. Even though the check block token was not currently done for other cmd, the third else will ensure proper handling is added if this is changed in future.

xiaoyuyao:
The acceptance test failure is caused by HDDS-4580. With HDDS-4580 committed, re-trigger CI.

xiaoyuyao:
The acceptance test passed LOCAL successfully after HDDS-4580. Not sure if the GITHUB action has timed out after 91 minutes due to slow VM instance. 

```

Status: | All tests passed
-- | --
00:46:56.998
log.html


```

xiaoyuyao:
Thanks @elek for reporting the issue and the review. The PR has been merged.

PR #1673, opened 2020-12-08
HDDS-4559. Avoid using hard coding uft-8 charset

lamber-ken:

From jdk1.7，jdk support `StandardCharsets.UTF_8`, so avoid using hard coding uft-8 charset

https://issues.apache.org/jira/browse/HDDS-4559

## How was this patch tested?

`mvn checkstyle:checkstyle`

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
@@ -23,6 +23,7 @@ import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.ratis.thirdparty.io.grpc.Context;
 import org.apache.ratis.thirdparty.io.grpc.Metadata;
 
+import java.nio.charset.StandardCharsets;
 import java.util.regex.Pattern;
 
 import static org.apache.ratis.thirdparty.io.grpc.Metadata.ASCII_STRING_MARSHALLER;
@@ -351,7 +352,7 @@ public final class OzoneConsts {
   public static final String GDPR_FLAG = "gdprEnabled";
   public static final String GDPR_ALGORITHM_NAME = "AES";
   public static final int GDPR_DEFAULT_RANDOM_SECRET_LENGTH = 16;
-  public static final String GDPR_CHARSET = "UTF-8";
+  public static final String GDPR_CHARSET = StandardCharsets.UTF_8.name();
   public static final String GDPR_LENGTH = "length";
   public static final String GDPR_SECRET = "secret";
   public static final String GDPR_ALGORITHM = "algorithm";
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
@@ -23,6 +23,7 @@ import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.ratis.thirdparty.io.grpc.Context;
 import org.apache.ratis.thirdparty.io.grpc.Metadata;
 
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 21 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
```diff
@@ -351,7 +352,7 @@ private OzoneConsts() {
   public static final String GDPR_FLAG = "gdprEnabled";
   public static final String GDPR_ALGORITHM_NAME = "AES";
   public static final int GDPR_DEFAULT_RANDOM_SECRET_LENGTH = 16;
-  public static final String GDPR_CHARSET = "UTF-8";
+  public static final String GDPR_CHARSET = StandardCharsets.UTF_8.name();
```
adoroszlai, line 355: Both usages of `GDPR_CHARSET` are in calls to methods where both `String` and `Charset` versions exist, so we can make this change:

```suggestion
  public static final Charset GDPR_CHARSET = StandardCharsets.UTF_8;
```

Since it is only used in `GDPRSymmetricKey`, I would also consider inlining this constant.  I don't think we need to define it globally.

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/client/TestDefaultCertificateClient.java
```diff
@@ -75,7 +76,7 @@
   private Path dnMetaDirPath;
   private SecurityConfig omSecurityConfig;
   private SecurityConfig dnSecurityConfig;
-  private final static String UTF = "UTF-8";
+  private final static String UTF = StandardCharsets.UTF_8.name();
```
adoroszlai, line 79: This is used in two kinds of calls:

1. `RandomStringUtils.random`: its `chars` parameter is not intended for charsets, rather the set of characters to use.  Eg. for `"UTF-8"` it will generate strings like `"--8F8T8U8T"`.  Since we don't need this special kind of random string, this usage can be simply removed.
2. `IOUtils.toInputStream`: can accept `Charset`, too.

After removing the first kind of usage, we can convert this constant to `Charset`.  (I would also inline it, ie. simply use `UTF_8`.)

ayushtkn (approved):

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/security/AWSV4AuthValidator.java
```diff
@@ -42,7 +42,7 @@
   private final static Logger LOG =
       LoggerFactory.getLogger(AWSV4AuthValidator.class);
   private static final String HMAC_SHA256_ALGORITHM = "HmacSHA256";
-  private static final Charset UTF_8 = Charset.forName("utf-8");
+  private static final Charset UTF_8 = StandardCharsets.UTF_8;
```
adoroszlai, line 45: Nit: this constant can be replaced with `import static java.nio.charset.StandardCharsets.UTF_8;`.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/SignatureProcessor.java
```diff
@@ -32,7 +33,7 @@
   String X_AMAZ_DATE = "X-Amz-Date";
   String CONTENT_MD5 = "content-md5";
   String AUTHORIZATION_HEADER = "Authorization";
-  Charset UTF_8 = Charset.forName("utf-8");
+  Charset UTF_8 = StandardCharsets.UTF_8;
```
adoroszlai, line 36: Nit: usages of this constant can be replaced with the one from `StandardCharsets`.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/scm/cli/SQLCLI.java
```diff
@@ -72,7 +73,7 @@
 
   private Options options;
   private BasicParser parser;
-  private final Charset encoding = Charset.forName("UTF-8");
+  private final Charset encoding = StandardCharsets.UTF_8;
```
adoroszlai, line 76: `encoding` is unused, can be removed.

adoroszlai (commented):
Thanks @lamber-ken for this cleanup.  I think there are a couple of places where it can be further improved a bit.

lamber-ken:
@adoroszlai Thanks for your suggestion!

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/client/TestDefaultCertificateClient.java
lamber-ken: done

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
lamber-ken: I think keep `GDPR_CHARSET` here is better, because define constants about GDPR here.

adoroszlai (approved):
Thanks @lamber-ken for updating the patch.

PR #1674, opened 2020-12-08
HDDS-4561: We should support suggested Ratis Server Leader in OM HA.

qizhu-lucas:

Now the OM HA not support suggested Leader, we should support it.

https://issues.apache.org/jira/browse/HDDS-4561

## How was this patch tested?

Later will add unit test.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -98,6 +98,10 @@ public final class OMConfigKeys {
       = "ozone.om.ratis.enable";
   public static final boolean OZONE_OM_RATIS_ENABLE_DEFAULT
       = false;
+  public static final String OZONE_OM_RATIS_SUGGESTED_LEADER_ENABLE_KEY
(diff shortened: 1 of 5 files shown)


ChenSammi:
@qizhu-lucas , thanks for this improvement.  I have following general questions,
1. Is Ratis Leader hint supported in ratis now?
2. If Ratis Leader hint is supported, what's the use case that "ozone.om.ratis.suggested.leader.enable" should be set as false?

qizhu-lucas:
@ChenSammi  
1. Ratis Leader suggested,  supported in ratis .
2. I just added false option for test, if we need roll back to original retry OMs based.

hanishakoneru:
We had previously removed failover to suggested leader because the nodeId is guaranteed consistency across servers and clients. For example, in server `nodeId1` could point to `OM1` and in client `nodeId1` could be pointing to `OM2`. In this scenario, when the client is told to failover to `nodeId1`, it will keep retrying the request on `OM2` and never reach the actual leader `OM1`. 
So instead of using the nodeId to failover, I think we should utilize the OM address to decide which node to failover to.

adoroszlai:
/pending "instead of using the nodeId to failover, I think we should utilize the OM address to decide which node to failover to"

PR #1675, opened 2020-12-08
HDDS-4563. Increase default value for key and block deletion limit.

lokeshj1703:

Currently ozone.key.deleting.limit.per.task defaults to 1000 and hdds.scm.block.deletion.per-interval.max defaults to 10000. These config values can be increased to 20000 based on the latest block deletion results.

Please refer these links for performance numbers.
https://issues.apache.org/jira/browse/HDDS-4344?focusedCommentId=17237974&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17237974 
https://issues.apache.org/jira/browse/HDDS-4344?focusedCommentId=17242233&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17242233

https://issues.apache.org/jira/browse/HDDS-4563

## How was this patch tested?

Existing UT

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -85,7 +85,7 @@ public final class OMConfigKeys {
 
   public static final String OZONE_KEY_DELETING_LIMIT_PER_TASK =
       "ozone.key.deleting.limit.per.task";
-  public static final int OZONE_KEY_DELETING_LIMIT_PER_TASK_DEFAULT = 1000;
+  public static final int OZONE_KEY_DELETING_LIMIT_PER_TASK_DEFAULT = 20000;
 
   public static final String OZONE_OM_METRICS_SAVE_INTERVAL =
       "ozone.om.save.metrics.interval";
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfig.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfig.java
@@ -79,7 +79,7 @@ public class ScmConfig {
 
   @Config(key = "block.deletion.per-interval.max",
       type = ConfigType.INT,
-      defaultValue = "10000",
+      defaultValue = "20000",
       tags = { ConfigTag.SCM, ConfigTag.DELETION},
       description =
           "Maximum number of blocks which SCM processes during an interval. "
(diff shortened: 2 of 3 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
```diff
@@ -85,7 +85,7 @@ private OMConfigKeys() {
 
   public static final String OZONE_KEY_DELETING_LIMIT_PER_TASK =
       "ozone.key.deleting.limit.per.task";
-  public static final int OZONE_KEY_DELETING_LIMIT_PER_TASK_DEFAULT = 1000;
+  public static final int OZONE_KEY_DELETING_LIMIT_PER_TASK_DEFAULT = 20000;
```
adoroszlai, line 88: Should also be changed in https://github.com/apache/ozone/blob/19dd94d594cfab0308b778dba367ecc62ed0f47e/hadoop-hdds/common/src/main/resources/ozone-default.xml#L396-L397

lokeshj1703: Fixed.

adoroszlai:
> These config values can be increased to 20000 based on the latest block deletion results.

@lokeshj1703 Can you please share these results or the rationale for the change?  I think it would be useful for admins to understand what affects these settings and how they can tune it based on the actual environment Ozone is deployed to.  Otherwise 1000/10000 and 20000 are just opaque numbers (at least to me).

lokeshj1703:
@adoroszlai The results were uploaded in HDDS-4344. 
https://issues.apache.org/jira/browse/HDDS-4344?focusedCommentId=17237974&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17237974 
https://issues.apache.org/jira/browse/HDDS-4344?focusedCommentId=17242233&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17242233

With 20000 keys/min deletion rate at OM, minimum time for OM to delete 1 million blocks is 1000000/20000 = 50 minutes. HDFS take around 60-70 minutes to delete 1 million blocks in a 10 node cluster.

lokeshj1703:
@adoroszlai I have updated these links in the PR description as well.

adoroszlai (approved):
Thanks @lokeshj1703 for pointing me to the comment (and sorry for not looking at the Jira earlier).

bharatviswa504 (approved):
+1 LGTM.
Question: SCM sending 20000 block deletes in a single RPC Call will it have any effect on client writes?

lokeshj1703:
> Question: SCM sending 20000 block deletes in a single RPC Call will it have any effect on client writes?

@bharatviswa504 Block deletion happens in background right now for datanode so the deletions do not happen instantaneously. Datanode would have to store these 20000 blocks in container rocksdb which would lead to some disk usage.

lokeshj1703:
@adoroszlai @bharatviswa504 Thanks for reviewing the PR! I have committed it to master branch.

PR #1676, opened 2020-12-08
HDDS-4560: Add ReadWriteLock into PipelineStateManagerV2Impl to protect contentions between RaftServer and PipelineManager

GlenGeng:

Add ReadWriteLock into PipelineStateManagerV2Impl to protect contentions between RaftServer and PipelineManager.

If current leader SCM (as leader of term N) steps down while there are on-going transactions, there might be read/write contentions that new leader SCM (as leader of term N+1) has replicated some raft log entries encapsulating method addPipeline()/removePipeline()/updatePipelineState() to the underlying RaftServer that are being applied by StateMachineUpdater, and at the same time PipelineManager is calling getPipelines().

https://issues.apache.org/jira/browse/HDDS-4560

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2Impl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineStateManagerV2Impl.java
@@ -35,6 +35,8 @@ import java.lang.reflect.Proxy;
 import java.util.Collection;
 import java.util.List;
 import java.util.NavigableSet;
+import java.util.concurrent.locks.ReadWriteLock;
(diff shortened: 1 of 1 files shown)


GlenGeng:
@ChenSammi @nandakumar131 @runzhiwang  Please take a look! Thanks.

runzhiwang:
without this patch, what will happen ? exception threw ?

GlenGeng:
Yes, since there is no concurrency control.

runzhiwang:
@GlenGeng  Is it possible to make all the data structure thread safe ?

GlenGeng:
> s it possible to make all the data structure thread safe ?

The key point here is synchronize the access to underlying container/pipeline state to ensure the causality, which will be broken by the thread safe data structure..

nandakumar131 (approved):

PR #1677, opened 2020-12-09
HDDS-4562. Old bucket needs to be accessible after the cluster was upgraded to the Quota version.

captainzmc:

long quotaInBytes are new fields in the bucketArgs, this field will go to 0 by default in the old bucket when the old cluster is upgraded. At this point, the data writes are rejected.

This is similar to creating a bucket without specifying quotaInBytes, where quotaInBytes is set to -1 by default via getQuotaValue. We can use 0 as a special term.

image

https://issues.apache.org/jira/browse/HDDS-4562

## How was this patch tested?

ut added

### DIFF
--- a/hadoop-hdds/docs/content/feature/Quota.md
+++ b/hadoop-hdds/docs/content/feature/Quota.md
@@ -35,13 +35,15 @@ Administrators should be able to define how much storage space a Volume or Bucke
 
 a. By default, the quota for volume and bucket is not enabled.
 
-b. When volume quota is enabled, the total size of bucket quota cannot exceed volume.
+b. When volume quota is enabled, the total quota of the buckets, cannot exceed the volume quota.
 
 c. Bucket quota can be set separately without enabling Volume quota. The size of bucket quota is unrestricted at this point.
 
 d. Volume quota is not currently supported separately, and volume quota takes effect only if bucket quota is set. Because ozone only check the usedBytes of the bucket when we write the key.
 
-e. If the cluster is upgraded from old version less than 1.1.0, use of quota on older volumes and buckets is not recommended. Since the old key is not counted to the bucket's usedBytes, the quota setting is inaccurate at this point.
+e. If the cluster is upgraded from old version less than 1.1.0, use of quota on older volumes and buckets(We can confirm by looking at the info for the volume or bucket, and if the quota value is -2 the volume or bucket is old) is not recommended. Since the old key is not counted to the bucket's usedBytes, the quota setting is inaccurate at this point.
+
+f. If volume's quota is enabled then bucket's quota cannot be cleared. 
 
 ## Client usage
 ### Storage Space level quota
@@ -71,9 +73,10 @@ This behavior changes the quota for Bucket1 to 10GB
 
 Total bucket quota should not be greater than its Volume quota. If we have a 10MB Volume, The sum of the sizes of all buckets under this volume cannot exceed 10MB, otherwise the bucket set quota fails.
 
-#### Clear the quota for Volume1. The Bucket cleanup command is similar.
+#### Clear the quota for volume and bucket
 ```shell
 bin/ozone sh volume clrquota --space-quota /volume1
+bin/ozone sh bucket clrquota --space-quota /volume1/bucket1
 ```
 
 #### Check quota and usedBytes for volume and bucket
--- a/hadoop-hdds/docs/content/feature/Quota.md
+++ b/hadoop-hdds/docs/content/feature/Quota.md
@@ -32,11 +32,16 @@ So far, we know that Ozone allows users to create volumes, buckets, and keys. A 
 1. Storage Space level quota
 
 Administrators should be able to define how much storage space a Volume or Bucket can use. The following Settings for Storage space quota are currently supported：
+
 a. By default, the quota for volume and bucket is not enabled.
+
 b. When volume quota is enabled, the total size of bucket quota cannot exceed volume.
+
 c. Bucket quota can be set separately without enabling Volume quota. The size of bucket quota is unrestricted at this point.
+
 d. Volume quota is not currently supported separately, and volume quota takes effect only if bucket quota is set. Because ozone only check the usedBytes of the bucket when we write the key.
 
+e. If the cluster is upgraded from old version less than 1.1.0, use of quota on older volumes and buckets is not recommended. Since the old key is not counted to the bucket's usedBytes, the quota setting is inaccurate at this point.
 
 ## Client usage
 ### Storage Space level quota
--- a/hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
+++ b/hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
@@ -363,7 +363,7 @@ message VolumeInfo {
     optional uint64 objectID = 8;
     optional uint64 updateID = 9;
     optional uint64 modificationTime = 10;
-    optional sint64 quotaInNamespace = 11 [default = -1];
+    optional int64 quotaInNamespace = 11 [default = -2];
     optional uint64 usedNamespace = 12;
 }
 
@@ -503,8 +503,8 @@ message BucketInfo {
     optional string sourceVolume = 12;
     optional string sourceBucket = 13;
     optional uint64 usedBytes = 14;
-    optional sint64 quotaInBytes = 15 [default = -1];
-    optional sint64 quotaInNamespace = 16 [default = -1];
+    optional int64 quotaInBytes = 15 [default = -2];
+    optional int64 quotaInNamespace = 16 [default = -2];
 }
 
 enum StorageTypeProto {
--- a/hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
+++ b/hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
@@ -363,7 +363,7 @@ message VolumeInfo {
     optional uint64 objectID = 8;
     optional uint64 updateID = 9;
     optional uint64 modificationTime = 10;
-    optional uint64 quotaInNamespace = 11;
+    optional sint64 quotaInNamespace = 11 [default = -1];
     optional uint64 usedNamespace = 12;
 }
 
@@ -503,8 +503,8 @@ message BucketInfo {
     optional string sourceVolume = 12;
     optional string sourceBucket = 13;
     optional uint64 usedBytes = 14;
-    optional uint64 quotaInBytes = 15;
-    optional uint64 quotaInNamespace = 16;
+    optional sint64 quotaInBytes = 15 [default = -1];
+    optional sint64 quotaInNamespace = 16 [default = -1];
 }
 
 enum StorageTypeProto {
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketCreateRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketCreateRequest.java
@@ -314,7 +314,7 @@ public class OMBucketCreateRequest extends OMClientRequest {
    */
   private void checkQuotaInNamespace(OmVolumeArgs omVolumeArgs,
       long allocatedNamespace) throws IOException {
-    if (omVolumeArgs.getQuotaInNamespace() != OzoneConsts.QUOTA_RESET) {
+    if (omVolumeArgs.getQuotaInNamespace() > 0) {
       long usedNamespace = omVolumeArgs.getUsedNamespace();
       long quotaInNamespace = omVolumeArgs.getQuotaInNamespace();
       long toUseNamespaceInTotal = usedNamespace + allocatedNamespace;
@@ -335,10 +335,10 @@ public class OMBucketCreateRequest extends OMClientRequest {
     long volumeQuotaInBytes = omVolumeArgs.getQuotaInBytes();
 
     long totalBucketQuota = 0;
-    if (quotaInBytes == OzoneConsts.QUOTA_RESET || quotaInBytes == 0) {
-      return false;
-    } else if (quotaInBytes > OzoneConsts.QUOTA_RESET) {
+    if (quotaInBytes > 0) {
       totalBucketQuota = quotaInBytes;
+    } else {
+      return false;
     }
 
     List<OmBucketInfo>  bucketList = metadataManager.listBuckets(
--- a/hadoop-hdds/docs/content/feature/Quota.zh.md
+++ b/hadoop-hdds/docs/content/feature/Quota.zh.md
@@ -30,10 +30,16 @@ menu:
 1. Storage space级别配额
 
  管理员应该能够定义一个Volume或Bucket可以使用多少存储空间。目前支持以下storage space quota的设置:
+ 
  a. 默认情况下volume和bucket的quota不启用。
+ 
  b. 当volume quota启用时，bucket quota的总大小不能超过volume。
+ 
  c. 可以在不启用volume quota的情况下单独给bucket设置quota。此时bucket quota的大小是不受限制的。
+ 
  d. 目前不支持单独设置volume quota，只有在设置了bucket quota的情况下volume quota才会生效。因为ozone在写入key时只检查bucket的usedBytes。
+ 
+ e. 如果集群从小于1.1.0的旧版本升级而来，则不建议在旧volume和bucket上使用配额。由于旧的key没有计算到bucket的usedBytes中，所以此时配额设置是不准确的。
 
 ## 客户端用法
 ### Storage space级别配额
--- a/hadoop-hdds/docs/content/feature/Quota.zh.md
+++ b/hadoop-hdds/docs/content/feature/Quota.zh.md
@@ -39,7 +39,9 @@ menu:
  
  d. 目前不支持单独设置volume quota，只有在设置了bucket quota的情况下volume quota才会生效。因为ozone在写入key时只检查bucket的usedBytes。
  
- e. 如果集群从小于1.1.0的旧版本升级而来，则不建议在旧volume和bucket上使用配额。由于旧的key没有计算到bucket的usedBytes中，所以此时配额设置是不准确的。
+ e. 如果集群从小于1.1.0的旧版本升级而来，则不建议在旧volume和bucket(可以通过查看volume或者bucket的info确认，如果quota值是-2，那么这个volume或者bucket就是旧的)上使用配额。由于旧的key没有计算到bucket的usedBytes中，所以此时配额设置是不准确的。
+ 
+ f. 如果volume quota被启用，那么bucket quota将不能被清除。
 
 ## 客户端用法
 ### Storage space级别配额
@@ -68,9 +70,10 @@ bin/ozone sh bucket setquota  --space-quota 10GB /volume1/bucket1
 
 bucket的总配额 不应大于其Volume的配额。让我们看一个例子，如果我们有一个10MB的Volume，该volume下所有bucket的大小之和不能超过10MB，否则设置bucket quota将失败。
 
-#### 清除Volume1的配额, Bucket清除命令与此类似
+#### 清除volume和bucket的配额
 ```shell
 bin/ozone sh volume clrquota --space-quota /volume1
+bin/ozone sh bucket clrquota --space-quota /volume1/bucket1
 ```
 #### 查看volume和bucket的quota值以及usedBytes
 ```shell
--- a/hadoop-ozone/dist/src/main/compose/upgrade/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/upgrade/test.sh
@@ -65,9 +65,9 @@ source "${COMPOSE_DIR}/../testlib.sh"
 export OZONE_KEEP_RESULTS=true
 start_docker_env
 execute_robot_test scm -v PREFIX:pre freon/validate.robot
-# TODO can be enabled with HDDS-4562:
-# execute_robot_test scm -v PREFIX:post freon/generate.robot
-# execute_robot_test scm -v PREFIX:post freon/validate.robot
+# test write key to old bucket after upgrade
+execute_robot_test scm -v PREFIX:post freon/generate.robot
+execute_robot_test scm -v PREFIX:post freon/validate.robot
 stop_docker_env
 
 generate_report


adoroszlai (commented):
Thanks @captainzmc for fixing this.  The old bucket is now writeable, reproduced the problem and verified the fix using `upgrade` compose environment.

bharatviswa504 (commented):
Hi @captainzmc 
Few questions:
1. Can we make use of proto default value to -1
2. So on old buckets, we cannot set quota, as we don't have any info of bytesUsed/namespace count, or if it can be set how this will be handled?
3. And how upgrades are handled for quota feature overall, like for older volumes and buckets under it?

bharatviswa504:
Not related to this PR, can you also provide info on how the clear space quota works and it's usage.

Because I see when bucket clear space quota comes, we reset to -1, but what will happen to volume quota (Will it reclaim bucket quota? or any information on clear space quota usage and how it works also can you share some info.

adoroszlai:
@captainzmc see #1691 for acceptance test with [two lines that are currently disabled](https://github.com/apache/ozone/pull/1691/files#diff-1a0d871a569dc91266016ce6723339dc51189088e204b3460b092d925e553c2eR68-R70) to verify this fix.  If you could review that PR, then uncommenting them in this PR would improve test coverage.

captainzmc:
> Hi @captainzmc
> Few questions:
> 
> 1. Can we make use of proto default value to -1 for quotaInBytes value in proto.
> 2. So on old buckets, we cannot set quota, as we don't have any info of bytesUsed/namespace count, or if it can be set how this will be handled?
> 3. And how upgrades are handled for quota feature overall, like for older volumes and buckets under it?

Thanks for @bharatviswa504’s advices.
1. For now, we use getQuotaValue(long quota) to handle the default case in [RpcClient.](https://github.com/apache/ozone/blob/master/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java#L463) And setting the default in proto is also a good way, I'll change that later.
2. After adding usedByets, the original keys of the old bucket cannot be counted, so we can only count the newly written keys. Therefore, we temporarily did not recommend old buckets to enable Quota (because their usedByets were inaccurate).
3. For old buckets we do not recommend that quota be enabled, but in order not to affect the write, I handle the default at [checkBucketQuotaInBytes](https://github.com/apache/ozone/pull/1677/files#diff-2841a017f8d700d802af1b1c49fea9ca281098728796384f5b2e79c23f09ab65R579) in this PR.

captainzmc:
> reclaim
Yes, I have submitted the use of the latest quota in Doc. Can be seen here:
https://ci-hadoop.apache.org/view/Hadoop%20Ozone/job/ozone-doc-master/lastSuccessfulBuild/artifact/hadoop-hdds/docs/public/feature/quota.html

captainzmc:
> Not related to this PR, can you also provide info on how the clear space quota works and it's usage.
> 
> Because I see when bucket clear space quota comes, we reset to -1, but what will happen to volume quota (Will it reclaim bucket quota? or any information on clear space quota usage and how it works also can you share some info.

Yes, I have submitted the use of the latest quota in Doc. Can be seen here:
https://ci-hadoop.apache.org/view/Hadoop%20Ozone/job/ozone-doc-master/lastSuccessfulBuild/artifact/hadoop-hdds/docs/public/feature/quota.html

captainzmc:
> @captainzmc see #1691 for acceptance test with [two lines that are currently disabled](https://github.com/apache/ozone/pull/1691/files#diff-1a0d871a569dc91266016ce6723339dc51189088e204b3460b092d925e553c2eR68-R70) to verify this fix. If you could review that PR, then uncommenting them in this PR would improve test coverage.

Thanks for @adoroszlai 's suggestion, I will deal with the acceptance test here.

bharatviswa504:
> https://ci-hadoop.apache.org/view/Hadoop%20Ozone/job/ozone-doc-master/lastSuccessfulBuild/artifact/hadoop-hdds/docs/public/feature/quota.html

1. When clear volume quota(Assuming volume quota was set before), it is just like setting the quota on the volume to -1, when clear volume quota is called?
2. When volume quota is set, bucket clear quota, it is like just setting the quota for that bucket to -1, and it will allow new buckets can be created if volume quota is still available(As previous bucket quota cleared quota)

When volume quota is not set and bucket quota is set, it is just reset quota of the bucket to -1. when clear bucket quota operation is performed

Is my understanding correct here? 

Example:
V1 - 100 MB
V1/B1 - 50MB
V1/B2- 50MB


Now clear quota on V1/B2 we set quota of V1/B2 to -1, and now a new bucket can be still created in this volume? 

New Bucket created with quota 50MB. (V1/B3)

So, in this volume there are 3 buckets, only 2 are considered in to count, as clearQuota is run(on V1/B2), and we are not considering this bucket in quota calculations, so this bucket will not be counted under quota, but for volume we have crossed the quota of that volume. (As user has run clear quota on V1/B2)

So, here clear quota means resetting the quota to -1? Or what is the real purpose of this clearQuota usage on clusters, in what scenarios this will be useful on the cluster?



>For old buckets, we do not recommend that quota be enabled, but in order not to affect the write, I handle the default at >checkBucketQuotaInBytes in this PR.

But I don't see any guards in the cluster to not allow this for older buckets.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
```diff
@@ -291,9 +291,10 @@ public boolean setVolumeOwner(String volumeName, String owner)
   public void setVolumeQuota(String volumeName, long quotaInCounts,
       long quotaInBytes) throws IOException {
     HddsClientUtils.verifyResourceName(volumeName);
-    verifyCountsQuota(quotaInCounts);
-    verifySpaceQuota(quotaInBytes);
-    ozoneManagerClient.setQuota(volumeName, quotaInCounts, quotaInBytes);
+    verifyCountsQuota(getQuotaValue(quotaInCounts));
```
bharatviswa504, line 294: Why do we need to do this during set Operation?

bharatviswa504:
Currently, Unsigned field can't have negative default value in proto. If we set [default = -1], an error will be compiled.
For now, we use getQuotaValue(long quota) to handle the default case in RpcClient..

Might be use int64, but protobuf document said it is inefficient for negative numbers, not sure it is good idea. Just thought of bringing this up here?

int64 | Uses variable-length encoding. Inefficient for encoding negative numbers – if your field is likely to have negative values, use sint64 instead. | int64 | long | int/long[3] | *int64
-- | -- | -- | -- | -- | --

I think it will also help if we plan to support quota on older buckets.

The default will be -2, all older buckets will have -2, new buckets created after this feature will have -1. So, if one some sets quota on older buckets in this way, we can figure it out.

captainzmc:
> Might be use int64, but protobuf document said it is inefficient for negative numbers, not sure it is good idea. Just thought of bringing this up here?

I'll change the type to sint64, which is both negative by default and more efficient than int64.

captainzmc:
> Example:
> V1 - 100 MB
> V1/B1 - 50MB
> V1/B2- 50MB
> 
> Now clear quota on V1/B2 we set quota of V1/B2 to -1, and now a new bucket can be still created in this volume?
> 
> New Bucket created with quota 50MB. (V1/B3)
> 
> So, in this volume there are 3 buckets, only 2 are considered in to count, as clearQuota is run(on V1/B2), and we are not considering this bucket in quota calculations, so this bucket will not be counted under quota, but for volume we have crossed the quota of that volume. (As user has run clear quota on V1/B2)
> 
> So, here clear quota means resetting the quota to -1? Or what is the real purpose of this clearQuota usage on clusters, in what scenarios this will be useful on the cluster?

Thanks to @bharatviswa504‘s advice, you found a very important point.
Before HDDS-4308, we had usedBytes in volume, so before writing the key, we checked both bucket quota and volume quota. Setting bucket quota to -1 alone at this point does not matter, since we are able to ensure that usedBytes do not exceed volume quota.
But we can't do that now because volume doesn't have usedBytes. I created a new JIRA HDDS-4588 to fix this problem. If volume's quota is enabled then bucket's quota cannot be cleared. We need to prompt the user to clear volume quota first.

> But I don't see any guards in the cluster to not allow quota operations for older buckets/volumes.
> 
> And also do you think, we need to document this in our docs?
I’ll  refine the docs to make it clear that older volumes/buckets are not recommended to use quota.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
captainzmc: This check simply ensures that when we use RpcClient set Quota, the value of quota must be legal (not less than -1). But the getQuotaValue conversion step inside is redundant,  I‘ll  delete this conversion.

captainzmc:
When the acceptance test upgrade, namespace will also have the problem that old volumes cannot create buckets. This also fixed in this PR.  R: @amaliujia @linyiqun 
image

captainzmc:
Updated PR and fixed review issues.

hadoop-hdds/docs/content/feature/Quota.md
```diff
@@ -32,11 +32,16 @@ So far, we know that Ozone allows users to create volumes, buckets, and keys. A
...
+
 a. By default, the quota for volume and bucket is not enabled.
+
 b. When volume quota is enabled, the total size of bucket quota cannot exceed volume.
+
 c. Bucket quota can be set separately without enabling Volume quota. The size of bucket quota is unrestricted at this point.
+
 d. Volume quota is not currently supported separately, and volume quota takes effect only if bucket quota is set. Because ozone only check the usedBytes of the bucket when we write the key.
 
+e. If the cluster is upgraded from old version less than 1.1.0, use of quota on older volumes and buckets is not recommended. Since the old key is not counted to the bucket's usedBytes, the quota setting is inaccurate at this point.
```
linyiqun, line 44: We can make this check in OM side and throw exception once user try to set quota on older version volumes/buckets.

hadoop-hdds/docs/content/feature/Quota.md
```diff
@@ -32,11 +32,16 @@ So far, we know that Ozone allows users to create volumes, buckets, and keys. A
 1. Storage Space level quota
```
linyiqun, line 32: The namespace quota introduction can also be added. We could file a new JIRA for tacking this.

linyiqun (commented):
@captainzmc , current PR change looks good to me. But I catch some places we can improve in follow-up JIRA.

bharatviswa504:
> > Example:
> > V1 - 100 MB
> > V1/B1 - 50MB
> > V1/B2- 50MB
> > Now clear quota on V1/B2 we set quota of V1/B2 to -1, and now a new bucket can be still created in this volume?
> > New Bucket created with quota 50MB. (V1/B3)
> > So, in this volume there are 3 buckets, only 2 are considered in to count, as clearQuota is run(on V1/B2), and we are not considering this bucket in quota calculations, so this bucket will not be counted under quota, but for volume we have crossed the quota of that volume. (As user has run clear quota on V1/B2)
> > So, here clear quota means resetting the quota to -1? Or what is the real purpose of this clearQuota usage on clusters, in what scenarios this will be useful on the cluster?
> 
> Thanks to @bharatviswa504‘s advice, you found a very important point.
> Before [HDDS-4308](https://issues.apache.org/jira/browse/HDDS-4308), we had usedBytes in volume, so before writing the key, we checked both bucket quota and volume quota. Setting bucket quota to -1 alone at this point does not matter, since we are able to ensure that usedBytes do not exceed volume quota.
> But we can't do that now because volume doesn't have usedBytes. I created a new JIRA [HDDS-4588](https://issues.apache.org/jira/browse/HDDS-4588) to fix this problem. If volume's quota is enabled then bucket's quota cannot be cleared. We need to prompt the user to clear volume quota first.
> 
> > But I don't see any guards in the cluster to not allow quota operations for older buckets/volumes.
> > And also do you think, we need to document this in our docs?
> 
> I’ll refine the docs to make it clear that older volumes/buckets are not recommended to use quota.

So, now if the volume quota is not set, the bucket Clearspace quota is just disabling quota on the buckets.

And also just setting quota at volume level will not be counted, until bucket level quotas are set.

For my understanding so volume quota is set, and bucket quota is set, clearing volume quota is just set to -1, so that now quota is tracked at bucket level. And now when bucket quota is cleared, when volume quota is set, we disallow the operation.
This will be the behavior of clear space quota, let me know if i am missing something.
 
> But we can't do that now because volume doesn't have usedBytes. I created a new JIRA [HDDS-4588 
>(https://issues.apache.org/jira/browse/HDDS-4588) to fix this problem. If volume's quota is enabled then bucket's quota >cannot be cleared. We need to prompt the user to clear volume quota first.

Now the example mentioned scenarios will not happen. Thanks for opening this Jira and tracking this issue.

bharatviswa504:
>The default will be -2, all older buckets will have -2, new buckets created after this feature will have -1. So, if some one sets >quota on older buckets in this way, we can figure it out, just an idea saying here.

If we want to disallow the quota support for older buckets, one idea to do at code level.

hadoop-hdds/docs/content/feature/Quota.md
```diff
@@ -32,11 +32,16 @@ So far, we know that Ozone allows users to create volumes, buckets, and keys. A
...
 
 Administrators should be able to define how much storage space a Volume or Bucket can use. The following Settings for Storage space quota are currently supported：
+
 a. By default, the quota for volume and bucket is not enabled.
+
 b. When volume quota is enabled, the total size of bucket quota cannot exceed volume.
```
bharatviswa504, line 38: Minor:
When volume quota is enabled, the total quota of the buckets, cannot exceed the volume quota.

hadoop-hdds/docs/content/feature/Quota.md
bharatviswa504, line 44: Can we also add documentation about the clear space quota and the behavior?

bharatviswa504 (commented):
Overall LGTM.
Few comments, and one question regarding handling old buckets/volumes.

amaliujia:
@captainzmc 

Can we use `0` to indicate that a quota is not set? Will this better handle the cluster upgrade issue that some buckets by default has 0 quota? 

If a case is really set "0" quota to a bucket thus lock the bucket from creating, then users can choose to use a negative value.

hadoop-hdds/docs/content/feature/Quota.md
amaliujia: I can add namespace quota to this documentation: https://issues.apache.org/jira/browse/HDDS-4594. Before adding to documentation, I will first finish namespace support on bucket: https://issues.apache.org/jira/browse/HDDS-4277

amaliujia:
O, actually from the code seems that setting default value as -1 is solved in proto. If so please ignore my previous comment (which assumed that proto cannot set -1 as default value for quota)

ChenSammi:
> > The default will be -2, all older buckets will have -2, new buckets created after this feature will have -1. So, if some one sets >quota on older buckets in this way, we can figure it out, just an idea saying here.
> 
> If we want to disallow the quota support for older buckets, one idea to do at code level.

Is it feasible that we provide a option of calculating the bytesInUsed for each bucket during 1.0.0 -> 1.1. upgrade to user?  @adoroszlai

captainzmc:
> So, now if the volume quota is not set, the bucket Clearspace quota is just disabling quota on the buckets.
> 
> And also just setting quota at volume level will not be counted, until bucket level quotas are set.
> 
> For my understanding so volume quota is set, and bucket quota is set, clearing volume quota is just set to -1, so that now quota is tracked at bucket level. And now when bucket quota is cleared, when volume quota is set, we disallow the operation.
> This will be the behavior of clear space quota, let me know if i am missing something.

Yes. 
And, the problem of clear quota will be fixed in HDDS-4588.

hadoop-hdds/docs/content/feature/Quota.md
captainzmc: > Can we also add documentation about the clear space quota and the behavior?

In HDDS-4588, I will modify the current clear logic, and also modify the document of clear space quota.

captainzmc:
> > > The default will be -2, all older buckets will have -2, new buckets created after this feature will have -1. So, if some one sets >quota on older buckets in this way, we can figure it out, just an idea saying here.
> > 
> > 
> > If we want to disallow the quota support for older buckets, one idea to do at code level.
> 
> User using old Ozone version may still has the desire to use the new Quota feature. It's better not shut the door completely.
> @adoroszlai , I'm not sure if it's feasible that we provide a option of calculating the bytesInUsed for each bucket during 1.0.0 -> 1.1. upgrade to user? So that user has the choice either tolerate the inaccurate bytesInUsed or a relative longer upgrade time to get accurate bytesInUsed of buckets.

Hi @ChenSammi, I think it's dangerous recalculating usedBytes in upgrade . If the amount of data is too large, it will greatly increase the time, and if the processing is not good, it will cause the upgrade to fail.
If we want use Quota in old buckets, we can add a command to recalculate bucket usedBytes to minimize the impact.

hadoop-hdds/docs/content/feature/Quota.md
captainzmc: Currently, we have not agreed on whether bucket can be enabled for the old quota. At present, we only put forward the suggestion in the document very briefly. If the requirement is clear, we can issue Jira and solve it.

bharatviswa504:
> > > The default will be -2, all older buckets will have -2, new buckets created after this feature will have -1. So, if some one sets >quota on older buckets in this way, we can figure it out, just an idea saying here.
> > 
> > 
> > If we want to disallow the quota support for older buckets, one idea to do at code level.
> 
> User using old Ozone version may still has the desire to use the new Quota feature. It's better not shut the door completely.
> @adoroszlai , I'm not sure if it's feasible that we provide a option of calculating the bytesInUsed for each bucket during 1.0.0 -> 1.1. upgrade to user? So that user has the choice either tolerate the inaccurate bytesInUsed or a relative longer upgrade time to get accurate bytesInUsed of buckets.

@ChenSammi Above idea is to detect older buckets/volumes in the Ozone cluster, from discussion as it is said not recommended, said we can fail the operation But, if we want to support, when they setQuota, we can know this is an old bucket, we can say a warning message to the user, older keys are not considered.

bharatviswa504:
> > > > The default will be -2, all older buckets will have -2, new buckets created after this feature will have -1. So, if some one sets >quota on older buckets in this way, we can figure it out, just an idea saying here.
> > > 
> > > 
> > > If we want to disallow the quota support for older buckets, one idea to do at code level.
> > 
> > 
> > User using old Ozone version may still has the desire to use the new Quota feature. It's better not shut the door completely.
> > @adoroszlai , I'm not sure if it's feasible that we provide a option of calculating the bytesInUsed for each bucket during 1.0.0 -> 1.1. upgrade to user? So that user has the choice either tolerate the inaccurate bytesInUsed or a relative longer upgrade time to get accurate bytesInUsed of buckets.
> 
> Hi @ChenSammi, I think it's dangerous recalculating usedBytes in upgrade . If the amount of data is too large, it will greatly increase the time, and if the processing is not good, it will cause the upgrade to fail.
> If we want use Quota in old buckets, we can add a command to recalculate bucket usedBytes to minimize the impact. User can decide for himself which buckets to recalculate.

Yes during upgrade might not be a feasible idea, it will take a long time for an upgrade if there are many buckets/keys in the cluster.


> If we want use Quota in old buckets, we can add a command to recalculate bucket usedBytes to minimize the impact. User can decide for himself which buckets to recalculate.
Even this will be a costly operation, during that operation we should acquire bucket read lock and do the calculation, so all writes will be stalled in the system. (As during this operation, we should not allow new writes to get accurate bytesused).

captainzmc:
Updated PR, set default to -2.  No matter whether or not we support old volume/bucket setting quota, we need to distinguish which volumes and buckets are old. So Here I take @bharatviswa504's advice.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketCreateRequest.java
```diff
@@ -314,7 +314,9 @@ private BucketEncryptionInfoProto getBeinfo(
    */
   private void checkQuotaInNamespace(OmVolumeArgs omVolumeArgs,
       long allocatedNamespace) throws IOException {
-    if (omVolumeArgs.getQuotaInNamespace() != OzoneConsts.QUOTA_RESET) {
+    // Client has ensured that no negative values other than -1 (clear quota)
+    // will occur when quota is set.
+    if (omVolumeArgs.getQuotaInNamespace() > OzoneConsts.QUOTA_RESET) {
```
linyiqun, line 319: Why we change above condition check? Above check just means that we enabled the quota, and current quota isn't equal to default value. This should be a more accurate check.

hadoop-hdds/docs/content/feature/Quota.md
linyiqun: Make sense to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketCreateRequest.java
captainzmc: @linyiqun Here with omVolumeArgs.getQuotaInNamespace () > 0 is enough? Because when it > 0 is certainly is not equal to -1.

linyiqun: @captainzmc , here my point is that we should firstly check if the quota was enabled by using one variable flag(here I understand is OzoneConsts.QUOTA_RESET). Then we check the quota value if it's valid. Is there a switch config for the quota now?  If the quota is not enabled, we don't need to do the quota check anymore.

captainzmc: Current there are two case  -1 or -2 (old volume/bucket) not check quota, actually we have guarantee on the client, [user cann’t set negative quota](https://github.com/apache/ozone/blob/master/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java#L194), Therefore, except for clear quota and Old Quoat, there is no negative case.

captainzmc: > Current there are two case -1 or -2 (old volume/bucket) not check quota, actually we have guarantee on the client, [user cann’t set negative quota](https://github.com/apache/ozone/blob/master/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java#L194), Therefore, except for clear quota and Old Quoat, there is no negative case.
> 
>If add omVolumeArgs. getQuotaInNamespace ()! = 2. The judgment condition will increase too much and the code will not be clear enough, we can use omVolumeArgs. getQuotaInNamespace () > 0.

linyiqun: Okay, go ahead for this.

hadoop-hdds/docs/content/feature/Quota.md
captainzmc: Yes. To prevent conflicts, I have updated the usage and behavior of clear space quota in this PR.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -503,8 +503,8 @@ message BucketInfo {
     optional string sourceVolume = 12;
     optional string sourceBucket = 13;
     optional uint64 usedBytes = 14;
-    optional uint64 quotaInBytes = 15;
-    optional uint64 quotaInNamespace = 16;
+    optional int64 quotaInBytes = 15 [default = -2];
```
bharatviswa504, line 506: https://github.com/apache/ozone/pull/1677#issuecomment-745019611
Any reason for moving back to int64?

captainzmc: Using sint64, I found that the default value changed to -9223372036854775808 after switching to long. This could be a compatibility issue. So I switched to int64.
image

bharatviswa504: Interesting, thanks for info.

bharatviswa504 (approved):
+1 LGTM.

bharatviswa504:
If no more comments, I will commit this by tomorrow.

bharatviswa504:
Thank you @captainzmc for the contribution and everyone for the review.

PR #1679, opened 2020-12-09
HDDS-4514. AllocateBlock : lookup and update open file table for the given path

rakeshadr:

This task is to use open file table for the allocate block operations. This has been identified as part of Hive TPCDS benchmark test

https://issues.apache.org/jira/browse/HDDS-4514

## How was this patch tested?
Added UT cases. Also, verified by running hive tpcds test

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
@@ -37,6 +37,7 @@ import org.apache.hadoop.hdds.utils.db.cache.CacheKey;
 import org.apache.hadoop.hdds.utils.db.cache.CacheValue;
 import org.apache.hadoop.ozone.OzoneAcl;
 import org.apache.hadoop.ozone.OzoneConsts;
+import org.apache.hadoop.ozone.om.OzoneManager;
 import org.apache.hadoop.ozone.om.OMMetadataManager;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
 import org.apache.hadoop.ozone.om.helpers.OmBucketInfo;
@@ -52,6 +53,8 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 
+import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.KEY_NOT_FOUND;
+import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.NOT_A_FILE;
 
 /**
  * Base class for file requests.
@@ -545,6 +548,7 @@ public final class OMFileRequest {
           throws IOException {
 
     OmKeyInfo dbOmKeyInfo;
+    OmKeyInfo omKeyInfoCopy = null;
     if (openFileTable) {
       dbOmKeyInfo = omMetadataMgr.getOpenKeyTable().get(dbOpenFileKey);
     } else {
@@ -556,9 +560,10 @@ public final class OMFileRequest {
     // For example, the user given key path is '/a/b/c/d/e/file1', then in DB
     // keyName field stores only the leaf node name, which is 'file1'.
     if (dbOmKeyInfo != null) {
-      dbOmKeyInfo.setKeyName(keyName);
+      omKeyInfoCopy = dbOmKeyInfo.copyObject();
+      omKeyInfoCopy.setKeyName(keyName);
     }
-    return dbOmKeyInfo;
+    return omKeyInfoCopy;
   }
 
   /**
@@ -850,5 +855,73 @@ public final class OMFileRequest {
 
   public static boolean isImmediateChild(long parentId, long ancestorId) {
     return parentId == ancestorId;
+  }
+
+
+  /**
+   * Check for directory exists with same name, if it exists throw error.
+   *
+   * @param keyName                  key name
+   * @param ozoneManager             Ozone Manager
+   * @param reachedLastPathComponent true if the path component is a fileName
+   * @throws IOException if directory exists with same name
+   */
+  private static void checkDirectoryAlreadyExists(String keyName,
+      OzoneManager ozoneManager, boolean reachedLastPathComponent)
+          throws IOException {
+    // Reached last component, which would be a file. Returns its parentID.
+    if (reachedLastPathComponent && ozoneManager.getEnableFileSystemPaths()) {
+      throw new OMException("Can not create file: " + keyName +
+              " as there is already directory in the given path", NOT_A_FILE);
+    }
+  }
+
+  /**
+   * Get parent id for the user given path.
+   *
+   * @param bucketId       bucket id
+   * @param pathComponents fie path elements
+   * @param keyName        user given key name
+   * @param ozoneManager   ozone manager
+   * @return lastKnownParentID
+   * @throws IOException DB failure or parent not exists in DirectoryTable
+   */
+  public static long getParentID(long bucketId, Iterator<Path> pathComponents,
+      String keyName, OzoneManager ozoneManager) throws IOException {
+
+    long lastKnownParentId = bucketId;
+
+    // If no sub-dirs then bucketID is the root/parent.
+    if(!pathComponents.hasNext()){
+      return bucketId;
+    }
+
+    OmDirectoryInfo omDirectoryInfo;
+    OMMetadataManager omMetadataManager = ozoneManager.getMetadataManager();
+    while (pathComponents.hasNext()) {
+      String nodeName = pathComponents.next().toString();
+      boolean reachedLastPathComponent = !pathComponents.hasNext();
+      String dbNodeName =
+              omMetadataManager.getOzonePathKey(lastKnownParentId, nodeName);
+
+      omDirectoryInfo = omMetadataManager.
+              getDirectoryTable().get(dbNodeName);
+      if (omDirectoryInfo != null) {
+        checkDirectoryAlreadyExists(keyName, ozoneManager,
+                reachedLastPathComponent);
+        lastKnownParentId = omDirectoryInfo.getObjectID();
+      } else {
+        // One of the sub-dir doesn't exists in DB. Immediate parent should
+        // exists for committing the key, otherwise will fail the operation.
+        if (!reachedLastPathComponent) {
+          throw new OMException("Failed to commit key, as parent directory of "
+                  + keyName + " entry is not found in DirectoryTable",
+                  KEY_NOT_FOUND);
+        }
+        break;
+      }
+    }
+
+    return lastKnownParentId;
   }
 }
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
@@ -37,7 +37,6 @@ import org.apache.hadoop.hdds.utils.db.cache.CacheKey;
 import org.apache.hadoop.hdds.utils.db.cache.CacheValue;
 import org.apache.hadoop.ozone.OzoneAcl;
 import org.apache.hadoop.ozone.OzoneConsts;
-import org.apache.hadoop.ozone.om.OzoneManager;
 import org.apache.hadoop.ozone.om.OMMetadataManager;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
 import org.apache.hadoop.ozone.om.helpers.OmBucketInfo;
@@ -548,7 +547,6 @@ public final class OMFileRequest {
           throws IOException {
 
     OmKeyInfo dbOmKeyInfo;
-    OmKeyInfo omKeyInfoCopy = null;
     if (openFileTable) {
       dbOmKeyInfo = omMetadataMgr.getOpenKeyTable().get(dbOpenFileKey);
     } else {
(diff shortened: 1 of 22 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMAllocateBlockResponseV1.java
```diff
@@ -0,0 +1,63 @@
...
+  public void addToDBBatch(OMMetadataManager omMetadataManager,
+      BatchOperation batchOperation) throws IOException {
+
+    OMFileRequest.addToOpenFileTable(omMetadataManager, batchOperation,
+            getOmKeyInfo(), getClientID());
+
+    // update volume usedBytes.
+    omMetadataManager.getVolumeTable().putWithBatch(batchOperation,
+            omMetadataManager.getVolumeKey(getOmVolumeArgs().getVolume()),
+            getOmVolumeArgs());
```
linyiqun, line 57: volume usedbytes update can be removed

linyiqun (commented):
Please address below minor comment and fix failed unit test, : ).

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMAllocateBlockResponseV1.java
rakeshadr: Thanks @linyiqun  for the review comments. Fixed in latest commit.

rakeshadr:
> Please address below minor comment and fix failed unit test, : ).

TestOzoneFileSystemV1#testListStatusOnLargeDirectory is failing in the build, interestingly it is passing consistently in my local env. Added log messages to debug it. Will monitor the results in next build.

rakeshadr:
@linyiqun I've addressed your comments. Could you please review it again when you get a chance. Thanks!

linyiqun (approved):
+1.
Thanks for addressing the comments, @rakeshadr .

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequestV1.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Iterator;
+
+/**
+ * Handles allocate block request layout version V1.
+ */
+public class OMAllocateBlockRequestV1 extends OMAllocateBlockRequest {
```
bharatviswa504, line 44: Just a question:
V1 only supports filesystem.paths.enabled true only and old code will be used when filesystem.paths.enabled false.
I think, we had a discussion some time back, can you provide info.

Is that is the reason only code for true mode is only handled in V1?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequestV1.java
```diff
@@ -0,0 +1,91 @@
...
+  protected String getOpenKeyName(String volumeName, String bucketName,
+      String keyName, long clientID, OzoneManager ozoneManager)
+          throws IOException {
+    OMMetadataManager omMetadataManager = ozoneManager.getMetadataManager();
+    String bucketKey = omMetadataManager.getBucketKey(volumeName, bucketName);
+    OmBucketInfo omBucketInfo =
+            omMetadataManager.getBucketTable().get(bucketKey);
+    long bucketId = omBucketInfo.getObjectID();
+    String fileName = OzoneFSUtils.getFileName(keyName);
+    Iterator<Path> pathComponents = Paths.get(keyName).iterator();
```
bharatviswa504, line 69: Not holding bucket lock during traversing of key to get parentID
Do you think we will have a race condition with delete or any parent directories?

bharatviswa504: On a side note anyway I think it will be handled during commit Key anyway.
Thoughts?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequestV1.java
rakeshadr: Thanks @bharatviswa504  for the review comments.

I've HDDS-4486 jira task to make proper feature configuration. Presently, I followed same pattern for the V1 classes. Hope thats fine for you?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequestV1.java
rakeshadr: Good catch. Yes, during commit it will throw exception. But somehow I feel it is safer to move #getOpenKeyInfo() under BUCKET_LOCK and this would avoid race conditions and will make consistency to the system. What do you say?

On the other side, moving getOpenKeyInfo() inside lock is adding any additional overhead and performance penalty ? If not, then I will move it inside the lock.

bharatviswa504: Yes, I agree the ideal thing should be doing this under bucket lock whole operation. But that will have performance impact, as previously allocate block was done with out lock. Previous code has a similar issue, even then during key commit, it might not be found out, and also before we don't have atomic delete, so nothing can be really guaranteed.

In new approach deletes are atomic, and when delete only parent dir is removed, and childs are deleted asynchronously. With new approach, this will be figured out during key commit (As we iterate the entire path again). But it will have impact on performance, and client may write some additional block data even when parent dir's are removed. (This is the only a problem for fs semantics, not object-store semantics)

So, here we need to make a choice, do we want to detect during allocate block or during commit Key.

cc @mukul1987 @arp7 for comments

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequestV1.java
bharatviswa504: Yes. I will take one final look at it (as previously skimmed the entire patch) and post my comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -851,4 +856,72 @@ private static boolean checkSubFileExists(OmKeyInfo omKeyInfo,
...
+   *
+   * @param keyName                  key name
+   * @param ozoneManager             Ozone Manager
+   * @param reachedLastPathComponent true if the path component is a fileName
+   * @throws IOException if directory exists with same name
+   */
+  private static void checkDirectoryAlreadyExists(String keyName,
+      OzoneManager ozoneManager, boolean reachedLastPathComponent)
+          throws IOException {
+    // Reached last component, which would be a file. Returns its parentID.
```
bharatviswa504, line 872: Minor: Comment says Returns its parentID.

In code we just check if lastComponent is true and fileSystem enabled, we return exception.

And also as current V1 always assumes fs enabled true, do we need && ozoneManager.getEnableFileSystemPaths() check?

And also these 3 lines can be in the calling method, as it is not a utility method which is called by different classes.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -216,17 +216,15 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+      addOpenTableCacheEntry(trxnLogIndex, omMetadataManager, openKeyName,
+          openKeyInfo);
       omBucketInfo.incrUsedBytes(preAllocatedSpace);
 
       omResponse.setAllocateBlockResponse(AllocateBlockResponse.newBuilder()
           .setKeyLocation(blockLocation).build());
-      omClientResponse = new OMAllocateBlockResponse(omResponse.build(),
-          openKeyInfo, clientID, omVolumeArgs, omBucketInfo.copyObject());
 
+      omClientResponse = getOmClientResponse(clientID, omResponse,
```
bharatviswa504, line 226: We don't update omVolumeArgs, but still reading and passing to Response, which is not required.
Not related to your patch BTW, we can fix this in master and get to branch.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/key/TestOMAllocateBlockResponseV1.java
```diff
@@ -0,0 +1,84 @@
...
+    return new OMAllocateBlockResponseV1(omResponse, omKeyInfo, clientID,
+            omVolumeArgs, omBucketInfo);
+  }
+
+  @NotNull
+  @Override
+  protected OzoneConfiguration getOzoneConfiguration() {
+    OzoneConfiguration config = super.getOzoneConfiguration();
+    config.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION, "V1");
+    // OzoneManager#start() is not invoked in this test cases. Hence need to
```
bharatviswa504, line 78: Not understood this comment and why we need to do this, and is there in many tests, can you provide info why we need to do this?

bharatviswa504: Is this for getKeyTable where we used this in metadatamanager?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequestV1.java
rakeshadr: As per the offline discussion with @mukul1987 and @arp7, it is not required to do the get parentID check under BUCKET_LOCK. With this current code flow, it won't make fast fail during race condition case, anyway later it will fail at the key commit operation time and there won't be any correctness issues. Basically it would be good to avoid any extra performance overhead and in future this part can be revisited if someone has any potential issue and required fast fail in allocate block request.

Following is the expected behavior during race condition. Here op-2 arrived first in OM server and simultaneously op-3 also arrived in OM server.
```
    op-1:   client-1) Creates openFile  - /a/b/c/d/file1 and written block1
    op-2:   client-2) Invoked delete /a/b/c. This acquired BUCKET_LOCK and performing deletion.
    op-3:   client-1) Invoked AllocateBlock - block2 for /a/b/c/d/file1. Since getParent("/a/b/c/d/file1") is not under BUCKET_LOCK, it saw "/a/b/c/d" parent ID and proceeds to successfully allocate block2.
    op-4:   client-1) CommitKey /a/b/c/d/file1 - Fail as parent doesn't exist.
```
@bharatviswa504, does this sound good to you?

bharatviswa504: I am fine with that, if we have consensus.
As, previous thought of not acquiring bucket lock for allocateBlock is the same reason.

rakeshadr: Thanks @bharatviswa504 for the quick response.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -556,9 +560,10 @@ public static OmKeyInfo getOmKeyInfoFromFileTable(boolean openFileTable,
     // For example, the user given key path is '/a/b/c/d/e/file1', then in DB
     // keyName field stores only the leaf node name, which is 'file1'.
     if (dbOmKeyInfo != null) {
-      dbOmKeyInfo.setKeyName(keyName);
```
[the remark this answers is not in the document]
rakeshadr: Agreed, Done the changes!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
rakeshadr: Agreed, Done the changes!

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/key/TestOMAllocateBlockResponseV1.java
rakeshadr: Yes, its basically for initializing prefix tables. I've updated comment, hope its clear now.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -219,14 +219,13 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
       omMetadataManager.getOpenKeyTable().addCacheEntry(
           new CacheKey<>(openKeyName),
           new CacheValue<>(Optional.of(openKeyInfo), trxnLogIndex));
-
```
bharatviswa504, line 222: Just space only changes.

bharatviswa504 (commented):
+1 LGTM

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
rakeshadr: Removed unnecessary spaces!

rakeshadr:
Thank you @linyiqun , @bharatviswa504 for the useful review comments. Merging it to the feature branch.

PR #1680, opened 2020-12-09
HDDS-4570. Reduce memory footprint of OMUpdateEventBatch for Recon

smengcl:

1. Reduce new object creation when Recon is consuming batches OM events (`consumeOMEvents()`)
2. ~~Improve `OMUpdateEventBatch#filter` to reduce memory usage.~~ Ended up removing `OMUpdateEventBatch#filter`. Instead just "filters" the events by task table name in `ReconOmTask#process`

https://issues.apache.org/jira/browse/HDDS-4570

## How was this patch tested?

All existing UTs shall pass.

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/ContainerKeyMapperTask.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/tasks/ContainerKeyMapperTask.java
@@ -113,8 +113,14 @@ public class ContainerKeyMapperTask implements ReconOmTask {
   public Pair<String, Boolean> process(OMUpdateEventBatch events) {
     Iterator<OMDBUpdateEvent> eventIterator = events.getIterator();
     int eventCount = 0;
+    final Collection<String> taskTables = getTaskTables();
+
     while (eventIterator.hasNext()) {
       OMDBUpdateEvent<String, OmKeyInfo> omdbUpdateEvent = eventIterator.next();
+      // Filter event inside process method to avoid duping
+      if (!taskTables.contains(omdbUpdateEvent.getTable())) {
+        continue;
+      }
       String updatedKey = omdbUpdateEvent.getKey();
       OmKeyInfo updatedKeyValue = omdbUpdateEvent.getValue();
       try {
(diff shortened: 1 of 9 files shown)


linyiqun (commented):
Here the problem is that we fetch too much DB update events to apply to recon db, right? How about doing the update event size limitation each time we get from OM? I see currently we just fetch all update events since from last sequence number.

smengcl:
> Here the problem is that we fetch too much DB update events to apply to recon db, right? How about doing the update event size limitation each time we get from OM? I see currently we just fetch all update events since from last sequence number.

Getting smaller batches is indeed another viable approach.

The current plan is to reduce the number of array copy when Recon is processing those events. At the moment the memory usage is very bloated (may use >2.0x memory).

avijayanhwx:
@linyiqun A good point. Currently, we can control the number of events implicitly through the Recon-OM delta updates sync interval. We are also bound by the RPC buffer limit (~128MB). But, in the future we may add more downstream tasks in Recon to act on OM data, and we cannot scale the number of copies of the event data based on number of tasks :)

smengcl:
Marking this PR as ready for review.

Existing UTs should cover. No new UTs added.

prashantpogde:
The changes look good.
is it possible to get some stats as to how much memory footprint is reduced for some sample run (pre/post these changes) ?

prashantpogde (approved):

prashantpogde:
Ci failures are not related with the changes.

avijayanhwx (approved):
Thanks for working on this @smengcl. Changes LGTM +1. Can we remove org.apache.hadoop.ozone.recon.tasks.ReconOmTask#getTaskTables since it is no longer at the interface level?

arp7:
@smengcl can you check if the integration test failures are related to the patch?

smengcl:
Thanks @prashantpogde and @arp7 for checking the CI. `TestContainerStateMachine` and `TestKeyInputStream` looks unrelated. I'm running them locally to add confidence to that.

I'm removing `ReconOmTask#getTaskTables` from the interface as @avijayanhwx suggested. Should just take another few hours to get a new CI run done.

smengcl:
> Here the problem is that we fetch too much DB update events to apply to recon db, right? How about doing the update event size limitation each time we get from OM? I see currently we just fetch all update events since from last sequence number.

btw just filed https://issues.apache.org/jira/browse/HDDS-4678 for this.

avijayanhwx:
> ~@smengcl can you check if the integration test failures are related to the patch?~
> 
> Looks like @prashantpogde already checked that. @avijayanhwx does your review comment need to be addressed prior to merge? Wasn't sure since you also added a +1. 🙂

I am OK with the approach in this patch. Just had a minor suggestion for @smengcl :)

smengcl:
Thanks @prashantpogde and @avijayanhwx for reviewing. Got a green run on the latest commit. I will merge this shortly.

PR #1682, opened 2020-12-10
HDDS-4574. Inconsistencies in case the parent of a open file is deleted/renamed.

ayushtkn:

Check whether parent exist before commit key to counter inconsistencies due to it. Scenario mentioned in the jira.

https://issues.apache.org/jira/browse/HDDS-4574

## How was this patch tested?

Added UT.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
@@ -30,6 +30,7 @@ import org.apache.hadoop.ozone.OmUtils;
 import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.om.OMConfigKeys;
 import org.apache.hadoop.ozone.om.helpers.OmBucketInfo;
+import org.apache.hadoop.ozone.om.helpers.OzoneFSUtils;
 import org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerDoubleBufferHelper;
 import org.apache.hadoop.ozone.om.request.util.OmResponseUtil;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer;
@@ -162,6 +163,14 @@ public class OMKeyCommitRequest extends OMKeyRequest {
             omMetadataManager)) {
           throw new OMException("Can not create file: " + keyName +
               " as there is already directory in the given path", NOT_A_FILE);
+        }
+        // Ensure the parent exist.
+        if (!"".equals(OzoneFSUtils.getParent(keyName))
+            && !checkDirectoryAlreadyExists(volumeName, bucketName,
+            OzoneFSUtils.getParent(keyName), omMetadataManager)) {
+          throw new OMException("Cannot create file : " + keyName
+              + " as parent directory doesn't exist",
+              OMException.ResultCodes.DIRECTORY_NOT_FOUND);
         }
       }
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemMissingParent.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystemMissingParent.java
@@ -1,0 +1,126 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.fs.ozone;
+
+import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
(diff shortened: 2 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -172,6 +173,16 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
         throw new OMException("Failed to commit key, as " + dbOpenKey +
             "entry is not found in the OpenKey table", KEY_NOT_FOUND);
       }
+
+      // Ensure the parent exist.
+      if (!"".equals(OzoneFSUtils.getParent(keyName))
```
rakeshadr, line 178: Can you move this check inside `ozoneManager.getEnableFileSystemPaths() == true` case to do this only for FileSystem api. 

Should we need this check for bucket#createKey & OzoneOutputStream#close() in turn make OMKeyCommitRequest call?

rakeshadr (commented):
Thanks @ayushtkn for the contribution. Added few comments.

[OMAllocateBlockRequest#L191](https://github.com/apache/ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java#L191) has similar case, how about failing there as well?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
ayushtkn: Thanx @rakeshadr for the review.
I have moved the check inside `ozoneManager.getEnableFileSystemPaths() == true` and added the same logic and tests for `OMAllocateBlockRequest` as well. Please have a look..

mukul1987:
@ayushtkn can you please rebase the patch ?

ayushtkn:
Thanx @mukul1987 I have rebased the PR

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMAllocateBlockRequest.java
```diff
@@ -192,6 +193,17 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
             KEY_NOT_FOUND);
       }
 
+      if (ozoneManager.getEnableFileSystemPaths()) {
+        // Ensure the parent exist.
+        if (!"".equals(OzoneFSUtils.getParent(keyName))
```
bharatviswa504, line 198: Just a question: Right now delete/rename is not an atomic operation, so will just checking the parent Dir is enough?


But this will help, once we have HDDS-2939 where we have atomic delete/renames.

bharatviswa504: And also for performance reasons, can we leave allocateBlock, as during KeyCommit anyway we shall figure it out?

As this will save one extra DB op, (this might not be a problem again once HDDS-2939 comes in, as it has directory cache)

ayushtkn: Thanx @bharatviswa504 for the review.
Rakesh suggested to add the check in `allocateBlock`, I think for a `fail-fast` approach, If we tend to fail during the commit phase and we know that at the `allocateBlock` phase, we should fail at an earlier stage itself, rather than going till the end. Might save a messed up client, if he is writing a big file.

But I am not very aware about the performance impacts and HDDS-2939 both, so I am ok both ways.

@rakeshadr any thoughts on this?

rakeshadr: >And also for performance reasons, can we leave allocateBlock, as during KeyCommit anyway we shall figure it out?

On a second thought, it makes sense to me to avoid any additional DB calls which can affect the OM performance. Most of the cases, parent directories would exists in RocksDB offheap cache even then I would agree to skip parentDir check considering the cost of DB call if not exist.

On the other side, I think there won't be any correctness issue if we skip parent dir check in AllocateBlockReq. Later, we can bring this` fast-fail` check based on user demand. Considering that, I'm +1 to skip extra DB check in AllocateBlockReq now.

Thanks a lot @ayushtkn for your patience and taking care this. I agree to revert the changes from OmAllocateBlockRequest.

>But I am not very aware about the performance impacts and HDDS-2939 both

In HDDS-2939, in order to fetch openFileName, it requires to traverse top-down model and parentDir check won't add any additional overhead. For example, say open file /a/b/c/d/file1.

_HDDS-2939 branch_: OM will fetch /a, /a/b, /a/b/c, /a/b/c/d from DirTable and finally get /a/b/c/d/file1 from OpenFileTable. Here, one idea is to build DirCache to reduce DB overhead. @bharatviswa504 was talking about this part in his previous comment.
_Master branch_: OM will do point look up by fetching /a/b/c/d from KeyTable. Here, existing RocksDB offheap cache would help.

ayushtkn: Thanx Rakesh for sharing the details, makes sense, Will explore more :-)

I have removed the check from `AllocateBlock` as suggested.

@bharatviswa504 Please give a check once agin, if things make sense now. Thanx...

bharatviswa504: >Is there anything else, that can be checked? Yahh, there might be some issue because of non atomic behaviour delete/rename. Will pulling this check inside lock, solve them or lead to a better state?

As due to non-atomic behavior whole path check can help here. But this issue will be solved once HDDS-2939 comes in. And also this check is done under lock even right now. Let's have this way for now.

bharatviswa504 (approved):
+1 LGTM.

bharatviswa504:
Thank You @ayushtkn for the contribution and @rakeshadr for the review.

PR #1683, opened 2020-12-10
HDDS-4575: Refactor SCMHAManager and SCMRatisServer with RaftServer.Divison

GlenGeng:

Simplify SCMHAManager and SCMRatisServer with RaftServer.Division

https://issues.apache.org/jira/browse/HDDS-4575

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
@@ -23,7 +23,6 @@ import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
@@ -42,6 +41,7 @@ import org.apache.ratis.protocol.RaftGroup;
 import org.apache.ratis.protocol.RaftGroupId;
 import org.apache.ratis.protocol.RaftPeer;
 import org.apache.ratis.protocol.RaftPeerId;
+import org.apache.ratis.protocol.exceptions.NotLeaderException;
 import org.apache.ratis.server.RaftServer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/MockSCMHAManager.java
```diff
@@ -204,23 +180,27 @@ private Message process(final SCMRatisRequest request)
...
     @Override
-    public RaftServer getServer() {
-      return null;
+    public void stop() {
     }
 
     @Override
-    public RaftGroupId getRaftGroupId() {
+    public RaftServer.Division getDivision() {
       return null;
```
timmylicheng, line 188: Do we allow Division to be null?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -69,41 +65,49 @@
...
+
+    RaftServer server = RaftServer.newBuilder()
+        .setServerId(haGrpBuilder.getPeerId())
+        .setGroup(haGrpBuilder.getRaftGroup())
         .setProperties(serverProperties)
-        .setStateMachine(scmStateMachine)
+        .setStateMachine(new SCMStateMachine())
         .build();
+
+    this.division = server.getDivision(haGrpBuilder.getRaftGroupId());
```
timmylicheng, line 80: Any way to guarantee division is not null?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/MockSCMHAManager.java
GlenGeng: No, we don't allow null Divison in real env.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
GlenGeng: You won't get a null, instead, the call will throw a `GroupMismatchException` in this situation.

ChenSammi:
+1

PR #1684, opened 2020-12-10
HDDS-4576. TestDefaultCertificateClient misuses chars param of random()

adoroszlai:

`TestDefaultCertificateClient` passes "UTF-8" to `RandomStringUtils.random(int, String)` with the intention to use UTF-8 encoding.  It was added in HDDS-1087 to fix CI failures.

However, the parameter is not for character encoding, rather:

```
     * @param chars  the String containing the set of characters to use,
     *  may be null, but must not be empty
```

So this results in values like: `"--8F8T8U8T..."`.

This PR proposes another fix:

1. Allow `random()` to generate strings using all characters.
2. Fix test failure by specifying UTF-8 encoding when calling `String.getBytes()`.

https://issues.apache.org/jira/browse/HDDS-4576

## How was this patch tested?

Reproduced HDDS-1087 by removing `chars` argument:
https://github.com/adoroszlai/hadoop-ozone/runs/1530939654#step:3:3638

Verified the new fix:
https://github.com/adoroszlai/hadoop-ozone/runs/1530664863#step:3:3779

### DIFF
--- a/hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/client/TestDefaultCertificateClient.java
+++ b/hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/client/TestDefaultCertificateClient.java
@@ -28,7 +28,6 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


xiaoyuyao (approved):

xiaoyuyao:
LGTM, +1.

adoroszlai:
Thanks @xiaoyuyao for reviewing and committing it.

PR #1688, opened 2020-12-11
HDDS-3378. OzoneManager group init failed because of incorrect snapshot directory location.

bharatviswa504:

Make ratis.snapshot.dir do not depend on ozone.om.ratis.storage.dir, so that ozone.om.ratis.storage.dir does not have any other directories.

And also if ratis.storage.dir is not defined, the default we fall back to ozone.metadata.dirs, then we will have ratis.storage.dir as ozone.metadata.dirs + "/ratis".

And for older clusters, the directory will be moved to new ratis.snapshot.dir, if snapshot directory exists in ozone.om.ratis.storage.dir.

https://issues.apache.org/jira/browse/HDDS-3378

## How was this patch tested?

Existing tests.


bharatviswa504:
Thank You @hanishakoneru for the offline discussion.
I have posted the PR, as discussed to solve the issue for fresh instllation and also for upgrades.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -464,16 +465,33 @@ private OzoneManager(OzoneConfiguration conf) throws IOException,
...
+      // Before starting ratis server check, if previous installation has
+      // snapshot directory in Ratis storage directory if so, move it to
+      // snapshot directory.
+      File[] dirs = new File(omRatisDirectory).listFiles();
+
+      if (dirs != null) {
+        for (File dir : dirs) {
+          if (dir.isDirectory() && dir.getName().equals("snapshot")) {
+            FileUtils.moveDirectory(dir.toPath(), omRatisSnapshotDir.toPath());
+            break;
```
hanishakoneru, line 482: Instead of listing the files and trying to find if snapshot subdir exists, why don't we construct the snapshot subdir path and check it that exists and move it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -464,16 +465,33 @@ private OzoneManager(OzoneConfiguration conf) throws IOException,
...
           OzoneManagerRatisServer.getOMRatisSnapshotDirectory(configuration));
 
+      // Before starting ratis server check, if previous installation has
+      // snapshot directory in Ratis storage directory if so, move it to
+      // snapshot directory.
+      File[] dirs = new File(omRatisDirectory).listFiles();
+
+      if (dirs != null) {
+        for (File dir : dirs) {
+          if (dir.isDirectory() && dir.getName().equals("snapshot")) {
```
hanishakoneru, line 480: Can we define "snapshot" as a constant in OzoneConsts and use that here and in getOMRatisSnapshotDirectory().

hanishakoneru (commented):
Thanks @bharatviswa504 for fixing this.
Patch LGTM overall. I have minor comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
bharatviswa504: Yes, it can be, thanks for the suggestion updated it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
bharatviswa504: Done

hanishakoneru:
Thanks Bharat for updating the patch. 
We can use the Const OM_RATIS_SNAPSHOT_DIR in OzoneManagerRatisServer#getOMRatisSnapshotDirectory() also. 

I am +1 with that fixed. Thanks.

bharatviswa504:
Thank You @hanishakoneru for the review. Fixed it.

hanishakoneru (approved):

bharatviswa504:

PR #1691, opened 2020-12-11
HDDS-4579. Add acceptance test for Ozone Client Key Validator

adoroszlai:

1. Add acceptance test for key validator (`freon ockv`).
2. Make current Freon robot test (of generators) reusable by setting prefix from parameter instead of random one.
3. Remove `freon rk` from this reusable test.
   a. it's already exercised in other tests
   b. it's also being deprecated in favor of newer generators
4. Remove useless `Wait Until Keyword Succeeds` from Freon robot test: no point in repeating the `Should contain` check every 10 second since `${result}` is not updated.
5. Exercise Freon robot test in `upgrade` environment instead of more "manual" steps from `topology` test.  This allows us to test generator+validator both before and after upgrade (post-upgrade generator can be enabled with/after HDDS-4562).
6. Fix the bug that re-loading `testlib.sh` defines invalid `OM_HA_PARAM` for non-HA OM setup.  It happens because empty `OM_SERVICE_ID` is set to `om` the first time, which makes `testlib.sh` think it's HA in the second round.

https://issues.apache.org/jira/browse/HDDS-4579

## How was this patch tested?

```
cd hadoop-ozone/dist/target/ozone-1.1.0-SNAPSHOT/compose/upgrade
./test.sh
```

https://github.com/adoroszlai/hadoop-ozone/runs/1537554565?check_suite_focus=true#step:7:1732
https://github.com/adoroszlai/hadoop-ozone/runs/1537554565?check_suite_focus=true#step:7:1742
https://github.com/adoroszlai/hadoop-ozone/runs/1537554565?check_suite_focus=true#step:7:1815


captainzmc (approved):
+1, LGTM. Thanks for the patch.

adoroszlai:
Thanks @captainzmc for reviewing and committing it.

PR #1692, opened 2020-12-11
HDDS-4564. Prepare client should check every OM individually for the prepared check based on Txn ID.

avijayanhwx:

- After getting a successful response for the OMPrepareRequest, the prepare client should use the Txn ID from the response to check every OM's preparation completeness.
- This JIRA is partially dependent on HDDS-4569 which plans to introduce a marker file at the end of preparation. In a follow up JIRA, this will be refined to take up that logic, along with an integration test.
- A basic acceptance test for prepare has been added. 
- User facing message screenshot has been attached (om1, om2, om3 are actually hostnames in the screenshot).

Screen Shot 2020-12-11 at 11 15 03 AM

https://issues.apache.org/jira/browse/HDDS-4564

## How was this patch tested?
Manually tested. 
Added acceptance test.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
@@ -609,7 +609,7 @@ public interface OzoneManagerProtocol
       throws IOException {
     return PrepareStatusResponse.newBuilder()
         .setCurrentTxnIndex(-1)
-        .setStatus(PrepareStatus.NOT_IN_PREPARE_STATE)
+        .setStatus(PrepareStatus.PREPARE_NOT_STARTED)
         .build();
   }
 }
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
@@ -49,6 +49,8 @@ import org.apache.hadoop.ozone.om.helpers.ServiceInfo;
 import org.apache.hadoop.ozone.om.helpers.ServiceInfoEx;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OzoneAclInfo;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.PrepareStatusResponse;
+import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.PrepareStatusResponse.PrepareStatus;
 import org.apache.hadoop.ozone.security.OzoneDelegationTokenSelector;
 import org.apache.hadoop.ozone.security.acl.OzoneObj;
 import org.apache.hadoop.ozone.upgrade.UpgradeFinalizer.StatusAndMessages;
@@ -601,5 +603,13 @@ public interface OzoneManagerProtocol
       long txnApplyWaitTimeoutSeconds, long txnApplyCheckIntervalSeconds)
       throws IOException {
     return -1;
+  }
+
+  default PrepareStatusResponse getOzoneManagerPrepareStatus(long txnId)
+      throws IOException {
+    return PrepareStatusResponse.newBuilder()
+        .setCurrentTxnIndex(-1)
+        .setStatus(PrepareStatus.NOT_IN_PREPARE_STATE)
+        .build();
   }
 }
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
@@ -605,6 +605,12 @@ public interface OzoneManagerProtocol
     return -1;
   }
 
+  /**
+   * Check if Ozone Manager is 'prepared' at a specific Txn Id.
+   * @param txnId passed in Txn Id
+   * @return PrepareStatus response
+   * @throws IOException on exception.
+   */
   default PrepareStatusResponse getOzoneManagerPrepareStatus(long txnId)
       throws IOException {
     return PrepareStatusResponse.newBuilder()
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
@@ -1574,6 +1574,17 @@ public final class OzoneManagerProtocolClientSideTranslatorPB
     return prepareResponse.getTxnID();
   }
 
+  public PrepareStatusResponse getOzoneManagerPrepareStatus(long txnId)
+      throws IOException {
+    PrepareStatusRequest prepareStatusRequest =
+        PrepareStatusRequest.newBuilder().setTxnID(txnId).build();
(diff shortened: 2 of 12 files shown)


avijayanhwx:
cc @errose28 for review.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
```diff
@@ -602,4 +604,12 @@ default long prepareOzoneManager(
       throws IOException {
     return -1;
   }
+
+  default PrepareStatusResponse getOzoneManagerPrepareStatus(long txnId)
```
linyiqun, line 608: Please also document this javadoc for this method.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
```diff
@@ -1574,6 +1574,17 @@ public long prepareOzoneManager(
     return prepareResponse.getTxnID();
   }
 
+  public PrepareStatusResponse getOzoneManagerPrepareStatus(long txnId)
```
linyiqun, line 1577: Can you add the @Override here that will let us know this is a protocol method?

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/PrepareSubCommand.java
```diff
@@ -67,13 +78,85 @@
...
+      }
+      if (currentNumPreparedOms < expectedNumPreparedOms) {
+        System.out.println("Waiting for " + prepareCheckInterval +
+            " seconds before retrying...");
+        Thread.sleep(pInterval.toMillis());
+      }
+    }
+    if (currentNumPreparedOms < expectedNumPreparedOms) {
+      throw new Exception("OM Preparation failed since all OMs are not " +
+          "prepared yet.");
```
linyiqun, line 151: As this is the command for admin users, can we print out this message instead of throwing an exception? That will be more friendly to users.

linyiqun (commented):
@avijayanhwx , Most of this looks good to me, few minor comments from me. Please have a check.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocol/OzoneManagerProtocol.java
avijayanhwx: Fixed.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/protocolPB/OzoneManagerProtocolClientSideTranslatorPB.java
avijayanhwx: Fixed.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/PrepareSubCommand.java
avijayanhwx: Thanks for the review @linyiqun. I believe throwing an exception is better since the return code for the command will be non-zero. This will help in higher level applications automating this call figure out the result of the command decisively.


> Exception while checking preparation completeness for [om2], Error : [Invalid host name: local host is: (unknown); >destination host is: "om2":9862; java.net.UnknownHostException; For more details see:  >http://wiki.apache.org/hadoop/UnknownHost]
>Waiting for 10 seconds before retrying...
>OM Preparation failed since all OMs are not prepared yet.
>bash-4.2$ echo $?
>255

linyiqun: Okay, make sense.

errose28 (commented):
Thanks for working on this @avijayanhwx, I just left a few comments/questions in line.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
```diff
@@ -1091,6 +1094,23 @@ message PrepareResponse {
...
+    enum PrepareStatus {
+        // TODO
+        // HDDS-4569 may introduce new states here, like marker file found
+        // but with different txn id. We can add them as make sense.
+        PREPARE_NOT_STARTED = 1;
+        PREPARE_IN_PROGRESS = 2;
+        PREPARE_COMPLETED = 3;
+    }
+    required PrepareStatus status = 1;
+    optional uint64 currentTxnIndex = 2;
```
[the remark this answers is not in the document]
avijayanhwx: @errose28 Yes, I agree. Since I expect some changes in the status enum fields as well as the logic in OzoneManagerRequestHandler#getPrepareStatus (usage of marker file vs OM transaction info), I believe we can punt the documentation to a follow up patch after HDDS-4569. That is the reason I have not added an integration test as well.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
```diff
@@ -623,6 +633,24 @@ private FinalizeUpgradeProgressResponse reportUpgradeProgress(
...
+  private PrepareStatusResponse getPrepareStatus(PrepareStatusRequest request)
+      throws IOException {
+    // TODO After HDDS-4569,
+    // When there is a global "prepared" state in OM, we can return
+    // PREPARE_NOT_STARTED instead of PREPARE_IN_PROGRESS appropriately.
+    PrepareStatus prepareStatus = null;
+    long txnID = request.getTxnID();
+    long ratisSnapshotIndex = impl.getRatisSnapshotIndex();
+    if (ratisSnapshotIndex != txnID) {
+      LOG.info("Last Txn Index = {}", ratisSnapshotIndex);
```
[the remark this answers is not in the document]
avijayanhwx: Yes, I can add that log line.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerRequestHandler.java
```diff
@@ -623,6 +633,24 @@ private FinalizeUpgradeProgressResponse reportUpgradeProgress(
...
+      prepareStatus =  PREPARE_IN_PROGRESS;
```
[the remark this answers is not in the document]
avijayanhwx: As mentioned in an earlier comment, this can undergo some minor changes after HDDS-4569. Currently, in my mind, an OM in prepared flag disabled state (state maintained in global OM level) does not know about a OmPrepareRequest (think slow follower). When the flag has been enabled but still is waiting for the flush, then it is likely in PREPARE_IN_PROGRESS mode. When the flag is enabled, and the flush is done (OMPR apply txn is fully complete), then we are expected to be PREPARE_COMPLETED state. Of course there are some edge cases that can be caused by subsequent OMPrepareRequests coming in the pipeline, but I don't expect them to be a big problem to the fundamental problem the prepare operation is trying to solve.

hadoop-ozone/interface-client/src/main/proto/OmClientProtocol.proto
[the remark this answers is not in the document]
errose28: Ok makes sense. Good for me to keep in mind while doing HDDS-4569.

avijayanhwx:
Ran the test 20 times and verified that no failures are seen (https://github.com/avijayanhwx/hadoop-ozone/runs/1553559592)  Please ignore the overall test run. It has failed due to incomplete github workflow changes. The test itself successfully ran 20 times.
[summary.txt](https://github.com/apache/ozone/files/5692222/summary.txt)

swagle (approved):

avijayanhwx:
Thanks for the reviews @linyiqun, @errose28 & @swagle!

PR #1693, opened 2020-12-11
HDDS-4581. Cleanup usage of volumeArgs in KeyRequests.

bharatviswa504:

Cleanup usage of VolumeArgs in Request/Response classes in Key requests.

https://issues.apache.org/jira/browse/HDDS-4581

## How was this patch tested?

Existing tests.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -279,7 +279,6 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           .collect(Collectors.toList());
       omKeyInfo.appendNewBlocks(newLocationList, false);
 
-      omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
```
linyiqun, line 282: Why we not remove omVolumeArgs variable defined in this method?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -193,7 +193,6 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       long scmBlockSize = ozoneManager.getScmBlockSize();
       int factor = omKeyInfo.getFactor().getNumber();
-      omVolumeArgs = getVolumeInfo(omMetadataManager, volumeName);
       omBucketInfo = getBucketInfo(omMetadataManager, volumeName, bucketName);
```
linyiqun, line 196: Same comment for above.

linyiqun (commented):
Few minor comments from me.
In additional, with compared the change in HDD-4308. S3MultipartUploadAbortRequest/TestS3MultipartUploadAbortRequest is missed updated here.

bharatviswa504:
Thank You @linyiqun for the review, addressed review comments in the latest update.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
bharatviswa504: I missed it, thanks for catching it.

linyiqun:
@bharatviswa504 , would you mind checking if the failed unit related to this PR? If not, I am +1 for this PR.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -279,7 +277,6 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
```
captainzmc, line 282: All the use of the getVolumeInfo method has been removed, so getVolumeInfo should not be needed, we can [delete it.]( https://github.com/apache/ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java#L620)

captainzmc (commented):
LGTM over all, just one comment. CI failure may have nothing to do with this PR and we can retrigger it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
bharatviswa504: Done

captainzmc (approved):
+1 LGTM.

linyiqun (approved):
+1.

captainzmc:
Thanks for @bharatviswa504’s patch. Also thanks for the review of @linyiqun. Merged this.

PR #1703, opened 2020-12-15
HDDS-4591. BasicOzoneFileSystem reports StorageUnit class not found.

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-4591

### DIFF
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -158,12 +158,7 @@ public class BasicOzoneFileSystem extends FileSystem {
           .build();
       LOG.trace("Ozone URI for ozfs initialization is {}", uri);
 
-      ConfigurationSource source;
-      if (conf instanceof OzoneConfiguration) {
-        source = (ConfigurationSource) conf;
-      } else {
-        source = new LegacyHadoopConfigurationSource(conf);
-      }
+      ConfigurationSource source = getConfSource();
       this.adapter =
           createAdapter(source, bucketStr,
               volumeStr, omHost, omPort);
@@ -703,14 +698,7 @@ public class BasicOzoneFileSystem extends FileSystem {
 
   @Override
   public long getDefaultBlockSize() {
-    Configuration conf = getConf();
-    ConfigurationSource source;
-    if (conf instanceof OzoneConfiguration) {
-      source = (ConfigurationSource) conf;
-    } else {
-      source = new LegacyHadoopConfigurationSource(conf);
-    }
-    return (long)source.getStorageSize(
+    return (long)getConfSource().getStorageSize(
         OZONE_SCM_BLOCK_SIZE, OZONE_SCM_BLOCK_SIZE_DEFAULT, StorageUnit.BYTES);
   }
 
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -703,7 +703,14 @@ public boolean mkdirs(Path f, FsPermission permission) throws IOException {
 
   @Override
   public long getDefaultBlockSize() {
-    return (long) getConf().getStorageSize(
+    Configuration conf = getConf();
+    ConfigurationSource source;
```
xiaoyuyao, line 707: NIT: can we wrap this as a helper function as I can see the initialize() has similar logic.

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -703,7 +703,14 @@ public boolean mkdirs(Path f, FsPermission permission) throws IOException {
 
   @Override
   public long getDefaultBlockSize() {
```
xiaoyuyao, line 705: We need a fix for this in BasicRootedOzoneFileSystem#getDefaultBlockSize() as well.

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -703,7 +703,14 @@ public boolean mkdirs(Path f, FsPermission permission) throws IOException {
...
-    return (long) getConf().getStorageSize(
+    Configuration conf = getConf();
+    ConfigurationSource source;
+    if (conf instanceof OzoneConfiguration) {
+      source = (ConfigurationSource) conf;
+    } else {
+      source = new LegacyHadoopConfigurationSource(conf);
+    }
```
adoroszlai, line 712: Instead of duplicating this logic from `initialize`:

https://github.com/apache/ozone/blob/09579756b0756fcaf1b22b860c1ee1eb927e82c2/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java#L161-L166

I think it would be cleaner to store this `ConfigurationSource`

* either in `BasicOzoneFileSystem`
* or in `BasicOzoneClientAdapterImpl` (and expose it via accessor method)

and then use it here.

xiaoyuyao (commented):
LGTM, a few minor comments inline.

adoroszlai (changes requested):
Thanks @ChenSammi for reporting this bug.  It turns out that acceptance test has false negatives (HDDS-4592).

We need similar fix for `BasicRootedOzoneFileSystem`, too.

ChenSammi:
Thanks @xiaoyuyao  and @adoroszlai  for the code review. I have uploaded a new commit to address the comments.

adoroszlai (approved):
Thanks @ChenSammi for updating the patch.  `hadoop27` test is now passing:

https://github.com/apache/ozone/pull/1703/checks?check_run_id=1562056505#step:7:208
https://github.com/apache/ozone/pull/1703/checks?check_run_id=1562056505#step:7:234

adoroszlai:

adoroszlai:
#1714 fixes the problem that this bug was not caught by acceptance tests.

PR #1705, opened 2020-12-15
HDDS-4569. Add pre append gate and marker file to OM prepare state.

errose28:

Add a gate to the pre append step of OzoneManagerStateMachine and submit request of OzoneManagerRatisServer to block further write requests from coming in when an ozone manager is prepared. Also add a marker file that is written when the OM is prepared so that if it is restarted without a special flag, it automatically enters prepare mode.

This pull request updates the Ratis snapshot to the latest version which includes the necessary RATIS-1216 fix, and addresses backwards incompatible changes made as a result of the update. The following 12 files were updated as a result of this Ratis change only; their changes are not related to the prepare gate or marker file:

- ContainerStateMachine
- XceiverServerRatis
- ContainerTestHelper
- TestHelper
- TestHadoopTreeDirGenerator
- TestHadoopNestedTreeDirGenerator
- TestOzoneClientKeyGenerator
- TestOMRatisSnapshots
- OMTransactionInfo
- OMRatisSnapshotInfo
- TestOzonemanagerRatisServer
- pom.xml

HDDS-4569

## How was this patch tested?

Unit and integration tests added.

## Notes

There are a few outstanding issues related to prepare, including HDDS-4610 and RATIS-1241 that are causing flakiness in the prepare integration tests. It may not be possible to get a clean CI run on this feature without these fixes.


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -138,75 +112,111 @@ public void testPrepareWithTransactions() throws Exception {
...
+              .stream()
+              .anyMatch((vol) -> vol.getName().equals(volumeName)));
+        } catch (ExecutionException ex) {
+          Throwable cause = ex.getCause();
+          Assert.assertTrue(cause instanceof OMException);
+          Assert.assertEquals(OMException.ResultCodes.NOT_SUPPORTED_OPERATION,
+              ((OMException) cause).getResult());
+        }
       }
     }
```
linyiqun, line 221: Can you closed executorService after the test.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -128,10 +129,25 @@ private static long nextCallId() {
...
+          createWriteRaftClientRequest(omRequest);
+      RaftClientReply raftClientReply = submitRequestToRatis(raftClientRequest);
+
+      return processReply(omRequest, raftClientReply);
+    } else {
+      String message = "Cannot apply write request " +
+          omRequest.getCmdType().name() + " when OM is in prepare mode.";
+      OMResponse.Builder omResponse = OMResponse.newBuilder()
+          .setMessage(message)
+          .setStatus(OzoneManagerProtocolProtos.Status.NOT_SUPPORTED_OPERATION)
```
linyiqun, line 145: Can we use a better status name for this, like OM_IN_PREPARE_MODE?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -193,6 +196,32 @@ public TransactionContext startTransaction(
     return handleStartTransactionRequests(raftClientRequest, omRequest);
   }
 
+  public TransactionContext preAppendTransaction(TransactionContext trx)
```
linyiqun, line 199: Can you add @Override here, so we could directly know this is an override method.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
```diff
@@ -184,26 +188,18 @@ private static void waitForLogIndex(long indexToWaitFor,
...
     long snapshotIndex = stateMachine.takeSnapshot();
-    RaftLog raftLog = impl.getState().getLog();
-    long raftLogIndex = raftLog.getLastEntryTermIndex().getIndex();
-
-    // Ensure that Ratis's in memory snapshot index is the same as the index
-    // of its last log entry.
-    if (snapshotIndex != raftLogIndex) {
-      throw new IOException("Snapshot index " + snapshotIndex + " does not " +
-          "match last log index " + raftLogIndex);
-    }
```
linyiqun, line 203: Not fully get this, why we remove this safe check?

linyiqun (commented):
@errose28 , some minor comments from me.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
errose28: Yes, just fixed in latest revision.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
errose28: Yes, I have created a new dedicated status for rejecting requests because the om is in prepare mode. Thanks for the suggestion!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
errose28: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
errose28: It is possible that an extra Ratis transaction to commit the prepare request happens after we take the snapshot. This entry is internal to Ratis so it is not sent to the state machine. This is normal, but with this check in place, it would cause the prepare to fail. HDDS-4610 will wait on this extra transaction and update the DB with its index when the snapshot is taken, which will provide consistent behavior here, instead of the transaction getting included only sometimes.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManagerPrepareState.java
```diff
@@ -0,0 +1,279 @@
...
+            "marker file.", expectedPrepareIndex, prepareMarkerIndex);
+      } else {
+        // Prepare state can only be restored if we read the expected index
+        // from the marker file.
+        prepareRestored = true;
+      }
+    }
+
+    if (prepareRestored) {
+      enablePrepareGate();
```
avijayanhwx, line 180: There is an 'enablePrepareGate()' in finishPrepare. Do we need the one in Line # 180?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -193,6 +196,32 @@ public TransactionContext startTransaction(
     return handleStartTransactionRequests(raftClientRequest, omRequest);
   }
 
+  @Override
+  public TransactionContext preAppendTransaction(TransactionContext trx)
```
avijayanhwx, line 200: Can we add a unit test for this method since it has a few different execution flows?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManagerPrepareState.java
```diff
@@ -0,0 +1,279 @@
...
+      throw new IOException("Failed to create necessary directories in " +
+          "marker file path: " + parentDir);
+    }
+
+    try(FileOutputStream stream = new FileOutputStream(markerFile)) {
+      stream.write(Long.toString(index).getBytes(StandardCharsets.UTF_8));
+    }
+  }
+
+  private void deletePrepareMarkerFile()
```
avijayanhwx, line 238: Can we add a log line when we create & delete the marker file? It will help us with troubleshooting issues with preparation.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerPrepare.java
```diff
@@ -138,77 +112,116 @@ public void testPrepareWithTransactions() throws Exception {
...
+          // If no exception is thrown, the volume should be created.
+          future.get();
+          String volumeName = VOLUME + i;
+          Assert.assertTrue(clientProtocol.listVolumes(volumeName, "", 1)
+              .stream()
+              .anyMatch((vol) -> vol.getName().equals(volumeName)));
+        } catch (ExecutionException ex) {
+          Throwable cause = ex.getCause();
+          Assert.assertTrue(cause instanceof OMException);
+          Assert.assertEquals(OMException.ResultCodes.NOT_SUPPORTED_OPERATION,
```
avijayanhwx, line 217: I believe this should be NOT_SUPPORTED_OPERATION_WHEN_PREPARED.

avijayanhwx (commented):
Thanks for working on this @errose28. The approach looks good to me. The management of the OM prepare state and related transition operations looks very clean. A few comments inline. 

Also, the CI failures are related to this patch.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/upgrade/OMPrepareRequest.java
linyiqun: Okay, get it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
errose28: Yes, done.

avijayanhwx:
Ignoring CI failure since it will be handled in HDDS-4610.

avijayanhwx (approved):

PR #1706, opened 2020-12-16
HDDS-4277.  Support Bucket Namespace Quota Updates

amaliujia:

add namespaceQuotaUsage and update it when create and delete key in a bucket

https://issues.apache.org/jira/browse/HDDS-4277

## How was this patch tested?

UT

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -946,7 +946,6 @@ public abstract class TestOzoneRpcClientAbstract {
     OzoneBucket bucket = null;
 
     String value = "sample value";
-    int countException = 0;
 
     store.createVolume(volumeName);
     volume = store.getVolume(volumeName);
@@ -964,19 +963,18 @@ public abstract class TestOzoneRpcClientAbstract {
 
     try {
       writeKey(bucket, key3, ONE, value, value.length());
+      Assert.fail("Write key should be failed");
     } catch (IOException ex) {
-      countException++;
       GenericTestUtils.assertExceptionContains("QUOTA_EXCEEDED", ex);
     }
 
-    // Write failed, bucket usedBytes should remain as 1
+    // Write failed, bucket usedBytes should remain as 2
     Assert.assertEquals(2L,
         store.getVolume(volumeName).getBucket(bucketName).getUsedNamespace());
 
     bucket.deleteKeys(Arrays.asList(key1, key2));
     Assert.assertEquals(0L,
         store.getVolume(volumeName).getBucket(bucketName).getUsedNamespace());
-    Assert.assertEquals(1, countException);
   }
 
   private void writeKey(OzoneBucket bucket, String keyName,
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -935,6 +935,50 @@ public abstract class TestOzoneRpcClientAbstract {
     Assert.assertEquals(0L, volume.getUsedNamespace());
   }
 
+  @Test
+  public void testBucketUsedNamespace() throws IOException {
+    String volumeName = UUID.randomUUID().toString();
+    String bucketName = UUID.randomUUID().toString();
+    String key1 = UUID.randomUUID().toString();
+    String key2 = UUID.randomUUID().toString();
+    String key3 = UUID.randomUUID().toString();
+    OzoneVolume volume = null;
+    OzoneBucket bucket = null;
+
+    String value = "sample value";
+    int countException = 0;
+
+    store.createVolume(volumeName);
+    volume = store.getVolume(volumeName);
+    volume.createBucket(bucketName);
+    bucket = volume.getBucket(bucketName);
+    bucket.setQuota(OzoneQuota.parseQuota(Long.MAX_VALUE + " Bytes", 2));
+
+    writeKey(bucket, key1, ONE, value, value.length());
+    Assert.assertEquals(1L,
+        store.getVolume(volumeName).getBucket(bucketName).getUsedNamespace());
+
+    writeKey(bucket, key2, ONE, value, value.length());
+    Assert.assertEquals(2L,
+        store.getVolume(volumeName).getBucket(bucketName).getUsedNamespace());
+
+    try {
+      writeKey(bucket, key3, ONE, value, value.length());
+    } catch (IOException ex) {
+      countException++;
+      GenericTestUtils.assertExceptionContains("QUOTA_EXCEEDED", ex);
+    }
+
+    // Write failed, bucket usedBytes should remain as 1
+    Assert.assertEquals(2L,
+        store.getVolume(volumeName).getBucket(bucketName).getUsedNamespace());
+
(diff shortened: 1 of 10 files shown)


amaliujia:
R @captainzmc

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -932,6 +932,50 @@ public void testVolumeUsedNamespace() throws IOException {
...
+        store.getVolume(volumeName).getBucket(bucketName).getUsedNamespace());
+
+    writeKey(bucket, key2, ONE, value, value.length());
+    Assert.assertEquals(2L,
+        store.getVolume(volumeName).getBucket(bucketName).getUsedNamespace());
+
+    try {
+      writeKey(bucket, key3, ONE, value, value.length());
+    } catch (IOException ex) {
+      countException++;
```
linyiqun, line 965: countException is not actually needed, we could add fail() after writeKey and ensure this exception should be happened.
```java
 try {
      writeKey(bucket, key3, ONE, value, value.length());
      Assert.fail("Write key should be failed")
    } catch (IOException ex) {
```

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -932,6 +932,50 @@ public void testVolumeUsedNamespace() throws IOException {
...
+      GenericTestUtils.assertExceptionContains("QUOTA_EXCEEDED", ex);
+    }
+
+    // Write failed, bucket usedBytes should remain as 1
+    Assert.assertEquals(2L,
```
linyiqun, line 970: This comment should be updated

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
```diff
@@ -299,6 +299,10 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       omBucketInfo.incrUsedBytes(preAllocatedSpace);
 
+      // Update namespace quota
+      checkBucketQuotaInNamespace(omBucketInfo, 1L);
+      omBucketInfo.incrUsedNamespace(1L);
```
linyiqun, line 304: We should move checkBucketQuotaInNamespace check after checkBucketQuotaInBytes. Here we add cache entry before quota check, this will lead the dirty data.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
```diff
@@ -309,6 +309,10 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
 
       omBucketInfo.incrUsedBytes(preAllocatedSpace);
 
+      // Update namespace quota
+      checkBucketQuotaInNamespace(omBucketInfo, 1L);
```
linyiqun, line 313: Same comment like above.

linyiqun (commented):
Most of change looks good to me. Some minor comments frome.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileCreateRequest.java
amaliujia: O I see.

amaliujia:
@captainzmc @linyiqun comments addressed. Can you take a another look please?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -961,19 +960,18 @@ public void testBucketUsedNamespace() throws IOException {
...
     try {
       writeKey(bucket, key3, ONE, value, value.length());
+      Assert.fail("Write key should be failed");
     } catch (IOException ex) {
-      countException++;
       GenericTestUtils.assertExceptionContains("QUOTA_EXCEEDED", ex);
     }
 
-    // Write failed, bucket usedBytes should remain as 1
+    // Write failed, bucket usedBytes should remain as 2
```
linyiqun, line 968: Nit: usedBytes -> usedNamespace

linyiqun (commented):
Catch one nit below, +1 for others.
 @captainzmc, please  help do the double check for this PR, thanks.

captainzmc (commented):
@amaliujia Thanks for the patch. I left two comments, we can discuss if there are any questions.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCreateRequest.java
```diff
@@ -308,6 +309,8 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
           new CacheValue<>(Optional.of(omKeyInfo), trxnLogIndex));
 
       omBucketInfo.incrUsedBytes(preAllocatedSpace);
+      // Update namespace quota
+      omBucketInfo.incrUsedNamespace(1L);
```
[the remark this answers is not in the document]
amaliujia: I see.

Do you know where is the place that does such failed key cleanup?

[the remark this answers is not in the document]
amaliujia: @captainzmc  do you have an idea how does the background cleanup happen on keys that failed to commit?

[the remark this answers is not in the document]
captainzmc: There is a PR #1511 currently cleaning open key. I think we can use this. Update quota during cleanup.
But this PR is not merged yet, and we do not need to block here. We can open an issue and work on it after #1511 is merged.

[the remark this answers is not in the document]
amaliujia: Got it. Have filed a jira to track it: https://issues.apache.org/jira/browse/HDDS-4620

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
```diff
@@ -589,6 +589,25 @@ protected void checkBucketQuotaInBytes(OmBucketInfo omBucketInfo,
     }
   }
 
+  /**
+   * Check namespace quota.
+   */
+  protected void checkBucketQuotaInNamespace(OmBucketInfo omBucketInfo,
+      long allocatedNamespace) throws IOException {
+    if (omBucketInfo.getQuotaInNamespace() != OzoneConsts.QUOTA_RESET) {
```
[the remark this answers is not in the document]
amaliujia: Done!

amaliujia:
@captainzmc 

Comments have addressed and CI has passed.

captainzmc (approved):
+1 LGTM.  If no one else has a new comment, I will merge this tomorrow.

captainzmc:
Merged this. Thanks for @amaliujia's patch, and thanks for the review of @linyiqun

PR #1707, opened 2020-12-16
HDDS-4595. RDBScanner should print table name if it is not present in the db definition.

mukul1987:
Recently discovered that with branch HDDS-2939, the newly added directory table is missing from the DB definition array.
So the tool wasn't able to parse the table which really exists on the rocksdb.

https://issues.apache.org/jira/browse/HDDS-4595

## How was this patch tested?
Manual testing


bharatviswa504 (approved):

sadanand48:
+1

PR #1710, opened 2020-12-16
HDDS-4530. Add authorization check for SCM

wycccccc:

Add authorization check about stopReplicationManager, forceExitSafeMode, startReplicationManager, deactivatePipeline, closePipeline in `SCMClientProtocolServer` .

https://issues.apache.org/jira/browse/HDDS-4530

## How was this patch tested?

Add unit test.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
@@ -509,6 +509,8 @@ public class SCMClientProtocolServer implements
   @Override
   public void deactivatePipeline(HddsProtos.PipelineID pipelineID)
       throws IOException {
+    String remoteUser = getRemoteUserName();
+    getScm().checkAdminAccess(remoteUser);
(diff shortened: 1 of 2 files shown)


vivekratnavel:
@wycccccc Can you please check if this integration test failure is related to this change?

```
Error:  Failures: 
Error:    TestDecommissionAndMaintenance.testContainerIsReplicatedWhenAllNodesGotoMaintenance:394 expected:<5> but was:<6>
[INFO] 
Error:  Tests run: 213, Failures: 1, Errors: 0, Skipped: 15
```

wycccccc:
@vivekratnavel I passed this test, no failure like this.So I think the failure is not related to this change.

elek:
> I passed this test, no failure like this.So I think the failure is not related to this change.

We prefer to commit only green builds.  If it's an intermittent test, there are two options:

 1. Add an empty commit and have a green build
 2. Disable the test with a simple commit on master and create a JIRA with the log which proves the intermittency.

Let me re-trigger this build...

elek (approved):
+1, thanks the patch @wycccccc 

LGTM

Might be interesting to create unit (or acceptance tests?) for it in the future. But it can be a more complex task, can do it in follow-ups.

wycccccc:
Thax @elek review, I will try to create a jira to solve it.

PR #1716, opened 2020-12-16
HDDS-4606. Detect and handle duplicate ACLs when set default ACLs

masstter:
HDDS-4606. Got wrong ACL(s) when set default ACL(s) on volume for repeatedly

ignored duplicate entity when adding volume default ACL(s)

https://issues.apache.org/jira/browse/HDDS-4606

## How was this patch tested?

updating ACL(s) repeatedly, then using ozone CLI to check ACL(s), it's not repeating -- passed
image
image


xiaoyuyao (commented):
Thanks for reporting the issue. The fix only handle one case for adding exactly same ACL. We still need to handle adding different acl rights to existing acl entry. This can be done similar to the way access acl is handled via BitSet.

masstter:
@xiaoyuyao Thanks for reviews, it's helped me a lot, added handle adding different acl rights to existing acl entry in my new changes

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmOzoneAclMap.java
```diff
@@ -100,28 +97,58 @@ private BitSet getAcl(OzoneAclType type, String user) {
...
+          defaultAclList.add(ozoneAclInfo);
+          return;
+        }
       }
-      getAccessAclMap(aclType).replace(acl.getName(), temp);
     }
+    defaultAclList.add(ozoneAclInfo);
+  }
+
+  private void aclExistsError(OzoneAcl acl) throws OMException {
```
ChenSammi, line 137: Could you replace other same expection codes in this class using this aclExistsError function?

ChenSammi (commented):
That patch overall looks good, just a few minor comments.

ChenSammi:
+1.  Thanks @masstter for the contribution and @xiaoyuyao for the code review.

PR #1719, opened 2020-12-16
HDDS-4597. OM Terminates when adding acls to S3v volume.

bharatviswa504:

When OM starts it creates default volume "s3v" where S3 buckets are created with all required permissions.

When volume is created it is set the updateID with maxTransactionID. This causes a problem, when user tries to perform setAcl/AddAcl on volume, we update the updateID during that we have below check. 

Now let's say, a transaction index 100 is trying to set ACL, it will fail in the preCondition check.

```
    if (isRatisEnabled) {
      Preconditions.checkArgument(updateId >= this.updateID, String.format(
          "Trying to set updateID to %d which is not greater than the " +
              "current value of %d for %s", updateId, this.updateID,
          getObjectInfo()));
    }
```

https://issues.apache.org/jira/browse/HDDS-4597

## How was this patch tested?

Added test.


vivekratnavel (approved):
+1 LGTM

jojochuang (approved):

bharatviswa504:
Thank You @vivekratnavel and @jojochuang for the review.

PR #1720, opened 2020-12-16
HDDS-4175. Implement Datanode Finalization.

prashantpogde:

Implement Datanode Finalization

https://issues.apache.org/jira/browse/HDDS-4175

## How was this patch tested?

Build/UT. I will address CI failures if any.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutFeatureCatalog.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutFeatureCatalog.java
@@ -19,6 +19,7 @@
 package org.apache.hadoop.hdds.upgrade;
 
 import java.util.Optional;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import org.apache.hadoop.ozone.upgrade.LayoutFeature;
 
@@ -52,11 +53,13 @@ public class HDDSLayoutFeatureCatalog {
       this.description = description;
     }
 
+    @SuppressFBWarnings("ME_ENUM_FIELD_SETTER")
     public void setSCMUpgradeAction(Optional<? extends HDDSUpgradeAction>
                                         scmAction) {
       this.scmUpgradeAction = scmAction;
     }
 
+    @SuppressFBWarnings("ME_ENUM_FIELD_SETTER")
     public void setDataNodeUpgradeAction(Optional<? extends HDDSUpgradeAction>
                                              datanodeAction) {
       this.datanodeUpgradeAction = datanodeAction;
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutFeatureCatalog.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutFeatureCatalog.java
@@ -52,12 +52,13 @@ public class HDDSLayoutFeatureCatalog {
       this.description = description;
     }
 
-    HDDSLayoutFeature(final int layoutVersion, String description,
-                      Optional<? extends HDDSUpgradeAction> scmAction,
-                      Optional<? extends HDDSUpgradeAction> datanodeAction) {
-      this.layoutVersion = layoutVersion;
-      this.description = description;
+    public void setSCMUpgradeAction(Optional<? extends HDDSUpgradeAction>
+                                        scmAction) {
       this.scmUpgradeAction = scmAction;
+    }
+
+    public void setDataNodeUpgradeAction(Optional<? extends HDDSUpgradeAction>
+                                             datanodeAction) {
       this.datanodeUpgradeAction = datanodeAction;
     }
 
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutFeatureCatalog.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutFeatureCatalog.java
@@ -27,6 +27,7 @@ import org.apache.hadoop.ozone.upgrade.LayoutFeature;
  */
 public class HDDSLayoutFeatureCatalog {
 
+
   /**
    * List of HDDS Features.
    */
@@ -39,7 +40,11 @@ public class HDDSLayoutFeatureCatalog {
 
     private int layoutVersion;
     private String description;
-    private Optional< ? extends HDDSUpgradeAction> hddsUpgradeAction =
+
+    private Optional<? extends HDDSUpgradeAction> scmUpgradeAction =
+        Optional.empty();
+
+    private Optional<? extends HDDSUpgradeAction> datanodeUpgradeAction =
         Optional.empty();
 
     HDDSLayoutFeature(final int layoutVersion, String description) {
@@ -48,10 +53,12 @@ public class HDDSLayoutFeatureCatalog {
     }
 
     HDDSLayoutFeature(final int layoutVersion, String description,
-                    HDDSUpgradeAction upgradeAction) {
+                      Optional<? extends HDDSUpgradeAction> scmAction,
+                      Optional<? extends HDDSUpgradeAction> datanodeAction) {
       this.layoutVersion = layoutVersion;
       this.description = description;
-      hddsUpgradeAction = Optional.of(upgradeAction);
+      this.scmUpgradeAction = scmAction;
+      this.datanodeUpgradeAction = datanodeAction;
     }
 
     @Override
@@ -64,9 +71,12 @@ public class HDDSLayoutFeatureCatalog {
       return description;
     }
 
-    @Override
-    public Optional<? extends HDDSUpgradeAction> onFinalizeAction() {
-      return hddsUpgradeAction;
+    public Optional<? extends HDDSUpgradeAction> onFinalizeSCMAction() {
+      return scmUpgradeAction;
+    }
+
+    public Optional<? extends HDDSUpgradeAction> onFinalizeDataNodeAction() {
+      return datanodeUpgradeAction;
     }
   }
 }
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
@@ -42,6 +42,7 @@ public final class OzoneConsts {
 
   public static final String STORAGE_ID = "storageID";
   public static final String DATANODE_UUID = "datanodeUuid";
+  public static final String DATANODE_STORAGE_DIR = "datanodeStorageConfig";
   public static final String CLUSTER_ID = "clusterID";
   public static final String LAYOUTVERSION = "layOutVersion";
   public static final String CTIME = "ctime";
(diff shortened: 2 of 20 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/DataNodeStorageConfig.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.NodeType;
+import org.apache.hadoop.hdds.server.ServerUtils;
+import org.apache.hadoop.ozone.common.Storage;
+
+/**
+ * DataNodeStorageConfig is responsible for management of the
+ * StorageDirectories used by the DataNode.
+ */
+public class DataNodeStorageConfig extends Storage {
```
avijayanhwx, line 37: The StorageInfo class generates a new "clusterId" for every Version file that it writes down. There is no clusterId with respect to DN now. Can we confirm the implications of adding a clusterId for the DN persistent state even when we need only the layout version?

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
```diff
@@ -145,7 +145,7 @@ protected StorageInfo getStorageInfo() {
   abstract protected Properties getNodeProperties();
 
   /**
-   * Sets the Node properties specific to OM/SCM.
+   * Sets the Node properties specific to OM/SCM/DataNode.
```
avijayanhwx, line 148: Can we add a unit test for the Datanode layout version startup validation like in TestScmStartupSlvLessThanMlv, TestOmStartupSlvLessThanMlv?

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
```diff
@@ -71,6 +69,11 @@ public synchronized StatusAndMessages preFinalize(String upgradeClientID,
       return FINALIZATION_IN_PROGRESS_MSG;
     case FINALIZATION_DONE:
     case ALREADY_FINALIZED:
+      if (versionManager.needsFinalization()) {
```
avijayanhwx, line 72: Can you please explain how this can happen, and how can a user recover from this?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DataNodeUpgradeFinalizer.java
```diff
@@ -0,0 +1,99 @@
...
+        emitStartingMsg();
+        versionManager.setUpgradeState(FINALIZATION_IN_PROGRESS);
+        /*
+         * Before we can call finalize the feature, we need to make sure that
+         * all existing pipelines are closed and pipeline Manger would freeze
+         * all new pipeline creation.
+         */
+
+        for (HDDSLayoutFeature f : versionManager.unfinalizedFeatures()) {
+          finalizeFeature(f, datanodeStateMachine.getDataNodeStorageConfig());
```
avijayanhwx, line 84: We currently have a common layout feature hierarchy for SCM and Datanode. It is likely that a feature can either be part of the SCM or the DN. If we call on the onFinalize action in both the SCM and DN for every unfinalized HDDS layout feature, then we may be left with unintended consequences of trying to call on finalize action on an SCM feature on the DN. We should either have a marker on every layout feature to say if it belongs to SCM or DN, or not support finalization actions for DN for the V1 upgrade implementation.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DataNodeLayoutVersionManager.java
```diff
@@ -69,41 +58,15 @@ public static synchronized LayoutVersionManager getInstance() {
...
-        if (props.isEmpty()) {
-          continue;
-        }
-        int storedVersion = HddsVolumeUtil.getLayOutVersion(props, versionFile);
-        if (storedVersion > layoutVersion) {
-          layoutVersion = storedVersion;
-        }
-      }
-      dataNodeLayoutVersionManager.init(layoutVersion,
+      dataNodeLayoutVersionManager.init(dataNodeStorage.getLayoutVersion(),
```
avijayanhwx, line 69: With these changes, I believe that _HDDSLayoutFeature#FIRST_UPGRADE_VERSION_ can be removed.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
```diff
@@ -42,6 +42,7 @@
 
   public static final String STORAGE_ID = "storageID";
   public static final String DATANODE_UUID = "datanodeUuid";
+  public static final String DATANODE_STORAGE_DIR = "datanodeStorageConfig";
```
avijayanhwx, line 45: There is also a "datanode.id" file being created in the Datanode that looks like a metadata yaml file. Can we look into whether we can add our layout version into that file? Reference -> _org.apache.hadoop.hdds.utils.HddsServerUtil#getDatanodeIdFilePath_

avijayanhwx (commented):
Thanks for working on this @prashantpogde. I have left a few comments inline.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/DataNodeStorageConfig.java
prashantpogde: It will remain a "don't care bit". storage config is used only for layout information.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
prashantpogde: it cant happen. just trying to catch if any future modification messes this up.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DataNodeLayoutVersionManager.java
prashantpogde: As discussed in the last call. Let us keep it as it can be used to test the upgrade feature itself when we role out this feature. It would be a self validating upgrade.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
prashantpogde: yup, storage config is used just for the upgrade and nothing else. We will continue to use datanode.id file as it is.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
prashantpogde: yes

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DataNodeUpgradeFinalizer.java
prashantpogde: yes. I added separate SCM and datanode actions. Please take a look.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DataNodeUpgradeAction.java
```diff
@@ -0,0 +1,31 @@
...
+
+package org.apache.hadoop.ozone.container.upgrade;
+
+import org.apache.hadoop.hdds.upgrade.HDDSUpgradeAction;
+import org.apache.hadoop.ozone.container.common.statemachine.DatanodeStateMachine;
+
+/**
+ * Upgrade Action for DataNode which takes in a 'DataNodeStateMachine' instance.
+ */
+public class DataNodeUpgradeAction extends
```
avijayanhwx, line 27: Can we have DataNodeUpgradeAction and SCMUpgradeAction as an interface or abstract class?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutFeatureCatalog.java
```diff
@@ -39,7 +40,11 @@
 
     private int layoutVersion;
     private String description;
-    private Optional< ? extends HDDSUpgradeAction> hddsUpgradeAction =
+
+    private Optional<? extends HDDSUpgradeAction> scmUpgradeAction =
+        Optional.empty();
+
+    private Optional<? extends HDDSUpgradeAction> datanodeUpgradeAction =
```
avijayanhwx, line 47: I am wondering whether the organization of modules will cause a problem here. The DatanodeStateMachine class is in container-service module, while this class (HDDSLayoutFeatureCatalog) is in hdds-common. The common module is usually shared across the the more specific modules. Where does one create an upgrade action for the Datanode? If the action is in container-service, then it cannot be accessed in hdds-common, if the action is in common, then it cannot access the DatanodeStateMachine class.

avijayanhwx (commented):
Thanks for the updated patch @prashantpogde. A couple of comments inline.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DataNodeUpgradeAction.java
prashantpogde: Yes, will do

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutFeatureCatalog.java
prashantpogde: Yes. I have updated the PR to add datanode action and scm action. There is an integration test for this, thats working but thats in a separate PR for integration test.

avijayanhwx: New approach looks good to me. Can we fix the CI issues? Also, do we need to include the sample SCM and DN action that you have added in the latest commit? Maybe we could use a different implementation that accepts an empty list (enum does not).

prashantpogde: Yes, uploaded the new changes to address CI failures. We can leave sample SCM and DN actions to be used as an example in future upgrades. This also makes it a self validating release.

prashantpogde:
these CI failures are unrelated.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutFeatureCatalog.java
avijayanhwx: @prashantpogde I am ok with that. If we are going to add these No-Op actions for the first upgrade release, I would suggest renaming the action to something more intuitive rather than "DatanodeAction1" and "SCMAction1". I am OK with doing these changes in a follow up JIRA.

avijayanhwx (approved):
Failures are unrelated.
LGTM +1 

A couple of suggested follow up items.

- Renaming actions (mentioned comment)
- Removing the HddsLayoutFeature and Action "Catalog" classes, and retaining the enum as the top level entity (similar to org.apache.hadoop.ozone.om.upgrade.OMLayoutFeature)

PR #1721, opened 2020-12-17
HDDS-4567: SCMContext Phase 1 - Raft Related Info

GlenGeng:

We want to provide SCMContext, which would be a single source of truth for some key information that is shared across all components within SCM.

SCMContext holds two kind of key information:
1. RaftServer related info: isLeader, term.
2. SafeMode related info: inSafeMode, preCheckComplete.

For details, please refer to
https://docs.google.com/document/d/1h_3gpC4o2EpuBlcQiJC7MMoZz9JmaMX9CxObSxWU614/edit#heading=h.4f1y2k1ojlx8

https://issues.apache.org/jira/browse/HDDS-4567

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/CloseContainerEventHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/CloseContainerEventHandler.java
@@ -23,6 +23,7 @@ import java.util.stream.Collectors;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.LifeCycleEvent;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.LifeCycleState;
+import org.apache.hadoop.hdds.scm.ha.SCMContext;
 import org.apache.hadoop.hdds.scm.pipeline.PipelineManager;
 import org.apache.hadoop.hdds.scm.pipeline.PipelineNotFoundException;
 import org.apache.hadoop.hdds.server.events.EventHandler;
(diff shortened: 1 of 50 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/CloseContainerEventHandler.java
```diff
@@ -74,14 +79,13 @@ public void onMessage(ContainerID containerID, EventPublisher publisher) {
           .getContainer(containerID);
       // Send close command to datanodes, if the container is in CLOSING state
       if (container.getState() == LifeCycleState.CLOSING) {
+        SCMCommand<?> command = new CloseContainerCommand(
+            containerID.getId(), container.getPipelineID());
+        command.setTerm(scmContext.getTerm());
```
amaliujia, line 84: Question:
when you do decide to throw the NotLeaderException out rather than handling immediately after `scmContext.getTerm()`?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * Used by non-HA mode SCM, Recon and Unit Tests.
+   */
+  public static SCMContext emptyContext() {
+    return EMPTY_CONTEXT;
+  }
+
+  private boolean isLeader;
+  private OptionalLong termOpt;
+  private final StorageContainerManager scm;
```
amaliujia, line 58: WIll it better to only hold reference to SCMHAManager than scm?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/CloseContainerEventHandler.java
GlenGeng: `NotLeaderException` is a subclass of `IOException` which is handled in this `onMessage()` method.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
GlenGeng: Good point! In SCMContext phase 2, it will also handle safe mode related things, and in future, SCMContext will be a mediator of all the XxxManagers in SCM. That's why I choose holding `scm` instead of holding `SCMHAManager`. I will revisit this code in my next PR: SCMContext phase 2.

PR #1723, opened 2020-12-18
HDDS-4588. If volume's quota is enabled then bucket's quota cannot be cleared

captainzmc:

If volume's quota is enabled then bucket's quota cannot be cleared. We need to prompt the user to clear volume quota first.
[For more details.](https://github.com/apache/ozone/pull/1677#issuecomment-745026394)

To prevent conflicts, I have updated the usage and behavior of clear space quota in #1677.

https://issues.apache.org/jira/browse/HDDS-4588

## How was this patch tested?

ut added

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -318,10 +318,6 @@ public abstract class TestOzoneRpcClientAbstract {
         " spaceQuota because volume spaceQuota is not cleared.",
         () -> ozoneBucket.clearSpaceQuota());
 
-    LambdaTestUtils.intercept(IOException.class, "Can not clear bucket" +
-        " namespaceQuota because volume namespaceQuota is not cleared.",
-        () -> ozoneBucket.clearCountQuota());
-
     store.getVolume(volumeName).clearSpaceQuota();
     store.getVolume(volumeName).clearNamespaceQuota();
     OzoneVolume clrVolume = store.getVolume(volumeName);
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -327,7 +327,7 @@ public abstract class TestOzoneRpcClientAbstract {
       Assert.fail("clear quota should be failed");
     } catch (IOException ex) {
       GenericTestUtils.assertExceptionContains("Before clear bucket " +
-          "quotaInNameSpace, volume quotaInNameSpace need to be clear first.",
+          "quotaInNamespace, volume quotaInNamespace need to be clear first.",
           ex);
     }
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -280,7 +280,7 @@ public abstract class TestOzoneRpcClientAbstract {
   }
 
   @Test
-  public void testSetAndClrQuota() throws IOException {
+  public void testSetAndClrQuota() throws Exception {
     String volumeName = UUID.randomUUID().toString();
     String bucketName = UUID.randomUUID().toString();
     OzoneVolume volume = null;
@@ -314,22 +314,13 @@ public abstract class TestOzoneRpcClientAbstract {
         ozoneBucket.getQuotaInBytes());
     Assert.assertEquals(1000L, ozoneBucket.getQuotaInNamespace());
 
-    try {
-      ozoneBucket.clearSpaceQuota();
-      Assert.fail("clear quota should be failed");
-    } catch (IOException ex) {
-      GenericTestUtils.assertExceptionContains("Before clear bucket " +
-          "quotaInBytes, volume quotaInBytes need to be clear first.", ex);
-    }
+    LambdaTestUtils.intercept(IOException.class, "Can not clear bucket" +
+        " spaceQuota because volume spaceQuota is not cleared.",
+        () -> ozoneBucket.clearSpaceQuota());
 
-    try {
-      ozoneBucket.clearCountQuota();
-      Assert.fail("clear quota should be failed");
-    } catch (IOException ex) {
-      GenericTestUtils.assertExceptionContains("Before clear bucket " +
-          "quotaInNamespace, volume quotaInNamespace need to be clear first.",
-          ex);
-    }
+    LambdaTestUtils.intercept(IOException.class, "Can not clear bucket" +
+        " namespaceQuota because volume namespaceQuota is not cleared.",
+        () -> ozoneBucket.clearCountQuota());
 
     store.getVolume(volumeName).clearSpaceQuota();
     store.getVolume(volumeName).clearNamespaceQuota();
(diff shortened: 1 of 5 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
```diff
@@ -228,6 +228,13 @@ public boolean checkQuotaBytesValid(OMMetadataManager metadataManager,
       throws IOException {
     long quotaInBytes = omBucketArgs.getQuotaInBytes();
 
+    if (quotaInBytes == OzoneConsts.QUOTA_RESET &&
+        omVolumeArgs.getQuotaInBytes() != OzoneConsts.QUOTA_RESET) {
+      throw new OMException("Before clear bucket quotaInBytes," +
```
amaliujia, line 233: Nit: will it better to use more descriptive names than variable names. For example, replace `quotaInBytes` by space quota?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
```diff
@@ -259,9 +266,16 @@ public boolean checkQuotaBytesValid(OMMetadataManager metadataManager,
...
 
   public boolean checkQuotaNamespaceValid(OmVolumeArgs omVolumeArgs,
-      OmBucketArgs omBucketArgs) {
+      OmBucketArgs omBucketArgs) throws IOException{
     long quotaInNamespace = omBucketArgs.getQuotaInNamespace();
 
+    if (quotaInNamespace == OzoneConsts.QUOTA_RESET &&
+        omVolumeArgs.getQuotaInNamespace() != OzoneConsts.QUOTA_RESET) {
+      throw new OMException("Before clear bucket " +
+          "quotaInNamespace, volume quotaInNamespace need to be clear first.",
```
amaliujia, line 275: Nit: same, maybe replace `quotaInNamespace` by namespace quota?

linyiqun (commented):
I am thinking for this case, it's okay to just clear bucket quota no matter if the volume quota was set? Since the volume quota check will still make the sense, when we clear one bucket quota.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
linyiqun: Agreed with @amaliujia 's suggestion. I also prefer to do below rename changing for more readable:

quotaInNamespace -> namespaceQuota
quotaInNamespace -> spaceQuota

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -311,6 +311,23 @@ public void testSetAndClrQuota() throws IOException {
         ozoneBucket.getQuotaInBytes());
     Assert.assertEquals(1000L, ozoneBucket.getQuotaInNamespace());
 
+    try {
+      ozoneBucket.clearSpaceQuota();
+      Assert.fail("clear quota should be failed");
+    } catch (IOException ex) {
+      GenericTestUtils.assertExceptionContains("Before clear bucket " +
+          "quotaInBytes, volume quotaInBytes need to be clear first.", ex);
+    }
```
ayushtkn, line 320: Can try something like this -
```
  LambdaTestUtils.intercept(IOException.class, "Before clear bucket "
            + "quotaInBytes, volume quotaInBytes need to be clear first.",
        () -> ozoneBucket.clearSpaceQuota());
```
Shall save lines and looks better

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
```diff
@@ -311,6 +311,23 @@ public void testSetAndClrQuota() throws IOException {
...
+
+    try {
+      ozoneBucket.clearCountQuota();
+      Assert.fail("clear quota should be failed");
+    } catch (IOException ex) {
+      GenericTestUtils.assertExceptionContains("Before clear bucket " +
+          "quotaInNamespace, volume quotaInNamespace need to be clear first.",
+          ex);
+    }
```
ayushtkn, line 329: Similarly here -
```
    LambdaTestUtils.intercept(IOException.class, "Before clear bucket "
            + "quotaInNamespace, volume quotaInNamespace need to be clear first.",
        () -> ozoneBucket.clearCountQuota());
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
ayushtkn, line 275: Seems some grammatical  error in the message, Can you check and reframe if possible, 
May be something like a warning, Can not clear bucket namespace quota because volume namespace quota is not unset. Or may be something better....

ayushtkn (commented):
Thanx @captainzmc for the work here.
dropped some minor comments,

captainzmc:
Fixed review issues.

captainzmc:
> I am thinking for this case, it's okay to just clear bucket quota no matter if the volume quota was set? Since the volume quota check will still make the sense, when we clear one bucket quota.

There is a very [detailed discussion](https://github.com/apache/ozone/pull/1677#issuecomment-745026394) of this.

captainzmc:
Fixed review issues.

linyiqun:
> > I am thinking for this case, it's okay to just clear bucket quota no matter if the volume quota was set? Since the volume quota check will still make the sense, when we clear one bucket quota.
> 
> There is a very [detailed discussion](https://github.com/apache/ozone/pull/1677#issuecomment-745026394) of this.

@captainzmc , although we removed the usedBytes because of the quota update issue. But instead of, we sum the buckets used bytes when bucket is created. This logic can still make the sense, [OMBucketCreateRequest.java#L331](https://github.com/apache/ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketCreateRequest.java#L331)

captainzmc:
> @captainzmc , although we removed the volume usedBytes because of the quota update issue. Can we also try to sum the buckets used bytes when bucket is created that will be used for the volume quota check? As I see we already do the similar check for the quota valid check, [OMBucketCreateRequest.java#L331](https://github.com/apache/ozone/blob/master/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketCreateRequest.java#L331)

Thanks for @linyiqun‘ advice.
We've considered this option before, but it can cause performance problems.
Let's assume that we have 1000 buckets under volume.
First set quota, which is not a very frequent action in a cluster, and unlike the write operation, has a low latency requirement. So we calculated that all bucket quota is acceptable.
However, each write operation calculating the usedBytes of all buckets for volume spaceQuotaChecker may affect write performance. The more buckets there are, the more performance is affected per write. This maybe unacceptable to the users.
And if we do this, we need to lock the volume when we count all the usedBytes of our buckets, because we don't want the usedBytes of our buckets to change back and forth when we calculate the total usedBytes, which will also affect the performance.

linyiqun:
If there is the performance consideration for this, I am okay for this.

linyiqun (commented):
+1 for this PR change.

ayushtkn (approved):
LGTM

captainzmc:
Removed the namespace restriction. cc @amaliujia

amaliujia:
LGTM

captainzmc:
Merged this patch, Thanks for @linyiqun @amaliujia @ayushtkn for the review.

PR #1718, opened 2020-12-16
HDDS-4608. Add a unit test to identify missing/extra tables in OMDBDefinition.

mukul1987:
## What changes were proposed in this pull request?
Add a unit test to identify missing/extra tables in OMDBDefinition

One example of such a case is the s3Table

## What is the link to the Apache JIRA
https://issues.apache.org/jira/browse/HDDS-4608

## How was this patch tested?
Added a new unit test

PR #1724, opened 2020-12-19
HDDS-4596. Directory table, fileTable and openFile Table is missing from the OM DB Definition.

mukul1987:

Directory table, fileTable and openFile Table is missing from the OM DB Definition, Hence the DBParser tools doesn't work against it.

https://issues.apache.org/jira/browse/HDDS-4596

## How was this patch tested?
A new unit test is added in master via #1718 , HDDS-4608


bharatviswa504 (approved):
+1 LGTM

bharatviswa504:
Thank You @mukul1987 for the contribution.

PR #1725, opened 2020-12-20
HDDS-3208. Implement Ratis snapshot on SCM

amaliujia:

Design doc: https://docs.google.com/document/d/1uy4_ER2V6nNQJ7_5455Wz8NmI142JHPnif6Y1OdPi8E/edit?usp=sharing

https://issues.apache.org/jira/browse/HDDS-3208

## How was this patch tested?

UT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
@@ -32,6 +32,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
@@ -110,6 +111,7 @@ public class ContainerManagerImpl implements ContainerManagerV2 {
         .setPipelineManager(pipelineManager)
         .setRatisServer(scmHaManager.getRatisServer())
         .setContainerStore(containerStore)
+        .setSCMDBTransactionBuffer(scmHaManager.getDBTransactionBuffer())
         .build();
 
     this.numContainerPerVolume = conf
@@ -117,8 +119,6 @@ public class ContainerManagerImpl implements ContainerManagerV2 {
             ScmConfigKeys.OZONE_SCM_PIPELINE_OWNER_CONTAINER_COUNT_DEFAULT);
 
     this.scmContainerManagerMetrics = SCMContainerManagerMetrics.create();
-
-
   }
 
   @Override
@@ -396,5 +396,10 @@ public class ContainerManagerImpl implements ContainerManagerV2 {
   @Deprecated
   protected ContainerStateManagerV2 getContainerStateManager() {
     return containerStateManager;
+  }
+
+  @VisibleForTesting
+  public SCMHAManager getSCMHAManager() {
+    return haManager;
   }
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerImpl.java
@@ -41,6 +41,7 @@ import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.container.states.ContainerState;
 import org.apache.hadoop.hdds.scm.container.states.ContainerStateMap;
 import org.apache.hadoop.hdds.scm.ha.CheckedConsumer;
+import org.apache.hadoop.hdds.scm.ha.DBTransactionBuffer;
 import org.apache.hadoop.hdds.scm.ha.ExecutionUtil;
 import org.apache.hadoop.hdds.scm.ha.SCMHAInvocationHandler;
 import org.apache.hadoop.hdds.scm.ha.SCMRatisServer;
@@ -99,7 +100,9 @@ public final class ContainerStateManagerImpl
   /**
    * Persistent store for Container States.
    */
-  private final Table<ContainerID, ContainerInfo> containerStore;
+  private Table<ContainerID, ContainerInfo> containerStore;
+
+  private final DBTransactionBuffer transactionBuffer;
 
   /**
    * PipelineManager instance.
@@ -135,7 +138,8 @@ public final class ContainerStateManagerImpl
    */
   private ContainerStateManagerImpl(final Configuration conf,
       final PipelineManager pipelineManager,
-      final Table<ContainerID, ContainerInfo> containerStore)
+      final Table<ContainerID, ContainerInfo> containerStore,
+      final DBTransactionBuffer buffer)
       throws IOException {
     this.pipelineManager = pipelineManager;
     this.containerStore = containerStore;
@@ -144,6 +148,7 @@ public final class ContainerStateManagerImpl
     this.containers = new ContainerStateMap();
     this.lastUsedMap = new ConcurrentHashMap<>();
     this.containerStateChangeActions = getContainerStateChangeActions();
+    this.transactionBuffer = buffer;
 
     initialize();
   }
@@ -298,12 +303,16 @@ public final class ContainerStateManagerImpl
     try {
       if (!containers.contains(containerID)) {
         ExecutionUtil.create(() -> {
-          containerStore.put(containerID, container);
+          containerStore.putWithBatch(
+              transactionBuffer.getCurrentBatchOperation(),
+              containerID, container);
           containers.addContainer(container);
           pipelineManager.addContainerToPipeline(pipelineID, containerID);
         }).onException(() -> {
           containers.removeContainer(containerID);
-          containerStore.delete(containerID);
+          containerStore.deleteWithBatch(
+              transactionBuffer.getCurrentBatchOperation(),
+              containerID);
         }).execute();
       }
     } finally {
@@ -339,13 +348,17 @@ public final class ContainerStateManagerImpl
         if (newState.getNumber() > oldState.getNumber()) {
           ExecutionUtil.create(() -> {
             containers.updateState(id, oldState, newState);
-            containerStore.put(id, containers.getContainerInfo(id));
+            containerStore.putWithBatch(
+                transactionBuffer.getCurrentBatchOperation(),
+                id, containers.getContainerInfo(id));
           }).onException(() -> {
-            containerStore.put(id, oldInfo);
+            containerStore.putWithBatch(
+                transactionBuffer.getCurrentBatchOperation(),
+                id, oldInfo);
             containers.updateState(id, newState, oldState);
           }).execute();
-          containerStateChangeActions.getOrDefault(event, info -> {})
-              .execute(oldInfo);
+          containerStateChangeActions.getOrDefault(event, info -> {
+          }).execute(oldInfo);
         }
       }
     } finally {
@@ -475,7 +488,9 @@ public final class ContainerStateManagerImpl
       final ContainerID cid = ContainerID.getFromProtobuf(id);
       final ContainerInfo containerInfo = containers.getContainerInfo(cid);
       ExecutionUtil.create(() -> {
-        containerStore.delete(cid);
+        containerStore.deleteWithBatch(
+            transactionBuffer.getCurrentBatchOperation(),
+            cid);
         containers.removeContainer(cid);
       }).onException(() -> containerStore.put(cid, containerInfo)).execute();
     } finally {
@@ -504,7 +519,12 @@ public final class ContainerStateManagerImpl
     private PipelineManager pipelineMgr;
     private SCMRatisServer scmRatisServer;
     private Table<ContainerID, ContainerInfo> table;
+    private DBTransactionBuffer transactionBuffer;
 
+    public Builder setSCMDBTransactionBuffer(DBTransactionBuffer buffer) {
+      this.transactionBuffer = buffer;
+      return this;
+    }
     public Builder setConfiguration(final Configuration config) {
       conf = config;
       return this;
@@ -533,7 +553,7 @@ public final class ContainerStateManagerImpl
       Preconditions.checkNotNull(table);
 
       final ContainerStateManagerV2 csm = new ContainerStateManagerImpl(
-          conf, pipelineMgr, table);
+          conf, pipelineMgr, table, transactionBuffer);
 
       final SCMHAInvocationHandler invocationHandler =
           new SCMHAInvocationHandler(RequestType.CONTAINER, csm,
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership.  The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.hadoop.hdds.scm.ha;
+
+import org.apache.hadoop.hdds.scm.metadata.SCMMetadataStore;
+import org.apache.hadoop.hdds.utils.db.BatchOperation;
+import org.apache.hadoop.hdds.utils.db.Table;
+import org.apache.ratis.statemachine.SnapshotInfo;
+
+import java.io.IOException;
+
+import static org.apache.hadoop.ozone.OzoneConsts.TRANSACTION_INFO_KEY;
+
+/**
+ * This is a transaction buffer that buffers SCM DB operations for Pipeline and
+ * Container. When flush this buffer to DB, a transaction info will also be
+ * written into DB to indicate the term and transaction index for the latest
+ * operation in DB.
+ */
+public class SCMDBTransactionBuffer implements DBTransactionBuffer {
+  private final SCMMetadataStore metadataStore;
+  private BatchOperation currentBatchOperation;
+  private SCMTransactionInfo latestTrxInfo;
+  private SnapshotInfo latestSnapshot;
+
+  public SCMDBTransactionBuffer(SCMMetadataStore store) throws IOException {
+    this.metadataStore = store;
+
+    // initialize a batch operation during construction time
+    currentBatchOperation = this.metadataStore.getStore().initBatchOperation();
+    latestTrxInfo = store.getTransactionInfoTable().get(TRANSACTION_INFO_KEY);
+    if (latestTrxInfo == null) {
+      // transaction table is empty
+      latestTrxInfo =
+          SCMTransactionInfo
+              .builder()
+              .setTransactionIndex(-1)
+              .setCurrentTerm(0)
+              .build();
+    }
+    latestSnapshot = latestTrxInfo.toSnapshotInfo();
+  }
+
+  @Override
+  public BatchOperation getCurrentBatchOperation() {
+    return currentBatchOperation;
+  }
+
+  @Override
+  public void updateLatestTrxInfo(SCMTransactionInfo info) {
+    if (info.compareTo(this.latestTrxInfo) <= 0) {
+      throw new IllegalArgumentException(
+          "Updating DB buffer transaction info by an older transaction info, "
+          + "current: " + this.latestTrxInfo + ", updating to: " + info);
+    }
+    this.latestTrxInfo = info;
+  }
+
+  @Override
+  public SCMTransactionInfo getLatestTrxInfo() {
+    return this.latestTrxInfo;
+  }
+
+  @Override
+  public SnapshotInfo getLatestSnapshot() {
+    return latestSnapshot;
+  }
+
+  @Override
+  public void setLatestSnapshot(SnapshotInfo latestSnapshot) {
+    this.latestSnapshot = latestSnapshot;
+  }
+
+  @Override
+  public void flush() throws IOException {
+    // write latest trx info into trx table in the same batch
+    Table<String, SCMTransactionInfo> transactionInfoTable
+        = metadataStore.getTransactionInfoTable();
+    transactionInfoTable.putWithBatch(currentBatchOperation,
+        TRANSACTION_INFO_KEY, latestTrxInfo);
+
+    metadataStore.getStore().commitBatchOperation(currentBatchOperation);
+    currentBatchOperation.close();
+    this.latestSnapshot = latestTrxInfo.toSnapshotInfo();
+    // reset batch operation
+    currentBatchOperation = metadataStore.getStore().initBatchOperation();
+  }
+
+  @Override
+  public String toString() {
+    return latestTrxInfo.toString();
+  }
+
+  @Override
+  public void close() throws IOException {
+
+  }
+}
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisSnapshotInfo.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisSnapshotInfo.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership. The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * <p>http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * <p>Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.hadoop.hdds.scm.ha;
+
+import org.apache.ratis.server.protocol.TermIndex;
+import org.apache.ratis.server.storage.FileInfo;
+import org.apache.ratis.statemachine.SnapshotInfo;
+
+import java.util.List;
+
+import static org.apache.hadoop.ozone.OzoneConsts.TRANSACTION_INFO_SPLIT_KEY;
+
+/**
+ * This class captures the snapshotIndex and term of the latest snapshot in
+ * the SCM
+ * Ratis server loads the snapshotInfo during startup and updates the
+ * lastApplied index to this snapshotIndex. SCM SnapshotInfo does not contain
+ * any files. It is used only to store/ update the last applied index and term.
+ */
+public class SCMRatisSnapshotInfo implements SnapshotInfo {
+  private final long term;
+  private final long snapshotIndex;
+
+  public SCMRatisSnapshotInfo(long term, long index) {
+    this.term = term;
+    this.snapshotIndex = index;
+  }
+
+  @Override
+  public TermIndex getTermIndex() {
+    return TermIndex.valueOf(term, snapshotIndex);
+  }
+
+  @Override
+  public long getTerm() {
+    return term;
+  }
+
+  @Override
+  public long getIndex() {
+    return snapshotIndex;
+  }
+
+  @Override
+  public List<FileInfo> getFiles() {
+    return null;
+  }
+
+  @Override
+  public String toString() {
+    StringBuilder stringBuilder = new StringBuilder();
+    stringBuilder.append(term);
+    stringBuilder.append(TRANSACTION_INFO_SPLIT_KEY);
+    stringBuilder.append(snapshotIndex);
+    return stringBuilder.toString();
+  }
+}
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
@@ -33,7 +33,6 @@ import org.apache.ratis.protocol.RaftGroupMemberId;
 import org.apache.ratis.protocol.RaftPeerId;
 import org.apache.hadoop.hdds.scm.exceptions.SCMException;
 import org.apache.hadoop.util.Time;
-import org.apache.ratis.protocol.Message;
 import org.apache.ratis.server.protocol.TermIndex;
 import org.apache.ratis.statemachine.SnapshotInfo;
 import org.apache.ratis.statemachine.TransactionContext;
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
@@ -57,12 +57,12 @@ public class SCMStateMachine extends BaseStateMachine {
   private final DBTransactionBuffer transactionBuffer;
 
   public SCMStateMachine(final StorageContainerManager scm,
-      final SCMRatisServer ratisServer)
+      final SCMRatisServer ratisServer, DBTransactionBuffer buffer)
       throws SCMException {
     this.scm = scm;
     this.ratisServer = ratisServer;
     this.handlers = new EnumMap<>(RequestType.class);
-    this.transactionBuffer = scm.getScmHAManager().getDBTransactionBuffer();
+    this.transactionBuffer = buffer;
     SCMTransactionInfo latestTrxInfo =
         this.transactionBuffer.getLatestTrxInfo();
     if (!latestTrxInfo.isInitialized()) {
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
@@ -55,16 +55,17 @@ public class SCMStateMachine extends BaseStateMachine {
   private final StorageContainerManager scm;
   private final SCMRatisServer ratisServer;
   private final Map<RequestType, Object> handlers;
-  private final SCMDBTransactionBuffer transactionBuffer;
+  private final DBTransactionBuffer transactionBuffer;
 
   public SCMStateMachine(final StorageContainerManager scm,
-      final SCMRatisServer ratisServer, final SCMDBTransactionBuffer buffer)
+      final SCMRatisServer ratisServer)
       throws SCMException {
     this.scm = scm;
     this.ratisServer = ratisServer;
     this.handlers = new EnumMap<>(RequestType.class);
-    this.transactionBuffer = buffer;
-    SCMTransactionInfo latestTrxInfo = buffer.getLatestTrxInfo();
+    this.transactionBuffer = scm.getScmHAManager().getDBTransactionBuffer();
+    SCMTransactionInfo latestTrxInfo =
+        this.transactionBuffer.getLatestTrxInfo();
     if (!latestTrxInfo.isInitialized()) {
       if (!updateLastAppliedTermIndex(latestTrxInfo.getTerm(),
           latestTrxInfo.getTransactionIndex())) {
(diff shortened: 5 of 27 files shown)


amaliujia:
I found one of the failed test is `TestMiniOzoneCluster.testDNstartAfterSCM`, which failed on loading RocksDB when restarting SCM. Will try investigate the root cause.

amaliujia:
cc @nandakumar131 @ChenSammi @GlenGeng

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
```diff
@@ -101,6 +102,16 @@ public ContainerManagerImpl(
       final PipelineManager pipelineManager,
       final Table<ContainerID, ContainerInfo> containerStore)
       throws IOException {
+    this(conf, scmHaManager, pipelineManager, containerStore, null);
```
GlenGeng, line 105: Could we avoid using this Ctor ? We may need to change the related test cases.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMTransactionInfoCodec.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class SCMTransactionInfoCodec implements Codec<SCMTransactionInfo> {
+
+  @Override
+  public byte[] toPersistedFormat(SCMTransactionInfo object)
+      throws IOException {
+    checkNotNull(object, "Null object can't be converted to byte array.");
```
GlenGeng, line 32: Ditto,  could we reuse `OMTransactionInfoCodec` ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
```diff
@@ -114,11 +115,21 @@ public static PipelineManagerV2Impl newPipelineManager(
       ConfigurationSource conf, SCMHAManager scmhaManager,
       NodeManager nodeManager, Table<PipelineID, Pipeline> pipelineStore,
       EventPublisher eventPublisher) throws IOException {
+    return newPipelineManager(conf, scmhaManager, nodeManager, pipelineStore,
```
GlenGeng, line 118: Ditto, may we remove this Ctor ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
```diff
@@ -0,0 +1,70 @@
...
+  public SCMDBTransactionBuffer(SCMMetadataStore store) {
+    this.metadataStore = store;
+
+    // initialize a batch operation during construction time
+    currentBatchOperation = this.metadataStore.getStore().initBatchOperation();
+    latestTrxInfo =
+        SCMTransactionInfo
+            .builder()
+            .setTransactionIndex(-1)
+            .setCurrentTerm(-1)
```
GlenGeng, line 41: `.setCurrentTerm(0)`. The first leader will be elected from term 1.

GlenGeng (commented):
LGTM. only some inline comments.

Could we rename the title to `implement Ratis takeSnapshot on SCM`?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMTransactionInfo.java
```diff
@@ -0,0 +1,111 @@
...
+package org.apache.hadoop.hdds.scm.ha;
+
+import com.google.common.base.Preconditions;
+import org.apache.hadoop.hdds.StringUtils;
+
+import java.util.Objects;
+
+import static org.apache.hadoop.ozone.OzoneConsts.TRANSACTION_INFO_SPLIT_KEY;
+
+final public class SCMTransactionInfo {
```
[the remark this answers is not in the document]
amaliujia: To reuse `OMTransactionInfo`, my current answer is probably no.

Though both `OMTransactionInfo` and `SCMTransactionInfo` seems are doing the same thing for now, decoupling is good when there is a need that either `OMTransactionInfo` and `SCMTransactionInfo` need to add new/drop existing fields while the other does not need. Especially in this time that SCM snapshot is still under development. 

I can create a JIRA though to track the merge of two transaction info.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerImpl.java
```diff
@@ -275,12 +278,24 @@ public void addContainer(final ContainerInfoProto containerInfo)
 
     if (!containers.contains(containerID)) {
       ExecutionUtil.create(() -> {
-        containerStore.put(containerID, container);
+        if (transactionBuffer != null) {
+          containerStore.putWithBatch(
+              transactionBuffer.getCurrentBatchOperation(),
+              containerID, container);
+        } else {
+          containerStore.put(containerID, container);
```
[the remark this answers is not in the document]
amaliujia: This is leave for MockHAManager code path in which the buffer is NULL :) 

If we choose to update MockHAManager to have a sort of MockBuffer implementation, then we can move this `else branch` because at that time I think all existing code will access this buffer.

I can create a JIRA to track that add buffer to MockHAManager and remove this else :-)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -89,4 +98,12 @@ private Message process(final SCMRatisRequest request)
     }
   }
 
+  @Override
+  public long takeSnapshot() throws IOException {
+    LOG.info("Current Snapshot Index {}", getLastAppliedTermIndex());
```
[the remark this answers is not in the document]
amaliujia: makes sense.

amaliujia:
With https://github.com/apache/ozone/pull/1733 we will be able to write tests for snapshot based on on single Ratis server SCM setup in MiniOzoneCluster.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
```diff
@@ -0,0 +1,70 @@
...
+            .build();
+  }
+
+  public BatchOperation getCurrentBatchOperation() {
+    return currentBatchOperation;
+  }
+
+  public void updateLatestTrxInfo(SCMTransactionInfo info) {
+    this.latestTrxInfo = info;
```
ChenSammi, line 50: Shall we add some checks here to guarantee new info is greater than current lastTrxInfo?

amaliujia: Good point. I will add a check or a LOG.error for that case.

Theoretically, the Ratis log will be applied with index monotonically increasing thus the new info is supposed to greater than current  lastTrxInfo. But have a check or a LOG will better handle rare cases or bugs.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
```diff
@@ -93,6 +104,6 @@ public String getLocationConfigKey() {
   @Override
   public DBColumnFamilyDefinition[] getColumnFamilies() {
     return new DBColumnFamilyDefinition[] {DELETED_BLOCKS, VALID_CERTS,
-        REVOKED_CERTS, PIPELINES, CONTAINERS};
```
ChenSammi, line 96: Since we add a new table in SCM DB,  we need a JIRA to track how to handle back compatibility during upgrade non HA SCM to HA SCM.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMMetadataStore.java
```diff
@@ -110,6 +111,11 @@
    */
   Table<PipelineID, Pipeline> getPipelineTable();
 
+  /**
+   *
+   */
+  Table <String, SCMTransactionInfo> getTransactionInfoTable();
```
ChenSammi, line 117: Are we missing some comments here?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -89,4 +98,12 @@ private Message process(final SCMRatisRequest request)
...
+    TermIndex lastTermIndex = getLastAppliedTermIndex();
```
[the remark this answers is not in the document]
amaliujia: I believe we don't need to. 

```
    public TermIndex getLastAppliedTermIndex() {
        return (TermIndex)this.lastAppliedTermIndex.get();
    }
```

```
    protected boolean updateLastAppliedTermIndex(long term, long index) {
        TermIndex newTI = TermIndex.newTermIndex(term, index);
        TermIndex oldTI = (TermIndex)this.lastAppliedTermIndex.getAndSet(newTI);
         ....
```

Both function access `this.lastAppliedTermIndex`.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
```diff
@@ -80,6 +82,15 @@
           ContainerInfo.class,
           new ContainerInfoCodec());
 
+  public static final DBColumnFamilyDefinition<String, SCMTransactionInfo>
+      TRANSACTIONINFO =
+      new DBColumnFamilyDefinition<>(
+          "scmTransactionInfoTable",
```
ChenSammi, line 88: scmTransactionInfoTable -> scmTransactionInfos

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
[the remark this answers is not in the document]
ChenSammi: Since all rocksdb writes are batched to write to disk in takesnapshot,  we need to change takesnapshot default frequency of ratis during scm ratis server creation.

ChenSammi:
@amaliujia, thanks for working on this,  I left a few comments.

amaliujia:
@ChenSammi  thank you! I will start to address comments since there are a bunch of actionable ones existing.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManagerImpl.java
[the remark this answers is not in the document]
amaliujia: https://issues.apache.org/jira/browse/HDDS-4634

Create JIRA to add buffer in MockHAManager thus we can remove this `null` check.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
amaliujia: Same https://issues.apache.org/jira/browse/HDDS-4634.

This is because MockHAManager does not have a mock buffer implementation.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
amaliujia: https://issues.apache.org/jira/browse/HDDS-4635

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
amaliujia: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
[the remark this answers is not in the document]
amaliujia: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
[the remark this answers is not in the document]
amaliujia: > Since all rocksdb writes are batched to write to disk in takesnapshot, we need to change takesnapshot default frequency of ratis during scm ratis server creation.

https://issues.apache.org/jira/browse/HDDS-4636 I will prefer to handle separately.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
```diff
@@ -0,0 +1,75 @@
...
+
+import org.apache.hadoop.hdds.scm.metadata.SCMMetadataStore;
+import org.apache.hadoop.hdds.utils.db.BatchOperation;
+import org.apache.hadoop.hdds.utils.db.Table;
+
+import java.io.IOException;
+
+import static org.apache.hadoop.ozone.OzoneConsts.TRANSACTION_INFO_KEY;
+
+public class SCMDBTransactionBuffer {
```
linyiqun, line 27: Could we document javadoc for this? That will be easier understood by others.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMTransactionInfo.java
```diff
@@ -0,0 +1,119 @@
...
```
linyiqun, line 26: Also would be great to add javadoc for this class.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -89,4 +99,15 @@ private Message process(final SCMRatisRequest request)
...
+  @Override
+  public long takeSnapshot() throws IOException {
+    LOG.info("Current Snapshot Index {}", getLastAppliedTermIndex());
+    long startTime = Time.monotonicNow();
+    TermIndex lastTermIndex = getLastAppliedTermIndex();
+    long lastAppliedIndex = lastTermIndex.getIndex();
+    transactionBuffer.flush();
+    LOG.debug("SCM takeSnapshot: {} took {} ms",
+        Time.monotonicNow() - startTime);
+    return lastAppliedIndex;
```
runzhiwang, line 111: Maybe flush should return SCMTransactionInfo, and this return SCMTransactionInfo#transactionIndex ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
```diff
@@ -0,0 +1,75 @@
...
+    currentBatchOperation = this.metadataStore.getStore().initBatchOperation();
+    latestTrxInfo =
+        SCMTransactionInfo
+            .builder()
+            .setTransactionIndex(-1)
+            .setCurrentTerm(0)
+            .build();
+  }
+
+  public BatchOperation getCurrentBatchOperation() {
```
runzhiwang, line 45: Expose currentBatchOperation has a risk, if someone commit this batch operation not by calling flush(), then applyIndex did not wrote into RocksDB.

amaliujia: hmm I think the goal here is commiting batch operations won't need to flush DB. The flushing is controlled by takeSnapshot().

Is there a case that committing will have a following flush?

amaliujia:
@ChenSammi @runzhiwang @GlenGeng @linyiqun 

I have rebased and updated this PR. Now this PR includes the takeSnapshot and loadSnapshot with a test.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
amaliujia: Code changes a bit. Maybe take another look?

amaliujia:
Will fix failing tests.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
runzhiwang: I'm afraid if someone do not know must call flush() to sync, but sync `currentBatchOperation `directly, then inconsistent will happen.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMTransactionInfo.java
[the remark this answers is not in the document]
runzhiwang: Sorry, I'm not clear why we can not reuse OMTransactionInfo. They has the same fields.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
amaliujia: Do you know is there a better mechanism to achieve the buffering? 

I studied what OM does and that is different from SCM. OM just buffer entry and then apply entries in a batch, but SCM need to router such entry to handlers and then different handler will apply the entry.

So if we change to the way that OM is doing, there will be a good amount of refactoring needs happen, which might be not appropriate in this PR. E.g. We will need to move handlers to the buffer class and insert entries into buffer. Buffer will be the place to trouter entries to right state managers.

What do you think? I can create a PR to track such refactoring thus we won't need to expose  `currentBatchOperation`

GlenGeng: How about return an anonymous subclass of `RDBBatchOperation` whose 'commit()' will throw `RuntimeException`, which means you can only write to the batch, but there is no way to commit this batch ?

`RDBBatchOperation#commit()` is called by `RDBStore#commitBatchOperation()`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMTransactionInfo.java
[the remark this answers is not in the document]
amaliujia: I think we are better to reuse `OMTransactionInfo` after merge HDDS-2823 back to master.

Not reusing `OMTransactionInfo` is to try to reduce the risk that have conflicts after syncing master int HDDS-2823. Also, to reuse `OMTransactionInfo`, I think we'd better to update the name to `TransactionInfo`, which will be a source of complicates, unless we can change it in the master branch first.

To conclude a bit, we can reuse it after merging SCM HA back to master, the goal is try to reduce potential conflicts if there could be any.

SCM HA is in developing branch anyway and I am sure we will need to clean up lots of stuff eventually. As long as we have JIRAs to not lose track of such work.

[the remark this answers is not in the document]
GlenGeng: `OMTransactionInfo` is written by Bharat, since SCM and OM has a quite similar requirement, we may have a talk with him, to determine whether OM and SCM can share the same `TransactionInfo` and the `Codec`. @amaliujia  What do you think ?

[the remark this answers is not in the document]
amaliujia: Sure we can do that. I will need add some changes in main branch, and then sync that to HDDS-2823 and then reuse in SCM.

I think we should do that separately as that will take time. I have created https://issues.apache.org/jira/browse/HDDS-4660 to track this effort.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
amaliujia: Glen's idea might be much easier. That's a good option.

I think I will separate it to:
1. add javadoc to `getCurrentBatchOperation()` to remind that do not use this batch to commit.
2. Apply a solution to fix this issue. Created https://issues.apache.org/jira/browse/HDDS-4661 to track it.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisSnapshotInfo.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public SCMRatisSnapshotInfo(long term, long index) {
+    this.term = term;
+    this.snapshotIndex = index;
+  }
+
+  public void updateTermIndex(long newTerm, long newIndex) {
+    this.term = newTerm;
+    this.snapshotIndex = newIndex;
+  }
```
linyiqun, line 46: I don't see this method is used in this PR. Will this be used in follow-up task? If not we could remove this and remove volatile keyword for term/snapshotIndex as well. Because I didn't find the concurrent update for this two variables.

amaliujia: got it. Will remove this method.

The `volatile` is a good point. I am also looking for comments to point me out whether there are potential concurrent operations.  I will remove `volatile` if there is no comment about this situation.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
GlenGeng: Could we implement the mock buffer in this PR ? Since missing this change will pollute the production code, and will also give a burden to the on-going PR of deleted block log.

GlenGeng (commented):
Thanks for the job!

Better solve following task in this PR, it should be part of the job of load snapshot.
https://issues.apache.org/jira/browse/HDDS-4533

amaliujia:
> Thanks for the job!
> 
> Better solve following task in this PR, it should be part of the job of load snapshot.
> https://issues.apache.org/jira/browse/HDDS-4533

@GlenGeng  I prefer to fix such issue separately so create a PR for it: https://github.com/apache/ozone/pull/1796

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -89,4 +128,33 @@ private Message process(final SCMRatisRequest request)
...
+          transactionBuffer.getLatestTrxInfo().toSnapshotInfo());
+    } else {
+      lastAppliedIndex =
+          transactionBuffer.getLatestTrxInfo().getTransactionIndex();
+    }
+
+    transactionBuffer.flush();
+    transactionBuffer.setLatestSnapshot(
+        transactionBuffer.getLatestTrxInfo().toSnapshotInfo());
+    LOG.debug("SCM takeSnapshot: {} took {} ms",
```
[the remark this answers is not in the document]
amaliujia: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -55,15 +84,25 @@ public void registerHandler(RequestType type, Object handler) {
     try {
       final SCMRatisRequest request = SCMRatisRequest.decode(
           Message.valueOf(trx.getStateMachineLogEntry().getLogData()));
-      applyTransactionFuture.complete(process(request));
+      applyTransactionFuture.complete(
+          process(
```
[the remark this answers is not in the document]
amaliujia: +1

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
```diff
@@ -0,0 +1,101 @@
...
+  public void flush() throws IOException {
+    // write latest trx info into trx table in the same batch
+    Table<String, SCMTransactionInfo> transactionInfoTable
+        = metadataStore.getTransactionInfoTable();
+    transactionInfoTable.putWithBatch(currentBatchOperation,
+        TRANSACTION_INFO_KEY, latestTrxInfo);
+
+    metadataStore.getStore().commitBatchOperation(currentBatchOperation);
+    currentBatchOperation.close();
+
```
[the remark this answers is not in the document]
amaliujia: The snapshot info is set also during the initialization. So not only `flush()` will need to update that information.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -55,15 +84,25 @@ public void registerHandler(RequestType type, Object handler) {
...
     } catch (Exception ex) {
       applyTransactionFuture.completeExceptionally(ex);
     }
     return applyTransactionFuture;
   }
 
-  private Message process(final SCMRatisRequest request)
-      throws Exception {
+  private boolean shouldUpdate(SCMTransactionInfo info) {
+    return !(info.getTransactionIndex() == -1 && info.getTerm() == 0);
```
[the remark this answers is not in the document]
amaliujia: Makes sense

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
```diff
@@ -111,15 +113,25 @@ private PipelineManagerV2Impl(ConfigurationSource conf,
     this.pipelineCreationAllowed = new AtomicBoolean(!this.isInSafeMode.get());
   }
 
-  public static PipelineManagerV2Impl newPipelineManager(
+  public static PipelineManagerV2Impl newPipelineManagerWithMockBuffer(
```
GlenGeng, line 116: `DBTransactionBuffer` can be fetched by calling `SCMHAManager#getDBTransactionBuffer`, so this function `newPipelineManagerWithMockBuffer` is not needed.
```
// Create PipelineStateManager
StateManager stateManager = PipelineStateManagerV2Impl
    .newBuilder().setPipelineStore(pipelineStore)
    .setRatisServer(scmhaManager.getRatisServer())
    .setNodeManager(nodeManager)
    .setSCMDBTransactionBuffer(scmhaManager.getDBTransactionBuffer())
    .build()
```

amaliujia: This is a very good point.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMTransactionInfo.java
```diff
@@ -0,0 +1,145 @@
...
+    term = Long.parseLong(tInfo[0]);
+    transactionIndex = Long.parseLong(tInfo[1]);
+  }
+
+  private SCMTransactionInfo(long currentTerm, long transactionIndex) {
+    this.term = currentTerm;
+    this.transactionIndex = transactionIndex;
+  }
+
+  public boolean shouldUpdate() {
```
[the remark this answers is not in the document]
amaliujia: `isInitialized` is a good naming.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -54,7 +54,7 @@ public SCMStateMachine(SCMDBTransactionBuffer buffer) throws SCMException {
     this.handlers = new EnumMap<>(RequestType.class);
     this.transactionBuffer = buffer;
     SCMTransactionInfo latestTrxInfo = buffer.getLatestTrxInfo();
-    if (latestTrxInfo.shouldUpdate()) {
+    if (!latestTrxInfo.isInitialized()) {
```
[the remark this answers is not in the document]
amaliujia: Actually it is `!latestTrxInfo.isInitialized()`.

`SCMTransactionInfo latestTrxInfo = buffer.getLatestTrxInfo();` comes from the buffer. The buffer will load the SCMTransactionInfo from the DB. If there was no snapshot taken before (e.g. brand new SCM), there isn't a SCMTransactionInfo from DB thus buffer will create a default one. 

So we only need to updateLastAppliedTermIndex when latestTrxInfo is not initialized (thus it is loaded from DB and it's meaningful).

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
bshashikant: In my opinion, we should stop the ratis server here of such a case arise. It is fatal case where we cannot move ahead from here.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -89,4 +119,30 @@ private Message process(final SCMRatisRequest request)
...
+          transactionBuffer.getLatestTrxInfo().toSnapshotInfo());
+    } else {
+      lastAppliedIndex =
+          transactionBuffer.getLatestTrxInfo().getTransactionIndex();
+    }
+
+    transactionBuffer.flush();
+    LOG.info("Current Snapshot Index {}, takeSnapshot took {} ms",
+        getLastAppliedTermIndex(), Time.monotonicNow() - startTime);
+    return lastAppliedIndex;
```
bshashikant, line 146: why the lastApplied index returned here is the one derived from transaction buffer? Once the snapshot is taken, the last applied index should be given by getLastAppliedTermIndex() not transaction buffer which may be behind actual last applied index in ratis?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
```diff
@@ -114,12 +114,14 @@ private PipelineManagerV2Impl(ConfigurationSource conf,
   public static PipelineManagerV2Impl newPipelineManager(
       ConfigurationSource conf, SCMHAManager scmhaManager,
       NodeManager nodeManager, Table<PipelineID, Pipeline> pipelineStore,
-      EventPublisher eventPublisher) throws IOException {
+      EventPublisher eventPublisher)
```
bshashikant, line 117: unintended change??

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
```diff
@@ -0,0 +1,112 @@
...
+  public void flush() throws IOException {
+    // write latest trx info into trx table in the same batch
+    Table<String, SCMTransactionInfo> transactionInfoTable
+        = metadataStore.getTransactionInfoTable();
+    transactionInfoTable.putWithBatch(currentBatchOperation,
+        TRANSACTION_INFO_KEY, latestTrxInfo);
+
+    metadataStore.getStore().commitBatchOperation(currentBatchOperation);
+    currentBatchOperation.close();
+    this.latestSnapshot = latestTrxInfo.toSnapshotInfo();
```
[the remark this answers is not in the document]
amaliujia: I agree with it :-)

In fact we have discussed not to expose batch operation above and I created  https://issues.apache.org/jira/browse/HDDS-4661.

As this PR becomes larger and larger so I am planning to address this batch operation related comments in HDDS-4661, including when to init and close it.

How do you feel? Do you agree with my idea?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
amaliujia: Yes :) will undo it.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
amaliujia: O the code here is keeping the lastAppliedIndex in buffer to equal to the getLastAppliedTermIndex() actually. 

```
    LOG.info("Current Snapshot Index {}, takeSnapshot took {} ms",
        getLastAppliedTermIndex(), Time.monotonicNow() - startTime);
```
is wrong.

I should replace `getLastAppliedTermIndex()` with `lastAppliedIndex` to make it clear.


We need the `lastAppliedIndex` in buffer keep updated so that will be written into DB.

amaliujia:
Failed UT is the decommission one, which is known to be flaky.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
bshashikant: I see the buffer getting updated with applied index only after applyTransaction is called. ApplyTransaction is not invoked for every log transaction in Ratis. So, its better to rely on getLastAppliedTermIndex() rather than the buffer for the latest applied index.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -26,27 +26,55 @@
...
     handlers.put(type, handler);
   }
 
+  @Override
+  public SnapshotInfo getLatestSnapshot() {
+    return transactionBuffer.getLatestSnapshot();
+  }
+
   @Override
   public CompletableFuture<Message> applyTransaction(
```
bshashikant, line 79: Can we rename applyTransaction to applyTansactionSerial as this is a serialized operation anyways?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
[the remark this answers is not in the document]
bshashikant: I am ok with addressing in a different PR.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
amaliujia: I created https://issues.apache.org/jira/browse/HDDS-4684.

After check the Ratis state machine interface, I can see two functions:
```
  /**
   * Called for transactions that have been committed to the RAFT log. This step is called
   * sequentially in strict serial order that the transactions have been committed in the log.
   * The SM is expected to do only necessary work, and leave the actual apply operation to the
   * applyTransaction calls that can happen concurrently.
   * @param trx the transaction state including the log entry that has been committed to a quorum
   *            of the raft peers
   * @return The Transaction context.
   */
  TransactionContext applyTransactionSerial(TransactionContext trx);

  /**
   * Apply a committed log entry to the state machine. This method can be called concurrently with
   * the other calls, and there is no guarantee that the calls will be ordered according to the
   * log commit order.
   * @param trx the transaction state including the log entry that has been committed to a quorum
   *            of the raft peers
   */
  CompletableFuture<Message> applyTransaction(TransactionContext trx);
```

So there is no API as the following 
```
  Message applyTransactionSerial(TransactionContext trx);
```

Basically a function to apply transaction in strict serial order where it returns a Message.

I am planning to send an email to Ratis community to discuss the intention of only having one applyTransaction that returns Message (CompletableFuture<Message>) but can be called concurrently.

We can address this API change in HDDS-4684.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerManagerImpl.java
```diff
@@ -397,4 +396,8 @@ public void close() throws IOException {
   protected ContainerStateManagerV2 getContainerStateManager() {
     return containerStateManager;
   }
+
+  public SCMHAManager getSCMHAManager() {
```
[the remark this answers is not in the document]
amaliujia: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisSnapshotInfo.java
amaliujia: That makes sense. Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMDBTransactionBuffer.java
amaliujia: Such case should be considered as a bug and be fixed. 

I created a JIRA to further discuss how to deal with such case: https://issues.apache.org/jira/browse/HDDS-4723

bshashikant (approved):
The patch looks ok. The changes discussed will be addressed in subsequent jira.

The transaction buffer flush now can only happen via a ratis snapshot , but in case if ratis is not enabled, there needs to be rocks db sync on every update to db or we need a way to periodically flush the buffer changes to db. This can be done as a part of separate jira.

bshashikant:
Thanks @amaliujia for the contribution.

PR #1730, opened 2020-12-23
HDDS-4619. Intermittent error exporting same container to multiple datanodes

adoroszlai:

Container export performs compaction while holding only read lock.  We need to acquire write lock to prevent concurrent export changing the files while others are packing the same files into tar archive.

This should fix intermittent exception which causes some of the `TestDecommissionAndMaintenance` failures:

```
java.lang.IllegalArgumentException: refCount: [1]
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:203)
	at org.apache.hadoop.ozone.container.common.utils.ContainerCache.removeDB(ContainerCache.java:203)
	at org.apache.hadoop.ozone.container.keyvalue.helpers.BlockUtils.removeDB(BlockUtils.java:138)
	at org.apache.hadoop.ozone.container.keyvalue.KeyValueContainer.exportContainerData(KeyValueContainer.java:533)
	at org.apache.hadoop.ozone.container.keyvalue.KeyValueHandler.exportContainer(KeyValueHandler.java:939)
	at org.apache.hadoop.ozone.container.ozoneimpl.ContainerController.exportContainer(ContainerController.java:145)
	at org.apache.hadoop.ozone.container.replication.OnDemandContainerReplicationSource.copyData(OnDemandContainerReplicationSource.java:59)
	at org.apache.hadoop.ozone.container.replication.GrpcReplicationService.download(GrpcReplicationService.java:56)
...
[grpc-default-executor-9] ERROR replication.GrpcReplicationClient (GrpcReplicationClient.java:onError(161)) - Download of container 4 was unsuccessful
```

https://issues.apache.org/jira/browse/HDDS-4619

## How was this patch tested?

* added unit test: https://github.com/adoroszlai/hadoop-ozone/runs/1599885042
* ran `TestDecommissionAndMaintenance` 10x: https://github.com/adoroszlai/hadoop-ozone/runs/1599907254

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
@@ -516,22 +516,27 @@ public class KeyValueContainer implements Container<KeyValueContainerData> {
   @Override
   public void exportContainerData(OutputStream destination,
       ContainerPacker<KeyValueContainerData> packer) throws IOException {
-    // Closed/ Quasi closed containers are considered for replication by
-    // replication manager if they are under-replicated.
-    ContainerProtos.ContainerDataProto.State state =
-        getContainerData().getState();
-    if (!(state == ContainerProtos.ContainerDataProto.State.CLOSED ||
-        state == ContainerDataProto.State.QUASI_CLOSED)) {
-      throw new IllegalStateException(
-          "Only closed/quasi closed containers could be exported: " +
-              "Where as ContainerId="
-              + getContainerData().getContainerID() + " is in state " + state);
+    writeLock();
+    try {
+      // Closed/ Quasi closed containers are considered for replication by
+      // replication manager if they are under-replicated.
+      ContainerProtos.ContainerDataProto.State state =
+          getContainerData().getState();
+      if (!(state == ContainerProtos.ContainerDataProto.State.CLOSED ||
+          state == ContainerDataProto.State.QUASI_CLOSED)) {
+        throw new IllegalStateException(
+            "Only closed/quasi closed containers could be exported: " +
+                "Where as ContainerId="
+                + getContainerData().getContainerID() + " is in state " + state);
+      }
+      compactDB();
+      // Close DB (and remove from cache) to avoid concurrent modification while
+      // packing it.
+      BlockUtils.removeDB(containerData, config);
+      packer.pack(this, destination);
+    } finally {
+      writeUnlock();
     }
-    compactDB();
-    // Close DB (and remove from cache) to avoid concurrent modification while
-    // packing it.
-    BlockUtils.removeDB(containerData, config);
-    packer.pack(this, destination);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lamber-ken:
hi @adoroszlai the changes are hard to understand, may only use `writeLock` is ok, like 
```java
@Override
public void exportContainerData(OutputStream destination,
    ContainerPacker<KeyValueContainerData> packer) throws IOException {
  
  try {
    writeLock();

    // Closed/ Quasi closed containers are considered for replication by
    // replication manager if they are under-replicated.
    ContainerProtos.ContainerDataProto.State state =
        getContainerData().getState();
    if (!(state == ContainerProtos.ContainerDataProto.State.CLOSED ||
        state == ContainerDataProto.State.QUASI_CLOSED)) {
      throw new IllegalStateException(
          "Only closed/quasi closed containers could be exported: " +
              "Where as ContainerId="
              + getContainerData().getContainerID() + " is in state " + state);
    }
    compactDB();
    // Close DB (and remove from cache) to avoid concurrent modification while
    // packing it.
    BlockUtils.removeDB(containerData, config);
    packer.pack(this, destination);
    
  } finally {
    writeUnlock();
  }
  
}
```

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
```diff
@@ -516,22 +516,36 @@ public void importContainerData(InputStream input,
...
-    // replication manager if they are under-replicated.
-    ContainerProtos.ContainerDataProto.State state =
-        getContainerData().getState();
-    if (!(state == ContainerProtos.ContainerDataProto.State.CLOSED ||
-        state == ContainerDataProto.State.QUASI_CLOSED)) {
-      throw new IllegalStateException(
-          "Only closed/quasi closed containers could be exported: " +
-              "Where as ContainerId="
-              + getContainerData().getContainerID() + " is in state " + state);
+    readLock();
```
sodonnel, line 519: The exception here should be a rare occurrence, so we are almost always going to have to escalate the read lock into a write lock. I think it would be simpler to just take the write lock at the start and forget about the read lock.

The part of the code under the read lock should be very fast, and the code under the write lock may be slow (compactDB) so I suspect we are saving very little by doing some work under the read lock. I think this lock is also only for this container (not for all containers on the DN) so contention should not be a big issue. 

What do you think?

sodonnel:
The change here looks mostly good to me. I just wonder if we should go with only a write lock, rather than read then write lock.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
adoroszlai: Agreed, the first read lock can be removed.  I think we need to keep the second one, though (for `packer.pack()`).

lamber-ken:
LGTM +1 👍

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
lamber-ken: > Agreed, the first read lock can be removed. I think we need to keep the second one, though (for `packer.pack()`).

Agree +1

adoroszlai: Thanks @sodonnel for the suggestion.  Patch is updated accordingly, can you please take another look?

sodonnel:
The latest change LGTM +1

adoroszlai:
Thanks @lamber-ken and @sodonnel for the reviews.

PR #1733, opened 2020-12-24
HDDS-4622. Use single ratis server in MiniOzoneCluster tests for SCM

amaliujia:

The intention is to replace MockRatisServer by production Ratis server in integration tests (tests that use  `MiniOzoneCluster`s)

https://issues.apache.org/jira/browse/HDDS-4622

## How was this patch tested?

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
@@ -38,7 +38,7 @@ public class SCMHAConfiguration {
 
   @Config(key = "ratis.storage.dir",
       type = ConfigType.STRING,
-      defaultValue = "",
+      defaultValue = "scm-ratis",
       tags = {OZONE, SCM, HA, RATIS},
       description = "Storage directory used by SCM to write Ratis logs."
   )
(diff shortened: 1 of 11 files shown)


amaliujia:
cc: @GlenGeng

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
```diff
@@ -38,7 +38,7 @@
 
   @Config(key = "ratis.storage.dir",
       type = ConfigType.STRING,
-      defaultValue = "",
+      defaultValue = "scm-ratis",
```
amaliujia, line 41: This is workaround to config SCM ratis storage dir value. The usual way that config value in SCMHAConfiguration seems break. I will file a JIRA to fix it.

amaliujia: https://issues.apache.org/jira/browse/HDDS-4624

amaliujia: We won't need this line of change after https://github.com/apache/ozone/pull/1739

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneClusterImpl.java
```diff
@@ -169,17 +169,21 @@ public void waitForClusterToBeReady()
       throws TimeoutException, InterruptedException {
     GenericTestUtils.waitFor(() -> {
       final int healthy = scm.getNodeCount(HEALTHY);
+      // When SCM HA is not enabled, scm is always leader.
+      final boolean checkScmLeader = scm.checkLeader();
```
GlenGeng, line 173: Good!

PR #1735, opened 2020-12-24
HDDS-4623. Update CountQuota to NamesapceQuota

amaliujia:

Use more description words "namespace quota" to replace "count quota".

https://issues.apache.org/jira/browse/HDDS-4623

## How was this patch tested?

UT

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
@@ -432,7 +432,7 @@ public class OzoneBucket extends WithMetadata {
    *
    * @throws IOException
    */
-  public void clearCountQuota() throws IOException {
+  public void clearNamespaceQuota() throws IOException {
     OzoneBucket ozoneBucket = proxy.getBucketDetails(volumeName, name);
     proxy.setBucketQuota(volumeName, name, QUOTA_RESET,
         ozoneBucket.getQuotaInBytes());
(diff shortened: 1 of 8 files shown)


amaliujia:
R: @captainzmc

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneVolume.java
```diff
@@ -300,7 +300,7 @@ public void clearSpaceQuota() throws IOException {
    *
    * @throws IOException
    */
-  public void clearCountQuota() throws IOException {
+  public void clearNamespaceQuota() throws IOException {
```
captainzmc, line 303: Same as above

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
```diff
@@ -432,7 +432,7 @@ public void clearSpaceQuota() throws IOException {
    *
    * @throws IOException
    */
-  public void clearCountQuota() throws IOException {
+  public void clearNamespaceQuota() throws IOException {
```
captainzmc, line 435: The method description also needs to be update.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/ClearSpaceQuotaOptions.java
```diff
@@ -30,14 +30,14 @@
 
   @CommandLine.Option(names = {"--namespace-quota"},
       description = "clear count quota")
```
captainzmc, line 32: update description.

captainzmc (commented):
Thanks for the update. LGTM overall, just some minor comments. You can also check if there are any other modifications needed.

amaliujia:
@captainzmc  

comments addressed.

ayushtkn (approved):
I tried to find any other usage of count quota, but couldn't find any.
Changes LGTM

captainzmc (approved):
LGTM +1. Thanks for @amaliujia's patch and @ayushtkn's review.

PR #1737, opened 2020-12-25
HDDS-4568. Add SCMContext to SCM HA

GlenGeng:

We want to provide SCMContext, which would be a single source of truth for some key information that is shared across all components within SCM.

SCMContext holds two kind of key information:
- RaftServer related info: isLeader, term.
- SafeMode related info: inSafeMode, preCheckComplete.

https://issues.apache.org/jira/browse/HDDS-4568

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
@@ -31,6 +31,7 @@ import org.apache.hadoop.hdds.scm.ScmConfig;
 import org.apache.hadoop.hdds.scm.container.ContainerID;
 import org.apache.hadoop.hdds.scm.container.ContainerManagerV2;
 import org.apache.hadoop.hdds.scm.events.SCMEvents;
+import org.apache.hadoop.hdds.scm.ha.SCMContext;
 import org.apache.hadoop.hdds.scm.node.NodeManager;
 import org.apache.hadoop.hdds.scm.node.NodeStatus;
 import org.apache.hadoop.hdds.server.events.EventPublisher;
@@ -40,6 +41,7 @@ import org.apache.hadoop.hdds.utils.BackgroundTaskQueue;
 import org.apache.hadoop.hdds.utils.BackgroundTaskResult.EmptyTaskResult;
 import org.apache.hadoop.ozone.protocol.commands.CommandForDatanode;
 import org.apache.hadoop.ozone.protocol.commands.DeleteBlocksCommand;
+import org.apache.hadoop.ozone.protocol.commands.SCMCommand;
 import org.apache.hadoop.util.Time;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -64,19 +66,22 @@ public class SCMBlockDeletingService extends BackgroundService {
   private final ContainerManagerV2 containerManager;
   private final NodeManager nodeManager;
   private final EventPublisher eventPublisher;
+  private final SCMContext scmContext;
 
   private int blockDeleteLimitSize;
 
+  @SuppressWarnings("parameternumber")
   public SCMBlockDeletingService(DeletedBlockLog deletedBlockLog,
       ContainerManagerV2 containerManager, NodeManager nodeManager,
-      EventPublisher eventPublisher, Duration interval, long serviceTimeout,
-      ConfigurationSource conf) {
+      EventPublisher eventPublisher, SCMContext scmContext,
+      Duration interval, long serviceTimeout, ConfigurationSource conf) {
     super("SCMBlockDeletingService", interval.toMillis(), TimeUnit.MILLISECONDS,
         BLOCK_DELETING_SERVICE_CORE_POOL_SIZE, serviceTimeout);
     this.deletedBlockLog = deletedBlockLog;
     this.containerManager = containerManager;
     this.nodeManager = nodeManager;
     this.eventPublisher = eventPublisher;
+    this.scmContext = scmContext;
 
     blockDeleteLimitSize =
         conf.getObject(ScmConfig.class).getBlockDeletionLimit();
@@ -146,9 +151,10 @@ public class SCMBlockDeletingService extends BackgroundService {
               // We should stop caching new commands if num of un-processed
               // command is bigger than a limit, e.g 50. In case datanode goes
               // offline for sometime, the cached commands be flooded.
+              SCMCommand<?> command = new DeleteBlocksCommand(dnTXs);
+              command.setTerm(scmContext.getTerm());
               eventPublisher.fireEvent(SCMEvents.DATANODE_COMMAND,
-                  new CommandForDatanode<>(dnId,
-                      new DeleteBlocksCommand(dnTXs)));
+                  new CommandForDatanode<>(dnId, command));
               if (LOG.isDebugEnabled()) {
                 LOG.debug(
                     "Added delete block command for datanode {} in the queue,"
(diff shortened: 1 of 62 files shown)


GlenGeng:
cc @nandakumar131 @ChenSammi @amaliujia

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
```diff
@@ -516,6 +515,15 @@ public RaftServer getServer() {
     return server;
   }
 
+  public RaftServer.Division getServerDivision() throws IOException {
```
amaliujia, line 518: This part of a change is contained in master (https://github.com/apache/ozone/pull/1728)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -89,4 +103,27 @@ private Message process(final SCMRatisRequest request)
...
+    if (!groupMemberId.getPeerId().equals(newLeaderId)) {
+      LOG.info("leader changed, yet current SCM is still follower.");
+      return;
+    }
+
+    long term = scm.getScmHAManager()
+        .getRatisServer()
+        .getDivision()
+        .getInfo()
+        .getCurrentTerm();
```
amaliujia, line 124: Will it make sense to use SCMContext to get current term?

amaliujia (commented):
I think some of the changes related to new Ratis API are in master branch already. It might be better to merge master to 2823 branch and then rebase this PR to remove those changes (e.g. changes in `XceiverServerRatis`)

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
amaliujia: I see. so this is from the cherry picked commit from master.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
```diff
@@ -146,9 +151,10 @@ public EmptyTaskResult call() throws Exception {
               // We should stop caching new commands if num of un-processed
               // command is bigger than a limit, e.g 50. In case datanode goes
               // offline for sometime, the cached commands be flooded.
+              SCMCommand<?> command = new DeleteBlocksCommand(dnTXs);
+              command.setTerm(scmContext.getTerm());
```
linyiqun, line 155: Could we also catch the NotLeaderException like other places does?  There are some other places we don't catch this exception that is swallowed by IOException.
```java
        } catch (NotLeaderException nle) {

        }  catch (IOException e) {
          // We may tolerate a number of failures for sometime
          // but if it continues to fail, at some point we need to raise
          // an exception and probably fail the SCM ? At present, it simply
          // continues to retry the scanning.
          LOG.error("Failed to get block deletion transactions from delTX log",
              e);
          return EmptyTaskResult.newResult();
        }
```

linyiqun (commented):
@GlenGeng , I left one minor comment below.

I see now we use the term index value as the SCM leader check. This is used across SCM internal components. Does this will cover the client request behaviour? 

For example, one client configured a single SCM address that is a Follower role and then send the request. Will it update the SCM metadata, like pipeline, containers?  I see there was a isLeader check before but now that was removed in HDDS-4551. Do you know the context for this?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
GlenGeng: We use SCMContext to encapsulate raft related info, so that components in SCM won't need to hold a reference of SCMHAManager or SCMRatisServer.

For non-HA mode or unit test, we just need an empty SCMContext, instead of a mocked SCMHAManager or a mocked SCMRatisServer.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
GlenGeng: Sure, I will fix them in the next patch.

GlenGeng:
> @GlenGeng , I left one minor comment below.
> 
> I see now we use the term index value as the SCM leader check. This is used across SCM internal components. Does this will cover the client request behaviour?
> 
> For example, one client configured a single SCM address that is a Follower role and then send the request. Will it update the SCM metadata, like pipeline, containers? I see there was a isLeader check before but now that was removed in [HDDS-4551](https://issues.apache.org/jira/browse/HDDS-4551). Do you know the context for this?

Hey Yiqun, thanks for the review! 
> For example, one client configured a single SCM address that is a Follower role and then send the request.

For now, we need client to know all the SCM instances that engage in the SCM raft cluster, if client send request to a follower SCM, it will get a NotLeaderException, and failover to the next SCM instance.

>  Will it update the SCM metadata, like pipeline, containers?

We removed the leader check in HDDS-4551, since all the metadata updates that will be saved into rocksdb will go through ratis, as a RaftClientRequest, so if underly Raft is in a non-Leader role, the replied RaftClientReply will be injected with a NotLeaderException. 

Please check `SCMHAInvocationHandler` to see how we implement this: for now, add container, remove container, update container state, add pipeline, remove pipeline, update pipeline state will go through ratis.

If needed, we can schedule a zoom meeting to discuss about current SCM HA design.

linyiqun:
@GlenGeng , thanks for the detailed explanation!

amaliujia:
Thanks Glen to rebase PR! Will try to give another pass on this PR.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/TestSCMContext.java
```diff
@@ -0,0 +1,67 @@
...
+import static org.junit.Assert.fail;
+
+/**
+ * Test for SCMContext.
+ */
+public class TestSCMContext {
+  @Test
+  public void testRaftOperations() {
+    // start as follower
+    SCMContext scmContext = new SCMContext(false, 0, null, null);
```
amaliujia, line 37: Nit: will it better to replace constructor by a builder for creating instances of SCMContext?

amaliujia (commented):
Overall LGTM. Left a comment.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/TestSCMContext.java
GlenGeng: Agree. Will fix in https://github.com/apache/ozone/pull/1784.

runzhiwang (approved):
LGTM

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/TestSCMContext.java
bshashikant: i agree.

bshashikant:
Thanks @GlenGeng for detailed explanation. The changes look good to me except one minor comment. Thanks for the efforts.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/TestSCMContext.java
GlenGeng: If you have reviewed `HDDS-4295. Add SCMServiceManager to SCM HA`, please ignore this PR, since the first 4 commits of SCMServiceManager are exactly from this PR.

btw, I have added a builder for SCMContext in the latest commit of SCMServiceManager.

bshashikant (approved):

PR #1738, opened 2020-12-25
HDDS-4622: Use singe server raft cluster in MiniOzoneCluster.

GlenGeng:

Use `SCMHAManagerImpl` and `SCMRatisServerImpl` instead of `MockSCMHAManager` and `MockRatisServer` in `MiniOzoneCluster`.
The `MockRatisServer` and `MockSCMHAManager` can not fully test the HA code.

https://issues.apache.org/jira/browse/HDDS-4622

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
@@ -76,7 +76,8 @@ public final class RatisUtil {
                                        final ConfigurationSource conf) {
     String storageDir = haConf.getRatisStorageDir();
     if (Strings.isNullOrEmpty(storageDir)) {
-      storageDir = ServerUtils.getDefaultRatisDirectory(conf);
+      File metaDirPath = ServerUtils.getOzoneMetaDirPath(conf);
+      storageDir = (new File(metaDirPath, "scm-ha")).getPath();
     }
     RaftServerConfigKeys.setStorageDir(properties,
         Collections.singletonList(new File(storageDir)));
@@ -100,9 +101,9 @@ public final class RatisUtil {
     Rpc.setRequestTimeout(properties, TimeDuration.valueOf(
         conf.getRatisRequestTimeout(), TimeUnit.MILLISECONDS));
     Rpc.setTimeoutMin(properties, TimeDuration.valueOf(
-        conf.getRatisRequestMinTimeout(), TimeUnit.MILLISECONDS));
+        conf.getLeaderElectionMinTimeout(), TimeUnit.MILLISECONDS));
     Rpc.setTimeoutMax(properties, TimeDuration.valueOf(
-        conf.getRatisRequestMaxTimeout(), TimeUnit.MILLISECONDS));
+        conf.getLeaderElectionMaxTimeout(), TimeUnit.MILLISECONDS));
     Rpc.setSlownessTimeout(properties, TimeDuration.valueOf(
         conf.getRatisNodeFailureTimeout(), TimeUnit.MILLISECONDS));
   }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
@@ -181,79 +181,46 @@ public final class PipelineManagerV2Impl implements PipelineManager {
   @Override
   public Pipeline getPipeline(PipelineID pipelineID)
       throws PipelineNotFoundException {
-    lock.readLock().lock();
-    try {
-      return stateManager.getPipeline(pipelineID);
-    } finally {
-      lock.readLock().unlock();
-    }
+    return stateManager.getPipeline(pipelineID);
   }
 
   @Override
   public boolean containsPipeline(PipelineID pipelineID) {
-    lock.readLock().lock();
     try {
       getPipeline(pipelineID);
       return true;
     } catch (PipelineNotFoundException e) {
       return false;
-    } finally {
-      lock.readLock().unlock();
     }
   }
 
   @Override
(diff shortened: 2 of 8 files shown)


amaliujia:
O seems like we are more or less doing the similar change :) https://github.com/apache/ozone/pull/1733

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
```diff
@@ -76,7 +76,8 @@ public static void setRaftStorageDir(final RaftProperties properties,
                                        final ConfigurationSource conf) {
     String storageDir = haConf.getRatisStorageDir();
     if (Strings.isNullOrEmpty(storageDir)) {
-      storageDir = ServerUtils.getDefaultRatisDirectory(conf);
+      File metaDirPath = ServerUtils.getOzoneMetaDirPath(conf);
+      storageDir = (new File(metaDirPath, "scm-ha")).getPath();
```
amaliujia, line 80: I think it's prefered to use SCMHA Configuration to set the storageDir

GlenGeng: agree

GlenGeng:
> O seems like we are more or less doing the similar change :) #1733

We can fix this issue together, and you can take away my changes if needed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
amaliujia: After https://github.com/apache/ozone/pull/1739 this can be set through HA config.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/TestUtils.java
```diff
@@ -475,7 +475,6 @@ public static void quasiCloseContainer(ContainerManagerV2 containerManager,
   public static StorageContainerManager getScmSimple(OzoneConfiguration conf)
       throws IOException, AuthenticationException {
     SCMConfigurator configurator = new SCMConfigurator();
-    configurator.setSCMHAManager(MockSCMHAManager.getInstance(true));
```
amaliujia, line 478: You will need this line: 
`conf.setBoolean(ScmConfigKeys.OZONE_SCM_HA_ENABLE_KEY, true);`

`SCMManagerImpl` will return isLeader immediately without enabling this HA:

```
    if (!SCMHAUtils.isSCMHAEnabled(conf)) {
      // When SCM HA is not enabled, the current SCM is always the leader.
      return Optional.of((long)0);
    }
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
```diff
@@ -100,9 +101,9 @@ private static void setRaftRpcProperties(final RaftProperties properties,
     Rpc.setRequestTimeout(properties, TimeDuration.valueOf(
         conf.getRatisRequestTimeout(), TimeUnit.MILLISECONDS));
     Rpc.setTimeoutMin(properties, TimeDuration.valueOf(
-        conf.getRatisRequestMinTimeout(), TimeUnit.MILLISECONDS));
```
amaliujia, line 103: I think you don't need to update these parameters. The `cluster.waitForClusterToBeReady` is supposed to wait for SCM become leader thus solve the not leader exception.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/TestUtils.java
amaliujia: So currently the `cluster.waitForClusterToBeReady` returns quickly even the SCM has not become leader yet. 

I am guessing CI can still pass become you also have changed the rpc timeout parameters. So either the cluster wait for ready or the parameter change can solve the `not leader problem`. 

I am not if we need both. Maybe we only need to keep `cluster.waitForClusterToBeReady`?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelineManagerV2Impl.java
```diff
@@ -181,126 +181,73 @@ public Pipeline createPipeline(ReplicationType type, ReplicationFactor factor,
...
 
   @Override
   public List<Pipeline> getPipelines(
       ReplicationType type, ReplicationFactor factor,
       Pipeline.PipelineState state, Collection<DatanodeDetails> excludeDns,
       Collection<PipelineID> excludePipelines) {
-    lock.readLock().lock();
-    try {
-      return stateManager
+    return stateManager
```
amaliujia, line 231: so the dead lock solution seems is removing the read write lock? Is there an explanation why we can remove locks here?

GlenGeng: I will create a jira for this issue, can give the explanation of this solution there.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
GlenGeng: You are right, we only need the wait leader in cluster.waitForClusterToBeReady. The problem here is, we've set the min/max timeout un-properly. I will create a separate jira for his fix.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/TestUtils.java
GlenGeng: Good catch!

PR #1739, opened 2020-12-26
HDDS-4624. Fix set configs from SCMHAConfiguration Object

amaliujia:

When setting SCMHAConfigration object to OzoneConfiguration `conf.setFromObject(scmhaConfiguration);`:  

```
java.lang.IllegalArgumentException: Attempt to get double field "org.apache.hadoop.hdds.scm.ha.SCMHAConfiguration.raftSegmentSize" with illegal data type conversion to long
```

In reflection, seems that a Double Field cannot be cast to Long at https://github.com/apache/ozone/blob/master/hadoop-hdds/config/src/main/java/org/apache/hadoop/hdds/conf/ConfigurationReflectionUtil.java#L247 

https://issues.apache.org/jira/browse/HDDS-4624

## How was this patch tested?

UT


amaliujia:
R: @nandakumar131 @GlenGeng @ChenSammi

GlenGeng:
+1
LGTM. Thanks for the fix!

PR #1740, opened 2020-12-28
HDDS-4616. Deleted TXID not found can be printed many times

sky76093016:

Delete LOG.debug().

https://issues.apache.org/jira/browse/HDDS-4616

## How was this patch tested?

Test pass.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImpl.java
@@ -130,15 +130,6 @@ public class DeletedBlockLogImpl
       try {
         DeletedBlocksTransaction block =
             scmMetadataStore.getDeletedBlocksTXTable().get(txID);
-        if (block == null) {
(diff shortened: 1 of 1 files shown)


ayushtkn (commented):
The log level is debug now, It shouldn't bother now.
HDDS-4356 changed it from `warn` to `debug`

sky76093016:
Regarding this issue, what is @lokeshj1703 's opinion?Thank you.

lokeshj1703:
@sky76093016 I made it debug in HDDS-4356 because it was being printed thousands of times during deletion. The idea of the jira was to investigate why the below code was getting executed. On receiving ACK from all datanodes SCM purges delete transaction from DB. If transaction is being resent after purge then this log can be printed. I wanted to understand the scenario because there might be a case where delete txn is being resent although it has been ACKed by the datanodes.

```
        if (block == null) {	
          if (LOG.isDebugEnabled()) {	
            // This can occur due to race condition between retry and old	
            // service task where old task removes the transaction and the new	
            // task is resending	
            LOG.debug("Deleted TXID {} not found.", txID);	
          }	
          continue;	
        }
```

adoroszlai:
/pending The idea of the jira was to investigate why the below code was getting executed.

PR #1742, opened 2020-12-29
HDDS-4628: min/max election timeout of SCMRatisServer is not set properly.

GlenGeng:

min/max election timeout of SCMRatisServer is not properly set.
Previously, it is `[ratisRequestTimeout - 1s, ratisRequestTimeout + 1s]`
It should be  `[ratisLeaderElectionTimeout, ratisLeaderElectionTimeout + 0.2s]`

https://issues.apache.org/jira/browse/HDDS-4628

## How was this patch tested?

CI


ChenSammi:
+1

PR #1745, opened 2020-12-30
HDDS-4632: Add term into SetNodeOperationalStateCommand

GlenGeng:

We need add term into SCMCommand to help DN distinguish SCMCommand from stale leader.
SetNodeOperationalStateCommand is a new added SCMCommand by Decommission, which misses the adding of the term.

https://issues.apache.org/jira/browse/HDDS-4632

## How was this patch tested?

CI


ChenSammi:
+1

PR #1746, opened 2020-12-30
HDDS-4631. Fix client set quota with 0.

captainzmc:

Currently setting Quota is invalid if set to 0. It's actually going to be set to -1.

`store.getVolume(volumeName).getBucket(bucketName).setQuota(`
`       OzoneQuota.parseQuota("0GB", 0));`
`  Assert.assertEquals(-1, bucket.getQuotaInBytes());`
`  Assert.assertEquals(-1, bucket.getQuotaInNamespace());`
This behavior is actually incorrect, and In general setting quota to 0 isn't a valid entry even in HDFS.  we can be consistent with HDFS on this part.

`java.lang.IllegalArgumentException: Invalid values for quota : 0 `

Because the original client quota type is long. If the user does not specify quota, the default value for this field is 0.
We cannot verify that the user did not specify quota or set quoat to 0, since both actions are passed to the backend to be 0.
So it would be more appropriate to change the long to a string. If the user does not specify quota, it is null. It's easier to judge.

https://issues.apache.org/jira/browse/HDDS-4631

## How was this patch tested?

ut added

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
@@ -201,8 +201,8 @@ public final class OzoneQuota {
           " value cannot be greater than Long.MAX_VALUE BYTES");
     }
 
-    if (nSize < 0) {
-      throw new IllegalArgumentException("Quota cannot be negative.");
+    if (nSize <= 0) {
+      throw new IllegalArgumentException("Invalid values for quota: " + nSize);
     }
 
     return new OzoneQuota(new RawQuotaInBytes(currUnit, nSize));
@@ -217,8 +217,12 @@ public final class OzoneQuota {
    * @return OzoneQuota object
    */
   public static OzoneQuota parseNameSpaceQuota(String quotaInNamespace) {
-    return new OzoneQuota(Long.parseLong(quotaInNamespace),
-        new RawQuotaInBytes(Units.BYTES, -1));
+    long nameSpaceQuota = Long.parseLong(quotaInNamespace);
+    if (nameSpaceQuota <= 0) {
+      throw new IllegalArgumentException(
+          "Invalid values for quota: " + nameSpaceQuota);
+    }
+    return new OzoneQuota(nameSpaceQuota, new RawQuotaInBytes(Units.BYTES, -1));
   }
 
   /**
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
@@ -18,6 +18,7 @@
 
 package org.apache.hadoop.hdds.client;
 
+import com.google.common.base.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -172,6 +173,11 @@ public final class OzoneQuota {
    */
   public static OzoneQuota parseSpaceQuota(String quotaInBytes) {
 
+    if (Strings.isNullOrEmpty(quotaInBytes)) {
+      throw new IllegalArgumentException(
+          "Quota string cannot be null or empty.");
+    }
+
     String uppercase = quotaInBytes.toUpperCase()
         .replaceAll("\\s+", "");
     String size = "";
@@ -217,6 +223,10 @@ public final class OzoneQuota {
    * @return OzoneQuota object
    */
   public static OzoneQuota parseNameSpaceQuota(String quotaInNamespace) {
+    if (Strings.isNullOrEmpty(quotaInNamespace)) {
+      throw new IllegalArgumentException(
+          "Quota string cannot be null or empty.");
+    }
     long nameSpaceQuota = Long.parseLong(quotaInNamespace);
     if (nameSpaceQuota <= 0) {
       throw new IllegalArgumentException(
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
@@ -144,7 +144,7 @@ public final class OzoneQuota {
   /**
    * Constructor for Ozone Quota.
    *
-   * @param quotaInNamespace Volume quota in counts
+   * @param quotaInNamespace ozone quota in counts
    * @param rawQuotaInBytes RawQuotaInBytes value
    */
   private OzoneQuota(long quotaInNamespace, RawQuotaInBytes rawQuotaInBytes) {
@@ -167,7 +167,7 @@ public final class OzoneQuota {
    * Parses a user provided string space quota and returns the
    * Quota Object.
    *
-   * @param quotaInBytes Volume quota in bytes
+   * @param quotaInBytes ozone quota in bytes
    *
    * @return OzoneQuota object
    */
@@ -219,7 +219,7 @@ public final class OzoneQuota {
    * Parses a user provided string Namespace quota and returns the
    * Quota Object.
    *
-   * @param quotaInNamespace Volume quota in counts
+   * @param quotaInNamespace ozone quota in counts
    *
    * @return OzoneQuota object
    */
@@ -240,8 +240,8 @@ public final class OzoneQuota {
    * Parses a user provided string and returns the
    * Quota Object.
    *
-   * @param quotaInBytes Volume quota in bytes
-   * @param quotaInNamespace Volume quota in counts
+   * @param quotaInBytes ozone quota in bytes
+   * @param quotaInNamespace ozone quota in counts
    *
    * @return OzoneQuota object
    */
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
@@ -18,7 +18,6 @@
 
 package org.apache.hadoop.hdds.client;
 
-import com.google.common.base.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -123,6 +122,25 @@ public final class OzoneQuota {
   }
 
   /**
+   * Constructor for Ozone Space Quota.
+   *
+   * @param rawQuotaInBytes RawQuotaInBytes value
+   */
+  private OzoneQuota(RawQuotaInBytes rawQuotaInBytes) {
+    this.rawQuotaInBytes = rawQuotaInBytes;
+    this.quotaInBytes = rawQuotaInBytes.sizeInBytes();
+  }
+
+  /**
+   * Constructor for Ozone NameSpace Quota.
+   *
+   * @param quotaInNamespace long value
+   */
+  private OzoneQuota(long quotaInNamespace) {
+    this.quotaInNamespace = quotaInNamespace;
+  }
+
+  /**
    * Constructor for Ozone Quota.
    *
    * @param quotaInNamespace Volume quota in counts
@@ -145,21 +163,14 @@ public final class OzoneQuota {
   }
 
   /**
-   * Parses a user provided string and returns the
+   * Parses a user provided string space quota and returns the
    * Quota Object.
    *
    * @param quotaInBytes Volume quota in bytes
-   * @param quotaInNamespace Volume quota in counts
    *
    * @return OzoneQuota object
    */
-  public static OzoneQuota parseQuota(String quotaInBytes,
-      long quotaInNamespace) {
-
-    if (Strings.isNullOrEmpty(quotaInBytes)) {
-      throw new IllegalArgumentException(
-          "Quota string cannot be null or empty.");
-    }
+  public static OzoneQuota parseSpaceQuota(String quotaInBytes) {
 
     String uppercase = quotaInBytes.toUpperCase()
         .replaceAll("\\s+", "");
@@ -194,10 +205,36 @@ public final class OzoneQuota {
       throw new IllegalArgumentException("Quota cannot be negative.");
     }
 
-    return new OzoneQuota(quotaInNamespace,
-        new RawQuotaInBytes(currUnit, nSize));
+    return new OzoneQuota(new RawQuotaInBytes(currUnit, nSize));
   }
 
+  /**
+   * Parses a user provided string Namespace quota and returns the
+   * Quota Object.
+   *
+   * @param quotaInNamespace Volume quota in counts
+   *
+   * @return OzoneQuota object
+   */
+  public static OzoneQuota parseNameSpaceQuota(String quotaInNamespace) {
+    return new OzoneQuota(Long.parseLong(quotaInNamespace),
+        new RawQuotaInBytes(Units.BYTES, -1));
+  }
+
+  /**
+   * Parses a user provided string and returns the
+   * Quota Object.
+   *
+   * @param quotaInBytes Volume quota in bytes
+   * @param quotaInNamespace Volume quota in counts
+   *
+   * @return OzoneQuota object
+   */
+  public static OzoneQuota parseQuota(String quotaInBytes,
+      String quotaInNamespace) {
+    return new OzoneQuota(parseNameSpaceQuota(quotaInNamespace)
+        .quotaInNamespace, parseSpaceQuota(quotaInBytes).rawQuotaInBytes);
+  }
 
   /**
    * Returns OzoneQuota corresponding to size in bytes.
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
@@ -208,7 +208,8 @@ public final class OzoneQuota {
     }
 
     if (nSize <= 0) {
-      throw new IllegalArgumentException("Invalid values for quota: " + nSize);
+      throw new IllegalArgumentException("Invalid values for space quota: "
+          + nSize);
     }
 
     return new OzoneQuota(new RawQuotaInBytes(currUnit, nSize));
@@ -230,7 +231,7 @@ public final class OzoneQuota {
     long nameSpaceQuota = Long.parseLong(quotaInNamespace);
     if (nameSpaceQuota <= 0) {
       throw new IllegalArgumentException(
-          "Invalid values for quota: " + nameSpaceQuota);
+          "Invalid values for namespace quota: " + nameSpaceQuota);
     }
     return new OzoneQuota(nameSpaceQuota, new RawQuotaInBytes(Units.BYTES, -1));
   }
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
@@ -239,7 +239,7 @@ public class OMBucketSetPropertyRequest extends OMClientRequest {
           OMException.ResultCodes.QUOTA_ERROR);
     }
 
-    if (quotaInBytes < OzoneConsts.QUOTA_RESET) {
+    if (quotaInBytes < OzoneConsts.QUOTA_RESET || quotaInBytes == 0) {
       return false;
     }
 
@@ -273,7 +273,7 @@ public class OMBucketSetPropertyRequest extends OMClientRequest {
       OmBucketArgs omBucketArgs) {
     long quotaInNamespace = omBucketArgs.getQuotaInNamespace();
 
-    if (quotaInNamespace < OzoneConsts.QUOTA_RESET) {
+    if (quotaInNamespace < OzoneConsts.QUOTA_RESET || quotaInNamespace == 0) {
       return false;
     }
     return true;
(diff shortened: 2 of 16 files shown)


ayushtkn (commented):
In general setting quota to 0 isn't a valid entry even in HDFS.
It throws an exception 
```
java.lang.IllegalArgumentException: Invalid values for quota : 0 and 0

	at org.apache.hadoop.hdfs.DFSClient.setQuota(DFSClient.java:2585)
	at org.apache.hadoop.hdfs.DistributedFileSystem$22.doCall(DistributedFileSystem.java:1055)
	at org.apache.hadoop.hdfs.DistributedFileSystem$22.doCall(DistributedFileSystem.java:1052)
	at org.apache.hadoop.fs.FileSystemLinkResolver.resolve(FileSystemLinkResolver.java:81)
```

captainzmc:
> In general setting quota to 0 isn't a valid entry even in HDFS.
> It throws an exception

@ayushtkn Thanks for the tip. I think we can be consistent with HDFS on this part. Later I will commit a new commit to fix it

captainzmc:
Fixed review issues. cc @ayushtkn

linyiqun:
@captainzmc , If setting quota to 0 is not allowed now, can we rollback unrelated change in this PR?

linyiqun (commented):
@captainzmc , If setting quota to 0 is not allowed now, can we rollback unrelated change in this PR?

captainzmc:
> @captainzmc , If setting quota to 0 is not allowed now, can we rollback unrelated change in this PR?

@linyiqun Thanks for the tip.
I think currently all changes to this PR are required。
First we need to be able to distinguish whether the user sets quota to 0 (users may not specify Quoat when creating volumes and buckets, long is undefined as 0 by default and should be successful), and then judge the validity based on this. 
Therefore, we need to change quotaInNamespace from long to String. In addition, it is also possible for users to specify either quotaInBytes or quotaInNamespace separately. In addition, many cases of original UT are not covered, so I made a supplement separately.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -190,14 +201,44 @@ public static OzoneQuota parseQuota(String quotaInBytes,
...
+   * Parses a user provided string Namespace quota and returns the
+   * Quota Object.
+   *
+   * @param quotaInNamespace Volume quota in counts
+   *
+   * @return OzoneQuota object
+   */
+  public static OzoneQuota parseNameSpaceQuota(String quotaInNamespace) {
+    long nameSpaceQuota = Long.parseLong(quotaInNamespace);
+    if (nameSpaceQuota <= 0) {
```
amaliujia, line 221: Question: how to set QUOTA.RESET? (basically how to disable quota by set the quota). In the past I guess we can set -2 but now it seems disallowed.

For example, is there a case that a bucket is enabled quota, later users decide to disable the quota by setting a quota value?

captainzmc: If the user specifies quota when creating bucket and volume,the default is -1, indicating that quota is not enabled.

A quota value of -2 is a special case, because older clusters such as previous versions of Ozone1.0, volume Info and bucket Info do not have quotaInBytes and quotaInNamespace in them. After such a cluster is upgraded to Ozone1.1, these fields become to 0 when getting volume Info and bucket Info because they didn't exist before, so in HDDS-4562 proto is set to -2 by default, indicating that these volumes and buckets are old, and their quota is also not enabled. Currently, it is not recommended to enable quota for these volumes and buckets because their usedbyte and usedNamespace are incorrect. So, -2 is more used to distinguish between old volumes and buckets. Later if we can support recalculation of usedbyte and usedNamespace. This lets us know which volumes and buckets need to be recalculated.

To disable quota, use the clrquota method instead of setquota.

linyiqun:
{quote}
First we need to be able to distinguish whether the user sets quota to 0 (users may not specify Quoat when creating volumes and buckets, long is undefined as 0 by default and should be successful), and then judge the validity based on this.
{quote}
I don't fully get why we need to distinguish whether the user sets quota to 0. If the long undefined value is 0, we could return a Quota.RESET value instead of while get the quota value for this case. I don't think this is a problem here. And for other cases, we don't allow 0 could be set by cli command after above change.

ayushtkn:
I am even not sure about converting to string type. If the user hasn't specified, Can't we send specific constant like QUOTA_DONT_SET(Something like this is there in HdfsConstants) to distinguish. We can even just validate at client side and leave the server side as is, Like don't bother with 0. Might help for compatibility reasons.
The present aim is like we don't support 0, it is not taking affect also. We just need to tell the client it won't work and save an RPC, if he tends to shoot 0.

captainzmc:
Hi @linyiqun @ayushtkn Sorry for the late reply, we can continue to discuss this problem.
Maybe I didn't make it clear before, let's do an example：
ozone sh bucket create vol1 （This should be success）
ozone sh bucket create vol1 --space-quota 0TB --namespace-quota 0 （This should be unsuccessful）

 These are two different cases. But in both cases in [CreateVolumeHandler quotaOptions.getQuotaInNamespace()](https://github.com/apache/ozone/blob/master/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/volume/CreateVolumeHandler.java#L68) is 0.
image

Different cases  receives the same parameters. Because the long variable is also 0 by default when it is not declared.

Let's look at another example，this also has the same problem：
 ozone sh volume setquota vol1 --space-quota 10TB （This should be success）
 ozone sh volume setquota vol1 --space-quota 10TB --namespace-quota 0 （This should be unsuccessful）

Because of this, I changed long to a string.

linyiqun:
>I am even not sure about converting to string type. If the user hasn't specified, Can't we send specific constant like QUOTA_DONT_SET(Something like this is there in HdfsConstants) to distinguish.

Suppose current logic already implement this logic(see below assert), if user hasn't specified quota. 0 means it comes from the long type default value as user.
```java
    store.getVolume(volumeName).setQuota(OzoneQuota.parseQuota(
        "0GB", 0L));
    volume = store.getVolume(volumeName);
    Assert.assertEquals(OzoneConsts.QUOTA_RESET, volume.getQuotaInBytes());
    Assert.assertEquals(OzoneConsts.QUOTA_RESET, volume.getQuotaInNamespace());
```

>Let's look at another example，this also has the same problem：
ozone sh volume setquota vol1 --space-quota 10TB （This should be success）
ozone sh volume setquota vol1 --space-quota 10TB --namespace-quota 0 （This should be unsuccessful）

@captainzmc , this will be not a problem. For first case that don't specified namespace quota, now the real meaning is the quota unset not means namepsace quota was 0. 

So I prefer only to make the adjustment that don't allow 0 can be set in this PR change.

captainzmc:
> > Let's look at another example，this also has the same problem：
> > ozone sh volume setquota vol1 --space-quota 10TB （This should be success）
> > ozone sh volume setquota vol1 --space-quota 10TB --namespace-quota 0 （This should be unsuccessful）
> 
> @captainzmc , this will be not a problem. 

We can reproduce the case I mentioned above using UT in the current master code. Add the following test code in UT：
image
Set the breakpoint in SetQuotaHandler. Execute the test. We can see that although namespace quota was not specified, 0 was passed in anyway. Because when this field is not defined get the default value (0) of long.
image

The first case can also reproduce with the testShQuota UT, adding a breakpoint at CreateVolumeHandler. Execute the test. 
image

> For first case that don't specified namespace quota, now the real meaning is the quota unset not means namepsace quota was 0.
> So I prefer only to make the adjustment that don't allow 0 can be set in this PR change.

I agree that we need to distinguish between these two different behaviors of the user. ”Don't allow 0 can be set“ It's our purpose，I also agree this. The point is, how do we do that?
We need to distinguish whether this '0' was set by the user actively or user didn't set it so that the code uses the default value of long. 
At the moment, I can't think of anything better than replacing a long with a string (which is easier to judge because the default value of a String is null not 0). Is any advice here?

linyiqun:
>I agree that we need to distinguish between these two different behaviors of the user. ”Don't allow 0 can be set“ It's our purpose，also agree this. The point is, how do we do that?

@captainzmc , here I mean don't allow 0 can be set in quota from the client side. So we can just add check in OzoneQuota#parseQuota and RpcClient#verifyCountsQuota/verifySpaceQuota. Does this make sense to you?

captainzmc:
> I see OmVolumeArgs resets the quotaInBytes/quotaInNamespace to OzoneConsts.QUOTA_RESET in its builder function. We can also make this change in OmBucketArgs# builder constructor. In addition, we should also reset quota when 0 was passed in constructor method.

Good advice. I found there are some bugs with creating volumes and buckets when quota is specified to 0 or not. This proposal can fix these problems. I will fix it in the new Commit.

The previous Commit had add check in OzoneQuota#parseQuota and RpcClient#verifyCountsQuota/verifySpaceQuota.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -190,14 +201,44 @@ public static OzoneQuota parseQuota(String quotaInBytes,
...
+   * Parses a user provided string and returns the
+   * Quota Object.
+   *
+   * @param quotaInBytes Volume quota in bytes
+   * @param quotaInNamespace Volume quota in counts
+   *
+   * @return OzoneQuota object
+   */
+  public static OzoneQuota parseQuota(String quotaInBytes,
+      String quotaInNamespace) {
```
linyiqun, line 238: @captainzmc ， can we rollback long to string convert for namspace quota now? As default quota 0 will all be converted to Quota.RESET based on above approach that we discussed.

captainzmc: The above advice only solves the problem of creating a Volume or Bucket.
The string quotainNamespace in parseQuota is still useful for verifying whether the user passed quotainNamespace.
Let's look at this case:
If the original volume space and namespace quoat were 100 and 10T, respectively, we reset the space quota separately.
ozone sh volume setquota vol1 --space-quota 20TB
OzoneConsts.QUOTA_RESET is the default quota for VolumeArgs. The space and namespace quota for vol1 will change to -1 and 20T, which is not expected.

I think it makes more sense to handle these cases in Ozonequota #parseQuota. Also, spaceQuota is currently a String in parseQuota, which makes our client-side processing logic more consistent.

ChenSammi: Hi @linyiqun , thanks for review the code.  We found bugs in production cluster with the previous implementation which doesn't distinguish the cases that user doesn't set the quota and user set the quota to 0.  So it's better to make it clear to avoid bugs caused by concept vague.  

In summary, for quota in bucket and volume,  
1) quota value will be -2  to indicates it's 1.0.0 bucket/volume 
2) quota value will be -1 if  bucket/volume created without explicitly set quota value in 1.1.0
3) quota > 0 indicates quota is enabled
4) throw exception if user set quota to 0 to keep consistent with current HDFS behavior

captainzmc:
Is any review update here？ @linyiqun @amaliujia @ChenSammi

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -145,21 +163,14 @@ public static String formatQuota(OzoneQuota quota) {
...
    *
    * @return OzoneQuota object
    */
-  public static OzoneQuota parseQuota(String quotaInBytes,
-      long quotaInNamespace) {
-
-    if (Strings.isNullOrEmpty(quotaInBytes)) {
-      throw new IllegalArgumentException(
-          "Quota string cannot be null or empty.");
-    }
```
ChenSammi, line 162: Shall we keep this null check？

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/volume/CreateVolumeHandler.java
```diff
@@ -59,13 +61,19 @@ protected void execute(OzoneClient client, OzoneAddress address)
...
         .setOwner(ownerName);
-    if (quotaOptions.getQuotaInBytes() != null) {
-      volumeArgsBuilder.setQuotaInBytes(OzoneQuota.parseQuota(
-          quotaOptions.getQuotaInBytes(),
-          quotaOptions.getQuotaInNamespace()).getQuotaInBytes());
+    if (!Strings.isNullOrEmpty(quotaOptions.getQuotaInBytes())) {
+      volumeArgsBuilder.setQuotaInBytes(OzoneQuota.parseSpaceQuota(
+          quotaOptions.getQuotaInBytes()).getQuotaInBytes());
+    } else {
+      volumeArgsBuilder.setQuotaInBytes(OzoneConsts.QUOTA_RESET);
```
ChenSammi, line 68: With default quotaInBytes and quotaInNamespace value, do we still need this and below else statement?

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/bucket/CreateBucketHandler.java
```diff
@@ -80,12 +81,19 @@ public void execute(OzoneClient client, OzoneAddress address)
...
     }
 
-    if (quotaOptions.getQuotaInBytes() != null) {
-      bb.setQuotaInBytes(OzoneQuota.parseQuota(quotaOptions.getQuotaInBytes(),
-          quotaOptions.getQuotaInNamespace()).getQuotaInBytes());
+    if (!Strings.isNullOrEmpty(quotaOptions.getQuotaInBytes())) {
+      bb.setQuotaInBytes(OzoneQuota.parseSpaceQuota(
+          quotaOptions.getQuotaInBytes()).getQuotaInBytes());
+    } else {
+      bb.setQuotaInBytes(OzoneConsts.QUOTA_RESET);
```
ChenSammi, line 88: Same as below comment.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeSetQuotaRequest.java
```diff
@@ -189,7 +189,7 @@ public boolean checkQuotaBytesValid(OMMetadataManager metadataManager,
       long volumeQuotaInBytes, String volumeName) throws IOException {
     long totalBucketQuota = 0;
 
-    if (volumeQuotaInBytes == 0) {
+    if (volumeQuotaInBytes < OzoneConsts.QUOTA_RESET) {
```
ChenSammi, line 192: (volumeQuotaInBytes < OzoneConsts.QUOTA_RESET || volumeQuotaInBytes == 0) ?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeSetQuotaRequest.java
```diff
@@ -213,8 +213,7 @@ public boolean checkQuotaBytesValid(OMMetadataManager metadataManager,
 
   public boolean checkQuotaNamespaceValid(long quotaInNamespace) {
 
-    if ((quotaInNamespace <= 0
-         && quotaInNamespace != OzoneConsts.QUOTA_RESET)) {
+    if (quotaInNamespace < OzoneConsts.QUOTA_RESET) {
```
ChenSammi, line 216: Same question as as above comment.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/volume/CreateVolumeHandler.java
captainzmc: This part is superfluous and can be deleted

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
linyiqun: Comments make sense to me, @captainzmc , @ChenSammi

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -190,14 +207,48 @@ public static OzoneQuota parseQuota(String quotaInBytes,
           " value cannot be greater than Long.MAX_VALUE BYTES");
     }
 
-    if (nSize < 0) {
-      throw new IllegalArgumentException("Quota cannot be negative.");
+    if (nSize <= 0) {
+      throw new IllegalArgumentException("Invalid values for quota: " + nSize);
```
linyiqun, line 211: Can you update quota to 'space quota' that will be more accurate for this exception?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -190,14 +207,48 @@ public static OzoneQuota parseQuota(String quotaInBytes,
...
+   */
+  public static OzoneQuota parseNameSpaceQuota(String quotaInNamespace) {
+    if (Strings.isNullOrEmpty(quotaInNamespace)) {
+      throw new IllegalArgumentException(
+          "Quota string cannot be null or empty.");
+    }
+    long nameSpaceQuota = Long.parseLong(quotaInNamespace);
+    if (nameSpaceQuota <= 0) {
+      throw new IllegalArgumentException(
+          "Invalid values for quota: " + nameSpaceQuota);
```
linyiqun, line 233: Similar comment: quota to namespace quota.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
```diff
@@ -239,7 +239,7 @@ public boolean checkQuotaBytesValid(OMMetadataManager metadataManager,
           OMException.ResultCodes.QUOTA_ERROR);
     }
 
-    if (quotaInBytes == 0) {
+    if (quotaInBytes < OzoneConsts.QUOTA_RESET) {
```
linyiqun, line 242: Seems this should be update to quotaInBytes < OzoneConsts.QUOTA_RESET || quotaInBytes == 0 as well.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/bucket/OMBucketSetPropertyRequest.java
```diff
@@ -273,8 +273,7 @@ public boolean checkQuotaNamespaceValid(OmVolumeArgs omVolumeArgs,
       OmBucketArgs omBucketArgs) {
     long quotaInNamespace = omBucketArgs.getQuotaInNamespace();
 
-    if ((quotaInNamespace <= 0
-         && quotaInNamespace != OzoneConsts.QUOTA_RESET)) {
+    if (quotaInNamespace < OzoneConsts.QUOTA_RESET) {
```
linyiqun, line 276: Same comment for this.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -145,16 +164,14 @@ public static String formatQuota(OzoneQuota quota) {
   }
 
   /**
-   * Parses a user provided string and returns the
+   * Parses a user provided string space quota and returns the
    * Quota Object.
    *
    * @param quotaInBytes Volume quota in bytes
```
linyiqun, line 170: volume quota should be ozone quota since this quota can also the bucket quota

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -190,14 +207,49 @@ public static OzoneQuota parseQuota(String quotaInBytes,
...
-    return new OzoneQuota(quotaInNamespace,
-        new RawQuotaInBytes(currUnit, nSize));
+    return new OzoneQuota(new RawQuotaInBytes(currUnit, nSize));
   }
 
+  /**
+   * Parses a user provided string Namespace quota and returns the
+   * Quota Object.
+   *
+   * @param quotaInNamespace Volume quota in counts
```
linyiqun, line 222: Same comment as above.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -190,14 +207,49 @@ public static OzoneQuota parseQuota(String quotaInBytes,
...
+    }
+    return new OzoneQuota(nameSpaceQuota, new RawQuotaInBytes(Units.BYTES, -1));
+  }
+
+  /**
+   * Parses a user provided string and returns the
+   * Quota Object.
+   *
+   * @param quotaInBytes Volume quota in bytes
+   * @param quotaInNamespace Volume quota in counts
```
linyiqun, line 244: Same comment as above.

linyiqun (commented):
PR looks great now, catch some nits below, @captainzmc .
Others looks to me.

linyiqun (approved):
LGTM, +1.
Thanks @captainzmc  for addressing the review comments.

captainzmc:
Fixed review issues.  cc @ChenSammi 
CI runs normally on my [personal branch](https://github.com/captainzmc/hadoop-ozone/runs/1707185948). Failed UT is irrelevant.

ChenSammi:
+ 1.

PR #1751, opened 2021-01-04
HDDS-4639. Safemode wait may end without checking

adoroszlai:

`ozone admin safemode wait` may conclude without actually checking status, if it takes more time to create SCM client than the timeout specified.  This change makes sure at least one check is performed after creating SCM client.

https://issues.apache.org/jira/browse/HDDS-4639

## How was this patch tested?

Added `sleep(5000)` in `createScmClient()`.  Executed with cluster already up:

```
$ time ozone admin safemode wait -t 1
SCM is out of safe mode.

real	0m8.302s
user	0m6.327s
sys	0m1.141s
```

and with only SCM running:

```
$ time ozone admin safemode wait -t 1
SCM is in safe mode. No more retries.
Safe mode is not ended within the timeout period.

real	0m8.568s
user	0m5.217s
sys	0m1.047s
```


elek (approved):
+1 thanks the patch @adoroszlai 

Just read the description in the Jira, nice catch.

adoroszlai:

PR #1650, opened 2020-12-02
HDDS-3684. Add tests for replication annotation

amaliujia:
## What changes were proposed in this pull request?

Add tests for replication annotation

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-3684

## How was this patch tested?

Unit Test

PR #1756, opened 2021-01-05
HDDS-4643. Ratis Snapshot should be loaded from the config

amaliujia:

Right now Ratis snapshot auto trigger threshold is a fixed number. We should load it from the HA configuration. 

https://issues.apache.org/jira/browse/HDDS-4643

## How was this patch tested?

N/A

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
@@ -149,7 +149,8 @@ public final class RatisUtil {
   private static void setRaftSnapshotProperties(
       final RaftProperties properties, final SCMHAConfiguration conf) {
     Snapshot.setAutoTriggerEnabled(properties, true);
-    Snapshot.setAutoTriggerThreshold(properties, 400000);
+    Snapshot.setAutoTriggerThreshold(properties,
+        conf.getRatisSnapshotThreshold());
   }
 
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
@@ -120,7 +120,7 @@ public class SCMHAConfiguration {
       tags = {SCM, OZONE, HA, RATIS},
       description = "The threshold to trigger a Ratis taking snapshot " +
           "operation")
-  private long ratisSnapshotThreshold = 100000L;
+  private long ratisSnapshotThreshold = 1000L;
 
   @Config(key = "ratis.request.timeout",
       type = ConfigType.TIME,
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
@@ -116,10 +116,10 @@ public class SCMHAConfiguration {
 
   @Config(key = "ratis.snapshot.threshold",
       type = ConfigType.LONG,
-      defaultValue = "100000L",
+      defaultValue = "1000L",
       tags = {SCM, OZONE, HA, RATIS},
       description = "The threshold to trigger a Ratis taking snapshot " +
-          "operation")
+          "operation for SCM")
   private long ratisSnapshotThreshold = 1000L;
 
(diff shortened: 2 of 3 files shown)


amaliujia:
@runzhiwang @GlenGeng @ChenSammi @nandakumar131

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
```diff
@@ -149,7 +149,8 @@ private static void setRaftRetryCacheProperties(
   private static void setRaftSnapshotProperties(
       final RaftProperties properties, final SCMHAConfiguration conf) {
     Snapshot.setAutoTriggerEnabled(properties, true);
-    Snapshot.setAutoTriggerThreshold(properties, 400000);
+    Snapshot.setAutoTriggerThreshold(properties,
+        conf.getRatisSnapshotThreshold());
```
amaliujia, line 153: This is useful for both writing unit tests for snapshot functionality and allow users to set their own threshold

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/TestReplicationAnnotation.java
```diff
@@ -88,8 +88,7 @@ public void testReplicateAnnotationBasic() throws Throwable {
 
     try {
       proxy.addContainer(HddsProtos.ContainerInfoProto.getDefaultInstance());
-      // Should have seen a IOException.
-      Assert.fail();
+      Assert.fail("Cannot reach here: should have seen a IOException");
```
amaliujia, line 91: Have this line of change due to https://github.com/apache/ozone/pull/1650#discussion_r551137844

runzhiwang (approved):
LGTM

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
```diff
@@ -114,6 +114,14 @@
   )
   private int raftLogPurgeGap = 1000000;
 
+  @Config(key = "ratis.snapshot.threshold",
+      type = ConfigType.LONG,
+      defaultValue = "100000L",
+      tags = {SCM, OZONE, HA, RATIS},
+      description = "The threshold to trigger a Ratis taking snapshot " +
+          "operation")
+  private long ratisSnapshotThreshold = 100000L;
```
GlenGeng, line 123: 100000L might be too large for SCM, since the WriteBatch needs hold them in memory and the flush will block the apply of transactions. How about 500 or 1000 ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
```diff
@@ -114,6 +114,14 @@
   )
   private int raftLogPurgeGap = 1000000;
 
+  @Config(key = "ratis.snapshot.threshold",
+      type = ConfigType.LONG,
+      defaultValue = "100000L",
+      tags = {SCM, OZONE, HA, RATIS},
+      description = "The threshold to trigger a Ratis taking snapshot " +
+          "operation")
```
GlenGeng, line 122: operation -> operation for SCM.

GlenGeng (commented):
LGTM.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
amaliujia: make senses. Set as 1000 for now. This is a configurable value anyway.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
amaliujia: seems not a typo?

GlenGeng: Not a typo, the description indicates that this config is for SCM.

"The threshold to trigger a Ratis taking snapshot operation" -> "The threshold to trigger a Ratis taking snapshot operation for SCM StateMachine" 
BTW, please change the `defaultValue` too.

amaliujia:
@runzhiwang  CI has passed.

runzhiwang:
@amaliujia Thanks the patch. @GlenGeng Thanks for review. I have merged this patch.

PR #1762, opened 2021-01-06
HDDS-4648. DN should set the block token when Read StateMachine data …

xiaoyuyao:
## What is proposed in this PR?

DN should set the block token when Read StateMachine data for WriteChunk

## What is the link to the Apache JIRA?

https://issues.apache.org/jira/browse/HDDS-4648

## How was this patch tested?

Fix existing acceptance tests.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java
@@ -573,16 +573,19 @@ public class ContainerStateMachine extends BaseStateMachine {
         ReadChunkRequestProto.newBuilder()
             .setBlockID(writeChunkRequestProto.getBlockID())
             .setChunkData(chunkInfo);
-    ContainerCommandRequestProto dataContainerCommandProto =
+    ContainerCommandRequestProto.Builder dataContainerCommandProtoBuilder =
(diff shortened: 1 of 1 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java
```diff
@@ -573,16 +573,19 @@ private ByteString readStateMachineData(
...
             .setBlockID(writeChunkRequestProto.getBlockID())
             .setChunkData(chunkInfo);
-    ContainerCommandRequestProto dataContainerCommandProto =
+    ContainerCommandRequestProto.Builder dataContainerCommandProtoBuilder =
         ContainerCommandRequestProto.newBuilder(requestProto)
-            .setCmdType(Type.ReadChunk).setReadChunk(readChunkRequestProto)
-            .build();
+            .setCmdType(Type.ReadChunk).setReadChunk(readChunkRequestProto);
+    if (requestProto.hasEncodedToken()) {
+      dataContainerCommandProtoBuilder.setEncodedToken(
```
adoroszlai, line 580: I think `newBuilder(requestProto)` sets all fields from `requestProto`, including encoded token, so this seems unnecessary.  Am I missing something?

xiaoyuyao: You are right, the constructor does copy the token if exist. I will close this one.

PR #1764, opened 2021-01-07
HDDS-2942. Putkey : create key table entries for intermediate directories in the key path

rakeshadr:

This task is to create entries in OM key table for every directory element occurring in the path.

https://issues.apache.org/jira/browse/HDDS-2942

## How was this patch tested?

Added UT cases.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
@@ -110,7 +110,8 @@ public class TestObjectStoreV1 {
     Assert.assertNotNull("Failed to find dir path: a/b/c", dirPathC);
 
     // after file creation
-    verifyKeyInFileTable(openKeyTable, file, dirPathC.getObjectID(), false);
+    verifyKeyInOpenFileTable(openKeyTable, file, dirPathC.getObjectID(),
+            false);
 
     ozoneOutputStream.write(data.getBytes(), 0, data.length());
     ozoneOutputStream.close();
@@ -118,13 +119,18 @@ public class TestObjectStoreV1 {
     Table<String, OmKeyInfo> keyTable =
             cluster.getOzoneManager().getMetadataManager().getKeyTable();
 
-    // after closing the file
+    // After closing the file. File entry should be removed from openFileTable
+    // and it should be added to fileTable.
     verifyKeyInFileTable(keyTable, file, dirPathC.getObjectID(), false);
+    verifyKeyInOpenFileTable(openKeyTable, file, dirPathC.getObjectID(),
+            true);
 
     ozoneBucket.deleteKey(key);
 
     // after key delete
     verifyKeyInFileTable(keyTable, file, dirPathC.getObjectID(), true);
+    verifyKeyInOpenFileTable(openKeyTable, file, dirPathC.getObjectID(),
+            true);
   }
 
   private OmDirectoryInfo getDirInfo(String volumeName, String bucketName,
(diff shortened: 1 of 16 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponseV1.java
```diff
@@ -32,12 +32,13 @@
...
 import java.util.List;
 
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.FILE_TABLE;
 import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.OPEN_FILE_TABLE;
 
 /**
  * Response for create file request layout version V1.
  */
-@CleanupTableInfo(cleanupTables = OPEN_FILE_TABLE)
+@CleanupTableInfo(cleanupTables = {FILE_TABLE, OPEN_FILE_TABLE})
```
linyiqun, line 41: The dir table and open file table is updated here, so the FILE_TABLE should be DIRECTORY_TABLE.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeyCreateResponseV1.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.annotation.Nonnull;
+import java.util.List;
+
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.FILE_TABLE;
+import static org.apache.hadoop.ozone.om.OmMetadataManagerImpl.OPEN_FILE_TABLE;
+
+/**
+ * Response for CreateKey request layout version V1.
+ */
+@CleanupTableInfo(cleanupTables = {OPEN_FILE_TABLE, FILE_TABLE})
```
linyiqun, line 37: Here should also be cleanupTables = {OPEN_FILE_TABLE, DIRECTORY_TABLE}

linyiqun (commented):
@rakeshadr , two minor comments from me, please have a check.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
```diff
@@ -0,0 +1,183 @@
...
+
+    ozoneOutputStream.write(data.getBytes(), 0, data.length());
+    ozoneOutputStream.close();
+
+    Table<String, OmKeyInfo> keyTable =
+            cluster.getOzoneManager().getMetadataManager().getKeyTable();
+
+    // after closing the file
+    verifyKeyInFileTable(keyTable, file, dirPathC.getObjectID(), false);
+
```
linyiqun, line 123: Can we additionally add key check for openKeyTable? Open key table should become empty after file is closed.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
```diff
@@ -0,0 +1,183 @@
...
+            = keyTable.iterator();
+
+    if (isEmpty) {
+      Assert.assertTrue("Table is not empty!", keyTable.isEmpty());
+    } else {
+      Assert.assertFalse("Table is empty!", keyTable.isEmpty());
+      while (iterator.hasNext()) {
+        Table.KeyValue<String, OmKeyInfo> next = iterator.next();
+        Assert.assertTrue("Invalid Key: " + next.getKey(),
+                next.getKey().startsWith(parentID + "/" + fileName));
```
linyiqun, line 162: Seems we could use the equals function to compare. There is only test file key in key table/open key table in this test.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/key/OMKeyCreateResponseV1.java
rakeshadr: Done!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/file/OMFileCreateResponseV1.java
rakeshadr: Done!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
rakeshadr: Sure, added.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
rakeshadr: Thanks a lot @linyiqun for the review comments. 

I've used startsWith because clientID is not visible to the test. The entry in openFileTable is <parentID>/filename/<clientID>. I have added comments and did equals separately for fileTable entries. Hope this would be fine for you.

linyiqun: Make sense to me.

linyiqun:
LGTM, +1. 
But would be better to check if failure tests is related, @rakeshadr .

linyiqun (approved):

rakeshadr:
> LGTM, +1.
> But would be better to check if failure tests is related, @rakeshadr .

Thanks @linyiqun , the test case failure is unrelated to the patch.

PR #1765, opened 2021-01-07
HDDS-4612. Create OMCancelPrepareRequest and Response to cancel the prepared state of an OM.

errose28:

Implement an OM request and response to cancel the prepared state of an ozone manager. Client protocol is added to allow integration testing. This will be connected to a client CLI command and acceptance tested in HDDS-4611.

HDDS-4612

## How was this patch tested?

Unit and integration tests added.


avijayanhwx (approved):
Thanks for working on this @errose28. Changes LGTM.

PR #1766, opened 2021-01-07
HDDS-4653. Support TDE for MPU Keys on Encrypted Buckets

bharatviswa504:

Support MPU keys in encrypted buckets.

Created a new OzoneInputStream - MultipartCryptoKeyInputStream for accessing MPU keys in encrypted buckets. This InputStream has a list of OzoneCryptoInputStreams, one for each part. OzoneCryptoInputStream wraps a CryptoInputStream (which in turn wraps a KeyInputStream) and provides an extra attribute of length (similar to LengthInputStream for KeyInputStream).

https://issues.apache.org/jira/browse/HDDS-4653

## How was this patch tested?

Added unit tests in TestOzoneAtRestEncryption.


hanishakoneru:
Thanks @bharatviswa504 for working on this. I have pushed an update to fix MPU in Encrypted bucket.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/MultipartCryptoKeyInputStream.java
```diff
@@ -0,0 +1,373 @@
...
+  // partOffsets[i] stores the index of the first data byte in
+  // partStream w.r.t the whole key data.
+  // For example, let’s say the part size is 200 bytes and part[0] stores
+  // data from indices 0 - 199, part[1] from indices 200 - 399 and so on.
+  // Then, partOffsets[0] = 0 (the offset of the first byte of data in
+  // part[0]), partOffsets[1] = 200 and so on.
+  private long[] partOffsets;
+
+  // Index of the partStream corresponding to the current position of the
+  // MultipartCryptoKeyInputStream i.e. offset of the data to be read next
```
xiaoyuyao, line 58: NIT: "offset of the data to be read next" is confusing and can be removed here.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/MultipartCryptoKeyInputStream.java
```diff
@@ -0,0 +1,373 @@
...
+  private int partIndex = 0;
+
+  // Tracks the partIndex corresponding to the last seeked position so that it
+  // can be reset if a new position is seeked.
+  private int partIndexOfPrevPos = 0;
```
xiaoyuyao, line 63: NIT: can re rename partIndexOfPrevPos=> prevPartIndex

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/MultipartCryptoKeyInputStream.java
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  @Override
+  public synchronized long getPos() throws IOException {
+    return length == 0 ? 0 : partOffsets[partIndex] +
+        partStreams.get(partIndex).getPos();
+  }
+
+  @Override
+  public boolean seekToNewSource(long targetPos) throws IOException {
```
xiaoyuyao, line 325: So seekToNewSource is not supported? Do we have use case requires this?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/MultipartCryptoKeyInputStream.java
```diff
@@ -0,0 +1,373 @@
...
+      partStreams.get(index).seek(0);
+    }
+    // 2. Seek the partStream to the adjusted position
+    partStreams.get(partIndex).seek(pos - partOffsets[partIndex]);
+    partIndexOfPrevPos = partIndex;
+  }
+
+  @Override
+  public synchronized long getPos() throws IOException {
+    return length == 0 ? 0 : partOffsets[partIndex] +
```
xiaoyuyao, line 320: Do we need to checkOpen() at the begining?

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/MultipartCryptoKeyInputStream.java
```diff
@@ -0,0 +1,373 @@
...
+   * to read more data than requested. The extra data will be filtered out
+   * before returning to the client.
+   */
+  private void adjustReadPosition(long cryptoBufferSize) throws IOException {
+    // Position of the buffer in current stream
+    long currentPosOfStream = partStreams.get(partIndex).getPos();
+    if (currentPosOfStream % cryptoBufferSize != 0) {
+      // Adjustment required.
+      // Update readPositionAdjustedBy and seek to the adjusted position
+      readPositionAdjustedBy = (int) (currentPosOfStream % cryptoBufferSize);
```
xiaoyuyao, line 232: NIT: readPositionAdjustedBy = (int) (currentPosOfStream % cryptoBufferSize);
can be moved before if to save a % operation.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/MultipartCryptoKeyInputStream.java
hanishakoneru: Done.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/MultipartCryptoKeyInputStream.java
hanishakoneru: The underlying InputStreams also have this as false. 
If we do want to support this, then it would need to be changed in underlying streams also.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/MultipartCryptoKeyInputStream.java
hanishakoneru: The underlying stream would perform the check. But I will add it here as well.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
```diff
@@ -546,6 +546,39 @@ protected void generateRequiredEncryptionInfo(KeyArgs keyArgs,
...
+
+    String volumeName = keyArgs.getVolumeName();
+    String bucketName = keyArgs.getBucketName();
+
+    boolean acquireLock = false;
+    OMMetadataManager omMetadataManager = ozoneManager.getMetadataManager();
+
+    if (ozoneManager.getKmsProvider() != null) {
+      try {
+        acquireLock = omMetadataManager.getLock().acquireReadLock(
```
xiaoyuyao, line 561: Can we move the acquireReadLock outside try block?

bharatviswa504: Lock release is handled in the finally block.
Any reason to move this outside try block, just trying to understand the reason.

bharatviswa504:
Thank You @hanishakoneru for continuing the work on this PR and making it work.
Fixed find bug issues in the latest commit, test failure looks unrelated, will wait for another CI run.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyRequest.java
xiaoyuyao: The suggestion is given based on the JDK lock document examples: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/locks/ReentrantReadWriteLock.html

xiaoyuyao:
+1, thanks @hanishakoneru and @bharatviswa504  for the contribution. I will merge the PR shortly.

PR #1768, opened 2021-01-08
HDDS-4498. Enable OM Ratis by default.

hanishakoneru:

Currently, by default, Ratis is not enabled on an OM single node cluster. This Jira proposes to change the default single node OM to use Ratis server. 

OM Ratis has been tested extensively and is stable now. To convert a single node cluster to HA, the first step would be to upgrade the cluster to Ratis enabled cluster. 

A non-ratis single node cluster can be upgraded to  ratis-enabled single node cluster by just setting the ozone.om.ratis.enable config to true and restarting the OM.

https://issues.apache.org/jira/browse/HDDS-4498

## How was this patch tested?

Config default value change. Tests already exist for this.

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1523,7 +1523,7 @@
 
   <property>
     <name>ozone.om.ratis.enable</name>
-    <value>false</value>
+    <value>true</value>
     <tag>OZONE, OM, RATIS, MANAGEMENT</tag>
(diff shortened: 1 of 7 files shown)


arp7 (approved):

arp7 (approved):
+1

hanishakoneru:
Ignoring testTrash in TestOzoneFileSystem and TestRootedOzoneFileSystem. Filed HDDS-4669 to fix testTrash to work when OM Ratis is enabled. 
TestKeyInputStream and TestCloseContainerHandlingByClient pass locally.

PR #1774, opened 2021-01-11
HDDS-4667. BlockInputStream should give up read retry if pipeline is not updated

adoroszlai:

HDDS-4320 added the following logic in `BlockInputStream`:

https://github.com/apache/ozone/blob/77bf7ad7c1a70e9e1d7aaeb3ec006f19ae8d3089/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockInputStream.java#L287-L292

`handleReadError` tries to refresh the pipeline.  If it gets a new one, read is retried on that.  Otherwise it re-throws the exception to exit the loop.

https://github.com/apache/ozone/blob/77bf7ad7c1a70e9e1d7aaeb3ec006f19ae8d3089/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/BlockInputStream.java#L171-L173

Pipeline equality includes the pipeline ID, but SCM constructs a new on-demand pipeline for reading from a single node.  The result is that the client thinks there is a new pipeline and retries.

This PR proposes to skip retry if the pipeline has the same set of nodes that failed in the previous iteration.

https://issues.apache.org/jira/browse/HDDS-4667

## How was this patch tested?

Added unit test that reproduces the infinite loop without the fix.

### DIFF
--- a/hadoop-hdds/client/pom.xml
+++ b/hadoop-hdds/client/pom.xml
@@ -57,5 +57,10 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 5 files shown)


bshashikant:
Thanks @adoroszlai for working on this. I feel , we should have a retry policy configured to bail out after certain cycles of refreshing the pipeline.

Also, the refresh pipeline should only happen in case the all the dns fail to serve the read. In such cases, on the client we should see StorageContainerException on the client. Any exception on the grpc channel we should terminate without retrying..

bshashikant (approved):

adoroszlai:
Thanks @elek for reporting this problem and @bshashikant for reviewing and committing it.

PR #1776, opened 2021-01-12
HDDS-4289. Throw exception from hadoop2 filesystem jar in HA environment

symious:

- Throw more meaningful error messages.

[https://issues.apache.org/jira/browse/HDDS-4289](https://issues.apache.org/jira/browse/HDDS-4289)

## How was this patch tested?

manual test. The new error shows as follows.
```
2021-01-12 10:27:45,626 [main] ERROR client.OzoneClientFactory: Couldn't create RpcClient protocol exception:
java.io.IOException: Hadoop2 Filesystem Jar doesn't support OM-HA, please update to use Hadoop3 Filesystem Jar.
	at org.apache.hadoop.fs.ozone.Hadoop27RpcTransport.submitRequest(Hadoop27RpcTransport.java:76)
	at org.apache.hadoop.ozone.om.protocolPB.OzoneManagerProtocolClientSideTranslatorPB.submitRequest(OzoneManagerProtocolClientSideTranslatorPB.java:218)
	at org.apache.hadoop.ozone.om.protocolPB.OzoneManagerProtocolClientSideTranslatorPB.getServiceInfo(OzoneManagerProtocolClientSideTranslatorPB.java:1066)
	at org.apache.hadoop.ozone.client.rpc.RpcClient.<init>(RpcClient.java:170)
	at org.apache.hadoop.ozone.client.OzoneClientFactory.getClientProtocol(OzoneClientFactory.java:246)
	at org.apache.hadoop.ozone.client.OzoneClientFactory.getRpcClient(OzoneClientFactory.java:113)
	at org.apache.hadoop.fs.ozone.BasicOzoneClientAdapterImpl.<init>(BasicOzoneClientAdapterImpl.java:149)
	at org.apache.hadoop.fs.ozone.BasicOzoneFileSystem.createAdapter(BasicOzoneFileSystem.java:186)
	at org.apache.hadoop.fs.ozone.BasicOzoneFileSystem.initialize(BasicOzoneFileSystem.java:163)
	at org.apache.hadoop.fs.FileSystem.createFileSystem(FileSystem.java:3260)
	at org.apache.hadoop.fs.FileSystem.access$200(FileSystem.java:121)
	at org.apache.hadoop.fs.FileSystem$Cache.getInternal(FileSystem.java:3309)
	at org.apache.hadoop.fs.FileSystem$Cache.get(FileSystem.java:3277)
	at org.apache.hadoop.fs.FileSystem.get(FileSystem.java:475)
	at org.apache.hadoop.fs.Path.getFileSystem(Path.java:356)
	at org.apache.hadoop.fs.shell.PathData.expandAsGlob(PathData.java:325)
	at org.apache.hadoop.fs.shell.Command.expandArgument(Command.java:245)
	at org.apache.hadoop.fs.shell.Command.expandArguments(Command.java:228)
	at org.apache.hadoop.fs.shell.FsCommand.processRawArguments(FsCommand.java:103)
	at org.apache.hadoop.fs.shell.Command.run(Command.java:175)
	at org.apache.hadoop.fs.FsShell.run(FsShell.java:317)
	at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:76)
	at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:90)
	at org.apache.hadoop.fs.FsShell.main(FsShell.java:380)
ls: Couldn't create RpcClient protocol
```

### DIFF
--- a/hadoop-ozone/ozonefs-hadoop2/src/main/java/org/apache/hadoop/fs/ozone/Hadoop27RpcTransport.java
+++ b/hadoop-ozone/ozonefs-hadoop2/src/main/java/org/apache/hadoop/fs/ozone/Hadoop27RpcTransport.java
@@ -28,6 +28,7 @@ import org.apache.hadoop.ipc.ProtobufRpcEngine;
 import org.apache.hadoop.ipc.RPC;
 import org.apache.hadoop.net.NetUtils;
 import org.apache.hadoop.ozone.OmUtils;
+import org.apache.hadoop.ozone.om.OMConfigKeys;
 import org.apache.hadoop.ozone.om.protocolPB.OmTransport;
 import org.apache.hadoop.ozone.om.protocolPB.OzoneManagerProtocolPB;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos.OMRequest;
@@ -48,6 +49,11 @@ public class Hadoop27RpcTransport implements OmTransport {
 
   public Hadoop27RpcTransport(
       ConfigurationSource conf) throws IOException {
+    // Hadoop27 doesn't support HA-OM
+    if (isHAOMConfig(conf)) {
+      throw new IOException("HA-OM config detected. Hadoop2 Filesystem jar " +
+          "doesn't support HA-OM, please use Hadoop3 Filesystem jar.");
+    }
     InetSocketAddress socket = OmUtils.getOmAddressForClients(conf);
     long version = RPC.getProtocolVersion(OzoneManagerProtocolPB.class);
     OzoneConfiguration ozoneConfiguration = OzoneConfiguration.of(conf);
@@ -83,5 +89,10 @@ public class Hadoop27RpcTransport implements OmTransport {
   private int getRpcTimeout(OzoneConfiguration conf) {
     return conf.getInt(CommonConfigurationKeys.IPC_CLIENT_RPC_TIMEOUT_KEY,
         CommonConfigurationKeys.IPC_CLIENT_RPC_TIMEOUT_DEFAULT);
+  }
+
+  private boolean isHAOMConfig(ConfigurationSource confSource) {
+    return OmUtils.isServiceIdsDefined(confSource) &&
+        confSource.getTrimmed(OMConfigKeys.OZONE_OM_ADDRESS_KEY) == null;
   }
 }


hadoop-ozone/ozonefs-hadoop2/src/main/java/org/apache/hadoop/fs/ozone/Hadoop27RpcTransport.java
```diff
@@ -67,6 +69,12 @@ public OMResponse submitRequest(OMRequest payload) throws IOException {
     try {
       return proxy.submitRequest(NULL_RPC_CONTROLLER, payload);
     } catch (ServiceException e) {
+      OMNotLeaderException notLeaderException =
```
elek, line 72: Not sure about this one, but not tested. Do we really get a `NotLeaderExecption` if a simple client tries to access an HA OM node? I think there is a chance to connect directly to the leader, and in that case the exception can be different. 

Or am I wrong?


It may be better to throw this exception

Another approach is using  `OmUtils.isOmHAServiceId(conf, omHost)`. For example, we can override `OzoneClientAdapter createAdapter(ConfigurationSource conf, String omHost, int omPort)` in the `ozonefs-hadoop2` version of `RootedOzFs` and  `OzFs`. After checking the HA config whith this utlity we can call the parent implementation (if not HA)

symious: @elek Thanks for the comment.

Can reproduce the error by the following steps:
1. Prepare an OM-HA cluster.
2. Prepare the config files in HADOOP_CONF_DIR, including the core-site.xml and the ozone-site.xml.
3. Run the command of "hdfs dfs -ls o3fs://bucket.volume.cluster1/" on a follower OM. The client would get the NotLeaderExecption.
4. Run the command of "hdfs dfs -ls o3fs://bucket.volume.cluster1/" on the leader OM. The client would get normal result.
5. Run the command of "hdfs dfs -ls o3fs://bucket.volume.cluster1/" on other servers. The client would be retring to connect to "0.0.0.0:9862".

When Hadoop27RpcTransport is trying to initialize the socket to OM, client would use the config of "ozone.om.address", and for an OM-HA cluster, this value should be default as "0.0.0.0". 
So when client initialize the proxy on a follower OM host, it will try to connect to the follower OM, and an NotLeaderException will be thrown. And if the client is on the leader server, the proxy would be successfully created and no exceptions should be thrown. And If the client is on other servers, it will try to connect to the local OM, while on OM is running, so the client would be retrying to connect "0.0.0.0:9862".

elek: > And if the client is on the leader server, the proxy would be successfully created and no exceptions should be thrown

Thanks the clarification (and sorry for the late answer).

Based on my understanding this code throws exception only if the client is connected to a follower OM. While the client can be used when the client is lucky and connected to the OM.

While I can accept this behavior, it seems to be confusing. One Ozone user may think that hadoop2 compatible Ozone client can be used together with OM HA, but in fact it's unsafe beacuse it can be failed after any of the leader election.

I would consider to always throw an exception when HA is configured (but I can be convinced from the opposite).

What do you think?

symious: Thanks for the reply. 
I agree with the solution to always throw an exception since the client can only work when it's on the same server as the leader OM.
I will change the PR later.
Also, I'd like to add HA support for Hadoop 2, hope you can have a review. Thanks a lot.

adoroszlai:
@symious @elek If I understand correctly, this is superseded by #1858 (HDDS-4764).  Can we close this?

symious:
Yes, I will close this PR.

PR #1777, opened 2021-01-12
HDDS-4671. Use mvn to get ozone.version instead of complex shell commands

ferhui:

## How was this patch tested?

### DIFF
--- a/hadoop-ozone/dev-support/checks/blockade.sh
+++ b/hadoop-ozone/dev-support/checks/blockade.sh
@@ -17,7 +17,7 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 cd "$DIR/../../.." || exit 1
 
-OZONE_VERSION=$(grep "<ozone.version>" "$DIR/../../pom.xml" | sed 's/<[^>]*>//g'|  sed 's/^[ \t]*//')
+OZONE_VERSION=$(mvn help:evaluate -Dexpression=ozone.version -q -DforceStdout)
 cd "$DIR/../../dist/target/ozone-$OZONE_VERSION/tests" || exit 1
 
 source ${DIR}/../../dist/target/ozone-${OZONE_VERSION}/compose/ozoneblockade/.env
--- a/pom.xml
+++ b/pom.xml
@@ -72,8 +72,8 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <hadoop.version>3.2.1</hadoop.version>
 
     <!-- version for hdds/ozone components -->
-    <hdds.version>${ozone.version}</hdds.version>
-    <ozone.version>1.1.0-SNAPSHOT</ozone.version>
+    <hdds.version>${project.version}</hdds.version>
+    <ozone.version>${project.version}</ozone.version>
     <ozone.release>Denali</ozone.release>
     <declared.hdds.version>${hdds.version}</declared.hdds.version>
     <declared.ozone.version>${ozone.version}</declared.ozone.version>
--- a/pom.xml
+++ b/pom.xml
@@ -72,8 +72,8 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <hadoop.version>3.2.1</hadoop.version>
 
     <!-- version for hdds/ozone components -->
-    <hdds.version>${project.version}</hdds.version>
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -72,8 +72,8 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <hadoop.version>3.2.1</hadoop.version>
 
     <!-- version for hdds/ozone components -->
-    <hdds.version>${ozone.version}</hdds.version>
-    <ozone.version>1.1.0-SNAPSHOT</ozone.version>
+    <hdds.version>${project.version}</hdds.version>
+    <ozone.version>${project.version}</ozone.version>
```
elek, line 76: This is unsafe. "${project.version}" points to the version of the current project not to the version of this pom.xml.

Imagine this example:
 
 1. create a new project (ozone-tools:2.0.1-SNAPSHOT) where the parent project is `org.apache.hadoop:hadoop-main-ozone:1.1.0-SNAPSHOT).
 
  2. `ozone.version` will be evaluated as `2.0.1-SNAPSHOT`
  
  We have seen this problem when we used Hadoop parent pom from Ozone.
  
  try with this pom:
  
  ```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <parent>
    <artifactId>hadoop-main-ozone</artifactId>
    <version>1.1.0-SNAPSHOT</version>
  </parent>
  <artifactId>testproject</artifactId>
  <version>2.1.0-SNAPSHOT</version>
  <packaging>pom</packaging>
  <dependencies>
  <dependency>
        <artifactId>hadoop-ozone-client</artifactId>
        <version>${ozone.version}</version>
  </dependency>
  </dependencies>
</project>
```

The error is:

```
[ERROR] Failed to execute goal on project testproject: Could not resolve dependencies for project org.apache.hadoop:testproject:pom:2.1.0-SNAPSHOT: Failure to find org.apache.hadoop:hadoop-ozone-client:jar:2.1.0-SNAPSHOT in https://repository.apache.org/content/repositories/snapshots was cached in the local repository, resolution will not be reattempted until the update interval of apache.snapshots.https has elapsed or updates are forced -> [Help 1]
```

File: hadoop-ozone/dev-support/checks/blockade.sh [hunk 2]
```diff
@@ -17,7 +17,7 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 cd "$DIR/../../.." || exit 1
 
-OZONE_VERSION=$(grep "<ozone.version>" "$DIR/../../pom.xml" | sed 's/<[^>]*>//g'|  sed 's/^[ \t]*//')
+OZONE_VERSION=$(mvn help:evaluate -Dexpression=ozone.version -q -DforceStdout)
```
elek, line 20: Smart trick, like it.

elek (changes requested):
Thanks the patch @ferhui. I like the modification in the shell scripts, but we had a problem earlier with using `${project.version}` inside `<ozone.version>`.

I attached a pom.xml to reproduce the problem. Do you have any idea how to fix/avoid it?

pom.xml
ferhui: Thanks for comments.
 I have seen your report on hadoop https://issues.apache.org/jira/browse/HADOOP-15369.
But right now when i run following command, it fails.
`mvn versions:set -DnewVersion=1.1.0 && mvn versions:commit && mvn clean package -DskipTests -Dmaven.javadoc.skip=true
`
I must modify ozone.version manually and then build, Do you have any suggestions?

ferhui:
@elek How about to just modify acceptance.sh, blockade.sh & kubernetes.sh and do not change ozone.version to project.version?
Right now i have no idea for your above question. And when i can resolve it, i will address a new jira to fix it.
If so I will change the title for PR and this jira

elek:
> How about to just modify acceptance.sh, blockade.sh & kubernetes.sh and do not change ozone.version to project.version?

Good idea. I am fine with committing just the shell scripts. And we can continue thinking how can we remove the ozone.version property in a safe way....

elek (approved):
+1, thanks the update @ferhui

elek:
Merging it. Coverage upload is hanging, but all the other tests were green...

ferhui:
@elek Thanks for review!

PR #1778, opened 2021-01-12
HDDS-2402. Adapt hadolint check to improved CI framework

symious:

- Save the list of failures to summary.txt (while also outputting everything to stdout)
- Save output number of failures to failures

https://issues.apache.org/jira/browse/HDDS-2402

## How was this patch tested?

manually tested. Failures and output numbers are written to summary.txt and failures correctly.

### DIFF
--- a/hadoop-ozone/dev-support/checks/hadolint.sh
+++ b/hadoop-ozone/dev-support/checks/hadolint.sh
@@ -21,26 +21,20 @@ REPORT_DIR=${OUTPUT_DIR:-"$DIR/../../../target/hadolint"}
 mkdir -p "$REPORT_DIR"
 REPORT_FILE="$REPORT_DIR/summary.txt"
 
-for Dockerfile in $(find "$REPO_DIR/hadoop-ozone" "$REPO_DIR/hadoop-hdds" -name Dockerfile | sort); do
-  echo "Checking $Dockerfile" | tee -a "$REPORT_FILE"
-
+for Dockerfile in $(find hadoop-ozone hadoop-hdds -name Dockerfile | sort); do
   result=$( hadolint $Dockerfile )
   if [ ! -z "$result" ]
   then
     echo "$result" | tee -a "$REPORT_FILE"
-    echo "" | tee -a "$REPORT_FILE"
   fi
 done
 
-cat "$REPORT_FILE" |  egrep -v '^$|^Checking' | wc -l | awk '{print $1}'> "$REPORT_DIR/failures"
+wc -l "$REPORT_FILE" | awk '{print $1}'> "$REPORT_DIR/failures"
 
 if [ -s "${REPORT_FILE}" ]
 then
   echo "" | tee -a "$REPORT_FILE"
   echo "" | tee -a "$REPORT_FILE"
   echo "Hadolint errors were found. Exit code: 1." | tee -a "$REPORT_FILE"
-fi
-
-if [[ -s "${REPORT_FILE}" ]]; then
-   exit 1
+  exit 1
 fi
--- a/hadoop-ozone/dev-support/checks/hadolint.sh
+++ b/hadoop-ozone/dev-support/checks/hadolint.sh
@@ -20,13 +20,10 @@ REPO_DIR="$DIR/../../.."
 REPORT_DIR=${OUTPUT_DIR:-"$DIR/../../../target/hadolint"}
 mkdir -p "$REPORT_DIR"
 REPORT_FILE="$REPORT_DIR/summary.txt"
+echo -n > "$REPORT_FILE"
 
 for Dockerfile in $(find hadoop-ozone hadoop-hdds -name Dockerfile | sort); do
-  result=$( hadolint $Dockerfile )
-  if [ ! -z "$result" ]
-  then
-    echo "$result" | tee -a "$REPORT_FILE"
-  fi
+  hadolint $Dockerfile | tee -a "$REPORT_FILE"
 done
 
 wc -l "$REPORT_FILE" | awk '{print $1}'> "$REPORT_DIR/failures"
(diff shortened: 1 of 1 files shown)


hadoop-ozone/dev-support/checks/hadolint.sh
```diff
@@ -17,25 +17,30 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
 
-ERROR=0
-
+REPORT_DIR=${OUTPUT_DIR:-"$DIR/../../../target/hadolint"}
+mkdir -p "$REPORT_DIR"
+REPORT_FILE="$REPORT_DIR/summary.txt"
 
 for Dockerfile in $(find "$REPO_DIR/hadoop-ozone" "$REPO_DIR/hadoop-hdds" -name Dockerfile | sort); do
-  echo "Checking $Dockerfile"
+  echo "Checking $Dockerfile" | tee -a "$REPORT_FILE"
```
adoroszlai, line 25: This would cause exit code 1, because even without any errors, at least `Checking ...` would be present in the report file, hence `-s` evaluates to true.  I think we can remove this line, as hadolint's error messages include the input file's path.

hadoop-ozone/dev-support/checks/hadolint.sh
```diff
@@ -17,25 +17,30 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 cd "$DIR/../../.." || exit 1
 REPO_DIR="$DIR/../../.."
 
-ERROR=0
-
+REPORT_DIR=${OUTPUT_DIR:-"$DIR/../../../target/hadolint"}
+mkdir -p "$REPORT_DIR"
+REPORT_FILE="$REPORT_DIR/summary.txt"
 
 for Dockerfile in $(find "$REPO_DIR/hadoop-ozone" "$REPO_DIR/hadoop-hdds" -name Dockerfile | sort); do
```
adoroszlai, line 24: I suggest the following improvement:

```suggestion
for Dockerfile in $(find hadoop-ozone hadoop-hdds -name Dockerfile | sort); do
```

This would provide cleaner output, eg.:

```
hadoop-ozone/dev-support/docker/Dockerfile
```

instead of

```
/opt/ozone/src/ozone/hadoop-ozone/dev-support/checks/../../../hadoop-ozone/dev-support/docker/Dockerfile
```

hadoop-ozone/dev-support/checks/hadolint.sh
```diff
@@ -17,25 +17,30 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
-  echo ""
-  echo ""
-  echo "Hadolint errors were found. Exit code: 1."
-  exit 1
+  echo "" | tee -a "$REPORT_FILE"
+  echo "" | tee -a "$REPORT_FILE"
+  echo "Hadolint errors were found. Exit code: 1." | tee -a "$REPORT_FILE"
+fi
+
+if [[ -s "${REPORT_FILE}" ]]; then
```
adoroszlai, line 44: Nit: these two `if`s can be collapsed (no need to check the condition twice).

adoroszlai (changes requested):
Thanks @symious for working on this.  It mostly looks good.

hadoop-ozone/dev-support/checks/hadolint.sh
symious: Sure. If we are removing the 'Cheking ...', I was wondering if we can remove the spaceline too.

symious:
@adoroszlai Thanks for the comment. Have updated the commit according to the suggestions, please have a check.

hadoop-ozone/dev-support/checks/hadolint.sh
```diff
@@ -17,25 +17,24 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
 REPO_DIR="$DIR/../../.."
 
-ERROR=0
-
-
-for Dockerfile in $(find "$REPO_DIR/hadoop-ozone" "$REPO_DIR/hadoop-hdds" -name Dockerfile | sort); do
-  echo "Checking $Dockerfile"
+REPORT_DIR=${OUTPUT_DIR:-"$DIR/../../../target/hadolint"}
+mkdir -p "$REPORT_DIR"
+REPORT_FILE="$REPORT_DIR/summary.txt"
```
adoroszlai, line 22: To make sure `hadolint.sh` can be run multiple times and still give correct results, rather than keep accumulating findings:

```suggestion
REPORT_FILE="$REPORT_DIR/summary.txt"
echo -n > "$REPORT_FILE"
```

hadoop-ozone/dev-support/checks/hadolint.sh
```diff
@@ -17,25 +17,24 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
 
+for Dockerfile in $(find hadoop-ozone hadoop-hdds -name Dockerfile | sort); do
   result=$( hadolint $Dockerfile )
   if [ ! -z "$result" ]
   then
-    echo "$result"
-    echo ""
-    ERROR=1
+    echo "$result" | tee -a "$REPORT_FILE"
   fi
```
adoroszlai, line 29: Loop body can be simplified to:

```
  hadolint $Dockerfile | tee -a "$REPORT_FILE"
```

since we no longer need to set `ERROR` at the same time.

adoroszlai (commented):
Thanks @symious for updating the patch, it works fine.  On second look I would like to suggest two more small improvements.

symious:
@adoroszlai Thanks for the comment. Have updated the patch, could you have a check?

adoroszlai (approved):
Thanks @symious for continuing to improve the patch.

elek:
Merging it Thanks the patch @symious and the review @adoroszlai

symious:
Thanks @elek @adoroszlai .

PR #1780, opened 2021-01-12
HDDS-3205. DeleteBlock via Ratis in SCM HA

runzhiwang:

DeleteBlock needs to add/update/delete DELETED_BLOCKS_TABLE in SCMMetadataDB. So SCM does these operations via ratis.

https://issues.apache.org/jira/browse/HDDS-3205

## How was this patch tested?

Existed UT.

### DIFF
--- a/hadoop-hdds/interface-server/src/main/proto/SCMRatisProtocol.proto
+++ b/hadoop-hdds/interface-server/src/main/proto/SCMRatisProtocol.proto
@@ -23,6 +23,7 @@ option java_generate_equals_and_hash = true;
 enum RequestType {
     PIPELINE = 1;
     CONTAINER = 2;
+    BLOCK = 3;
 }
 
 message Method {
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/BlockManagerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/BlockManagerImpl.java
@@ -101,8 +101,11 @@ public class BlockManagerImpl implements BlockManager, BlockmanagerMXBean {
     mxBean = MBeans.register("BlockManager", "BlockManagerImpl", this);
 
     // SCM block deleting transaction log and deleting service.
-    deletedBlockLog = new DeletedBlockLogImpl(conf, scm.getContainerManager(),
-        scm.getScmMetadataStore());
+    deletedBlockLog = new DeletedBlockLogImplV2(conf, scm.getContainerManager(),
+        scm.getScmHAManager().getRatisServer(),
+        scm.getScmMetadataStore().getDeletedBlocksTXTable(),
+        scm.getScmHAManager().getDBTransactionBuffer(),
+        scm.getScmContext());
     Duration svcInterval = conf.getObject(
             ScmConfig.class).getBlockDeletionInterval();
     long serviceTimeout =
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLog.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLog.java
@@ -81,17 +81,6 @@ public interface DeletedBlockLog extends Closeable {
       UUID dnID);
 
   /**
-   * Creates a block deletion transaction and adds that into the log.
-   *
-   * @param containerID - container ID.
-   * @param blocks - blocks that belong to the same container.
-   *
-   * @throws IOException
-   */
-  void addTransaction(long containerID, List<Long> blocks)
-      throws IOException;
-
-  /**
    * Creates block deletion transactions for a set of containers,
    * add into the log and persist them atomically. An object key
    * might be stored in multiple containers and multiple blocks,
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
@@ -41,10 +41,12 @@ import org.apache.hadoop.hdds.scm.container.ContainerManagerV2;
 import org.apache.hadoop.hdds.scm.container.ContainerID;
 import org.apache.hadoop.hdds.scm.container.ContainerNotFoundException;
 import org.apache.hadoop.hdds.scm.container.ContainerReplica;
-import org.apache.hadoop.hdds.scm.metadata.SCMMetadataStore;
+import org.apache.hadoop.hdds.scm.ha.DBTransactionBuffer;
+import org.apache.hadoop.hdds.scm.ha.SCMContext;
+import org.apache.hadoop.hdds.scm.ha.SCMRatisServer;
 import org.apache.hadoop.hdds.server.events.EventHandler;
 import org.apache.hadoop.hdds.server.events.EventPublisher;
-import org.apache.hadoop.hdds.utils.db.BatchOperation;
+import org.apache.hadoop.hdds.utils.UniqueId;
 import org.apache.hadoop.hdds.utils.db.Table;
 import org.apache.hadoop.hdds.utils.db.TableIterator;
 
@@ -52,6 +54,7 @@ import com.google.common.collect.Lists;
 import static java.lang.Math.min;
 import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_BLOCK_DELETION_MAX_RETRY;
 import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_BLOCK_DELETION_MAX_RETRY_DEFAULT;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -65,7 +68,7 @@ import org.slf4j.LoggerFactory;
  * equally same chance to be retrieved which only depends on the nature
  * order of the transaction ID.
  */
-public class DeletedBlockLogImpl
+public class DeletedBlockLogImplV2
     implements DeletedBlockLog, EventHandler<DeleteBlockStatus> {
 
   public static final Logger LOG =
@@ -73,18 +76,23 @@ public class DeletedBlockLogImpl
 
   private final int maxRetry;
   private final ContainerManagerV2 containerManager;
-  private final SCMMetadataStore scmMetadataStore;
   private final Lock lock;
   // Maps txId to set of DNs which are successful in committing the transaction
   private Map<Long, Set<UUID>> transactionToDNsCommitMap;
+  // The access to DeletedBlocksTXTable is protected by
+  // DeletedBlockLogStateManager.
+  private final DeletedBlockLogStateManager deletedBlockLogStateManager;
+  private final SCMContext scmContext;
 
-  public DeletedBlockLogImpl(ConfigurationSource conf,
-                             ContainerManagerV2 containerManager,
-                             SCMMetadataStore scmMetadataStore) {
+  public DeletedBlockLogImplV2(ConfigurationSource conf,
+      ContainerManagerV2 containerManager,
+      SCMRatisServer ratisServer,
+      Table<Long, DeletedBlocksTransaction> deletedBlocksTXTable,
+      DBTransactionBuffer dbTxBuffer,
+      SCMContext scmContext) {
     maxRetry = conf.getInt(OZONE_SCM_BLOCK_DELETION_MAX_RETRY,
         OZONE_SCM_BLOCK_DELETION_MAX_RETRY_DEFAULT);
     this.containerManager = containerManager;
-    this.scmMetadataStore = scmMetadataStore;
     this.lock = new ReentrantLock();
 
     // transactionToDNsCommitMap is updated only when
@@ -92,6 +100,14 @@ public class DeletedBlockLogImpl
 
     // maps transaction to dns which have committed it.
     transactionToDNsCommitMap = new ConcurrentHashMap<>();
+    this.deletedBlockLogStateManager = DeletedBlockLogStateManagerImpl
+        .newBuilder()
+        .setConfiguration(conf)
+        .setDeletedBlocksTable(deletedBlocksTXTable)
+        .setRatisServer(ratisServer)
+        .setSCMDBTransactionBuffer(dbTxBuffer)
+        .build();
+    this.scmContext = scmContext;
   }
 
 
@@ -103,7 +119,7 @@ public class DeletedBlockLogImpl
       final List<DeletedBlocksTransaction> failedTXs = Lists.newArrayList();
       try (TableIterator<Long,
           ? extends Table.KeyValue<Long, DeletedBlocksTransaction>> iter =
-               scmMetadataStore.getDeletedBlocksTXTable().iterator()) {
+               deletedBlockLogStateManager.getReadOnlyIterator()) {
         while (iter.hasNext()) {
           DeletedBlocksTransaction delTX = iter.next().getValue();
           if (delTX.getCount() == -1) {
@@ -125,47 +141,18 @@ public class DeletedBlockLogImpl
    */
   @Override
   public void incrementCount(List<Long> txIDs) throws IOException {
-    for (Long txID : txIDs) {
-      lock.lock();
-      try {
-        DeletedBlocksTransaction block =
-            scmMetadataStore.getDeletedBlocksTXTable().get(txID);
-        if (block == null) {
-          if (LOG.isDebugEnabled()) {
-            // This can occur due to race condition between retry and old
-            // service task where old task removes the transaction and the new
-            // task is resending
-            LOG.debug("Deleted TXID {} not found.", txID);
-          }
-          continue;
-        }
-        DeletedBlocksTransaction.Builder builder = block.toBuilder();
-        int currentCount = block.getCount();
-        if (currentCount > -1) {
-          builder.setCount(++currentCount);
-        }
-        // if the retry time exceeds the maxRetry value
-        // then set the retry value to -1, stop retrying, admins can
-        // analyze those blocks and purge them manually by SCMCli.
-        if (currentCount > maxRetry) {
-          builder.setCount(-1);
-        }
-        scmMetadataStore.getDeletedBlocksTXTable().put(txID,
-            builder.build());
-      } catch (IOException ex) {
-        LOG.warn("Cannot increase count for txID " + txID, ex);
-        // We do not throw error here, since we don't want to abort the loop.
-        // Just log and continue processing the rest of txids.
-      } finally {
-        lock.unlock();
-      }
+    lock.lock();
+    try {
+      deletedBlockLogStateManager
+          .increaseRetryCountOfTransactionInDB(new ArrayList<>(txIDs));
+    } finally {
+      lock.unlock();
     }
   }
 
 
-  private DeletedBlocksTransaction constructNewTransaction(long txID,
-                                                           long containerID,
-                                                           List<Long> blocks) {
+  private DeletedBlocksTransaction constructNewTransaction(
+      long txID, long containerID, List<Long> blocks) {
     return DeletedBlocksTransaction.newBuilder()
         .setTxID(txID)
         .setContainerID(containerID)
@@ -187,6 +174,7 @@ public class DeletedBlockLogImpl
       List<DeleteBlockTransactionResult> transactionResults, UUID dnID) {
     lock.lock();
     try {
+      ArrayList<Long> txIDsToBeDeleted = new ArrayList<>();
       Set<UUID> dnsWithCommittedTxn;
       for (DeleteBlockTransactionResult transactionResult :
           transactionResults) {
@@ -229,7 +217,7 @@ public class DeletedBlockLogImpl
               if (LOG.isDebugEnabled()) {
                 LOG.debug("Purging txId={} from block deletion log", txID);
               }
-              scmMetadataStore.getDeletedBlocksTXTable().delete(txID);
+              txIDsToBeDeleted.add(txID);
             }
           }
           if (LOG.isDebugEnabled()) {
@@ -240,6 +228,12 @@ public class DeletedBlockLogImpl
           LOG.warn("Could not commit delete block transaction: " +
               transactionResult.getTxID(), e);
         }
+      }
+      try {
+        deletedBlockLogStateManager.removeTransactionsFromDB(txIDsToBeDeleted);
+      } catch (IOException e) {
+        LOG.warn("Could not commit delete block transactions: "
+            + txIDsToBeDeleted, e);
       }
     } finally {
       lock.unlock();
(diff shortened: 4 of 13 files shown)


runzhiwang:
@GlenGeng @amaliujia Could you help review this ? Thanks a lot.

amaliujia:
Will take a look! Thanks for this PR.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManager.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.hadoop.hdds.protocol.proto.StorageContainerDatanodeProtocolProtos.DeletedBlocksTransactionIDs;
+import org.apache.hadoop.hdds.scm.metadata.Replicate;
+import org.apache.hadoop.hdds.utils.db.TableIterator;
+import org.apache.hadoop.hdds.utils.db.Table.KeyValue;
+
+import java.io.IOException;
+
+public interface DeletedBlockLogStateManager {
+  @Replicate
+  void addTransactionsToDB(DeleteBlocksCommandProto txs) throws IOException;
```
amaliujia, line 31: @GlenGeng 

Per the design doc here should be a `List`. Is it better to use a `List` or just a repeated field in the proto?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
```diff
@@ -0,0 +1,412 @@
...
+   */
+  @Override
+  public void addTransactions(Map<Long, List<Long>> containerBlocksMap)
+      throws IOException {
+    lock.lock();
+    try {
+      List<DeletedBlocksTransaction> txs = new ArrayList<>();
+      for (Map.Entry< Long, List< Long > > entry :
+          containerBlocksMap.entrySet()) {
+        // TODO(runzhiwang): Should use distributed sequence id generator
```
amaliujia, line 320: +1

I am waiting https://github.com/apache/ozone/pull/1737 before sending the ID gen PR.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
```diff
@@ -0,0 +1,412 @@
...
+   * @throws IOException
+   */
+  @Override
+  public void incrementCount(List<Long> txIDs) throws IOException {
+    lock.lock();
+    try {
+      DeletedBlocksTransactionIDs transactionIDs =
+          DeletedBlocksTransactionIDs.newBuilder().addAllTxID(txIDs).build();
+      deletedBlockLogStateManager
+          .increaseRetryCountOfTransactionDB(transactionIDs);
```
bshashikant, line 144: Do we really need to propagate the retry count via Ratis? Let's double check on this..

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
```diff
@@ -337,6 +337,10 @@ message DeletedBlocksTransaction {
   required int32 count = 4;
 }
 
+message DeletedBlocksTransactionIDs {
+  repeated int64 txID = 1;
```
bshashikant, line 341: what would be the upgrade implications here after this proto change?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
```diff
@@ -0,0 +1,228 @@
...
+  public static final Logger LOG =
+      LoggerFactory.getLogger(DeletedBlockLogStateManagerImpl.class);
+
+  private final Table<Long, DeletedBlocksTransaction> deletedTable;
+  private final BatchOperationHandler batchHandler;
+  private final int maxRetry;
+
+  public DeletedBlockLogStateManagerImpl(
+      ConfigurationSource conf,
+      Table<Long, DeletedBlocksTransaction> deletedTable,
```
GlenGeng, line 54: just pass in the `ScmMetadataStore` to the Ctor.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
```diff
@@ -0,0 +1,412 @@
...
+
+  /**
+   * {@inheritDoc}
+   *
+   * @param containerID - container ID.
+   * @param blocks      - blocks that belong to the same container.
+   * @throws IOException
+   */
+  @Override
+  public void addTransaction(long containerID, List<Long> blocks)
```
GlenGeng, line 264: We will not support this method `addTransaction()` in HA code, since it is only used in test code, we need change affected tests as well.

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
GlenGeng: What does `upgrade implications` means ? could you please explain more ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManager.java
GlenGeng: We may need a better method signature here to handle the list type parameter.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
```diff
@@ -0,0 +1,228 @@
...
+      public void removeFromDB() throws IOException {
+        throw new UnsupportedOperationException("read-only");
+      }
+    };
+  }
+
+  @Override
+  public void addTransactionsToDB(DeleteBlocksCommandProto txs)
+      throws IOException {
+    BatchOperation batch = batchHandler.initBatchOperation();
```
GlenGeng, line 121: You need try-with-resource together with the `BatchOperation`, otherwise the rocksdb might crash.

See
```
  public void addTransactions(Map<Long, List<Long>> containerBlocksMap)
      throws IOException {
    lock.lock();
    try {
      try(BatchOperation batch =
          scmMetadataStore.getStore().initBatchOperation()) {
        for (Map.Entry< Long, List< Long > > entry :
            containerBlocksMap.entrySet()) {
          long nextTXID = scmMetadataStore.getNextDeleteBlockTXID();
          DeletedBlocksTransaction tx = constructNewTransaction(nextTXID,
              entry.getKey(), entry.getValue());
          scmMetadataStore.getDeletedBlocksTXTable().putWithBatch(batch,
              nextTXID, tx);
        }
        scmMetadataStore.getStore().commitBatchOperation(batch);
      }
    } finally {
      lock.unlock();
    }
  }
```

GlenGeng (commented):
LGTM. left some inline comments. Thanks for the work!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
amaliujia: Agreed.

I will send an email to master to discuss this idea.

amaliujia: In fact, there is a comment to explain the reason of persisting this retry count:

```
        // if the retry time exceeds the maxRetry value
        // then set the retry value to -1, stop retrying, admins can
        // analyze those blocks and purge them manually by SCMCli.
```

So the purpose actually makes sense. 

I think there is optimization room though. For example, only write to DB when the retry count becomes the maximum limit.

amaliujia: I sent an email to dev@ozone to discuss this count anyway.

amaliujia: Per the discussion in dev@, we can reduce the frequency of writing this count into DB. E.g. write once in every 100 retries. Created a JIRA for it: https://issues.apache.org/jira/browse/HDDS-4708

I think we can continue what we have here. Later we can apply HDDS-4708 to SCM HA to reduce the load on Ratis.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/BlockManagerImpl.java
```diff
@@ -105,8 +105,10 @@ public BlockManagerImpl(final ConfigurationSource conf,
     mxBean = MBeans.register("BlockManager", "BlockManagerImpl", this);
 
     // SCM block deleting transaction log and deleting service.
-    deletedBlockLog = new DeletedBlockLogImpl(conf, scm.getContainerManager(),
-        scm.getScmMetadataStore());
+    deletedBlockLog = new DeletedBlockLogImplV2(conf, scm.getContainerManager(),
+        scm.getScmHAManager().getRatisServer(),
+        scm.getScmMetadataStore().getDeletedBlocksTXTable(),
```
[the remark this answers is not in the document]
runzhiwang: I pass deletedBlockTxTable to BlockManagerImpl, because I think BlockManagerImpl should not see the entire metastore.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManager.java
runzhiwang: Will fix this later

hadoop-hdds/interface-server/src/main/proto/SCMRatisProtocol.proto
```diff
@@ -23,6 +23,7 @@ option java_generate_equals_and_hash = true;
 enum RequestType {
     PIPELINE = 1;
     CONTAINER = 2;
+    DELETE_BLOCK = 3;
 }
```
bshashikant, line 27: should we make this request type to be just BLOCK rather than DELETE_BLOCK ?
For now, for blocks we only need to handle delete block info, but in future if we have to extend this for some block specific attributes, it would be more simple.

For understanding as well, we can think of requests opearting at 3 levels, pipeline, container and blocks.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
```diff
@@ -0,0 +1,412 @@
...
+    implements DeletedBlockLog, EventHandler<DeleteBlockStatus> {
+
+  public static final Logger LOG =
+      LoggerFactory.getLogger(DeletedBlockLogImpl.class);
+
+  private final int maxRetry;
+  private final ContainerManagerV2 containerManager;
+  private final Lock lock;
+  // Maps txId to set of DNs which are successful in committing the transaction
+  private Map<Long, Set<UUID>> transactionToDNsCommitMap;
```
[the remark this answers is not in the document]
bshashikant: I agree.

runzhiwang:
@GlenGeng @bshashikant @amaliujia Thanks for review. I have updated the patch.

GlenGeng:
@runzhiwang  Is the CI failure related to your changes ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -152,6 +152,10 @@ public void notifyLeaderChanged(RaftGroupMemberId groupMemberId,
 
     LOG.info("current SCM becomes leader of term {}.", term);
     scm.getScmContext().updateIsLeaderAndTerm(true, term);
+    if (scm.getScmContext().isLeader()) {
```
GlenGeng, line 155: Why we need this isLeader check ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLog.java
```diff
@@ -104,4 +104,9 @@ void addTransactions(Map<Long, List<Long>> containerBlocksMap)
    * @throws IOException
    */
   int getNumOfValidTransactions() throws IOException;
+
+  /**
+   * Clear the transactionToDNsCommitMap.
+   */
+  void clearTransactionToDNsCommitMap();
```
GlenGeng, line 111: This method should not be seen by the non-ratis DeleteBlockLog.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
```diff
@@ -71,6 +74,14 @@ public static SCMContext emptyContext() {
     this.scm = scm;
   }
 
+  public void setDeletedBlockLog(DeletedBlockLog deletedBlockLog) {
```
GlenGeng, line 77: You should not pollute SCMContext with DeletedBlockLog, please find a better place.

hadoop-hdds/interface-server/src/main/proto/ScmServerDatanodeHeartbeatProtocol.proto
```diff
@@ -337,6 +337,10 @@ message DeletedBlocksTransaction {
   required int32 count = 4;
 }
 
+message DeletedBlocksTransactionIDs {
```
GlenGeng, line 340: Long type has been supported, can we remove this new added message ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/BlockManagerImpl.java
```diff
@@ -101,8 +103,11 @@ public BlockManagerImpl(final ConfigurationSource conf,
...
+        scm.getScmHAManager().getDBTransactionBuffer());
+    scmContext.setDeletedBlockLog(deletedBlockLog);
```
GlenGeng, line 110: This change ` scmContext.setDeletedBlockLog(deletedBlockLog);` is not needed.

You can use scm as a mediator.
```
StorageContainerManager.getScmBlockManager()
BlockManagerImpl.getDeletedBlockLog()
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManager.java
```diff
@@ -0,0 +1,43 @@
...
+  @Replicate
+  void addTransactionsToDB(ArrayList<DeletedBlocksTransaction> txs)
+      throws IOException;
+
+  @Replicate
+  void removeTransactionsFromDB(ArrayList<Long> txIDs)
+      throws IOException;
+
+  @Replicate
+  void increaseRetryCountOfTransactionDB(ArrayList<Long> txIDs)
```
GlenGeng, line 38: `increaseRetryCountOfTransactionDB` to `increaseRetryCountOfTransactionInDB`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
```diff
@@ -0,0 +1,221 @@
...
+  private final int maxRetry;
+
+  public DeletedBlockLogStateManagerImpl(
+      ConfigurationSource conf,
+      Table<Long, DeletedBlocksTransaction> deletedTable,
+      DBTransactionBuffer buffer) {
+    this.maxRetry = conf.getInt(OZONE_SCM_BLOCK_DELETION_MAX_RETRY,
+        OZONE_SCM_BLOCK_DELETION_MAX_RETRY_DEFAULT);
+    this.deletedTable = deletedTable;
+    this.transactionBuffer = buffer;
```
GlenGeng, line 57: We need add a cache for deleted keys, and clear the cache when transactionBuffer is flushed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
runzhiwang: you are right, we do not need this check.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -152,6 +154,13 @@ public void notifyLeaderChanged(RaftGroupMemberId groupMemberId,
 
     LOG.info("current SCM becomes leader of term {}.", term);
     scm.getScmContext().updateIsLeaderAndTerm(true, term);
+
+    DeletedBlockLog deletedBlockLog = scm.getScmBlockManager()
+        .getDeletedBlockLog();
+    if (deletedBlockLog instanceof DeletedBlockLogImplV2) {
```
GlenGeng, line 160: Preconditions.checkArgument(deletedBlockLog instanceof DeletedBlockLogImplV2)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -176,6 +185,11 @@ public long takeSnapshot() throws IOException {
     }
 
     transactionBuffer.flush();
+    DeletedBlockLog deletedBlockLog = scm.getScmBlockManager()
+        .getDeletedBlockLog();
+    if (deletedBlockLog instanceof DeletedBlockLogImplV2) {
+      ((DeletedBlockLogImplV2) deletedBlockLog).clearDeletingTxIDs();
```
GlenGeng, line 191: This should be called in `SCMDBTransactionBuffer.flush`

GlenGeng (commented):
LGTM, some typo related comments.

GlenGeng:
LGTM.

@bshashikant @amaliujia @nandakumar131 Please take one more look at this PR. Thanks!

This PR has implemented all the functionality according to the design: https://docs.google.com/document/d/166Aea2EowSGWtAFWNlDv0gu4rA06dQ2rJAsBd-l210Q/edit?usp=sharing.

After it is merged. all the features of Phase 2.0 of SCM HA will be finished.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
```diff
@@ -0,0 +1,398 @@
...
+        .setSCMDBTransactionBuffer(dbTxBuffer)
+        .build();
+    this.scmContext = scmContext;
+  }
+
+
+  @Override
+  public List<DeletedBlocksTransaction> getFailedTransactions()
+      throws IOException {
+    lock.lock();
```
amaliujia, line 117: Why lock in this function since this function does not update DB state?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
```diff
@@ -0,0 +1,398 @@
...
+      return transactions;
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  @Override
+  public void onMessage(
+      DeleteBlockStatus deleteBlockStatus, EventPublisher publisher) {
+    if (!scmContext.isLeader()) {
```
amaliujia, line 388: I am a bit confused here: so a delete block message could be still in-flight, and before it is processed, SCM leader could change, thus when `onMessage` is called, we can check whether `scmContext.isleader()` to confirm whether to proceed?

Is this  `scmContext.isleader()` just an optimization? E.g. because leader change can still happen after this check, thus those Ratis operations will fail by a NotLeaderException?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
runzhiwang: This lock only to keeps the origin code. I did not add new lock.

amaliujia: I see. Thanks for clarification.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
runzhiwang: It is just an optimization, if without this check, when leader change, it only waste a ratis call which will fail

amaliujia: Got it. Thanks for clarification.

amaliujia:
LGTM

GlenGeng:
@runzhiwang  Please rebase the HDDS-2823, since SCMServiceManager is merged.

runzhiwang:
@GlenGeng I have rebased it.

bshashikant (approved):

bshashikant:
Thanks @runzhiwang for the contribution.

linyiqun:
Sorry for the late comments.

I have one concern about current SCM HA handling mechanism. As the Blocks/Container/Pipeline update ops are cached in BatchOperation, and only be flushed during SCM snapshot take. But if user use the stand alone SCM mode to run based on current HA code. Is that will lead OOM error since stand alone SCM won't take any snapshot by itself while the request update will still go into SCM Transaction  Buffer in our current implementation.

@runzhiwang @GlenGeng , is this a real problem of my concern?

GlenGeng:
Hey @linyiqun, thanks for the review !

We planed to use single server raft cluster for the stand alone mode, which means the taking snapshot functionality works too.

linyiqun:
>We planed to use single server raft cluster for the stand alone mode, which means the taking snapshot functionality works too.

Okay, after that, that's not a problem.

PR #1781, opened 2021-01-12
HDDS-4672. Add warning log when old volume and bucket set quota

captainzmc:
https://github.com/apache/ozone/blob/master/hadoop-hdds/docs/content/feature/Quota.md
image

Currently we only advise users in the doc document not to enable quota for old volumes and buckets. Sometimes the user might not notice. So we added a warning on the client side, when user set quota on old volume and bucket.
image

https://issues.apache.org/jira/browse/HDDS-4672

## How was this patch tested?

Add log no need UT

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
@@ -299,8 +299,9 @@ public class RpcClient implements ClientProtocol {
     // that it is not recommended to enable quota.
     OmVolumeArgs omVolumeArgs = ozoneManagerClient.getVolumeInfo(volumeName);
     if (omVolumeArgs.getQuotaInNamespace() == OLD_QUOTA_DEFAULT) {
-      LOG.warn("Volume: {} is a old volume, usedNamespace may be " +
-          "inaccurate and it is not recommended to enable quota.", volumeName);
+      LOG.warn("Volume {} is created before version 1.1.0, usedNamespace " +
+          "may be inaccurate and it is not recommended to enable quota.",
+          volumeName);
     }
     ozoneManagerClient.setQuota(volumeName, quotaInNamespace, quotaInBytes);
   }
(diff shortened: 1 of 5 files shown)


hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
```diff
@@ -588,8 +597,19 @@ public void setBucketQuota(String volumeName, String bucketName,
...
+    if (omBucketInfo.getQuotaInNamespace() == OLD_QUOTA_DEFAULT) {
+      LOG.warn("Bucket: {} is a old bucket, usedNamespace may be " +
+          "inaccurate and it is not recommended to enable quota.", bucketName);
+    }
+    if (omBucketInfo.getUsedBytes() == OLD_QUOTA_DEFAULT) {
+      LOG.warn("Bucket: {} is a old bucket, usedBytes may be " +
+          "inaccurate and it is not recommended to enable quota.", bucketName);
+    }
     ozoneManagerClient.setBucketProperty(builder.build());
-
```
amaliujia, line 592: nit: better to undo this line of change.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
```diff
@@ -293,6 +295,13 @@ public void setVolumeQuota(String volumeName, long quotaInNamespace,
     HddsClientUtils.verifyResourceName(volumeName);
     verifyCountsQuota(quotaInNamespace);
     verifySpaceQuota(quotaInBytes);
+    // If the volume is old, we need to remind the user on the client side
+    // that it is not recommended to enable quota.
+    OmVolumeArgs omVolumeArgs = ozoneManagerClient.getVolumeInfo(volumeName);
+    if (omVolumeArgs.getQuotaInNamespace() == OLD_QUOTA_DEFAULT) {
+      LOG.warn("Volume: {} is a old volume, usedNamespace may be " +
```
amaliujia, line 302: I feel like "old" is not descriptive enough. 

Will it better to say that " {} is a volume created before version 1.1.0"? Same for the following warning messages.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
```diff
@@ -588,6 +598,20 @@ public void setBucketQuota(String volumeName, String bucketName,
...
+    if (omBucketInfo.getQuotaInNamespace() == OLD_QUOTA_DEFAULT) {
+      LOG.warn("Bucket {} is created before version 1.1.0, usedNamespace " +
+          "may be inaccurate and it is not recommended to enable quota.",
+          bucketName);
+    }
+    if (omBucketInfo.getUsedBytes() == OLD_QUOTA_DEFAULT) {
+      LOG.warn("Bucket {} is created before version 1.1.0, usedBytes " +
+          "may be inaccurate and it is not recommended to enable quota.",
+          bucketName);
+    }
```
ChenSammi, line 614: Can we pass this warnning message to client side?  Put it in log is not enough to raise the alert to user.

captainzmc: Yes, the warning here is able to be displayed directly to the user client
image

amaliujia:
overall LGTM

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/rpc/RpcClient.java
ChenSammi: @captainzmc ， we can see this output on console because LOG is enabled. When LOG is disabled on client side, we cannot see any message.  My suggestion is we return this message as the command response, and then output to console.

captainzmc:
Had added message as the command response, and output to console. [CI runs normally on my personal branch.](https://github.com/captainzmc/hadoop-ozone/runs/1699555309) cc @ChenSammi

ChenSammi:
+1

PR #1782, opened 2021-01-12
HDDS-4673. NodeStateMap leaks internal representation of container sets

elek:
JIRA: https://issues.apache.org/jira/browse/HDDS-4673

@jojochuang  [reported](https://issues.apache.org/jira/browse/HDDS-3918?focusedCommentId=17223103&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17223103) that some `ConcurrentModificationException` is visible even after #1174 (https://issues.apache.org/jira/browse/HDDS-3918) 

The root cause of this issue is a leak of internal representation of sets in `NodeStateMap`. 

`NodeStateMap`  returns an unmodifiable copy of containers for one specific node:

```
Collections.unmodifiableSet(nodeToContainer.get(uuid))
```

With this approach the users of the `nodeStateMap.getContainers` method couldn't modify the set anymore. But `nodeStateMap` can modify the original set which is referenced (instead of copy) by the `unmodifiableSet`. Even if it's unmodifiable it throws an exception during the iteration if the original set is modified in the background.

## How was this patch tested?

There is a new unit test method which reproduce the specific race condition with the help of `CountDownLatch`-es

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/states/NodeStateMap.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/states/NodeStateMap.java
@@ -18,18 +18,23 @@
 
 package org.apache.hadoop.hdds.scm.node.states;
 
-import org.apache.hadoop.hdds.protocol.DatanodeDetails;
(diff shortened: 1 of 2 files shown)


nandakumar131 (approved):

nandakumar131:
Thanks @elek for taking care of this.

PR #1783, opened 2021-01-12
HDDS-4674. Bucket usedBytes inaccurate because of preallocated blocks

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-4674

Currently, for every createKey request, OM will preallocate at least a block for the key and returned in the createKey response. When commit empty key, client will filter out the block location with 0 offsettt, so OM doesn't have the chance to remove the preallocated block size from bucket's usedbytes statics

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
@@ -184,6 +184,8 @@ public class OMKeyCommitRequest extends OMKeyRequest {
       omKeyInfo.setModificationTime(commitKeyArgs.getModificationTime());
 
       // Update the block length for each block
+      List<OmKeyLocationInfo> allocatedLocationInfoList =
+          omKeyInfo.getLatestVersionLocations().getLocationList();
       omKeyInfo.updateLocationInfoList(locationInfoList);
 
       // Set the UpdateID to current transactionLogIndex
@@ -206,7 +208,7 @@ public class OMKeyCommitRequest extends OMKeyRequest {
       // the actual Key size, and the total Block size applied before should
       // be subtracted.
       long correctedSpace = omKeyInfo.getDataSize() * factor -
-          locationInfoList.size() * scmBlockSize * factor;
+          allocatedLocationInfoList.size() * scmBlockSize * factor;
       omBucketInfo.incrUsedBytes(correctedSpace);
 
       omClientResponse = new OMKeyCommitResponse(omResponse.build(),
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneRpcClientAbstract.java
@@ -892,6 +892,7 @@ public abstract class TestOzoneRpcClientAbstract {
     // blocks will succeed，while the later block will fail.
     bucket.setQuota(OzoneQuota.parseQuota(
         4 * blockSize + " Bytes", 100));
+
     try {
       OzoneOutputStream out = bucket.createKey(UUID.randomUUID().toString(),
           valueLength, STAND_ALONE, ONE, new HashMap<>());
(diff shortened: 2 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -179,6 +174,20 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+      long blockStorageSize = ozoneManager.getScmBlockSize() *
+          omKeyInfo.getFactor().getNumber();
+      for (KeyLocation keyLocation : commitKeyArgs.getKeyLocationsList()) {
+        OmKeyLocationInfo locationInfo =
+            OmKeyLocationInfo.getFromProtobuf(keyLocation);
+        if (locationInfo.getLength() > 0) {
+          locationInfoList.add(OmKeyLocationInfo.getFromProtobuf(keyLocation));
+        }
+        allocatedSize += blockStorageSize;
+      }
```
captainzmc, line 189: Changes of the OMKeyCommitRequest are not required. The original logic has ensure correctedSpace is correct.

`long correctedSpace = omKeyInfo.getDataSize() * factor - allocatedSize;`
When the key length is 0:
`long correctedSpace = 0 * factor - allocatedSize;`

Just change BlockOutputStreamEntryPool, the added UT should also run normally.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
```diff
@@ -198,15 +207,13 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-      long scmBlockSize = ozoneManager.getScmBlockSize();
       int factor = omKeyInfo.getFactor().getNumber();
       omBucketInfo = getBucketInfo(omMetadataManager, volumeName, bucketName);
       // Block was pre-requested and UsedBytes updated when createKey and
       // AllocatedBlock. The space occupied by the Key shall be based on
       // the actual Key size, and the total Block size applied before should
       // be subtracted.
-      long correctedSpace = omKeyInfo.getDataSize() * factor -
-          locationInfoList.size() * scmBlockSize * factor;
+      long correctedSpace = omKeyInfo.getDataSize() * factor - allocatedSize;
```
captainzmc, line 216: Changes of the OMKeyCommitRequest are not required. The original logic has ensure correctedSpace is correct.

`long correctedSpace = omKeyInfo.getDataSize() * factor - allocatedSize;`
When the key length is 0:
`long correctedSpace = 0 * factor - allocatedSize;`

Just change BlockOutputStreamEntryPool, the added UT should also run normally.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyCommitRequest.java
captainzmc, line 189: Changes of the OMKeyCommitRequest are not required. The original logic has ensure correctedSpace is correct.

`long correctedSpace = omKeyInfo.getDataSize() * factor - allocatedSize;`
When the key length is 0:
`long correctedSpace = 0 * factor - allocatedSize;`

Just change BlockOutputStreamEntryPool, the added UT should also run normally.

ChenSammi: @captainzmc ,  I got your point.  But the key point is we should not persist zero 0 location Info to OM DB and need to filter it out.  Following is the code. 

if (locationInfo.getLength() > 0) {
          locationInfoList.add(OmKeyLocationInfo.getFromProtobuf(keyLocation));
}

captainzmc: Thanks for the explanation. Got your point, I agree with the changes here.

captainzmc:
Thanks @ChenSammi  for the fix.
LGTM overall, There are several UT failures, could confirm whether it is related to this PR?

ChenSammi:
> Thanks @ChenSammi for the fix.
> LGTM overall, There are several UT failures, could confirm whether it is related to this PR?

Yes, the failed UTs are related.   I'm working on a complete solution to resolve the impact of pre-allocated blocks to usedBytes.

ChenSammi:
The failed UT TestKeyInputStream.testSkip is irrelevant.

captainzmc (approved):
LGTM +1, Thanks for @ChenSammi's fix.

PR #1744, opened 2020-12-29
HDDS-4622: Use singe server raft cluster in MiniOzoneCluster.

GlenGeng:
## What changes were proposed in this pull request?

Use SCMHAManagerImpl and SCMRatisServerImpl instead of MockSCMHAManager and MockRatisServer in MiniOzoneCluster.
The MockRatisServer and MockSCMHAManager can not fully test the HA code.

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-4622

## How was this patch tested?

CI

PR #1784, opened 2021-01-12
HDDS-4295. Add SCMServiceManager to SCM HA.

GlenGeng:

SCM ServiceManager is going to control all the SCM background service so that they are only serving as the leader. 

Design doc: https://docs.google.com/document/d/1DbbqP0m3g_iEpY9qkSGOuQgcCN-QqlSNgWpvBOLv5h0/edit

https://issues.apache.org/jira/browse/HDDS-4295

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
@@ -23,6 +23,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
@@ -32,6 +34,8 @@ import org.apache.hadoop.hdds.scm.container.ContainerID;
 import org.apache.hadoop.hdds.scm.container.ContainerManagerV2;
 import org.apache.hadoop.hdds.scm.events.SCMEvents;
 import org.apache.hadoop.hdds.scm.ha.SCMContext;
+import org.apache.hadoop.hdds.scm.ha.SCMService;
+import org.apache.hadoop.hdds.scm.ha.SCMServiceManager;
 import org.apache.hadoop.hdds.scm.node.NodeManager;
 import org.apache.hadoop.hdds.scm.node.NodeStatus;
 import org.apache.hadoop.hdds.server.events.EventPublisher;
@@ -57,7 +61,8 @@ import org.slf4j.LoggerFactory;
  * SCM HB thread polls cached commands and sends them to datanode for physical
  * processing.
  */
-public class SCMBlockDeletingService extends BackgroundService {
+public class SCMBlockDeletingService extends BackgroundService
+    implements SCMService {
 
   public static final Logger LOG =
       LoggerFactory.getLogger(SCMBlockDeletingService.class);
@@ -71,11 +76,18 @@ public class SCMBlockDeletingService extends BackgroundService {
 
   private int blockDeleteLimitSize;
 
+  /**
+   * SCMServiceManager.SCMService related variables.
+   */
+  private final Lock serviceLock = new ReentrantLock();
+  private ServiceStatus serviceStatus = ServiceStatus.PAUSING;
+
   @SuppressWarnings("parameternumber")
   public SCMBlockDeletingService(DeletedBlockLog deletedBlockLog,
       ContainerManagerV2 containerManager, NodeManager nodeManager,
       EventPublisher eventPublisher, SCMContext scmContext,
-      Duration interval, long serviceTimeout, ConfigurationSource conf) {
+      SCMServiceManager serviceManager, Duration interval, long serviceTimeout,
+      ConfigurationSource conf) {
     super("SCMBlockDeletingService", interval.toMillis(), TimeUnit.MILLISECONDS,
         BLOCK_DELETING_SERVICE_CORE_POOL_SIZE, serviceTimeout);
     this.deletedBlockLog = deletedBlockLog;
@@ -88,6 +100,9 @@ public class SCMBlockDeletingService extends BackgroundService {
         conf.getObject(ScmConfig.class).getBlockDeletionLimit();
     Preconditions.checkArgument(blockDeleteLimitSize > 0,
         "Block deletion limit should be " + "positive.");
+
+    // register SCMBlockDeletingService to SCMServiceManager
+    serviceManager.register(this);
   }
 
   @Override
@@ -121,6 +136,10 @@ public class SCMBlockDeletingService extends BackgroundService {
 
     @Override
     public EmptyTaskResult call() throws Exception {
+      if (!shouldRun()) {
+        return EmptyTaskResult.newResult();
+      }
+
       long startTime = Time.monotonicNow();
       // Scan SCM DB in HB interval and collect a throttled list of
       // to delete blocks.
@@ -153,7 +172,7 @@ public class SCMBlockDeletingService extends BackgroundService {
               // command is bigger than a limit, e.g 50. In case datanode goes
               // offline for sometime, the cached commands be flooded.
               SCMCommand<?> command = new DeleteBlocksCommand(dnTXs);
-              command.setTerm(scmContext.getTerm());
+              command.setTerm(scmContext.getTermOfLeader());
               eventPublisher.fireEvent(SCMEvents.DATANODE_COMMAND,
                   new CommandForDatanode<>(dnId, command));
               if (LOG.isDebugEnabled()) {
@@ -199,5 +218,37 @@ public class SCMBlockDeletingService extends BackgroundService {
   @VisibleForTesting
   public void setBlockDeleteTXNum(int numTXs) {
     blockDeleteLimitSize = numTXs;
+  }
+
+
+  @Override
+  public void notifyRaftStatusOrSafeModeStatusChanged(
+      RaftStatus raftStatus, SafeModeStatus safeModeStatus) {
+    serviceLock.lock();
+    try {
+      // leader SCM
+      if (raftStatus == RaftStatus.LEADER) {
+        serviceStatus = ServiceStatus.RUNNING;
+      } else {
+        serviceStatus = ServiceStatus.PAUSING;
+      }
+    } finally {
+      serviceLock.unlock();
+    }
+  }
+
+  @Override
+  public boolean shouldRun() {
+    serviceLock.lock();
+    try {
+      return serviceStatus == ServiceStatus.RUNNING;
+    } finally {
+      serviceLock.unlock();
+    }
+  }
+
+  @Override
+  public String getServiceName() {
+    return SCMBlockDeletingService.class.getSimpleName();
   }
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
@@ -77,7 +77,7 @@ public class SCMBlockDeletingService extends BackgroundService
   private int blockDeleteLimitSize;
 
   /**
-   * SCMServiceManager.SCMService related variables.
+   * SCMService related variables.
    */
   private final Lock serviceLock = new ReentrantLock();
   private ServiceStatus serviceStatus = ServiceStatus.PAUSING;
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
@@ -220,14 +220,11 @@ public class SCMBlockDeletingService extends BackgroundService
     blockDeleteLimitSize = numTXs;
   }
 
-
   @Override
-  public void notifyRaftStatusOrSafeModeStatusChanged(
-      RaftStatus raftStatus, SafeModeStatus safeModeStatus) {
+  public void notifyStatusChanged() {
     serviceLock.lock();
     try {
-      // leader SCM or running without Ratis.
-      if (raftStatus == RaftStatus.LEADER || raftStatus == null) {
+      if (scmContext.isLeader()) {
         serviceStatus = ServiceStatus.RUNNING;
       } else {
         serviceStatus = ServiceStatus.PAUSING;
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
@@ -226,8 +226,8 @@ public class SCMBlockDeletingService extends BackgroundService
       RaftStatus raftStatus, SafeModeStatus safeModeStatus) {
     serviceLock.lock();
     try {
-      // leader SCM
-      if (raftStatus == RaftStatus.LEADER) {
+      // leader SCM or running without Ratis.
+      if (raftStatus == RaftStatus.LEADER || raftStatus == null) {
         serviceStatus = ServiceStatus.RUNNING;
       } else {
         serviceStatus = ServiceStatus.PAUSING;
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
@@ -71,8 +71,7 @@ public final class SCMContext {
   private final ReadWriteLock lock = new ReentrantReadWriteLock();
 
   private SCMContext(boolean isLeader, long term,
-                     final SafeModeStatus safeModeStatus,
-                     final StorageContainerManager scm) {
+      final SafeModeStatus safeModeStatus, final StorageContainerManager scm) {
     this.isLeader = isLeader;
     this.term = term;
     this.safeModeStatus = safeModeStatus;
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
@@ -17,10 +17,9 @@
 
 package org.apache.hadoop.hdds.scm.ha;
 
+import com.google.common.base.Preconditions;
 import org.apache.hadoop.hdds.scm.safemode.SCMSafeModeManager.SafeModeStatus;
 import org.apache.hadoop.hdds.scm.server.StorageContainerManager;
-import org.apache.hadoop.hdds.server.events.EventHandler;
-import org.apache.hadoop.hdds.server.events.EventPublisher;
 import org.apache.ratis.protocol.exceptions.NotLeaderException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -38,7 +37,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
  * check will always return true, and {@link SCMContext#getTermOfLeader} will
  * return INVALID_TERM.
  */
-public final class SCMContext implements EventHandler<SafeModeStatus> {
+public final class SCMContext {
   private static final Logger LOG = LoggerFactory.getLogger(SCMContext.class);
 
   /**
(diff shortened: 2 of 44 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMService.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Provide a fine-grained way to manipulate the status of these background
+ * services.
+ */
+public interface SCMService {
+  /**
+   * @param raftCondition latest raft condition
+   * @param safeModeCondition latest safe mode condition
+   */
+  void notifyRaftCondOrSafeModeCondChanged(
```
bshashikant, line 31: Can we add some more info here, regarding what would be one time transitional events and what would be recurring which needs to be notified ?

bshashikant:
Thanks @GlenGeng for working on this. I went through the patch once and given some comments inline. Still need to look through in more details.

I also want to know, how are we handling case of a stale leader here? I know there is an effort inside ratis to address/detect a case of stale leader., but don't we really need anything specific to handle such case here ?

Also, the transitional cases seem to be handled for LEADER_TO_FOLLOWER and FOLLOWER_TO_LEADER? How about transition to CANDIDATE state in general?

amaliujia:
cc @amaliujia

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMServiceManager.java
```diff
@@ -0,0 +1,125 @@
...
+    raftCondition = RaftCondition.FOLLOWER_TO_LEADER;
+    notifyRaftCondOrSafeModeCondChanged();
+  }
+
+  /**
+   * Current SCM steps down.
+   */
+  public synchronized void leaderToFollower() {
+    LOG.info("leaderToFollower is called.");
+    raftCondition = RaftCondition.FOLLOWER_TO_LEADER;
```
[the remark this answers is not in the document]
GlenGeng: Thanks for the catch.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+
+  /**
+   * Get term of current leader SCM.
+   *
+   * @return term
+   * @throws NotLeaderException if isLeader is false
+   */
+  public long getTerm() throws NotLeaderException {
```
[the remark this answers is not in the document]
GlenGeng: Agree.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
```diff
@@ -0,0 +1,164 @@
...
+    this(false, 0, new SafeModeStatus(true, false), scm);
+    Preconditions.checkNotNull(scm, "scm is null");
+  }
+
+  /**
+   *
+   * @param newIsLeader : is leader or not
+   * @param newTerm     : term if current SCM becomes leader
+   */
+  public void updateIsLeaderAndTerm(boolean newIsLeader, long newTerm) {
```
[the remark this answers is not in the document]
GlenGeng: Agree, will find a better name. Previously it is `isLeader` and `term`, but can not pass checkstyle.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMService.java
GlenGeng: Sure.

GlenGeng:
@bshashikant  Thanks for reviewing this patch!

> I also want to know, how are we handling case of a stale leader here? I know there is an effort inside ratis to address/detect a case of stale leader., but don't we really need anything specific to handle such case here ?

The request inside SCM can be divided into two categories, read/write request and read only request.

The read/write request will go through ratis as a RaftClientRequest, so we don’t need to care about whether the underlying RaftServer is a still leader or has already stepped down. These read/write requests include add/remove/update pipeline/container/deleted block.

For the read only request, they will not go through ratis, a stale leader SCM might make a dangerous decision if it misses the latest committed writes, a.k.a the split-brain issues occurred, two leaders of different terms live together.

Nanda pointed out such a case:
> Our major concern here is the DeleteContainerCommand sent by the stale leader. It is possible that the ICR of the recently created container is received by the stale leader which has no idea about the new container. The stale leader will end up sending DeleteContainerCommand to the datanode in this case.

See [SCM HA Handling Stale Leader](https://docs.google.com/document/d/1-5-KpR2GYIwWXGRH_C8IUVbFsm8RiETOVNYsMB5W8Ic/edit?usp=sharing) drafted by @nandakumar131 for more details.

With help of the lease solution, leader SCM can handle the read only request, then call `isLeader`/`getTermOfLeader` in `SCMContext` to confirm its leadership, ensuring that when it handles the read only request, it has seen the latest commit writes.

> Also, the transitional cases seem to be handled for LEADER_TO_FOLLOWER and FOLLOWER_TO_LEADER? How about transition to CANDIDATE state in general?

The candidate state is not expose by Ratis, and SCM does not cares about this state either. SCM just treat candidate the same as follower: it should stop the work if not leader.

`SCMServiceManager#leaderToFollower` is called in `StateMachine#notifyNotLeader` and `SCMServiceManager#followerToLeader` is called in  `StateMachine#notifyLeaderChanged` when `groupMemberId.getPeerId().equals(newLeaderId)`

We may have a offline talk to discuss more about the subtle cases in how we integrate SCM HA with ratis.

GlenGeng:
@bshashikant @amaliujia  I've update according to your comments, please take one more look Thanks!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
```diff
@@ -0,0 +1,208 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * SCMContext is the single source of truth for some key information shared
+ * across all components within SCM, including:
+ *  - RaftServer related info, e.g., isLeader, term.
+ *  - SafeMode related info, e.g., inSafeMode, preCheckComplete.
+ */
+public final class SCMContext implements EventHandler<SafeModeStatus> {
+  private static final Logger LOG = LoggerFactory.getLogger(SCMContext.class);
```
bshashikant, line 38: In case where SCM HA feature is not enabled , SCM context still seem to be relevant. In such cases, leader info and term info don't make sense. Can we add some constants to define leader state and term when HA is not enabled?

GlenGeng: Hey @bshashikant, this is a good point.

Due to PR https://github.com/apache/ozone/pull/1744, integration test which is bases on MiniOzoneCluster has been running upon single-server raft cluster. But we still have cases that run without Ratis, e.g., UnitTest, Recon.

Here is the design decision of SCMContext: 

The initial value of `currentTerm` in raft is 0. After RaftServer starts, they are all followers of term 0, then election timeout happens, one becomes candidate of term 1, finally the first leader will be elected, which is on a term that is greater or equal to 1.

SCMContext has two raft related field, `term` and `isLeader`. 
- For the cases using Ratis (no matter single server or multi server), the initial value of `term` and `isLeader` for SCMContext is `0` and `false`, they will be updated by `SCMStateMachine#notifyNotLeader` and `SCMStateMachine#notifyLeaderChanged`.
- For the cases not using Ratis, the initial value of `term` and `isLeader` for SCMContext is `0` and `true`, means leader of term 0. 

Since all the `isLeader` and `getTermOfLeader` check are going through `SCMConext`, and they are spreading across the whole SCM, for Ratis mode, it returns real raft info, for non Ratis mode, it behaviors as leader on term 0, so that we treat them in the same way.

What do you think ? We can discuss offline if needed.

GlenGeng:
cc @bshashikant Please take one more look !

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMServiceManager.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Manipulate background services in SCM, including ReplicationManager,
+ * SCMBlockDeletingService and BackgroundPipelineCreator.
+ */
+public final class SCMServiceManager {
+  private static final Logger LOG =
+      LoggerFactory.getLogger(SCMServiceManager.class);
+
+  // Cached latest raft status and safe mode status.
+  // If raftStatus equals null, it means current SCM is running without Ratis.
```
bshashikant, line 37: I think its better to use SCMContext object info here instead of maintaing raft status as well as safe mode status.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMServiceManager.java
```diff
@@ -0,0 +1,148 @@
...
+    notifyRaftStatusOrSafeModeStatusChanged();
+  }
+
+  /**
+   * Current SCM leaves safe mode.
+   */
+  public synchronized void leavingSafeMode() {
+    LOG.info("SCM leaves SafeMode.");
+    safeModeStatus = SafeModeStatus.OUT_OF_SAFE_MODE;
+    notifyRaftStatusOrSafeModeStatusChanged();
```
bshashikant, line 102: Let's separate out the leader change and safe mode notification as these will be two independent events,

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/BackgroundPipelineCreatorV2.java
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    LOG.info("Starting {}.", THREAD_NAME);
+
+    thread = new ThreadFactoryBuilder()
+        .setDaemon(false)
+        .setNameFormat(THREAD_NAME + " - %d")
+        .setUncaughtExceptionHandler((Thread t, Throwable ex) -> {
+          String message = "Terminate SCM, encounter uncaught exception"
+              + " in RatisPipelineUtilsThread";
```
bshashikant, line 125: instead of terminate , can we gracefully shutdown???

bshashikant (changes requested):

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
```diff
@@ -200,4 +219,36 @@ public EmptyTaskResult call() throws Exception {
   public void setBlockDeleteTXNum(int numTXs) {
     blockDeleteLimitSize = numTXs;
   }
+
+
+  @Override
+  public void notifyRaftStatusOrSafeModeStatusChanged(
+      RaftStatus raftStatus, SafeModeStatus safeModeStatus) {
```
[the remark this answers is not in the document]
bshashikant: > is safeModeStatus really required. The function seems to care about just leader info.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
```diff
@@ -117,9 +122,13 @@ public boolean isLeader() {
    * @return term
    * @throws NotLeaderException if isLeader is false
    */
-  public long getTerm() throws NotLeaderException {
+  public long getTermOfLeader() throws NotLeaderException {
     lock.readLock().lock();
     try {
+      if (term == INVALID_TERM) {
```
bshashikant, line 128: Is this check required??

bshashikant (changes requested):

GlenGeng:
@bshashikant Comments for SCMServiceManager has been addressed, Please take one more look. Thanks!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMService.java
```diff
@@ -0,0 +1,65 @@
...
+   * @return name of the Service.
+   */
+  String getServiceName();
+
+  /**
+   * Status of Service.
+   */
+  enum ServiceStatus {
+    RUNNING,
+    PAUSING
```
bshashikant, line 54: I think, we should add a status called STOPPED as well here. For example, if i want to suspend replication i can issue, ozone admin replicationManager stop. This state might need to be propagated over Ratis as well. Even after leader re-relection, the service should not auto start on the new leader.

bshashikant (changes requested):
Thanks @GlenGeng for updating the patch.The changes look good to me. I need a clarification regarding replicationManager.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMService.java
GlenGeng: Thanks for the insight. 
I will handle `bin/ozone admin replicationmanager start|stop|status` in this PR. 
I agree that the status of RM should also be replicated through ratis. I created a jira https://issues.apache.org/jira/browse/HDDS-4740 to track this issue, will schedule a discussion in this week's sync.

GlenGeng: create jira for Add  STOP state to SCMService.
https://github.com/apache/ozone/pull/1784

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
```diff
@@ -71,11 +76,18 @@
 
   private int blockDeleteLimitSize;
 
+  /**
+   * SCMService related variables.
+   */
+  private final Lock serviceLock = new ReentrantLock();
+  private ServiceStatus serviceStatus = ServiceStatus.PAUSING;
```
amaliujia, line 83: Instead of setting default value as `ServiceStatus.PAUSING`, will it make sense to initialize this variable as RUNNING during construction?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMService.java
amaliujia: I guess the JIRA link of adding STOP state to SCM service was not properly linked?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMContext.java
```diff
@@ -62,35 +70,26 @@ public static SCMContext emptyContext() {
   private final StorageContainerManager scm;
   private final ReadWriteLock lock = new ReentrantReadWriteLock();
 
-  SCMContext(boolean isLeader, long term,
-             final SafeModeStatus safeModeStatus,
-             final StorageContainerManager scm) {
+  private SCMContext(boolean isLeader, long term,
+                     final SafeModeStatus safeModeStatus,
+                     final StorageContainerManager scm) {
```
amaliujia, line 75: nit: undo indent change?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMServiceManager.java
```diff
@@ -0,0 +1,67 @@
...
+      LoggerFactory.getLogger(SCMServiceManager.class);
+
+  private final List<SCMService> services = new ArrayList<>();
+
+  /**
+   * Register a SCMService to SCMServiceManager.
+   */
+  public synchronized void register(SCMService service) {
+    Preconditions.checkNotNull(service);
+    LOG.info("register Service {}.", service.getServiceName());
```
amaliujia, line 43: Nit: `register Service` -> `Registering service`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMServiceManager.java
```diff
@@ -0,0 +1,67 @@
...
+      service.notifyStatusChanged();
+    }
+  }
+
+  /**
+   * Notify event triggered, which may affect SCMService.
+   */
+  public synchronized void notifyEventTriggered(Event event) {
+    for (SCMService service : services) {
+      LOG.info("Notify service:{} with event:{}.",
```
amaliujia, line 62: Nit: do we need a LOG.info for this information?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMServiceManager.java
```diff
@@ -0,0 +1,67 @@
...
+    LOG.info("register Service {}.", service.getServiceName());
+    services.add(service);
+  }
+
+  /**
+   * Notify raft or safe mode related status changed.
+   */
+  public synchronized void notifyStatusChanged() {
+    for (SCMService service : services) {
+      LOG.info("Notify service:{}.", service.getServiceName());
```
amaliujia, line 52: Nit: do we need a LOG.info just a `notifyStatusChanged` call?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMServiceManager.java
```diff
@@ -0,0 +1,67 @@
...
+  public synchronized void register(SCMService service) {
+    Preconditions.checkNotNull(service);
+    LOG.info("register Service {}.", service.getServiceName());
+    services.add(service);
+  }
+
+  /**
+   * Notify raft or safe mode related status changed.
+   */
+  public synchronized void notifyStatusChanged() {
```
amaliujia, line 50: Question: Does `notifyStatusChanged` has to be a blocking call?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/BackgroundPipelineCreatorV2.java
```diff
@@ -0,0 +1,303 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+import static org.apache.hadoop.hdds.scm.ha.SCMService.Event.UNHEALTHY_TO_HEALTHY_NODE_HANDLER_TRIGGERED;
+import static org.apache.hadoop.hdds.scm.ha.SCMService.Event.NEW_NODE_HANDLER_TRIGGERED;
+import static org.apache.hadoop.hdds.scm.ha.SCMService.Event.PRE_CHECK_COMPLETED;
+
+/**
+ * Implements api for running background pipeline creation jobs.
+ */
+public class BackgroundPipelineCreatorV2 implements SCMService {
```
amaliujia, line 50: I might have missed: is this class tested somehow?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMService.java
GlenGeng: Thanks for the catch.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
GlenGeng: Actually not.  BlockDeletingService should not run on follower SCMs.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMServiceManager.java
GlenGeng: The execution of this method is quite quick. We may consider relax it if we meet performance issue in future.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/BackgroundPipelineCreatorV2.java
GlenGeng: `BackgroundPipelineCreator` is replaced by `BackgroundPipelineCreatorV2`, and it has been fully test in integration test and docker test.

bshashikant (approved):

amaliujia:
LGTM!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
amaliujia: O yes thanks!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMServiceManager.java
amaliujia: SG!

runzhiwang:
@GlenGeng Thanks the patch. @bshashikant @amaliujia Thanks for review. I have merged it.

PR #1786, opened 2021-01-12
HDDS-4675. Support scanning content of DN rocksdb instances with current scheme.

elek:
JIRA: https://issues.apache.org/jira/browse/HDDS-4675

I realized that `ozone debug ldb` command doesn't support the container dbs. It's hard to support the old version (everything is in the same column family) but seems to be fine to support only the `DatanodeSchemaTwoDBDefinition` as all the new Ozone clusters will use this scheme. 

## How was this patch tested?

Unit test is updated, but also tested with:

```
./ozone debug --verbose --set hdds.datanode.dir=/home/elek/tmp/datanode3/storage  ldb --db=/home/elek/tmp/datanode1/storage/hdds/fbab0cc1-7973-435c-b25a-1097d6b60d95/current/containerDir0/1/metadata/1-dn-container.db scan  --column_family=blockData
```


adoroszlai (approved):
Thanks @elek for the improvement.  I also verified it still works for OM DB.

PR #1788, opened 2021-01-13
HDDS-2212. Genconf tool should generate config files for secure clust…

symious:
…er setup

- Add the option of "--security" to the tool of "genconf".

https://issues.apache.org/jira/browse/HDDS-2212

## How was this patch tested?

Manually tested. The usage of the tool is now as follows.
```
Usage: ozone genconf [-hV] [--security] [--verbose] [-conf=<configurationPath>]
                     [-D=<String=String>]... <path>
Tool to generate template ozone-site.xml
      <path>       Directory path where ozone-site file should be generated.
      --security   Generate security config.
      --verbose    More verbose output. Show the stack trace of the errors.
      -conf=<configurationPath>

  -D, --set=<String=String>

  -h, --help       Show this help message and exit.
  -V, --version    Print version information and exit.
```

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -434,7 +434,7 @@
   </property>
   <property>
     <name>ozone.om.service.ids</name>
-    <value></value>
+    <value/>
     <tag>OM, HA</tag>
     <description>
       Comma-separated list of OM service Ids. This property allows the client
@@ -443,7 +443,7 @@
   </property>
   <property>
     <name>ozone.om.internal.service.id</name>
-    <value></value>
+    <value/>
     <tag>OM, HA</tag>
     <description>
       Service ID of the Ozone Manager. If this is not set fall back to
@@ -452,7 +452,7 @@
   </property>
   <property>
     <name>ozone.om.nodes.EXAMPLEOMSERVICEID</name>
-    <value></value>
+    <value/>
     <tag>OM, HA</tag>
     <description>
       Comma-separated list of OM node Ids for a given OM service ID (eg.
@@ -470,7 +470,7 @@
   </property>
   <property>
     <name>ozone.om.node.id</name>
-    <value></value>
+    <value/>
     <tag>OM, HA</tag>
     <description>
       The ID of this OM node. If the OM node ID is not configured it
@@ -1262,7 +1262,7 @@
 
   <property>
     <name>hdds.datanode.plugins</name>
-    <value></value>
+    <value/>
     <description>
       Comma-separated list of HDDS datanode plug-ins to be activated when
       HDDS service starts as part of datanode.
@@ -1455,6 +1455,37 @@
   </property>
 
   <property>
+    <name>hdds.scm.kerberos.keytab.file</name>
+    <value/>
+    <tag>SCM, SECURITY, KERBEROS</tag>
+    <description> The keytab file used by SCM daemon to login as its service principal.
+    </description>
+  </property>
+  <property>
+    <name>hdds.scm.kerberos.principal</name>
+    <value/>
+    <tag>SCM, SECURITY, KERBEROS</tag>
+    <description>The SCM service principal. e.g. scm/_HOST@REALM.COM</description>
+  </property>
+  <property>
+    <name>hdds.scm.http.auth.kerberos.principal</name>
+    <value>HTTP/_HOST@REALM</value>
+    <tag>SCM, SECURITY, KERBEROS</tag>
+    <description>
+      SCM http server service principal if SPNEGO is enabled for SCM http server.
+    </description>
+  </property>
+  <property>
+    <name>hdds.scm.http.auth.kerberos.keytab</name>
+    <value>/path/to/HTTP.keytab</value>
+    <tag>SCM, SECURITY, KERBEROS</tag>
+    <description>
+      The keytab file used by SCM http server to login as its service
+      principal if SPNEGO is enabled for SCM http server.
+    </description>
+  </property>
+
+  <property>
     <name>ozone.s3g.volume.name</name>
     <value>s3v</value>
     <tag>OZONE, S3GATEWAY</tag>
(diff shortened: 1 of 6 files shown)


amaliujia:
@symious is it possible to attach a screenshot for the output of this command flag.

symious:
Sure. The screen shot is as follows.
Screenshot 2021-01-13 at 2 00 19 PM

elek:
Thanks the patch @symious. It looks a simple change for me.

My only fear is that I am not sure if the generated config file is enough to start Ozone in secure environment. 

For unsecure environment all the required attributes are filled with some reasonable default (like using /tmp for metadata).

I don't know what is a good approach, but the overall goal is to make the configuration easier for the users.

symious:
> Thanks the patch @symious. It looks a simple change for me.
> 
> My only fear is that I am not sure if the generated config file is enough to start Ozone in secure environment.
> 
> For unsecure environment all the required attributes are filled with some reasonable default (like using /tmp for metadata).
> 
> I don't know what is a good approach, but the overall goal is to make the configuration easier for the users.

@elek Thanks for the reply. IMHO, secure config is for advanced users, that they do have the ability and requirement to build a secure cluster. The default config may also mislead the user somehow.

adoroszlai:
> My only fear is that I am not sure if the generated config file is enough to start Ozone in secure environment.

Thanks @elek for sharing your thoughts.  I also had the same doubt: most of the security-related configs are empty by default.  That's why I pinged @dineshchitlangia, the reporter of this task, for review, in the hope that we can clarify the intended scope.

symious:
@elek @adoroszlai Thanks for the reply. I think it will be good to wait for @dineshchitlangia 's reply. Meanwhile, I can try to build a secure cluster and fill in the required configs.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/genconf/GenerateOzoneRequiredConfigurations.java
```diff
@@ -61,6 +62,9 @@
       description = "Directory path where ozone-site file should be generated.")
   private String path;
 
+  @Option(names = "--security", description = "Generate security config.")
```
dineshchitlangia, line 65:
```suggestion
  @Option(names = "--security", description = "Generates security config template, update template with proper values before use.")
```

dineshchitlangia (changes requested):
@symious  Thanks for working on this.

I also reviewed thoughts from @adoroszlai and @elek .

My thoughts on this are:
1. We certainly need to make it easier for users as @elek mentioned.
2. I think we can start by generating a security config template and that is why I suggested the inline change.
3. We can file another jira to generate conf with predefined security config values(if at all possible).

symious:
@dineshchitlangia Thanks for the review. Added a new commit with the following update:
1. Added the unit test for security config generation, and update the command option description.
2. After checking the configurations with the tag of "SECURITY", I found a lot of unrelated or fundamental configs that users shouldn't use usually, such as`hdds.x509.signature.algorithm`. And then I walk through the document of Ozone Security, I found that all configs are related to kerberos, so my idea is to add a new tag named as "KERBEROS" to track the common used security configs.
3. After checking the docker file of ozonesecure, I found that the principal and keytab are coordinated with the KDC server, which means the KDC server and the keytab file has to be generated and prepared before we start ozone cluster,  so the configs must be updated manually unless we set up a KDC server and update the keytab files for the users, which I think is quite strange.

In docker, flokkr/issuer helps to initial KDC and update the keytab files, so for testing, it would be quite convenient for the user to just use docker. And for `genconf --security`, I think it's mainly for the users who has Kerberos set up already and want to fill the required security configs template for a productive secure cluster.

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -2143,57 +2187,65 @@
...
     <description>
       The keytab file for HTTP Kerberos authentication in Recon.
     </description>
   </property>
   <property>
     <name>ozone.recon.http.auth.kerberos.principal</name>
-    <value/>
-    <tag>RECON</tag>
+    <value>HTTP/_HOST@REALM</value>
+    <tag>RECON, KERBEROS</tag>
```
dineshchitlangia, line 2199: Add tag SECURITY  in line with other similar configs.

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -1699,31 +1740,34 @@
...
+    <tag>OZONE, SECURITY, KERBEROS</tag>
     <description>
-      OzoneManager http server kerberos principal.
+      Ozone Manager http server service principal if SPNEGO is enabled for om http server.
     </description>
   </property>
   <property>
     <name>ozone.om.http.auth.kerberos.keytab</name>
-    <value>/etc/security/keytabs/HTTP.keytab</value>
+    <value>/path/to/HTTP.keytab</value>
```
dineshchitlangia, line 1766: I wouldn't recommend changing the default value of the path to keytab on all such occurrences.
From what I have seen, most admins keep the keytabs at /etc/security/keytabs.

If we avoid changing this, then only a very few subset of users would have to change this in the template before using.

dineshchitlangia (changes requested):
@symious Thank you for updating the PR.
Minor changes requested inline.

@elek what are your thoughts on the new KERBEROS tag? For me, it is fine to add the new tag as it is easy to identify which configs to deal with when you want to implement kerberos but I would like to hear your opinion too.

hadoop-hdds/common/src/main/resources/ozone-default.xml
symious: @dineshchitlangia Thanks for the review, the keytabs directory has been changed to "/etc/security/keytabs".

hadoop-hdds/common/src/main/resources/ozone-default.xml
dineshchitlangia: @symious any thoughts about this?

elek (approved):
+1 let's merge it after a green build

I asked others on the community meeting and the agreement was that it's better to have this even if it doesn't generate all the required configs (keytabs should be provided, etc...)

Comments from @dineshchitlangia are adressed as far as I see, the only thing what we need is a green build...

symious:
@elek Thanks for the review, I just found the other comment @dineshchitlangia mentioned above, let me add the "SECURITY" tag, too.

hadoop-hdds/common/src/main/resources/ozone-default.xml
symious: @dineshchitlangia Thanks for the review, SECURITY tag added.

dineshchitlangia (approved):
Thanks for making the changes. +1

elek:
Merging it now. Thanks the update @symious and the review @dineshchitlangia and @adoroszlai

PR #1790, opened 2021-01-13
HDDS-4593. Refine IAccessAuthorizer interface to do recursive ACL check on a path

rakeshadr:

The idea/proposal is to refine #checkAccess interface by giving "recursive" hint and make only one single call during recursive delete or list operations. Now, the implementation classes(native or external entity) can do recursive logic in an optimized way.

https://issues.apache.org/jira/browse/HDDS-4593

## How was this patch tested?

Modified existing UTs to check the flag values. Will add more test cases along with atomic delete ops.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/RequestContext.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/RequestContext.java
@@ -125,8 +125,8 @@ public class RequestContext {
       return this;
     }
 
-    public Builder setRecursiveAccessCheck(boolean recursiveAccessCheck) {
-      this.recursiveAccessCheck = recursiveAccessCheck;
+    public Builder setRecursiveAccessCheck(boolean recursiveAccessCheckFlag) {
+      this.recursiveAccessCheck = recursiveAccessCheckFlag;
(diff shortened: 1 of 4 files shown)


mukul1987 (changes requested):
I think we also need changes to use or set the resrusiveAccessCheck flag in our code. Will that be covered as part of another patch ?

rakeshadr:
> I think we also need changes to use or set the resrusiveAccessCheck flag in our code. Will that be covered as part of another patch ?

Thanks @mukul1987 for the reviews. This PR is targeting to the master branch and I am planning to do the recursiveCheck=true on the feature branch code. Once this PR is merged then I will rebase the feature branch with master and then will do the settings along with HDDS-4691 task.

Hope this makes sense to you.

xiaoyuyao (approved):
LGTM, +1 pending CI.

rakeshadr:
> LGTM, +1 pending CI.

Thanks @xiaoyuyao for the reviews. CI failure is unrelated to the patch, anyway I've triggered another build.
I could see HDDS-4693 is raised to fix TestContainerStateMachine failure.

rakeshadr:
Thank you @mukul1987 , @xiaoyuyao for the reviews!

PR #1791, opened 2021-01-13
HDDS-4162. Use BeforeClass to init test cluster in TestOzoneFileSystem

cku328:

1. Modify to initialize a MiniOzoneCluster in `@BeforeClass`, then tear it down in `@AfterClass`.
2. Tweaked the test case cleanup process to prevent the failure of one test case from affecting other tests. (my approach is to delete all files and directories in `@After`.)

https://issues.apache.org/jira/browse/HDDS-4162

## How was this patch tested?

Unit tests passed.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -163,7 +163,7 @@ public class TestOzoneFileSystem {
       for (FileStatus fileStatus : fileStatuses) {
         fs.delete(fileStatus.getPath(), true);
       }
-    }catch (IOException ex){
+    } catch (IOException ex){
       fail("Failed to cleanup files.");
     }
   }
@@ -258,7 +258,7 @@ public class TestOzoneFileSystem {
     try{
       LambdaTestUtils.intercept(InvalidPathException.class, "Invalid path Name",
           () -> fs.create(path, false));
-    }catch (AssertionError e){
+    } catch (AssertionError e){
       fail("testCreateWithInvalidPaths failed for path" + path);
     }
   }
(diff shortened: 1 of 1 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -187,119 +236,46 @@ public void testMakeDirsWithAnExistingDirectoryPath() throws Exception {
...
   }
 
   private void checkInvalidPath(Path path) throws Exception {
-    FSDataOutputStream outputStream = null;
-    try  {
-      outputStream = fs.create(path, false);
+    try{
+      LambdaTestUtils.intercept(InvalidPathException.class, "Invalid path Name",
+          () -> fs.create(path, false));
+    }catch (AssertionError e){
```
smengcl, line 261: same nit: `} catch (AssertionError e) {`

smengcl (approved):
Thanks for working on this. Overall lgtm

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -107,17 +107,68 @@ public TestOzoneFileSystem(boolean setDefaultFs, boolean enableOMRatis) {
...
+
+  @After
+  public void cleanup() {
+    try {
+      Path root = new Path("/");
+      FileStatus[] fileStatuses = fs.listStatus(root);
+      for (FileStatus fileStatus : fileStatuses) {
+        fs.delete(fileStatus.getPath(), true);
+      }
+    }catch (IOException ex){
```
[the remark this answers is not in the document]
cku328: done.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
cku328: done.

cku328:
Thanks @smengcl for the review.
I have updated the patch based on comments.

smengcl:
Thanks @cku328 . I will merge this in a moment.

PR #1792, opened 2021-01-13
HDDS-4679. Intermittent failure in TestSCMRestart.

bshashikant:

Set the datanode pipeline limit to 1 as required for the test.

https://issues.apache.org/jira/browse/HDDS-4679

## How was this patch tested?

Verified the unit test works.


adoroszlai (approved):
Thanks @bshashikant for the fix, LGTM.

PR #1793, opened 2021-01-13
HDDS-4680. Change default OM Node ID from UUID to a constant

hanishakoneru:

If a nodeID is not set explicitly (for a single node OM cluster), then it defaults to the OM storage ID which is an UUID string. Proposing to change this default to a constant (such as "om1") instead. 
This would help when a cluster is upgraded to HA for example. It is not straightforward to change the nodeID after Ratis server has been instantiated (as the nodeID is used to generate the RaftPeerID). Hence, it would be good to have a more readable nodeID by default. 

Note that existing single node Ratis enabled OM clusters will have to set the nodeId explicitly if upgraded.

https://issues.apache.org/jira/browse/HDDS-4680

## How was this patch tested?

Manually tested.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
@@ -343,6 +343,7 @@ public final class OzoneConsts {
 
   // Default OMServiceID for OM Ratis servers to use as RaftGroupId
   public static final String OM_SERVICE_ID_DEFAULT = "omServiceIdDefault";
+  public static final String OM_DEFAULT_NODE_ID = "om1";
 
   // Dummy OMNodeID for OM Clients to use for a non-HA OM setup
   public static final String OM_NODE_ID_DUMMY = "omNodeIdDummy";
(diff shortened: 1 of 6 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
```diff
@@ -343,6 +343,7 @@ private OzoneConsts() {
 
   // Default OMServiceID for OM Ratis servers to use as RaftGroupId
   public static final String OM_SERVICE_ID_DEFAULT = "omServiceIdDefault";
+  public static final String OM_DEFAULT_NODE_ID = "om1";
 
   // Dummy OMNodeID for OM Clients to use for a non-HA OM setup
   public static final String OM_NODE_ID_DUMMY = "omNodeIdDummy";
```
arp7, line 349: I assume this `omNodeIdDummy` will only be use if someone disables Ratis?

arp7 (approved):
+1

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
hanishakoneru: No, this would be used for single node Ratis also.

arp7: Sorry I didn't understand then - how do we use both `OM_DEFAULT_NODE_ID` and `OM_NODE_ID_DUMMY ` with single node Ratis?

hanishakoneru: OM_NODE_ID_DUMMY was being used just as a dummy nodeId in OMFailoverProxyProvider for non-HA cluster.
But now that we have a default, we should remove the dummy. It would not affect the client in anyway as this is only for the non-HA and also. 
Thanks for pointing this out Arpit. I will update the PR.

arp7 (approved):
+1

PR #1795, opened 2021-01-14
HDDS-4688. Update Hadoop version to 3.2.2

symious:

- Changed hadoop.version from 3.2.1 to 3.2.2 in pom.xml
- Changed hadoop.version from 3.2.1 to 3.2.2 in other related files.

https://issues.apache.org/jira/browse/HDDS-4688

## How was this patch tested?

Manual test. Maven package passed.

### DIFF
--- a/hadoop-hdds/hadoop-dependency-client/README.md
+++ b/hadoop-hdds/hadoop-dependency-client/README.md
@@ -35,20 +35,20 @@ mvn dependency:tree
 [INFO]
 [INFO] --- maven-dependency-plugin:3.0.2:tree (default-cli) @ hadoop-hdds-hadoop-dependency-client ---
 [INFO] org.apache.hadoop:hadoop-hdds-hadoop-dependency-client:jar:1.0.0
-[INFO] +- org.apache.hadoop:hadoop-annotations:jar:3.2.1:compile
+[INFO] +- org.apache.hadoop:hadoop-annotations:jar:3.2.2:compile
 [INFO] |  \- jdk.tools:jdk.tools:jar:1.8:system
-[INFO] +- org.apache.hadoop:hadoop-common:jar:3.2.1:compile
+[INFO] +- org.apache.hadoop:hadoop-common:jar:3.2.2:compile
 [INFO] |  +- org.apache.httpcomponents:httpclient:jar:4.5.2:compile
 [INFO] |  |  \- org.apache.httpcomponents:httpcore:jar:4.4.4:compile
 [INFO] |  +- org.apache.commons:commons-configuration2:jar:2.1.1:compile
 [INFO] |  +- com.google.re2j:re2j:jar:1.1:compile
 [INFO] |  +- com.google.protobuf:protobuf-java:jar:2.5.0:compile
-[INFO] |  +- org.apache.hadoop:hadoop-auth:jar:3.2.1:compile
+[INFO] |  +- org.apache.hadoop:hadoop-auth:jar:3.2.2:compile
 [INFO] |  +- com.google.code.findbugs:jsr305:jar:3.0.0:compile
 [INFO] |  +- org.apache.htrace:htrace-core4:jar:4.1.0-incubating:compile
 [INFO] |  +- org.codehaus.woodstox:stax2-api:jar:3.1.4:compile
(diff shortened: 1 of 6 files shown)


hadoop-hdds/hadoop-dependency-client/README.md
```diff
@@ -35,20 +35,20 @@ mvn dependency:tree
 [INFO]
 [INFO] --- maven-dependency-plugin:3.0.2:tree (default-cli) @ hadoop-hdds-hadoop-dependency-client ---
 [INFO] org.apache.hadoop:hadoop-hdds-hadoop-dependency-client:jar:1.0.0
-[INFO] +- org.apache.hadoop:hadoop-annotations:jar:3.2.1:compile
+[INFO] +- org.apache.hadoop:hadoop-annotations:jar:3.2.2:compile
```
adoroszlai, line 38: This dependency tree is for Ozone 1.0.0, which still uses Hadoop 3.2.1.  I think we can keep it as is.  Alternatively, we might update the whole dependency tree (I'm sure there are other version upgrades since 1.0.0).  But I suggest doing that only right before 1.1.0 release.

hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop32/.env
```diff
@@ -16,6 +16,6 @@
 
 HDDS_VERSION=@hdds.version@
 HADOOP_IMAGE=flokkr/hadoop
-HADOOP_VERSION=3.2.1
+HADOOP_VERSION=3.2.2
```
adoroszlai, line 19: @elek this change [needs](https://github.com/apache/ozone/pull/1795/checks?check_run_id=1699324055#step:7:442) a new `flokkr/hadoop:3.2.2` image, too.  I have [opened a PR](https://github.com/flokkr/docker-hadoop/pull/8) for that.  (Or we can keep MR acceptance tests on Hadoop 3.2.1.)

adoroszlai (commented):
Thanks @symious for taking this up.

`TestStorageContainerManagerHttpServer` failure seems related.  Can you please check?

symious:
@adoroszlai Thanks for the comment. 
Checking the error of "TestStorageContainerManagerHttpServer".

symious:
The error of "TestStorageContainerManagerHttpServer" was caused by the change of MetricsSystemImp.java. The commit is [HADOOP-17081|https://github.com/apache/hadoop/commit/2f500e4635ea4347a55693b1a10a4a4465fe5fac#diff-299fd2f06692c12bf10a572a87be5010de9c4cd2563a3d366ba0d9b76f038d3aR277]
I think it's a bug here? Maybe we need to check if the config is null before invoking _registerSink_?

symious:
Have rasied a ticket for this issue in project Hadoop(https://issues.apache.org/jira/plugins/servlet/mobile#issue/HADOOP-17468). Maybe we need to wait before other actions?

hadoop-ozone/dist/src/main/compose/ozone-mr/hadoop32/.env
elek: Thanks, I merged the pr. Let's give it a few minutes and we can restart the builds.

adoroszlai: Thanks @elek.  There is still no new image built.  May be caused by the [problem with Hadoop 3.2.2 mirrors](https://mail-archives.apache.org/mod_mbox/hadoop-general/202101.mbox/%3C07913537-6683-414b-94c9-6846d0967db0%40www.fastmail.com%3E).

symious:
@adoroszlai Manually invoke DefaultMetricsSystem.initialize(...) to solve the NPE of _**config**_.

symious:
@adoroszlai Test passed in the local environment, could you have a check?

hadoop-hdds/hadoop-dependency-client/README.md
symious: Have revert the changes.

adoroszlai (approved):
Thanks @symious for updating the patch and @elek for creating the necessary docker image.

symious:
@adoroszlai Thanks for the review.

PR #1798, opened 2021-01-14
HDDS-4645. Starting OM with the --upgrade flag should delete the prepare marker file.

errose28:

When an OM is put in prepare mode, a marker file is written to disk so that it remains in prepare mode upon restart. After an OM has been upgraded, it can be restarted with a --upgrade flag to automatically delete the marker file and take the OM out of prepare mode.

This patch incorporates work in progress changes from HDDS-4610 to make sure that the commit transaction after prepare updates the transaction index in the OM DB, so it matches the index of the prepare marker file when the OM comes back up.

HDDS-4645

## Testing

Acceptance test added.

The added test uses a similar approach to the existing *upgrade* acceptance test to restart the OMs. All containers in the cluster write their state to mounted volumes, and nodes are restarted by restarting their containers. These restarted containers have the previous on disk state still persistent in the volume when it is mounted on restart.

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozone-om-prepare/.env
+++ b/hadoop-ozone/dist/src/main/compose/ozone-om-prepare/.env
@@ -1,0 +1,21 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+HDDS_VERSION=${hdds.version}
+OZONE_RUNNER_VERSION=${docker.ozone-runner.version}
+OZONE_IMAGE=apache/ozone-runner:${docker.ozone-runner.version}
+OZONE_DIR=/opt/hadoop
+OZONE_VOLUME=.
--- a/hadoop-ozone/dist/src/main/compose/ozone-om-prepare/.env
+++ b/hadoop-ozone/dist/src/main/compose/ozone-om-prepare/.env
@@ -14,8 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-HDDS_VERSION=${hdds.version}
-OZONE_RUNNER_VERSION=${docker.ozone-runner.version}
-OZONE_IMAGE=apache/ozone-runner:${docker.ozone-runner.version}
+HDDS_VERSION=1.1.0-SNAPSHOT
+OZONE_RUNNER_VERSION=20200625-1
+OZONE_IMAGE=apache/ozone-runner:20200625-1
 OZONE_DIR=/opt/hadoop
 OZONE_VOLUME=.
+OM_HA_ARGS=--
--- a/hadoop-ozone/dist/src/main/compose/ozone-om-prepare/.env
+++ b/hadoop-ozone/dist/src/main/compose/ozone-om-prepare/.env
@@ -19,4 +19,7 @@ OZONE_RUNNER_VERSION=20200625-1
 OZONE_IMAGE=apache/ozone-runner:20200625-1
 OZONE_DIR=/opt/hadoop
 OZONE_VOLUME=.
+# Inidcate no arguments to the OM.
+# This variable must be set to some non-empty value, or docker compose will
+# expand it to an empty string and pass that to the OM as an argument.
 OM_HA_ARGS=--
--- a/hadoop-ozone/dist/src/main/compose/ozone-om-prepare/test.sh
+++ b/hadoop-ozone/dist/src/main/compose/ozone-om-prepare/test.sh
@@ -39,22 +39,17 @@ export OZONE_DIR=/opt/hadoop
 # shellcheck source=/dev/null
 source "${COMPOSE_DIR}/../testlib.sh"
 
-# prepare pre-upgrade cluster
-#export OM_HA_ARGS='--upgrade'
+# Write data and prepare cluster.
 start_docker_env
-docker-compose ps
-echo "Before prepare test"
 execute_robot_test scm omha/om-prepare.robot
-echo "After prepare test"
-#KEEP_RUNNING=false stop_docker_env
-#stop_docker_env
+KEEP_RUNNING=false stop_docker_env
 
 # re-start cluster with --upgrade flag to take it out of prepare.
-#export OM_HA_ARGS='--upgrade'
-#export OZONE_KEEP_RESULTS=true
-#start_docker_env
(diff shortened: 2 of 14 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManagerStarter.java
```diff
@@ -99,29 +99,25 @@ public void initOm()
...
-//    commonInit();
-//    boolean result = receiver.prepareForUpgrade(conf);
-//    if (!result) {
-//      throw new Exception("Prepare OM For Upgrade failed.");
-//    }
-//    System.exit(0);
-//  }
+  /**
+   * This function implements a sub-command to allow the OM to be
+   * "prepared for upgrade".
```
avijayanhwx, line 104: Wrong Javadoc?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManagerStarter.java
```diff
@@ -99,29 +99,25 @@ public void initOm()
...
+   */
+  @CommandLine.Command(name = "--upgrade",
```
avijayanhwx, line 106: Can we add an alias for downgrade as well?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -343,6 +347,7 @@ public void pause() {
    */
   public void unpause(long newLastAppliedSnaphsotIndex,
       long newLastAppliedSnapShotTermIndex) {
+    LOG.info("Unpausing State machine at {}", this.ozoneManager.getOMNodeId());
```
avijayanhwx, line 350: Left over log line?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
```diff
@@ -376,7 +381,15 @@ public long takeSnapshot() throws IOException {
...
     snapshotInfo.updateTermIndex(lastTermIndex.getTerm(),
         lastAppliedIndex);
+    OMTransactionInfo build = new OMTransactionInfo.Builder()
+        .setTransactionIndex(lastAppliedIndex)
+        .setCurrentTerm(lastTermIndex.getTerm()).build();
+    Table<String, OMTransactionInfo> txnInfoTable =
+        ozoneManager.getMetadataManager().getTransactionInfoTable();
+    txnInfoTable.put(TRANSACTION_INFO_KEY, build);
     ozoneManager.getMetadataManager().getStore().flushDB();
+    LOG.info("OM TransactionInfo Table : {}",
```
avijayanhwx, line 391: Left over log line?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManagerStarter.java
errose28: Yes, forgot to change it, thanks.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
errose28: Yes, removed.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerStateMachine.java
errose28: Yes, removed.

hadoop-ozone/dist/src/main/compose/ozone-om-prepare/test.sh
```diff
@@ -0,0 +1,55 @@
...
+# shellcheck source=/dev/null
+source "${COMPOSE_DIR}/../testlib.sh"
+
+# Write data and prepare cluster.
+start_docker_env
+execute_robot_test scm omha/om-prepare.robot
+KEEP_RUNNING=false stop_docker_env
+
+# re-start cluster with --upgrade flag to take it out of prepare.
+export OM_HA_ARGS='--upgrade'
```
avijayanhwx, line 48: Do we need to add a test case to verify that a regular restart of OM does not bring it out of "prepared" state?

avijayanhwx (commented):
Thanks for addressing the review comments. Approach looks good to me. A few more comments inline after reviewing the acceptance test.

hadoop-ozone/dist/src/main/compose/ozone-om-prepare/test.sh
```diff
@@ -0,0 +1,55 @@
...
+  fi
+fi
+
+export OZONE_DIR=/opt/hadoop
+# shellcheck source=/dev/null
+source "${COMPOSE_DIR}/../testlib.sh"
+
+# Write data and prepare cluster.
+start_docker_env
+execute_robot_test scm omha/om-prepare.robot
```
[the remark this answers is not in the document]
errose28: Sure. I actually added this while doing HDDS-4611, I will put it in here so it gets in first.

hadoop-ozone/dist/src/main/compose/ozone-om-prepare/test.sh
errose28: Yes, I can add this.

hadoop-ozone/dist/src/main/compose/ozone-om-prepare/.env
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+HDDS_VERSION=1.1.0-SNAPSHOT
+OZONE_RUNNER_VERSION=20200625-1
+OZONE_IMAGE=apache/ozone-runner:20200625-1
+OZONE_DIR=/opt/hadoop
+OZONE_VOLUME=.
+OM_HA_ARGS=--
```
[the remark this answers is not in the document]
errose28: I should probably add a comment here to explain this. The command line args are passed to the OM in the docker compose file. When docker compose expands an environment variable and the variable is unset or empty, it will expand it to an empty string `''` and pass that literally to OM, which is an invalid argument.

To get around this, I used `--` as the default value instead of empty, which tells the CLI parser that optional flags are done and everything following is an argument. For example, `ls -- -l` gives: `ls: cannot access '-l': No such file or directory`. In this case, since there are no arguments on either side, it is used to indicate that the OM has no arguments, without using an empty string.

If you have a better way to implement this, let me know. This is just what I came up with to make it work.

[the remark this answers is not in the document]
avijayanhwx: @errose28 Thanks for the explanation, I am OK with this approach. We can revisit it later if needed.

avijayanhwx (approved):

avijayanhwx:
CI failure is unrelated. Merging the changes.

PR #1801, opened 2021-01-14
HDDS-4704. Add permission check in OMDBCheckpointServlet

smengcl:

Add permission check in `OMDBCheckpointServlet`

https://issues.apache.org/jira/browse/HDDS-4704

## How was this patch tested?

1. Add new UT in `TestOMDbCheckpointServlet` to test with different "login" users.
2. Pending manual testing on a cluster.


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
```diff
@@ -106,6 +132,37 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) {
       return;
     }
 
+    // Check ACL for dbCheckpoint only when global Ozone ACL is enable
+    if (om.getAclsEnabled()) {
+      final String remoteUser = request.getRemoteUser();
+      final java.security.Principal userPrincipal = request.getUserPrincipal();
+      if (userPrincipal == null) {
+        // Fallback to checking login user if userPrincipal is null.
+        // Note: In prod, a secure cluster would deploy Kerberos so this case
```
elek, line 141: Kerberos may or may not be used for web endpoints in secure cluster. There are multiple `AuthenticationHandler` which can be configured. Are you sure that userPrincipal can never be null in real case? Based on the Javadoc of `HttpServletRequest` `null` means un-authenticated. Wouldn't be more safe to update the unit test instead of doing this check in the producation code?

Is there any unit test which request this branch?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
```diff
@@ -84,11 +88,33 @@ public void init() throws ServletException {
         OMConfigKeys.OZONE_DB_CHECKPOINT_TRANSFER_RATE_KEY,
         OMConfigKeys.OZONE_DB_CHECKPOINT_TRANSFER_RATE_DEFAULT);
 
+    // TODO: Add servlet
```
elek, line 91: Do we need this?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
```diff
@@ -106,6 +132,37 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) {
...
+        }
+      } else {
+        final String userPrincipalName = userPrincipal.getName();
+        if (!checkAcls(userPrincipalName)) {
+          LOG.error("Permission denied: User principal '{}' doesn't have the "
+              + "permission to access /dbCheckpoint.", userPrincipalName);
+          response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+          return;
+        } else {
+          LOG.info("Granted user principal '{}' access to /dbCheckpoint.",
```
elek, line 160: I am not sure if we need this info level log. If you think it's required for security, we can add audit log event.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
```diff
@@ -84,11 +88,33 @@ public void init() throws ServletException {
...
 
+    // TODO: Add servlet
+//    getServletContext().addServlet(OMDBCheckpointAuthorizedServlet);
+
     if (transferBandwidth > 0) {
       throttler = new DataTransferThrottler(transferBandwidth);
     }
   }
 
+  private boolean checkAcls(String username) {
```
elek, line 99: NIT: For me the name is confusing. We check the admin permission not the ACLs. (But it's just a personal opinion)

elek (commented):
Thanks the patch @smengcl. Overall it looks good to me, but I have a few questions.

And one more: Can you please add at least two lines to the documentation to make it clear which users should be added to the admins?

I think it's a very tricky requirement, and we need to make it clear from day 1 what are the requirements...

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
smengcl: Hmm, when Kerberos is enabled, if `request.getUserPrincipal()` is null, I don't see a way to get the principal name (`request.getUserPrincipal().getName()`) to that I can check against. Therefore, I assumed when Kerberos is enabled, `request.getUserPrincipal()` should not be null. And when I test this chunk of code manually locally (unsecured), `request.getUserPrincipal()` is indeed null.

When Kerberos is not in use, the code just check with `request.getRemoteUser()`.

smengcl: I agree I should add some tests. I'm going to add some lines in the ozone secure robot test.

smengcl: > Is there any unit test which require this branch?

Yes. `TestOMDbCheckpointServlet#testDoGet`, which doesn't have Kerberos but have ACL enabled.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
smengcl: I will remove this

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
smengcl: Yep I will set it to debug level then

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
smengcl: You're right. renamed to `hasPermission`.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
smengcl: Looks like we had some robot tests for dbCheckpoint in `web.robot`:
https://github.com/apache/ozone/blob/4ad03188ed4fbbe8d6dce1e8e0c8d91518904fc1/hadoop-ozone/dist/src/main/smoketest/spnego/web.robot#L59-L60

in SPNEGO testing.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
```diff
@@ -47,12 +48,25 @@
...
 /**
  * Provides the current checkpoint Snapshot of the OM DB. (tar.gz)
+ *
+ * When Ozone ACL is enabled (`ozone.acl.enabled`=`true`), only users/principals
+ * configured in `ozone.administrator` (along with the user that starts OM,
+ * which automatically becomes an Ozone administrator but not necessarily in
+ * the config) are allowed to access this endpoint.
+ *
+ * If Kerberos is enabled, the principal should be appended to
+ * `ozone.administrator`, e.g. `testuser/scm@EXAMPLE.COM`
```
xiaoyuyao, line 67: NIT: can you change the testuser to scm as it is typically used in production deployment.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
```diff
@@ -106,6 +139,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) {
...
     }
 
+    // Check ACL for dbCheckpoint only when global Ozone ACL is enable
+    if (om.getAclsEnabled()) {
+      final java.security.Principal userPrincipal = request.getUserPrincipal();
+      if (userPrincipal == null) {
+        // Fallback to checking login user if getUserPrincipal returns null.
+        // Note: In prod, a secure cluster with Kerberos should not hit this
+        // branch. This is mostly here for UT and dev testing.
+        final String remoteUser = request.getRemoteUser();
```
xiaoyuyao, line 149: null check for the remoteUser

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
```diff
@@ -106,6 +139,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) {
...
+          LOG.error("Permission denied: Current login user '{}' does not have"
+              + " access to /dbCheckpoint.", remoteUser);
+          response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+          return;
+        }
+        LOG.debug("Granted login user '{}' access to /dbCheckpoint.",
+            remoteUser);
+      } else {
+        final String userPrincipalName = userPrincipal.getName();
+        if (!hasPermission(userPrincipalName)) {
```
xiaoyuyao, line 160: Can we refactor like below to dedup

String remoteUser = null;
final java.security.Principal userPrincipal = request.getUserPrincipal();
      if (userPrincipal == null) {
        remoteUser = request.getRemoteUser();
      } else {
        remoteuser = userPrincipal.getName();
     }
    if (remoteUser == null  || !hasPermission(remoteUser)) {
      LOG.error("Permission denied: User remoteUser '{}' does not have"
               " access to /dbCheckpoint.", remoteUser);
          response.setStatus(HttpServletResponse.SC_FORBIDDEN);
          return;
    }

xiaoyuyao (commented):
Thanks @smengcl  for working on this. The change LGTM, just few minor comments inline.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java
elek: > I agree. But since you mentioned Kerberos might not be in-use on a secure cluster, we would still need some sort of user name verification, no? If that's the case I can adjust the comment.

I think, it's fine as is, the only thing what I would suggest is to remove the unit test related part from the production code unless it's fully required (can be convinced, but let me know which test requires is). Just throw an exception when user is null.

Instead of this:
```
     final java.security.Principal userPrincipal = request.getUserPrincipal();
      if (userPrincipal == null) {
        // Fallback to checking login user if getUserPrincipal returns null.
        // Note: In prod, a secure cluster with Kerberos should not hit this
        // branch. This is mostly here for UT and dev testing.
 
        final String remoteUser = request.getRemoteUser();
         if (!hasPermission(remoteUser)) {
          LOG.error("Permission denied: Current login user '{}' does not have"
              + " access to /dbCheckpoint.", remoteUser);
          response.setStatus(HttpServletResponse.SC_FORBIDDEN);
          return;
        }
```

I would use this:
```
    final java.security.Principal userPrincipal = request.getUserPrincipal();
      if (userPrincipal == null) {
          LOG.error("Permission denied: Current login user '{}' does not have"
              + " access to /dbCheckpoint.", remoteUser);
          response.setStatus(HttpServletResponse.SC_FORBIDDEN);
          return;
        }
 }
 ```

Comment (smengcl) [reply hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java]: done

File: hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java (hunk 5 above)
Comment (smengcl) [reply hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java:67]: done

File: hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java (hunk 6 above)
Comment (smengcl) [reply hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java]: should be fine as the string isn't deferenced. just printed in the log.

File: hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java (hunk 7 above)
Comment (smengcl) [reply hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OMDBCheckpointServlet.java:153]: I have removed the fallback login user check as Marton suggested.

**Review by elek — CHANGES_REQUESTED** [review]
Thank you very much the update. @smengcl. Looks good to me.

One of my comments still not addressed:

> And one more: Can you please add at least two lines to the documentation to make it clear which users should be added to the admins?

If you wouldn't like to add it to the markdown docs, can you please explain how can it be fixed in the error log?

Like this:

```
LOG.error("Permission denied: User principal '{}' does not have"
          + " access to /dbCheckpoint. In case the Ozone services started with different users, the principal of other services should be added to the Ozone administrators. Please add {} user to the 'ozone.administrators' configuration value.", userPrincipalName, userPrincipalName);```

smengcl:
@elek Done.

elek (approved):
Thanks the update @smengcl (and the review @xiaoyuyao).

PR #1804, opened 2021-01-15
HDDS-4700. Add JUnit5 dependency for integration tests

adoroszlai:

JUnit5 makes it easier to [run tests multiple times](https://junit.org/junit5/docs/current/user-guide/#writing-tests-repeated-tests) via `@RepeatedTest(N)` annotation.  Adding JUnit5 dependency for integration tests would allow upgrading flaky tests one by one for easier debug.

Also upgrade a single test class to JUnit5 to be able to verify that such tests are run, and JUnit4 tests are still executed as well.

https://issues.apache.org/jira/browse/HDDS-4700

## How was this patch tested?

Regular CI - verified that both JUnit5 and JUnit4 tests are executed.

### DIFF
--- a/hadoop-ozone/integration-test/pom.xml
+++ b/hadoop-ozone/integration-test/pom.xml
@@ -104,6 +104,21 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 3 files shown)


hadoop-ozone/integration-test/pom.xml
```diff
@@ -103,6 +103,21 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
...
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
```
ayushtkn, line 120: Should we add `junit-platform-launcher` also?

adoroszlai: The doc says `junit-platform-launcher` is required only for older IDEs, but I don't know what's "old".  I had no problem running tests in IDEA 2020.2 without it.

ayushtkn: 2020.2 is new enough :-)
It would create problems for IDE bundled without junit-5 I guess, IntelliJ 2017 or before might get affected.
Do you any issues adding `junit-platform-launcher`? else having support for older IDE's won't hurt.

ayushtkn (approved):
LGTM

adoroszlai:
Thanks @ayushtkn for the review and suggestion.

PR #1807, opened 2021-01-15
HDDS-4430. OM failover timeout is too short

hanishakoneru:

The current OM has one second failover timeout. This is too short as any network hiccup, system I/O or JVM GC pause could easily trigger a failover. Increasing Ratis rpc timeout to 5 sec by default.
Also, Ratis does not have a leader election timeout. Hence, deprecating Ozone OM's leader election timeout.
Instead, setting the leader step down wait time to 30 sec.

https://issues.apache.org/jira/browse/HDDS-4430

## How was this patch tested?

No new tests

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1620,18 +1620,27 @@
 
   <property>
     <name>ozone.om.ratis.minimum.timeout</name>
-    <value>1s</value>
+    <value>5s</value>
     <tag>OZONE, OM, RATIS, MANAGEMENT</tag>
     <description>The minimum timeout duration for OM's Ratis server rpc.
     </description>
   </property>
 
   <property>
+    <name>ozone.om.ratis.leader.step.down.wait.time</name>
+    <value>30s</value>
+    <tag>OZONE, OM, RATIS, MANAGEMENT, DEPRECATED</tag>
+    <description>The time OM Ratis Leader waits before stepping down on loosing
+      majority heartbeats from followers.
+    </description>
+  </property>
+
+  <property>
     <name>ozone.om.leader.election.minimum.timeout.duration</name>
-    <value>1s</value>
-    <tag>OZONE, OM, RATIS, MANAGEMENT</tag>
-    <description>The minimum timeout duration for OM ratis leader election.
-      Default is 1s.
+    <value>5s</value>
+    <tag>OZONE, OM, RATIS, MANAGEMENT, DEPRECATED</tag>
+    <description>DEPRECATED. Leader election timeout uses ratis
+      rpc timeout which can be set via ozone.om.ratis.minimum.timeout.
     </description>
   </property>
 
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -1627,15 +1627,6 @@
   </property>
 
   <property>
-    <name>ozone.om.ratis.leader.step.down.wait.time</name>
-    <value>30s</value>
-    <tag>OZONE, OM, RATIS, MANAGEMENT, DEPRECATED</tag>
-    <description>The time OM Ratis Leader waits before stepping down on loosing
-      majority heartbeats from followers.
-    </description>
-  </property>
-
-  <property>
     <name>ozone.om.leader.election.minimum.timeout.duration</name>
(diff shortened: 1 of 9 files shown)


hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -1620,18 +1620,27 @@
...
   <property>
     <name>ozone.om.leader.election.minimum.timeout.duration</name>
-    <value>1s</value>
-    <tag>OZONE, OM, RATIS, MANAGEMENT</tag>
-    <description>The minimum timeout duration for OM ratis leader election.
-      Default is 1s.
+    <value>5s</value>
+    <tag>OZONE, OM, RATIS, MANAGEMENT, DEPRECATED</tag>
+    <description>DEPRECATED. Leader election timeout uses ratis
+      rpc timeout which can be set via ozone.om.ratis.minimum.timeout.
```
amaliujia, line 1643: OUt of curiosity:

so the deprecation of a  config entry is to put `DEPRECATED` in the description?

amaliujia (commented):
+1 overall LGTM

hadoop-hdds/common/src/main/resources/ozone-default.xml
hanishakoneru: I took reference from Hadoop. There is a DeprecationDelta we can add in OzoneConfiguration but since this config is not being replaced by a new config, I was not sure if it makes sense to add it there. 

@adoroszlai, thoughts? should this config be added to OzoneConfiguration#deprecated keys?

arp7: I don't think we need to add it to deprecated keys. This should not be a widely used config key.

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -1620,18 +1620,27 @@
 
   <property>
     <name>ozone.om.ratis.minimum.timeout</name>
-    <value>1s</value>
+    <value>5s</value>
```
arp7, line 1623: Should this be the same as `ozone.om.ratis.leader.step.down.wait.time`? IIUC, if leader is partitioned from its peers then a new leader can be elected within 5 seconds however the old leader won't step down for 30 seconds. So there is a large interval when the old leader can be serving reads thinking that it is still the leader.

hanishakoneru: I misunderstood leader.step.down.time. It sets the time an old leader, after stepping down, has to wait before instantiating another election. 
The leader step down on losing majority heartbeats is based on the server rpc min and max timeout only. Removed this setting in new commit.

arp7 (approved):

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerRatisServer.java
```diff
@@ -498,23 +498,6 @@ private RaftProperties newRaftProperties(ConfigurationSource conf) {
...
-        OMConfigKeys.OZONE_OM_LEADER_ELECTION_MINIMUM_TIMEOUT_DURATION_KEY,
-        OMConfigKeys.OZONE_OM_LEADER_ELECTION_MINIMUM_TIMEOUT_DURATION_DEFAULT
-            .getDuration(), leaderElectionMinTimeoutUnit);
-    final TimeDuration leaderElectionMinTimeout = TimeDuration.valueOf(
-        leaderElectionMinTimeoutduration, leaderElectionMinTimeoutUnit);
-    RaftServerConfigKeys.Rpc.setTimeoutMin(properties,
-        leaderElectionMinTimeout);
-    long leaderElectionMaxTimeout = leaderElectionMinTimeout.toLong(
-        TimeUnit.MILLISECONDS) + 200;
-    RaftServerConfigKeys.Rpc.setTimeoutMax(properties,
```
arp7, line 515: How are we setting the timeout after this change?

hanishakoneru: There is no specific config for leader election timeout in Ratis. It uses Rpc timeout only. 
Before also, for leader election timeout, actually `RaftServerConfigKeys.Rpc.setTimeoutMax()` was being called.

arp7 (approved):

PR #1810, opened 2021-01-17
HDDS-4651. Implement a sequence ID generator

amaliujia:

Design doc: https://docs.google.com/document/d/13qSwLuDBw7mytJtF9h2oauuLocFfr-XJSBdTfc1YvxM/edit#heading=h.toe54twhlpm2

https://issues.apache.org/jira/browse/HDDS-4651

## How was this patch tested?

UT


amaliujia:
R: @GlenGeng @runzhiwang @bshashikant @ChenSammi

amaliujia:
To replace existing UniqueID by this ID generator, tracking in https://issues.apache.org/jira/browse/HDDS-4712

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIDGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+  private final AtomicLong counter = new AtomicLong(0L);
+
+  public SequenceIDGenerator(long term) {
+    // move term to higher 30 bits and save it into curTermOnHigher30Bits.
+    curTermOnHigher30Bits = term << 34;
+  }
+
+  public long nextID() {
+    long l = counter.getAndIncrement();
+    return l & 0x3FFFFFFF | curTermOnHigher30Bits;
```
amaliujia, line 38: I have thought about whether we need a check here to see if the next unique ID exceeds the 34 bits (thus become non-unique). My current thought is no need to add such a check:

34 bits can give 1,735,166,787,583 (>1.7 trillion) numbers for a term. It is very very very less likely that another leader election won't happen for the specific term. Meanwhile, adding a check means we will need to do >1.7 trillion checks before hitting the rare case.

Combine both I think we don't need to do that check here (save the cost as the rare case will not happen).

GlenGeng: Thanks @amaliujia for the work !

The up limit of 34 bits is 17 billion, and I consider we have to do such a check, and need to exit SCM when it reaches the limit. Since customer might deploy a single-server raft cluster in the non-HA mode, there will be no leader election after SCM starts.

GlenGeng: and it should be `l & 0x3FFFFFFFF`, seems missing one F

amaliujia: Yes your are right that I have missed a `F`. 

Also add the check to prevent non unique ID generated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIDGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+    curTermOnHigher30Bits = term << 34L;
+  }
+
+  public long nextID() throws SCMException {
+    long l = counter.getAndIncrement();
+    long countOnLower34Bits = l & lower34BitsMask;
+    if (countOnLower34Bits != l) {
+      throw new SCMException(
+          String.format("ID generator generates a non unique id, " +
+              "term:{}, count:{} ", curTermOnHigher30Bits >> 34L, l),
```
lamber-ken, line 45: here, String format should use `%`, not `{}`.

amaliujia: Makes sense!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIDGenerator.java
```diff
@@ -0,0 +1,52 @@
...
+  public SequenceIDGenerator(long term) {
+    // move term to higher 30 bits and save it into curTermOnHigher30Bits.
+    curTermOnHigherBits = term << 34L;
+  }
+
+  public long nextID() throws SCMException {
+    long l = counter.getAndIncrement();
+    long countOnLower34Bits = l & LOWER_BITS_MASK;
+    if (countOnLower34Bits != l) {
+      throw new SCMException(
```
GlenGeng, line 45: I suggest to exit SCM, since except stepping down, there is no way to recover from this case.

amaliujia: @GlenGeng 

Any example to how to exit SCM?

GlenGeng: Please check the following code in ReplicationManager
```
    } catch (Throwable t) {
      // When we get runtime exception, we should terminate SCM.
      LOG.error("Exception in Replication Monitor Thread.", t);
      ExitUtil.terminate(1, t);
    }
```

amaliujia: Makes sense. By doing so we can enforce a leader re-election.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIDGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+    long countOnLower34Bits = l & LOWER_BITS_MASK;
+    if (countOnLower34Bits != l) {
+      Throwable t = new SCMException(
+          String.format("ID generator generates a non unique id, " +
+              "term:%s, count:%s", curTermOnHigherBits >> 34L, l),
+          SCMException.ResultCodes.INTERNAL_ERROR);
+      // When IDs are exhausted for a term, we should terminate SCM to force
+      // a leader re-election.
+      LOG.error("Facing a fatal problem in SequenceIDGenerator", t);
+      ExitUtil.terminate(1, t);
```
bshashikant, line 58: instead of exiting SCM, probably it would be better to just step down from leadership by catching the exception where the ID needs to be used. Shutting down the ratis server will itself enforce leader reelection while the current SCM can still participate and act as a follower.

amaliujia: @GlenGeng  what do you think? 

I guess when ID is exhausted, a SCM can only be serve as a read-only SCM?

amaliujia: How about let us solve this problem in https://issues.apache.org/jira/browse/HDDS-4712? 

Right now this PR just implements a ID gen like a library. In HDDS-4712 we will start to use this ID gen then we can think about when ID space is exhausted, how when and where to stop Ratis Server or SCM?

bshashikant: makes sense.

bshashikant (approved):

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIDGenerator.java
GlenGeng: This is a feasible way that when ID is exhausted,  calling `RaftServerImpl#close()` to shutdown the underlying RaftServer and then calling `RaftServerImpl#start()` to start it, so that current SCM can become a follower and rejoin current raft clusters. The risk is we might need to think through the corner cases we might meet while current SCM is still running.

Although exit SCM is obviously a clean way, it might shock customer. I propose to let Ratis to expose its `LeaderStateImpl#stepDown()`. We can create a new jira to track this issue.

amaliujia: The idea makes sense. 

We can apply these ideas in HDDS-4712

bshashikant:
Changes discussed will be taken up in HDDS-4712. Committing this.

PR #1811, opened 2021-01-17
HDDS-4683. ListKeys: do lookup in dir and file tables

rakeshadr:

This task is to perform #listKeys by searching the startKey, keyPrefix in Dir & File Tables.

https://issues.apache.org/jira/browse/HDDS-4683

## How was this patch tested?

Added UT.


linyiqun (commented):
@rakeshadr , haven't started the review. As I find there is one another similar task HDDS-4332(ListFileStatus - do lookup in directory and file tables). What's the major differences in these two task implementation? That would be helpful to review this PR if I know this.

rakeshadr:
> @rakeshadr , haven't started the review. As I find there is one another similar task [HDDS-4332](https://issues.apache.org/jira/browse/HDDS-4332)(ListFileStatus - do lookup in directory and file tables). What's the major differences in these two task implementation? That would be helpful to review this PR if I know this.

Thank you @linyiqun for your help in reviews. Following are the points I learned after comparing listKeys and listStatus:

1. ListKeys will skip the given startKey and won't be added to the finalList
2. ListKeys will add trailing slash for the directory key names. 
3. ListKeys will recursively list all the keys starting from keyPrefix and startKey.

```
For ex: FS structure like, 

/a
/a/b
/a/b/c
/a/b/d

/a/b/file1
a/b/c/file2
```

#listKeys("/a/b") returns array values like, [a/b/, a/b/file1, a/b/c/, a/b/d/, a/b/c/file2]

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -942,17 +944,159 @@ private boolean isKeyEmpty(OmKeyInfo keyInfo) {
...
+      // For example, startKey="dir1/dir2/dir3/file3" and keyPrefix="dir1/".
+      // Iterate then move startkey and parentKeyPrefixPath positions till the
+      // given keyPrefix and lists all its sup paths...
+
+      // Iteration-1:
+      // startKeyPos="dir1/dir2/dir3/file3", parentKeyPrefix="dir1/dir2/dir3"
+      // Iteration-2:
+      // startKeyPos="dir1/dir2/dir3", parentKeyPrefix="dir1/dir2"
+      // Iteration-3:
+      // startKeyPos="dir1/dir2", parentKeyPrefix="dir1"
```
linyiqun, line 1028: Not understand why we divided the startKey to search. In addition, why do we have to check if startKey is started with keyPrefix?
Can we reuse listOzoneFileStatusesV1 method like listKeysRecursivelyFromKeyPrefixV1 does? One difference is that now startKey is not empty.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2317,28 +2461,42 @@ private void listStatusFindKeyInTableCache(
...
+    return listOzoneFileStatusesV1(recursive, startKey, 0, numEntries,
+            clientAddress, volumeName, bucketName, keyName, sortAddress,
+            false);
+  }
+
+  @NotNull
+  protected List<OzoneFileStatus> listOzoneFileStatusesV1(boolean recursive,
+      String startKey, int countEntries, long numEntries, String clientAddress,
+      String volumeName, String bucketName, String keyName,
+      boolean sortAddress, boolean listKeys) throws IOException {
```
linyiqun, line 2477: Can we update this as a private method, I see this is privately used in this class.

linyiqun (commented):
Thanks @rakeshadr  for the comments.
I take the initial review. Some places still not fully get the point. Please have a look.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -942,17 +944,159 @@ private boolean isKeyEmpty(OmKeyInfo keyInfo) {
     // snapshot of the data, so we don't need these locks at a higher level
     // when we iterate.
 
+    boolean preserveTrailingSlash = enableFileSystemPaths &&
+                    !OzoneManagerRatisUtils.isOmLayoutVersionV1();
     if (enableFileSystemPaths) {
-      startKey = OmUtils.normalizeKey(startKey, true);
-      keyPrefix = OmUtils.normalizeKey(keyPrefix, true);
+      startKey = OmUtils.normalizeKey(startKey, preserveTrailingSlash);
+      keyPrefix = OmUtils.normalizeKey(keyPrefix, preserveTrailingSlash);
```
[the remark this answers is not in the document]
rakeshadr: Thanks @linyiqun for the comments.

Yes, this logic is not required. I will remove this unnecessary `isOmLayoutVersionV1` check in the PR.

I was having a thinking that, in V1 there won't be any trailing slash in the "keyName". As we know, in existing code a trailing slash in the keyName represents a directory path in KeyTable.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -942,17 +944,159 @@ private boolean isKeyEmpty(OmKeyInfo keyInfo) {
...
+      keyList.add(omKeyInfo);
+      if (status.isDirectory()) {
+        // add trailing slash to represent directory
+        String dirName = omKeyInfo.getKeyName();
+        if (!dirName.endsWith("/")) {
+          omKeyInfo.setKeyName(dirName + OM_KEY_PREFIX);
+        }
+        if (!StringUtils.equals(dirName, keyPrefix)) {
+          listKeysRecursivelyFromKeyPrefixV1(volumeName, bucketName, dirName,
+                  maxKeys, keyList);
```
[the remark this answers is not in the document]
rakeshadr: `maxKeys` represents the limit and the batch size. During iteration and seeks, keyList.size() shouldn't exceed this limit. So, maxKeys is static.

`keyList.size()  <= maxKeys`

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Sure.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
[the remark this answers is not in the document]
linyiqun: Okay, one improvement I am thinking: can we do a check if keyList.size() already reach maxKeys, if reached, no need to call listKeysRecursivelyFromKeyPrefixV1 for dir. This can avoid unnecessary listKeysRecursivelyFromKeyPrefixV1 be invoked for subsequent sub-dirs.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
```diff
@@ -216,6 +230,116 @@ public void testLookupKey() throws Exception {
             true);
   }
 
+  @Test
+  public void testListKeysWithNotNormalizedPath() throws Exception {
```
linyiqun, line 234: Current test case only cover few cases of listKeysV1. Actually like startKey, maxKeys of listKeys are not tested. However, I also didn't find a good unit test added for original listKeys call.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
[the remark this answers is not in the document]
rakeshadr: Agreed, will do the check.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
rakeshadr: Yes, I will add more test cases. I'm thinking something like large dirs and sub-dirs/files.

Also, there is an open point: #listKeys("key-a-"), basically partial keyName and startKey. For example, actual keyName stored in DB are "key-a-10", "key-a-11", "key-a-12"..."key-z-10", "key-z-11", "key-z-12".
I will raise separate jira and work it on.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -942,17 +944,159 @@ private boolean isKeyEmpty(OmKeyInfo keyInfo) {
...
+                bucketName, startKeyPos, parentKeyPrefixPath, maxKeys,
+                keyList);
+
+        startKeyPos = parentKeyPrefixPath;
+        parentKeyPrefixPath = OzoneFSUtils.getParentDir(parentKeyPrefixPath,
+                true);
+      }
+    } else {
+      // TODO: Handle startKey not starting with keyPrefix. Should we suport
+      //  this case?
```
linyiqun, line 1043: If startKey does not start with key prefix, we can just compare the their value.

* If key prefix already behind the startKey, this search equals to the listKeys(key prefix , startKey='') as all entries with key prefix should behind the startKey.
* If key prefix not behind the startKey,  the empty list expected to be returned since startKey behind all the prefix key entries.

Does this handling make the sense here?

linyiqun (commented):
@rakeshadr , I read the doc of listKey algorithm. And some comments I left below. Please have a look.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Thanks @linyiqun for the example case. Please see my below reply.

> In Iteration-2, will it search for the dir1/dir2/dir3 again? I think it will be skipped as dir1/dir2/dir3 is the new startKey will be skipped.
Yes, it will skip the startKey from the resultList when startKey is not empty.

Secondly, the `listKeysV1()` algorithm is written like a template fashion with the following steps irrespective of startKey empty or non-empty:
step-1) It will fetch all the files under the given path.
step-2) It will fetch all the directories under the given path.

If the startKey is a directory, then the assumption is that all the files under this key was already visited in previous batch. So, in your example, you have passed `startKeyPos="dir1/dir2/dir3"`. Here it will check any nextString directory exists after the startKeyPos in DirTable. In your example case, we don't have any , so will return empty dir list. Assume, there was "dir1/dir2/dir4" exists, then it would have added to the result list.
Now, it won't go to FileTable and list all keys after `startKeyPos="dir1/dir2/dir3"`, instead it will move startKeyPos pointer to previous dir "dir1/dir2" and do a check ` !StringUtils.equals(parentStartKeyPath, keyPrefix)`. This return false.

So the resultList in this case is, `empty result list []`

Hope this explanation is clear to you.

mukul1987:
@rakeshadr I think the patch needs to be rebased, there are multiple unrelated changes.

rakeshadr:
Thanks a lot @linyiqun for the detailed reviewing on this patch.
 
On a second thought, I'd like to avoid more complex logic on the server side. I thought of re-using the existing proxy#listStatus() client API and making recursion logic at the client side. I've created a new [PR-1954](https://github.com/apache/ozone/pull/1954). I'd greatly appreciate your feedback on that. Thanks again!

Its just a first implementation and I think few more follow up jira tasks required to fully refine the client side logic by introducing batch fetching on levels incrementally etc etc. Will plan that work one-by-one.

linyiqun:
>On a second thought, I'd like to avoid more complex logic on the server side. I thought of re-using the existing proxy#listStatus() client API and making recursion logic at the client side. I've created a new PR-1954. I'd greatly appreciate your feedback on that. Thanks again!

 @rakeshadr . Will have a look recent days.

PR #1812, opened 2021-01-17
HDDS-4713. Add build-helper-maven-plugin to interface-client module

lamber-ken:

In ozone/interface-client module, sources files are generated by protobuf  plugin. But when other modules depend on this module, idea always prompts "Cannot resolve symbol" message.

image

https://issues.apache.org/jira/browse/HDDS-4713

## How was this patch tested?

Intellig IDEA.

### DIFF
--- a/hadoop-ozone/interface-client/pom.xml
+++ b/hadoop-ozone/interface-client/pom.xml
@@ -65,6 +65,26 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
           </execution>
         </executions>
       </plugin>
+
+      <!-- Adding protobuf generated classes to build classpath. -->
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>add-source</id>
+            <phase>generate-sources</phase>
+            <goals>
+              <goal>add-source</goal>
+            </goals>
+            <configuration>
+              <sources>
+                <source>${project.build.directory}/generated-sources/protobuf/java</source>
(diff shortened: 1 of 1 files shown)


amaliujia:
I found if run `mvn install` then the Intellij will be able to find such proto classes.

Does this `build-helper-maven-plugin` skip the step of `mvn install` in command line?

lamber-ken:
> I found if run `mvn install` then the Intellij will be able to find such proto classes.
> 
> Does this `build-helper-maven-plugin` help skipping the step of `mvn install` in command line (thus better)?

Hi, @amaliujia, `mvn install` seams not work in my local machines(at least two computers, withing Intellij IDEA Ultimate 2020.3). 

When using `build-helper-maven-plugin`, Intellij Idea works ok (need run mvn package )

amaliujia:
Actually I guess there are two things you can try:

1. Mark the directory as the generated source by intellij
截屏2021-01-18 上午12 40 24
截屏2021-01-18 上午12 40 42


2. It is possible that the intellij has marked the generated source, however the file size parameter is too small thus intellij does not load the specific large class/java file (the generated proto file sometimes can be very large).

lamber-ken:
Thanks for your detail @amaliujia . You're right, we need increate the `idea.max.intellisense.filesize`. 
By the way, if we use `build-helper-maven-plugin`, the idea will automatically recognizes the source files.

amaliujia:
Though I haven't checked what `build-helper-maven-plugin` does, I believe Intellij does not require to manually mark those source directories (at least on version 2020) :-)

lamber-ken:
ok, thanks @amaliujia : )

PR #1813, opened 2021-01-18
HDDS-4714. OzoneQuota change BYTES to B

captainzmc:

OzoneQuota change BYTES to B. Make it easier to use.

https://issues.apache.org/jira/browse/HDDS-4714

## How was this patch tested?

UT added

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
@@ -549,7 +549,7 @@ public class TestOzoneShellHA {
             .getQuotaInNamespace());
 
     // Test --quota option.
-    args = new String[]{"volume", "create", "vol1", "--quota", "100BYTES"};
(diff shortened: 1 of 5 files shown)


amaliujia:
LGTM

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
```diff
@@ -687,7 +687,7 @@ public void testShQuota() throws Exception {
 
     // Test set bucket spaceQuota or nameSpaceQuota to normal value.
     String[] bucketArgs3 = new String[]{"bucket", "setquota", "vol4/buck4",
-        "--space-quota", "1000BYTES"};
+        "--space-quota", "1000B"};
```
amaliujia, line 690: Does quota support default unit now? like if only use 1000 it will use `B` as the default unit?

captainzmc: At present does not support, need to let the user display the specified unit.

ChenSammi:
+1

PR #1814, opened 2021-01-18
HDDS-4650. Refactor ACL operation error handling

sky76093016:

Use the exception to replace the boolean failure.

https://issues.apache.org/jira/browse/HDDS-4650

## How was this patch tested?

No test.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/exceptions/OMException.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/exceptions/OMException.java
@@ -233,7 +233,9 @@ public class OMException extends IOException {
 
     QUOTA_EXCEEDED,
 
-    QUOTA_ERROR
+    QUOTA_ERROR,
+
+    ACL_SET_FAILED
 
   }
 }
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAWithACL.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerHAWithACL.java
@@ -35,6 +35,7 @@ import static org.apache.hadoop.ozone.OzoneAcl.AclScope.DEFAULT;
 import static org.apache.hadoop.ozone.security.acl.IAccessAuthorizer.ACLIdentityType.USER;
 import static org.apache.hadoop.ozone.security.acl.IAccessAuthorizer.ACLType.READ;
 import static org.apache.hadoop.ozone.security.acl.IAccessAuthorizer.ACLType.WRITE;
+import static org.junit.Assert.assertThrows;
 
 /**
  * Test Ozone Manager ACL operation in distributed handler scenario.
@@ -286,8 +287,8 @@ public class TestOzoneManagerHAWithACL extends TestOzoneManagerHA {
     Assert.assertTrue(containsAcl(userAcl, acls));
 
     // Add an already existing acl.
-    addAcl = objectStore.addAcl(ozoneObj, userAcl);
-    Assert.assertFalse(addAcl);
+    OzoneAcl finalUserAcl = userAcl;
+    assertThrows(OMException.class, () -> objectStore.addAcl(ozoneObj, finalUserAcl));
 
     // Add an acl by changing acl type with same type, name and scope.
     userAcl = new OzoneAcl(USER, remoteUserName,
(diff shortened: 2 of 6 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/VolumeManagerImpl.java
```diff
@@ -452,7 +452,7 @@ public boolean addAcl(OzoneObj obj, OzoneAcl acl) throws IOException {
         if (LOG.isDebugEnabled()) {
           LOG.debug("Add acl failed.", ex);
         }
-        return false;
+        throw new OMException(volume, ResultCodes.ACL_ADD_FAILED);
```
ayushtkn, line 455: Why this is only done for volume not for bucket and keys?
only in case of Volume there would be this Result Code, but won't be there for bucket and keys.
A general observation for bucket and keys, if the ACL entry exists already it returns false, but I think in case of Volume there is nothing of that sort. In fact it looks like post this change it can either return a true or an exception

sky76093016: Thank you for your suggestion. I did not notice the bucket and keys and will try to modify it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
```diff
@@ -431,10 +432,11 @@ public boolean addAcl(OzoneObj obj, OzoneAcl acl) throws IOException {
             BUCKET_NOT_FOUND);
       }
 
-      changed = bucketInfo.addAcl(acl);
-      if (changed) {
-        metadataManager.getBucketTable().put(dbBucketKey, bucketInfo);
+      if (!bucketInfo.addAcl(acl)) {
+        throw new OMException("Add acl operation failed" + bucket,
+            ACL_ADD_FAILED);
```
ayushtkn, line 437: will this be an incompatible change, instead of returning false throw exception.
I thought you tend to change the ex caought at L440

sky76093016: So @ayushtkn  think how to modify the incompatible changes is correct?Thank you.

ayushtkn: So, will returning false cause any issue?
We are catching exception bellow at L440, For that we can throw this ACL_ADD_FAILED exception. does that make sense?

sky76093016: What is @ChenSammi 's opinion on this modification?Thank you.

jojochuang: Ayush, IMHO strictly speaking, incompatible change implies public API method signature change. OMException is a subclass of IOException so it will be caught below.

That said, I don't think this is a refactor either. So maybe Ayush was misled.

ayushtkn: Couldn't check again, Wei-Chiu if you are convinced, Please proceed, nothing blocking from my side

xiaoyuyao: This seems to be a behavior change. Client previous expect false without exception for add existing acl or 
remove non-existing acl now need to handle the exception.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/BucketManagerImpl.java
```diff
@@ -419,7 +421,6 @@ public boolean addAcl(OzoneObj obj, OzoneAcl acl) throws IOException {
     }
```
xiaoyuyao, line 421: These are legacy addAcl/removeAcl/setAcl before HA is introduced. It is deprecated and not used. 
The actual addAcl/removeAcl/setAcl requests are handled in OMVolumeAclRequest/OMBucketAclRequest/OMKeyAclRequest classes.

xiaoyuyao (commented):
Thanks @sky76093016 for working on this. A few comments are added for the change.

cxorm:
Thank you @sky76093016 for the work.

Overall it looks well to me,
let's wait till all CI-checks pass.

cxorm:
Hi @sky76093016, the CI-checks failure seems to be related (in `testAddACL` part),
could you be so kind as to check it again ?

sky76093016:
Thanks for the reminder from @cxorm , I have observed this part.I am trying to modify it.

PR #1815, opened 2021-01-18
HDDS-4717. Fix TestOzoneFileSystemV1 and TestObjectStoreV1 cases

rakeshadr:

Few test cases failing when OM raits enabled to true. Also, delete file was wrongly updated keyName.

https://issues.apache.org/jira/browse/HDDS-4717

## How was this patch tested?
Updated existing UT cases.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -615,10 +615,17 @@ public class TestOzoneFileSystem {
     ArrayList<String> actualPathList = new ArrayList<>();
     if (rootItemCount != fileStatuses.length) {
       for (int i = 0; i < fileStatuses.length; i++) {
-        actualPaths.add(fileStatuses[i].getPath().getName());
+        boolean duplicate =
+                actualPaths.add(fileStatuses[i].getPath().getName());
+        if (!duplicate) {
+          LOG.info("Duplicate path:{} in FileStatusList",
+                  fileStatuses[i].getPath().getName());
+        }
         actualPathList.add(fileStatuses[i].getPath().getName());
       }
       if (rootItemCount != actualPathList.size()) {
+        LOG.info("actualPathsSize: {}", actualPaths.size());
+        LOG.info("actualPathListSize: {}", actualPathList.size());
         actualPaths.removeAll(paths);
         actualPathList.removeAll(paths);
         LOG.info("actualPaths: {}", actualPaths);
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
@@ -162,7 +162,7 @@ public class TestObjectStoreV1 {
             (KeyOutputStream) ozoneOutputStream.getOutputStream();
     long clientID = keyOutputStream.getClientID();
 
-    OmDirectoryInfo dirPathC = getDirInfo(volumeName, bucketName, parent);
+    OmDirectoryInfo dirPathC = getDirInfo(parent);
     Assert.assertNotNull("Failed to find dir path: a/b/c", dirPathC);
 
     // after file creation
@@ -214,7 +214,7 @@ public class TestObjectStoreV1 {
             (KeyOutputStream) ozoneOutputStream.getOutputStream();
     long clientID = keyOutputStream.getClientID();
 
-    OmDirectoryInfo dirPathC = getDirInfo(volumeName, bucketName, parent);
+    OmDirectoryInfo dirPathC = getDirInfo(parent);
     Assert.assertNotNull("Failed to find dir path: a/b/c", dirPathC);
 
     // after file creation
@@ -264,8 +264,7 @@ public class TestObjectStoreV1 {
             dirPathC.getObjectID(), true);
   }
 
-  private OmDirectoryInfo getDirInfo(String volumeName, String bucketName,
-      String parentKey) throws Exception {
+  private OmDirectoryInfo getDirInfo(String parentKey) throws Exception {
     OMMetadataManager omMetadataManager =
             cluster.getOzoneManager().getMetadataManager();
     long bucketId = TestOMRequestUtils.getBucketId(volumeName, bucketName,
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
@@ -17,17 +17,23 @@
 package org.apache.hadoop.ozone.om;
 
 import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hdds.client.ReplicationFactor;
 import org.apache.hadoop.hdds.client.ReplicationType;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.utils.db.Table;
-import org.apache.hadoop.hdds.utils.db.TableIterator;
 import org.apache.hadoop.ozone.MiniOzoneCluster;
+import org.apache.hadoop.ozone.OzoneConsts;
+import org.apache.hadoop.ozone.TestDataUtil;
 import org.apache.hadoop.ozone.client.ObjectStore;
 import org.apache.hadoop.ozone.client.OzoneBucket;
 import org.apache.hadoop.ozone.client.OzoneClient;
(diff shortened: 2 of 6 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -615,10 +615,17 @@ protected void testListStatusOnLargeDirectory() throws Exception {
     ArrayList<String> actualPathList = new ArrayList<>();
     if (rootItemCount != fileStatuses.length) {
       for (int i = 0; i < fileStatuses.length; i++) {
-        actualPaths.add(fileStatuses[i].getPath().getName());
+        boolean duplicate =
+                actualPaths.add(fileStatuses[i].getPath().getName());
+        if (!duplicate) {
+          LOG.info("Duplicate path:{} in FileStatusList",
```
mukul1987, line 621: Do we expect Duplicate paths in ListStatus ?

mukul1987 (changes requested):

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
rakeshadr: Thanks @mukul1987 for the reviews. No, there shouldn't be any duplicated path in the list, but I suspected some duplicate elements which makes the test case failure. So, I have added logs to capture this part.

In the latest run, its again failed and from the logs I saw that 5010 has been duplicated. I am checking it further to find out the cause. Interestingly it is always passing locally and failed to re-produce it. Do you have any thoughts?
```
2021-01-18 13:46:39,800 [Listener at 127.0.0.1/41009] INFO  ozone.TestOzoneFileSystem (TestOzoneFileSystem.java:testListStatusOnLargeDirectory(621)) - Duplicate path:5110 in FileStatusList
2021-01-18 13:46:39,802 [Listener at 127.0.0.1/41009] INFO  ozone.TestOzoneFileSystem (TestOzoneFileSystem.java:testListStatusOnLargeDirectory(627)) - actualPathsSize: 5111
2021-01-18 13:46:39,802 [Listener at 127.0.0.1/41009] INFO  ozone.TestOzoneFileSystem (TestOzoneFileSystem.java:testListStatusOnLargeDirectory(628)) - actualPathListSize: 5112
2021-01-18 13:46:39,809 [Listener at 127.0.0.1/41009] INFO  ozone.TestOzoneFileSystem (TestOzoneFileSystem.java:testListStatusOnLargeDirectory(631)) - actualPaths: []
2021-01-18 13:46:39,810 [Listener at 127.0.0.1/41009] INFO  ozone.TestOzoneFileSystem (TestOzoneFileSystem.java:testListStatusOnLargeDirectory(632)) - actualPathList: []
```

rakeshadr: Test is creating 0 to 5110 directories and `5110` directory is the last one. 
I am suspecting that `5110` exists in table cache and its again visible while iterating over DB, do you think so?

rakeshadr: I have updated the patch by using two separate maps - cacheDirMap and cacheFileMap to make unique OmKeyInfos. I have got a green build now.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
```diff
@@ -78,107 +90,143 @@ public static void init() throws Exception {
...
+    if (fileStatuses != null) {
+      Assert.assertEquals("Delete root failed!", 0, fileStatuses.length);
+    }
   }
 
   @Test
   public void testCreateKey() throws Exception {
-    String volumeName = "volume" + RandomStringUtils.randomNumeric(5);
-    String bucketName = "bucket" + RandomStringUtils.randomNumeric(5);
     String parent = "a/b/c/";
```
linyiqun, line 138: @rakeshadr , can you explain the change included in this unit test? I'm not directly get this.

rakeshadr: Following are the changes done in TestObjectStore.
1) Removed the logic of directly iterating over the file and openfile tables. I've replaced iterator with `table.get(key)`, this will fetch from the table cache, if exists.  Like I described, with ratis enabled(edit logs), table iterator won't have most recent entries and only table cache contains the most recent entries. This was the reason for test case failure. As part of HDDS-4498, now by default ratis is enabled.
2) `ClientID` is needed for preparing openFileKey, the open file format is <parentID>/fileName/<clientID>. So, I have exposed `keyOutputStream.getClientID();` method by adding @VisibleForTesting.
3) I'm creating volume and bucket `TestDataUtil.createVolumeAndBucket(cluster)` at the beginning of the test instead of creating at every unit test case in this class. This is basically for the cleanup convenience and you can see `tearDown()` has been added to delete all the dirs from the root.

rakeshadr: Thanks @linyiqun for the reviews.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -2312,17 +2311,40 @@ private void listStatusFindKeyInTableCache(
...
+     * StartKey should be added to the final listStatuses, so if we combine
+     * files and dirs into a single map then directory with lower precedence
+     * will appear at the top of the list even if the startKey is given as
+     * fileName.
+     *
+     * For example, startKey="a/file1". As per the seek order, first fetches
+     * all the files and then it will start seeking all the directories.
+     * Assume a directory name exists "a/b". With one map, the sorted list will
+     * be ["a/b", "a/file1"]. But the expected list is: ["a/file1", "a/b"],
+     * startKey element should always be at the top of the listStatuses.
```
linyiqun, line 2343: A little confused with this example. startKey is used to list keys which are behinds this key. "a/b" should not be expected in the return list, right? I also checked the logic in getDirectories method. It also doesn't use startKey to do the key comparison, [KeyManagerImpl.java#L2633](https://github.com/apache/ozone/blob/HDDS-2939/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L2633)

linyiqun (commented):
@rakeshadr ， PR looks good to me.  Only one thing that I have a different understanding of the startKey used here.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
linyiqun: Make the sense to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: Earlier the files and dirs were kept in single keyTable it was easy for the traversal and easy to do lexicographically greater than the startKey argument and RocksDB filters were helping us to maintain sorted order. 

But in new code, since we have two different tables(file & dir) we are iterating separately and programatically I've a search algorithm were it first fetches entries from the file and then fetches from dir Table. So, the combined list won't be in sorted order.
In V1 : the final list will be like [all files, all sub-dirs]. ["a/file1", "a/b"]. 

[Reference](https://github.com/apache/ozone/blob/HDDS-2939/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L2424)

I understand, its hard to understand. I think, we can ensure correctness in first phase. Later, in folow-up tasks we need to re-visit this part and try to optimize it further for better performance or so.

Doe this make sense to you?

Thanks again for the detailed reviews.

rakeshadr: For example, there are keys in FS:
a/b/c/file-1

linyiqun: Thanks for detailed explanation, @rakeshadr . Suppose there is a different meaning of the startKey under separate  file/dir table here. I will take a deep look for this.
In order not to unblock subsequent PRs, I will give my approval here.

linyiqun (approved):
+1 from me.
We can continue to discuss more detailed implementation in follow-up JIRAs.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
rakeshadr: @linyiqun Thanks a lot. Sure, will continue the discussions and refine it, if there are any gaps in the `startKey` semantics.

rakeshadr:
> +1 from me.
> We can continue to discuss more detailed implementation in follow-up JIRAs.

Thank you @linyiqun, @mukul1987 for the review comments. Will merge it shortly.

PR #1817, opened 2021-01-18
HDDS-1997. Support copy-source-if-(un)modified-since headers for MPU …

symious:
…key creation with copy

- Add the header check of copy-source-if-(un)modified-since.

https://issues.apache.org/jira/browse/HDDS-1997

## How was this patch tested?

Unit test.

### DIFF
--- a/hadoop-ozone/dist/src/main/smoketest/s3/MultipartUpload.robot
+++ b/hadoop-ozone/dist/src/main/smoketest/s3/MultipartUpload.robot
@@ -17,6 +17,7 @@
 Documentation       S3 gateway test with aws cli
 Library             OperatingSystem
 Library             String
+Library             DateTime
 Resource            ../commonlib.robot
 Resource            commonawslib.robot
 Test Timeout        5 minutes
@@ -274,13 +275,13 @@ Test Multipart Upload Put With Copy and range
 
                         Compare files           /tmp/part1        /tmp/part-result
 
-Test Multipart Upload Put With Copy and IfModifiedSince
+Test Multipart Upload Put With Copy and range with IfModifiedSince
     Run Keyword         Create Random file      10
-    ${beforeCreate} =   Execute and checkrc     date 0
+    ${curDate} =        Get Current Date
+    ${beforeCreate} =   Subtract Time From Date     ${curDate}  1 day
+    ${afterCreate} =    Add Time To Date        ${curDate}  1 day
 
     ${result} =         Execute AWSS3APICli     put-object --bucket ${BUCKET} --key ${PREFIX}/copyrange/source --body /tmp/part1
-
-    ${afterCreate} =    Execute and checkrc     date 0
 
     ${result} =         Execute AWSS3APICli     create-multipart-upload --bucket ${BUCKET} --key ${PREFIX}/copyrange/destination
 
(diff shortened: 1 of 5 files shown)


symious:
@elek Could you help to review this PR?

elek (changes requested):
Thanks the patch @symious (and sorry for the late review). I tested it with aws cli, and I think the headers can be strings not just longs:

```
aws s3api --endpoint=http://localhost:9878 create-bucket --bucket bucket1
aws --endpoint=http://localhost:9878  s3api put-object --bucket bucket1 --key key1 --body /tmp/qwe

aws s3api --endpoint=http://localhost:9878 create-multipart-upload --bucket bucket1 --key mptest
export UPLOAD_ID=ec6856e9-0c44-4b94-8055-67cf4f33f51e-105622884754849793
201364  aws s3api --endpoint=http://localhost:9878 upload-part-copy --bucket bucket1 --key mptest --upload-id $UPLOAD_ID --part-number 1 --copy-source bucket1/key1 --copy-source-if-modified-since 2020-01-27 
```

Last command is failed and the exception is visible in the log:

```
s3g_1       | Caused by: java.lang.NumberFormatException: For input string: "Mon, 27 Jan 2020 00:00:00 GMT"
s3g_1       | 	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
s3g_1       | 	at java.base/java.lang.Long.parseLong(Long.java:692)
s3g_1       | 	at java.base/java.lang.Long.valueOf(Long.java:1144)
s3g_1       | 	at org.apache.hadoop.ozone.s3.endpoint.ObjectEndpoint.checkCopySourceModificationTime(ObjectEndpoint.java:860)
s3g_1       | 	at org.apache.hadoop.ozone.s3.endpoint.ObjectEndpoint.createMultipartKey(ObjectEndpoint.java:596)
```

I think the content of the headers should be parsed.

Did you test it with another AWS tool? We may need to support both long and date.

Other parts looks good. The error message was too generic for me, but I found that AWS throws the same generic message ;-)

symious:
@elek Thanks for the comment.
I think I was doing the test with s3api too, but I was testing with Long only. Will update the patch to support String also.
For the error message, I used the error message from AWS, I will add more details in the patch.

elek:
>  I used the error message from AWS, 

Yes, I just learned it. It's fine to keep the current message to mimic the behavior of AWS.

Just FYI: You can create command line compatibility test (Test which executes AWS CLI and checks results) with modifying this file:

https://github.com/apache/ozone/blob/master/hadoop-ozone/dist/src/main/smoketest/s3/MultipartUpload.robot

(And executing `./test.sh` from `compose/ozone` folder from the build).

It's more like a good to have, not a must-have (IMHO), but if you are interested, you can check, and create other Jira/patch with improving acceptance tests.

symious:
Sure, thanks for the suggestions. 
I'm a little confused here, I have tried different typs of date format with was s3api, like "01-12-2020", "13-01-2020" and Long type timestamp, I found AWS is handling them all well. 
I am trying to use parseDate from common-lang3 which can parse date according to a list of format, but seems a little hard to list all possible DateFormat, do you have any suggestions on this?

symious:
Fixed the patch to check the date from the header.
Will raise another commit for the robot test. Not familiar with the robot test, need some time to work on it.

symious:
@elek The robot script is updated, could you have a check?

elek (approved):
+1 thanks for the update @symious 

Tested with the same method as before (using aws cli) and works well. (And it's already tested by robot tests, thanks for that...)

symious:
My pleasure.

PR #1819, opened 2021-01-18
HDDS-4721. Allow Default Space Quota Unit as Bytes

amaliujia:

Allow Default Space Quota Unit as Bytes.

https://issues.apache.org/jira/browse/HDDS-4721

## How was this patch tested?

UT

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
@@ -196,7 +196,6 @@ public final class OzoneQuota {
       // there might be no unit specified.
       if (size.equals("")) {
         size = uppercase;
-        currUnit = Units.B;
       }
       nSize = Long.parseLong(size);
     } catch (NumberFormatException e) {
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
@@ -182,7 +182,7 @@ public final class OzoneQuota {
         .replaceAll("\\s+", "");
     String size = "";
     long nSize = 0;
-    Units currUnit = Units.MB;
+    Units currUnit = Units.B;
 
     try {
       for (String quota : quotaList.getOzoneQuotaArray()) {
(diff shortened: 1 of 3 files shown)


amaliujia:
R: @captainzmc

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/shell/TestOzoneShellHA.java
```diff
@@ -704,6 +704,16 @@ public void testShQuota() throws Exception {
...
+    // test whether supports default quota unit as bytes.
+    String[] bucketArgs5 = new String[]{"bucket", "setquota", "vol4/buck4",
+        "--space-quota", "500"};
+    execute(ozoneShell, bucketArgs5);
+    out.reset();
+    assertEquals(500, objectStore.getVolume("vol4")
+        .getBucket("buck4").getQuotaInBytes());
+    assertEquals(100, objectStore.getVolume("vol4")
+        .getBucket("buck4").getQuotaInNamespace());
+
```
ayushtkn, line 716: Isn't this same as L704? If so, I don't think we changed anything on the namespace quota side, so no need of this?

amaliujia: O yes. That's more like a extra check to see the namespace quota remains the same. But removing it won't hurt anything.

amaliujia: Removed the namespace check as it is not relevant

captainzmc:
Thanks for @amaliujia‘s PR.  LGTM overall. 
Can you also add test case in TestOzoneRpcClientAbstract.testSetAndClrQuota() for test Java API?

amaliujia:
@captainzmc done!

captainzmc (approved):
LGTM +1
If there is no new comment, let's merge the PR tomorrow.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/OzoneQuota.java
```diff
@@ -193,6 +193,11 @@ public static OzoneQuota parseSpaceQuota(String quotaInBytes) {
           break;
         }
       }
+      // there might be no unit specified.
+      if (size.equals("")) {
+        size = uppercase;
+        currUnit = Units.B;
```
ayushtkn, line 199: if size is empty, currUnit will be Units.B already, do we need to re assign it here

amaliujia: Done!

ayushtkn (approved):
LGTM +1

ayushtkn:
Thanx @amaliujia for the contribution and @captainzmc for the review!!!

PR #1820, opened 2021-01-18
HDDS-4722. Creating RDBStore fails due to RDBMetrics instance race

jojochuang:

Make RDBMetrics#unRegister() synchronized. Accessing the instance object inside RDMBetrics#create() is synchronized, and therefore RDBMetrics#unRegister() should be made synchronized to protect the instance object.

https://issues.apache.org/jira/browse/HDDS-4722

## How was this patch tested?
Manually tested.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBMetrics.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/RDBMetrics.java
@@ -101,7 +101,7 @@ public class RDBMetrics {
     return numDBKeyMayExistMisses.value();
   }
 
-  public static void unRegister() {
+  public static synchronized void unRegister() {
     instance = null;
     MetricsSystem ms = DefaultMetricsSystem.instance();
     ms.unregisterSource(SOURCE_NAME);


adoroszlai (commented):
@jojochuang Synchronizing `unRegister` may fix the race condition, but I'm not sure using a single shared instance is OK in the first place.  Imagine the following steps:

1. `RDBStore` is created with new `RDBMetrics` instance
2. Another `RDBStore` is created, reuses `RDBMetrics` instance
3. One of the `RDBStores` is closed, shared `RDBMetrics` is unregistered
4. Other `RDBStore` is still used, but can it report events to the metrics system?

Shouldn't each `RDBStore` register its `RDBMetrics` instance with a unique name?

jojochuang:
Great question.... IMO we want a single RDBMetric instance. With the current architecture we will see thousands of containers per DataNode. Having metrics for each and every container... that seems too excessive to me. Plus, the metrics are kept only during the lifetime of a container. I feel like we want a singleton RDBMetric for the duration of the process.

@vivekratnavel  thoughts? you wrote the original code.

adoroszlai:
> I feel like we want a singleton RDBMetric for the duration of the process.

I also think this was the intention of introducing the shared instance.  In this case I think it should be reference counted, and the last one should clean up.

vivekratnavel:
Thanks @jojochuang and @adoroszlai for the patch and review! 

@avijayanhwx will also have a good context on this issue. If I remember correctly, we discussed some of the potential issues while making the RDBMetrics a shared instance, but I think we overlooked the example scenario that @adoroszlai has stated in his comment above. Please correct me if I am wrong.

@adoroszlai You are right about our intention for the shared Metrics instance. Before sharing the instance, we were getting the wrong metrics since each store was keeping its own count. I think it's time to think about all the scenarios thoroughly and redesign RDBMetrics if needed.

adoroszlai:
Let's merge this to make synchronized access consistent.  Then we can improve the shared instance with reference counting.

PR #1821, opened 2021-01-19
HDDS-4710. PipelinePlacementPolicy altered to randomly pick

Xushaohong:
fix Randomly-pick policy in chooseDatanodes.
Problem:
I was trying to configure a four-node cluster to test SCM HA, after configuring the ozone.datanode.pipeline.limit to 3 and 5 separately, the pipeline existed only chose the first three data nodes, the fourth node was never chosen.(They are equal nodes)

Solution:
The implementation of chooseNode() is replaced with a random policy as the same as the one in class SCMContainerPlacementRandom.
The other usages of chooseNode in class PipelinePlacementPolicy won't affect the topology & rack-awareness based node-choose policy

https://issues.apache.org/jira/browse/HDDS-4710

## How was this patch tested?

manually I tested the chooseDatanodes() method with a test method similar to testChooseNodeWithSingleNodeRack() where i added several equal nodes. The pipeline members are randomly generated.


Xushaohong:
@bshashikant @linyiqun  Please take a look

GlenGeng:
LGTM. Thanks for the work!

bshashikant (approved):

bshashikant:
Thanks @Xushaohong for the work.

PR #1823, opened 2021-01-20
HDDS-4725. Change metrics unit from nanosecond to millisecond

ferhui:

 Change metrics unit from nanosecond to millisecond uniformly 

Some metrics unit is nanosecond, and some is millisecond. The metrics are following

```
EventWatcherMetrics   ms

ProtocolMessageMetrics ns

CSMMetrics ns

ContainerMetrics ns

XceiverClientMetrics ns

OzoneManagerDoubleBufferMetrics ns

OzoneManagerSyncMetrics ns

ContainerCacheMetrics ms
```

https://issues.apache.org/jira/browse/HDDS-4725

## How was this patch tested?

manual tests

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/XceiverClientGrpc.java
@@ -468,7 +468,7 @@ public class XceiverClientGrpc extends XceiverClientSpi {
     final CompletableFuture<ContainerCommandResponseProto> replyFuture =
         new CompletableFuture<>();
     semaphore.acquire();
-    long requestTime = System.nanoTime();
+    long requestTime = System.currentTimeMillis();
     metrics.incrPendingContainerOpsMetrics(request.getCmdType());
     // create a new grpc stream for each non-async call.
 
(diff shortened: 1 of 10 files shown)


ferhui:
@ChenSammi @elek Could you please take a look?Thanks

amaliujia:
So the reason to convert nano to millis is because that most of the unit are already millis?

Could there a reason that nano as unit actually could be better?

ferhui:
> So the reason to convert nano to millis is because that most of the unit are already millis?
> 
> Could there be a reason that nano as unit actually could be better?

@amaliujia Thanks for comments. As mentioned in JIRA HDDS-4725, I think there are some reasons that it's better to use millis.

- Mostly the metrics are the order of millisecond.

- It's millis in hadoop, and It's easier to understand the metric.

- there are both nano and milli right now, we need to unify them.

amaliujia:
Thanks @ferhui for you explanation! Let's see if other people have comments on this.

amaliujia:
+1 LGTM

ferhui:
@amaliujia Thanks for review! I will merge this if no other comments later.

runzhiwang (approved):
LGTM

ferhui:
@runzhiwang Thanks for review & merge

PR #1825, opened 2021-01-20
HDDS-4726. Debug utility to export container data

elek:
JIRA: https://issues.apache.org/jira/browse/HDDS-4726

A new `ozone debug export-container --container=1` command which exports the container to a tar file.

The code path is the same as the closed container replication.

Can be useful for:

 1. Measure speed of export part of the closed container replication
 2. Debug container related problems locally

## How was this patch tested?

Executed on real Ozone cluster and measured the performance (and found HDDS-4687)


bshashikant:
Thanks @elek for the patch. The changes look good. Can we extend the tool to work against a datanode  UUID as well a specific volume path to probably read all containers from these. This will be useful to simulate container replication in case a disk goes bad/node goes down scenario and verify the replication performance?

elek:
> an we extend the tool to work against a datanode UUID as well a specific volume path to probably read all containers from these. This will be useful to simulate container replication in case a disk goes bad/node goes down scenario and verify the replication performance?

I am not fully sure what do you mean. This tool reads the data directly from the disk using the same layout as the datanode. It's always user only one datanode UUID, and all the configured disks are used. You can configure any sub-set of the available disks if you would like to use less disks.

end2end container replication can be used with https://github.com/apache/ozone/blob/master/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ClosedContainerReplicator.java

bshashikant:
> > an we extend the tool to work against a datanode UUID as well a specific volume path to probably read all containers from these. This will be useful to simulate container replication in case a disk goes bad/node goes down scenario and verify the replication performance?
> 
> I am not fully sure what do you mean. This tool reads the data directly from the disk using the same layout as the datanode. It's always user only one datanode UUID, and all the configured disks are used. You can configure any sub-set of the available disks if you would like to use less disks.
> 
> end2end container replication can be used with https://github.com/apache/ozone/blob/master/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/freon/ClosedContainerReplicator.java

Thanks @elek . I understand that, this can only be run on datanode and it makes sense. +1

elek:
Thanks the review @bshashikant, merging it based on your +1

PR #1827, opened 2021-01-20
HDDS-4486. Feature Config : Make proper enableFSPaths and OFS optimized flag combinations

rakeshadr:

This jira proposal is to simplify the ofs optimization cases. OM can support FS optimization only if both are TRUE(ofsOptimized=true && enableFSPaths=true)

https://issues.apache.org/jira/browse/HDDS-4486

## How was this patch tested?
Used existing unit tests.


rakeshadr:
Closed as there are many conflicts after branch rebase.

PR #1832, opened 2021-01-22
HDDS-4518. Add metrics around Trash Operations.

mukul1987:

This jira adds metrics for Trash Operations

https://issues.apache.org/jira/browse/HDDS-4518

## How was this patch tested?
Modified an existing test to cover this change.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -1236,15 +1236,17 @@ public class TestRootedOzoneFileSystem {
       }
     }, 1000, 180000);
 
-    Assert.assertTrue(getOMMetrics()
-        .getNumTrashDeletes() >= prevNumTrashDeletes);
+    // This condition should pass after the checkpoint
     Assert.assertTrue(getOMMetrics()
         .getNumTrashRenames() > prevNumTrashRenames);
     Assert.assertTrue(getOMMetrics()
-        .getNumTrashFilesDeletes() >= prevNumTrashFileDeletes);
-    Assert.assertTrue(getOMMetrics()
         .getNumTrashFilesRenames() > prevNumTrashFileRenames);
 
+    // This condition should succeed once the checkpoint directory is deleted
+    GenericTestUtils.waitFor(
+        () -> getOMMetrics().getNumTrashDeletes() > prevNumTrashDeletes
+            && getOMMetrics().getNumTrashFilesDeletes()
+            > prevNumTrashFileDeletes, 100, 180000);
     // Cleanup
     ofs.delete(trashRoot, true);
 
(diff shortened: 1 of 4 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashOzoneFileSystem.java
```diff
@@ -386,6 +393,7 @@ boolean processKeyPath(List<String> keyPathList) {
         OzoneManagerProtocolProtos.OMRequest omRequest =
             getRenameKeyRequest(src, dst);
         try {
+          ozoneManager.getMetrics().incNumTrashFilesRenames();
```
amaliujia, line 396: Out of curiosity: what is the difference between `incNumTrashFilesRenames` and ` incNumTrashRenames`, does the latter contains the former or vice verse?

sadanand48: `numTrashRenames` would count the renames on the directory while `numTrashFileRenames` would count renames on each file of the directory.

sadanand48 (commented):
+1 Looks good to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashOzoneFileSystem.java
amaliujia: Thanks for the clarification!

amaliujia:
+1 LGTM

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -1225,6 +1236,15 @@ public void testTrash() throws Exception {
       }
     }, 1000, 180000);
 
+    Assert.assertTrue(getOMMetrics()
+        .getNumTrashDeletes() >= prevNumTrashDeletes);
+    Assert.assertTrue(getOMMetrics()
+        .getNumTrashRenames() > prevNumTrashRenames);
+    Assert.assertTrue(getOMMetrics()
+        .getNumTrashFilesDeletes() >= prevNumTrashFileDeletes);
```
ayushtkn, line 1244: If for `prevNumTrashDeletes` and `prevNumTrashFileDeletes ` the equality hold true, That would mean the metrics didn't increment, which won't serve the purpose of the test. Can we ditch the `=` in the assertion and confirm the metric value gets incremented. Something like -
```
    GenericTestUtils.waitFor(
        () -> getOMMetrics().getNumTrashDeletes() > prevNumTrashDeletes
            && getOMMetrics().getNumTrashFilesDeletes()
            > prevNumTrashFileDeletes, 100, 180000);
```
Unless there is some catch here, in that case it should be ok..

ayushtkn (commented):
Thanx @mukul1987 for the work here.
There is a complain from checkstyle which doesn't seems to be introduced here. So should be fine.
Changes LGTM

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
mukul1987: Thanks, Excellent comment. Incorporated it.

ayushtkn (approved):
+1

PR #1833, opened 2021-01-22
HDDS-4660. Update OMTransactionInfo to TransactionInfo with functions added

amaliujia:

Requirements:
1. Rename OMTransactionInfo to TransactionInfo (to be used by both OM and SCM), same for the codec.
2. Add extra functions in https://github.com/apache/ozone/blob/HDDS-2823/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMTransactionInfo.java to OMTransactionInfo.

By doing so, we can use the TransactionInfo in HDDS-2823 and remove duplicate code

https://issues.apache.org/jira/browse/HDDS-4660

## How was this patch tested?

UT

### DIFF
--- a/hadoop-ozone/interface-storage/src/main/java/org/apache/hadoop/ozone/om/ratis/TransactionInfo.java
+++ b/hadoop-ozone/interface-storage/src/main/java/org/apache/hadoop/ozone/om/ratis/TransactionInfo.java
@@ -32,7 +32,7 @@ import static org.apache.hadoop.ozone.OzoneConsts.TRANSACTION_INFO_SPLIT_KEY;
 /**
  * TransactionInfo which is persisted to OM DB.
  */
-public final class OMTransactionInfo {
+public final class TransactionInfo {
 
   // Term associated with Ratis Log index in Ratis enabled cluster. In
   // non-Ratis cluster, term is set to -1.
@@ -42,7 +42,7 @@ public final class OMTransactionInfo {
   // non-Ratis cluster
   private long transactionIndex;
 
-  private OMTransactionInfo(String transactionInfo) {
+  private TransactionInfo(String transactionInfo) {
     String[] tInfo =
         transactionInfo.split(TRANSACTION_INFO_SPLIT_KEY);
     Preconditions.checkState(tInfo.length==2,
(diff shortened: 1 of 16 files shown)


amaliujia:
R: @hanishakoneru

amaliujia:
also R: @GlenGeng @runzhiwang

bshashikant:
@hanishakoneru , can you please check this ??

hadoop-ozone/interface-storage/src/main/java/org/apache/hadoop/ozone/om/ratis/TransactionInfo.java
```diff
@@ -52,11 +52,23 @@ private OMTransactionInfo(String transactionInfo) {
...
   }
 
-  private OMTransactionInfo(long currentTerm, long transactionIndex) {
+  private TransactionInfo(long currentTerm, long transactionIndex) {
     this.term = currentTerm;
     this.transactionIndex = transactionIndex;
   }
 
+  public boolean isInitialized() {
+    return transactionIndex == -1 && term == 0;
```
hanishakoneru, line 61: This function is not being called anywhere. 
Also, if initialized, then txIndex should not be equal to -1, right?

hadoop-ozone/interface-storage/src/main/java/org/apache/hadoop/ozone/om/ratis/TransactionInfo.java
```diff
@@ -52,11 +52,23 @@ private OMTransactionInfo(String transactionInfo) {
...
+  }
+
+  public int compareTo(TransactionInfo info) {
+    if (info.getTerm() == this.getTerm()) {
+      return this.getTransactionIndex() <= info.getTransactionIndex() ? -1 : 1;
+    } else {
+      return this.getTerm() < info.getTerm() ? -1 : 1;
+    }
```
hanishakoneru, line 69: I am not sure if I am missing something but couldn't find the usage for this function.
And shouldn't compareTo() return 0 if the TransactionInfo's are same?

hanishakoneru (commented):
Thanks @amaliujia for working on this. 
Patch LGTM overall. 
Did you test for compatibility? It does not look like there is incompatibility but better to be sure.

bshashikant (commented):
@amaliujia , can we move TransactionInfo class to hdds-common probbaly (or another suitable package), so that reusability in both SCM and OM becomes easier.

amaliujia:
@bshashikant @hanishakoneru 

I guess maybe I will directly make this change in HDDS-2823. Based on current feedback, it seems people got confused on unused functions, which is useful in HDDS-2823. Also moving stuff to `hdds-common` is happening in HDDS-2823 as well.

PR #1834, opened 2021-01-23
HDDS-4733. Remove leveldb from codebase

lamberken:

- LevelDB support was removed for OM, SCM, DN dbs. And users can no longer be able to use levedb through configuration options.
- `LevelDBStore.java` and `LevelDBStoreIterator.java` are marked as `@Deprecated`

Related links:
https://issues.apache.org/jira/browse/HDDS-3914

https://issues.apache.org/jira/browse/HDDS-4733

## How was this patch tested?

No new unit tests required.


elek (changes requested):
Thanks the patch @lamber-ken. It looks good to me. As far as I remember we agreed to keep the vendor neutral interface, but support only rocksdb for now.

Some minor comments:

Can you please remove the `org.fusesource.leveldbjni` dependency from the pom files if it's no longer required?

`MetadataStore` is an old interface and replaced by `DBStore` everywhere (the later can support multiple column families).

Which means that:

 1. `RocksDBStore`, `RocksDBStoreIterator`, `MetadataStore` and `MetadataStoreBuild` can also be removed (IMHO)
 2. `SQLCLI` can be removed because the functionality is replaced by `ozone debug ldb`.
 3. As `MetdataStore` is not required, the related Genesis tests can be removed `BencMark...` files

I would suggest including at least the dependency update in this patch(unless leveldb is really required by any other code, it can be removed from `pom.xml`)

1 + 3 and 2 can be included in this patch or followup patches... Depends on from your preference.

lamber-ken:
> Thanks the patch @lamber-ken. It looks good to me. As far as I remember we agreed to keep the vendor neutral interface, but support only rocksdb for now.
> 
> Some minor comments:
> 
> Can you please remove the `org.fusesource.leveldbjni` dependency from the pom files if it's no longer required?
> 
> `MetadataStore` is an old interface and replaced by `DBStore` everywhere (the later can support multiple column families).
> 
> Which means that:
> 
> 1. `RocksDBStore`, `RocksDBStoreIterator`, `MetadataStore` and `MetadataStoreBuild` can also be removed (IMHO)
> 2. `SQLCLI` can be removed because the functionality is replaced by `ozone debug ldb`.
> 3. As `MetdataStore` is not required, the related Genesis tests can be removed `BencMark...` files
> 
> I would suggest including at least the dependency update in this patch(unless leveldb is really required by any other code, it can be removed from `pom.xml`)
> 
> 1 + 3 and 2 can be included in this patch or followup patches... Depends on from your preference.

sure, will do, thanks.

elek (approved):
Sorry for the late answer, and thanks for the update. 

LGTM. I am fine with it, if we can have a green build.

elek:
Hi @lamberken. Are you planning to finish this PR?

adoroszlai:
> Hi @lamberken. Are you planning to finish this PR?

I would like to pick this up if @lamberken does not object.

adoroszlai:
I intended to update this PR by pushing to the original source branch, but it no longer exists, even the repo is gone.  I'll open a new PR instead.

PR #1835, opened 2021-01-23
HDDS-4734. Validate update state in PipelineStateMap.updatePipelineState

symious:

- Check the update state in PipelineStateMap.updatePipelineState

https://issues.apache.org/jira/browse/HDDS-4734

## How was this patch tested?

manual test.


elek (approved):
+1. Thanks for the patch @symious 

As far as I understood the `ContainerStateMap` change is not strictly required (logic should work even with repeated message) but I have no problem with adding the check.

`updatePipelineState` seems to be more dangerous, and I agree that it's the repsonsibility of the lower layer to validate preconditions.

(Note: The problem with `updatePipelineState` should be possible to be tested with unit test, which can make the reviews faster as makes easier the understanding ;-) )

symious:
@elek Thanks for the review. Will keep in mind to write unit test in future's work.

PR #1838, opened 2021-01-24
HDDS-4736. Intermittent failure in testExpiredCertificate

symious:

- Set the token expiration to a past time.

https://issues.apache.org/jira/browse/HDDS-4736

## How was this patch tested?

Manual test.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/security/TestOzoneBlockTokenSecretManager.java
+++ b/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/security/TestOzoneBlockTokenSecretManager.java
@@ -329,8 +329,8 @@ public class TestOzoneBlockTokenSecretManager {
       KeyPair pair, String algorithm) throws CertificateException,
       IllegalStateException, IOException, OperatorCreationException {
     Date from = new Date();
-    Date to = from;
-    return generateTestCert(dn, pair, algorithm, from, to);
+    // Set end date same as start date to make sure the cert is expired.
+    return generateTestCert(dn, pair, algorithm, from, from);
   }
 
   private X509Certificate generateNotValidYetCert(String dn,
--- a/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/security/TestOzoneBlockTokenSecretManager.java
+++ b/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/security/TestOzoneBlockTokenSecretManager.java
@@ -328,7 +328,7 @@ public class TestOzoneBlockTokenSecretManager {
   private X509Certificate generateExpiredCert(String dn,
       KeyPair pair, String algorithm) throws CertificateException,
       IllegalStateException, IOException, OperatorCreationException {
-    Date from = new Date();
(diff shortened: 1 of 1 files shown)


hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/security/TestOzoneBlockTokenSecretManager.java
```diff
@@ -328,7 +328,7 @@ public void testNetYetValidCertificate() throws Exception {
   private X509Certificate generateExpiredCert(String dn,
       KeyPair pair, String algorithm) throws CertificateException,
       IllegalStateException, IOException, OperatorCreationException {
-    Date from = new Date();
+    Date from = new Date(Instant.now().toEpochMilli() - 100L);
     Date to = new Date(from.getTime() + 100L);
     return generateTestCert(dn, pair, algorithm, from, to);
```
ayushtkn, line 333: Can we just do -
`` generateTestCert(dn, pair, algorithm, from, from);``
Means eliminate the `to` and pass `from` only instead, rather than this?

symious: Thanks for the suggestion, have set endDate same as startDate.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/security/TestOzoneBlockTokenSecretManager.java
```diff
@@ -329,7 +329,7 @@ private X509Certificate generateExpiredCert(String dn,
       KeyPair pair, String algorithm) throws CertificateException,
       IllegalStateException, IOException, OperatorCreationException {
     Date from = new Date();
-    Date to = new Date(from.getTime() + 100L);
+    Date to = from;
     return generateTestCert(dn, pair, algorithm, from, to);
```
ayushtkn, line 333: you can directly pass `from` instead of `to` in `generateTestCert`
something like -  ```generateTestCert(dn, pair, algorithm, from, from);```
no need to have
``Date to = from;``

symious: Yes, but I think it may be more readable to keep the fields of "from" and "to"?

ayushtkn: For readability better to add a comment explaining why we are passing `from` for both arguments.
Though trivial assigning unnecessary would have cost and isn't a good practice IMO

symious: Got it. Have added a comment for readability.

adoroszlai (approved):

symious:
@adoroszlai @ayushtkn Thanks for the review.

ayushtkn (approved):

ayushtkn:
Thanx @symious for the contribution and @adoroszlai for the review!!!

PR #1840, opened 2021-01-25
HDDS-4389.Disable Rocksdb caching when performing Trash specific oper…

sadanand48:
…ations

This change would avoid caching keys in the keyTable that are renamed to Trash making an assumption that a large proportion of keys that are moved to trash would be deleted eventually .

https://issues.apache.org/jira/browse/HDDS-4389

## How was this patch tested?
Existing trash unit tests


mukul1987:
Closing

PR #1841, opened 2021-01-25
HDDS-4486. Feature Config : Make proper enableFSPaths and OFS optimized flag combinations

rakeshadr:

This jira proposal is to simplify the ofs optimization cases. OM can support FS optimization only if both are TRUE(ofsOptimized=true && enableFSPaths=true)

https://issues.apache.org/jira/browse/HDDS-4486

## How was this patch tested?

Through existing test cases


hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
```diff
@@ -545,7 +545,17 @@ private FileStatusAdapter toFileStatusAdapter(OzoneFileStatus status,
   }
 
   @Override
-  public String getBucketLayoutVersion() {
-    return bucket.getMetadata().get(OMConfigKeys.OZONE_OM_LAYOUT_VERSION);
+  public boolean isFSOptimizedBucket() {
+    boolean layoutVersionEnabled =
+            StringUtils.equalsIgnoreCase(
```
mukul1987, line 550: Lets add a comment here and around the original variable that Layout Version V1 in OM implies optimized FS Paths

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/utils/OzoneManagerRatisUtils.java
```diff
@@ -140,41 +141,41 @@ public static OMClientRequest createClientRequest(OMRequest omRequest) {
...
       return new OMKeysRenameRequest(omRequest);
     case CreateDirectory:
-      if (omLayoutVersionV1) {
+      if (isFsOptimizedEnabled()) {
         return new OMDirectoryCreateRequestV1(omRequest);
       }
       return new OMDirectoryCreateRequest(omRequest);
     case CreateFile:
-      if (omLayoutVersionV1) {
+      if (isFsOptimizedEnabled()) {
```
mukul1987, line 178: lets change this to isBucketFSOptimized()

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3680,14 +3682,18 @@ public boolean getEnableFileSystemPaths() {
...
             OZONE_OM_LAYOUT_VERSION_DEFAULT);
-    boolean omLayoutVersionV1 =
-            StringUtils.equalsIgnoreCase(version, "V1");
-    OzoneManagerRatisUtils.setOmLayoutVersionV1(omLayoutVersionV1);
+    boolean omLayoutVersionV1 = StringUtils.equalsIgnoreCase(version,
+            OZONE_OM_LAYOUT_VERSION_V1);
     LOG.info("Configured {}={} and enabled:{} optimized OM FS operations",
             OZONE_OM_LAYOUT_VERSION, version, omLayoutVersionV1);
+
+    boolean enabledFSO = omLayoutVersionV1 && getEnableFileSystemPaths();
```
mukul1987, line 3693: enabledFSO -> isFSOptimized ?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneDirectory.java
```diff
@@ -150,6 +150,7 @@ private void setupOzoneFileSystem()
     OzoneConfiguration conf = new OzoneConfiguration();
     conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
     conf.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION, "V1");
+    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS, true);
```
mukul1987, line 153: We should add a function in OzoneTestUtils to enabled fsPaths.
Something like optimizeFSPaths(conf)

mukul1987 (changes requested):

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneClientAdapterImpl.java
rakeshadr: Thanks a lot @mukul1987 for the review comments. I have updated this in my latest commit.
I'm planning one more follow-up jira task to modify config "`ozone.om.layout.version`" name to something more readable which reflects feature name correctly :-)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/utils/OzoneManagerRatisUtils.java
rakeshadr: Done

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
rakeshadr: Done

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneDirectory.java
rakeshadr: I have add OzoneTestUtils.configureFSOptimizedPaths, which is visible to all `integration-test` cases. Also, I've add the same function into TestOmRequestUtils.configureFSOptimizedPaths, this is basically added to avoid adding new project dependency to `integration-test`. All the `ozone-manager` project test cases using `TestOmRequestUtils` function. 

Do you agree to this changes ?

rakeshadr:
@mukul1987  FYI, test case `TestOzoneFileSystem#testTrash` failure is unrelated to this PR, raised HDDS-4751 issue to address it.

linyiqun (commented):
@rakeshadr , I'm fine for this change, but one concern that does this change will break some logic in below 3 cases? I remember we considered and support these 3 cases when implemented ozone fs feature tasks before.

* ofsOptimized=true && enableFSPaths=false
* ofsOptimized=false && enableFSPaths=true
* ofsOptimized=false && enableFSPaths=false

rakeshadr:
> @rakeshadr , I'm fine for this change, but one concern that does this change will break some logic in below 3 cases? I remember we considered and support these 3 cases when implemented ozone fs feature tasks before.
> 
> * ofsOptimized=true && enableFSPaths=false
> * ofsOptimized=false && enableFSPaths=true
> * ofsOptimized=false && enableFSPaths=false

Thanks a lot @linyiqun for the reviews. Yes, we won't support all these three cases and support only if both the flags are TRUE. As we are exposing this feature freshly in the upcoming Ozone release, we can make it simple now and based on high user demands/usecases we can extend supporting complex modes. Now, this makes it clear to the users that OM supports either `100% FS mode(true) or 100% S3 mode(false)`. Like I mentioned in this patch, OM will expose this as a bucket level property. I will work on few more follow up tasks to ensure completeness (For ex: HDDS-4727).

Hope this makes sense to you.

linyiqun:
Makes sense to me, : ).

rakeshadr:
> Makes sense to me, : ).

Glad to see your reply, please let me know your feedback. Appreciate, if you could approve the PR. Thanks!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfacesV1.java
```diff
@@ -44,9 +43,7 @@ public TestOzoneFileInterfacesV1(boolean setDefaultFs,
   @Override
   protected OzoneConfiguration getOzoneConfiguration() {
     OzoneConfiguration conf = new OzoneConfiguration();
-    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS,
-            enableFileSystemPaths);
-    conf.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION, "V1");
```
linyiqun, line 49: Is here correct to just delete this rather than OzoneTestUtils.configureFSOptimizedPaths(conf, enableFileSystemPaths, OMConfigKeys.OZONE_OM_LAYOUT_VERSION_V1);?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileOps.java
```diff
@@ -73,8 +74,8 @@ public void setupOzoneFileSystem()
           throws IOException, TimeoutException, InterruptedException {
     OzoneConfiguration conf = new OzoneConfiguration();
     conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
-    conf.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION, "V1");
-    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS, false);
+    OzoneTestUtils.configureFSOptimizedPaths(conf,
+            true, OMConfigKeys.OZONE_OM_LAYOUT_VERSION_V1);
```
linyiqun, line 78: Original OZONE_OM_ENABLE_FILESYSTEM_PATHS is false, but now is expected to pass true?

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestOMRequestUtils.java
```diff
@@ -927,4 +962,11 @@ public static long addParentsToDirTable(String volumeName, String bucketName,
     }
     return parentId;
   }
+
+  public static void configureFSOptimizedPaths(Configuration conf,
+      boolean enableFileSystemPaths, String version) {
+    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS,
+            enableFileSystemPaths);
+    conf.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION, version);
+  }
```
linyiqun, line 971: We already defined one completely same method OzoneTestUtils#configureFSOptimizedPaths, we can reuse that and don't need to define this method again.

linyiqun (commented):
@rakeshadr , I did a detailed review, some minor comments from me.  Please have a check.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileOps.java
rakeshadr: Yes, this test class is to verify HDDS-2939 cases. With this PR, we are ensuring that FSO(FileSystemOptimized) feature will be enabled if both flags are TRUE.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestOMRequestUtils.java
rakeshadr: Could you please go through the [comment link](https://github.com/apache/ozone/pull/1841#discussion_r563886377), here I have explained the reason. Thanks!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileInterfacesV1.java
rakeshadr: Good catch. I have completely missed adding new function.

rakeshadr:
Sorry, I've used wrong branch name `4658-01`, instead of `4486-01`, with some other jira issue number. Once all the review comments are fixed, then I will create fresh PR which can be used for merging. Hope that makes sense to everyone.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestOMRequestUtils.java
rakeshadr: On a second thought, I will retain `TestOMRequestUtils` and remove the duplicate function from `OzoneTestUtils`. I will do this change in next update.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneDirectory.java
rakeshadr: I have removed duplicate method from OzoneTestUtils  and retained TestOmRequestUtils#configureFSOptimizedPaths

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestOMRequestUtils.java
```diff
@@ -439,6 +460,20 @@ public static void addBucketToDB(String volumeName, String bucketName,
...
+    metadataList.add(HddsProtos.KeyValue.newBuilder().setKey("key1").setValue(
+            "value1").build());
+    metadataList.add(HddsProtos.KeyValue.newBuilder().setKey("key2").setValue(
+            "value2").build());
+    metadataList.add(HddsProtos.KeyValue.newBuilder().setKey(
+            OMConfigKeys.OZONE_OM_LAYOUT_VERSION).setValue(
+            OMConfigKeys.OZONE_OM_LAYOUT_VERSION_V1).build());
+    metadataList.add(HddsProtos.KeyValue.newBuilder().setKey(
+            OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS).setValue(
+            "false").build());
```
linyiqun, line 474: Not get this, why we pass L1 version but not enabled FILESYSTEM_PATHS here

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/bucket/TestOMBucketCreateRequestV1.java
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Tests OMBucketCreateRequest class, which handles CreateBucket request.
+ */
+public class TestOMBucketCreateRequestV1 extends TestOMBucketCreateRequest {
+
+  @Test
+  public void testValidateAndUpdateCacheWithFSOBucket() throws Exception {
+    when(ozoneManager.getOMLayoutVersion()).thenReturn(
+            OMConfigKeys.OZONE_OM_LAYOUT_VERSION_V1);
```
linyiqun, line 43: Similar comment here. What's the purpose of this test.

linyiqun (commented):
The PR almost looks good to me. Only below nits.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/bucket/TestOMBucketCreateRequestV1.java
rakeshadr: This test is basically to validate the number of items stored in the metadata map `bucketInfoFromProto.getStorageType() and dbBucketInfo.getMetadata()`. Existing test stores two values : key1 and key2. With HDDS-2939 feature, we are making use of bucket's metadata structure to store enable.fspaths and layoutversion details. 

As we saw in below example, the purpose is basically to validate the number of key-value pairs stored in the hashmap and not specific values.

```
result = {HashMap@4191}  size = 4
 "key1" -> "value1"
 "key2" -> "value2"
 "ozone.om.enable.filesystem.paths" -> "false"
 "ozone.om.layout.version" -> "V1"
```


```
TestOMBucketCreateRequestV1: line number 106

Assert.assertEquals(bucketInfoFromProto.getMetadata(),
        dbBucketInfo.getMetadata());
```


In existing code, it adds two items. Please refer [code ](https://github.com/apache/ozone/blob/master/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/bucket/TestOMBucketCreateRequest.java#L180). But in V1 test, it is adding two more elements.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/request/TestOMRequestUtils.java
rakeshadr: Like I explained in below comment, test is to verify the stored elements can be matched against bucketInfo and there is no relation with FSO enabled or not. If required I can make `OZONE_OM_ENABLE_FILESYSTEM_PATHS=true` value, should I change to true?

linyiqun: Okay, get it.

linyiqun (approved):
LGTM , +1.

rakeshadr:
Closing this PR as the branch name is wrong. Created [new PR](https://github.com/apache/ozone/pull/1848) for commiting to the branch.

Thanks a lot @linyiqun and @mukul1987 for the useful review comments.

PR #1844, opened 2021-01-25
HDDS-4731. Fix compatibility issue caused by new enum DatanodeDetails.Port.Name.REPLICATION

adoroszlai:

1. Introduce `version` number to client messages.  Old client does not send version number, in this case it defaults to 0 at the other side.  Client with this change sends `version = 1`.
2. Only include pre-existing ports of datanode (ie. `STANDALONE`, `RATIS`, `REST`) in `DatanodeDetails` protobuf message sent to v0 client, which cannot handle new enum constants.
3. Make client more future-proof by ignoring unknown `Port.Name` values.  This in itself is not enough as a fix, since old client does not have this logic.  But this allows the server to send any new port from now on to `v >= 1` clients.

https://issues.apache.org/jira/browse/HDDS-4731

## How was this patch tested?

Added acceptance test which verifies cross-compatibility of key write/read among 0.5.0, 1.0.0, 1.1.0 versions.

Tested closed container replication using `ozone-topology` compose environment to check for regression.

https://github.com/adoroszlai/hadoop-ozone/actions/runs/510263073

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
@@ -445,6 +445,7 @@ public class DatanodeDetails extends NodeImpl implements
         ipAddress +
         ", host: " +
         hostName +
+        ", ports: " + ports +
         ", networkLocation: " +
         getNetworkLocation() +
         ", certSerialId: " + certSerialId +
@@ -784,6 +785,11 @@ public class DatanodeDetails extends NodeImpl implements
         return name.equals(((Port) anObject).name);
       }
       return false;
+    }
+
+    @Override
+    public String toString() {
+      return name + "=" + value;
     }
 
     public HddsProtos.Port toProto() {
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
@@ -19,9 +19,12 @@
 package org.apache.hadoop.hdds.protocol;
 
 import java.util.ArrayList;
+import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
 import java.util.UUID;
 
+import com.google.common.collect.ImmutableSet;
 import org.apache.hadoop.hdds.annotation.InterfaceAudience;
 import org.apache.hadoop.hdds.annotation.InterfaceStability;
 import org.apache.hadoop.hdds.protocol.DatanodeDetails.Port.Name;
@@ -31,6 +34,8 @@ import org.apache.hadoop.hdds.scm.net.NodeImpl;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * DatanodeDetails class contains details about DataNode like:
@@ -45,6 +50,10 @@ import com.google.common.base.Strings;
 @InterfaceStability.Evolving
 public class DatanodeDetails extends NodeImpl implements
     Comparable<DatanodeDetails> {
+
+  private static final Logger LOG =
+      LoggerFactory.getLogger(DatanodeDetails.class);
+
   /**
    * DataNode's unique identifier in the cluster.
    */
@@ -280,8 +289,7 @@ public class DatanodeDetails extends NodeImpl implements
     }
     for (HddsProtos.Port port : datanodeDetailsProto.getPortsList()) {
       try {
-        Port.Name name = Port.Name.valueOf(port.getName().toUpperCase());
-        builder.addPort(newPort(name, port.getValue()));
+        builder.addPort(Port.fromProto(port));
       } catch (IllegalArgumentException ignored) {
         // ignore unknown port type
       }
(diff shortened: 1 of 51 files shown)


adoroszlai:
@GlenGeng

GlenGeng:
We can help to do a upgrade test to verify whether this issue has been solved in the real environment. Please let me know when I can start the verification. 
cc @Xushaohong who helped to find this issue.

adoroszlai:
> We can help to do a upgrade test to verify whether this issue has been solved in the real environment. Please let me know when I can start the verification.

Thanks @GlenGeng for the offer.  Is it repeatable (ie. after the test can you restore prior state), or do we have only one shot?

GlenGeng:
It is repeatable. Just setup an ozone cluster with your dev branch, and run a testdfsio with old jars upon it.

adoroszlai:
> It is repeatable. Just setup an ozone cluster with your dev branch, and run a testdfsio with old jars upon it.

Thanks, but I meant your test in real environment.  I already added a test to reproduce the bug in CI.

GlenGeng:
Thanks @adoroszlai.
The upgrade test was done in a temporary cluster. During the test, an empty ozone cluster with the new version is setup, a yarn and mapreduce with classpath of jars from old ozone is setup, and we just check the compatibility by running a testDFSIO. Everything is in sandbox.

adoroszlai:
On second thought it seems necessary to detect at server-side whether client is old (supports only specific port types) or new (can handle unknown ports).  Otherwise no new ports for client can be introduced.  The current patch should be fine for the REPLICATION port, since it is for datanodes only.

elek:
> On second thought it seems necessary to detect at server-side whether client is old (supports only specific port types) or new (can handle unknown ports). Otherwise no new ports for client can be introduced. The current patch should be fine for the REPLICATION port, since it is for datanodes only.

Yes, I have the same thoughts. Current patch makes the new port internal, which is fine. But even after this patch we can introduce new public ports only after Ozone 2.0 (but I prefer to not break compatibility even between Ozone 1.x client and Ozone 2.x servers.)

adoroszlai:
@GlenGeng @Xushaohong can you please review, too?

hadoop-hdds/interface-client/src/main/proto/DatanodeClientProtocol.proto
```diff
@@ -196,7 +196,9 @@ message ContainerCommandRequestProto {
   optional   PutSmallFileRequestProto putSmallFile = 20;
   optional   GetSmallFileRequestProto getSmallFile = 21;
   optional   GetCommittedBlockLengthRequestProto getCommittedBlockLength = 22;
+
   optional   string encodedToken = 23;
+  optional uint32 version = 24;
```
GlenGeng, line 201: better keep the style consistent. Up to you.

hadoop-hdds/interface-server/src/main/proto/ScmServerProtocol.proto
```diff
@@ -48,6 +48,7 @@ message SCMBlockLocationRequest {
   optional string traceID = 2;
 
   optional UserInfo userInfo = 3;
+  optional uint32 version = 4;
```
GlenGeng, line 51: Will clientVersion be more straightforward than version in these protos,  in case we might add other version into proto message ?

GlenGeng (commented):
LGTM. Thanks for the fix.

hadoop-hdds/interface-server/src/main/proto/ScmServerProtocol.proto
adoroszlai: They are the same.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
```diff
@@ -333,6 +358,16 @@ public static DatanodeDetails getFromProtoBuf(
...
   public HddsProtos.DatanodeDetailsProto getProtoBufMessage() {
+    return toProto(CURRENT_VERSION);
+  }
+
+  public HddsProtos.DatanodeDetailsProto toProto(int clientVersion) {
+    return toProtoBuilder(clientVersion).build();
+  }
+
+  public HddsProtos.DatanodeDetailsProto.Builder toProtoBuilder(
+      int clientVersion) {
```
xiaoyuyao, line 369: Can we embed the version info inside the PORT Enum without passing clientVersion as parameter?
 
This way, we can make the the handling of the version specific deserialization easier for future changes.

xiaoyuyao (commented):
Thanks @adoroszlai for working on this. The patch looks good to me, especially I like the new forward/backward compatibility testes added. Just one minor suggestion added inline.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
adoroszlai: Thanks @xiaoyuyao for the review.  I'm not sure I understand the proposal.  Can you please clarify?

We can embed the minimum version required for handling each port in the enum constants (eg. `0` for existing ports and `1` for the new replication port).  But don't we still need the client version to be able to decide whether to include it in the proto message or not?

elek (approved):
+1 the concept looks good to me. Let's merge when others have no more concerns.

But I couldn't resist to add some more thoughts about concepts:

Originally we had a pattern where the serialization are added to each of the objects (under different method names). 

After a while we introduced the `Codec` interface to separate the serialization logic from POJOs, but still we call back to the original methods in most of the cases (just because it would be a big refactor to move everything to the codec).

And `Codec` implementations are used mainly for the persistence.

But long-term I would prefer to use Codec for network serialization, too, to separate the serialization logic and manage in the same way.

At that time, we can introduce `VersionedCodec` which can use different `Codec` (depends on the versions).

But this is just a dream, would be great to have more time to improve the code style/class hierarchy....

But even if it's just a dream, I am interested about your opinions ;-)

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
xiaoyuyao: Yes. We still need the client version. But that can be a static helper without many interface changes to introduce extra parameter.

arp7:
I merged this based on +1s from @xiaoyuyao and @elek.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
adoroszlai: > We still need the client version. But that can be a static helper

This conversion must be done on the server-side, and server can have clients of various versions.  How would a static helper work?

adoroszlai:
Thanks @GlenGeng, @elek and @xiaoyuyao for the review, @arp7 for merging it.

arp7:
It looks like I missed a minor remaining comment from @xiaoyuyao. Let's address in a follow up jira if needed.

PR #389, opened 2019-12-24
HDDS-2534. scmcli container delete not working

chimney-lee:
## What changes were proposed in this pull request?

What changes were proposed in this pull request?
fix scmcli container delete not working

What is the link to the Apache JIRA
https://issues.apache.org/jira/browse/HDDS-2534
#214 

How was this patch tested?
exec bin/ozone scmcli container delete  <containerId>

PR #214, opened 2019-11-18
HDDS-2534. scmcli container delete not working

chimney-lee:
## What changes were proposed in this pull request?

fix scmcli container delete not working

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-2534

## How was this patch tested?

exec `bin/ozone scmcli container delete <containerId>`

PR #1845, opened 2021-01-26
HDDS-4746. Fix delete container occurs unknown command type.

captainzmc:

An exception "Unknown command type: DeleteContainer" occurs when we delete a container. This is a bug that needs to be fixed.
In fact, this part of the interface backend is already implemented, just not called.
image

https://issues.apache.org/jira/browse/HDDS-4746

## How was this patch tested?

UT added

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
@@ -298,6 +298,13 @@ public final class StorageContainerLocationProtocolServerSideTranslatorPB
             .setStartMaintenanceNodesResponse(startMaintenanceNodes(
                 request.getStartMaintenanceNodesRequest()))
           .build();
+      case DeleteContainer:
+        return ScmContainerLocationResponse.newBuilder()
+            .setCmdType(request.getCmdType())
+            .setStatus(Status.OK)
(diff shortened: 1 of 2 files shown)


avijayanhwx:
Thanks for working on this @captainzmc. Can we find out in why and in which JIRA this command was removed?

adoroszlai:
Please see #214 and #389 for previous attempts to fix this.

captainzmc:
> Thanks for working on this @captainzmc. Can we find out why and in which JIRA this command was removed?

hi @avijayanhwx This interface is not removed, this interface didn't work properly from the beginning.

captainzmc:
> Please see #214 and #389 for previous attempts to fix this.

@adoroszlai Thanks for the reminding. I will confirm the current status of #389.

captainzmc:
Hi @maobaolong,  Are you currently working on fixing #389? If not, I hope to take this Task, because we will use this interface.

maobaolong:
@captainzmc Sure, please feel free to take it, hope you can finish this.

captainzmc:
By looking at the issues mentioned in #214 and #389, I found that in #1338(this PR is used to remove empty containers) these issues were well addressed and explained, by fixed ContainerReportHandler.
I have tested deleting the container locally, metadata and data can be deleted normally, also DataNode and SCM logs are normal. There are no report exceptions.
So I think we can continue with the current PR review. cc @maobaolong @adoroszlai

captainzmc:
We use this feature because HDDS-4334 relies on it. Now #1338 has implemented the remove all empty containers. So we can delete all the files through OFS, and the empty container will be deleted automatically. So we don't think it's necessary anymore, so I'm going to turn off the PR.
If there are other features needed, we can start this PR again.

PR #1847, opened 2021-01-27
HDDS-4720. RenameKey : add unit test to verify bucket#renameKey

rakeshadr:

This task is to verify the functionality of ozonebucket#renameKey client API. Basically the existing o3fs#rename(path, path) logic is internally invoking bucket#renameKey, so functionality should work fine.

https://issues.apache.org/jira/browse/HDDS-4720

## How was this patch tested?

Added new UTs

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
@@ -31,6 +31,7 @@ import org.apache.hadoop.ozone.TestDataUtil;
 import org.apache.hadoop.ozone.client.ObjectStore;
 import org.apache.hadoop.ozone.client.OzoneBucket;
 import org.apache.hadoop.ozone.client.OzoneClient;
+import org.apache.hadoop.ozone.client.OzoneKey;
 import org.apache.hadoop.ozone.client.OzoneKeyDetails;
 import org.apache.hadoop.ozone.client.OzoneVolume;
 import org.apache.hadoop.ozone.client.io.KeyOutputStream;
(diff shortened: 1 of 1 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
```diff
@@ -264,6 +268,95 @@ public void testLookupKey() throws Exception {
...
+    OMException oe = null;
+    String toKeyName = "";
+    try {
+      bucket.renameKey(fromKeyName, toKeyName);
+      fail("Rename to empty string should fail!");
+    } catch (OMException e) {
+      oe = e;
+    }
+    Assert.assertEquals(OMException.ResultCodes.INVALID_KEY_NAME,
+            oe.getResult());
```
linyiqun, line 292: Can we simplified all this check to below? I didn't see _oe_ is used in other place.
```java
    try {
      bucket.renameKey(fromKeyName, toKeyName);
      fail("Rename to empty string should fail!");
    } catch (OMException e) {
      Assert.assertEquals(OMException.ResultCodes.INVALID_KEY_NAME,
            e.getResult());
    }
```

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
```diff
@@ -264,6 +268,95 @@ public void testLookupKey() throws Exception {
...
+    try {
+      bucket.getKey(fromKeyName);
+      fail("Lookup for old from key name should fail!");
+    } catch (OMException e) {
+      oe = e;
+    }
+    Assert.assertEquals(KEY_NOT_FOUND, oe.getResult());
+
+    OzoneKey key = bucket.getKey(toKeyName);
+    Assert.assertEquals(toKeyName, key.getName());
```
linyiqun, line 307: Can we additionally add a case for renamed to an existed key? There should be returned KEY_ALREADY_EXISTS error.

rakeshadr: Sure, I will add this case. 

As we know `OMKeyRenameRequestV1` is common for both FS#rename and bucket#renameKey I hope TestOzoneFileSystemV1 cases are covering most of the other cases.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
rakeshadr: Agreed!

linyiqun (approved):
LGTM, +1. @rakeshadr , please double check if failed ut is related with this, thanks.

rakeshadr:
> LGTM, +1. @rakeshadr , please double check if failed ut is related with this, thanks.

Thanks @linyiqun for the reviews. TestOzoneFileSystem is failing, which is unrelated to this PR. Failure will be analysed separately along with HDDS-4486 or other jira tasks.

PR #1851, opened 2021-01-29
HDDS-4751. TestOzoneFileSystem#testTrash failed when enabledFileSystemPaths and omRatisDisabled

sadanand48:
TestOzoneFileSystem testTrash fails while setting `enableFsPaths` to true. This is because of using `DeleteKeyRequest `for delete. 
There are 2 operations that happen when a key is moved to Trash 1. rename 2 .delete .
 While Renaming a directory key,  the renamedKey is stored  in KeyTable with  "/"  (`RenameKeyRequest`)  irrespective of `enableFsPaths` value but while deleting a key via `DeleteKeyRequest` the "/" is removed and a lookup is done on KeyTable which results in KEY_NOT_FOUND.  Using `DeleteKeys` instead of `DeleteKey` will solve this issue as a lookup is not done on the normalised key.
[This PR](https://github.com/apache/ozone/pull/1256) explains why normalisation is not done while rename
https://issues.apache.org/jira/browse/HDDS-4751
## How was this patch tested?
Existing Test Trash unit test.


mukul1987 (approved):
+1, the change looks good to me.

PR #1852, opened 2021-01-29
HDDS-4761. Implement increment count optimization in DeletedBlockLog V2

amaliujia:

Implement increment count optimization in DeletedBlockLog V2. We only write retry count to DB in every 100 times.

This optimization is already in master branch: https://github.com/apache/ozone/commit/39027e4b33eaa2c233c10ced2863fed7d937e944

https://issues.apache.org/jira/browse/HDDS-4761

## How was this patch tested?

UT

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
@@ -279,9 +279,10 @@ public class DeletedBlockLogImplV2
 
   /**
    * Called in SCMStateMachine#notifyLeaderChanged when current SCM becomes
-   *  leader.
+   * leader.
    */
-  public void clearTransactionToDNsCommitMap() {
+  public void clear() {
+    deletedBlockLogStateManager.clear();
     transactionToDNsCommitMap.clear();
   }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManager.java
@@ -42,4 +42,6 @@ public interface DeletedBlockLogStateManager {
       KeyValue<Long, DeletedBlocksTransaction>> getReadOnlyIterator();
 
   void onFlush();
+
+  void clear();
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
@@ -216,6 +216,9 @@ public class DeletedBlockLogStateManagerImpl
     }
   }
 
+  public void clear() {
+    transactionRetryCountMap.clear();
+  }
 
   public void onFlush() {
     deletingTxIDs.clear();
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
@@ -33,6 +33,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.lang.reflect.Proxy;
 import java.util.ArrayList;
+import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -50,6 +51,8 @@ public class DeletedBlockLogStateManagerImpl
   private final DBTransactionBuffer transactionBuffer;
   private final int maxRetry;
   private Set<Long> deletingTxIDs;
+  // Maps txId to its retry counts;
+  private Map<Long, Integer> transactionRetryCountMap;
 
   public DeletedBlockLogStateManagerImpl(
       ConfigurationSource conf,
@@ -60,6 +63,7 @@ public class DeletedBlockLogStateManagerImpl
     this.deletedTable = deletedTable;
     this.transactionBuffer = txBuffer;
     this.deletingTxIDs = ConcurrentHashMap.newKeySet();
+    this.transactionRetryCountMap = new ConcurrentHashMap<>();
   }
 
   public TableIterator<Long, TypedTable.KeyValue<Long,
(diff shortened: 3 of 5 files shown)


amaliujia:
R: @runzhiwang @GlenGeng

amaliujia:
Without this optimization, next time when master is synced to HDDS-2823, I think current V2 implementation won't pass the unit test in `TestDeletedBlockLog`

GlenGeng:
@amaliujia  Thanks for back port this fix to HDDS-2823.

I consider we have to move `transactionRetryCountMap` from `DeletedBlockLogStateManagerImpl#increaseRetryCountOfTransactionInDB` to `DeletedBlockLogImplV2#incrementCount`, and clear it when SCM becomes leader.

The reason is, content of `DeletedBlockLogImplV2` should be kept consistent and deterministic across SCM instance, which is not necessary and can not be maintained during SCM failover.

amaliujia:
@GlenGeng  I have updated this PR based on discussion. Also update the UT as the UT was updated during the master to 2823 sync thus before this PR, the UT was not functionally correct.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
```diff
@@ -279,9 +279,10 @@ public int getNumOfValidTransactions() throws IOException {
 
   /**
    * Called in SCMStateMachine#notifyLeaderChanged when current SCM becomes
-   *  leader.
+   * leader.
    */
-  public void clearTransactionToDNsCommitMap() {
+  public void clear() {
```
GlenGeng, line 284: Please rename to `onBecomeLeader()`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManager.java
```diff
@@ -42,4 +42,6 @@ void increaseRetryCountOfTransactionInDB(ArrayList<Long> txIDs)
       KeyValue<Long, DeletedBlocksTransaction>> getReadOnlyIterator();
 
   void onFlush();
+
+  void clear();
```
GlenGeng, line 46: ditto. `onBecomeLeader()`

GlenGeng (commented):
LGTM

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
```diff
@@ -186,22 +191,34 @@ public void increaseRetryCountOfTransactionInDB(
...
-      if (currentCount > maxRetry) {
-        builder.setCount(-1);
+        int nextCount = currentCount + 1;
+        DeletedBlocksTransaction.Builder builder = block.toBuilder();
+        if (nextCount > maxRetry) {
+          // if the retry time exceeds the maxRetry value
+          // then set the retry value to -1, stop retrying, admins can
+          // analyze those blocks and purge them manually by SCMCli.
+          builder.setCount(-1);
+          deletedTable.putWithBatch(
```
GlenGeng, line 204: better ` transactionRetryCountMap.remove(txID);` ? Otherwise might be a leak.

GlenGeng:
Hey Rui, after a second thought, I realized that my suggestion (the second commit) is wrong, sorry for that.
Your first commit is deterministic for `increaseRetryCountOfTransactionInDB `.

Given accross different SCMs, 
- the rocksDB is the same at the same raft log index, 
- and they has the same raft log, which means they will apply the same `increaseRetryCountOfTransactionInDB

The generated `transactionRetryCountMap` is deterministic, under all kinds of failovers, including restart, become leader and step down.

The real problem here is, this solution does not decrease the number of fsync. In non-HA mode, each write to DB will call fsync, and in HA mode, the writes will be buffered, but each write will trigger a raft client request, which will call fsync on raft log.

Let's go througth this problem again.

GlenGeng:
After a third thought, I consider both the first commit and first + second commit can not ensure the deterministic of the retry counter map, and the optimization in HA mode should be reduce the raft call. What do you think ?

amaliujia:
@GlenGeng  

Optimize to reduce the Ratis call make sure. So in that case, we will need both in-memory state in DeletedBlockLog and state manager. The former is to reduce ratis call, and the latter is required because of the transaction buffer.

What do you think?

GlenGeng:
Agree. The key point of a deterministic solution is to let each ratis call write rocksDB.

amaliujia:
@GlenGeng  unfortunately with transaction buffer that we cannot flush retry count into DB immediately thus we probably have to accept the non-deterministic result for this moment.

First of all,  the current implementation is already bad because it reads from DB and then try to write into DB, thus in the worst case, DB value never increase thus for a txID it retries forever. So we have to introduce the in-memory state in state manager.

Then, 
I think the non-deterministic here might not be a big deal. Even though DB state could be different, but
1. The count in DB will not be overwrite back because we have in-memory state now. (or it becomes consistent through installSnapshot) 
2. there is a maxRetry as a bound.

At least current PR will avoid the worst case. What do you think?

GlenGeng:
> unfortunately with transaction buffer that we cannot flush retry count into DB immediately thus we probably have to accept the non-deterministic result for this moment.

Not flush retry count into DB immediately may not lead to non-deterministic. Let's try to prove it:
1, if current SCM(both leader and follower) has not touched the txID, if will load the count from DB.
2, if current SCM(both leader and follower) touched the txID, then if will use the count in `transactionRetryCountMap`, which will be accurate if SCM is not restarted.
3, For a restarted SCM, loading DB and replaying corresponding log entry will help build a `transactionRetryCountMap` that is   exactly the same as other SCMs.

What we need take care is to move the counter check logic into ReadOnlyIterator, e.g, 

```
            if (txn.getCount() > -1 && txn.getCount() <= maxRetry
                && !containerManager.getContainer(id).isOpen()) {
              numBlocksAdded += txn.getLocalIDCount();
              getTransaction(txn, transactions);
              transactionToDNsCommitMap
                  .putIfAbsent(txn.getTxID(), new LinkedHashSet<>());
            }
```

```
      private void findNext() {
        while (iter.hasNext()) {
          TypedTable.KeyValue<Long, DeletedBlocksTransaction> next = iter
              .next();
          long txID;
          try {
            txID = next.getKey();
          } catch (IOException e) {
            throw new IllegalStateException("");
          }

         retryCount = transactionRetryCountMap.getOrDefault(txID, 0) 
          if (!deletingTxIDs.contains(txID) && retryCount  > -1 && retryCount <= maxRetry) {
            nextTx = next;
            if (LOG.isTraceEnabled()) {
              LOG.trace("DeletedBlocksTransaction matching txID:{}",
                  txID);
            }
            return;
          }
        }
        nextTx = null;
      }
```

Given the `transactionRetryCountMap.getOrDefault`, we don't need add new in-memory staff, What do you think ?

GlenGeng:
Hey Rui, the trxBuffer introduce some issues into the delete block log, let me draft a design for it, we can discuss there.
Please hold on this PR for now, thanks!

amaliujia:
I will open another PR to apply the optimization on retry count that we have discussed.

PR #1864, opened 2021-01-30
HDDS-4768. Refactor the repeated exception conversion method into a common util class

lamber-ken:

The following method was defined in `RocksDBStore.java`, `RDBStore.java`, `RDBTable.java`, this pr aims to move it to `HddsServerUtil` which is a util class for server side components.
```java
/**
 * Converts RocksDB exception to IOE.
 * @param msg  - Message to add to exception.
 * @param e - Original Exception.
 * @return  IOE.
 */
public static IOException toIOException(String msg, RocksDBException e) {
  String statusCode = e.getStatus() == null ? "N/A" :
      e.getStatus().getCodeString();
  String errMessage = e.getMessage() == null ? "Unknown error" :
      e.getMessage();
  String output = msg + "; status : " + statusCode
      + "; message : " + errMessage;
  return new IOException(output, e);
}
```

https://issues.apache.org/jira/browse/HDDS-4768

## How was this patch tested?

No need more new tests.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HddsServerUtil.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HddsServerUtil.java
@@ -70,6 +70,8 @@ import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_HEARTBEAT_RPC_R
 import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_STALENODE_INTERVAL;
 import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_STALENODE_INTERVAL_DEFAULT;
 import static org.apache.hadoop.hdds.server.ServerUtils.sanitizeUserArgs;
+
+import org.rocksdb.RocksDBException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


ayushtkn (commented):
Changes Looks good and safe.
Can you give a check to the build failures, shouldn't be related though...

lamber-ken:
Thanks @ayushtkn. These failures are caused by missing required dependencies.

image

lamber-ken:
hi @ayushtkn, the ci go pass.

adoroszlai (commented):
Thanks @lamber-ken for getting rid of this code duplication.

I'd prefer adding a static import for the new utility method instead of importing the class itself:

```
import static org.apache.hadoop.hdds.utils.HddsServerUtil.toIOException;
```

This would keep the code more readable and minimize change.

The patch is fine in its current form, too.

lamber-ken:
> Thanks @lamber-ken for getting rid of this code duplication.
> 
> I'd prefer adding a static import for the new utility method instead of importing the class itself:
> 
> ```
> import static org.apache.hadoop.hdds.utils.HddsServerUtil.toIOException;
> ```
> 
> This would keep the code more readable and minimize change.
> 
> The patch is fine in its current form, too.

Thanks for your review. Let's  leave it just as it is now. 😄

lamber-ken:
hi @adoroszlai @ayushtkn, it seems that we can only use IOException directly, like following

image

ayushtkn:
@lamber-ken that would not append the status code, to do so only, we have this util method, else everywhere checking null and then....
So, we should use the util itself and keep the exception message pattern same.

Can you change to using `static` imports as @adoroszlai suggested? Post that we can push this.

lamber-ken:
Got，follow you @ayushtkn.

ayushtkn (approved):
LGTM

adoroszlai (approved):
Thanks @lamber-ken for updating the patch.

adoroszlai:
Thanks @ayushtkn for the review.

PR #1866, opened 2021-01-31

smengcl:
3rd time's the charm

Use ClientProtocol directly in Adapter and FS. Completely remove the usage of OzoneClientAdapter interface in OFS.

## How was this patch tested?

All existing UTs/robots should pass as-is.

## What is the difference of this PR since #1363?

1. Rebased to the latest master branch
2. Renamed internal class `BasicRootedOzoneClientAdapterImpl` to `BasicRootedOzoneFileSystemHelper` as [suggested](https://github.com/apache/ozone/pull/1363#issuecomment-694144300).

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/ObjectStore.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/ObjectStore.java
@@ -81,7 +81,9 @@ public class ObjectStore {
     proxy = null;
   }
 
-  @VisibleForTesting
+  /**
+   * Returns the ClientProtocol of the ObjectStore.
+   */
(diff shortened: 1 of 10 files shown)


elek (commented):
Sorry for the late review @smengcl, and thanks for the patch. The concept looks good to me. 

In fact, I am very interested about the mentioned future:

```
 * Previously BasicRootedOzoneClientAdapterImpl. Might be merged with
 * BasicRootedOzoneFileSystem in the future.
 ```
 
 But it seems to be a reasonable step forward to it.
 
 As far as I see it has never been passed the CI, would be great to get a green build...

elek:
/pending As far as I see it has never been passed the CI, would be great to get a green build...

PR #1870, opened 2021-02-01
HDDS-4773. Add functionality to transfer Rocks db checkpoint from leader to follower

bshashikant:
Aims to add the transfer rocks db checkpointing functionality for SCM HA.

https://issues.apache.org/jira/browse/HDDS-4773

## How was this patch tested?
Added Unit tests

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
@@ -42,9 +42,9 @@ import java.util.concurrent.TimeUnit;
  * Grpc client to download a Rocks db checkpoint from leader node
  * in SCM HA ring.
  */
-public class SCMGrpcClient implements SCMSnapshotDownloader{
+public class InterSCMGrpcClient implements SCMSnapshotDownloader{
   private static final Logger LOG =
-      LoggerFactory.getLogger(SCMGrpcClient.class);
+      LoggerFactory.getLogger(InterSCMGrpcClient.class);
 
   private final ManagedChannel channel;
 
@@ -53,7 +53,7 @@ public class SCMGrpcClient implements SCMSnapshotDownloader{
 
   private final long timeout;
 
-  public SCMGrpcClient(final String host, final ConfigurationSource conf) {
+  public InterSCMGrpcClient(final String host, final ConfigurationSource conf) {
     Preconditions.checkNotNull(conf);
     int port = conf.getObject(SCMHAConfiguration.class).getGrpcBindPort();
     timeout =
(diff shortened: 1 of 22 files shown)


amaliujia:
cc @amaliujia

GlenGeng (commented):
LGTM. Thanks for the job!
The code style is neat, similar to that in Ratis.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMGrpcProtocolService.java
```diff
@@ -0,0 +1,82 @@
...
+    nettyServerBuilder.addService(service);
+    server = nettyServerBuilder.build();
+  }
+
+  public int getPort() {
+    return this.port;
+  }
+
+  public void start() throws IOException {
+    if (!isStarted) {
```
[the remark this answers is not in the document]
bshashikant: I think we probably don't need to make it thread safe.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMSnapshotDownloader.java
```diff
@@ -0,0 +1,35 @@
...
+ * Contract to download a SCM Snapshot from remote server..
+ * <p>
+ *
+ * The underlying implementation is supposed to download SCM snapshot via
+ * any chosen protoclol(for now its Grpc).
+ * images.
+ */
+public interface SCMSnapshotDownloader {
+
+  CompletableFuture<Path> download(final Path destination) throws IOException;
```
amaliujia, line 32: Question: is `destination` a file or a directory? 

Nit: maybe add a javadoc here to explain.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
```diff
@@ -0,0 +1,172 @@
...
+        withDeadlineAfter(timeout, TimeUnit.SECONDS);
+  }
+
+
+  @Override
+  public CompletableFuture<Path> download(final Path outputPath) {
+    // By default on every checkpoint, the rocks db will be flushed
+    InterSCMProtocolProtos.CopyDBCheckpointRequestProto request =
+        InterSCMProtocolProtos.CopyDBCheckpointRequestProto.newBuilder()
+            .setFlush(true)
```
amaliujia, line 75: Why we need a flush flag than always flush the DB before take a checkpoint? Or when we want to set this flush flag as false?

bshashikant: By default, flush should always be true. A flag like this will help in benchmarking later.

amaliujia:
Overall LGTM

amaliujia (approved):

bshashikant:
Thanks @amaliujia for the review.

PR #1871, opened 2021-02-01
HDDS-4763. Owner field of S3AUTHINFO type delegation token should be validated

elek:

Adding a check for S3AUTH token validation. 

Please see the jira for the motivation. 

https://issues.apache.org/jira/browse/HDDS-4763

## How was this patch tested?

Test method is explained in the Jira (with the code to test it).

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
@@ -472,7 +472,11 @@ public class OzoneDelegationTokenSecretManager
   private byte[] validateS3AuthInfo(OzoneTokenIdentifier identifier)
       throws InvalidToken {
     LOG.trace("Validating S3AuthInfo for identifier:{}", identifier);
-    if (!identifier.getOwner().equals(identifier.getAwsAccessId())) {
+    if (identifier.getOwner() == null) {
+      throw new InvalidToken(
+          "Owner is missing from the S3 auth token");
+    }
+    if (!identifier.getOwner().toString().equals(identifier.getAwsAccessId())) {
       LOG.error(
           "Owner and AWSAccessId is different in the S3 token. Possible "
               + " security attack: {}",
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
```diff
@@ -472,6 +472,19 @@ public boolean verifySignature(OzoneTokenIdentifier identifier,
   private byte[] validateS3AuthInfo(OzoneTokenIdentifier identifier)
       throws InvalidToken {
     LOG.trace("Validating S3AuthInfo for identifier:{}", identifier);
+    if (identifier.getOwner() == null) {
+      throw new InvalidToken(
+          "Owner is missing from the S3 auth token");
+    }
+    if (!identifier.getOwner().toString().equals(identifier.getAwsAccessId())) {
```
xiaoyuyao, line 479: Thanks @elek  for working on this. I have one question about the awsaccessId from identifier or owner, they can all be modified by the attacker as well. What we need to check is compare awsaccessId from the identifier.stringtosign using s3 signatureparser (after verify the secret) with the one from the owner/awsid from the identifier.

xiaoyuyao (commented):
One question added inline. Otherwise, LGTM.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/security/OzoneDelegationTokenSecretManager.java
elek: _awsaccessId_ is used as a primary key to find the stored secret in the rockdb. If you use any wrong accessId the signature validation will be failed. If the string2sign contains reference to a different accessId, the validation will be failed as the signature won't be matched. 

(At least this is my understanding, but let me know if I missed something...)

xiaoyuyao: Make sense to me. I further check the hadoop rpc code and the UGI of token user is taken from the owner field of the identifier. Because S3 token identifier is not protected with a token signature like Hadoop delegation token, the proposed check of owner against awsaccessid is the right fix.

xiaoyuyao (approved):

PR #1874, opened 2021-02-02
HDDS-4482. SCM should be able to persist CRL

vivekratnavel:

- Add a new table to persist CRL information
- Add a new table to keep track of the last CRL sequence ID
- Update the revoke certificates method to create a new CRL to revoke certificates
- Add Unit tests

https://issues.apache.org/jira/browse/HDDS-4482

## How was this patch tested?

Unit tests

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
@@ -31,6 +31,7 @@ import java.io.IOException;
 import java.math.BigInteger;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
+import java.util.Date;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.Future;
@@ -107,11 +108,13 @@ public interface CertificateServer {
    * @param serialIDs       - List of serial IDs of Certificates to be revoked.
    * @param reason          - Reason for revocation.
    * @param securityConfig  - Security Configuration.
+   * @param revocationTime  - Revocation time for the certificates.
    * @return Future that gives a list of certificates that were revoked.
    */
   Future<Optional<Long>> revokeCertificates(
       List<BigInteger> serialIDs,
       CRLReason reason,
+      Date revocationTime,
       SecurityConfig securityConfig);
 
   /**
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
@@ -23,13 +23,16 @@ import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.security.exception.SCMSecurityException;
 import org.apache.hadoop.hdds.security.x509.SecurityConfig;
 import org.apache.hadoop.hdds.security.x509.certificate.authority.CertificateApprover.ApprovalType;
+import org.bouncycastle.asn1.x509.CRLReason;
 import org.bouncycastle.cert.X509CertificateHolder;
 import org.bouncycastle.pkcs.PKCS10CertificationRequest;
 
 import java.io.IOException;
+import java.math.BigInteger;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.Future;
 
 /**
@@ -46,7 +49,7 @@ public interface CertificateServer {
    * @throws SCMSecurityException - Throws if the init fails.
    */
   void init(SecurityConfig securityConfig, CAType type)
-      throws SCMSecurityException;
+      throws IOException;
 
   /**
    * Returns the CA Certificate for this CA.
@@ -101,14 +104,15 @@ public interface CertificateServer {
   /**
    * Revokes a Certificate issued by this CertificateServer.
    *
-   * @param certificates - List of Certificates to revoke.
-   * @param reason - Reason for revocation.
-   * @param securityConfig - Security Configuration.
-   * @return Future that tells us what happened.
+   * @param serialIDs       - List of serial IDs of Certificates to be revoked.
+   * @param reason          - Reason for revocation.
+   * @param securityConfig  - Security Configuration.
+   * @return Future that gives a list of certificates that were revoked.
    */
-  Future<Boolean> revokeCertificates(List<X509Certificate> certificates,
-                                     int reason,
-                                     SecurityConfig securityConfig);
+  Future<Optional<Long>> revokeCertificates(
+      List<BigInteger> serialIDs,
+      CRLReason reason,
+      SecurityConfig securityConfig);
 
   /**
    * List certificates.
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
@@ -101,26 +101,22 @@ public interface CertificateServer {
   /**
    * Revokes a Certificate issued by this CertificateServer.
    *
-   * @param certificate - Certificate to revoke
-   * @param approver - Approval process to follow.
+   * @param certificates - List of Certificates to revoke.
+   * @param reason - Reason for revocation.
+   * @param securityConfig - Security Configuration.
    * @return Future that tells us what happened.
-   * @throws SCMSecurityException - on Error.
    */
-  Future<Boolean> revokeCertificate(X509Certificate certificate,
-      ApprovalType approver) throws SCMSecurityException;
-
-  /**
-   * TODO : CRL, OCSP etc. Later. This is the start of a CertificateServer
-   * framework.
-   */
+  Future<Boolean> revokeCertificates(List<X509Certificate> certificates,
+                                     int reason,
+                                     SecurityConfig securityConfig);
 
   /**
    * List certificates.
    * @param type            - node type: OM/SCM/DN
    * @param startSerialId   - start certificate serial id
    * @param count           - max number of certificates returned in a batch
-   * @return
-   * @throws IOException
+   * @return List of X509 Certificates.
+   * @throws IOException - On Failure
    */
   List<X509Certificate> listCertificate(HddsProtos.NodeType type,
       long startSerialId, int count, boolean isRevoked) throws IOException;
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateStore.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateStore.java
@@ -20,9 +20,12 @@
 package org.apache.hadoop.hdds.security.x509.certificate.authority;
 
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
+import org.apache.hadoop.hdds.security.x509.SecurityConfig;
+import org.bouncycastle.cert.X509CertificateHolder;
 
 import java.io.IOException;
 import java.math.BigInteger;
+import java.security.KeyPair;
 import java.security.cert.X509Certificate;
 import java.util.List;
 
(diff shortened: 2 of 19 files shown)


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
```diff
@@ -101,26 +101,22 @@ X509Certificate getCertificate(String certSerialId)
...
-   */
-  Future<Boolean> revokeCertificate(X509Certificate certificate,
-      ApprovalType approver) throws SCMSecurityException;
-
-  /**
-   * TODO : CRL, OCSP etc. Later. This is the start of a CertificateServer
-   * framework.
    */
+  Future<Boolean> revokeCertificates(List<X509Certificate> certificates,
+                                     int reason,
```
xiaoyuyao, line 110: Can the reason be a Enum?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
```diff
@@ -101,26 +101,22 @@ X509Certificate getCertificate(String certSerialId)
...
+   * @param securityConfig - Security Configuration.
    * @return Future that tells us what happened.
-   * @throws SCMSecurityException - on Error.
-   */
-  Future<Boolean> revokeCertificate(X509Certificate certificate,
-      ApprovalType approver) throws SCMSecurityException;
-
-  /**
-   * TODO : CRL, OCSP etc. Later. This is the start of a CertificateServer
-   * framework.
```
xiaoyuyao, line 114: NIT: javadoc?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultCAServer.java
```diff
@@ -272,20 +272,22 @@ private X509CertificateHolder signAndStoreCertificate(LocalDate beginDate,
...
   @Override
-  public Future<Boolean> revokeCertificate(X509Certificate certificate,
-      CertificateApprover.ApprovalType approverType)
-      throws SCMSecurityException {
+  public Future<Boolean> revokeCertificates(List<X509Certificate> certificates,
+                                            int reason,
+                                            SecurityConfig securityConfig) {
     CompletableFuture<Boolean> revoked = new CompletableFuture<>();
-    if (certificate == null) {
+    if (certificates == null || certificates.isEmpty()) {
```
xiaoyuyao, line 279: NIT: we can use Collection#isEmptyOrNull here.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateStore.java
```diff
@@ -47,12 +50,21 @@ void storeValidCertificate(BigInteger serialID,
...
+   * @param reason - CRLReason for revocation.
+   * @param securityConfig - Security Configuration.
+   * @param keyPair - Public and Private key of the CA.
    * @throws IOException
    */
-  void revokeCertificate(BigInteger serialID) throws IOException;
+  void revokeCertificates(List<X509Certificate> certificates,
+                          X509CertificateHolder caCertificateHolder,
+                          int reason, SecurityConfig securityConfig,
+                          KeyPair keyPair)
```
xiaoyuyao, line 66: can we abstract the CRL handling into a separate class (like Approver for CSR) without bringing CA keypair into the certificate store interface and class?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/crl/CRLInfo.java
```diff
@@ -0,0 +1,160 @@
...
+   * first argument is less than, equal to, or greater than the
+   * second.
+   * @throws NullPointerException if an argument is null and this
+   *                              comparator does not permit null arguments
+   * @throws ClassCastException   if the arguments' types prevent them from
+   *                              being compared by this comparator.
+   */
+  @Override
+  public int compare(CRLInfo o1, CRLInfo o2) {
+    return 0;
```
xiaoyuyao, line 114: This does not seem to be implemented yet?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/crl/CRLInfo.java
```diff
@@ -0,0 +1,160 @@
...
+
+    return this.getX509CRL().equals(that.x509CRL) &&
+        this.creationTimestamp == that.creationTimestamp;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(x509CRL);
+  }
+
```
xiaoyuyao, line 137: Can we add a toString override for debugging output, etc?

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/authority/TestDefaultCAServer.java
```diff
@@ -213,6 +224,55 @@ public void testRequestCertificateWithInvalidSubject() throws IOException,
...
+    X509Certificate certificate =
+        new JcaX509CertificateConverter().getCertificate(holder.get());
+    List<X509Certificate> certs = new ArrayList<>();
+    certs.add(certificate);
+    Future<Boolean> revoked = testCA.revokeCertificates(certs,
+        CRLReason.keyCompromise, new SecurityConfig(conf));
+
+    // Revoking a valid certificate should return true.
+    assertTrue(revoked.get());
+
```
xiaoyuyao, line 263: can we verify that after revoke, getCertificateByID from ca with valid  and invalid type, respectively?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
```diff
@@ -73,13 +75,30 @@
...
+  public static final DBColumnFamilyDefinition<Long, CRLInfo> CRL_INFO =
+      new DBColumnFamilyDefinition<>(
+          "crlInfo",
+          Long.class,
+          new LongCodec(),
+          CRLInfo.class,
+          new CRLInfoCodec());
+
+  public static final DBColumnFamilyDefinition<String, Long>
+      CRL_SEQUENCE_ID =
```
xiaoyuyao, line 94: It's a bit overkill to use a table to just save the crl sequence id. Any chance to put it into the SCM version file?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
```diff
@@ -73,13 +75,30 @@
...
       CONTAINERS =
-      new DBColumnFamilyDefinition<ContainerID, ContainerInfo>(
+      new DBColumnFamilyDefinition<>(
           "containers",
           ContainerID.class,
           new ContainerIDCodec(),
           ContainerInfo.class,
           new ContainerInfoCodec());
 
+  public static final DBColumnFamilyDefinition<Long, CRLInfo> CRL_INFO =
```
xiaoyuyao, line 85: NIT: CRL_INFO=>CRLS to be consistent with other table definition.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/server/TestSCMCertStore.java
```diff
@@ -0,0 +1,235 @@
...
+
+    Set<? extends X509CRLEntry> revokedCertificates =
+        crlInfo.getX509CRL().getRevokedCertificates();
+    Assert.assertEquals(1L, revokedCertificates.size());
+    Assert.assertEquals(x509Certificate.getSerialNumber(),
+        revokedCertificates.iterator().next().getSerialNumber());
+
+    // Now trying to revoke the already revoked certificate should result in
+    // a warning message and no-op. It should not create a new CRL.
+    scmCertStore.revokeCertificates(certs,
```
xiaoyuyao, line 148: How does the caller know that has already been revoked? a warning message won't work at API level.

xiaoyuyao (commented):
Thanks @vivekratnavel  for working on this. The patch LGTM overall. Just few minor comments inline.

hadoop-hdds/framework/src/test/java/org/apache/hadoop/hdds/security/x509/certificate/authority/TestDefaultCAServer.java
vivekratnavel: I have added unit tests to test all those scenarios in TestSCMCertStore.java. Those kinds of tests are not possible for TestDefaultCAServer because we use a MockCAServer instance to unit test the CA server.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
vivekratnavel: Good suggestion @xiaoyuyao. But, having this sequence id stored in RocksDB gives us the ability to batch transactions when revoke certificates operation is executed. If we use the version file to keep track of this id, we will lose this ability and might cause conflicts with the CRL sequence id.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/server/TestSCMCertStore.java
vivekratnavel: Good catch @xiaoyuyao. I have now changed the signature of this method to return an optional CRL sequence id. It will return empty if no new CRL was generated. Otherwise, it will return the sequence id which can be returned to the users to keep track of the status of revocations in the future.

vivekratnavel:
@xiaoyuyao Thanks for the detailed review! I have posted a patch to resolve most of your comments. Please review again and let me know of any more changes needed.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
xiaoyuyao: Good point. I'm OK with CRL sequence id in the table for atomic CRL operations using rocksDB.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMCertStore.java
```diff
@@ -70,35 +85,69 @@ public void storeValidCertificate(BigInteger serialID,
   }
 
   @Override
-  public void revokeCertificate(BigInteger serialID) throws IOException {
+  public Optional<Long> revokeCertificates(
+      List<BigInteger> serialIDs,
+      X509CertificateHolder caCertificateHolder,
+      CRLReason reason,
```
xiaoyuyao, line 91: Can we add a revoke time parameter for this interface method so that we don't need to modify the interface later to support scheduled revoke? The caller can pass now() until the scheduled revoke is implemented.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMCertStore.java
```diff
@@ -70,35 +85,69 @@ public void storeValidCertificate(BigInteger serialID,
...
+      CRLApprover crlApprover)
+      throws IOException {
+    Date now = new Date();
+    X509v2CRLBuilder builder =
+        new X509v2CRLBuilder(caCertificateHolder.getIssuer(), now);
+    List<X509Certificate> certsToRevoke = new ArrayList<>();
+    X509CRL crl;
+    Optional<Long> sequenceId = Optional.empty();
```
xiaoyuyao, line 99: Can you add some javadoc on the interface for the case where non of the certs are revoked due to either not found or already revoked? 
In this case, the caller need to check the returned Optional for empty.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/server/TestSCMCertStore.java
```diff
@@ -0,0 +1,248 @@
...
+            .findAny());
+
+    assertNotNull(
+        revokedCertificates.stream().filter(c ->
+            c.getSerialNumber().equals(newSerialIDs.get(1)))
+            .findAny());
+
+    // Valid certs table should have 1 cert
+    size = 0;
+    TableIterator<BigInteger,
```
xiaoyuyao, line 219: NIT: this can be a helper function like getTableSize() to avoid duplicated code.

xiaoyuyao (commented):
Thanks @vivekratnavel  for the update. PR looks good to me overall. Just few minor comments inline.

vivekratnavel:
@xiaoyuyao I have fixed all the concerns you had. Can you please take another look? Thank you!

xiaoyuyao (approved):
LGTM, +1.

PR #1877, opened 2021-02-02
HDDS-4771. [FSO]S3MultiPart: Implement InitiateMultiPartUpload

rakeshadr:

This task is to implement Initiate multipart upload for a specified key.

https://issues.apache.org/jira/browse/HDDS-4771

## How was this patch tested?

Added S3 multipart UT cases.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
@@ -35,6 +35,7 @@ public class OmMultipartKeyInfo extends WithObjectID {
   private final ReplicationType replicationType;
   private final ReplicationFactor replicationFactor;
   private TreeMap<Integer, PartKeyInfo> partKeyInfoList;
+  private long parentID;
 
   /**
    * Construct OmMultipartKeyInfo object which holds multipart upload
@@ -50,6 +51,29 @@ public class OmMultipartKeyInfo extends WithObjectID {
     this.partKeyInfoList = new TreeMap<>(list);
     this.objectID = objectID;
     this.updateID = updateID;
+  }
+
+  /**
+   * Construct OmMultipartKeyInfo object which holds multipart upload
+   * information for a key.
+   */
+  @SuppressWarnings("parameternumber")
+  public OmMultipartKeyInfo(String id, long creationTime,
+      ReplicationType replicationType, ReplicationFactor replicationFactor,
+      Map<Integer, PartKeyInfo> list, long objectID, long updateID,
+      long parentObjId) {
+    this(id, creationTime, replicationType, replicationFactor, list, objectID,
+            updateID);
+    this.parentID = parentObjId;
+  }
+
+  /**
+   * Returns parentID.
+   *
+   * @return long
+   */
+  public long getParentID() {
+    return parentID;
   }
 
   /**
@@ -95,6 +119,7 @@ public class OmMultipartKeyInfo extends WithObjectID {
     private TreeMap<Integer, PartKeyInfo> partKeyInfoList;
     private long objectID;
     private long updateID;
+    private long parentID;
 
     public Builder() {
       this.partKeyInfoList = new TreeMap<>();
@@ -144,9 +169,14 @@ public class OmMultipartKeyInfo extends WithObjectID {
       return this;
     }
 
+    public Builder setParentID(long parentObjId) {
+      this.parentID = parentObjId;
+      return this;
+    }
+
     public OmMultipartKeyInfo build() {
       return new OmMultipartKeyInfo(uploadID, creationTime, replicationType,
-          replicationFactor, partKeyInfoList, objectID, updateID);
+          replicationFactor, partKeyInfoList, objectID, updateID, parentID);
     }
   }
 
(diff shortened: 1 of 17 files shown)


rakeshadr:
@linyiqun , appreciate if you could review and give your feedback. Thanks in advance!

linyiqun (commented):
@rakeshadr , I see original S3InitiateMultipartUploadRequest doesn't support FS semantic (not add parent key if it doesn't exist). So  it's still intend to implement V1 version here?

rakeshadr:
> till intend to implement V1 version here?

Good point @linyiqun. Yes, V1 version will make 100% FS semantics for multi part request as well. IMHO, its good to follow same pattern for all the ObjectStore APIs.
Does it make sense to you?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
```diff
@@ -35,6 +35,7 @@
   private final ReplicationType replicationType;
   private final ReplicationFactor replicationFactor;
   private TreeMap<Integer, PartKeyInfo> partKeyInfoList;
+  private long parentID;
```
linyiqun, line 38: Can we add some comments for parentID? Something like what's the meaning of this and when this  parentID will make sense.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -510,6 +512,28 @@ public static void addToOpenFileTable(OMMetadataManager omMetadataMgr,
...
+  /**
+   * Adding omKeyInfo to open file table.
+   *
+   * @param omMetadataMgr OM Metadata Manager
+   * @param batchOp       batch of db operations
+   * @param omFileInfo    omKeyInfo
+   * @param uploadID      uploadID
+   * @throws IOException DB failure
+   */
+  public static void addToOpenFileTable(OMMetadataManager omMetadataMgr,
```
linyiqun, line 525: I didn't see this method is used in current PR change. Suppose this should be used in S3InitiateMultipartUploadResponseV1#addToDBBatch, right?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
```diff
@@ -0,0 +1,182 @@
...
+   * Ozone is made active by setting OZONE_ENABLED = true
+   *
+   * @throws IOException
+   */
+  @BeforeClass
+  public static void init() throws Exception {
+    OzoneConfiguration conf = new OzoneConfiguration();
+    conf.setBoolean(OMConfigKeys.OZONE_OM_ENABLE_FILESYSTEM_PATHS, true);
+    conf.set(OMConfigKeys.OZONE_OM_LAYOUT_VERSION,
+            OMConfigKeys.OZONE_OM_LAYOUT_VERSION_V1);
```
linyiqun, line 74: Can we reuse existed method TestOMRequestUtil#configureFSOptimizedPaths for this?

linyiqun (commented):
Haven't taken a deep look for the unit tests, some minor comments below. Please have a check @rakeshadr .

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/helpers/OmMultipartKeyInfo.java
rakeshadr: Sure, I have pushed the changes in new commit. I updated same comments in OmKeyInfo class as well.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
rakeshadr: Done!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
rakeshadr: Done!

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/s3/multipart/TestS3InitiateMultipartUploadResponseV1.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.UUID;
+
+/**
+ * Class tests S3 Initiate MPU response.
+ */
+public class TestS3InitiateMultipartUploadResponseV1
+    extends TestS3InitiateMultipartUploadResponse {
+
+  @Test
+  public void addDBToBatch() throws Exception {
```
linyiqun, line 39: Can we rename this to testAddDBToBatch

linyiqun (commented):
Looks great now, one nit for the unit test.

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/s3/multipart/TestS3InitiateMultipartUploadResponseV1.java
rakeshadr: Sure, I've renamed it.

rakeshadr:
> Looks great now, one nit for the unit test.

Thanks, Could you please approve the PR after final review.

linyiqun (approved):
>Thanks, Could you please approve the PR after final review.
I'm +1 for this.

rakeshadr:
Thanks a lot @linyiqun for the useful review comments. I've merged the PR.

PR #1882, opened 2021-02-02
HDDS-4774. Disable Prevote in Ratis for Ozone by default.

bshashikant:
Pre-vote and Leader Lease are new features in Ratis which may not be required/not tested enough with ozone currently. The idea here is to disable these by default in ozone.

Leader lease is yet be committed : https://issues.apache.org/jira/browse/RATIS-1273

https://issues.apache.org/jira/browse/HDDS-4774

## How was this patch tested?
Config change. No tests needed.


adoroszlai (approved):
Thanks @bshashikant for working this.

amaliujia:
LGTM thanks!

PR #1884, opened 2021-02-02
HDDS-4784. TestRootedOzoneFileSystem is not picking the configured 'Parameters'

rakeshadr:

This is a follow-up Jira to update TestRootedOzoneFileSystem test class and run against all the configured parameters combinations.

https://issues.apache.org/jira/browse/HDDS-4784

## How was this patch tested?

Through existing TestRootedOzoneFileSystem case.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -55,7 +55,6 @@ import org.apache.hadoop.test.GenericTestUtils;
 import org.apache.hadoop.test.LambdaTestUtils;
 import org.junit.AfterClass;
 import org.junit.Assert;
-import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.Timeout;
@@ -87,6 +86,7 @@ import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.BUCK
 import static org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes.VOLUME_NOT_FOUND;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 /**
  * Ozone file system tests that are not covered by contract tests.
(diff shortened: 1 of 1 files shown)


rakeshadr:
**Note:** Following test cases are failing in my local environment. It is expected to fail in QA as well and need to analyse it for the cause.

```
Parameters:
0: new Object[]{true, true}
1: new Object[]{true, false}
```
```

testDeleteCreatesFakeParentDir[0]
testMkdirOnNonExistentVolumeBucketDir[0]
testListStatusOnLargeDirectory[0]
testListStatusInBucket[0]
```
```

testDeleteCreatesFakeParentDir[1]
testMkdirOnNonExistentVolumeBucketDir[1]
testListStatusOnLargeDirectory[1]
testListStatusInBucket[1]
```

ayushtkn:
Thanx @rakeshadr for taking this up.
I think only two `testDeleteCreatesFakeParentDir` and `testMkdirOnNonExistentVolumeBucketDir` are broken, the other two are failing because, these two failed test didn't clean up properly.

Doesn't look like a bug, should be a test only issue. Changing to `fs.exists(key)` from getting key and validating fixes all of them(for me). Give a check if that does for you too.
Something like this, Just for reference:
```
diff --git a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
index 98e535496..777760675 100644
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -247,12 +247,8 @@ public void testDeleteCreatesFakeParentDir() throws Exception {
     // Delete the child key
     Assert.assertTrue(fs.delete(child, false));
 
-    // Deleting the only child should create the parent dir key if it does
-    // not exist
-    OFSPath parentOFSPath = new OFSPath(parent);
-    String parentKey = parentOFSPath.getKeyName() + "/";
-    OzoneKeyDetails parentKeyInfo = getKey(parent, true);
-    Assert.assertEquals(parentKey, parentKeyInfo.getName());
+    // Deleting the only child should not delete the parent
+    Assert.assertNotNull(fs.exists(parent));
 
     // Recursive delete with DeleteIterator
     Assert.assertTrue(fs.delete(grandparent, true));
@@ -339,11 +335,7 @@ public void testMkdirOnNonExistentVolumeBucketDir() throws Exception {
 
     // Check volume and bucket existence, they should both be created.
     OzoneVolume ozoneVolume = objectStore.getVolume(volumeNameLocal);
-    OzoneBucket ozoneBucket = ozoneVolume.getBucket(bucketNameLocal);
-    OFSPath ofsPathDir1 = new OFSPath(dir12);
-    String key = ofsPathDir1.getKeyName() + "/";
-    OzoneKeyDetails ozoneKeyDetails = ozoneBucket.getKey(key);
-    Assert.assertEquals(key, ozoneKeyDetails.getName());
+    Assert.assertTrue(fs.exists(dir12));
 
     // Verify that directories are created.
     FileStatus[] fileStatuses = ofs.listStatus(root);

```
HDDS-4102, seems to have introduced this with `enabledFileSystemPaths` it `normalizes` the key path.

elek:
/pending pending on comments from @ayushtkn and failing tests

PR #1887, opened 2021-02-02
HDDS-4730. Use separate Ratis admin and client ports

adoroszlai:

Move Datanode's Ratis server-to-server and admin communication to other ports and use separate TLS config.

 * Add new ports with default config.
 * Only use new ports if datanode has not been initialized previously.  Ratis stores peer addresses with port in Raft log, so if the ring is already created, we cannot simply move server-to-server communication to another port.
 * Store "separate ports" as feature in datanode details YAML.  Currently this is the only feature, but I think it's better to have a list of features than to introduce "random" fields.
 * Add acceptance test for restarting Ozone (with the same version, no upgrade).
 * Enabled TLS for datanode GRPC in `ozonesecure` cluster.

https://issues.apache.org/jira/browse/HDDS-4730

## How was this patch tested?

Regular CI:
https://github.com/adoroszlai/hadoop-ozone/actions/runs/531656927


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/protocol/DatanodeDetails.java
```diff
@@ -73,6 +74,7 @@
   private String buildDate;
   private HddsProtos.NodeOperationalState persistedOpState;
   private long persistedOpStateExpiryEpochSec = 0;
+  private final List<String> features = new ArrayList<>();
```
amaliujia, line 77: I might have missed something:

Is there a JIRA or design doc that says what features in Datanode are?

adoroszlai: This is just a way to help distinguish newly initialized datanodes from existing ones being restarted with new code version.  No further features planned currently.

bshashikant (approved):

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
```diff
@@ -129,6 +129,8 @@
   public static final String OZONE_DB_CHECKPOINT_REQUEST_FLUSH =
       "flushBeforeCheckpoint";
 
+  public static final String SEPARATE_RATIS_PORTS_AVAILABLE = "RatisPorts";
+
```
lokeshj1703, line 133: Can we define features in a separate class? We could also define them as enums.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
```diff
@@ -413,71 +444,68 @@ public static XceiverServerRatis newXceiverServerRatis(
...
+      adminPort = getRealPort(serverRpc.getAdminServerAddress(),
+          Port.Name.RATIS_ADMIN);
+      serverPort = getRealPort(serverRpc.getInetSocketAddress(),
+          Port.Name.RATIS_SERVER);
 
       isStarted = true;
     }
   }
 
+  private int getRealPort(InetSocketAddress address, Port.Name name) {
```
lokeshj1703, line 501: The real port should be same as configured port?

lokeshj1703 (commented):
@adoroszlai Thanks for working on this! The changes look good to me. I have few minor comments inline.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
adoroszlai: Yes, I can move it to separate class.  I'd rather avoid enums, which are more type-safe, but worse for compatibility.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
adoroszlai: They usually are the same, but can be different when random ports are used.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
adoroszlai: Although not sure if it's worth extracting to another class at this time.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
lokeshj1703: Got it.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
lokeshj1703: I was suggesting separate class so that in future if new features are defined, it could all be in the same class for better accessibility.

adoroszlai: Thanks @lokeshj1703 for the suggestion.  Extracted to separate class.  Also changed to using version numbers instead of "features".  I think this will be simpler in the long run.  Features of the datanode can be mapped at runtime (eg. if version > 0, then we have this "separate ports" feature).

adoroszlai:
@elek @bshashikant @lokeshj1703 @amaliujia Thanks for the reviews so far.  Do you have any comment on the latest patch, can I merge it?

lokeshj1703 (approved):
@adoroszlai Thanks for updating the PR! The changes look good to me. +1.

adoroszlai:
Thanks all for the review.

PR #1888, opened 2021-02-03
HDDS-4787. Trash emptier fails to create checkpoints in a secure setup

sadanand48:
Since in `TrashOzoneFilesystem` the calls are internal and not through RPC, NPE is generated during `checkacls`(), Since `Server.getRemoteIp` and other such parameters would be null, the security authorizers too generate an NPE. The fix here is if `Server.getRemoteIp()` is null replace it with ip of the node that starts OM

https://issues.apache.org/jira/browse/HDDS-4787#

## How was this patch tested?
Existing unit tests. also tried on cluster [https://github.com/sadanand48/hadoop-ozone/runs/1825648963?check_suite_focus=true](https://github.com/sadanand48/hadoop-ozone/runs/1825648963?check_suite_focus=true)

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -3618,11 +3618,11 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
     Pair<String, String> resolved;
     try {
       if (isAclEnabled) {
+        InetAddress remoteIp = Server.getRemoteIp();
         resolved = resolveBucketLink(requested, new HashSet<>(),
             Server.getRemoteUser(),
-            Server.getRemoteIp(),
-            Server.getRemoteIp() != null ?
-                Server.getRemoteIp().getHostName() :
+            remoteIp,
+            remoteIp != null ? remoteIp.getHostName() :
                 omRpcAddress.getHostName());
       } else {
         resolved = resolveBucketLink(requested, new HashSet<>(),
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -1698,16 +1698,12 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
   private void checkAcls(ResourceType resType, StoreType store,
       ACLType acl, String vol, String bucket, String key)
       throws IOException {
+    UserGroupInformation user = ProtobufRpcEngine.Server.getRemoteUser();
+    InetAddress remoteIp = ProtobufRpcEngine.Server.getRemoteIp();
     checkAcls(resType, store, acl, vol, bucket, key,
-        ProtobufRpcEngine.Server.getRemoteUser() != null ?
-            ProtobufRpcEngine.Server.getRemoteUser() :
-            getRemoteUser(),
-        ProtobufRpcEngine.Server.getRemoteIp() != null ?
-            ProtobufRpcEngine.Server.getRemoteIp() :
-            omRpcAddress.getAddress(),
-        ProtobufRpcEngine.Server.getRemoteIp() != null ?
-            ProtobufRpcEngine.Server.getRemoteIp().getHostName() :
-            omRpcAddress.getHostName(),
+        user != null ? user : getRemoteUser(),
+        remoteIp != null ? remoteIp : omRpcAddress.getAddress(),
+        remoteIp != null ? remoteIp.getHostName() : omRpcAddress.getHostName(),
         true, getVolumeOwner(vol, acl, resType));
   }
 
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -31,6 +31,7 @@ import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.security.KeyPair;
 import java.security.PrivateKey;
+import java.security.PrivilegedExceptionAction;
 import java.security.PublicKey;
 import java.security.cert.CertificateException;
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashOzoneFileSystem.java
```diff
@@ -490,15 +497,38 @@ boolean processKeyPath(List<String> keyPathList) {
...
+    InetAddress remoteAddress = ozoneManager.getOmRpcServerAddr().getAddress();
+    OzoneManagerProtocolProtos.UserInfo.Builder userInfo =
+        OzoneManagerProtocolProtos.UserInfo.newBuilder();
+    if (user != null) {
+      userInfo.setUserName(user.getUserName());
+    }
+
+    if (remoteAddress != null) {
+      userInfo.setHostName(remoteAddress.getHostName());
+      userInfo.setRemoteAddress(remoteAddress.getHostAddress()).build();
```
adoroszlai, line 526:
```suggestion
      userInfo.setRemoteAddress(remoteAddress.getHostAddress());
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1688,11 +1697,17 @@ public void createVolume(OmVolumeArgs args) throws IOException {
...
   private void checkAcls(ResourceType resType, StoreType store,
       ACLType acl, String vol, String bucket, String key)
-      throws OMException {
+      throws IOException {
     checkAcls(resType, store, acl, vol, bucket, key,
-        ProtobufRpcEngine.Server.getRemoteUser(),
-        ProtobufRpcEngine.Server.getRemoteIp(),
-        ProtobufRpcEngine.Server.getRemoteIp().getHostName(),
+        ProtobufRpcEngine.Server.getRemoteUser() != null ?
+            ProtobufRpcEngine.Server.getRemoteUser() :
```
adoroszlai, line 1703: Nit: can you please store these in local variables instead of multiple calls?

adoroszlai (changes requested):
Thanks @sadanand48 for working on this.

I think secure acceptance test failure is related:

```
om_1        | 2021-02-03 05:18:00,113 [OM StateMachine ApplyTransaction Thread - 0] ERROR ratis.OzoneManagerStateMachine: Terminating with exit status 1: Request cmdType: RenameKey
om_1        | clientId: "client-6670A68B0F8D"
om_1        | userInfo {
om_1        | }
om_1        | renameKeyRequest {
om_1        |   keyArgs {
om_1        |     volumeName: "fstest1-src"
om_1        |     bucketName: "link1-o3fs-src"
om_1        |     keyName: ".Trash/root/Current/"
om_1        |     modificationTime: 1612329480031
om_1        |   }
om_1        |   toKeyName: ".Trash/root/210203051800/"
om_1        | }
om_1        | failed with exception
om_1        | java.lang.NullPointerException
om_1        | 	at org.apache.hadoop.ozone.security.acl.OzoneNativeAuthorizer.isAdmin(OzoneNativeAuthorizer.java:222)
om_1        | 	at org.apache.hadoop.ozone.security.acl.OzoneNativeAuthorizer.checkAccess(OzoneNativeAuthorizer.java:92)
om_1        | 	at org.apache.hadoop.ozone.om.OzoneManager.checkAcls(OzoneManager.java:1797)
om_1        | 	at org.apache.hadoop.ozone.om.request.OMClientRequest.checkAcls(OMClientRequest.java:176)
om_1        | 	at org.apache.hadoop.ozone.om.request.OMClientRequest.checkAcls(OMClientRequest.java:154)
om_1        | 	at org.apache.hadoop.ozone.om.request.key.OMKeyRequest.checkKeyAcls(OMKeyRequest.java:437)
om_1        | 	at org.apache.hadoop.ozone.om.request.key.OMKeyRenameRequest.validateAndUpdateCache(OMKeyRenameRequest.java:143)
```

https://github.com/apache/ozone/runs/1819931384#step:7:549
https://github.com/apache/ozone/suites/1956204276/artifacts/38766607

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -3605,14 +3620,20 @@ public ResolvedBucket resolveBucketLink(Pair<String, String> requested)
...
-    if (isAclEnabled) {
-      resolved = resolveBucketLink(requested, new HashSet<>(),
-              Server.getRemoteUser(),
-              Server.getRemoteIp(),
-              Server.getRemoteIp().getHostName());
-    } else {
-      resolved = resolveBucketLink(requested, new HashSet<>(),
-          null, null, null);
+    try {
+      if (isAclEnabled) {
```
xiaoyuyao, line 3620: agree with @adoroszlai can we use a local variable to hold the result of Server.getRemoteIp().

xiaoyuyao (commented):
Thanks @sadanand48 for working on this. Patch LGTM, just some minor comments. Can you add a test to cover this?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
sadanand48: Done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/TrashOzoneFileSystem.java
sadanand48: Done.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
sadanand48: done

sadanand48:
Thanks @xiaoyuyao for reviewing this .` TestOzoneFilesystem testTrash()` is the unit test for this and this issue would arise when `OZONE_ACL_ENABLED `is set to true. I have made that change.  There is also a robot test case for trash deletion inside `ozone-fs.robot` . The secure acceptance CI check will test that in a secure environment with native acls enabled.

sadanand48:
Thanks @adoroszlai for the review and pointing out the error in the secure acceptance check. It's passing now.

adoroszlai (approved):

adoroszlai:
Thanks @sadanand48 for updating the patch.

mukul1987 (approved):
+1

PR #1889, opened 2021-02-03
HDDS-4789. Improve usability of ozone s3 getsecret output

elek:
JIRA: https://issues.apache.org/jira/browse/HDDS-4789

I am testing the S3g a lot in secure environment.

It's slightly painful to set the AWS acces key Id and secret every time.

I propose to improve the usability to print out the credentials in a read-to-use format:

```
ozone s3 getsecret 
AWS_ACCESS_KEY_ID=testuser/scm@EXAMPLE.COM
AWS_SECRET_ACCESS_KEY=...
```

Instead of:
```
ozone s3 getsecret 
awsAccessKey=testuser/scm@EXAMPLE.COM
awsSecret=...
```

The `-e` option even adds the required `export` prefixes:

```
ozone s3 getsecret -e 
export AWS_ACCESS_KEY_ID=testuser/scm@EXAMPLE.COM
export AWS_SECRET_ACCESS_KEY=...
```

Which makes the usage as easy as :

```
eval $(ozone s3 getsecret -e)
```

or for docker-compose based environments:

```
eval $(docker-compose exec scm ozone s3 getsecret -e)
```

## How was this patch tested?

From CLI and from acceptance tests

### DIFF
--- a/hadoop-ozone/dist/src/main/smoketest/s3/commonawslib.robot
+++ b/hadoop-ozone/dist/src/main/smoketest/s3/commonawslib.robot
@@ -62,9 +62,9 @@ Setup v4 headers
 
 Setup secure v4 headers
     ${result} =         Execute                    ozone s3 getsecret
-    ${accessKey} =      Get Regexp Matches         ${result}     (?<=AWS_ACCESS_KEY_ID=).*
+    ${accessKey} =      Get Regexp Matches         ${result}     (?<=awsAccessKey=).*
     ${accessKey} =      Get Variable Value         ${accessKey}  sdsdasaasdasd
-    ${secret} =         Get Regexp Matches         ${result}     (?<=AWS_SECRET_ACCESS_KEY=).*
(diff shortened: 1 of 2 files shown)


adoroszlai (commented):
Thanks @elek for this improvement, I think it's very nice.  However, I'd like to ask you to make this new format optional: add another flag to turn it on (maybe the same `-e` flag being [repeatable](https://picocli.info/#_repeated_boolean_options): `-e` for env-var format, `-ee` (or `-e -e`) for additional `export` prefix).  I know it's a bit of extra complexity, but lets us avoid breaking some downstream scripts we are not aware of.

elek:
Thanks for the feedback @adoroszlai. If we couldn't improve the default format, I don't see any benefit for changing it. I don't think that anybody would turn on a flag just to have better usability. 

In that case I would suggest to keep only the `-e` parameter from the patch which introduces the new, `export ...` output, and the old output can be kept as before...

adoroszlai (approved):

PR #1892, opened 2021-02-04
HDDS-4791. StateContext.getReports may return list with size larger t…

symious:
…he maxLimit

- Fix the logic to return the correct size of the result list.

https://issues.apache.org/jira/browse/HDDS-4791

## How was this patch tested?

Unit Test.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
@@ -318,14 +318,7 @@ public class StateContext {
     return reportsToReturn;
   }
 
-  /**
-   * Returns available reports from the report queue with a max limit on
-   * list size, or empty list if the queue is empty.
-   *
-   * @return List of reports
-   */
-  public List<GeneratedMessage> getReports(InetSocketAddress endpoint,
-                                           int maxLimit) {
+  List<GeneratedMessage> getNonIncrementalReports() {
     List<GeneratedMessage> nonIncrementalReports = new LinkedList<>();
     GeneratedMessage report = containerReports.get();
     if (report != null) {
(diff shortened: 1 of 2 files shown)


symious:
@elek @adoroszlai Could you help to review this PR? Thanks in advance.

dineshchitlangia (approved):
+1 @symious  thanks for the improvement

adoroszlai (commented):
Thanks @symious for finding this discrepancy introduced in HDDS-4404.

The proposed solution has a problem (`IndexOutOfBoundsException`) in edge case: if `maxLimit` is smaller than the number of available misc. reports.  And if they are equal, `getIncrementalReports` is called unnecessarily as it will be limited to 0 items.

Also, I'm not sure behavior for `maxLimit < 3` can be well defined other than throwing some exception.

Since `getReports` is called from only one place, with unlimited list size:

https://github.com/apache/ozone/blob/df38294e34544d073fe2e3c85f329d53f88e70c3/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java#L302

I think it would be better to just remove `maxLimit` parameter.

smengcl:
Thanks @symious for the patch. Yes this was an oversight.

I have the same concern as @adoroszlai mentioned. `maxLimit - nonIncrementalReports.size() < 0` can be a problem.

I agree that we should just remove the `maxLimit` param from `getReports()`:

https://github.com/apache/ozone/blob/6b67ae0c40110af195b897fb70a87dc6e82d467e/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java#L327-L330

and put `getIncrementalReports(endpoint, Integer.MAX_VALUE - 3)` inside, maybe.

symious:
@adoroszlai @smengcl Thanks for the review, added a commit to handle the edge case.

smengcl (approved):
+1 on the latest change.

A caveat is if maxLimit <= 3, the getReports may always return the same set of reports because once containerReports/nodeReport/pipelineReports are set, they are only updated but not removed. It makes most sense if the caller of getReports specifies maxLimit > 4.

symious:
@smengcl @adoroszlai Thanks for the review.

PR #1893, opened 2021-02-04
HDDS-4740. Admin command should take effect on all SCM instance

GlenGeng:

Scope
admin command includes rm start/stop and safe mode exit.
 
Requirement
1, When admin stops rm, rm in all SCM should stop, re-election should not trigger rm to start in the new leader.
2, When admin starts rm, only rm in leader and out of safe mode should take effect. Given leader is in safe mode, even if admin starts rm explicitly, it does not take effect.
3, This admin rm start/stop can not survive restart for a SCM instance. When admin decides to stop rm of the SCM cluster, he should pay attention if any of the SCM crashes.
 
Status
1, For now, admin rm start/stop will create/destroy the rm thread.
2, SCMContainerLocationFailoverProxyProvider has been proxied by FailoverProxyProvider, it will round robin SCMs in ozone.scm.names, until it is successfully handled. In ServerSide, whenever receiving a client request, it do isLeader check first, return nle to trigger fpp to failover to the next SCM.
3, SCMService decides the next iteration of rm to take effect or not by changing RUNNING and PAUSING.
 
Solution:
When receiving a rm stop/start request on the server side, SCM skip the isLeader check, just destroys/creates rm thread, client side fake an exception to trigger fpp to try the next SCM in a round robin way.
The Running and PAUSING status and rm start/stop can be treated separately. The admin operations and the raft status are requirements of two dimensions.
 
We can achieve above requirements:
1, When admin stops rm, rm in all SCM should stop, re-election should not trigger rm to start in the new leader.
Meet, admin rm start destroy rm thread in all SCM.
 
2, When admin starts rm, only rm in leader and out of safe mode should take effect. Given leader is in safe mode, even if admin starts it explicitly, rm does not take effect.
Meet, admin rm stop create rm thread in all SCM, but SCMStatus is decided by leader and safe mode.
 
3, This admin rm start/stop can not survive restart for a SCM instance. When admin decides to stop rm of the SCM cluster, he should pay attention if any of the SCM crashes.
Meet. The is actually a relax item. 

https://issues.apache.org/jira/browse/HDDS-4740

## How was this patch tested?

CI and 3 SCM cluster.

### DIFF
--- a/hadoop-hdds/common/pom.xml
+++ b/hadoop-hdds/common/pom.xml
@@ -189,6 +189,10 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <artifactId>hadoop-hdds-interface-client</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-hdds-interface-admin</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.hamcrest</groupId>
       <artifactId>hamcrest-all</artifactId>
       <scope>test</scope>
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
@@ -18,6 +18,7 @@
 package org.apache.hadoop.hdds.scm.protocol;
 
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.hadoop.hdds.protocol.proto.StorageContainerLocationProtocolProtos.Type;
 import org.apache.hadoop.hdds.scm.ScmConfig;
 import org.apache.hadoop.hdds.scm.ScmInfo;
 import org.apache.hadoop.hdds.scm.container.common.helpers.ContainerWithPipeline;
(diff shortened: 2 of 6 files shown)


GlenGeng:
@bshashikant @nandakumar131  @amaliujia @runzhiwang Please take a look! This is a followup of Phase 2.0.

I cherry-picked a commit from master, will reset it after HDDS-2823 is fixed.

bshashikant:
@GlenGeng , can you please rebase now as the build issue with CI is fixed?

bshashikant (commented):
The changes look good. @GlenGeng , is there a race condition where, replicationManager can go from stop to start state bcoz of leader election and admin command execution concurrently?

bshashikant:
@GlenGeng , can you please rebase?

GlenGeng:
Sure. Will rebase upon the latest HDDS-2823

GlenGeng:
> The changes look good. @GlenGeng , is there a race condition where, replicationManager can go from stop to start state bcoz of leader election and admin command execution concurrently?

There will be no race condition here. start/stop RM and becoming leader/step down are two different conditions protected by two locks.

stop/start ReplicationManager will kill/create the thread which are all synchronized methods of ReplicationManager, meanwhile becoming leader or step down will call notifyStatusChanged() to change `serviceStatus`, and this modification is protected by the `serviceLock`. 

When the thread of ReplicationManager is running, it call `shouldRun()` to decide if it should take effect or skip, and `shouldRun()`  need lock `serviceLock`.

GlenGeng:
@bshashikant Comments addressed. Please take another look!

bshashikant:
@GlenGeng , can you plz rebase?

GlenGeng:
> @GlenGeng , can you plz rebase?

will do. After catch up with the changes in 2823 during last 10 days.

GlenGeng:
@bshashikant Please take a look.

bshashikant (approved):

bshashikant:
Thanks @GlenGeng for the contribution.

PR #1894, opened 2021-02-04
HDDS-4792. Remove redundant state check for handleDeleteChunk

symious:

- Remove redundant state check for handleDeleteChunk.

https://issues.apache.org/jira/browse/HDDS-4792

## How was this patch tested?

Locally test.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueHandler.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueHandler.java
@@ -653,14 +653,6 @@ public class KeyValueHandler extends Handler {
       return malformedRequest(request);
     }
 
-    // The container can become unhealthy after the lock is released.
-    // The operation will likely fail/timeout in that happens.
-    try {
-      checkContainerIsHealthy(kvContainer);
-    } catch (StorageContainerException sce) {
-      return ContainerUtils.logAndReturnError(LOG, sce, request);
-    }
-
     try {
       checkContainerOpen(kvContainer);
 


symious:
@elek @adoroszlai Could you help to review this PR? Thanks in advance.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueHandler.java
```diff
@@ -653,14 +653,6 @@ ContainerCommandResponseProto handleDeleteChunk(
       return malformedRequest(request);
     }
 
-    // The container can become unhealthy after the lock is released.
-    // The operation will likely fail/timeout in that happens.
-    try {
-      checkContainerIsHealthy(kvContainer);
-    } catch (StorageContainerException sce) {
-      return ContainerUtils.logAndReturnError(LOG, sce, request);
-    }
```
ayushtkn, line 662: Where is this check available in ``handleDeleteChunk`` path? I tried checking callers of the method, couldn't find where this is checked before. Can you point me to that

symious: Sorry, I didn't get the question? Do you mean where the function is called before?

ayushtkn: Yes

symious: I didn't find the usage of DeleteChunk either. 
I think the method of `deleteChunk` in `ChunkManager` is designed for `FilePerChunkStrategy` first that it will only delete one chunk file, in `FilePerBlockStrategy` the method should throw the exception of unsupported request.

elek (approved):
+1. Thanks the patch @symious 

Checked the code and I agree, this check is redundant.

PR #1898, opened 2021-02-05
HDDS-3974. StateContext#addPipelineActionIfAbsent does not work as expected

smengcl:
https://issues.apache.org/jira/browse/HDDS-3974

The `break` in the loop doesn't actually prevent the pipeline action from being added twice. Should be fixed.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
@@ -411,12 +411,22 @@ public class StateContext {
     }
   }
 
-  boolean isSamePipelineAction(PipelineAction action1, PipelineAction action2) {
-    return action1.getAction() == action2.getAction()
-        && action1.hasClosePipeline()
-        && action2.hasClosePipeline()
-        && action1.getClosePipeline().getPipelineID()
-        .equals(action2.getClosePipeline().getPipelineID());
+  /**
+   * Helper function for addPipelineActionIfAbsent that check if inputs are the
+   * same close pipeline action.
+   *
+   * Important Note: Make sure to double check for correctness before using this
+   * helper function for other purposes!
+   *
+   * @return true if a1 and a2 are the same close pipeline action,
+   *         false otherwise
+   */
+  boolean isSameClosePipelineAction(PipelineAction a1, PipelineAction a2) {
+    return a1.getAction() == a2.getAction()
+        && a1.hasClosePipeline()
+        && a2.hasClosePipeline()
+        && a1.getClosePipeline().getPipelineID()
+        .equals(a2.getClosePipeline().getPipelineID());
   }
 
   /**
@@ -438,7 +448,7 @@ public class StateContext {
         final Queue<PipelineAction> actionsForEndpoint =
             pipelineActions.get(endpoint);
         if (actionsForEndpoint.stream().noneMatch(
-            action -> isSamePipelineAction(action, pipelineAction))) {
+            action -> isSameClosePipelineAction(action, pipelineAction))) {
           actionsForEndpoint.add(pipelineAction);
         }
       }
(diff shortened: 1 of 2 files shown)


symious:
I think the action should also be sent to other endpoints?

smengcl:
> I think the action should also be sent to other endpoints?

You are right. Should use a flag. Also working on a UT.

symious:
I think it may be more concise to put the addition in the if clause?

smengcl:
@symious done, tho in a slightly different way :)

symious:
> @symious done, tho in a slightly different way :)

Oh, I was proposing
```
        for (PipelineAction pipelineActionIter : actionsForEndpoint) {
          if (pipelineActionIter.getAction() == pipelineAction.getAction()
              && pipelineActionIter.hasClosePipeline() && pipelineAction
              .hasClosePipeline()
              && pipelineActionIter.getClosePipeline().getPipelineID()
              .equals(pipelineAction.getClosePipeline().getPipelineID())) {
            actionsForEndpoint.add(pipelineAction);
            break;
          }
```

smengcl:
> > @symious done, tho in a slightly different way :)
> 
> Oh, I was proposing
> 
> ```
>         for (PipelineAction pipelineActionIter : actionsForEndpoint) {
>           if (pipelineActionIter.getAction() == pipelineAction.getAction()
>               && pipelineActionIter.hasClosePipeline() && pipelineAction
>               .hasClosePipeline()
>               && pipelineActionIter.getClosePipeline().getPipelineID()
>               .equals(pipelineAction.getClosePipeline().getPipelineID())) {
>             actionsForEndpoint.add(pipelineAction);
>             break;
>           }
> ```

hmm, IMO this loop wouldn't actually work as we expect. The goal of the loop inside was to iterate through all the existing actions before deciding if we need to queue it, IIUC?

symious:
> > > @symious done, tho in a slightly different way :)
> > 
> > 
> > Oh, I was proposing
> > ```
> >         for (PipelineAction pipelineActionIter : actionsForEndpoint) {
> >           if (pipelineActionIter.getAction() == pipelineAction.getAction()
> >               && pipelineActionIter.hasClosePipeline() && pipelineAction
> >               .hasClosePipeline()
> >               && pipelineActionIter.getClosePipeline().getPipelineID()
> >               .equals(pipelineAction.getClosePipeline().getPipelineID())) {
> >             actionsForEndpoint.add(pipelineAction);
> >             break;
> >           }
> > ```
> 
> hmm, IMO the proposed loop wouldn't actually work as we expect. The goal of the loop inside was to iterate through all the existing actions before deciding if we need to queue it, IIUC?

Oh, sorry, my mistake. Since we can't use labels, your options are both fine with me.

smengcl:
@symious No worries! Please take a look at the latest patch if you are available. Thanks!

Also pending CI.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
```diff
@@ -411,6 +411,14 @@ public void addContainerActionIfAbsent(ContainerAction containerAction) {
     }
   }
 
+  boolean isSamePipelineAction(PipelineAction action1, PipelineAction action2) {
```
avijayanhwx, line 414: Although there is only a "Close" pipeline action now, can we add a comment saying that this only works for close pipeline action? The method name seems generic even though it is not.

avijayanhwx (approved):
Thanks for working on this @smengcl. Logic looks good to me.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/StateContext.java
smengcl: Yes! Good suggestion

smengcl:
Thanks @symious and @avijayanhwx for reviewing this. Will merge this later in the day.

PR #1904, opened 2021-02-07
HDDS-4802. Fix an internal variable always not null in PipelinePlacementPolicy

lamberken:

`pick` always not null. `fallBackPickNodes` will throw exception if unable to find fall back node.

it's not a good way that throw exception in `fallBackPickNodes`. we can move the check logic to the end.

https://issues.apache.org/jira/browse/HDDS-4802

## How was this patch tested?

Update related tests.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/PipelinePlacementPolicy.java
@@ -237,8 +237,7 @@ public final class PipelinePlacementPolicy extends SCMCommonPlacementPolicy {
 
   // Fall back logic for node pick up.
   DatanodeDetails fallBackPickNodes(
-      List<DatanodeDetails> nodeSet, List<DatanodeDetails> excludedNodes)
-      throws SCMException{
+      List<DatanodeDetails> nodeSet, List<DatanodeDetails> excludedNodes) {
     DatanodeDetails node;
     if (excludedNodes == null || excludedNodes.isEmpty()) {
       node = chooseNode(nodeSet);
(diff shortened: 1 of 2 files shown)


ayushtkn (commented):
Makes sense to me, the other chooseNode methods seems to be returning null only when the can't find a datanode.
We need to update TestPipelineUpdataPolicy#testFallBackPickNodes too.
It is checking for the exception and error message from exception in other case, which it won't hit post this change.

lamber-ken:
> Makes sense to me, the other chooseNode methods seems to be returning null only when the can't find a datanode.
> We need to update TestPipelineUpdataPolicy#testFallBackPickNodes too.
> It is checking for the exception and error message from exception in other case, which it won't hit post this change.

OK，thanks for review，will check TestPipelineUpdataPolicy#testFallBackPickNodes.

lamber-ken:
> It is checking for the exception and error message from exception in other case, which it won't hit post this change.

The exception is from following logic, the logic has been moved to PipelinePlacementPolicy#getResultSet. 
So in TestPipelineUpdataPolicy#testFallBackPickNodes, already use `Assert.assertNotNull(node);` to check.

If my understanding is wrong, please correct me, thanks
```
    if (node == null) {
      String msg = String.format("Unable to find fall back node in" +
          " pipeline allocation. nodeSet size: %d", nodeSet.size());
      LOG.warn(msg);
      throw new SCMException(msg,
          SCMException.ResultCodes.FAILED_TO_FIND_SUITABLE_NODE);
    }
```

ayushtkn:
The test is asserting the exception also which will not happen now
```
catch (SCMException e) {
      Assert.assertEquals(SCMException.ResultCodes.FAILED_TO_FIND_SUITABLE_NODE,
          e.getResult());
    } 
```
This we can remove now, since it won't hit post this change.

In `testFallBackPickNodes` also we can get away from catching the SCMException, since that won't be thrown now.

elek (approved):
+1. The mentioned unit tests seems to be already fixed on the branch. 

Thanks the patch @lamberken, merging it now...

PR #1909, opened 2021-02-08
HDDS-4801. Skip coverage check for PRs and in forks

adoroszlai:

Currently _coverage_ CI check:

 1. calculates combined test coverage
 2. uploads it to Sonar only for Apache Ozone repo and only for builds on push/schedule
 3. stores combined coverage in GitHub Actions artifact

Thus for PR in Apache Ozone and for all builds in forks, it only stores coverage in the artifact.  These expire in 30 days and I don't think anybody really checks them manually.

I propose to completely skip _coverage_ check for PRs and in forks, instead of only skipping upload to Sonar.  This would save ~12 minutes for such builds.

https://issues.apache.org/jira/browse/HDDS-4801

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/runs/1849668371

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -256,6 +256,7 @@ jobs:
         continue-on-error: true
   coverage:
     runs-on: ubuntu-18.04
+    if: github.repository == 'apache/ozone' && github.event_name != 'pull_request'
     needs:
       - acceptance
       - integration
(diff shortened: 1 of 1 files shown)


elek (approved):
+1 I am fine with disabling it.

I have never had any plan to do something with the coverage data, but when I read the description of this PR I had an idea ;-) : it would be easy to create a Github actions scripts to compare the coverage artifacts between PR and master and create a report.

But we can turn it back anytime when we will have useful usage...

adoroszlai:
> it would be easy to create a Github actions scripts to compare the coverage artifacts between PR and master and create a report.

That's also what Codecov was supposed to do, but it had been disabled earlier due to unreliability.

adoroszlai:

PR #1912, opened 2021-02-08
HDDS-4788. Enable mTLS for Ratis in OM HA

adoroszlai:

Let OM HA enable TLS for its internal Ratis if `SecurityConfig#isGrpcTlsEnabled` is true (ie. if `hdds.grpc.tls.enabled=true`)

To make this work in docker-compose cluster, we need a small tweak in the OM certificate request.  Normally hostname is included in the alternative names list, and subject name can be anything (in our case it's in the form of `user@host`, eg. `root@om1`).  However, if reverse lookup for OM host's IP does not work, then subject name will be the only information available for certificate verification to match DNS.  In this case we should omit the username part, because it will cause DNS match to fail.

https://issues.apache.org/jira/browse/HDDS-4788

## How was this patch tested?

Enabled `hdds.grpc.tls.enabled=true` in `ozonesecure-om-ha` cluster.

https://github.com/adoroszlai/hadoop-ozone/actions/runs/547953101

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificates/utils/CertificateSignRequest.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificates/utils/CertificateSignRequest.java
@@ -190,12 +190,24 @@ public final class CertificateSignRequest {
       return this;
     }
 
-    // Support SAN extenion with DNS and RFC822 Name
+    // Support SAN extension with DNS and RFC822 Name
     // other name type will be added as needed.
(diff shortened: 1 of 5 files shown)


xiaoyuyao (approved):
LGTM, +1 pending CI.

amaliujia:
@bshashikant @xiaoyuyao 

Do we need the same change on SCM Ratis?

bshashikant:
> @bshashikant @xiaoyuyao
> 
> Do we need the same change on SCM Ratis?

yes , we need and we will enable it when we implement security in SCM HA.

adoroszlai:

PR #1913, opened 2021-02-08
HDDS-4182. Onboard HDDS-3869 into Layout version management.

errose28:

In HDDS-3869 (Use different column families for datanode block and metadata), a new schema version for container layout was introduced. Use the non-rolling upgrade framework to make sure that this new schema version is not used before the cluster is finalized.

HDDS-4182

## How was this patch tested?

With the current code, existing tests will use schema V1 after this change. After HDDS-4817, existing tests will use schema V2 as before. Moving from schema V1 to schema V2 will be tested as part of the upgrade acceptance testing added in HDDS-4181.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DatanodeMetadataFeatures.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DatanodeMetadataFeatures.java
@@ -3,19 +3,27 @@ package org.apache.hadoop.ozone.container.upgrade;
 import org.apache.hadoop.hdds.upgrade.HDDSLayoutVersionManager;
 import org.apache.hadoop.ozone.OzoneConsts;
 
+/**
+ * Utility class to retrieve the version of a feature that corresponds to the
+ * metadata layout version specified by the provided
+ * {@link HDDSLayoutVersionManager}.
+ */
 public final class DatanodeMetadataFeatures {
   private static HDDSLayoutVersionManager versionManager;
 
-  public static void setHDDSLayoutVersionManager(HDDSLayoutVersionManager manager) {
+  private DatanodeMetadataFeatures() { }
+
+  public static synchronized void setHDDSLayoutVersionManager(
+      HDDSLayoutVersionManager manager) {
     versionManager = manager;
   }
 
-  public static String getSchemaVersion() {
+  public static synchronized String getSchemaVersion() {
     if (versionManager == null ||
         versionManager.getMetadataLayoutVersion() < 1) {
       return OzoneConsts.SCHEMA_V1;
     } else {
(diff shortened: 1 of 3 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DatanodeMetadataFeatures.java
```diff
@@ -0,0 +1,46 @@
...
+ * Utility class to retrieve the version of a feature that corresponds to the
+ * metadata layout version specified by the provided
+ * {@link HDDSLayoutVersionManager}.
+ */
+public final class DatanodeMetadataFeatures {
+  private static HDDSLayoutVersionManager versionManager;
+
+  private DatanodeMetadataFeatures() { }
+
+  public static synchronized void setHDDSLayoutVersionManager(
```
avijayanhwx, line 33: nit. Rename to initialize()?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DatanodeMetadataFeatures.java
```diff
@@ -0,0 +1,46 @@
...
+      HDDSLayoutVersionManager manager) {
+    versionManager = manager;
+  }
+
+  public static synchronized String getSchemaVersion() {
+    if (versionManager == null ||
+        versionManager.getMetadataLayoutVersion() < 1) {
```
avijayanhwx, line 40: Can we use the actual LayoutFeature.layoutVersion here?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/upgrade/DatanodeMetadataFeatures.java
```diff
@@ -0,0 +1,46 @@
...
+
+  private DatanodeMetadataFeatures() { }
+
+  public static synchronized void setHDDSLayoutVersionManager(
+      HDDSLayoutVersionManager manager) {
+    versionManager = manager;
+  }
+
+  public static synchronized String getSchemaVersion() {
+    if (versionManager == null ||
```
avijayanhwx, line 39: If versionManager is null, we should throw an exception (maybe IOException).

avijayanhwx (commented):
Thanks for working on this @errose28. A few comments inline.

errose28:
Thanks for the review @avijayanhwx. I have updated the code with your suggestions.

avijayanhwx (approved):

avijayanhwx:
CI failure '_org.apache.hadoop.ozone.om.TestOzoneManagerPrepare_' is unrelated to this patch, and is being fixed in HDDS-4610 and RATIS-1305.

PR #1916, opened 2021-02-09
HDDS-4812. Move Ratis group creation to scm --init phase

bshashikant:

Move ratis initilaization to scm --init phase

https://issues.apache.org/jira/browse/HDDS-4812

## How was this patch tested?

Added test.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
@@ -38,7 +38,7 @@ public final class OzoneConsts {
 
   public static final String STORAGE_DIR = "scm";
   public static final String SCM_ID = "scmUuid";
-  public static final String SCM_NODES = "scmNodes";
+  public static final String SCM_NODES_INFO = "scmNodesInfo";
   public static final String SCM_NODES_SEPARATOR = "-";
 
   public static final String OZONE_SIMPLE_ROOT_USER = "root";
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
@@ -38,6 +38,8 @@ public final class OzoneConsts {
 
   public static final String STORAGE_DIR = "scm";
   public static final String SCM_ID = "scmUuid";
+  public static final String SCM_NODES = "scmNodes";
+  public static final String SCM_NODES_SEPARATOR = "-";
 
   public static final String OZONE_SIMPLE_ROOT_USER = "root";
   public static final String OZONE_SIMPLE_HDFS_USER = "hdfs";
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
@@ -74,15 +74,19 @@ public final class RatisUtil {
   public static void setRaftStorageDir(final RaftProperties properties,
                                        final SCMHAConfiguration haConf,
                                        final ConfigurationSource conf) {
+    RaftServerConfigKeys.setStorageDir(properties,
+        Collections.singletonList(new File(getRatisStorageDir(haConf, conf))));
+  }
+
+  public static String getRatisStorageDir(final SCMHAConfiguration haConf,
+      final ConfigurationSource conf) {
     String storageDir = haConf.getRatisStorageDir();
     if (Strings.isNullOrEmpty(storageDir)) {
       File metaDirPath = ServerUtils.getOzoneMetaDirPath(conf);
       storageDir = (new File(metaDirPath, "scm-ha")).getPath();
     }
-    RaftServerConfigKeys.setStorageDir(properties,
-        Collections.singletonList(new File(storageDir)));
+    return storageDir;
   }
-
   /**
    * Set properties related to Raft RPC.
    *
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHANodeDetails.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHANodeDetails.java
@@ -124,11 +124,13 @@ public class SCMHANodeDetails {
     for (String serviceId : scmServiceIds) {
       Collection<String> scmNodeIds = ScmUtils.getSCMNodeIds(conf, serviceId);
 
+      // TODO: need to fall back to ozone.scm.names in case scm node ids are
+      // not defined.
       if (scmNodeIds.size() == 0) {
         throw new IllegalArgumentException(
             String.format("Configuration does not have any value set for %s " +
-                "for the service %s. List of OM Node ID's should be " +
-                "specified for an OM service",
+                "for the service %s. List of SCM Node ID's should be " +
+                "specified for an SCM service",
                 ScmConfigKeys.OZONE_SCM_NODES_KEY, serviceId));
       }
       // TODO: load Ratis peers configuration
@@ -166,7 +168,7 @@ public class SCMHANodeDetails {
         }
 
         if (addr.isUnresolved()) {
-          LOG.error("Address for OM {} : {} couldn't be resolved. Proceeding " +
+          LOG.error("Address for SCM {} : {} couldn't be resolved. Proceeding " +
                   "with unresolved host to create Ratis ring.", nodeId,
               rpcAddrStr);
         }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHANodeDetails.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHANodeDetails.java
@@ -168,8 +168,8 @@ public class SCMHANodeDetails {
         }
 
         if (addr.isUnresolved()) {
-          LOG.error("Address for SCM {} : {} couldn't be resolved. Proceeding " +
-                  "with unresolved host to create Ratis ring.", nodeId,
+          LOG.error("Address for SCM {} : {} couldn't be resolved. Proceeding "
+                  + "with unresolved host to create Ratis ring.", nodeId,
               rpcAddrStr);
         }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
@@ -17,6 +17,7 @@
 
 package org.apache.hadoop.hdds.scm.ha;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
@@ -26,13 +27,18 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
+import java.util.Optional;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.apache.hadoop.hdds.HddsUtils;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.protocol.proto.SCMRatisProtocol.RequestType;
 import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.server.StorageContainerManager;
@@ -46,7 +52,7 @@ import org.apache.ratis.protocol.RaftPeer;
 import org.apache.ratis.protocol.RaftPeerId;
 import org.apache.ratis.protocol.exceptions.NotLeaderException;
 import org.apache.ratis.server.RaftServer;
-import org.apache.ratis.server.storage.RaftStorageImpl;
+import org.apache.ratis.server.RaftServerConfigKeys;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -70,28 +76,27 @@ public class SCMRatisServerImpl implements SCMRatisServer {
       final DBTransactionBuffer buffer) throws IOException {
     this.scm = scm;
     this.address = haConf.getRatisBindAddress();
-
-    RaftServer server =
-        newRaftServer(scm.getClusterId(), scm.getSCMHANodeDetails(), haConf,
-            conf).setStateMachine(new SCMStateMachine(scm, this, buffer))
-            .build();
+    RaftServer server = newRaftServer(scm.getClusterId(), scm.getScmId(),
+        scm.getSCMHANodeDetails(), conf)
+        .setStateMachine(new SCMStateMachine(scm, this, buffer)).build();
 
     this.division =
         server.getDivision(server.getGroups().iterator().next().getGroupId());
   }
 
   public static RaftServer.Builder newRaftServer(final String clusterId,
-      final SCMHANodeDetails haDetails, final SCMHAConfiguration haConf,
+      final String scmId, final SCMHANodeDetails haDetails,
       final ConfigurationSource conf) throws IOException {
     final String scmNodeId = haDetails.getLocalNodeDetails().getNodeId();
+    final SCMHAConfiguration haConf = conf.getObject(SCMHAConfiguration.class);
     SCMHAGroupBuilder haGrpBuilder = scmNodeId != null ?
-        new SCMHAGroupBuilder(haDetails, clusterId) :
+        new SCMHAGroupBuilder(haDetails, clusterId, scmId) :
         new SCMHAGroupBuilder(haConf, conf, clusterId);
     final RaftProperties serverProperties =
         RatisUtil.newRaftProperties(haConf, conf);
     return RaftServer.newBuilder().setServerId(haGrpBuilder.getPeerId())
-            .setGroup(haGrpBuilder.getRaftGroup())
-            .setProperties(serverProperties).setStateMachine(null);
+        .setGroup(haGrpBuilder.getRaftGroup()).setProperties(serverProperties)
+        .setStateMachine(new SCMStateMachine());
   }
 
   @Override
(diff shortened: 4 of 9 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -608,6 +614,11 @@ private void loginAsSCMUser(ConfigurationSource conf)
   }
 
 
+  // TODO: Ratis is always enabled now. In case, ratis gets disabled, this
+  // switch should be adjusted accordingly.
+  public static boolean isRatisEnabled(OzoneConfiguration conf) {
```
bharatviswa504, line 619: There is a utility method SCMHAUtils#isSCMHAEnabled we can use this and change the default value of OZONE_SCM_HA_ENABLE_DEFAULT to true, as now ratis is default looks like. (I see there is no usage for this method)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -699,6 +711,12 @@ public static boolean scmInit(OzoneConfiguration conf,
             + ";cid={};layoutVersion={}", scmStorageConfig.getStorageDir(),
             scmStorageConfig.getClusterID(),
             scmStorageConfig.getLayoutVersion());
+        //TODO: this check need not be done if ratis is not used.
+        // Initialize the raft group Id withe cluster Id.
+        if (isRatisEnabled(conf)) {
+          SCMRatisServerImpl.newRaftServer(clusterId, haDetails,
```
bharatviswa504, line 717: Question do we need to initialize RaftServer during this step, as anyway during SCM start, we create RaftServer using SCM ClusterID.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -227,12 +240,62 @@ public RaftPeerId getPeerId() {
...
+      List<RaftPeer> raftPeers = new ArrayList<>();
+      // Add this Ratis server to the Ratis ring
+      raftPeers.add(localRaftPeer);
+
+      for (SCMNodeDetails peerInfo : details.getPeerNodeDetails()) {
+        String peerNodeId = peerInfo.getNodeId();
+        RaftPeerId raftPeerId = RaftPeerId.valueOf(peerNodeId);
+        RaftPeer raftPeer;
+        if (peerInfo.isHostUnresolved()) {
+          raftPeer = RaftPeer.newBuilder().setId(raftPeerId)
```
bharatviswa504, line 278: Question: When the host is unresolved, we still construct raftPeer, do we fail here if it might be due to incorrect hostname. I see OM also followed similar approach, just trying to understand reason.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -227,12 +240,62 @@ public RaftPeerId getPeerId() {
               "localHost: {}, OZONE_SCM_NAMES: {}, selfPeerId: {}",
           localHost, conf.get(ScmConfigKeys.OZONE_SCM_NAMES), selfPeerId);
 
-      raftGroupId = RaftGroupId.valueOf(UUID.nameUUIDFromBytes(
-          SCM_SERVICE_ID.getBytes(StandardCharsets.UTF_8)));
+      String groupId = clusterId == null ? SCM_SERVICE_ID : clusterId;
```
bharatviswa504, line 243: Question: In which scenario clusterID will be null, both in HA/non-HA clusterID would be there and it will be generated during init step right?

bharatviswa504 (commented):
Thank You @bshashikant for the patch. I have a few comments/questions.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
bshashikant: Thanks @bharatviswa504 . The idea here is to check HA enablement but, whether ratis is being used even for non-HA mode. Its a placeholder code for now and the logic to invoke code with/without ratis for non-HA setup needs to be worked upon yet.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
bshashikant: It was added to address some unit test failures.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
bshashikant: Yes, we need to do iniilization.

bharatviswa504: This part of initialization I have not understood why we need during init, as anyway during SCM start we use clusterID and start RaftServer. Can you share more info, why we need this during init step.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
bharatviswa504: If that is the use case isSCMHAEnabled we can use it right, as that is the config to let know whether ratis is used in SCM or not.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
bshashikant: The idea here, is to just create and initialize the ratis group in init phase. When a ratis server starts up, the group should already exists. Thats way , we differentiate an scm node was initialized and this will immediately start a single node server and will become leader.

The other nodes won't be initialised and after starting up, if figure out they don't belong to any group, first thing to do there is contact the leader by going through each SCM round-robin , initiate an alredy exiting leader to execute setConf to add , it to the ratis ring.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -69,19 +71,27 @@
...
+      final ConfigurationSource conf) throws IOException {
+    final String scmNodeId = haDetails.getLocalNodeDetails().getNodeId();
+    SCMHAGroupBuilder haGrpBuilder = scmNodeId != null ?
+        new SCMHAGroupBuilder(haDetails, clusterId) :
+        new SCMHAGroupBuilder(haConf, conf, clusterId);
+    final RaftProperties serverProperties =
+        RatisUtil.newRaftProperties(haConf, conf);
+    return RaftServer.newBuilder().setServerId(haGrpBuilder.getPeerId())
+            .setGroup(haGrpBuilder.getRaftGroup())
+            .setProperties(serverProperties).setStateMachine(null);
```
amaliujia, line 94: Why not pass StateMachine as a parameter to `newRaftServer`?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -189,8 +199,11 @@ public RaftPeerId getPeerId() {
       return selfPeerId;
     }
 
+    /**
+     * This will be used when the SCM HA config are not defined.
+     */
     SCMHAGroupBuilder(final SCMHAConfiguration haConf,
-                      final ConfigurationSource conf) throws IOException {
+        final ConfigurationSource conf, String clusterId) throws IOException {
```
amaliujia, line 206: Can you clarify why use `clusterId` as RaftGroup id?  Is it useful somehow later?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -227,12 +240,62 @@ public RaftPeerId getPeerId() {
...
 
+    SCMHAGroupBuilder(final SCMHANodeDetails details, final String clusterId) {
+
+      Preconditions.checkNotNull(clusterId);
+      // RaftGroupId is the clusterId
+      raftGroupId = RaftGroupId.valueOf(buildRaftGroupId(clusterId));
+      final String scmNodeId = details.getLocalNodeDetails().getNodeId();
+      Preconditions.checkNotNull(scmNodeId);
+      selfPeerId = RaftPeerId.getRaftPeerId(scmNodeId);
+      Preconditions.checkNotNull(scmNodeId);
```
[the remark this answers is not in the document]
bshashikant: Addressed

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
bshashikant: Addressed this in the patch.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
bshashikant: The idea is to use any id defined by user config like service id. ClusterId will be generated by only one SCM in the cluster and propagated to whole system.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
bshashikant: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
bshashikant: Its following OM model.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHANodeDetails.java
```diff
@@ -166,7 +168,7 @@ public static SCMHANodeDetails loadSCMHAConfig(OzoneConfiguration conf)
         }
 
         if (addr.isUnresolved()) {
-          LOG.error("Address for OM {} : {} couldn't be resolved. Proceeding " +
+          LOG.error("Address for SCM {} : {} couldn't be resolved. Proceeding " +
```
amaliujia, line 171: O nice catch :)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/RatisUtil.java
```diff
@@ -74,15 +74,19 @@ public static RaftProperties newRaftProperties(
...
     String storageDir = haConf.getRatisStorageDir();
     if (Strings.isNullOrEmpty(storageDir)) {
       File metaDirPath = ServerUtils.getOzoneMetaDirPath(conf);
       storageDir = (new File(metaDirPath, "scm-ha")).getPath();
     }
-    RaftServerConfigKeys.setStorageDir(properties,
-        Collections.singletonList(new File(storageDir)));
+    return storageDir;
   }
-
```
amaliujia, line 85: nit: undo this line

amaliujia:
overall LGTM.

mukul1987 (approved):

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -713,9 +704,12 @@ public static boolean scmInit(OzoneConfiguration conf,
             scmStorageConfig.getLayoutVersion());
         //TODO: this check need not be done if ratis is not used.
```
bharatviswa504, line 705: Minor: confusing, can we remove the TODO?
As we start SCM Ratis Server only when ratis is enabled, that checks makes it clear.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -80,13 +80,18 @@ public SCMStateMachine(final StorageContainerManager scm,
     }
   }
 
+  public SCMStateMachine() {
```
bharatviswa504, line 83: Instead of doing this, can we create StateMachinee Object like normal creation flow, so that we don' need to add any null checks, as all objects will be null.

I see from code to do this, we need to create metadata store and SCM, do you think we need to do this way, to handle any NPE, if any other method is called.

bshashikant: I think its ok to not create instantiate any SCM sever instance. If any other method gets called suring ratis server spin up, we need to handle that in SCM anyways..

bharatviswa504: Fine with me, as now no usage of ratis server instances we should be fine for now.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -697,32 +698,36 @@ public static boolean scmInit(OzoneConfiguration conf,
...
+          // ensure the ratis group exists
           server.start();
           server.close();
+          scmStorageConfig
+              .setScmNodeInfo(haDetails.getLocalNodeDetails().getHostName());
         }
+        scmStorageConfig.initialize();
+        LOG.info("SCM initialization succeeded. Current cluster id for sd={}"
+                + ";cid={};layoutVersion={}", scmStorageConfig.getStorageDir(),
+            scmStorageConfig.getClusterID(),
```
bharatviswa504, line 716: Minor: Can we also add scmId also?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -697,32 +698,36 @@ public static boolean scmInit(OzoneConfiguration conf,
...
+        // TODO: set the scm node info in the version file during upgrade
+        //  from non-HA SCM to SCM HA set up.
         if (SCMHAUtils.isSCMHAEnabled(conf)) {
           RaftServer server = SCMRatisServerImpl
               .newRaftServer(clusterId, scmStorageConfig.getScmId(), haDetails,
                   conf).build();
+          // ensure the ratis group exists
           server.start();
           server.close();
+          scmStorageConfig
```
bharatviswa504, line 710: Just a question can we also set scmNodeInfo in non-HA also?
So that all fresh installs of non-HA also will have this info.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMStorageConfig.java
```diff
@@ -76,6 +98,11 @@ protected Properties getNodeProperties() {
     }
     Properties scmProperties = new Properties();
     scmProperties.setProperty(SCM_ID, scmId);
+    String scmNodeInfo = getScmNodeInfo();
+    if (scmNodeInfo != null) {
+      // FOR NON-HA setup, SCM_NODES can be null
```
bharatviswa504, line 103: If we add nodeInfo for non-HA, this can be avoided.

bharatviswa504 (commented):
@bshashikant thanks for the update.
Overall LGTM, few minor comments/questions.

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConsts.java
```diff
@@ -38,6 +38,8 @@
 
   public static final String STORAGE_DIR = "scm";
   public static final String SCM_ID = "scmUuid";
+  public static final String SCM_NODES = "scmNodes";
```
bharatviswa504, line 41: Minor: Can we name this as scmNodesInfo

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMStorageConfig.java
bshashikant: I guess, we need to handle upgrades if we use this for non-HA cluster. Not sure we should this right away.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
bshashikant: Upgrade scenarios will need to be handled in case we want to add it for non-HA clsuter.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMStorageConfig.java
bharatviswa504: Fine with me, If we want to revisit this later.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
bharatviswa504: Fine with me, If we want to revisit this later.

bshashikant: Handling upgrades will be an issue as usual. Will revisit this later.

bharatviswa504 (approved):
+1 LGTM.

bshashikant:
Thanks @mukul1987 and @bharatviswa504 for the reviews.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -208,12 +267,12 @@ public RaftPeerId getPeerId() {
         String host = hosts.get(i);
         if (InetAddress.getByName(host).equals(localHost)) {
           selfPeerId = peerId;
+          raftPeers.add(RaftPeer.newBuilder()
+              .setId(peerId)
+              .setAddress(host + ":" + port)
+              .build());
+          break;
```
GlenGeng, line 274: why need this change ? `raftPeers` should contain all the SCMs in `ozone.scm.names`. With it, we can not build a 3 SCM cluster any more.

PR #1917, opened 2021-02-09
HDDS-4815. Close the corresponding ChunkInputStream when BlockInputStream is being closed.

jojochuang:
Close the corresponding ChunkInputStream when BlockInputStream is being closed, reverting the behavior changed by HDDS-4320.

https://issues.apache.org/jira/browse/HDDS-4815

## How was this patch tested?
Manually verified using SparkSQL TPC-DS workload. Also verified by a btrace script that intercepts input stream construction/close calls and socket creation.


adoroszlai (approved):
Thanks @jojochuang for identifying the problem and providing the fix.

PR #1920, opened 2021-02-15
HDDS-4824. Remove repeated value set of reports in StateContext.addReport

symious:

- Use for loop only for incremental reports

https://issues.apache.org/jira/browse/HDDS-4824

## How was this patch tested?

The original test can validate the correctness already. This update is mainly for code improvement.


elek (approved):
+1. Nice catch. Looks reasonable.

Thanks the patch @symious

PR #1921, opened 2021-02-15
HDDS-4822. Implement scm --bootstrap command.

bshashikant:
Implemented SCM --bootstrap command.

During SCM --bootstrap, the bootstrapping SCM node will connect to primary SCM node (already running) and get the cluster Id. Once security is implemented, it will also fetch the CSR root certificates from primary SCM during SCM --bootstrap phase. 

https://issues.apache.org/jira/browse/HDDS-4822

## How was this patch tested?
Added basic Unit tests. Will add more tests once SCM HA config details are finalized.


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -681,6 +688,60 @@ public static String buildRpcServerStartMessage(String description,
...
+      }
+    } else {
+      try {
+        scmStorageConfig.setClusterId(fetchedId);
+        // It will write down the cluster Id fetched from already
+        // running SCM as well as the local SCM Id.
+
+        // SCM Node info containing hostname to scm Id mappings
+        // will be persisted into the version file once this node gets added
+        // to existing SCM ring post node regular start up.
```
bharatviswa504, line 735: Question: Now we have scmId and scmNodeInfo with scmId and host mapping.
Do we plan to remove scmId for the newer version or how it will be handled for upgrade?

bharatviswa504: SCM ID exchange and persist is missing?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManagerStarter.java
```diff
@@ -112,6 +112,25 @@ public void initScm(@CommandLine.Option(names = { "--clusterid" },
     }
   }
 
+  /**
+   * This function implements a sub-command to allow the SCM to be
+   * initialized from the command line.
+   */
+  @CommandLine.Command(name = "--bootstrap",
+      customSynopsis = "ozone scm [global options] --bootstrap",
```
bharatviswa504, line 120: Question: [global options] what are these global options?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -681,6 +688,60 @@ public static String buildRpcServerStartMessage(String description,
...
+  public static boolean scmBootstrap(OzoneConfiguration conf)
+      throws IOException {
+    if (!SCMHAUtils.isSCMHAEnabled(conf)) {
+      LOG.error("Bootstrap is not supported without SCM HA.");
+      return false;
+    }
+    // The node here will try to fetch the cluster id from any of existing
+    // running SCM instances.
+    // TODO: need to avoid failover to local SCM Node here
+    final ScmInfo scmInfo = HAUtils.getScmInfo(conf);
```
bharatviswa504, line 712: If we remove the localSCMNodeId and construct failOverProxy provider we should be good here right?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -681,6 +688,60 @@ public static String buildRpcServerStartMessage(String description,
...
+    final String persistedClusterId = scmStorageConfig.getClusterID();
+    final String fetchedId = scmInfo.getClusterId();
+    Preconditions.checkNotNull(fetchedId);
+    StorageState state = scmStorageConfig.getState();
+    if (state != StorageState.INITIALIZED) {
+      Preconditions.checkNotNull(scmStorageConfig.getScmId());
+      if (fetchedId != persistedClusterId) {
+        LOG.error(
+            "Could not bootstrap as SCM is already initialized with cluster "
+                + "id {} but cluster id for existing running SCM instance "
```
bharatviswa504, line 723: Can we reword this
but clusterId of SCM is not matching with leader SCM clusterId. 

(Generally, this might be due to change in config ?)

bharatviswa504 (commented):
Over all approach LGTM.
Few comments/questions.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
bshashikant: SCMID will be exchanged during start up. Its not required during Bootstrap.

bshashikant: > Question: Now we have scmId and scmNodeInfo with scmId and host mapping.
> Do we plan to remove scmId for the newer version or how it will be handled for upgrade?
> Will these be handled in new Jira?

We will remove it , while upgrading It needs to write down the SCMID:NODE Mappings for existing clusters.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManagerStarter.java
bshashikant: [global options]  are hadoop generic options . I guess, maintained for legacy same as in scm --init.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
bshashikant: Yes, it will work. I did not find anyway to do that in failOverProxy provider.

bharatviswa504 (approved):
+1 LGTM, pending CI issues, need to be fixed.

bshashikant:
Thanks @bharatviswa504 for thr review.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -681,6 +688,60 @@ public static String buildRpcServerStartMessage(String description,
...
+    // The node here will try to fetch the cluster id from any of existing
+    // running SCM instances.
+    // TODO: need to avoid failover to local SCM Node here
+    final ScmInfo scmInfo = HAUtils.getScmInfo(conf);
+    SCMStorageConfig scmStorageConfig = new SCMStorageConfig(conf);
+    final String persistedClusterId = scmStorageConfig.getClusterID();
+    final String fetchedId = scmInfo.getClusterId();
+    Preconditions.checkNotNull(fetchedId);
+    StorageState state = scmStorageConfig.getState();
+    if (state != StorageState.INITIALIZED) {
```
GlenGeng, line 718: `if (state == StorageState.INITIALIZED) {` ?

bharatviswa504: @GlenGeng 
Shashii pointed out this during offline discussion and it is being addressed in HDDS-4718
https://github.com/apache/ozone/pull/1953/files#diff-90e25a3409b1f5c4234a04a51db63e702b217aebe51e15e08525c76b2c6137d6R720

GlenGeng: Thanks @bharatviswa504 for the clarification.

PR #1922, opened 2021-02-15
HDDS-4825. Trash checkpoints are not getting deleted from multiple buckets.

sadanand48:
Trash checkpoints are not getting deleted from multiple buckets . Since the trash emptier is multi-threaded and share the same fs instance,` ofsPathPrefix` which is used as an instance variable in `TrashOzoneFilesystem` is getting set my  multiple threads at once (each thread operates on a bucket) and wrong bucket name is obtained . This change resolves this issue by removing ` ofsPathPrefix` as an instance variable for TrashOzoneFilesystem which is unneccessary.

https://issues.apache.org/jira/browse/HDDS-4825

## How was this patch tested?
Added unit tests where deletion from multiple buckets is ensured.


mukul1987 (changes requested):
Lets remove the variable  ofsPathPrefix as a member variable

mukul1987 (approved):

PR #1923, opened 2021-02-16
HDDS-4813. [FSO]S3Multipart: Implement UploadCompleteRequest

rakeshadr:

This task is to implement prefix based FSO for the S3MultipartUploadCompleteRequest. 

https://issues.apache.org/jira/browse/HDDS-4813

## How was this patch tested?

Added UT cases

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
@@ -354,8 +354,7 @@ public class TestOzoneClientMultipartUploadV1 {
 
     uploadPart(bucket, keyName, uploadID, 1, "data".getBytes(UTF_8));
 
-    // We have not uploaded any parts, but passing some list it should throw
-    // error.
+    // passing with an incorrect part name, should throw INVALID_PART error.
     TreeMap<Integer, String> partsMap = new TreeMap<>();
     partsMap.put(1, UUID.randomUUID().toString());
 
@@ -379,8 +378,7 @@ public class TestOzoneClientMultipartUploadV1 {
 
     uploadPart(bucket, keyName, uploadID, 1, "data".getBytes(UTF_8));
 
-    // We have not uploaded any parts, but passing some list it should throw
-    // error.
+    // passing with an incorrect part number, should throw INVALID_PART error.
     TreeMap<Integer, String> partsMap = new TreeMap<>();
     partsMap.put(3, "random");
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
@@ -310,7 +310,7 @@ public class TestOzoneClientMultipartUploadV1 {
 
     // Complete multipart upload
     OzoneTestUtils.expectOmException(OMException.ResultCodes.ENTITY_TOO_SMALL,
-            () -> completeMultipartUpload(bucket, keyName, uploadID, partsMap));
+        () -> completeMultipartUpload(bucket, keyName, uploadID, partsMap));
   }
 
   @Test
@@ -334,7 +334,7 @@ public class TestOzoneClientMultipartUploadV1 {
     partsMap.put(1, UUID.randomUUID().toString());
 
     OzoneTestUtils.expectOmException(OMException.ResultCodes.INVALID_PART,
-            () -> completeMultipartUpload(bucket, keyName, uploadID, partsMap));
+        () -> completeMultipartUpload(bucket, keyName, uploadID, partsMap));
   }
 
   @Test
@@ -360,7 +360,7 @@ public class TestOzoneClientMultipartUploadV1 {
     partsMap.put(1, UUID.randomUUID().toString());
 
     OzoneTestUtils.expectOmException(OMException.ResultCodes.INVALID_PART,
-            () -> completeMultipartUpload(bucket, keyName, uploadID, partsMap));
+        () -> completeMultipartUpload(bucket, keyName, uploadID, partsMap));
   }
 
   @Test
@@ -385,7 +385,7 @@ public class TestOzoneClientMultipartUploadV1 {
     partsMap.put(3, "random");
 
     OzoneTestUtils.expectOmException(OMException.ResultCodes.INVALID_PART,
-            () -> completeMultipartUpload(bucket, keyName, uploadID, partsMap));
+        () -> completeMultipartUpload(bucket, keyName, uploadID, partsMap));
   }
 
   @Test
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
@@ -17,24 +17,29 @@
 
 package org.apache.hadoop.ozone.client.rpc;
 
+import org.apache.commons.lang3.RandomUtils;
+import org.apache.hadoop.hdds.client.ReplicationFactor;
 import org.apache.hadoop.hdds.client.ReplicationType;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.ozone.MiniOzoneCluster;
+import org.apache.hadoop.ozone.OzoneTestUtils;
 import org.apache.hadoop.ozone.client.ObjectStore;
 import org.apache.hadoop.ozone.client.OzoneBucket;
 import org.apache.hadoop.ozone.client.OzoneClient;
 import org.apache.hadoop.ozone.client.OzoneClientFactory;
 import org.apache.hadoop.ozone.client.OzoneVolume;
+import org.apache.hadoop.ozone.client.io.OzoneInputStream;
 import org.apache.hadoop.ozone.client.io.OzoneOutputStream;
 import org.apache.hadoop.ozone.om.OMConfigKeys;
+import org.apache.hadoop.ozone.om.exceptions.OMException;
 import org.apache.hadoop.ozone.om.helpers.OmMultipartCommitUploadPartInfo;
 import org.apache.hadoop.ozone.om.helpers.OmMultipartInfo;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.apache.hadoop.hdds.StringUtils.string2Bytes;
 import static org.apache.hadoop.hdds.client.ReplicationFactor.THREE;
-import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertNotNull;
 
+import org.apache.hadoop.ozone.om.helpers.OmMultipartUploadCompleteInfo;
 import org.apache.hadoop.ozone.om.request.TestOMRequestUtils;
 import org.junit.AfterClass;
 import org.junit.Assert;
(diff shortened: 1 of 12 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
```diff
@@ -264,12 +274,232 @@ public void testUploadPartOverrideWithRatis() throws IOException {
...
+
+    String uploadID = initiateMultipartUpload(bucket, keyName, STAND_ALONE,
+            ONE);
+
+    uploadPart(bucket, keyName, uploadID, 1, "data".getBytes(UTF_8));
+
+    // We have not uploaded any parts, but passing some list it should throw
+    // error.
+    TreeMap<Integer, String> partsMap = new TreeMap<>();
+    partsMap.put(1, UUID.randomUUID().toString());
```
linyiqun, line 360: The comment should be updated, this case is passing with an incorrect part name.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
```diff
@@ -264,12 +274,232 @@ public void testUploadPartOverrideWithRatis() throws IOException {
...
+
+    String uploadID = initiateMultipartUpload(bucket, keyName, STAND_ALONE,
+            ONE);
+
+    uploadPart(bucket, keyName, uploadID, 1, "data".getBytes(UTF_8));
+
+    // We have not uploaded any parts, but passing some list it should throw
+    // error.
+    TreeMap<Integer, String> partsMap = new TreeMap<>();
+    partsMap.put(3, "random");
```
linyiqun, line 385: The comment should be updated, this case is passing with an incorrect part number.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
```diff
@@ -264,12 +274,232 @@ public void testUploadPartOverrideWithRatis() throws IOException {
...
+    Assert.assertNotNull(commitUploadPartInfo.getPartName());
 
     // PartName should be different from old part Name.
-    assertNotEquals("Part names should be different", partName,
+    Assert.assertNotEquals("Part names should be different", partName,
             commitUploadPartInfo.getPartName());
   }
 
+  @Test
+  public void testMultipartUploadWithPartsLessThanMinSize() throws Exception {
```
linyiqun, line 286: Can we additionally add one another case for OMException.ResultCodes.INVALID_PART_ORDER? I didn't this case was covered in this unit test.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCompleteRequest.java
```diff
@@ -351,8 +267,151 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+              Collections.singletonList(keyLocationInfoGroup))
+          .setAcls(OzoneAclUtil.fromProtobuf(keyArgs.getAclsList()));
+      // Check if db entry has ObjectID. This check is required because
+      // it is possible that between multipart key uploads and complete,
+      // we had an upgrade.
+      if (dbOpenKeyInfo.getObjectID() != 0) {
+        builder.setObjectID(dbOpenKeyInfo.getObjectID());
+      }
+      builder.setParentObjectID(dbOpenKeyInfo.getParentObjectID());
+      builder.setFileName(dbOpenKeyInfo.getFileName());
```
linyiqun, line 313: Can we add isBucketFSOptimized() check when we setting the parentObjectID/FileName? These two field are only used when BucketFSO is enabled.

linyiqun (commented):
@rakeshadr , some minor comments below. Please have a check.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
rakeshadr: Done!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
rakeshadr: Done!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneClientMultipartUploadV1.java
rakeshadr: Sure. As `bucket.completeMultipartUpload(keyName, uploadID, partsMap); ` used hashmap to store parts, which will always be in sorted order. Inorder to make part list with incorrect order, I have added UT `testInvalidPartOrderError` in `TestS3MultipartUploadCompleteRequest` class.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCompleteRequest.java
rakeshadr: Done!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCommitPartRequestV1.java
```diff
@@ -147,8 +148,18 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
+       * "fileName + ClientID".
+       *
+       * Contract is that all part names present in a multipart info will
+       * have same key prefix path.
+       *
+       * For example:
+       *        /vol1/buck1/a/b/c/part-1, /vol1/buck1/a/b/c/part-2,
+       *        /vol1/buck1/a/b/c/part-n
+       */
+      dbPartName = fileName + clientID;
```
linyiqun, line 162: Any specific reason for changing the format of part name?  With above format change, we have to get a full part name again in S3MultipartUploadCompleteRequestV1 request.

rakeshadr: In the V0 (existing code), `PartName` is constructed using the `partName = ozoneKey + clientID;`.

Now, in FSOBucket feature code, we store only the `fileName` in the `fileTable` and reconstructed `fullKeyName` while reading back from it. I thought of following the same format pattern the `multipartFileInfoTable` as well. Yes, like you said, we need to reconstruct it during #complete or #abort or #listParts etc.

Following is the PartKeyInfo proto object. There we can see KeyInfo, which inturn has `parentID` and is having similar structure with other V1(FSOBucket feature) tables. I think, it would be helpful if we could follow same pattern everywhere in the new tables for better debugging unless if there is any functional challenges. Does that make sense to you?
```
message PartKeyInfo {
    required string partName = 1;
    required uint32 partNumber = 2;
    required KeyInfo partKeyInfo = 3;
}
```

linyiqun: @rakeshadr , I see current format  in multipartFileInfoTable, openFileTable, FileTable is like below.

   * |  fileTable         | parentId/fileName -> KeyInfo                    |
   * |----------------------------------------------------------------------|
   * |  openFileTable     | parentId/fileName/id -> KeyInfo                 |
   * |----------------------------------------------------------------------|
   * |  multipartFileInfoTable | parentId/fileName/uploadId ->...  


 So the format of part name is already follow above pattern now. So still need to do above part name change here?
```java
String ozoneKey = omMetadataManager.getOzonePathKey(parentID, fileName);
partName = ozoneKey + clientID;
```

rakeshadr: @linyiqun Yes, there will be multiple parts during the upload operations. Finally openFileTable, multiPartFileInfoTable entries will be deleted and then add fileTable entry corresponding to the user given KeyName. I have tried adding internals especially focussing on DB operations with an example for better understanding, can you please go through it and hope that helps to proceed further. Thanks a lot!

KeyName = "a/b/c/file1". 
Assume there are three parts, which will be having part number assigned like, 1, 2 and 3.

**User Operation-1)** S3InitiateMultipartUpload : KeyName="a/b/c/file1"
```
        1) UploadID will be generated by OM : UUID.randomUUID().toString() + "-" + UniqueId.next()
	2) Add "a/b/c" parent dirs into DirectoryTable
	3) Add entry to openFileTable:
			TableKey = multipartKey, which is nothing but <parentId/fileName/uploadId>
			TableValue = omKeyInfo, which will store fileName and parentID.
	4) Add entry to multipartFileInfoTable:		
			TableKey = multipartKey, which is nothing but <parentId/fileName/uploadId>
			TableValue = omKeyInfo, which will store uploadID and parentID. Since there is no real upload started, partKeyInfoList will be empty.
	
	Say, uploadID = 12345-0000. This is the unique identifier for uploading parts for the corresponding keyname.
```


**User Operation-2)** S3CreateMultipartKey : KeyName="a/b/c/file1", PartNumber=1, uploadID="12345-0000"
```
	1) ClientID will be generated by OM : UniqueId.next()
	2) Since it has uploadID, it will check OmKeyInfo in openFileTable.
	2) Add entry to OpenFileTable:
			TableKey = openFileKey, which is nothing but <parentId/fileName/clientID>
			TableValue = omKeyInfo, which will store fileName and parentID.
	
	This returns KeyOutputStream with unique ClientID. Say, ClientID=9777 and user streams data to it.
	Finally during close, will invoke commitMultipartUploadPart.
```

**User Operation-3)** S3CommitMultipartUploadPart: KeyName="a/b/c/file1", PartNumber=1, uploadID="12345-0000", ClientID=9777
```
	1) Get <parentId/fileName/clientID> entry from openFileTable. Say, we got OmKeyInfo1 object from DB.
	2) Create 'PartKeyInfo' using the OmKeyInfo1 object.
			PartNumber = 1
			PartName   = file1 + 9777; // fileName + ClientID;
			keyInfo    = OmKeyInfo1    // here it stores only the fileName and parentID.
	3) Get multipartKey <parentId/fileName/uploadId> entry from multipartFileInfoTable. Say, we got multipartKeyInfo1 object from DB.
	4) Append/Add 'PartKeyInfo' to multipartKeyInfo1 object like,
			multipartKeyInfo.addPartKeyInfo(partNumber, partKeyInfo.build());
	5) Update the above multipartKeyInfo into the multipartFileInfoTable.
	6) Delete <parentId/fileName/clientID> entry from openFileTable.
```
	

**User Operation-4)** 
```
       Repeat S3CreateMultipartKey and S3CommitMultipartUploadPart for PartNumber=2 and PartNumber=3.
        After these three parts upload operations, now the 'multipartKeyInfo' entry in multipartFileInfoTable will have three 'PartKeyInfo' items in the list.
```

**User Operation-5)** S3CompleteMultipartUpload: KeyName="a/b/c/file1", partsMap={1,2,3}, uploadID="12345-0000".
```
	1) Get multipartKey <parentId/fileName/uploadId> entry from multipartFileInfoTable. Say, we got multipartKeyInfo1 object from DB.
	2) Compare all the user given part numbers and data size calculation etc using all the parts. I'm not exploring those info now as it is existing logic.
	3) Prepare OmKeyInfo using <parentId/fileName> and add new entry into FileTable. This is now become the reference to the user given KeyName="a/b/c/file1".
	4) Delete multipartKey <parentId/fileName/uploadId> entry from openFileTable.
	5) Delete multipartKey <parentId/fileName/uploadId> entry from multipartFileInfoTable.
```

linyiqun: Thanks @rakeshadr for very detailed explanation, make the sense to me.

linyiqun (approved):
LGTM, +1.

rakeshadr:
Thanks a lot @linyiqun for the reviews and useful comments. I'm merging it to the branch.

PR #1929, opened 2021-02-17
HDDS-4811.Use OM style config to construct RaftGroup and initialize Raft Servers

amaliujia:

https://issues.apache.org/jira/browse/HDDS-4811

## How was this patch tested?

N/A

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
@@ -313,6 +313,13 @@ public class SCMRatisServerImpl implements SCMRatisServer {
       List<RaftPeer> raftPeers = new ArrayList<>();
       // Add this Ratis server to the Ratis ring
       raftPeers.add(localRaftPeer);
+      for (SCMNodeDetails nodeDetails : details.getPeerNodeDetails()) {
+        raftPeers.add(
+            RaftPeer.newBuilder()
+                .setId(nodeDetails.getNodeId())
+                .build()
+        );
+      }
       // always start as a standalone server
       raftGroup = RaftGroup.valueOf(raftGroupId, raftPeers);
     }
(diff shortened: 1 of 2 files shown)


amaliujia:
R: @mukul1987 @bshashikant

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -313,6 +313,13 @@ public RaftPeerId getPeerId() {
       List<RaftPeer> raftPeers = new ArrayList<>();
       // Add this Ratis server to the Ratis ring
       raftPeers.add(localRaftPeer);
+      for (SCMNodeDetails nodeDetails : details.getPeerNodeDetails()) {
```
bshashikant, line 316: I think , we need to add the peers depending upon whether SCM needs to start as single node server(1st regular start) or participate in 3 node ring(subsequent restarts). Let's hold on to this change for now and i will take care of this in the bootstrap sequence.

bshashikant (commented):
@amaliujia , thanks for the patch. I see a TODO marked in SCMHANodeDetails#174 regarding details missing for Raft peer configuration.

Can you please specify what is missing here and address it in this patch?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
amaliujia: Sure sounds good.

amaliujia:
@bshashikant 

Regrading  the SCMHANodeDetails#174 TODO, I used to think we load Raft config into SCMHANodeDetails directly. however, now the code is refactored as https://github.com/apache/ozone/blob/b60a2306374533a616123d1f9da33e0506bac4d6/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java#L93

So I guess there is no need for the SCMHANodeDetails#174 TODO.

At this moment HDDS-4811 maybe is already just a no-op JIRA

bshashikant:
This change is not required any more as the required changes are done as part of https://github.com/apache/ozone/pull/1953.

Thanks @amaliujia for the efforts.

PR #1931, opened 2021-02-17
HDDS-4834. Replication failure in secure environment

adoroszlai:

`GrpcReplicationClient` currently tries to read DN certificate with the wrong filename and missing path.  This change makes it get the certificate from `DNCertificateClient`, which already has logic to find and load the it.

https://issues.apache.org/jira/browse/HDDS-4834

## How was this patch tested?

Reproduced the bug in `ozonesecure` acceptance test:

```
extradn_1   | 2021-02-17 08:41:00,481 [ContainerReplicationThread-0] INFO replication.DownloadAndImportReplicator: Starting replication of container 1 from [c61bf116-5ceb-4ed7-a94e-9a156433e2fe{ip: 192.168.240.5, host: ozonesecure_datanode_1.ozonesecure_default, ports: [REPLICATION=9886, RATIS=9858, RATIS_ADMIN=9857, RATIS_SERVER=9856, STANDALONE=9859], networkLocation: /default-rack, certSerialId: null, persistedOpState: IN_SERVICE, persistedOpStateExpiryEpochSec: 0}, b9d535d9-3a6e-44e6-b212-333f73c73c38{ip: 192.168.240.10, host: ozonesecure_datanode_2.ozonesecure_default, ports: [REPLICATION=9886, RATIS=9858, RATIS_ADMIN=9857, RATIS_SERVER=9856, STANDALONE=9859], networkLocation: /default-rack, certSerialId: null, persistedOpState: IN_SERVICE, persistedOpStateExpiryEpochSec: 0}]
extradn_1   | 2021-02-17 08:41:00,486 [ContainerReplicationThread-0] ERROR replication.SimpleContainerDownloader: Container 1 download from datanode c61bf116-5ceb-4ed7-a94e-9a156433e2fe{ip: 192.168.240.5, host: ozonesecure_datanode_1.ozonesecure_default, ports: [REPLICATION=9886, RATIS=9858, RATIS_ADMIN=9857, RATIS_SERVER=9856, STANDALONE=9859], networkLocation: /default-rack, certSerialId: null, persistedOpState: IN_SERVICE, persistedOpStateExpiryEpochSec: 0} was unsuccessful. Trying the next datanode
extradn_1   | java.lang.IllegalArgumentException: File does not contain valid certificates: certificate.crt
extradn_1   | 	at org.apache.ratis.thirdparty.io.netty.handler.ssl.SslContextBuilder.keyManager(SslContextBuilder.java:345)
extradn_1   | 	at org.apache.ratis.thirdparty.io.netty.handler.ssl.SslContextBuilder.keyManager(SslContextBuilder.java:294)
extradn_1   | 	at org.apache.hadoop.ozone.container.replication.GrpcReplicationClient.<init>(GrpcReplicationClient.java:81)
extradn_1   | 	at org.apache.hadoop.ozone.container.replication.SimpleContainerDownloader.downloadContainer(SimpleContainerDownloader.java:135)
extradn_1   | 	at org.apache.hadoop.ozone.container.replication.SimpleContainerDownloader.getContainerDataFromReplicas(SimpleContainerDownloader.java:88)
extradn_1   | 	at org.apache.hadoop.ozone.container.replication.DownloadAndImportReplicator.replicate(DownloadAndImportReplicator.java:110)
extradn_1   | 	at org.apache.hadoop.ozone.container.replication.MeasuredReplicator.replicate(MeasuredReplicator.java:69)
extradn_1   | 	at org.apache.hadoop.ozone.container.replication.ReplicationSupervisor$TaskRunner.run(ReplicationSupervisor.java:139)
...
extradn_1   | 2021-02-17 08:41:00,488 [ContainerReplicationThread-0] ERROR replication.ReplicationSupervisor: Container 1 can't be downloaded from any of the datanodes.
```

https://github.com/adoroszlai/hadoop-ozone/runs/1918037139

Verified that replication is successful with the fix:

```
extradn_1   | 2021-02-17 11:53:44,371 [ContainerReplicationThread-0] INFO replication.DownloadAndImportReplicator: Starting replication of container 1 from [60529ea2-54c0-4882-95fa-930afa556d98{ip: 172.18.0.2, host: ozonesecure_datanode_2.ozonesecure_default, ports: [REPLICATION=9886, RATIS=9858, RATIS_ADMIN=9857, RATIS_SERVER=9856, STANDALONE=9859], networkLocation: /default-rack, certSerialId: null, persistedOpState: IN_SERVICE, persistedOpStateExpiryEpochSec: 0}, f26be9a7-5084-438a-b4cb-5f875036030c{ip: 172.18.0.3, host: ozonesecure_datanode_1.ozonesecure_default, ports: [REPLICATION=9886, RATIS=9858, RATIS_ADMIN=9857, RATIS_SERVER=9856, STANDALONE=9859], networkLocation: /default-rack, certSerialId: null, persistedOpState: IN_SERVICE, persistedOpStateExpiryEpochSec: 0}]
extradn_1   | 2021-02-17 11:53:53,815 [grpc-default-executor-2] INFO replication.GrpcReplicationClient: Container 1 is downloaded to /tmp/container-copy/container-1.tar.gz
extradn_1   | 2021-02-17 11:53:53,818 [ContainerReplicationThread-0] INFO replication.DownloadAndImportReplicator: Container 1 is downloaded with size 214343220, starting to import.
extradn_1   | 2021-02-17 11:53:56,222 [ContainerReplicationThread-0] INFO replication.DownloadAndImportReplicator: Container 1 is replicated successfully
extradn_1   | 2021-02-17 11:53:56,222 [ContainerReplicationThread-0] INFO replication.ReplicationSupervisor: Container 1 is replicated.
```

https://github.com/adoroszlai/hadoop-ozone/runs/1918041342

### DIFF
--- a/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
+++ b/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
@@ -29,8 +29,10 @@ import org.apache.hadoop.hdds.scm.container.common.helpers
     .ContainerWithPipeline;
 
 import com.google.common.base.Preconditions;
+import org.apache.hadoop.hdds.server.JsonUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import picocli.CommandLine;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Model.CommandSpec;
 import picocli.CommandLine.Parameters;
@@ -52,6 +54,11 @@ public class InfoSubcommand extends ScmSubcommand {
   @Spec
   private CommandSpec spec;
 
+  @CommandLine.Option(names = { "--json" },
+      defaultValue = "false",
+      description = "Format output as JSON")
+  private boolean json;
+
   @Parameters(description = "Decimal id of the container.")
   private long containerID;
 
@@ -61,22 +68,26 @@ public class InfoSubcommand extends ScmSubcommand {
         getContainerWithPipeline(containerID);
     Preconditions.checkNotNull(container, "Container cannot be null");
 
-    // Print container report info.
-    LOG.info("Container id: {}", containerID);
-    boolean verbose = spec.root().userObject() instanceof GenericParentCommand
-        && ((GenericParentCommand) spec.root().userObject()).isVerbose();
-    if (verbose) {
-      LOG.info("Pipeline Info: {}", container.getPipeline());
+    if (json) {
+      LOG.info(JsonUtils.toJsonStringWithDefaultPrettyPrinter(container));
     } else {
-      LOG.info("Pipeline id: {}", container.getPipeline().getId().getId());
-    }
-    LOG.info("Container State: {}", container.getContainerInfo().getState());
+      // Print container report info.
+      LOG.info("Container id: {}", containerID);
+      boolean verbose = spec.root().userObject() instanceof GenericParentCommand
+          && ((GenericParentCommand) spec.root().userObject()).isVerbose();
+      if (verbose) {
+        LOG.info("Pipeline Info: {}", container.getPipeline());
+      } else {
+        LOG.info("Pipeline id: {}", container.getPipeline().getId().getId());
+      }
+      LOG.info("Container State: {}", container.getContainerInfo().getState());
 
-    // Print pipeline of an existing container.
-    String machinesStr = container.getPipeline().getNodes().stream().map(
-        InfoSubcommand::buildDatanodeDetails)
-        .collect(Collectors.joining(",\n"));
-    LOG.info("Datanodes: [{}]", machinesStr);
+      // Print pipeline of an existing container.
+      String machinesStr = container.getPipeline().getNodes().stream().map(
+          InfoSubcommand::buildDatanodeDetails)
+          .collect(Collectors.joining(",\n"));
+      LOG.info("Datanodes: [{}]", machinesStr);
+    }
   }
 
   private static String buildDatanodeDetails(DatanodeDetails details) {
--- a/hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
+++ b/hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
@@ -48,18 +48,6 @@ services:
     environment:
       KERBEROS_KEYTABS: dn HTTP
       OZONE_OPTS:
-  extradn:
-    image: apache/ozone-runner:${OZONE_RUNNER_VERSION}
-    volumes:
-      - ../..:/opt/hadoop
-    ports:
-      - 9864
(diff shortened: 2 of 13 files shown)


hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
```diff
@@ -48,6 +48,18 @@ services:
     environment:
       KERBEROS_KEYTABS: dn HTTP
       OZONE_OPTS:
+  extradn:
```
elek, line 51: Thanks for the fix @adoroszlai 

One question here: do really need this `extradn`. I have mixed feelings as this docker-compose environment is used not only for testing but as an example environment. It may be confusing for the users who would like to see a simple secure cluster, that we have this `extradn`.

As far as I see, when you do a `docker-compose scale datanode=2` and `docker-compose scale datanode=3`, the original `/data` volume is recreated, so replication could be tested.

Or (as an alternative, but may be a more complex approach) we can improve the `ClosedContainerReplication` freon test to use certificates...

sodonnel: The logic changes look good to me, but I agree with @elek that it would be better to avoid the `extradn` if possible.

One other possible idea - start a cluster with 4 DNs. Create a container and identify one of the DNs which has the container. Then decommission the host and wait for decommission to complete. For decommission to complete, replication must happen, otherwise decommission will hang forever.

sodonnel:
I think these changes look good, but I am not really familiar with this code area, or these robot tests, so it would be good for @elek to check as well.

+1 from me pending CI and @elek checking too.

hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
adoroszlai: Thanks @elek and @sodonnel for the review.  I wasn't happy about adding `extradn`, either.  Thanks for the idea about scaling down then up.  The only issue is that it's harder to state our expectation about where the container should be located if the original and new datanodes have the same name.

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java
```diff
@@ -52,6 +54,11 @@
   @Spec
   private CommandSpec spec;
 
+  @CommandLine.Option(names = { "--json" },
+      defaultValue = "false",
+      description = "Format output as JSON")
+  private boolean json;
+
```
elek, line 61: Let me stop here for a moment and celebrate this flag. I really like it. I think it should be the standard in all the debug output: we should use either json or multi-line one-json per-line everywhere.

It's not just for the robot test but for everyday usage, jq is very powerful.

:heart:  :heart:

elek (approved):
+1 Thanks the update @adoroszlai and the review @sodonnel 

Merging it...

PR #1936, opened 2021-02-19
HDDS-4807. Add install checkpoint in SCMStateMachine.

bshashikant:

Added Install Checkpoint functionality

https://issues.apache.org/jira/browse/HDDS-4807

## How was this patch tested?

Unit tests added.


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
```diff
@@ -248,4 +248,9 @@ public boolean shouldRun() {
   public String getServiceName() {
     return SCMBlockDeletingService.class.getSimpleName();
   }
+
+  @Override
+  public void stop() {
```
GlenGeng, line 253: A shutdown thread pool can not be reused again, since after calling shutdown on a ExecutorService, no new Task will be accepted.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/states/ContainerStateMap.java
```diff
@@ -114,11 +114,7 @@ public void addContainer(final ContainerInfo info)
     Preconditions.checkNotNull(info, "Container Info cannot be null");
     final ContainerID id = info.containerID();
     if (!contains(id)) {
-      containerMap.put(id, info);
-      lifeCycleStateMap.insert(info.getState(), id);
-      ownerMap.insert(info.getOwner(), id);
-      factorMap.insert(info.getReplicationFactor(), id);
-      typeMap.insert(info.getReplicationType(), id);
+      build(info);
```
GlenGeng, line 117: Can we have a better function name ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/states/ContainerStateMap.java
```diff
@@ -128,6 +124,21 @@ public void addContainer(final ContainerInfo info)
     }
   }
 
+  /**
+   * The idea here is to rebuild the containerStateMap once the SCM StateMachine
+   * reloads from the the DBCheckpoint downloaded from leader SCM.
+   *
+   * Replica Map won't be rebuild but will remain as it is.
+   */
+  public void build(final ContainerInfo info) throws SCMException {
```
GlenGeng, line 133: Will it be used at other place than `addContainer` ?

bshashikant: NO, it won't be.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
bshashikant: Yes., but only if the service is stopped.

I think , we don't need to stop the background services while reloading the SCM state during install snapshot, as reload will only happen on follower and on followers, these won't do anything on the db. What do you think?


However, a start and stop method in such services are desired so that, all the registered services can be started and stopped via service manager itself.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogImplV2.java
```diff
@@ -277,6 +277,12 @@ public int getNumOfValidTransactions() throws IOException {
     }
   }
 
+  @Override
+  public void reinitialize(
+      Table<Long, DeletedBlocksTransaction> deletedBlocksTXTable) {
+  this.deletedBlocksTXTable = deletedBlocksTXTable;
```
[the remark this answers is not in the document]
bshashikant: Thanks for pointing out. Will address that.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/SCMBlockDeletingService.java
GlenGeng: > What do you think?
I agree.

> However, a start and stop method in such services are desired so that, all the registered services can be started and stopped via service manager itself.
In order to support safely stop() then start(), these service should be rewritten to avoid use ExecutorService. This can be done as a newbie task in future.

bshashikant:
Filed https://issues.apache.org/jira/browse/RATIS-1326 to address the related Ratis issue.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/SCMPipelineManager.java
```diff
@@ -759,6 +759,11 @@ public boolean getSafeModeStatus() {
     return this.isInSafeMode.get();
   }
 
+  @Override
+  public void reinitialize(Table<PipelineID, Pipeline> store)
+      throws IOException {
+  }
```
GlenGeng, line 765: For safety,
```
throw new RuntimeException("Un-supported operation.");
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -1110,7 +1114,6 @@ public void stop() {
 
     scmSafeModeManager.stop();
   }
-
```
GlenGeng, line 1113: better have the empty line

GlenGeng (commented):
LGTM. Thanks for the job! Just comments of typo.

GlenGeng:
+1. Thanks for the work!

bshashikant:
Thanks @GlenGeng for the review.

PR #1946, opened 2021-02-19
HDDS-4851. More compatibility problem with DatanodeDetails.Port.Name.REPLICATION

adoroszlai:

HDDS-4731 missed some places related to file status where client version needs to be considered for protobuf conversion.  These can be reproduced by `ozone fs` operations:

```
$ ozone fs -mkdir o3fs://bucket1.vol1/dir/
$ ozone fs -ls o3fs://bucket1.vol1/dir/
$ ozone fs -put /etc/passwd o3fs://bucket1.vol1/dir/
-put: No enum constant org.apache.hadoop.hdds.protocol.DatanodeDetails.Port.Name.REPLICATION
$ ozone fs -ls o3fs://bucket1.vol1/dir/
-ls: No enum constant org.apache.hadoop.hdds.protocol.DatanodeDetails.Port.Name.REPLICATION
```

https://issues.apache.org/jira/browse/HDDS-4851

## How was this patch tested?

Added acceptance test cases.  0.5.0 client had to be removed from the test due to another incompatibility:

```
Message missing required fields: getFileStatusResponse.status.status, ...
```

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/xcompat/clients.yaml
+++ b/hadoop-ozone/dist/src/main/compose/xcompat/clients.yaml
@@ -17,15 +17,6 @@
 version: "3.4"
 
 services:
-  old_client_0_5_0:
-    image: apache/ozone:0.5.0
-    env_file:
-      - docker-config
-    volumes:
-      - ../..:/opt/ozone
-    environment:
-      HADOOP_OPTS:
-    command: ["sleep","1000000"]
   old_client_1_0_0:
     image: apache/ozone:1.0.0
     env_file:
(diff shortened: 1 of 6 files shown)


hadoop-ozone/dist/src/main/compose/xcompat/clients.yaml
```diff
@@ -17,15 +17,6 @@
 version: "3.4"
 
 services:
-  old_client_0_5_0:
```
avijayanhwx, line 20: Can you provide some context on why this has been removed?

avijayanhwx (approved):
LGTM +1. One minor qn inline.

swagle:
Thanks for the review @avijayanhwx committing this change.

adoroszlai:
Thanks @avijayanhwx and @swagle for the review.

hadoop-ozone/dist/src/main/compose/xcompat/clients.yaml
adoroszlai: From PR description:

> 0.5.0 client had to be removed from the test due to another incompatibility:
> 
> ```
> Message missing required fields: getFileStatusResponse.status.status, ...
> ```

PR #1950, opened 2021-02-22
HDDS-4138. Improve crc efficiency by using Java.util.zip.CRC when available

sodonnel:

Based on the results in  HDDS-4808 (https://github.com/apache/ozone/pull/1910#issuecomment-775165462) it is clear we should switch the CRC32 implementations used by Ozone, as there are other better performing implementations.

Specifically:

`org.apache.hadoop.ozone.common.PureJavaCrc32ByteBuffer` to `Java.util.zip.CRC32`

`org.apache.hadoop.ozone.common.PureJavaCrc32CByteBuffer` to `Java.util.zip.CRC32C` if Java 9 or greater, otherwise `org.apache.hadoop.util.PureJavaCrc32C`

https://issues.apache.org/jira/browse/HDDS-4138

## How was this patch tested?

Small change to an existing test

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChecksumByteBufferImpl.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChecksumByteBufferImpl.java
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.ozone.common;
 
+import org.apache.hadoop.hdds.JavaUtils;
 import org.apache.hadoop.util.PureJavaCrc32C;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -33,22 +34,8 @@ public class ChecksumByteBufferImpl implements ChecksumByteBuffer {
   private static final Logger LOG =
       LoggerFactory.getLogger(ChecksumByteBufferImpl.class);
 
-  private static volatile boolean useJava9Crc32C = isJavaVersionAtLeast(9);
-
-  // "1.8"->8, "9"->9, "10"->10
-  private static final int JAVA_SPEC_VER = Math.max(8, Integer.parseInt(
-      System.getProperty("java.specification.version").split("\\.")[0]));
-
-  /**
-   * Query to see if major version of Java specification of the system
-   * is equal or greater than the parameter.
-   *
-   * @param version 8, 9, 10 etc.
-   * @return comparison with system property, always true for 8
-   */
-  public static boolean isJavaVersionAtLeast(int version) {
-    return JAVA_SPEC_VER >= version;
-  }
+  private static volatile boolean useJava9Crc32C
+      = JavaUtils.isJavaVersionAtLeast(9);
 
   public static class Java9Crc32CFactory {
     private static final MethodHandle NEW_CRC32C_MH;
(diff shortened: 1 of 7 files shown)


adoroszlai:
@sodonnel `hadoop27` test is failing due to:

```
NoSuchMethodError: org.apache.hadoop.util.Shell.isJavaVersionAtLeast(I)Z
	at org.apache.hadoop.ozone.common.ChecksumByteBufferImpl.<clinit>(ChecksumByteBufferImpl.java:38)
```

sodonnel:
@adoroszlai Thanks for checking it. Is this because the Hadoop27 tests runs with a hadoop-common 2.x which I guess does not have the Shell.isJavaVersionAtLeast method I am using? I will figure out a way around this, perhaps just creating a method to the Java check, rather than reusing what is in Hadoop common.

sodonnel:
For now (to ensure it fixes the test) I lifted the methods out of hadoop-common into the ChecksumByteBufferImpl class. However we should probably move them into a Ozone util class somewhere, so they are more discoverable / reusable.

adoroszlai:
> However we should probably move them into a Ozone util class somewhere, so they are more discoverable / reusable.

Totally agree.  Maybe `JavaUtils` in `hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds`.  (Normally I'd suggest keeping the class name from Hadoop in case we might need to copy more code later.  But the `Shell` class itself does not seem to be very much related to the Java version check, so I think we can do better.)

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/ChecksumByteBufferImpl.java
```diff
@@ -54,6 +79,23 @@
...
   };
 
+  public static ChecksumByteBuffer crc32Impl() {
+    return new ChecksumByteBufferImpl(new CRC32());
+  }
+
+  public static ChecksumByteBuffer crc32Cimpl() {
+    if (useJava9Crc32C) {
+      try {
+        return new ChecksumByteBufferImpl(Java9Crc32CFactory.createChecksum());
```
swagle, line 89: Should we consider a static factory here for creating the ChecksumByteBufferImpl?

sodonnel: I'm not sure what you mean - the crc32Cimpl is like a static factory I think, and the code which needs to get a checksum impl calls either `checksumByteBufferImpl.crc32impl or `checksumByteBufferImpl..crc32cimpl`.

Can you post a code snippet of what you are thinking?

swagle: What I meant was to separate out the Factory class from the interface impl to encapsulate the object creation logic separate from the behavior that the impl provides, this suggestion was based on the fact that crc32Cimpl is like a static factory method but it is co-located with the impl.

swagle (approved):

sodonnel:
@adoroszlai Have you any further comments or am I good to commit this?

adoroszlai (approved):

PR #1951, opened 2021-02-23
HDDS-4853. libexec/entrypoint.sh might copy from wrong path

sky76093016:

I think it may be the previous mistake that caused the copy path error, so I modified the copy path.

https://issues.apache.org/jira/browse/HDDS-4853

## How was this patch tested?

no test.


adoroszlai (approved):
Thanks @sky76093016.  I don't think I have ever encountered any problem due to this mismatch, but the fix looks safe to me.

PR #1953, opened 2021-02-23
HDDS-4718. Bootstrap new SCM node

bshashikant:

Implemented sequence for adding new SCMs to existing cluster.

https://issues.apache.org/jira/browse/HDDS-4718

## How was this patch tested?
Added UT


GlenGeng (commented):
LGTM.

One question is, will there be following task to add scm to DNs SCMConnectionMananger ?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocol/ScmBlockLocationProtocol.java
```diff
@@ -77,6 +78,11 @@
    */
   ScmInfo getScmInfo() throws IOException;
 
+  /**
+   * Request to add SCM instance to HA group.
+   */
+  boolean addSCM(AddSCMRequest request) throws IOException;
```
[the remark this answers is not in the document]
bshashikant: i was thinking about doing this.. But seems, like OM also uses the block client protocol to connect to SCM. That's why reused it. We can discuss over this as we move along and conclude.

There has been discussions to merge the block protocol with container protocol as well.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
```diff
@@ -311,8 +311,8 @@ public static InetSocketAddress getReconAddresses(
   public static InetSocketAddress getSingleSCMAddress(
       ConfigurationSource conf) {
     Collection<InetSocketAddress> singleton = getSCMAddresses(conf);
-    Preconditions.checkArgument(singleton.size() == 1,
-        MULTIPLE_SCM_NOT_YET_SUPPORTED);
+   // Preconditions.checkArgument(singleton.size() == 1,
```
vivekratnavel, line 314: Can we remove this?

bshashikant: Thanks @vivekratnavel . We will remove this once the the proper SCM address validation patch is in . Its a work in progress.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/XceiverServerRatis.java
```diff
@@ -765,12 +765,18 @@ public void addGroup(HddsProtos.PipelineID pipelineId,
         clientId, server.getId(), nextCallId(), group);
 
     RaftClientReply reply;
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("Received create pipeline request for pipeline {}", pipelineID);
```
GlenGeng, line 769: `create pipeline request` seem not accurate

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocol/ScmBlockLocationProtocol.java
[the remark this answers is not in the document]
bharatviswa504: I also believe this is not the correct place for addSCM, as StorageContainerLocation is used for all admin ops, that might be good place. We need to conclude moving this to StorageContainerLocationProtocol before merge.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -179,25 +193,79 @@ public long takeSnapshot() throws IOException {
...
+      } else {
+        lastAppliedIndex =
+            transactionBuffer.getLatestTrxInfo().getTransactionIndex();
+      }
+
+      transactionBuffer.flush();
+      LOG.info("Current Snapshot Index {}, takeSnapshot took {} ms",
+          lastAppliedIndex, Time.monotonicNow() - startTime);
+    }
+    super.takeSnapshot();
```
bharatviswa504, line 215: Do we need this call?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -179,25 +193,79 @@ public long takeSnapshot() throws IOException {
...
+      RaftServer server, RaftGroupId id, RaftStorage raftStorage)
+      throws IOException {
+    super.initialize(server, id, raftStorage);
+    storage.init(raftStorage);
+    loadSnapshot(storage.getLatestSnapshot());
+  }
+
+  private long loadSnapshot(SingleFileSnapshotInfo snapshot)
+      throws IOException {
+    if (snapshot == null) {
```
bharatviswa504, line 230: Here load snapshot should be loaded from TransactionInfo table?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -179,25 +193,79 @@ public long takeSnapshot() throws IOException {
     long startTime = Time.monotonicNow();
     TermIndex lastTermIndex = getLastAppliedTermIndex();
     long lastAppliedIndex = lastTermIndex.getIndex();
-    TransactionInfo lastAppliedTrxInfo =
-        TransactionInfo.fromTermIndex(lastTermIndex);
-    if (transactionBuffer.getLatestTrxInfo()
-        .compareTo(lastAppliedTrxInfo) < 0) {
+    if (isInitialized.get()) {
```
bharatviswa504, line 196: Here there is a chance of missing DB update.

Lets take 1,2,4 metadata/confUpdate and 3 is applyTransaction (Still not added to TransactionDBBuffer)

And now when **takeSnapshot** lastAppliedTermIndex will be 4, (As notifyConfigurationChanged will be called) and we flush. Restart SCM, we think 4 is applied but it is not.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
```diff
@@ -0,0 +1,188 @@
...
+        .setFactor(HddsProtos.ReplicationFactor.ONE).setKeyName(keyName)
+        .setRefreshPipeline(true).build();
+    final OmKeyInfo keyInfo = cluster.getOzoneManager().lookupKey(keyArgs);
+    final List<OmKeyLocationInfo> keyLocationInfos =
+        keyInfo.getKeyLocationVersions().get(0).getBlocksLatestVersionOnly();
+    final long containerID = keyLocationInfos.get(0).getContainerID();
+    for (int k = 0; k < numOfSCMs; k++) {
+      StorageContainerManager scm =
+          cluster.getStorageContainerManagers().get(k);
+      // flush to DB on each SCM
```
bharatviswa504, line 179: Question: Does without flush also container should be there right?

bharatviswa504: Can we check if with out flush also, it is available on all SCMS.

And also for this test, do you think we need to make sure first that each SCM catch up with transaction and then do checks. (As when request is processed, only leader applies to StateMachine)

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestSecureOzoneCluster.java
```diff
@@ -326,6 +329,10 @@ private void initSCM() throws IOException {
     scmStore.setScmId(scmId);
     // writes the version file properties
     scmStore.initialize();
+    if (!SCMHAUtils.isSCMHAEnabled(conf)) {
```
bharatviswa504, line 332: Question: Why do we need this step here?

bharatviswa504 (commented):
Need to review still newly added HACluster for SCM.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
bshashikant: yes, to ensure if there are any updates to sm directory by the base stateMachine impelementation.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
bshashikant: This cannot happen. applyTransaction is executed in the log order and applTransaction completely sequential in SCM stateMachine.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
bshashikant: That's why explicit take snapshot call is introduced so that, it first waits for all the appended log entries to complete apply .

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/TestSecureOzoneCluster.java
bshashikant: This step is required bcoz, otherwise, scm start won't elect a ratis leader as this is initializing scm version file without calling explicit scmInit.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
bshashikant: > That's why explicit take snapshot call is introduced so that, it first waits for all the appended log entries to complete apply  on each node. It will ensure all are upto date.

bharatviswa504: This is an explicit call that is being performed here, lets say 3rd SCM still not got in to applyphase for that transaction, then 3rd SCM will still not have right?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -1169,11 +1168,12 @@ public ReplicationManager getReplicationManager() {
...
 
   /**
-   * Check if the current scm is the leader.
-   * @return - if the current scm is the leader.
+   * Check if the current scm is the leader and ready for accepting requests.
+   * @return - if the current scm is the leader and is ready.
    */
   public boolean checkLeader() {
-    return scmContext.isLeader();
+    return scmContext.isLeader() && getScmHAManager().getRatisServer()
```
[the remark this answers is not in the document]
bshashikant: Its for a case, where ratis server is not initialised yet. By default scmContext.isLeader() will be false.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocol/ScmBlockLocationProtocol.java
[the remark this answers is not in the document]
bshashikant: I agree. Let's discuss it further.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
bshashikant: If 3rd SCM is not in the apply phase, takeSnapshot() will fail.

bshashikant: > If 3rd SCM is not in the apply phase, takeSnapshot() will fail. Snapshot guarantees whatever existing the logs commit is applied before snapshot is taken.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
```diff
@@ -0,0 +1,738 @@
...
+        conf.set(scmHttpsAddrKey, "127.0.0.1:" + (port + 3));
+        conf.setInt(scmRatisPortKey, port + 4);
+        //conf.setInt("ozone.scm.ha.ratis.bind.port", port + 4);
+        conf.set(dnPortKey, "127.0.0.1:" + (port + 5));
+        conf.set(blockClientKey, "127.0.0.1:" + (port + 6));
+        conf.set(ssClientKey, "127.0.0.1:" + (port + 7));
+        //conf.setInt("ozone.scm.ha.ratis.grpc.bind.port", port + 8);
+        conf.setInt(scmGrpcPortKey, port + 8);
+        scmNames.append(",").append("localhost:" + (port + 5));
+        conf.set(ScmConfigKeys.
```
bharatviswa504, line 587: Minor: I think not required

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
bshashikant: Snapshot ensures whatever is in the log is applied.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
bharatviswa504: This is completely removed now. I don't see this fixed.

bshashikant: Yes, its not required.

bshashikant: for the latest snapshot, getLatestSnapshot is already overriden in SCMStateMachine.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/MiniOzoneHAClusterImpl.java
```diff
@@ -0,0 +1,738 @@
...
+          scmList.clear();
+          ++retryCount;
+          LOG.info("MiniOzoneHACluster port conflicts, retried {} times",
+              retryCount);
+        }
+      }
+      return scmList;
+    }
+
+    protected void configureSCM() {
```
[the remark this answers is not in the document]
bshashikant: It will be called from super class.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
bshashikant: Addressed  in the recent. It is waiting for follower catch up before flush.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
bharatviswa504: But this is not setting lastAppliedIndex, how before loadSnapshot used to set

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -63,69 +61,120 @@
...
+            .setGroup(RaftGroup.valueOf(groupId)).build();
+  }
+
+  public static void initialize(String clusterId, String scmId,
+      SCMNodeDetails details, OzoneConfiguration conf) throws IOException {
+    final RaftGroup group = buildRaftGroup(details, scmId, clusterId);
+    RaftServer server = newRaftServer(scmId, conf)
+        .setGroup(group).build();
+    server.start();
+    waitforLeaderToBeReady(server, 60000, group);
```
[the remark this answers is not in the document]
bshashikant: yes..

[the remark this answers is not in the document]
bshashikant: I am adding TODO to make the timeout as well as sleep interval configurable. I think we should make it dependent on Ratis Leader election timeout.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -63,69 +61,120 @@
...
   }
 
   @Override
   public SCMRatisResponse submitRequest(SCMRatisRequest request)
       throws IOException, ExecutionException, InterruptedException {
     final RaftClientRequest raftClientRequest = RaftClientRequest.newBuilder()
         .setClientId(clientId)
-        .setServerId(division.getId())
-        .setGroupId(division.getGroup().getGroupId())
+        .setServerId(getDivision().getId())
```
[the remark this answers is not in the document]
bshashikant: Division info can change depending upon server state. Its safe to use this via Ratis. Its a cached object away inside ratisServerImpl.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -708,14 +708,14 @@ public static boolean scmBootstrap(OzoneConfiguration conf)
     }
     // The node here will try to fetch the cluster id from any of existing
     // running SCM instances.
-    // TODO: need to avoid failover to local SCM Node here
-    final ScmInfo scmInfo = HAUtils.getScmInfo(conf);
```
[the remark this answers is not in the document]
bshashikant: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -63,69 +61,120 @@
...
-    ((SCMStateMachine) division.getStateMachine())
-        .registerHandler(handlerType, handler);
+    stateMachine.registerHandler(handlerType, handler);
   }
 
   @Override
   public SCMRatisResponse submitRequest(SCMRatisRequest request)
       throws IOException, ExecutionException, InterruptedException {
     final RaftClientRequest raftClientRequest = RaftClientRequest.newBuilder()
         .setClientId(clientId)
```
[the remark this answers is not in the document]
bshashikant: Thanks for pointing this out. Addressed this.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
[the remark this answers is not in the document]
bshashikant: Done.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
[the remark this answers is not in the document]
bharatviswa504: I don't see this being addressed?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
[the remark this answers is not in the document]
bharatviswa504: Sure. Let's open a new Jira.
> I think we should make it dependent on Ratis Leader election timeout.
Do you mean for interval check?

bharatviswa504 (commented):
Overall LGTM. One comment regarding usage of clientID and callID usage need to be set from Server API's that looks missing.
And also getting localNodeID need to be fetched, as setNodeSpecificConfigKeys does not ozone.scm.node.id.

bshashikant:
> Overall LGTM. One comment regarding usage of clientID and callID usage need to be set from Server API's that looks missing.
> And also getting localNodeID need to be fetched, as setNodeSpecificConfigKeys does not ozone.scm.node.id.

There is only one central place of submitting ratis requests and there is only client Id which will be same for all requests submitted to the server on each node. I don't think , we need to get the clientId from server here as there is only one client.
Same goes for "callId" as well.

However, this change is not related to the patch. We can discuss it further and make change accordingly, in a different jira.

bshashikant:
@bharatviswa504 , i have filed https://issues.apache.org/jira/browse/HDDS-4877 and https://issues.apache.org/jira/browse/HDDS-4878 to address the issues discussed.

bshashikant:
Thanks @bshashikant , @vivekratnavel . @GlenGeng for the reviews. I have committed this.

PR #1954, opened 2021-02-23
HDDS-4683. [FSO]ListKeys: do lookup in dir and file tables

rakeshadr:

This task is to perform #listKeys by searching the startKey, keyPrefix in Dir & File Tables.

https://issues.apache.org/jira/browse/HDDS-4683

## How was this patch tested?

Added only one UT now.  I will add more test cases. I'm thinking something like large dirs and sub-dirs/files.

Also, there is an open point: #listKeys("key-a-"), basically partial keyName and startKey. For example, actual keyName stored in DB are "key-a-10", "key-a-11", "key-a-12"..."key-z-10", "key-z-11", "key-z-12".

### DIFF
--- a/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
+++ b/hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
@@ -21,12 +21,14 @@ package org.apache.hadoop.ozone.client;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.hdds.client.OzoneQuota;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.protocol.StorageType;
 import org.apache.hadoop.hdds.client.ReplicationFactor;
 import org.apache.hadoop.hdds.client.ReplicationType;
 import org.apache.hadoop.hdds.scm.client.HddsClientUtils;
+import org.apache.hadoop.ozone.OmUtils;
 import org.apache.hadoop.ozone.OzoneConfigKeys;
 import org.apache.hadoop.ozone.client.io.OzoneInputStream;
 import org.apache.hadoop.ozone.client.io.OzoneOutputStream;
@@ -35,7 +37,9 @@ import org.apache.hadoop.ozone.OzoneAcl;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
 import org.apache.hadoop.ozone.om.helpers.OmMultipartInfo;
 import org.apache.hadoop.ozone.om.helpers.OmMultipartUploadCompleteInfo;
+import org.apache.hadoop.ozone.om.helpers.OmKeyInfo;
 import org.apache.hadoop.ozone.om.helpers.OzoneFileStatus;
+import org.apache.hadoop.ozone.om.helpers.OzoneFSUtils;
 import org.apache.hadoop.ozone.om.helpers.WithMetadata;
 import org.apache.hadoop.ozone.security.acl.OzoneObj;
 import org.apache.hadoop.ozone.security.acl.OzoneObjInfo;
@@ -43,13 +47,16 @@ import org.apache.hadoop.util.Time;
 
 import java.io.IOException;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Stack;
 import java.util.NoSuchElementException;
 
 import static org.apache.hadoop.ozone.OzoneConsts.QUOTA_RESET;
+import static org.apache.hadoop.ozone.OzoneConsts.OZONE_URI_DELIMITER;
 
 /**
  * A class that encapsulates OzoneBucket.
@@ -544,6 +551,10 @@ public class OzoneBucket extends WithMetadata {
    */
   public Iterator<? extends OzoneKey> listKeys(String keyPrefix,
       String prevKey) throws IOException {
+
+    if(OzoneFSUtils.isFSOptimizedBucket(getMetadata())){
+      return new KeyIteratorV1(keyPrefix, prevKey);
+    }
     return new KeyIterator(keyPrefix, prevKey);
   }
 
@@ -788,6 +799,13 @@ public class OzoneBucket extends WithMetadata {
     private Iterator<OzoneKey> currentIterator;
     private OzoneKey currentValue;
 
+    String getKeyPrefix() {
+      return keyPrefix;
+    }
+
+    void setKeyprefix(String keyPrefixPath) {
+      keyPrefix = keyPrefixPath;
+    }
 
     /**
      * Creates an Iterator to iterate over all keys after prevKey in the bucket.
@@ -796,7 +814,7 @@ public class OzoneBucket extends WithMetadata {
      * @param keyPrefix
      */
     KeyIterator(String keyPrefix, String prevKey) throws IOException{
-      this.keyPrefix = keyPrefix;
+      setKeyprefix(keyPrefix);
       this.currentValue = null;
       this.currentIterator = getNextListOfKeys(prevKey).iterator();
     }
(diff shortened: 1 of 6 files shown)


hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
```diff
@@ -828,10 +846,273 @@ public OzoneKey next() {
...
+      // TODO: Back and Forth seek all the files & dirs, starting from
+      //  startKey till keyPrefix.
+
+      return keysResultList;
+    }
+
+    /**
+     * List children under the given keyPrefix path. This doesn't guarantee
+     * order. Internally, it does recursive #listStatus calls to list all the
+     * sub-keysResultList.
```
linyiqun, line 928: >This doesn't guarantee order.

The listStatus result list returned should already in order I think. Why here say doesn't guarantee order?

linyiqun (commented):
@rakeshadr , haven't taken a deep review for this PR, but catch one thing makes me a little confused.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
rakeshadr: Thanks a lot @linyiqun for the reviews and your comments greatly help to push this feature fast. 

**[FileSystem#listStatus](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L1890)** its not guaranteeing sorted order. `ozone#listStatus` user final status list is also not maintaining sorted order. 

We are sorting files and dirs list [separately in OM KeyManagerImpl](https://github.com/apache/ozone/blob/HDDS-2939/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L2353), this is done for batch-wise key iteration. OM maintains only seek order to avoid infinite looping - first it will seek files(in Filetable) and then dirs(in DirTable). If the `startKey` is a directory that means, all the files has been fetched in previous iteration and it can stop seeking after fetching all the dirs from DirTable.


`Ozone#listKeys `also not expects sorted order and existing javadoc didn't have anything specific about the ordering. But user can experience a sorted list in V0 code because RocksDB maintains order. So, I thought of explicitly mention about the order part in V1 code base.

rakeshadr:
FYI, raised HDDS-4859 jira to handle the back and forth search.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
```diff
@@ -828,10 +846,285 @@ public OzoneKey next() {
...
+      // 2. Get immediate children of keyPrefix, starting with startKey
+      List<OzoneFileStatus> statuses = proxy.listStatus(volumeName, name,
+              keyPrefix, false, startKey, listCacheSize);
+
+      // 3. Special case: ListKey expects keyPrefix element should present in
+      // the resultList, only if startKey is blank. If startKey is not blank
+      // then resultList shouldn't contain the startKey element.
+      // Since proxy#listStatus API won't return keyPrefix element in the
+      // resultList. So, this is to add user given keyPrefix to the return list.
+      addKeyPrefixInfoToResultList(keyPrefix, startKey, keysResultList);
```
linyiqun, line 992: Can you share the code link related with this case? I didn't actually catch this point, the code I am seeing is [KeyManagerImpl.java#L2173](https://github.com/apache/ozone/blob/HDDS-2939/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L2173). But seems the code I am checking is not the place that match the above proxy.listStatus call.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
```diff
@@ -275,6 +285,246 @@ public void testLookupKey() throws Exception {
...
+        ozoneBucket.listKeys("a/b2/d2", "a/b2/d2/d21.tx");
+    expectedKeys = new LinkedList<>();
+    expectedKeys.add("a/b2/d2/d22.tx");
+    checkKeyList(ozoneKeyIterator, expectedKeys);
+
+    // Boundary case - last node in the depth-first-traversal
+    ozoneKeyIterator =
+        ozoneBucket.listKeys("a/b3/e3", "a/b3/e3/e31.tx");
+    expectedKeys = new LinkedList<>();
+    checkKeyList(ozoneKeyIterator, expectedKeys);
```
linyiqun, line 411: Can you add one more case for that keyPrefix is null meanwhile startKey is not  null, like listKeys(null,"xxx");

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
linyiqun: Okay, get it.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
rakeshadr: #listKeys API expects the user given `keyPrefix` element should be there in the `returnKeysList`. Since client is calling #listStatus API, it won't include the `keyPrefix` element in returnKeysList. Please refer [KeyManagerImpl code](https://github.com/apache/ozone/blob/HDDS-2939/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java#L2229) where OM is skipping the keyPrefix element. So, I have made an explicit call to getFileStatus("keyPrefix") and add to the `returnKeysList`.

Please refer [existing test case for the existence of keyPrefix](https://github.com/apache/ozone/blob/master/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFSWithObjectStoreCreate.java#L376) in the listKeys return values.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
rakeshadr: Sure

linyiqun: Makes sense to me.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
linyiqun: Get it, thanks @rakeshadr .

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/OzoneBucket.java
```diff
@@ -828,10 +846,286 @@ public OzoneKey next() {
...
+      addedKeyPrefix = true;
+
+      // not required to addKeyPrefix
+      // case-1) if keyPrefix is null or empty
+      // case-2) if startKey is not null or empty
+      if (StringUtils.isBlank(keyPrefix) || StringUtils.isNotBlank(startKey)) {
+        return;
+      }
+
+      // TODO: Handle a case, where startKey not started with keyPrefix.
```
linyiqun, line 1106: Will we also addressed on this case in HDDS-4859?

rakeshadr: Yes. Since listKeys internally using listStatus, HDDS-4364 will take care both listKeys & listStatus.
I will include jira ID in TODO.

linyiqun:
@rakeshadr , the PR looks good to me now. Please check above minor comment.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestObjectStoreV1.java
```diff
@@ -275,6 +285,259 @@ public void testLookupKey() throws Exception {
...
+    createKeys(ozoneBucket, keys);
+
+    // Root level listing keys
+    Iterator<? extends OzoneKey> ozoneKeyIterator =
+        ozoneBucket.listKeys(null, null);
+    verifyFullTreeStructure(ozoneKeyIterator);
+
+    ozoneKeyIterator =
+        ozoneBucket.listKeys(null, null);
+    verifyFullTreeStructure(ozoneKeyIterator);
```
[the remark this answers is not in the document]
rakeshadr: Sure, I will remove it.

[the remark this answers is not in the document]
rakeshadr: Done

rakeshadr:
> @rakeshadr , the PR looks good to me now. Please check above minor comment.

Thanks a lot @linyiqun for the review comments. I will wait for QA report and then merge it.

PR #1955, opened 2021-02-23
HDDS-4840. Make datanode db profile configurable with existing hdds.d…

guihecheng:
…b.profile

Make datanode use the config item `hdds.db.profile`, instead of statically default to use spinning disk profile.

https://issues.apache.org/jira/browse/HDDS-4840

## How was this patch tested?
UT:TestKeyValueContainer:testDBProfileAffectsDBOptions


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/metadata/AbstractDatanodeStore.java
```diff
@@ -89,11 +91,14 @@ protected AbstractDatanodeStore(ConfigurationSource config, long containerID,
...
 
+    dbProfile = config.getEnum(HDDS_DB_PROFILE,
+        HDDS_DEFAULT_DB_PROFILE);
+
     // The same config instance is used on each datanode, so we can share the
     // corresponding column family options, providing a single shared cache
     // for all containers on a datanode.
     if (!OPTIONS_CACHE.containsKey(config)) {
-      OPTIONS_CACHE.put(config, buildColumnFamilyOptions(config));
+      OPTIONS_CACHE.put(config, buildColumnFamilyOptions(config, dbProfile));
```
adoroszlai, line 101: Nit: instead of passing `dbProfile` member variable, can you please make `buildColumnFamilyOptions` non-static?

adoroszlai (commented):
Thanks @guihecheng for working on this.  Can you please add some unit tests, eg. in `TestKeyValueContainer`?  (`testContainersShareColumnFamilyOptions` might be a good place to start.)

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/metadata/AbstractDatanodeStore.java
guihecheng: oh, thanks for your advice, I'll fix it and try to add a simple test case~

guihecheng:
@adoroszlai updated for review, thanks~

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/DBProfile.java
```diff
@@ -105,8 +105,26 @@ public ColumnFamilyOptions getColumnFamilyOptions() {
       columnFamilyOptions.setCompactionStyle(CompactionStyle.LEVEL);
       return columnFamilyOptions;
     }
+  },
+  // NOTE: just for unit test now, don't use it in production
+  TEST {
```
adoroszlai, line 110: I think we should try to avoid introducing test-specific items in production code.  Why not test with existing `DISK` and `SSD` profiles instead?

adoroszlai (commented):
Thanks @guihecheng for updating the patch.  Mostly looks good to me, but I think `TEST` profile is unnecessary.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/DBProfile.java
guihecheng: @adoroszlai oh, it is simply because the ColumnFamilyOptions for DISK & SSD happen to be the same, as I tested, so a new profile TEST is introduced.

adoroszlai: I don't think they are intended to be the same:

https://github.com/apache/ozone/blob/e658f40a5cf1263a7b8036f9f24bac74c8e90372/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/DBProfile.java#L104-L106

If they are, then it's probably a bug.

guihecheng: Yes, I've read this line that setting the compactionStyle to CompactionStyle.LEVEL, but actually it is just the default style, so the results are still the same.

adoroszlai: Good point.  How about checking non-equality of `getDBOptions().compactionReadaheadSize()` between the two profiles?  Those _are_ different.

guihecheng: @adoroszlai well, I admit that there merely any difference to check the DBOptions or ColumnFamilyOptions, here I intended to check the ColumnFamilyOptions just because it is convenient to use the existing OPTIONS_CACHE in AbstractDatanodeStore.
So to check the DBOptions, I'll try to export the DBOptions as well(or the DBProfile directly), just for test, what do you think?

adoroszlai: @guihecheng Thanks for checking.  I think providing an accessor for `AbstractDatanodeStore`'s `DBProfile` for the test rather is much less intrusive than the fake profile.

guihecheng:
@adoroszlai updated the utest case for review, now we get rid of the TEST profile.

adoroszlai (approved):
Thanks @guihecheng for updating the patch.

adoroszlai:
@errose28 Would you like to review?

errose28:
Sure, thanks @adoroszlai and @guihecheng I will take a look.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/metadata/AbstractDatanodeStore.java
```diff
@@ -243,8 +253,7 @@ private static ColumnFamilyOptions buildColumnFamilyOptions(
         .setPinL0FilterAndIndexBlocksInCache(true)
```
errose28, line 253: nit: I think we can remove lines 238-239 `RocksDB.loadLibrary` since this method is no longer static.

errose28 (commented):
Just a minor inline comment, otherwise LGTM. Thanks @guihecheng for working on this!

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/metadata/AbstractDatanodeStore.java
guihecheng: > nit: I think we can remove lines 238-239 `RocksDB.loadLibrary` since this method is no longer static.

oh, by reading the comment in the code, it seems to be reasonable to remove this line, but I'll check and test it.

guihecheng: @errose28 It seems that you are right, the `RocksDB.loadLibrary` is not needed anymore as I testd, will update, thx~

guihecheng:
well, there's seem to be a unrelated case failure.

elek:
Merging it as we have green build. Thanks the review @adoroszlai and @errose28

PR #1958, opened 2021-02-24
HDDS-4861. [SCM HA Security] Implement generate SCM certificate.

bharatviswa504:

This Jira is to implement generatePeerSCMCertificate which will be called during bootStrap and also during init by primary SCM. This is similar to how OM and DN gets certificate.

During implementation as the validCerTable key is SerialId which is BigInteger, created a new table validScmCertsTable to store scm certs. This is being done so that we can implement querySCMCertificates which can return all SCM certs.

Also, I see there is a TODO for list Certs based on role. Not sure if there is a way in the current code to return certs from the table. If there is any approach, will rework this PR based on review comments.

https://issues.apache.org/jira/browse/HDDS-4861

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/protocol/SCMSecurityProtocol.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/protocol/SCMSecurityProtocol.java
@@ -67,7 +67,7 @@ public interface SCMSecurityProtocol {
   /**
    * Get signed certificate for SCM.
    *
-   * @param scmNodeDetails       - SCM Node Details.
+   * @param scmNodeDetails  - SCM Node Details.
    * @param certSignReq     - Certificate signing request.
    * @return String         - pem encoded SCM signed
    *                          certificate.
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/protocol/SCMSecurityProtocol.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/protocol/SCMSecurityProtocol.java
@@ -23,6 +23,7 @@ import org.apache.hadoop.hdds.annotation.InterfaceAudience;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.DatanodeDetailsProto;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.OzoneManagerDetailsProto;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.ScmNodeDetailsProto;
 import org.apache.hadoop.hdds.scm.ScmConfig;
 import org.apache.hadoop.security.KerberosInfo;
 
@@ -60,6 +61,18 @@ public interface SCMSecurityProtocol {
    *                          certificate.
    */
   String getOMCertificate(OzoneManagerDetailsProto omDetails,
+      String certSignReq) throws IOException;
+
+
+  /**
+   * Get signed certificate for SCM.
+   *
+   * @param scmNodeDetails       - SCM Node Details.
+   * @param certSignReq     - Certificate signing request.
+   * @return String         - pem encoded SCM signed
+   *                          certificate.
+   */
+  String getSCMCertificate(ScmNodeDetailsProto scmNodeDetails,
       String certSignReq) throws IOException;
 
   /**
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
@@ -20,6 +20,7 @@
 package org.apache.hadoop.hdds.security.x509.certificate.authority;
 
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.NodeType;
 import org.apache.hadoop.hdds.security.exception.SCMSecurityException;
 import org.apache.hadoop.hdds.security.x509.SecurityConfig;
 import org.apache.hadoop.hdds.security.x509.certificate.authority.CertificateApprover.ApprovalType;
@@ -82,7 +83,7 @@ public interface CertificateServer {
    */
   Future<X509CertificateHolder> requestCertificate(
       PKCS10CertificationRequest csr,
-      CertificateApprover.ApprovalType type)
+      CertificateApprover.ApprovalType type, NodeType nodeType)
       throws SCMSecurityException;
 
 
@@ -96,7 +97,7 @@ public interface CertificateServer {
    * @throws SCMSecurityException - on Error.
    */
   Future<X509CertificateHolder> requestCertificate(String csr,
-      ApprovalType type) throws IOException;
+      ApprovalType type, NodeType nodeType) throws IOException;
 
   /**
    * Revokes a Certificate issued by this CertificateServer.
@@ -122,7 +123,7 @@ public interface CertificateServer {
    * @return
    * @throws IOException
    */
-  List<X509Certificate> listCertificate(HddsProtos.NodeType type,
+  List<X509Certificate> listCertificate(NodeType type,
       long startSerialId, int count, boolean isRevoked) throws IOException;
 
   /**
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
@@ -19,7 +19,6 @@
 
 package org.apache.hadoop.hdds.security.x509.certificate.authority;
 
-import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos.NodeType;
 import org.apache.hadoop.hdds.security.exception.SCMSecurityException;
 import org.apache.hadoop.hdds.security.x509.SecurityConfig;
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
@@ -77,6 +77,7 @@ public interface CertificateServer {
    *
    * @param csr  - Certificate Signing Request.
    * @param type - An Enum which says what kind of approval process to follow.
+   * @param nodeType: OM/SCM/DN
    * @return A future that will have this certificate when this request is
    * approved.
    * @throws SCMSecurityException - on Error.
@@ -92,6 +93,7 @@ public interface CertificateServer {
    *
    * @param csr - Certificate Signing Request as a PEM encoded String.
    * @param type - An Enum which says what kind of approval process to follow.
+   * @param nodeType: OM/SCM/DN
    * @return A future that will have this certificate when this request is
    * approved.
    * @throws SCMSecurityException - on Error.
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultCAServer.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultCAServer.java
@@ -23,6 +23,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.apache.commons.validator.routines.DomainValidator;
 import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos.NodeType;
 import org.apache.hadoop.hdds.security.exception.SCMSecurityException;
 import org.apache.hadoop.hdds.security.x509.SecurityConfig;
 import org.apache.hadoop.hdds.security.x509.certificate.authority.PKIProfiles.DefaultProfile;
@@ -206,7 +207,7 @@ public class DefaultCAServer implements CertificateServer {
   @Override
   public Future<X509CertificateHolder> requestCertificate(
       PKCS10CertificationRequest csr,
-      CertificateApprover.ApprovalType approverType) {
+      CertificateApprover.ApprovalType approverType, NodeType nodeType) {
     LocalDate beginDate = LocalDate.now().atStartOfDay().toLocalDate();
     LocalDateTime temp = LocalDateTime.of(beginDate, LocalTime.MIDNIGHT);
     LocalDate endDate =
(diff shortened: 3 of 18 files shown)


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
```diff
@@ -82,7 +82,7 @@ X509Certificate getCertificate(String certSerialId)
    */
   Future<X509CertificateHolder> requestCertificate(
       PKCS10CertificationRequest csr,
-      CertificateApprover.ApprovalType type)
+      CertificateApprover.ApprovalType type, boolean scmCert)
```
xiaoyuyao, line 85: can we use a enum for role instead with om/scm/dn for now, that might help query certificate by role as well.

hadoop-hdds/interface-server/src/main/proto/ScmServerSecurityProtocol.proto
```diff
@@ -49,6 +49,8 @@ message SCMSecurityRequest {
     optional SCMGetCertificateRequestProto getCertificateRequest = 5;
     optional SCMGetCACertificateRequestProto getCACertificateRequest = 6;
     optional SCMListCertificateRequestProto listCertificateRequest = 7;
+    optional SCMGeneratePeerSCMCertRequestProto
```
xiaoyuyao, line 52: the root ca scm also need to call this, can we remove the Peer here?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
```diff
@@ -105,6 +114,6 @@ public String getLocationConfigKey() {
   @Override
   public DBColumnFamilyDefinition[] getColumnFamilies() {
     return new DBColumnFamilyDefinition[] {DELETED_BLOCKS, VALID_CERTS,
-        REVOKED_CERTS, PIPELINES, CONTAINERS, TRANSACTIONINFO};
+        VALID_SCM_CERTS, REVOKED_CERTS, PIPELINES, CONTAINERS, TRANSACTIONINFO};
```
xiaoyuyao, line 117: we might also need a REVOKED_SCM_CERTS with the CRL work given we are putting cert of different roles into different table with this change.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMCertStore.java
```diff
@@ -93,6 +114,8 @@ public void revokeCertificate(BigInteger serialID) throws IOException {
                scmMetadataStore.getStore().initBatchOperation();) {
         scmMetadataStore.getRevokedCertsTable()
             .putWithBatch(batch, serialID, cert);
+        scmMetadataStore.getValidSCMCertsTable().deleteWithBatch(batch,
```
xiaoyuyao, line 117: we only need this if the revoked cert is a scm cert.

xiaoyuyao (commented):
Thanks @bharatviswa504  for the patch. LGTM overall, a few comments added inline.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMSecurityProtocolServer.java
```diff
@@ -156,6 +159,44 @@ public String getOMCertificate(OzoneManagerDetailsProto omDetails,
...
+      String certSignReq) throws IOException {
+    Objects.requireNonNull(scmNodeDetails);
+    LOGGER.info("Processing CSR for scm {}, nodeId: {}",
+        scmNodeDetails.getHostName(), scmNodeDetails.getScmNodeId());
+
+    // Check clusterID
+    if (storageContainerManager.getClusterId().equals(
+        scmNodeDetails.getClusterId())) {
+      throw new IOException("SCM ClusterId mismatch. Peer SCM ClusterId, " +
+          scmNodeDetails.getClusterId() + "leader SCM ClusterId "
```
adoroszlai, line 181: Isn't "leader" confusing here?  If I understand correctly, this might precede Ratis leader election.

Also, a space is needed between clusterID and "leader" (or its replacement).

```suggestion
      throw new IOException("SCM ClusterId mismatch. Peer SCM ClusterId " +
          scmNodeDetails.getClusterId() + ", primary SCM ClusterId "
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMSecurityProtocolServer.java
```diff
@@ -156,6 +159,44 @@ public String getOMCertificate(OzoneManagerDetailsProto omDetails,
...
+            KERBEROS_TRUSTED, true);
+
+    try {
+      return CertificateCodec.getPEMEncodedString(future.get());
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new IOException("generateSCMPeerCertificate operation failed. ", e);
+    } catch (ExecutionException e) {
+      throw new IOException("generateSCMPeerCertificate operation failed. ", e);
+    }
```
adoroszlai, line 197: Can you please extract this repeated block to a method?  Now appears three times with slight differences.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultApprover.java
```diff
@@ -88,7 +88,8 @@ public  X509CertificateHolder sign(
       Date validTill,
       PKCS10CertificationRequest certificationRequest,
       String scmId,
-      String clusterId) throws IOException, OperatorCreationException {
+      String clusterId) throws IOException,
+      OperatorCreationException {
```
adoroszlai, line 92: Nit: can we omit this formatting-only change?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultCAServer.java
```diff
@@ -252,23 +252,32 @@ private KeyPair getCAKeys() throws IOException {
...
         getCAKeys().getPrivate(),
         getCACertificate(), java.sql.Date.valueOf(beginDate),
         java.sql.Date.valueOf(endDate), csr, scmID, clusterID);
-    store.storeValidCertificate(xcert.getSerialNumber(),
-        CertificateCodec.getX509Certificate(xcert));
+    if (scmCert) {
+      // If scm cert store in scm cert table and valid cert table.
+      // This is to help to return scm certs when get certs call.
+      store.storeValidScmCertificate(xcert.getSerialNumber(),
+          CertificateCodec.getX509Certificate(xcert));
```
adoroszlai, line 266: Can you please explain why SCM certs need to be also stored in a separate table?

adoroszlai (changes requested):

hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/TestHddsUtils.java
```diff
@@ -44,6 +46,8 @@
 
   @Test
   public void testGetHostName() {
+    UUID uii = UUID.randomUUID();
+    UUID.nameUUIDFromBytes(uii.toString().getBytes(StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
bharatviswa504: I was debugging some issue yesterday, by mistake, the code came in this commit.
Will remove it

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/protocol/SCMSecurityProtocol.java
```diff
@@ -62,6 +63,18 @@ String getDataNodeCertificate(
...
+
+  /**
+   * Get SCM signed certificate for OM.
+   *
+   * @param scmNodeDetails       - DataNode Details.
+   * @param certSignReq     - Certificate signing request.
+   * @return String         - pem encoded SCM signed
+   *                          certificate.
+   */
+  String generateSCMPeerCertificate(ScmNodeDetailsProto scmNodeDetails,
```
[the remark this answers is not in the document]
bharatviswa504: Yes, the other APIs should also start with generate, the correct naming is to start with generate. I will use generate for this new API. I can open a new Jira to change  for others.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/protocol/SCMSecurityProtocol.java
```diff
@@ -62,6 +63,18 @@ String getDataNodeCertificate(
   String getOMCertificate(OzoneManagerDetailsProto omDetails,
       String certSignReq) throws IOException;
 
+
+  /**
+   * Get SCM signed certificate for OM.
+   *
+   * @param scmNodeDetails       - DataNode Details.
```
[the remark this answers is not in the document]
bharatviswa504: Done.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/CertificateServer.java
bharatviswa504: Yes for now in this PR will only address for SCM, I can open a new Jira to do for other roles. So, that it will help during list with role type.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultApprover.java
bharatviswa504: Done

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultCAServer.java
bharatviswa504: SCM Certs are stored in scmCertTable because we want to queryCerts for scm Role.  This will come in help during SCM Startup, where each SCM need to get all SCM certs of its quorum.
And also added to fall back table validCertTables because to not to break list API, where it is listing all certs.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
bharatviswa504: Right now added scmCert to validCertsTable, so that revoke will remove from both for Scmcerts.
 I think once we have a separate table for each role, we can add that.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMCertStore.java
bharatviswa504: Done

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/protocol/SCMSecurityProtocol.java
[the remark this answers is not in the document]
adoroszlai: Can we name it `get...` now, and rename with the rest?  I know it's a bit more work, but there are some benefits:

1. They are consistent with one another both before and after the rename.
2. If renaming should turn out to be not possible (eg. because it might introduce incompatibility) or not desired (eg. someone might argue that `get` reflects the operation more closely), then we are not stuck with inconsistent names.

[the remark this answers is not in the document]
bharatviswa504: I don't see why it will be an incompatible change. For now, I followed ur suggestion

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMSecurityProtocolServer.java
bharatviswa504: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMSecurityProtocolServer.java
bharatviswa504: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/SCMSecurityProtocolServerSideTranslatorPB.java
```diff
@@ -146,6 +154,31 @@ public SCMGetCertResponseProto getDataNodeCertificate(
 
   }
 
+  /**
+   * Get SCM signed certificate for DataNode.
+   *
+   * @param request
+   * @return SCMGetDataNodeCertResponseProto.
```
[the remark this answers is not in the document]
bharatviswa504: Removed it

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/SCMSecurityProtocolServerSideTranslatorPB.java
```diff
@@ -146,6 +154,31 @@ public SCMGetCertResponseProto getDataNodeCertificate(
 
   }
 
+  /**
+   * Get SCM signed certificate for DataNode.
+   *
+   * @param request
```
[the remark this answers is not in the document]
bharatviswa504: Added now

adoroszlai (approved):
Thanks @bharatviswa504 for updating the patch, LGTM.

bshashikant (approved):
Looks good with a minor sugggestion regarding comments.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultCAServer.java
```diff
@@ -252,23 +253,34 @@ private KeyPair getCAKeys() throws IOException {
...
+      throws IOException,
       OperatorCreationException, CertificateException {
     X509CertificateHolder xcert = approver.sign(config,
         getCAKeys().getPrivate(),
         getCACertificate(), java.sql.Date.valueOf(beginDate),
         java.sql.Date.valueOf(endDate), csr, scmID, clusterID);
-    store.storeValidCertificate(xcert.getSerialNumber(),
-        CertificateCodec.getX509Certificate(xcert));
+    if (nodeType.equals(NodeType.SCM)) {
+      // If scm cert store in scm cert table and valid cert table.
```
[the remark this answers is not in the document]
bharatviswa504: Done

bharatviswa504:
@xiaoyuyao I have addressed review comments.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
bharatviswa504: Yes, might be for list for role SCM for certtype RevokedCerts we might need this. Figured out during implementation of listCertificates. As for security work this is not required, will open a new Jira to fix this later.

xiaoyuyao (approved):
LGTM, +1

bharatviswa504:
Thank You everyone for the review.

PR #1962, opened 2021-02-25
HDDS-4868. Ozone datanode initraftlog fail due to bad disk so can not communicate to SCM

cchenax:
When datanode initraftlog fail due to bad disk,the datanode can not connect to SCM,and the datanode will always throw the exception,so we should let datanode throw the runtime exception once,and SCM will remove the raftgroup 

https://issues.apache.org/jira/browse/HDDS-4868

## How was this patch tested?
I will mock the datanode initraftlog fail and the process of the datanode will not always throw the exception

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/endpoint/VersionEndpointTask.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/endpoint/VersionEndpointTask.java
@@ -109,7 +109,11 @@ public class VersionEndpointTask implements
           }
 
           // Start the container services after getting the version information
-          ozoneContainer.start(scmId);
+          try {
+            ozoneContainer.start(scmId);
+          } catch (IOException ex) {
+            rpcEndPoint.logIfNeeded(ex);
+          }
         }
         EndpointStateMachine.EndPointStates nextState =
             rpcEndPoint.getState().getNextState();


adoroszlai (commented):
Thanks @cchenax for proposing a fix.  However, simply catching the exception does not work, because the datanode would not be functional without its container services.  (You can simulate the error by throwing an `IOException` from `XceiverServerRatis.start()`).

cchenax:
> Thanks @cchenax for proposing a fix. However, simply catching the exception does not work, because the datanode would not be functional without its container services. (You can simulate the error by throwing an `IOException` from `XceiverServerRatis.start()`).

when the datanode was failed,the scm would through heartbeat to detect the datanode which failed.So the scm would remove the container.This datanode really not work.

adoroszlai:
> when the datanode was failed,the scm would through heartbeat to detect the datanode which failed.So the scm would remove the container.This datanode really not work.

SCM is expected to mark the datanode first `STALE`, then `DEAD`, after not receiving heartbeat for some time.  I don't see why keeping the datanode alive as a zombie would be necessary.

Also, I think 8a0b530 is unrelated and intended for another issue (HDDS-4983).

cchenax:
> > when the datanode was failed,the scm would through heartbeat to detect the datanode which failed.So the scm would remove the container.This datanode really not work.
> 
> SCM is expected to mark the datanode first `STALE`, then `DEAD`, after not receiving heartbeat for some time. I don't see why keeping the datanode alive as a zombie would be necessary.
> 
> Also, I think [8a0b530](https://github.com/apache/ozone/commit/8a0b5308032ae91d37ce5620c6cee501992b3515) is unrelated and intended for another issue ([HDDS-4983](https://issues.apache.org/jira/browse/HDDS-4983)).

cchenax:
> > when the datanode was failed,the scm would through heartbeat to detect the datanode which failed.So the scm would remove the container.This datanode really not work.
> 
> SCM is expected to mark the datanode first `STALE`, then `DEAD`, after not receiving heartbeat for some time. I don't see why keeping the datanode alive as a zombie would be necessary.
> 
> Also, I think [8a0b530](https://github.com/apache/ozone/commit/8a0b5308032ae91d37ce5620c6cee501992b3515) is unrelated and intended for another issue ([HDDS-4983](https://issues.apache.org/jira/browse/HDDS-4983)).

Ok，I know what you mean，the datanode initraftgroup would skip the init failed raft group,the other normal pipeline raft group would init success，and the datanode would not fail.I would change the ratis code.

ChenSammi:
Close this PR. It issue will be handled in ratis RATIS-1350.

PR #1965, opened 2021-02-25
HDDS-4490.[FSO]RenameAndDelete : make ofs#rename and ofs#delete an atomic operation.

sadanand48:
Add atomic rename and delete api's to BasicRootedOzoneFileSystem (OFS).  

https://issues.apache.org/jira/browse/HDDS-4490

## How was this patch tested?
Unit tests. (Same as TestRootedOzoneFileSystem) but with FSO configs.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -111,6 +111,14 @@ public class TestRootedOzoneFileSystem {
     omRatisEnabled = enableOMRatis;
   }
 
+  public static FileSystem getFs() {
+    return fs;
+  }
+
+  public static Path getBucketPath() {
+    return bucketPath;
+  }
+
   @Rule
   public Timeout globalTimeout = new Timeout(300_000);
 
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
@@ -62,7 +62,10 @@ import org.junit.Test;
 import org.junit.rules.Timeout;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -72,12 +75,12 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
+import java.util.Random;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_TRASH_INTERVAL_KEY;
-import static org.apache.hadoop.fs.FileSystem.LOG;
 import static org.apache.hadoop.fs.FileSystem.TRASH_PREFIX;
 import static org.apache.hadoop.fs.ozone.Constants.LISTING_PAGE_SIZE;
 import static org.apache.hadoop.ozone.OzoneAcl.AclScope.ACCESS;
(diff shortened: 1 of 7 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestReadRetries.java
```diff
@@ -241,16 +233,6 @@ public void testPutKeyAndGetKeyThreeNodes()
     factory.releaseClient(clientSpi, false);
   }
 
-  private void verifyIntermediateDir(OzoneBucket bucket,
```
sadanand48, line 244: removed this because this expects an Exception while performing getKey on a directory which has been changed as part of this PR.

rakeshadr: Thanks @sadanand48 for the clear explanation. This test were running with below params. This gives me the impression that with existing code `enableFileSystemFlag=true`, it expects exception should be thrown for a dir.

enableFileSystemFlag=true and layoutVersion=V0
enableFileSystemFlag=true and layoutVersion=V1

I think, we need to get the correct functionality or the expectation with enableFileSystemFlag=true or false.

rakeshadr (commented):
Thanks @sadanand48 for working on this.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
```diff
@@ -1830,14 +1837,23 @@ private OzoneFileStatus getOzoneFileStatus(String volumeName,
       // Check if the key is a file.
       String fileKeyBytes = metadataManager.getOzoneKey(
               volumeName, bucketName, keyName);
+      // if the queried key is already a directory key (For eg: Keyname= a/b/c/)
+      if (fileKeyBytes.endsWith(OZONE_URI_DELIMITER)){
```
rakeshadr, line 1841: Should we need this additional check for the trailing slash? Will the below line fetch the given directory a/b/c/ or file a/b/c key from KeyTable, right?
`fileKeyInfo = metadataManager.getKeyTable().get(fileKeyBytes);`

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystemV1.java
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * OFS: Test recursive listStatus on root and volume.
+   */
+  @Override
+  @Ignore("TODO:HDDS-4360")
+  public void testListStatusRootAndVolumeRecursive() throws IOException {
+  }
+
+  @Test
+  public void testRenameDir() throws Exception {
```
rakeshadr, line 74: Appreciate if you could check the feasibility of adding these new rename test cases to `TestRootedOzoneFileSystem` something similar to TestOzoneFileSystem? 

If you have already tried, please point me to the test failures. Thanks!

rakeshadr (commented):
Thanks you @sadanand48 for taking care this patch. Overall the patch looks really nice. Added two minor comments, please go through it.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -1276,7 +1296,7 @@ public void testTrash() throws Exception {
     GenericTestUtils.waitFor(
         () -> getOMMetrics().getNumTrashDeletes() > prevNumTrashDeletes
             && getOMMetrics().getNumTrashFilesDeletes()
-            > prevNumTrashFileDeletes, 100, 180000);
+            >= prevNumTrashFileDeletes, 100, 180000);
```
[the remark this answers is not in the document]
rakeshadr: Its taken care in new commit. Thanks @sadanand48

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystem.java
```diff
@@ -1257,8 +1276,9 @@ public void testTrash() throws Exception {
     // This condition should pass after the checkpoint
     Assert.assertTrue(getOMMetrics()
         .getNumTrashRenames() > prevNumTrashRenames);
+    // With new layout version, file renames wouldn't be counted
```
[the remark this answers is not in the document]
rakeshadr: Thanks for addressing the comment based on offline discussion.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/KeyManagerImpl.java
sadanand48: Removed this check.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestRootedOzoneFileSystemV1.java
sadanand48: Added tests to TestRootedOzoneFileSystem.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestReadRetries.java
sadanand48: Resolving this based on offline discussion

rakeshadr (approved):
+1 LGTM, thanks @sadanand48 for the good work!

PR #1970, opened 2021-03-01
HDDS-4879. Support reserved space for single dir

runzhiwang:
HDFS support reserved space for all dirs: https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/fsdataset/impl/ReservedSpaceCalculator.java

But we also need to support reserved space for single dir, because different dir maybe need different reserved space, for example, maybe only one dir needs reserved space.

https://issues.apache.org/jira/browse/HDDS-4879

## How was this patch tested?

new ut

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeInfo.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeInfo.java
@@ -21,8 +21,6 @@ package org.apache.hadoop.ozone.container.common.volume;
 import java.io.File;
 import java.io.IOException;
 import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
 
 import org.apache.hadoop.fs.StorageType;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeInfo.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeInfo.java
@@ -20,6 +20,9 @@ package org.apache.hadoop.ozone.container.common.volume;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.apache.hadoop.fs.StorageType;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
@@ -29,6 +32,8 @@ import org.apache.hadoop.hdds.fs.SpaceUsageCheckParams;
 import com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.HDDS_DATANODE_DIR_DU_RESERVED;
 
 /**
  * Stores information about a disk/volume.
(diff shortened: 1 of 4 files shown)


ferhui (approved):
+1

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeInfo.java
```diff
@@ -83,6 +90,18 @@ public VolumeInfo build() throws IOException {
     }
   }
 
+  private long getReserved(ConfigurationSource conf) {
+    Collection<String> reserveList = conf.getTrimmedStringCollection(
+        HDDS_DATANODE_DIR_DU_RESERVED);
+    Map<String, Long> reserveMap = new HashMap<>();
```
adoroszlai, line 96: Nit: building the map of all directories is unnecessary, since only the value for `rootDir` is used.

runzhiwang:
@adoroszlai Thanks for review. I have updated the patch.

ferhui:
@adoroszlai could you please review this again? Thanks
If no other comments, will merge this one or two days later.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeInfo.java
```diff
@@ -83,6 +88,24 @@ public VolumeInfo build() throws IOException {
...
   }
 
+  private long getReserved(ConfigurationSource conf) {
+    Collection<String> reserveList = conf.getTrimmedStringCollection(
+        HDDS_DATANODE_DIR_DU_RESERVED);
+    for (String reserve : reserveList) {
+      String[] words = reserve.split(":");
+      if (words.length < 2) {
+        throw new IllegalArgumentException(
+            "Reserved space should config in pair");
```
adoroszlai, line 98: Throwing `IllegalArgumentException` instead of the `ArrayIndexOutOfBoundsException` still shuts down datanode during startup.

I think if the config is invalid, it should be ignored with a warning log message.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeInfo.java
```diff
@@ -83,6 +88,24 @@ public VolumeInfo build() throws IOException {
...
+      }
+
+      if (words[0].trim().equals(rootDir)) {
+        return Long.parseLong(words[1].trim());
```
adoroszlai, line 102: This throws exception if the specified value is not a number.  Also, what about negative values?

Now reserved space needs to be bytes.  Can we add support for storage units (eg. KB, MB, etc.)?  `StorageSize.parse` may be helpful here.

adoroszlai (commented):

adoroszlai (approved):

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/VolumeInfo.java
```diff
@@ -83,6 +89,31 @@ public VolumeInfo build() throws IOException {
...
+        HDDS_DATANODE_DIR_DU_RESERVED);
+    for (String reserve : reserveList) {
+      String[] words = reserve.split(":");
+      if (words.length < 2) {
+        LOG.error("Reserved space should config in pair, but current is {}",
+            reserve);
+        continue;
+      }
+
+      if (words[0].trim().equals(rootDir)) {
```
amaliujia, line 103: Do we need to LOG something when `words[0].trim().equals(rootDir)` is not true?

runzhiwang: We do not need to LOG, because not equal always happen.

ferhui:
@runzhiwang Thanks for contribution! @adoroszlai Thanks for review!
Merged

PR #1980, opened 2021-03-02
HDDS-4651: Distributed Sequence ID Gen.

GlenGeng:

Monotonically Increased ID Supporting Switch Between Non-Ratis SCM and Ratis-Based SCM.
1. SCM allocates ids in a batch way. It maintains two fields: lastId and nextId. It saves the lastId in rocksDB. The initial value of nextId is firstId.
2. When getNextId() is called, if nextId is less than lastId, SCM just returns nextId and increases it, otherwise it calls allocateBatch(expectedLastId, newLastId) to allocate a batch of ids, then serves the request.
3. In allocateBatch(expectedLastId, newLastId) , it works in CAS way:
  it reads lastId from rocksDB.
  if lastId equals expectedLastId, it saves newLastId into rocksDB, returns success, otherwise it rejects the allocation request.
  It also provides a getLastId() to read lastId from rocksDB. The allocation works in a loop way:

See https://docs.google.com/document/d/13qSwLuDBw7mytJtF9h2oauuLocFfr-XJSBdTfc1YvxM/edit?usp=sharing

https://issues.apache.org/jira/browse/HDDS-4651

## How was this patch tested?

CI


GlenGeng:
@mukul1987  @bshashikant @amaliujia  Please take look!
Monotonically Increased ID Supporting Switch Between Non-Ratis SCM and Ratis-Based SCM.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/io/BooleanCodec.java
```diff
@@ -0,0 +1,34 @@
...
+ * the License.
+ */
+
+package org.apache.hadoop.hdds.scm.ha.io;
+
+import com.google.protobuf.ByteString;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class BooleanCodec implements Codec {
```
amaliujia, line 24: Why do you need BoloeanCodec? Is it used?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/TestSequenceIDGen.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.hadoop.ozone.container.common.SCMTestUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_HA_ENABLE_KEY;
+import static org.apache.hadoop.hdds.scm.ScmConfigKeys.OZONE_SCM_SEQUENCE_ID_BATCH_SIZE;
+
+public class TestSequenceIDGen {
+  @Test
+  public void testSequenceIDGenUponNonRatis() throws Exception {
```
amaliujia, line 31: Do we have SCM HA in MiniOzoneCluster now? If so it will be great to have a leader switching tests?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/io/BooleanCodec.java
GlenGeng: We have this method, thus we need this `BooleanCodec`
```
@Replicate
Boolean allocateBatch(Long expectedLastId, Long newLastId);
```

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/ha/TestSequenceIDGen.java
GlenGeng: Agree. Will create a jira this. It is a good newbie job.

GlenGeng (commented):
Thanks @amaliujia for the review!

bshashikant (commented):
The changes look good with few minor comments. I have few questions which are not very clear from the doc.

1) Do we need to handle overflow for "long" used here for lastId?

2) With these approach, we won't have monotonically increasing containerIds or block ids. IDs will be monotonically increasing within a batch.

For Container Id, why can't we use (lastAppliedIndex + 1) from Ratis in case of HA? In NON-HA case, a monotonically increasing container Id will suffice .
For BlockId, we can use the distributed sequence Id generator

3) We need to handle the upgrade scenario. This needs to be handled before the branch gets merged otherwise all existing setups won't work.

What is the plan here to handle the same.?

GlenGeng:
> Do we need to handle overflow for "long" used here for lastId?

Thanks for the advice, I've add the check.

> With these approach, we won't have monotonically increasing containerIds or block ids. IDs will be monotonically increasing within a batch.

Could you find a situation that containerId or blockId is not  monotonically increasing ? For non-HA mode, it is straightword that the id is monotonically increasing. For HA mode, since only leader can request a nextID, and leader will always refresh lastId from db and allocate a new batch, a new leader will always generate larger id than previous leader, thus monotonically increasing too.

> We need to handle the upgrade scenario. This needs to be handled before the branch gets merged otherwise all existing setups won't work.

We have a discussion in the design doc: 
> need a hook to be used in upgrade, to ensure the monotonicity of id during switch from previous ids to current solution:
localId : the initial value should be UniqueId() + 1
containerID: scan the container db, figure the largest containerID ever used.
delTx: read from DeletedBlocksTXTable, see HDDS-4477 contributed by Lokesh.

So, we can simply use `UniqueId() + 1`, since containerID and delTx can not be larger than `UniqueId() + 1`.

bshashikant:
> > Do we need to handle overflow for "long" used here for lastId?
> 
> Thanks for the advice, I've add the check.
> 
> > With these approach, we won't have monotonically increasing containerIds or block ids. IDs will be monotonically increasing within a batch.
> 
> Could you find a situation that containerId or blockId is not monotonically increasing ? For non-HA mode, it is straightword that the id is monotonically increasing. For HA mode, since only leader can request a nextID, and leader will always refresh lastId from db and allocate a new batch, a new leader will always generate larger id than previous leader, thus monotonically increasing too.

What i meant was, containerIds are not sequentially increasing which is a departure from the current behaviour which is strictly sequentially increasing containerIds. and we are using the same seqID generator for both Container as well as block.
This may become the reason for scale limitation.

Instead of that, why can't we use lastAppliedIndex as base for containerIds.
> 
> > We need to handle the upgrade scenario. This needs to be handled before the branch gets merged otherwise all existing setups won't work.
> 
> We have a discussion in the design doc:
> 
> > need a hook to be used in upgrade, to ensure the monotonicity of id during switch from previous ids to current solution:
> > localId : the initial value should be UniqueId() + 1
> > containerID: scan the container db, figure the largest containerID ever used.
> > delTx: read from DeletedBlocksTXTable, see HDDS-4477 contributed by Lokesh.
> 
> So, we can simply use `UniqueId() + 1` as `INVALID_SEQUENCE_ID`, since containerID and delTx are starting from 1, practically they can not be larger than `UniqueId() + 1`.

Are u going to address these changes in subsequent PR as this will be required for branch merge?

bshashikant:
> > > Do we need to handle overflow for "long" used here for lastId?
> > 
> > 
> > Thanks for the advice, I've add the check.
> > > With these approach, we won't have monotonically increasing containerIds or block ids. IDs will be monotonically increasing within a batch.
> > 
> > 
> > Could you find a situation that containerId or blockId is not monotonically increasing ? For non-HA mode, it is straightword that the id is monotonically increasing. For HA mode, since only leader can request a nextID, and leader will always refresh lastId from db and allocate a new batch, a new leader will always generate larger id than previous leader, thus monotonically increasing too.
> 
> What i meant was, containerIds are not sequentially increasing which is a departure from the current behaviour which is strictly sequentially increasing containerIds. and we are using the same seqID generator for both Container as well as block.
> This may become the reason for scale limitation.
> 
> Instead of that, why can't we use lastAppliedIndex as base for containerIds? Sequence Ids can be used for block Ids and can be rest to zero once all the containers are closed.
> 
> > > We need to handle the upgrade scenario. This needs to be handled before the branch gets merged otherwise all existing setups won't work.
> > 
> > 
> > We have a discussion in the design doc:
> > > need a hook to be used in upgrade, to ensure the monotonicity of id during switch from previous ids to current solution:
> > > localId : the initial value should be UniqueId() + 1
> > > containerID: scan the container db, figure the largest containerID ever used.
> > > delTx: read from DeletedBlocksTXTable, see HDDS-4477 contributed by Lokesh.
> > 
> > 
> > So, we can simply use `UniqueId() + 1` as `INVALID_SEQUENCE_ID`, since containerID and delTx are starting from 1, practically they can not be larger than `UniqueId() + 1`.
> 
> Are u going to address these changes in subsequent PR as this will be required for branch merge?

bshashikant (approved):

bshashikant:
The upgrade work will be handled in a different PR. Thanks @GlenGeng for the work.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIdGenerator.java
```diff
@@ -0,0 +1,308 @@
...
+
+      public Builder setDBTransactionBuffer(DBTransactionBuffer trxBuffer) {
+        buffer = trxBuffer;
+        return this;
+      }
+
+      public StateManager build() {
+        Preconditions.checkNotNull(table);
+        Preconditions.checkNotNull(buffer);
+        Preconditions.checkNotNull(ratisServer);
```
GlenGeng, line 247: @bshashikant Please take care.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIdGenerator.java
```diff
@@ -0,0 +1,308 @@
...
+          });
+
+      if (!lastId.equals(expectedLastId)) {
+        LOG.warn("Failed to allocate a batch for {}, expected lastId is {}," +
+            " actual lastId is {}.", sequenceIdName, expectedLastId, lastId);
+        return false;
+      }
+
+      try {
+        sequenceIdTable.putWithBatch(transactionBuffer
```
GlenGeng, line 205: @bshashikant Please take care.

PR #1981, opened 2021-03-02
HDDS-4890. SCM Ratis enable/disable switch

bshashikant:

Added SCM Ratis enable/disable switch

https://issues.apache.org/jira/browse/HDDS-4890

## How was this patch tested?

Existing unit tests should work with the patch on and off.


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBTransactionBufferImpl.java
```diff
@@ -0,0 +1,34 @@
...
+public class SCMDBTransactionBufferImpl implements DBTransactionBuffer {
+
+  public SCMDBTransactionBufferImpl() {
+
+  }
+
+  @Override
+  public void addToBuffer(Table table, Object key, Object value)
+      throws IOException {
+    table.put(key, value);
```
GlenGeng, line 21: Quite straightforward, I like it.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/MockSCMHAManager.java
```diff
@@ -204,32 +199,29 @@ private Message process(final SCMRatisRequest request)
...
       return null;
     }
 
-    @Override
-    public List<String> getRatisRoles() {
-      return Arrays.asList(
-          "180.3.14.5:9865",
-          "180.3.14.21:9865",
-          "180.3.14.145:9865");
+    @Override public List<String> getRatisRoles() {
```
GlenGeng, line 209: The code style is weird, why put override and function name in the same line ?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBTransactionBufferImpl.java
amaliujia: Actually very nice! Thanks!

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/block/TestDeletedBlockLog.java
```diff
@@ -85,23 +86,24 @@
...
     conf = new OzoneConfiguration();
+    conf.setBoolean(ScmConfigKeys.OZONE_SCM_HA_ENABLE_KEY, true);
     conf.setInt(OZONE_SCM_BLOCK_DELETION_MAX_RETRY, 20);
     conf.set(HddsConfigKeys.OZONE_METADATA_DIRS, testDir.getAbsolutePath());
     scm = TestUtils.getScm(conf);
     containerManager = Mockito.mock(ContainerManagerV2.class);
-    dbTransactionBuffer =
-        new MockDBTransactionBuffer(scm.getScmMetadataStore().getStore());
+    SCMHADBTransactionBuffer =
+        new MockSCMHADBTransactionBuffer(scm.getScmMetadataStore().getStore());
```
amaliujia, line 102: Is it possible to implement "flush on every write" in `MockSCMHADBTransactionBuffer` so in tests we no longer need to call flush()?

amaliujia: Or, even better, `MockSCMHADBTransactionBuffer` just pass through the buffer and write to table directly?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/metadata/DBTransactionBuffer.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * DB transaction that abstracts the updates to the underlying datastore.
+ */
+public interface DBTransactionBuffer<KEY, VALUE> extends Closeable {
+
+  void addToBuffer(Table<KEY, VALUE> table, KEY key, VALUE value) throws
+      IOException;
+
+  void removeFromBuffer(Table<KEY, VALUE> table, KEY key) throws
```
GlenGeng, line 33: generic method may be better than generic class here. We can fix it later if it is too complicated.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
```diff
@@ -170,8 +169,7 @@ public void removeTransactionsFromDB(ArrayList<Long> txIDs)
       throws IOException {
     deletingTxIDs.addAll(txIDs);
     for (Long txID : txIDs) {
-      deletedTable.deleteWithBatch(
-          transactionBuffer.getCurrentBatchOperation(), txID);
+      transactionBuffer.removeFromBuffer(deletedTable, txID);
```
GlenGeng, line 172: Should not call `deletingTxIDs.addAll(txIDs);` if ratis is not enabled.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/MockSCMHADBTransactionBuffer.java
```diff
@@ -20,22 +20,22 @@
...
+  public MockSCMHADBTransactionBuffer() {
   }
 
-  public MockDBTransactionBuffer(DBStore store) {
+  public MockSCMHADBTransactionBuffer(DBStore store) {
     this.dbStore = store;
   }
 
-  @Override
   public BatchOperation getCurrentBatchOperation() {
```
GlenGeng, line 39: change getCurrentBatchOperation  to private.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAInvocationHandler.java
```diff
@@ -50,16 +50,20 @@ public SCMHAInvocationHandler(final RequestType requestType,
...
   }
 
   @Override
   public Object invoke(final Object proxy, final Method method,
                        final Object[] args) throws Throwable {
     try {
       long startTime = Time.monotonicNow();
-      final Object result = method.isAnnotationPresent(Replicate.class) ?
-          invokeRatis(method, args) : invokeLocal(method, args);
+      final Object result =
```
GlenGeng, line 63: Please add a Preconditions.checkNotNull for ratisHandler at invokeRatis too. Thanks!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAManagerImpl.java
```diff
@@ -85,25 +94,27 @@ public SCMHAManagerImpl(final ConfigurationSource conf,
...
-      // It will first try to add itself to existing ring
-      boolean success = HAUtils.addSCM(OzoneConfiguration.of(conf),
-          new AddSCMRequest.Builder().setClusterId(scm.getClusterId())
-              .setScmId(scm.getScmId()).setRatisAddr(
-              scm.getSCMHANodeDetails().getLocalNodeDetails()
-                  // TODO : Should we use IP instead of hostname??
-                  .getRatisHostPortStr()).build(), scm.getSCMNodeId());
-      if (!success) {
-        throw new IOException("Adding SCM to existing HA group failed");
+    if (ratisServer != null) {
```
GlenGeng, line 97: avoid too deep indent.
How about 

```
  if (ratisServer == null) {
    return
  }

  // original code

```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -454,10 +454,11 @@ private void initializeSystemManagers(OzoneConfiguration conf,
     if (configurator.getScmContext() != null) {
       scmContext = configurator.getScmContext();
     } else {
+      long term = SCMHAUtils.isSCMHAEnabled(conf) ? 0 : SCMContext.INVALID_TERM;
```
GlenGeng, line 457: Good !
Better add a comment here.
```
// When term equals SCMContext.INVALID_TERM, the isLeader() and getTermOfLeader() will always pass.
```

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/TestUtils.java
```diff
@@ -476,7 +476,7 @@ public static void quasiCloseContainer(ContainerManagerV2 containerManager,
   public static StorageContainerManager getScmSimple(OzoneConfiguration conf)
       throws IOException, AuthenticationException {
     SCMConfigurator configurator = new SCMConfigurator();
-    conf.setBoolean(ScmConfigKeys.OZONE_SCM_HA_ENABLE_KEY, true);
+    // conf.setBoolean(ScmConfigKeys.OZONE_SCM_HA_ENABLE_KEY, true);
```
GlenGeng, line 479: I guess, this is the flag to enable ratis SCM in test ? Better add a comment, since it is easy to overlook.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/block/TestDeletedBlockLog.java
GlenGeng: Agree. Can `MockSCMHADBTransactionBuffer` share the same logic `addToBuffer/removeFromBuffer` as `SCMDBTransactionBufferImpl`, and let its `flush` to throw exception ?

bshashikant: I think, MockInstances should mimic every functional behaviour of the class. I would prefer to keep it as it is for now. We should be able to test the batch behaviour of SCMHADBTransaction behaviour with this.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
```diff
@@ -68,12 +68,7 @@
 
 import java.io.IOException;
 import java.net.InetSocketAddress;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
```
GlenGeng, line 71: expand it ?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/block/TestDeletedBlockLog.java
```diff
@@ -163,26 +165,24 @@ public void tearDown() throws Exception {
 
   private void addTransactions(Map<Long, List<Long>> containerBlocksMap)
       throws IOException {
-    dbTransactionBuffer.getCurrentBatchOperation();
+    //SCMHADBTransactionBuffer.getCurrentBatchOperation();
```
GlenGeng, line 168: remove it ?

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/block/TestDeletedBlockLog.java
amaliujia: I see. Sure it sounds good.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/MockSCMHADBTransactionBuffer.java
amaliujia: +1

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/metadata/DBTransactionBuffer.java
bshashikant: Will address it in a different PR.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
```diff
@@ -193,14 +198,19 @@ public void increaseRetryCountOfTransactionInDB(
...
+        skippingRetryTxIDs.add(txID);
+      }
     }
   }
 
 
   public void onFlush() {
-    deletingTxIDs.clear();
-    skippingRetryTxIDs.clear();
+    if (deletingTxIDs != null) {
```
GlenGeng, line 209: Better to be
```
// onFlush() can be invoked only when ratis is enabled.
Preconditions.checkNotNull(deletingTxIDs);
Preconditions.checkNotNull(skippingRetryTxIDs);
deletingTxIDs.clear();
skippingRetryTxIDs.clear();
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
```diff
@@ -89,8 +92,8 @@ private void findNext() {
             throw new IllegalStateException("");
           }
 
-          if (!deletingTxIDs.contains(txID) &&
-              !skippingRetryTxIDs.contains(txID)) {
+          if (deletingTxIDs != null && !deletingTxIDs.contains(txID) &&
```
GlenGeng, line 95:
```
 (deletingTxIDs == null || !deletingTxIDs.contains(txID)) &&
 (skippingRetryTxIDs == null || !skippingRetryTxIDs.contains(txID))
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -175,9 +175,9 @@ public SCMRatisResponse submitRequest(SCMRatisRequest request)
         .build();
     final RaftClientReply raftClientReply =
         server.submitClientRequestAsync(raftClientRequest).get();
-    if (LOG.isDebugEnabled()) {
-      LOG.debug("request {} Reply {}", raftClientRequest, raftClientReply);
-    }
+   // if (LOG.isDebugEnabled()) {
```
GlenGeng, line 178: this should be a temporary change ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/DeletedBlockLogStateManagerImpl.java
bshashikant: Thanks for pointing out. Will address in the next patch.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/block/TestDeletedBlockLog.java
GlenGeng: One tricky thing is that, recon uses `MockSCMHAManager` and `MockSCMHADBTransactionBuffer`.
Since Recon will not use ratis, `flush` is not called, thus there will finally be OOM for recon.

```
Ctor of ReconStorageContainerManagerFacade
    this.scmhaManager = MockSCMHAManager.getInstance(
        true, new MockSCMHADBTransactionBuffer(dbStore));

```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIdGenerator.java
```diff
@@ -202,8 +203,8 @@ public Boolean allocateBatch(String sequenceIdName,
       }
 
       try {
-        sequenceIdTable.putWithBatch(transactionBuffer
-            .getCurrentBatchOperation(), sequenceIdName, newLastId);
+        transactionBuffer
+            .addToBuffer(sequenceIdTable, sequenceIdName, newLastId);
```
GlenGeng, line 207: You need also remove `Preconditions.checkNotNull(ratisServer);` in line 248

GlenGeng: My fault. It should be fine. non-ratis SCM will not go to there.

GlenGeng:
Thanks @bshashikant  for the work! 

One more thing about recon, `MockSCMHAManager` and `SCMHADBTransactionBuffer` is used by `ReconStorageContainerManagerFacade`, as
```
    this.scmhaManager = MockSCMHAManager.getInstance(
        true, new MockSCMHADBTransactionBuffer(dbStore));
```

Need replace `MockSCMHADBTransactionBuffer ` with `SCMDBTransactionBufferImpl`, otherwise recon will OOM, since flush is called.

GlenGeng:
+1. Thanks @bshashikant for the contribution.

bshashikant:
Thanks @GlenGeng for the review. I have committed this.

PR #1983, opened 2021-03-02
HDDS-4891. Avoid latest/master in Github Actions

adoroszlai:

Avoid versions `latest` (for runners) and `master` (for actions) in Github Actions workflows.

https://issues.apache.org/jira/browse/HDDS-4891

## How was this patch tested?

Tested a command comment on my fork:
https://github.com/adoroszlai/hadoop-ozone/pull/8#issuecomment-788974590
https://github.com/adoroszlai/hadoop-ozone/runs/2014250387

### DIFF
--- a/.github/workflows/close-pending.yaml
+++ b/.github/workflows/close-pending.yaml
@@ -21,10 +21,10 @@ on:
 jobs:
   close-pending:
     name: close-pending
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     steps:
       - name: Checkout code
-        uses: actions/checkout@master
+        uses: actions/checkout@v2
       - name: Execute close-pending script
         if: github.repository == 'apache/ozone'
         run: ./.github/close-pending.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/close-pending.yaml
```diff
@@ -21,10 +21,10 @@ on:
 jobs:
   close-pending:
     name: close-pending
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
```
ayushtkn, line 24: Any specific reason for using 18.04? I guess 20.04 is also LTS

adoroszlai: `latest` [is currently 18.04](https://github.com/apache/ozone/runs/2020674273?check_suite_focus=true#step:1:4), but [will be updated to 20.04](https://github.com/actions/virtual-environments/issues/1816) sometime soon.  I don't expect any problem from the update.  But changing to a more specific version lets us make the upgrade when required in a controlled way, instead of getting it unsolicitedly.

ayushtkn: makes sense

elek (approved):
+1 good idea to pin all the versions

Thanks the patch @adoroszlai

adoroszlai:

PR #1984, opened 2021-03-02
HDDS-4888. Merge basic CI checks

adoroszlai:

1. Merge basic CI checks (`author`, `bats`, etc.) into a single check with matrix build.  This reduces code duplication.
2. ~Add scripts to install check-specific dependencies (eg. `bats` or `spotbugs`).  This reduces complexity of Github-specific code in the workflow definition, and allows simpler prototyping/testing.~ (Extracted to HDDS-4969.)
3. Use Github Actions cache instead of building inside `ozone-build` container.  This reduces the effect of outdated dependencies in the image.
4. Prefer check-specific cache because:
   * various checks require slightly different dependencies (mostly related to Maven plugins)
   * concurrent checks may have problem sharing cache (noticed with _compile (8)_ and _compile (11)_ checks)
   If check-specific cache is not available, fallback to shared cache at restore time.

https://issues.apache.org/jira/browse/HDDS-4888

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/actions/runs/732631557


hadoop-ozone/dev-support/checks/install/_install.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
elek, line 1: Tricky name, as this script is responsible both for install and executing the scripts.

I am wondering if it would be easier to do a laze download at the beginning of each of the scripts. For example at the beginning of the findbugs.sh we can check if spotbugs exists and download if not.

elek (commented):
Thanks, the patch @adoroszlai. Overall I like all the suggested items.

I have some proposal to the tools downloading part:

For me, it seems to be easier to lazy-download the required tools from the beginning of each script.

We can also use a specific directory under ozone for all the tools. For example `./.dev-tools` which can contain subdirectories like spotbugs, bats, ...

With this approach:

 1. we can cache the `./.dev-tools` directory itself which can make next builds faster
 2. `./.dev-tools` can be populated during the local script executions. Easy to run any test as required tool will be downloaded (and persisted) first time.
 
 But it's just an idea, I am fine with the current approach as a first step, just interested what do you think...

hadoop-ozone/dev-support/checks/install/_install.sh
adoroszlai: > Tricky name, as this script is responsible both for install and executing the scripts.

This is just a wrapper to invoke a specific install script.  It does not run the check.  The only reason for its existence: the workflow file can blindly invoke it without checking for the existence of the specific install script (I'm not even sure it would be possible).

adoroszlai:
Thanks @elek for the review.  I like the idea of integrated lazy download.  I would make it optional (and enable in CI), as developers may have their own preferences for installing tools (eg. via package manager).

adoroszlai:
> With this approach:
> 
> 1. we can cache the `./.dev-tools` directory itself which can make next builds faster
> 2. `./.dev-tools` can be populated during the local script executions. Easy to run any test as required tool will be downloaded (and persisted) first time.

Unfortunately this does not work for the _kubernetes_ check.  K3s installs stuff to global directories (`/etc`).

It also does not apply to Robot Framework installed via `pip`.

elek:
> Unfortunately this does not work for the kubernetes check. K3s installs stuff to global directories (/etc).
> 
> It also does not apply to Robot Framework installed via pip.

robotframework can be installed with virtualenv + pip to local directory (As far as I remember it worked for me, but let me know if it doesn't)

Kubernetes is a more tricky question:

flekszible can be installed to local dir

I think the installation of k3s can remain in the github actions (if possible). K3s is just one way to provide kubernetes cluster. I think `hadoop-ozone/dev-support/checks/kubernetes.sh` should assume that we have working kubernetes install (similar that we assume that java and maven are available, we don't install them).

I can imagine that someone would like to execute tests on minikube (or any other one node local k8s cluster).

In github action environment we know that k3s should be installed, but locally it's not so straightforward, therefore I would skip the installation from the script. (We can print out a warning and recommend k3s installation if kubectl is not on the path)

elek (approved):
+1. Thanks the update (and the patch) @adoroszlai

adoroszlai:
Thanks @elek for the review.

PR #1986, opened 2021-03-03
HDDS-4873. Re-replication failure throws NPE

sky76093016:

Increase the log that helps users to use.

https://issues.apache.org/jira/browse/HDDS-4873

## How was this patch tested?

No test.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/DownloadAndImportReplicator.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/DownloadAndImportReplicator.java
@@ -113,8 +113,8 @@ public class DownloadAndImportReplicator implements ContainerReplicator {
       task.setStatus(Status.FAILED);
     } else {
       try {
-        //wait for the download. This thread pool is limiting the parallel
-        //downloads, so it's ok to block here and wait for the full download.
+        // Wait for the download. This thread pool is limiting the parallel
+        // downloads, so it's ok to block here and wait for the full download.
         Path path = tempTarFile.get();
         long bytes = Files.size(path);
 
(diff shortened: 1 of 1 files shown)


sky76093016:
@jojochuang Can you review this PR for me?Thank you.

adoroszlai (changes requested):
Thanks @sky76093016 for working on this.

We should not simply log "please retry one more time", as container replication is performed by datanodes automatically.  It will be retried eventually when datanode receives the replication command from SCM again.

If download fails currently the code reaches the exception handler with NPE.  The only problem with this is that it's noisy, and we could handle the condition more gracefully.  I think if `tempTarFile == null`, it should simply set task status to failed, and skip the code in the `try` block.

jojochuang:
what Attila said. Thanks for the review!

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/DownloadAndImportReplicator.java
```diff
@@ -109,23 +109,26 @@ public void replicate(ReplicationTask task) {
...
-      task.setTransferredBytes(bytes);
-
-      importContainer(containerID, path);
-      LOG.info("Container {} is replicated successfully", containerID);
-      task.setStatus(Status.DONE);
-    } catch (Exception e) {
-      LOG.error("Container {} replication was unsuccessful.", containerID, e);
+    if (tempTarFile == null) {
       task.setStatus(Status.FAILED);
+    }else{
```
adoroszlai, line 114:
```suggestion
    } else {
```

adoroszlai (approved):
Thanks @sky76093016 for updating the patch.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/DownloadAndImportReplicator.java
```diff
@@ -109,23 +109,26 @@ public void replicate(ReplicationTask task) {
...
-      LOG.info("Container {} is replicated successfully", containerID);
-      task.setStatus(Status.DONE);
-    } catch (Exception e) {
-      LOG.error("Container {} replication was unsuccessful.", containerID, e);
+    if (tempTarFile == null) {
       task.setStatus(Status.FAILED);
+    } else {
+      try {
+        //wait for the download. This thread pool is limiting the parallel
+        //downloads, so it's ok to block here and wait for the full download.
```
cku328, line 117:
```suggestion
        // Wait for the download. This thread pool is limiting the parallel
        // downloads, so it's ok to block here and wait for the full download.
```
The comments in code should begin with a whitespace after the leading forward slashes.

cku328 (approved):
LGTM. Thanks @sky76093016 for working on this.

PR #1988, opened 2021-03-03
HDDS-4897. [SCM HA Security] Create SCM CertClient and use it during startup and bootStrap

bharatviswa504:

This Jira is to implement
1. Create CertClient, which generates a public key, private key, and generates CSR with ClusterID, SCM1ID.
2. Use SCM SecurityFailOverProxyProvider to getSCMCertificate and RootCA certificate and store it to a local disk.

https://issues.apache.org/jira/browse/HDDS-4897

## How was this patch tested?
Need changes in DefaultProfile to make this work, as right now BasicConstraint is not validated in DefaultProfile, so when the new RootCA server starts, will use a new profile/make changes to make the CSR and request certificate work. Will add tests in further jira.


bharatviswa504:
Closing this, will be implemented as part of HDDS-4915.

PR #1992, opened 2021-03-04
HDDS-4901. Remove OmOzoneAclMap from OmVolumeArgs to avoid OzoneAcl c…

xiaoyuyao:
…onversions.

Remove OmOzoneAclMap to consolidate acl operation among volume/bucket/prefix

https://issues.apache.org/jira/browse/HDDS-4901

## How was this patch tested?

Existing UT


xiaoyuyao:
The test timeout failure TestNode2PipelineMap seems unrelated.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/protocol/ClientProtocol.java
```diff
@@ -124,6 +124,7 @@ OzoneVolume getVolumeDetails(String volumeName)
    * @return Boolean - True if the user with a role can access the volume.
    * This is possible for owners of the volume and admin users
    * @throws IOException
+   * TODO: deprecate and remove this API.
```
ChenSammi, line 127: Shall we add a @Deprecated annotation on this API?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/volume/OMVolumeCreateRequest.java
```diff
@@ -117,21 +116,17 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
-      // The Object ID will never change, but update
-      // ID will be set to transactionID each time we update the object.
-      omVolumeArgs.setObjectID(
-          ozoneManager.getObjectIdFromTxId(transactionLogIndex));
-      omVolumeArgs.setUpdateID(transactionLogIndex,
-          ozoneManager.isRatisEnabled());
-
-
-      auditMap = omVolumeArgs.toAuditMap();
+    OmVolumeArgs omVolumeArgs = OmVolumeArgs.getFromProtobuf(volumeInfo);
```
ChenSammi, line 119: getFromProtobuf has the chance to throw exceptions, so shall we keep into the try scope?

ChenSammi:
@xiaoyuyao , thanks for consolidating native ACL logics. The patch overall looks good. 

Besides some inline comments,  there is "TestOmOzoneAclMap.java".  
This UT is introduced in HDDS-4606, which fixed the issue that duplicate default ACLs are added on volume.  I don't find existing UT which covers this case.  So can you refine the TestOmOzoneAclMap.java to keep it, or add a new duplicate default ACL check UT somewhere in existing test class?

xiaoyuyao:
> @xiaoyuyao , thanks for consolidating native ACL logics. The patch overall looks good.
> 
> Besides some inline comments, there is "TestOmOzoneAclMap.java".
> This UT is introduced in [HDDS-4606](https://issues.apache.org/jira/browse/HDDS-4606), which fixed the issue that duplicate default ACLs are added on volume. I don't find existing UT which covers this case. So can you refine the TestOmOzoneAclMap.java to keep it, or add a new duplicate default ACL check UT somewhere in existing test class?

Good catch. I've ported the test in the update.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/protocol/ClientProtocol.java
xiaoyuyao: Added.

ChenSammi:
The last patch LGTM.   @xiaoyuyao would you check the failed acceptance.  They seems irrelevant.

xiaoyuyao:
@ChenSammi It seems to be some transient timeout issue. The latest run is clean.

ChenSammi:
+1.  Thanks @xiaoyuyao.

PR #1993, opened 2021-03-05
HDDS-4902 Add Layout version information to Recon datanode info API.

avijayanhwx:
Simple change to add Layout version information to Recon datanode info API.

https://issues.apache.org/jira/browse/HDDS-4902

## How was this patch tested?
Manually tested.
Added unit test.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
@@ -540,37 +540,35 @@ public class SCMNodeManager implements NodeManager {
           layoutVersionReport.toString().replaceAll("\n", "\\\\n"));
     }
 
-    if (layoutVersionReport != null) {
-      int scmSlv = scmLayoutVersionManager.getSoftwareLayoutVersion();
-      int scmMlv = scmLayoutVersionManager.getMetadataLayoutVersion();
-      int dnSlv = layoutVersionReport.getSoftwareLayoutVersion();
-      int dnMlv = layoutVersionReport.getMetadataLayoutVersion();
+    int scmSlv = scmLayoutVersionManager.getSoftwareLayoutVersion();
+    int scmMlv = scmLayoutVersionManager.getMetadataLayoutVersion();
+    int dnSlv = layoutVersionReport.getSoftwareLayoutVersion();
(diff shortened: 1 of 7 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -540,37 +540,35 @@ public void processLayoutVersionReport(DatanodeDetails datanodeDetails,
...
-            datanodeDetails.getHostName(), dnSlv, scmSlv);
-      }
+    int scmSlv = scmLayoutVersionManager.getSoftwareLayoutVersion();
+    int scmMlv = scmLayoutVersionManager.getMetadataLayoutVersion();
+    int dnSlv = layoutVersionReport.getSoftwareLayoutVersion();
+    int dnMlv = layoutVersionReport.getMetadataLayoutVersion();
+
+    // If the data node slv is > scm slv => log error condition
+    if (dnSlv > scmSlv) {
+      LOG.error("Rogue data node in the cluster : {}. " +
```
swagle, line 550: Rouge is generally used in the security context, I'd suggest using Outdated or something similar.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/api/types/DatanodeMetadata.java
```diff
@@ -67,6 +67,9 @@
   @XmlElement(name = "buildDate")
   private String buildDate;
 
+  @XmlElement(name = "layoutVersion")
```
swagle, line 70: General question: Why is this POJO class replicated inside recon vs having just one version in hdds-common or somewhere higher.

avijayanhwx: You are correct, this was added as a Recon specific entity that may be formed by picking metadata pieces from multiple internal representations, and to support JAXB (I believe). We can revisit it and see if we can just reuse internal entities.

swagle: Make sense for another time but I think having all pojos in one place makes code so much more readable.

swagle (approved):

avijayanhwx:
Thanks for the review @swagle.

PR #1994, opened 2021-03-05
HDDS-4910. Cp command not working on Ozone ofs mounted using hdfs-fuse or hcfsfuse.	

captainzmc:

The question is same to: HDDS-3492

Now, we are able to support the cp operation of FUSE using o3fs via HDDS-3492. But ofs does not support, and so increase the PosixRootedOzoneFileSystem to make ofs support FUSE cp , just like what we did in HDDS-3492.

https://issues.apache.org/jira/browse/HDDS-4910

## How was this patch tested?

Used [hcfsfuse](https://github.com/opendataio/hcfsfuse) or hdfs-fuse to mount the OFS tested


maobaolong (approved):
Thanks for this feature, It LTGM

PR #1997, opened 2021-03-05
HDDS-4835. [FSO]S3Multipart: Implement UploadAbortRequest

rakeshadr:

This task is to implement prefix based FSO for the S3MultipartUploadAbortRequest.

https://issues.apache.org/jira/browse/HDDS-4835

## How was this patch tested?

Added UTs.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/s3/multipart/S3MultipartUploadAbortResponseV1.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/s3/multipart/S3MultipartUploadAbortResponseV1.java
@@ -18,6 +18,7 @@
 
 package org.apache.hadoop.ozone.om.response.s3.multipart;
 
+import org.apache.hadoop.hdds.utils.db.BatchOperation;
 import org.apache.hadoop.ozone.OmUtils;
 import org.apache.hadoop.ozone.om.OMMetadataManager;
 import org.apache.hadoop.ozone.om.helpers.OmBucketInfo;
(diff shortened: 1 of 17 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/response/s3/multipart/S3MultipartUploadAbortResponseV1.java
```diff
@@ -0,0 +1,100 @@
...
+    TreeMap<Integer, PartKeyInfo > partKeyInfoMap =
+        getOmMultipartKeyInfo().getPartKeyInfoMap();
+    for (Map.Entry<Integer, PartKeyInfo > partKeyInfoEntry :
+        partKeyInfoMap.entrySet()) {
+      PartKeyInfo partKeyInfo = partKeyInfoEntry.getValue();
+      OmKeyInfo currentKeyPartInfo =
+          OmKeyInfo.getFromProtobuf(partKeyInfo.getPartKeyInfo());
+
+      RepeatedOmKeyInfo repeatedOmKeyInfo =
+          omMetadataManager.getDeletedTable().get(partKeyInfo.getPartName());
```
linyiqun, line 85: Seems here addToDBBatch logic can completely reuse S3MultipartUploadAbortResponse#addToDBBatch. 
The only difference is that: In V1 layout, partName stored in part keyInfo is not a full part key name (now is filename+clientID, [link](https://github.com/apache/ozone/blob/HDDS-2939/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCommitPartRequestV1.java#L162)), but we should reconstruct the full part key name to find in delete table.

rakeshadr: Thanks a lot @linyiqun for the reviews. 

Very good point. I have made the changes. I've made few more assertions for the upload parts.

For the deleted table part entries, there is a unit test case and this patch extending TestS3MultipartUploadAbortResponseV1#testAddDBToBatchWithParts. Hope thats fine for you. [UT Reference](https://github.com/apache/ozone/blob/master/hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/response/s3/multipart/TestS3MultipartUploadAbortResponse.java#L133)

linyiqun: Thanks for addressing the comment, @rakeshadr .
The reference UT verified that the part key entries are correctly put into the delete table.
But the key name here might be a full part key name I think, as delete table format is like below:

```java
|----------------------------------------------------------------------|
| deletedTable       | /volumeName/bucketName/keyName->RepeatedKeyInfo |
|----------------------------------------------------------------------|
```

Now part name stored in omMultipartKeyInfo is filename+clientID, use this part name as a key name in delete table doesn't match above format.

Correct me If I am wrong.

rakeshadr: Thanks again for the detailed explanation. Good catch!

I have updated the partName logic and the partName format is now `<parentID>/fileName + ClientID`.

Since the `partKeyInfo.getPartName()` stored in DB is in the above format all the DB operations will happen using this dbPartName. So, no specific conversion required in response class. 

To the user after commit, it will return the` fullKeyPartName` . [Reference code:](https://github.com/apache/ozone/blob/HDDS-2939/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCommitPartRequestV1.java#L225).

linyiqun (approved):
LGTM, +1.
As we use the new format key  parentID/fileName + ClientID in delete table, suppose we might need to do some adjustment in current key delete service to support this type of key in follow-up task.

rakeshadr:
> LGTM, +1.
> As we use the new format key parentID/fileName + ClientID in delete table, suppose we might need to do some adjustment in current key delete service to support this type of key in follow-up task.

Sure, we have to work on the deletion keys.

Thanks once again @linyiqun for the continuous support by providing useful review comments. I will merge it.

PR #1999, opened 2021-03-08
HDDS-4896. Need a tool to upgrade current non-HA SCM node to single node HA cluster

bshashikant:
The idea here is to add  an optional config "ozone.scm.primordial.node.id".  if the config is set, it will cause scm --init to only take effect on
the specific node and ignore scm --bootstrap cmd. Similarly, scm --init will be ignored on the non-primordial scm nodes.

If the cluster is upgraded from non-ratis to ratis version , scm --init needs to re-run for switching from non-ratis based SCM to ratis-based SCM on the primary node.

https://issues.apache.org/jira/browse/HDDS-4896

## How was this patch tested?
Unit tests added

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
@@ -101,7 +101,7 @@ public class SCMRatisServerImpl implements SCMRatisServer {
       server = newRaftServer(scmId, conf).setGroup(group).build();
       server.start();
       // TODO: Timeout and sleep interval should be made configurable
-      waitforLeaderToBeReady(server, 60000, group);
+      waitForLeaderToBeReady(server, 60000, group);
     } finally {
       if (server != null) {
         server.close();
@@ -137,7 +137,7 @@ public class SCMRatisServerImpl implements SCMRatisServer {
     }
   }
 
-  private static void waitforLeaderToBeReady(RaftServer server, long timeout,
+  private static void waitForLeaderToBeReady(RaftServer server, long timeout,
       RaftGroup group)
       throws IOException {
     boolean ready;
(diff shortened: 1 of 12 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -103,6 +104,27 @@ public static void initialize(String clusterId, String scmId,
...
+      SCMNodeDetails details, OzoneConfiguration conf) throws IOException {
+    RaftServer server = newRaftServer(scmId, conf).build();
+    RaftGroup group = null;
+    Iterator<RaftGroup> iter = server.getGroups().iterator();
+    if (iter.hasNext()) {
+      group = iter.next();
+    }
+    // close the server instance so that pending locks on raft storage directory
+    // gets released if any.
+    server.close();
```
mukul1987, line 117: Lets wrap the earlier code in try and do the server close in finally block ?

mukul1987 (approved):

GlenGeng:
How about move `validateRatisGroupExists` to `TestStorageContainerManager`, since it is not used in production code any more.

GlenGeng:
One question about the optional config `ozone.scm.primordial.node.id`, In which situation will this flag be used ?

With this flag, people can apply init, bootstrap and start on all of the 3 SCMs, so that manual operation will be safer. Is this the  requirement ?

bshashikant:
> One question about the necessity of the optional config `ozone.scm.primordial.node.id`, In which situation will this flag be used ?
> 
> With this flag, people can apply init, bootstrap and start on all of the 3 SCMs, so that manual operation will be safer. Is this the requirement ?

yes. This is a requirement for Kubernetes setups .

GlenGeng:
> > One question about the necessity of the optional config `ozone.scm.primordial.node.id`, In which situation will this flag be used ?
> > With this flag, people can apply init, bootstrap and start on all of the 3 SCMs, so that manual operation will be safer. Is this the requirement ?
> 
> yes. This is a requirement for Kubernetes setups .

Thanks for the explanation. Could you add the explanation for the scenario in comments and ozone-site.xml ?

GlenGeng:
+1.

PR #2001, opened 2021-03-08
HDDS-4911. List container by container state.

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-4911 

There are a lot of unhealthy (CLOSING) state container in our production cluster, add this command line filter to find them all.


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/client/ScmClient.java
```diff
@@ -98,12 +98,13 @@ void deleteContainer(long containerId, Pipeline pipeline, boolean force)
    *
    * @param startContainerID start containerID.
    * @param count count must be {@literal >} 0.
+   * @param state Container of this state will be returned.
    *
    * @return a list of pipeline.
    * @throws IOException
    */
   List<ContainerInfo> listContainer(long startContainerID,
```
xiaoyuyao, line 106: Can we add a new listContainer API with state and keep the existing one for better backward compatibility?

xiaoyuyao (commented):
Thanks @ChenSammi  for adding this. I just have some minor suggestion inline. Otherwise, LGTM.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/client/ScmClient.java
xiaoyuyao: This also helps avoiding some unnecessary unit test changes.

ChenSammi: Sure.  I will add a new listContainer API.  Actually that's the way of my first patch.  Then I thought a consolidated API looks like more neat.

adoroszlai (commented):
Thanks @ChenSammi for working on this.  Functionality LGTM.  Instead of the integration test (which seems to have no assertions), can you please add test cases to existing acceptance test?

https://github.com/apache/ozone/blob/master/hadoop-ozone/dist/src/main/smoketest/admincli/container.robot

ChenSammi:
@adoroszlai , thanks for the commets.  I added one new test step in container.robot.

adoroszlai (approved):
> I added one new test step in container.robot.

Thank you.

ChenSammi:
Thanks @adoroszlai @xiaoyuyao for the code review.

PR #2002, opened 2021-03-08
HDDS-4881. Pre-install awscli in ozone-runner

adoroszlai:

Upgrade `ozone-runner` to newest image with pre-installed `awscli`.

https://issues.apache.org/jira/browse/HDDS-4881

## How was this patch tested?

https://github.com/adoroszlai/hadoop-ozone/actions/runs/631673202

### DIFF
--- a/hadoop-ozone/dist/src/main/smoketest/s3/commonawslib.robot
+++ b/hadoop-ozone/dist/src/main/smoketest/s3/commonawslib.robot
@@ -39,6 +39,18 @@ Execute AWSS3Cli
     ${output} =       Execute                     aws s3 --endpoint-url ${ENDPOINT_URL} ${command}
     [return]          ${output}
 
+Install aws cli
+    ${rc}              ${output} =                 Run And Return Rc And Output           which apt-get
+    Run Keyword if     '${rc}' == '0'              Install aws cli s3 debian
+    ${rc}              ${output} =                 Run And Return Rc And Output           yum --help
+    Run Keyword if     '${rc}' == '0'              Install aws cli s3 centos
+
+Install aws cli s3 centos
+    Execute            sudo -E yum install -y awscli
+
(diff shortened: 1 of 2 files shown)


hadoop-ozone/dist/src/main/smoketest/s3/commonawslib.robot
```diff
@@ -39,18 +39,6 @@ Execute AWSS3Cli
     ${output} =       Execute                     aws s3 --endpoint-url ${ENDPOINT_URL} ${command}
     [return]          ${output}
 
-Install aws cli
```
xiaoyuyao, line 42: Thanks @adoroszlai  for the improvement. That will save us unnecessary CI execution time. 
I have one question: can we check and install only if the base image does not have it in the acceptance test? This way, the test itself is still self-contained without depending on specific docker images.

xiaoyuyao (commented):
LGTM overall, one minor question inline.

hadoop-ozone/dist/src/main/smoketest/s3/commonawslib.robot
adoroszlai: Thanks @xiaoyuyao for the review.  The installation logic is idempotent (`yum` does nothing if the package is already installed), so we could simply keep it without any new check.  I have considered this option, but we already have other tools which are not installed during tests, rather assumed to be present in the environment (eg. `boto3` most recently).

xiaoyuyao: Thanks @adoroszlai  for the explanation. Can keep the s3 install as-is and add those idempotent installation (like boto3) to the test to make it more resilient? This can be done in follow up JIRAs.

xiaoyuyao:
+1, will merge it shortly.

PR #2005, opened 2021-03-08
Rename Apache Hadoop Ozone to Apache Ozone in pom and markdown files

elek:

Replace mentioning of `Apache Hadoop Ozone` with `Apache Ozone` in pom.xml files and markdown files.

## How was this patch tested?

:eye:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,13 +1,9 @@
-Apache Hadoop Ozone Contribution guideline
+Apache Ozone Contribution guideline
 ===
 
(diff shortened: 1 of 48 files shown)


dineshchitlangia (approved):
Overall +1 @elek, thanks for the improvement.
I was wondering if we should also rename the modules appropriately, like currently we have modules as hadoop-hdds-xxx, we could possibly rename it to ozone-hdds-xxx in a new jira? What do you think? If it sounds good, I can take this activity.

ChenSammi:
+1.  Thanks @elek for continuously improving the code base.

PR #2006, opened 2021-03-09
HDDS-4921. Replace the usage of deprecated Junit#timeout() in Ozone.

xiaoyuyao:

Replace the deprecated Junit#timeout call with the one with explicit timeunit. 

https://issues.apache.org/jira/browse/HDDS-4921

## How was this patch tested?

Exiting UT.

### DIFF
--- a/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/scm/net/TestNodeSchemaManager.java
+++ b/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/scm/net/TestNodeSchemaManager.java
@@ -50,7 +50,7 @@ public class TestNodeSchemaManager {
   }
 
   @Rule
-  public Timeout testTimeout = new Timeout(30000);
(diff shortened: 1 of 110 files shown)


hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/scm/net/TestNodeSchemaManager.java
```diff
@@ -50,7 +50,7 @@ public TestNodeSchemaManager() {
   }
 
   @Rule
-  public Timeout testTimeout = new Timeout(30000);
+  public Timeout testTimeout = Timeout.seconds(300);
```
ayushtkn, line 53: Will it be `Timeout.seconds(30)` instead of 300?

ayushtkn (approved):
Thanx @xiaoyuyao for the work here.
Changes LGTM

adoroszlai:
Thanks @xiaoyuyao for the improvement and @ayushtkn for the review.

PR #2007, opened 2021-03-09
HDDS-4922. refactor code in SCMStateMachine.

GlenGeng:

refactor code in SCMStateMachine.

https://issues.apache.org/jira/browse/HDDS-4922

## How was this patch tested?

CI


bshashikant:
@GlenGeng , can you please rebase?

bshashikant (commented):
@GlenGeng , i think we need to do isInitiliazed check during SCMStateMachine#close() otherwise we may see a NULL pointer exception during scm -init. Can you please take care of that?

PR #2008, opened 2021-03-09
HDDS-4932. [FSO] Provide list subpaths function to perform recursive ACL check during delete and rename op

rakeshadr:

This task is to provide list subpaths interface to perform recursive ACL check during delete and rename operation.

https://issues.apache.org/jira/browse/HDDS-4932

## How was this patch tested?

Added UT.

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
@@ -1,0 +1,59 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership.  The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.hadoop.ozone.security.acl;
+
+import org.apache.hadoop.ozone.om.helpers.OzoneFileStatus;
+
+import java.io.IOException;
+import java.util.List;
+
+public interface OzonePrefixPath {
+
+  /**
+   * Lists immediate children(files or a directories) of the given keyPrefix.
+   * It won't do recursive traversal.
+   *
+   * Assume following is the Ozone FS tree structure.
+   *
+   *                  buck-1
+   *                    |
+   *                    a
+   *                    |
+   *      -----------------------------------
+   *     |           |                       |
+   *     b1          b2                      b3
+   *   -----       --------               ----------
+   *   |    |      |    |   |             |    |     |
+   *  c1   c2     d1   d2  d3             e1   e2   e3
+   *                   |                  |
+   *               --------               |
+   *              |        |              |
+   *           d21.txt   d22.txt        e11.txt
+   *
+   * Say, KeyPrefix = "a" will return immediate children [a/b1, a/b2, a/b3].
+   * Say, KeyPrefix = "a/b2" will return children [a/b2/d1, a/b2/d2, a/b2/d3].
+   *
+   * @param volumeName volume name
+   * @param bucketName bucket name
+   * @param keyPrefix  keyPrefix name
+   * @return list of immediate files or directories under the given keyPrefix.
+   * @throws IOException
+   */
+  List<OzoneFileStatus> getChildren(String volumeName,
+      String bucketName, String keyPrefix) throws IOException;
+
+}
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
@@ -27,6 +27,13 @@ import java.util.Iterator;
 public interface OzonePrefixPath {
 
   /**
+   * Returns file status for the given key path.
+   *
+   * @return OzoneFileStatus for the given path.
+   */
+  OzoneFileStatus getOzonePrefixPath();
+
+  /**
    * Lists immediate children(files or a directories) of the given keyPrefix.
    * It won't do recursive traversal. The given keyPrefix parameter should be a
    * directory type.
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
@@ -31,7 +31,7 @@ public interface OzonePrefixPath {
    *
    * @return OzoneFileStatus for the given path.
    */
-  OzoneFileStatus getOzonePrefixPath();
+  OzoneFileStatus getOzoneFileStatus();
 
   /**
    * Lists immediate children(files or a directories) of the given keyPrefix.
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
@@ -21,11 +21,15 @@ import org.apache.hadoop.ozone.om.helpers.OzoneFileStatus;
 import java.io.IOException;
 import java.util.Iterator;
 
+/**
+ * Interface used to lists immediate children(sub-paths) for a given keyPrefix.
+ */
 public interface OzonePrefixPath {
 
   /**
    * Lists immediate children(files or a directories) of the given keyPrefix.
-   * It won't do recursive traversal.
+   * It won't do recursive traversal. The given keyPrefix parameter should be a
+   * directory type.
    *
    * Assume following is the Ozone FS tree structure.
    *
@@ -51,5 +55,6 @@ public interface OzonePrefixPath {
    * @return list of immediate files or directories under the given keyPrefix.
    * @throws IOException
    */
-  Iterator<? extends OzoneFileStatus> getChildren(String keyPrefix) throws IOException;
+  Iterator<? extends OzoneFileStatus> getChildren(String keyPrefix)
+      throws IOException;
 }
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
@@ -19,7 +19,7 @@ package org.apache.hadoop.ozone.security.acl;
 import org.apache.hadoop.ozone.om.helpers.OzoneFileStatus;
 
 import java.io.IOException;
-import java.util.List;
+import java.util.Iterator;
 
 public interface OzonePrefixPath {
 
@@ -47,13 +47,9 @@ public interface OzonePrefixPath {
    * Say, KeyPrefix = "a" will return immediate children [a/b1, a/b2, a/b3].
    * Say, KeyPrefix = "a/b2" will return children [a/b2/d1, a/b2/d2, a/b2/d3].
    *
-   * @param volumeName volume name
-   * @param bucketName bucket name
    * @param keyPrefix  keyPrefix name
    * @return list of immediate files or directories under the given keyPrefix.
    * @throws IOException
    */
-  List<OzoneFileStatus> getChildren(String volumeName,
-      String bucketName, String keyPrefix) throws IOException;
-
+  Iterator<? extends OzoneFileStatus> getChildren(String keyPrefix) throws IOException;
 }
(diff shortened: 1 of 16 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -1828,6 +1828,57 @@ public boolean checkAcls(ResourceType resType, StoreType storeType,
...
+      boolean throwIfPermissionDenied, String volumeOwner,
+      boolean recursiveAccessCheck)
+      throws OMException {
+    OzoneObj obj = OzoneObjInfo.Builder.newBuilder()
+        .setResType(resType)
+        .setStoreType(storeType)
+        .setVolumeName(vol)
+        .setBucketName(bucket)
+        .setKeyName(key).build();
+    RequestContext context = RequestContext.newBuilder()
```
xiaoyuyao, line 1851: It seems every time the RequestContext changes, we will have to add a chain of checkAcls change. Can we pass a RequestContext (or its Builder) instead here for checkAcls to avoid future churns?

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
```diff
@@ -0,0 +1,59 @@
...
+   * Say, KeyPrefix = "a" will return immediate children [a/b1, a/b2, a/b3].
+   * Say, KeyPrefix = "a/b2" will return children [a/b2/d1, a/b2/d2, a/b2/d3].
+   *
+   * @param volumeName volume name
+   * @param bucketName bucket name
+   * @param keyPrefix  keyPrefix name
+   * @return list of immediate files or directories under the given keyPrefix.
+   * @throws IOException
+   */
+  List<OzoneFileStatus> getChildren(String volumeName,
```
xiaoyuyao, line 56: It seems that HDDS-4691 should use this API to do the recursion. Also can we provide a batching parameters so that if the number keys are huge, we won't hit RPC packet size limitation or memory limitation.

xiaoyuyao (commented):
Thanks @rakeshadr  for working on this. Patch LGTM overall, a few comments added inline...

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
rakeshadr: Thanks @xiaoyuyao for the review comments. Very good point.

How about to change the List to Iterator ?
```

interface OzonePrefixPath {

   Iterator<? extends OzoneFileStatus> getChildren(String keyPrefix) throws IOException;

   class PathIterator implements Iterator<OzoneFileStatus> {

      @Override
      public boolean hasNext() {
        // batch wise fetching....constant batchSize=1000
        return false;
      }

      @Override
      public OzoneFileStatus next() {
        return null;
      }
    }

  }
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
rakeshadr: > RequestContext changes, we will have to add a chain of checkAcls change

@xiaoyuyao Are you pointing to a case where user performed delete/rename on a parentDir which has millions of subpaths, now for every subpath ozoneManager needs to invoke checkAcls by creating a `RequestContext context = RequestContext.newBuilder()` ?

Please correct me if I missed something. Thanks!

xiaoyuyao: No I mean we have many checkAcls() with lots of parameters and each time, we add one more, the chain will be extended longer. Instead of doing that, we can pass the wrapper RequestContext or its Builder for checkAcls parameter.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
xiaoyuyao: That looks good to me.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
rakeshadr: Thanks @xiaoyuyao for the clear explanation. I will refactor it.

rakeshadr:
@xiaoyuyao , I have addressed the comments. Could you please review it again, thanks!

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/OzonePrefixPath.java
rakeshadr: Done!

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
rakeshadr: Done!

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/security/acl/RequestContext.java
```diff
@@ -43,20 +43,23 @@
...
+            false, null);
   }
 
   @SuppressWarnings("parameternumber")
   public RequestContext(String host, InetAddress ip,
       UserGroupInformation clientUgi, String serviceId,
       ACLIdentityType aclType, ACLType aclRights,
-      String ownerName, boolean recursiveAccessCheck) {
+      String ownerName, boolean recursiveAccessCheck,
+      OzonePrefixPath ozonePrefixPath) {
```
xiaoyuyao, line 62: Should we put the OzonePrefixPath in the OzoneObject?

rakeshadr: Thanks @xiaoyuyao for the comment. Agreed.

I hope you are thinking that, OzoneObject has reference to `volume, bucket and keyname` attributes, so we can also add `ozonePrefixPath` to OzoneObject to make it more sense. Am I correct?

Is that `recursiveAccessCheck` flag in RequestContext work instead of adding a new flag to indicates the check for prefixPath? 

Now, `recursiveAccessCheck` flag will be true for any directory and false otherwise. I'm setting the value in OMKeyRenameRequestV1 and OMKeyDeleteRequestV1.

rakeshadr: @xiaoyuyao , I have updated the patch by addressing your comment. Thanks for the reviews!

rakeshadr:
Note: Test failure is urelated to the patch, triggered another build run.

```
Warning:  Tests run: 1, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 0.002 s - in org.apache.hadoop.ozone.client.rpc.TestContainerStateMachineFailureOnRead
[INFO] Running org.apache.hadoop.ozone.client.rpc.TestOzoneRpcClientWithRatis
Error:  Tests run: 80, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 62.65 s <<< FAILURE! - in org.apache.hadoop.ozone.client.rpc.TestOzoneRpcClientWithRatis
Error:  testValidateBlockLengthWithCommitKey  Time elapsed: 0.019 s  <<< FAILURE!
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
```

rakeshadr:
@xiaoyuyao test case failure is unrelated to the patch. I've raised jira HDDS-5012 to address it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
```diff
@@ -186,6 +189,64 @@ public void checkAcls(OzoneManager ozoneManager,
...
+        .setStoreType(OzoneObj.StoreType.OZONE)
+        .setVolumeName(volumeName)
+        .setBucketName(bucketName)
+        .setKeyName(keyName)
+        .setOzonePrefixPath(pathViewer).build();
+
+    boolean isDirectory = pathViewer.getOzoneFileStatus().isDirectory();
+
+    RequestContext.Builder contextBuilder = RequestContext.newBuilder()
+        .setAclRights(IAccessAuthorizer.ACLType.DELETE)
```
xiaoyuyao, line 243: why do we hard code ACLType to DELETE here?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
```diff
@@ -607,11 +609,20 @@ public static OzoneFileStatus getOMKeyInfoIfExists(
...
       String keyName, long scmBlockSize) throws IOException {
 
+    OMFileRequest.validateBucket(omMetadataMgr, volumeName, bucketName);
+
     Path keyPath = Paths.get(keyName);
     Iterator<Path> elements = keyPath.iterator();
     String bucketKey = omMetadataMgr.getBucketKey(volumeName, bucketName);
     OmBucketInfo omBucketInfo =
             omMetadataMgr.getBucketTable().get(bucketKey);
+    if (null == omBucketInfo) {
```
xiaoyuyao, line 619: Does validateBucket handle the non-exist bucket case?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequestV1.java
```diff
@@ -94,16 +91,13 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
...
     OmBucketInfo omBucketInfo = null;
     try {
       keyArgs = resolveBucketLink(ozoneManager, keyArgs, auditMap);
       volumeName = keyArgs.getVolumeName();
       bucketName = keyArgs.getBucketName();
 
-      // check Acl
-      checkKeyAcls(ozoneManager, volumeName, bucketName, keyName,
-          IAccessAuthorizer.ACLType.DELETE, OzoneObj.ResourceType.KEY);
+      checkACLs(ozoneManager, volumeName, bucketName, keyName);
```
xiaoyuyao, line 100: If this is intended for delete only, should we rename it to checkACLsForDelete?

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/file/OMFileRequest.java
rakeshadr: OK, will remove this. Its a duplicate check.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/OMClientRequest.java
rakeshadr: Sure, will pass `ACLType` as an argument to this function and will make it generic.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/OMKeyDeleteRequestV1.java
rakeshadr: Same as above. I will pass `ACLType` as an argument to this function. Hope this make sense to you.

rakeshadr:
Thanks a lot @xiaoyuyao for the useful review comments. Updated PR by addressing the comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzonePrefixPathImpl.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Override
+    public boolean hasNext() {
+      if (!currentIterator.hasNext() && currentValue != null) {
+        try {
+          currentIterator =
+              getNextListOfKeys(currentValue.getTrimmedName()).iterator();
+        } catch (IOException e) {
+          throw new RuntimeException(e);
```
xiaoyuyao, line 110: Can we avoid throw RuntimeException here?

xiaoyuyao (commented):
Thanks @rakeshadr  for the update, just have one minor comments inline.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzonePrefixPathImpl.java
rakeshadr: OK. Then I will return false if any exception thrown by `getNextListOfKeys(currentValue.getTrimmedName())`  function call.

rakeshadr:
> Thanks @rakeshadr for the update, just have one minor comments inline.

Thanks a lot @xiaoyuyao  for the review comments. Sure, hope I can proceed to merge the patch once that is handled.

rakeshadr:
Updated the patch by addressing the comment. Thank you @xiaoyuyao!

rakeshadr:
Thanks again @xiaoyuyao for the great support and the review comments helped a lot:-)
I've addressed those comments and got a clean QA report.

I would like to merge this PR today sometime(IST) and I'm planning to rebase the feature branch to make the branch maintenance easy as it is 27 commits behind master. I hope you agree to this idea.

Please feel free to comments(if any), I'm really happy to work on those and make the changes.

xiaoyuyao:
+1. Thanks @rakeshadr  for the update, LGTM.

rakeshadr:
> +1. Thanks @rakeshadr for the update, LGTM.

Thanks @xiaoyuyao

PR #2009, opened 2021-03-09
HDDS-4916. No ACL check when uploading object through multi-upload.

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-4916


xiaoyuyao:
Thanks @ChenSammi  for working on this. The patch LGTM overall. One concern I have: After this change, we will need corresponding changes on the s3v and bucket acls. This can be either manual or as part of (get s3 secret) so that when the acl check is enforced, MPU will not fail unexpectedly.

ChenSammi:
> Thanks @ChenSammi for working on this. The patch LGTM overall. One concern I have: After this change, we will need corresponding changes on the s3v and bucket acls. This can be either manual or as part of (get s3 secret) so that when the acl check is enforced, MPU will not fail unexpectedly.

@xiaoyuyao , I didn't understand your concern clearly.  Would you explain it a little more?

xiaoyuyao:
> > Thanks @ChenSammi for working on this. The patch LGTM overall. One concern I have: After this change, we will need corresponding changes on the s3v and bucket acls. This can be either manual or as part of (get s3 secret) so that when the acl check is enforced, MPU will not fail unexpectedly.
> 
> @xiaoyuyao , I didn't understand your concern clearly. Would you explain it a little more?

My concerns is that previous, S3 secret gives users permission to access buckets under s3v. Now, additional permission must be provisioned on top of it to allow user to read/write using the S3 secret. 

Should this one wait to use S3 acl mapping or expect ACL for s3 users configured via Kerberos?

ChenSammi:
> 
> 
> > > Thanks @ChenSammi for working on this. The patch LGTM overall. One concern I have: After this change, we will need corresponding changes on the s3v and bucket acls. This can be either manual or as part of (get s3 secret) so that when the acl check is enforced, MPU will not fail unexpectedly.
> > 
> > 
> > @xiaoyuyao , I didn't understand your concern clearly. Would you explain it a little more?
> 
> My concerns is that previous, S3 secret gives users permission to access buckets under s3v. Now, additional permission must be provisioned on top of it to allow user to read/write using the S3 secret.
> 
> Should this one wait to use S3 acl mapping or expect ACL for s3 users configured via Kerberos?

@xiaoyuyao , current ACL check depends on whether "ozone.acl.enabled" is true,  it doesn't check whether "ozone.security.enabled" is true or not.   In our production cluster, "ozone.acl.enabled" is true and "ozone.security.enabled" is false.  I did a little more test,  small objects uploading through S3G are already controled by ACL of bucket.  Only the large object uploading now doesn't have the ACL check. 

HDFS uses "hadoop.security.authentication" property to control the authentication,  "simple" or "kerberos".  "simple" equals to "ozone.security.enabled" false. "Kerberos' equals to "ozone.security.enabled" true.  HDFS uses "dfs.permissions.enabled" property to control whether permission check should be enforeced, similar to Ozone "ozone.acl.enabled" property.  So generally, I think Ozone has the similar security control pattern as HDFS.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCompleteRequest.java
```diff
@@ -126,7 +128,9 @@ public OMClientResponse validateAndUpdateCache(OzoneManager ozoneManager,
       multipartKey = omMetadataManager.getMultipartKey(volumeName,
           bucketName, keyName, uploadID);
 
-      // TODO to support S3 ACL later.
+      // check Acl
+      checkKeyAcls(ozoneManager, volumeName, bucketName, keyName,
+          IAccessAuthorizer.ACLType.WRITE, OzoneObj.ResourceType.KEY);
```
xiaoyuyao, line 133: Should we check for CREATE as we are completing the key write by adding new entry in the KeyTable?

xiaoyuyao (commented):
Agree, it makes sense to check ACL consistently between small upload and MPU. The small upload goes through normal write key RPC and get the acl check without additional code. MPU has to do that during the different phases. I have one last comment added. Otherwise, LGTM.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/s3/multipart/S3MultipartUploadCompleteRequest.java
ChenSammi: Xiaoyu, from user's point of view,  to upload a object, both CREATE and WRITE permission are needed.  Currently OmKeyCreateRequest checks CREATE permission and OmKeyCommitRequest checks WRITE permission.  Multi-upload follows the same rule,  CREATE permission is checked when initiating a multi-upload request.

xiaoyuyao: Sounds good tome. +1.

PR #2012, opened 2021-03-09
HDDS-4913. Refine the native authorizer parent context right check.

xiaoyuyao:

Change the parent context right check based on READ/WRITE.

https://issues.apache.org/jira/browse/HDDS-4913

## How was this patch tested?

Draft patch based on current UT and acceptance test. Will add more test with followup changes.


ChenSammi:
@xiaoyuyao ,  currently parent context check of a key applies to both bucket and volume.  While the volume is the parent of bucket, not the parent of key,  I think we need to have seperate parent context for bucket and volume.  Say, If I want to create key1 under volume1/bucket1,  write permission on bucket1 is required, but write permission on volume1 is not necessary.

xiaoyuyao:
> @xiaoyuyao , currently parent context check of a key applies to both bucket and volume. While the volume is the parent of bucket, not the parent of key, so shall we consider to have seperate parent context for bucket and volume? Say, If I want to create key1 under volume1/bucket1, write permission on bucket1 is required, but write permission on volume1 seems not necessary.

ChenSammi:
> > @xiaoyuyao , currently parent context check of a key applies to both bucket and volume. While the volume is the parent of bucket, not the parent of key, so shall we consider to have seperate parent context for bucket and volume? Say, If I want to create key1 under volume1/bucket1, write permission on bucket1 is required, but write permission on volume1 seems not necessary.
> 
> Based on offline discussion with @arp7, he raised a good point that the requirement of write permission on volume gives admin a quick way to prevent user access to all the resources (buckets/keys) in one change at volume level instead of individual changes at buckets/keys level.

For @arp7 's requirement, we can have a tool or CLI command to achieve this.  If we think of volume/bucket/key as a FileSystem tree, it's not very make sense that write a grand-child node need the write permission of it's grand-parent.

xiaoyuyao:
Thanks @ChenSammi  for the review. This JIRA fix the issue with the parent context of READ_ACL/WRITE_ACL/CREATE right. We can address that in followup discussions. Can you take another look? Thanks!

xiaoyuyao:
@rakeshadr, can you help reviewing? Thanks!

xiaoyuyao:
@rakeshadr, thanks for the review. Please find the UT added in this commit a3527b9.  The related tasks are captured under umbrella JIRA HDDS-4655.

rakeshadr (approved):
+1 LGTM

xiaoyuyao:
Thanks @rakeshadr for the review. PR has been merged.

PR #2013, opened 2021-03-09
HDDS-4856. Ruby S3 SDK never get authenticated by Ozone

elek:

As reported the `Content-Type: ""` is not handled by Ozone S3g but handled by AWS. 

With a simple filter we can make Ozone more similar to AWS and mimic this strange behavior.

https://issues.apache.org/jira/browse/HDDS-4856

## How was this patch tested?

Started locally the S3g, reproduced the problem and tested with the new filter and worked well...


swagle (approved):
+1 LGTM

elek:
Thanks the review @swagle and the report  UENISHI Kota, I am mergint it now...

PR #2015, opened 2021-03-09
HDDS-4946. Do not wait one heartbeat to move newly registered datanodes that match SCM's MLV from HEALTHY_READONLY to HEALTHY

errose28:

Currently, when a datanode is registered with SCM, it is added as HEALTHY_READONLY, regardless of its metadata layout version. On the next heartbeat, the layout is checked against that of SCM's to determine if it matches and can be moved to HEALTHY. This Jira will add the check for layout match when the datanode is registered, as well as on each heartbeat, so nodes whose MLV matches SCM's on registration will move directly from HEALTHY_READONLY to HEALTHY without needing to wait for a heartbeat.

HDDS-4946

## How was this patch tested?

Recon `TestEndpoints` unit test updated to pass layout version on heartbeat.
`TestNodeStateManager` updated to test the new behavior.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
@@ -180,7 +180,7 @@ public class NodeStateManager implements Runnable, Closeable {
     this.state2EventMap = new HashMap<>();
     initialiseState2EventMap();
     Set<NodeState> finalStates = new HashSet<>();
-    this.nodeHealthSM = new StateMachine<>(NodeState.HEALTHY_READONLY,
+    this.nodeHealthSM = new StateMachine<>(NodeState.HEALTHY,
         finalStates);
     initializeStateMachines();
     heartbeatCheckerIntervalMs = HddsServerUtil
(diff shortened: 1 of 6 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
```diff
@@ -278,10 +292,21 @@ private void initializeStateMachines() {
...
    */
   public void addNode(DatanodeDetails datanodeDetails,
-                      LayoutVersionProto layoutInfo)
-      throws NodeAlreadyExistsException {
+      LayoutVersionProto layoutInfo) throws NodeAlreadyExistsException {
     NodeStatus newNodeStatus = newNodeStatus(datanodeDetails);
     nodeStateMap.addNode(datanodeDetails, newNodeStatus, layoutInfo);
+    UUID dnID = datanodeDetails.getUuid();
+    try {
+      updateLastKnownLayoutVersion(datanodeDetails, layoutInfo);
```
avijayanhwx, line 300: Do we need to add a state == HEALTHY_READONLY check here?

avijayanhwx: Just an alternative suggestion, can we init state as HEALTHY for the Datanode, and move it to READONLY based on the alternative condition (layoutMismatch)? It is more likely for a new DN to be in a correct rather than a wrong layout version.

avijayanhwx (commented):
Thanks for working on this @errose28. A couple of comments inline.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeStateManager.java
errose28: Starting nodes as healthy and moving to healthy readonly when node added if there is a layout mismatch makes sense. I updated the code.

errose28: `updateLastKnownLayoutVersion`  and `updateNodeLayoutVersionState` I believe will handle the state transitions without an extra check.

avijayanhwx (approved):
LGTM +1

PR #2022, opened 2021-03-10
HDDS-4951. Return with exit code 0 in case of optional scm bootstrap/init 

elek:

HDDS-4896 provides a very elegant way to initialize SCM HA in containerized environment:

Both `-init` and `-bootstrap` can be executed, but when `ozone.scm.primordial.node.id` is set, the `--init` is ignored on non-primordial, the `--bootstrap` ignored on primordial node(s).

To make it even easier to use in k8s environments I suggest to return with exit code 0 in these cases as init proces shouldn't be repeated as failed: in these cases ignoring the init/bootstrap is part of the expected workflow.

https://issues.apache.org/jira/browse/HDDS-4951

## How was this patch tested?

Deployed the Ozone to a k8s cluster with the help of https://issues.apache.org/jira/browse/HDDS-4950 yaml files.


adoroszlai:
@elek I think this should be opened with `HDDS-2823` (SCM HA) as the base branch, not `master`.

PR #2024, opened 2021-03-10
HDDS-4917.[FSO]Implement ACL requests for new layout

sadanand48:
Implement ACL api's like getAcl, addAcl, setAcl ,removeAcl in the new layout V1.

https://issues.apache.org/jira/browse/HDDS-4917

## How was this patch tested?
Added Unit test

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/acl/OMKeyAclRequestV1.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/acl/OMKeyAclRequestV1.java
@@ -31,6 +31,7 @@ import org.apache.hadoop.ozone.om.ratis.utils.OzoneManagerDoubleBufferHelper;
 import org.apache.hadoop.ozone.om.request.file.OMFileRequest;
 import org.apache.hadoop.ozone.om.request.util.ObjectParser;
 import org.apache.hadoop.ozone.om.response.OMClientResponse;
+import org.apache.hadoop.ozone.om.response.key.acl.OMKeyAclResponseV1;
 import org.apache.hadoop.ozone.protocol.proto.OzoneManagerProtocolProtos;
 import org.apache.hadoop.ozone.security.acl.IAccessAuthorizer;
 import org.apache.hadoop.ozone.security.acl.OzoneObj;
@@ -142,6 +143,19 @@ public abstract class OMKeyAclRequestV1 extends OMKeyAclRequest {
         ozoneManager.getAuditLogger(), auditMap);
 
     return omClientResponse;
+  }
+
+  /**
+   * Get the om client response on failure case with lock.
+   *
+   * @param omResponse
+   * @param exception
+   * @return OMClientResponse
+   */
+  @Override OMClientResponse onFailure(
+      OzoneManagerProtocolProtos.OMResponse.Builder omResponse,
+      IOException exception) {
+    return new OMKeyAclResponseV1(createErrorOMResponse(omResponse, exception));
   }
 
   abstract OMClientResponse onSuccess(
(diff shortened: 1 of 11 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/acl/OMKeyAddAclRequestV1.java
```diff
@@ -0,0 +1,149 @@
...
+    return super.validateAndUpdateCache(ozoneManager, trxnLogIndex,
+        omDoubleBufferHelper);
+  }
+
+  @Override OMClientResponse onSuccess(
+      OzoneManagerProtocolProtos.OMResponse.Builder omResponse,
+      OmKeyInfo omKeyInfo, boolean operationResult, boolean isDirectory) {
+    omResponse.setSuccess(operationResult);
+    omResponse.setSetAclResponse(
+        OzoneManagerProtocolProtos.SetAclResponse.newBuilder()
```
linyiqun, line 145: Seems here should be:
```java
omResponse.setAddAclResponse(
        OzoneManagerProtocolProtos.AddAclResponse.newBuilder()
```

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/acl/OMKeyRemoveAclRequestV1.java
```diff
@@ -0,0 +1,153 @@
...
+    return super.validateAndUpdateCache(ozoneManager, trxnLogIndex,
+        omDoubleBufferHelper);
+  }
+
+  @Override OMClientResponse onSuccess(
+      OzoneManagerProtocolProtos.OMResponse.Builder omResponse,
+      OmKeyInfo omKeyInfo, boolean operationResult, boolean isDirectory) {
+    omResponse.setSuccess(operationResult);
+    omResponse.setSetAclResponse(
+        OzoneManagerProtocolProtos.SetAclResponse.newBuilder()
```
linyiqun, line 149: The similar problem, expected:  
 omResponse.setRemoveAclResponse(
        OzoneManagerProtocolProtos.RemoveAclResponse.newBuilder()

sadanand48: Thanks @linyiqun for pointing this out.

linyiqun (approved):
LGTM, +1. Thanks for addressing the comments.
Let's wait for others to have a double check for this PR before the merge, : ).

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/request/key/acl/OMKeyAclRequestV1.java
```diff
@@ -0,0 +1,151 @@
...
+        omMetadataManager.getKeyTable().addCacheEntry(new CacheKey<>(dbKey),
+            new CacheValue<>(Optional.of(omKeyInfo), trxnLogIndex));
+      }
+      omClientResponse =
+          onSuccess(omResponse, omKeyInfo, operationResult, isDirectory);
+      result = Result.SUCCESS;
+    } catch (IOException ex) {
+      result = Result.FAILURE;
+      exception = ex;
+      omClientResponse = onFailure(omResponse, ex);
```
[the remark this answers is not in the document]
sadanand48: Thanks @rakeshadr for the review comment . Addressed it.

rakeshadr:
+1 patch looks good to me.

Thanks @sadanand48 for the contribution. Thanks a lot @linyiqun for the reviews.

rakeshadr:
@xiaoyuyao this patch is expecting prefixes and the key should exist in DB. Otw, it will throw Key_Not_Found exception. IIRC, we have discussed this case of non-existent path and not required to handle this case. Could you please add your thoughts. Thanks!

IMHO, I am open to discuss this point and if we need to support then it can be handled separately in a new jira. I'm merging this PR now.

PR #2028, opened 2021-03-11
HDDS-4877. Make SCM ratis server spin up time during initialization c…

JacksonYao287:

Make SCM ratis server spin up time during initialization configurable

https://issues.apache.org/jira/browse/HDDS-4877

## How was this patch tested?

unit test

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
@@ -156,6 +156,24 @@ public class SCMHAConfiguration {
   )
   private long ratisLeaderElectionTimeout = 5 * 1000L;
 
+  @Config(key = "ratis.leader.ready.wait.timeout",
+      type = ConfigType.TIME,
+      defaultValue = "60s",
+      tags = {SCM, OZONE, HA, RATIS},
+      description = "The minimum timeout duration for waiting for" +
+                    "leader readiness"
+  )
+  private long ratisLeaderReadyWaitTimeout = 60 * 1000L;
+
+  @Config(key = "ratis.leader.ready.check.interval",
+      type = ConfigType.TIME,
+      defaultValue = "2s",
+      tags = {SCM, OZONE, HA, RATIS},
+      description = "The interval between ratis server performing" +
+                    "a leader readiness check"
+  )
+  private long ratisLeaderReadyCheckInterval = 2 * 1000L;
+
   @Config(key = "ratis.server.failure.timeout.duration",
       type = ConfigType.TIME,
       defaultValue = "120s",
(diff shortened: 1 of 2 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
```diff
@@ -156,6 +156,22 @@
   )
   private long ratisLeaderElectionTimeout = 5 * 1000L;
 
+  @Config(key = "ratis.leader.ready.wait.timeout",
+          type = ConfigType.TIME,
+          defaultValue = "60s",
```
GlenGeng, line 161: indent. just 4 spaces.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
```diff
@@ -280,6 +296,10 @@ public long getLeaderElectionMaxTimeout() {
     return ratisLeaderElectionTimeout + 200L;
   }
 
+  public long getLeaderReadyWaitTimeout() { return ratisLeaderReadyWaitTimeout;}
```
GlenGeng, line 299:
```
public long getLeaderReadyWaitTimeout() {
  return ratisLeaderReadyWaitTimeout;
}
```

Better has setter too.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -94,8 +94,10 @@ public static void initialize(String clusterId, String scmId,
     try {
       server = newRaftServer(scmId, conf).setGroup(group).build();
       server.start();
-      // TODO: Timeout and sleep interval should be made configurable
-      waitForLeaderToBeReady(server, 60000, group);
+      final SCMHAConfiguration haConf = conf.getObject(SCMHAConfiguration.class);
+      long waitTimeout = haConf.getLeaderReadyWaitTimeout();
+      long retryInterval = haConf.getLeaderReadyCheckInterval();
+      waitForLeaderToBeReady(server, waitTimeout, retryInterval, group);
```
GlenGeng, line 100: How about just pass `conf` to `waitForLeaderToBeReady`, and read out `retryInterval` and `waitTimeout` there ?

GlenGeng (commented):
Thanks @JacksonYao287. Overall LGTM. Just some inline comments.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
JacksonYao287: thanks , fixed

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAConfiguration.java
JacksonYao287: @GlenGeng  thanks , a setter has been added

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
JacksonYao287: here, conf is an OzoneConfiguration object  which can not read out retryInterval and waitTimeout directly. So, for now , i have to get a SCMHAConfiguration to do this. It seems better to pass conf to waitForLeaderToBeReady,done.
maybe we can unify SCMHAConfiguration to OzoneConfiguration later, and then there will be only one Configuration for scm. what do you think about it?

amaliujia:
LGTM

Thanks @JacksonYao287 for your contribution! Thanks @GlenGeng  for your review!

amaliujia (approved):

PR #2031, opened 2021-03-12
HDDS-4971. Define Ozone specific trash-interval and checkpoint interval

sadanand48:
This change adds 2 new configs in ozone. One thing to note here is that ozone trash is not fully independent from the hadoop-configurations i.e (`fs.trash.interval`) . This config (`fs.trash.interval`) will still control enabling/disabling of trash feature whereas the newly added configs (`ozone.fs.trash.interval` & `ozone.fs.trash.checkpoint.interval` ) would set values if trash is enabled . These will add flexibility if ozone user wants to set ozone specific interval period. 
 
https://issues.apache.org/jira/browse/HDDS-4971

## How was this patch tested?
Unit test

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
@@ -133,7 +133,8 @@ public class TestOzoneFileSystem {
 
   private void init() throws Exception {
     OzoneConfiguration conf = new OzoneConfiguration();
-    conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
+    conf.setInt(FS_TRASH_INTERVAL_KEY, 2);
+    conf.setInt(OMConfigKeys.OZONE_FS_TRASH_INTERVAL_KEY, 1);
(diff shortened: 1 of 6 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneFileSystem.java
```diff
@@ -798,10 +799,11 @@ public void testTrash() throws Exception {
     String testKeyName = "testKey2";
     Path path = new Path(OZONE_URI_DELIMITER, testKeyName);
     ContractTestUtils.touch(fs, path);
-    Assert.assertTrue(trash.getConf().getClass(
-        "fs.trash.classname", TrashPolicy.class).
+    Assert.assertTrue(trash.getConf().getClass("fs.trash.classname",
```
mukul1987, line 802: this line is an unrelated indentation related change ?

sadanand48: Yes, addressed comment.Thanks

mukul1987 (approved):
+1, the change looks good to me.

PR #2034, opened 2021-03-12
HDDS-4953. [SCM HA Security] Make CertStore DB updates for StoreValidateCertificate go via Ratis

bharatviswa504:

This Jira is to make DB updates of CertStore go via ratis, so that all SCM's can be in sync.

In this way, OM/DN/SCM Certs will be in sync across Ratis.

https://issues.apache.org/jira/browse/HDDS-4953

## How was this patch tested?

Added tests for codec.


GlenGeng:
Please hold on the merge util we finish merging master back to 2823. Thanks!

bshashikant (commented):
Thanks @bharatviswa504 for the patch. The changes look good. Can you please add a miniOzoneHACluster based test and verify the storeCertificate behaviour with ratis replication?

bharatviswa504:
> Thanks @bharatviswa504 for the patch. The changes look good. Can you please add a miniOzoneHACluster based test and verify the storeCertificate behaviour with ratis replication?

@bshashikant I think We don't have a MiniOzoneSecure cluster, due to the limitation of not able to login Kerberos credential in a single JVM (I think this was the reason, but not 100% sure cc @xiaoyuyao ). Once we have complete Security work I can try to add tests and verify this via Robot tests.

bshashikant:
> > Thanks @bharatviswa504 for the patch. The changes look good. Can you please add a miniOzoneHACluster based test and verify the storeCertificate behaviour with ratis replication?
> 
> @bshashikant I think We don't have a MiniOzoneSecure cluster, due to the limitation of not able to login multiple Kerberos credential in a single JVM (I think this was the reason, but not 100% sure cc @xiaoyuyao ). Once we have complete Security work I can try to add tests and verify this via Robot tests.

@bharatviswa504 , can't we directly use server API's i MiniOzoneHACuster to verify the behaviour of storeValidCertificate bypassing the other security checks?

bharatviswa504:
> storeValidCertificate

We don't start the security protocol Server in non-secure. So, not sure what do you mean by directly use server API here.

bharatviswa504:
> > > Thanks @bharatviswa504 for the patch. The changes look good. Can you please add a miniOzoneHACluster based test and verify the storeCertificate behaviour with ratis replication?
> > 
> > 
> > @bshashikant I think We don't have a MiniOzoneSecure cluster, due to the limitation of not able to login multiple Kerberos credential in a single JVM (I think this was the reason, but not 100% sure cc @xiaoyuyao ). Once we have complete Security work I can try to add tests and verify this via Robot tests.
> 
> @bharatviswa504 , can't we directly use server API's i MiniOzoneHACuster to verify the behaviour of storeValidCertificate bypassing the other security checks?

We don't start the security protocol Server in non-secure. So, not sure what do you mean by directly use server API here.

bharatviswa504:
**Just a side question**

Do we have any specific tests to test the behavior of replicate annotation(looks like no) if those work we should be good right, as this is using existing Replicate annotation.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMCertStore.java
```diff
@@ -56,20 +61,21 @@
...
 
-  public SCMCertStore(SCMMetadataStore dbStore, long sequenceId) {
+  private SCMCertStore(SCMMetadataStore dbStore, long sequenceId) {
     this.scmMetadataStore = dbStore;
     lock = new ReentrantLock();
     crlSequenceId = new AtomicLong(sequenceId);
   }
 
   @Override
+  @Replicate
```
GlenGeng, line 78: You'd better place `@Replicate` inside the interface `CertificateStore`. Please check `ContainerStateManagerV2` and `ContainerStateManagerImpl` as an example.

GlenGeng (commented):
LGTM.  Just one inline comments.

GlenGeng:
> **Just a side question**
> 
> Do we have any specific tests to test the behavior of replicate annotation(looks like no) if those work we should be good right, as this is using existing Replicate annotation, and the codec serialization/deserialization tests are already added. Do you think we need to add tests?

We have one `TestReplicationAnnotation`, yet it is a unit test.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMCertStore.java
bharatviswa504: Replicate class in hadoop-hdds-server-scm and CertificateStore in hdds-server-framework. 
hadoop-hdds-server-scm depends on framework. So, taken a simple approach and placed annotation in SCMCertSttore. 
So, to move @Replication to CertificateStore I need to move the Replicate Class to hdds-server-framework. Let me do it.

GlenGeng: I just run `TestSCMCertStore` locally, and not seen the trace 
```
    LOG.info("Invoking method {} on target {}, cost {}us",
        method, ratisHandler, (Time.monotonicNowNanos() - startTime) / 1000.0);
```
seems the annotation does not work if placing in subclass.

@bshashikant Above trace is sufficient to indicate whether annotated method goes through ratis, even in unit test.

bshashikant: Its not preferrable solution i feel, to check for specific log messages in a test. That's why , i thought replicating over ratis and verifying the db is a better choice.  In this case, i am ok with any of the approaches mentioned.

@GlenGeng , thanks for testing out replicate annotation over the sub class and interface.

bharatviswa504: TestSCMCertStore is not constructing RatisServer that is the reason, we don't see the log.

I have updated TestReplicateAnnotation itself, to call storeAndValidateCertificate.

GlenGeng:
Thanks for the verification. LGTM. Please rebase with the latest 2823.

bharatviswa504:

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMCertStore.java
```diff
@@ -56,14 +60,14 @@
...
-public class SCMCertStore implements CertificateStore {
+public final class SCMCertStore implements CertificateStore {
   private static final Logger LOG =
       LoggerFactory.getLogger(SCMCertStore.class);
   private SCMMetadataStore scmMetadataStore;
   private final Lock lock;
   private AtomicLong crlSequenceId;
 
-  public SCMCertStore(SCMMetadataStore dbStore, long sequenceId) {
+  private SCMCertStore(SCMMetadataStore dbStore, long sequenceId) {
```
GlenGeng, line 70: @bharatviswa504 After integrate with Ratis, the db writes should go to `transactionBuffer` and async flush to db by `takeSnapshot`, seems this part is missing.

bharatviswa504: This was an intentional choice not to make CertStore DB updates use transactionalBuffer. The reason is, there is no in-memory map(Like how containerstateManager has), to get immediate reads work (Read operation after a write). (Or we should use table cache and add to table cache, and then listCerts need to be modified for it). As the request for certs in a cluster is a number of nodes in the cluster, so made this choice of directly persisting to DB.

When I am about to type it is idempotent, I see there is a problem, as for replay transaction because of below check looks 

    if ((getCertificateByID(serialID, VALID_CERTS) != null) ||
        (getCertificateByID(serialID, CertType.REVOKED_CERTS) != null)) {
      throw new SCMSecurityException("Conflicting certificate ID");
    }

If we move this check to DefaultCAServer, then storeValidCertificate will be an idempotent operation, and we will not have a replay issue. Let me know if this makes sense, I will open a Jira to fix this.

GlenGeng: Thanks for the verification!  Will inform you if find any issue.

PR #2040, opened 2021-03-15
HDDS-4990. Introduce First upgrade startup action and Pre-finalized state validation in Layout Feature.

avijayanhwx:

- Introduce 2 new phases of upgrade action hooks (per layout feature)
   **Prefinalized state validation** - A layout feature (version) can use this to validate that the component is not started up in a way to use it before finalization. For example, an SCM HA validation action can make sure HA is not enabled before finalization.
**First Upgrade Start action** - Run exactly once when a component is started up after an upgrade with an unfinalized layout feature.
- Annotation based registration of layout actions to the layout features. After this change, an HDDS upgrade action be created like this.

> @UpgradeActionHdds(feature = SCM_HA, component = DATANODE, type = ON_FIRST_UPGRADE_START)
> public class ScmHAFirstUpgradeLayoutChangeAction implements HDDSUpgradeAction<DatanodeStateMachine> {`

https://issues.apache.org/jira/browse/HDDS-4990

## How was this patch tested?
Added test cases.
Manually verified on docker.


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/upgrade/HDDSLayoutVersionManager.java
```diff
@@ -31,12 +43,61 @@
...
   public static int maxLayoutVersion() {
     HDDSLayoutFeature[] features = HDDSLayoutFeature.values();
     return features[features.length - 1].layoutVersion();
   }
+
+  /**
+   * Scan classpath and register all actions to layout features.
+   */
+  @VisibleForTesting
+  void registerUpgradeActions(String[] packageNames) {
```
swagle, line 66: As a general practice classpath scan and reflective class loading is done from static methods so that classpath configuration errors can be thrown before initializing the application.

avijayanhwx: Yes, I agree. But, in this case the classpath loading is more investigative. It does not need to find any upgrade action at all. If it does find one, we will register it.

avijayanhwx:
cc @errose28 for review.

errose28 (commented):
Thanks for the feature @avijayanhwx LGTM +1

swagle (approved):

avijayanhwx:
Thanks for the review @errose28 and @swagle.

PR #2043, opened 2021-03-15
HDDS-4978. [SCM HA Security] Ozone services should be disabled in SCM HA enabled and security enabled cluster

bharatviswa504:

Do not start ozone services when security and SCM HA is enabled

https://issues.apache.org/jira/browse/HDDS-4978

## How was this patch tested?
Tested it with docker-compose ozone-ha cluster by settingozone.security.enabled=true
Verified that all services failed to come up

Snippet of partial log.
```
om2_1       | Ozone Services cannot be started on a secure SCM HA enabled cluster
om2_1       | 2021-03-15 12:09:15,429 [shutdown-hook-0] INFO om.OzoneManagerStarter: SHUTDOWN_MSG:
om2_1       | /************************************************************
om2_1       | SHUTDOWN_MSG: Shutting down OzoneManager at om2/172.27.0.4
om2_1       | ************************************************************/
scm1_1      | Ozone Services cannot be started on a secure SCM HA enabled cluster
scm1_1      | 2021-03-15 12:09:15,686 [shutdown-hook-0] INFO server.StorageContainerManagerStarter: SHUTDOWN_MSG:
scm1_1      | /************************************************************
scm1_1      | SHUTDOWN_MSG: Shutting down StorageContainerManager at 2a2fc5196315/172.27.0.2
scm1_1      | ************************************************************/
````

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
@@ -490,6 +490,14 @@ public final class ScmConfigKeys {
   public static final boolean HDDS_DATANODE_UPGRADE_LAYOUT_INLINE_DEFAULT =
       true;
 
+
+  // Temporary config which will be used only for test only purposes until
+  // SCM HA Security work is completed. This config should not be modified by
+  // users.
+  public static final String OZONE_SCM_HA_SECURITY_ENABLE =
+      "hdds.scm.ha.security.enable";
+  public static final boolean OZONE_SCM_HA_SECURITY_ENABLE_DEFAULT = false;
+
   /**
    * Never constructed.
    */
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
@@ -494,9 +494,9 @@ public final class ScmConfigKeys {
   // Temporary config which will be used only for test only purposes until
   // SCM HA Security work is completed. This config should not be modified by
   // users.
-  public static final String OZONE_SCM_HA_SECURITY_ENABLE =
+  public static final String OZONE_SCM_HA_SECURITY_SUPPORTED =
       "hdds.scm.ha.security.enable";
-  public static final boolean OZONE_SCM_HA_SECURITY_ENABLE_DEFAULT = false;
+  public static final boolean OZONE_SCM_HA_SECURITY_SUPPORTED_DEFAULT = false;
 
   /**
    * Never constructed.
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HAUtils.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HAUtils.java
@@ -20,6 +20,7 @@ import org.apache.hadoop.hdds.HddsConfigKeys;
 import com.google.protobuf.ServiceException;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.scm.AddSCMRequest;
+import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.ScmInfo;
 import org.apache.hadoop.hdds.scm.ha.SCMHAUtils;
 import org.apache.hadoop.hdds.scm.ha.SCMNodeInfo;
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HAUtils.java [hunk 1]
```diff
@@ -308,4 +311,15 @@ public static boolean isAccessControlException(Exception ex) {
     }
     return false;
   }
+
+  public static void checkSecurityAndSCMHAEnabled(OzoneConfiguration conf) {
+    if (OzoneSecurityUtil.isSecurityEnabled(conf)) {
+      List<SCMNodeInfo> scmNodeInfo = SCMNodeInfo.buildNodeInfo(conf);
+      if (scmNodeInfo.size() > 1) {
+        System.out.println("Ozone Services cannot be started on a secure SCM " +
```
xiaoyuyao, line 319: Should we print to System.err?

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/Gateway.java
```diff
@@ -54,6 +56,7 @@ public Void call() throws Exception {
     TracingUtil.initTracing("S3gateway", ozoneConfiguration);
     OzoneConfigurationHolder.setConfiguration(ozoneConfiguration);
     UserGroupInformation.setConfiguration(ozoneConfiguration);
+    checkSecurityAndSCMHAEnabled(ozoneConfiguration);
```
xiaoyuyao, line 59: S3G does not talk to SCM directly? We can skip the check here.

xiaoyuyao (commented):
Thanks @bharatviswa504  for the patch. LGTM overall. Few comments added inline.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/Gateway.java
bharatviswa504: I have added the check, because when secure SCM HA is enabled, to not to start services in ozone.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HAUtils.java
```diff
@@ -308,4 +311,15 @@ public static boolean isAccessControlException(Exception ex) {
     }
     return false;
   }
+
+  public static void checkSecurityAndSCMHAEnabled(OzoneConfiguration conf) {
```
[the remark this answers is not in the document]
bharatviswa504: Can you tell me why secure SCM HA acceptance tests will fail?

As already we have a check for checking node list size > 1.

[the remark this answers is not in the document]
xiaoyuyao: I understand the goal is to prevent user from running into issues when before SCM HA+ security is fully supported? However, this also prevents us running any acceptance test with SCM HA + security in the mean while unless we wont add any here.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/Gateway.java
xiaoyuyao: The other check makes sense because they are related to SCM. But S3G does not need to know SCM HA + security is enabled. The S3G node itself may not need to upgrade with the SCM HA change.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HAUtils.java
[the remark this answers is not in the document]
bharatviswa504: Got it, introduced a new config which we can set in docker tests.

hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/Gateway.java
bharatviswa504: When other services failed to start, even starting S3G is of no real use when OM and SCM are down in the cluster.

>The S3G node itself may not need to upgrade with the SCM HA change.
I am fine with removing the check here.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HAUtils.java
bharatviswa504: Done

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
```diff
@@ -490,6 +490,14 @@
   public static final boolean HDDS_DATANODE_UPGRADE_LAYOUT_INLINE_DEFAULT =
       true;
 
+
+  // Temporary config which will be used only for test only purposes until
+  // SCM HA Security work is completed. This config should not be modified by
+  // users.
+  public static final String OZONE_SCM_HA_SECURITY_ENABLE =
```
xiaoyuyao, line 497: NIT: can we rename this to OZONE_SCM_HA_SECURITY_SUPPORTED as this is compared against other keys has defined for security enabled and HA enabled? Otherwise, LGTM.

bharatviswa504: Done

xiaoyuyao:
Thanks @bharatviswa504  for the update. LGTM, +1.

bharatviswa504:

PR #2047, opened 2021-03-16
HDDS-4940: (Not for review) With assumptions, EC write POC

umamaheswararao:
This is not for review. This patch kept assumption for EC 3:2 and hard coded the values. This patch is for reference to figure out where the changes needed to support write EC.


adoroszlai:
Hi @umamaheswararao, can you please clarify:

> This is not for review.

Do you mean it's still draft, or PR is opened only to run CI?

umamaheswararao:
@adoroszlai This is just for reference to understand where the potential changes needed. In one of our EC discussion Marton requested to upload this POC part to use as ref. Once we have the parts like container id index, pipeline provider changes, this will be refined to get the actual patch. Also not intended for CI as well.

adoroszlai:
Is this PR needed?  I don't see any activity on it and it has conflicts.

PR #2048, opened 2021-03-16
HDDS-4984. Support async profiler 2.0

sky76093016:

Due to the update of Async profiler 2.0, the output format is changed from svg to html.

https://issues.apache.org/jira/browse/HDDS-4984

## How was this patch tested?

Local test.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/ProfileServlet.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/ProfileServlet.java
@@ -75,16 +75,16 @@ import org.slf4j.LoggerFactory;
  * //  --ttsp            time-to-safepoint profiling
  * Example:
  * - To collect 30 second CPU profile of current process (returns FlameGraph
- * html)
+ * svg)
  * curl "http://localhost:10002/prof"
  * - To collect 1 minute CPU profile of current process and output in tree
  * format (html)
  * curl "http://localhost:10002/prof?output=tree&duration=60"
  * - To collect 30 second heap allocation profile of current process (returns
- * FlameGraph html)
+ * FlameGraph svg)
  * curl "http://localhost:10002/prof?event=alloc"
  * - To collect lock contention profile of current process (returns
- * FlameGraph html)
+ * FlameGraph svg)
  * curl "http://localhost:10002/prof?event=lock"
  * Following event types are supported (default is 'cpu') (NOTE: not all
  * OS'es support all events)
@@ -379,7 +379,9 @@ public class ProfileServlet extends HttpServlet {
               + "ready..");
     } else {
       if (safeFileName.endsWith(".html")) {
-        resp.setContentType("image/html+xml");
+        resp.setContentType("text/html");
+      } else if (safeFileName.endsWith(".svg")) {
+        resp.setContentType("image/svg+xml");
       } else if (safeFileName.endsWith(".tree")) {
         resp.setContentType("text/html");
       }
@@ -442,10 +444,10 @@ public class ProfileServlet extends HttpServlet {
       try {
         return Output.valueOf(outputArg.trim().toUpperCase());
       } catch (IllegalArgumentException e) {
-        return Output.HTML;
+        return Output.SVG;
       }
     }
-    return Output.HTML;
+    return Output.SVG;
   }
 
   private void setResponseHeader(final HttpServletResponse response) {
@@ -511,6 +513,7 @@ public class ProfileServlet extends HttpServlet {
     TRACES,
     FLAT,
     COLLAPSED,
+    SVG,
     TREE,
     JFR,
     HTML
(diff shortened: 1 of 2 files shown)


adoroszlai (changes requested):
Thanks @sky76093016 for working on this.  Can you please keep SVG support, too?  For example `ozone-runner` docker image has Async Profiler 1.5.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/ProfileServlet.java
```diff
@@ -365,8 +378,8 @@ protected void doGetDownload(String fileName, final HttpServletRequest req,
           "This page will auto-refresh every 2 second until output file is "
               + "ready..");
     } else {
-      if (safeFileName.endsWith(".svg")) {
-        resp.setContentType("image/svg+xml");
+      if (safeFileName.endsWith(".html")) {
+        resp.setContentType("image/html+xml");
```
cku328, line 382:
```suggestion
        resp.setContentType("text/html");
```

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/ProfileServlet.java
```diff
@@ -500,7 +515,8 @@ public static Event fromInternalName(final String name) {
     COLLAPSED,
     SVG,
     TREE,
-    JFR
+    JFR,
+    HTML
```
adoroszlai, line 519: Async Profiler 2.0 does not seem to create useful output with `-o html`:

```
$ /opt/profiler/profiler.sh -d 10 -o html -f /tmp/prof-output/async-prof-pid-7-cpu-1.html 7
Profiling for 10 seconds
Done
$ cat /tmp/prof-output/async-prof-pid-7-cpu-1.html
Stopped profiling after 10 seconds. No dump options specified
```

We need `-o flamegraph` instead:

```
$ /opt/profiler/profiler.sh -d 10 -o flamegraph -f /tmp/prof-output/async-prof-pid-7-cpu-2.html 7
Profiling for 10 seconds
Done
$ head -3 /tmp/prof-output/async-prof-pid-7-cpu-2.html
<?xml version="1.0" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg version="1.1" width="1200" height="518" onload="init(evt)" viewBox="0 0 1200 518" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```

adoroszlai (commented):
Thanks @sky76093016 for updating the patch.  I think output format needs some tweak for 2.0.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/ProfileServlet.java
sky76093016: Thanks for tips. :")

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/ProfileServlet.java
sky76093016: Done.

jojochuang:
@adoroszlai  is it? i think it threw an error at me when i specified svg output...
regardless, the html report is easier to use.

jojochuang:
So... looks like even with the change, you would still need to point the browser to /prof?output=flamegraph in order to use async profiler 2.0, otherwise the servlet will attempt to generate svg output and that will fail.

jojochuang:
how about adding another system property  to specify the async profiler version, and defaults to flamegraph if it's 2.0?

adoroszlai:
> > 2.0 works with -o svg, although profiler's help does not mention it as a valid output format.
> is it? i think it threw an error at me when i specified svg output...

I think I was fooled by some docker-specific behavior (tried to replace `/opt/profiler` at runtime...).

Now with proper 2.0 installation I see the problem with `-o svg`.

It might be improved by checking the subprocess for error result instead of waiting forever for the output file.

> how about adding another system property to specify the async profiler version, and defaults to flamegraph if it's 2.0?

We could also run the profiler with `-v` first to get the version number.

jojochuang (approved):
+1 as discussed offline, let's merge this first, and we can work on the follow-up later.

cku328 (approved):
+1 for this PR. It is currently at least initially compatible with version 2.0.

elek:
> +1 as discussed offline, let's merge this first, and we can work on the follow-up later.

Did we open the follow-up issues? Looks like ideal newbie issues....

sky76093016:
Thanks @elek . I have issued jira and will finish it in a few days. The website of jira is: https://issues.apache.org/jira/browse/HDDS-5009

PR #2052, opened 2021-03-16
HDDS-4985. [SCM HA Security] When Ratis enable, SCM secure cluster is not working.

bharatviswa504:

Use actual method parameter type when invoking a method, instead of deriving it from actual argument. (This will help when the actual arg param type is subclass, where as method param type is superclass. If we don't use actual method param type during invoke, we get MethodNotFoundException.

https://issues.apache.org/jira/browse/HDDS-4985

## How was this patch tested?

Updated docker ozone-secure to run with ratis enabled.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisRequest.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisRequest.java
@@ -20,6 +20,7 @@ package org.apache.hadoop.hdds.scm.ha;
 import java.util.ArrayList;
 import java.util.List;
 
+import com.google.common.base.Preconditions;
 import com.google.protobuf.InvalidProtocolBufferException;
 
 import org.apache.hadoop.hdds.scm.ha.io.CodecFactory;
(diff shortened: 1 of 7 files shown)


bshashikant (approved):

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisRequest.java
```diff
@@ -86,9 +94,14 @@ public Message encode() throws InvalidProtocolBufferException {
     methodBuilder.setName(operation);
 
     final List<MethodArgument> args = new ArrayList<>();
+
+    int paramCounter = 0;
     for (Object argument : arguments) {
       final MethodArgument.Builder argBuilder = MethodArgument.newBuilder();
-      argBuilder.setType(argument.getClass().getName());
+      // Set actual method parameter type, not actual argument type.
+      // This is done to avoid MethodNotFoundException in case if argument is
```
GlenGeng, line 102: Nice change!

bharatviswa504:

hadoop-ozone/dist/src/main/compose/ozonesecure/test.sh
```diff
@@ -27,39 +27,42 @@ export SECURITY_ENABLED=true
 
 : ${OZONE_BUCKET_KEY_NAME:=key1}
 
-start_docker_env
+for enable in true false; do
```
elek, line 30: This trick doubled the execution time of the secure acceptance test which was already 50 min. It increased the full build time with ~50 mins and (because the parallel execution assumed all the tests finish in 1 hour) the full build is also slower with 30-40 minutes. 

Do we really need this trick?

(By the way the `enable` variable name is not very expressive, and it should be added to the `-N` because different tests will overwrite the results of each other) 

cc @adoroszlai @bharatviswa504

bharatviswa504: Thanks @elek It is added to test in a stand alone SCM cluster with ratis enable/disable.
As some of the integration issues were missed to identify in UTs. Later we have added a new acceptance test secure-ha, which tests end-to-end HA. (It would be good to have, but if we think it is expensive or any other way we can run this it would be good)
I think now we can remove this if we think this is causing a longer CI duration. Thanks, @adoroszlai for opening a Jira to fix this.

elek: Thanks to explain it @bharatviswa504. I think it's too expensive to run ALL the tests in two modes. For example, we can assume that the linked and real buckets working at the same way as they already tested. We don't need to test all of them in both cases (IMHO).

I think it's fine to test different settings to make sure if it works. But we can adjust the required tests. For the main/default case we can execute all the tests and the other mode we can execute only a minimum set.

PR #2053, opened 2021-03-16
HDDS-4988. Cancel duplicate PR workflows

adoroszlai:

CI runs for pull requests with multiple commits may waste resources if pushed one after another, not in a single batch.  Only the last commit is considered for PR status, so we should try to cancel any prior runs not yet completed.

`push` and `schedule` builds are not affected.

Custom Github Action from Jarek Potiuk is added as a submodule like [in Apache Airflow](https://github.com/apache/airflow/commit/f115983550833b1516ce58e66b49fe4fb1a0eff7) to avoid code duplication but still retain control over consuming new versions.

References:

 * item 1 in [Apache Airflow PR workflow description](https://github.com/apache/airflow/blob/master/PULL_REQUEST_WORKFLOW.rst#why-non-standard-pull-request-workflow)
 * [discussion](https://lists.apache.org/thread.html/ra806c0fd58fd55813a0f6a98bf3a0ec3e94d0db37224f877e9ea6ac8%40%3Cbuilds.apache.org%3E) on `builds` list

https://issues.apache.org/jira/browse/HDDS-4988

## How was this patch tested?

Tested on dummy PRs in my fork:

https://github.com/adoroszlai/hadoop-ozone/pull/10
https://github.com/adoroszlai/hadoop-ozone/pull/11

Regular CI:
https://github.com/adoroszlai/hadoop-ozone/actions/runs/658416539


elek (commented):
Thanks to create this patch @adoroszlai 

I think it's very important to have it, especially as we use the shared resources of common apache organization.

One thing what I am not sure about it: using git modules.

By default, I agree, that using standard features is a good thing. But earlier, when I used git submodules I had some problems with the usability (warnings, additional commands, conflict resolutions). With this approach everybody will see the git subodules after an Ozone checkout even if this feature is required only for one specific github action.

Wouldn't be it better to just checkout directly a specific revision of the target repository without using submodules?

adoroszlai:
Thanks @elek for the review.

> One thing what I am not sure about it: using git modules.
> 
> By default, I agree, that using standard features is a good thing. But earlier, when I used git submodules I had some problems with the usability (warnings, additional commands, conflict resolutions). With this approach everybody will see the git subodules after an Ozone checkout even if this feature is required only for one specific github action.

I have no prior experience using submodules, but now read a bit about it and tried some commands.

I think warnings and additional commands apply only if I need to interact with the submodule (eg. if it's required for compilation).  Since this one is intended for CI only, most developers do not need that.  In this case `.github/actions/cancel-workflow-runs` will be just an empty dir when on a branch that already has this commit, and removed when checking out another, older branch.

Only if I run `git submodule init` and `git submodule update`, and then switch to another branch, will I run into warnings and untracked files:

```
warning: unable to rmdir '.github/actions/cancel-workflow-runs': Directory not empty
```

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.github/actions/
```

In this case I need to remember to `git submodule deinit --all` before the switch.

elek:
Yes, something like this. 

Still I am not sure the benefit of using git submodules here (even if airflow uses this pattern).

When I clone the repository. I don't need the source code of this optional tool. It is not required when Intellij indexes all the files, it's not required when I do a grep, it's not required when I execute any tool (like counting lines,...) and so on...

The linked repository is not governed by Ozone PMC and not an Apache project, even if can be included in the same local source tree (if git submodules are updated)  

My -- very subjective, I admit  -- feeling is using a normal checkout instead of submodule has a stronger separation:

```
jobs:
  cancel-duplicate-ci-runs:
    name: "Cancel duplicate CI runs"
    runs-on: ubuntu-18.04
    steps:
      - name: Checkout project
        uses: actions/checkout@v2
        with:
          repository: https://github.com/potiuk/cancel-workflow-runs.git
          ref:  a81b3c4d59c61e27484cfacdc13897dd908419c9
          persist-credentials: false
```          

But I should admit that this is more like a personal preference, or feeling. While I think there is no doubt that checkout provides better separation, I have no **strong technical** arguments against using submodules.

adoroszlai:
> When I clone the repository. I don't need the source code of this optional tool. It is not required when Intellij indexes all the files, it's not required when I do a grep, it's not required when I execute any tool (like counting lines,...) and so on...

The files for the submodule would not be present until/unless you manually execute the git command to fetch them.

> using a normal checkout instead of submodule has a stronger separation

Is checkout better than using the third-party action directly?

elek (approved):
> Is checkout better than using the third-party action directly?

No, I think checkout is the easiest. Thanks for the update.

To be honest I was not sure about `git submodules`. I was in a mode where I was thinking loudly without very strong arguments.

But I am pretty **sure** that the current solution (using whitelisted github action) is the simplest way to do it ;-)

adoroszlai:
Thanks @elek for the review.

PR #2054, opened 2021-03-17
Deleting at bucket root throws NPE

cchenax:

https://issues.apache.org/jira/browse/HDDS-4838

## How was this patch tested?

hdfs hdfs dfs -rm -r -skipTrash o3fs://bucket1.vol1/
WARN ozone.BasicOzoneFileSystem: Cannot delete root directory.

### DIFF
--- a/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
+++ b/hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
@@ -496,6 +496,11 @@ public class BasicOzoneFileSystem extends FileSystem {
     String key = pathToKey(f);
     boolean result;
 
+    if (f.isRoot()) {
+      LOG.warn("Cannot delete root directory.");
+      return false;
+    }
+
     if (status.isDirectory()) {
       LOG.debug("delete: Path is a directory: {}", f);
 


hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
```diff
@@ -496,6 +496,11 @@ public boolean delete(Path f, boolean recursive) throws IOException {
     String key = pathToKey(f);
     boolean result;
 
+    if (f.isRoot()) {
```
ayushtkn, line 499: There is a similar check below inside `innerDelete`-
```
    try {
      DeleteIterator iterator = new DeleteIterator(f, recursive);

      if (f.isRoot()) {
        return false;
      }
```
Does that not work? Or will that be redundant post this? 
In any case this should be inside `    if (status.isDirectory()) {` block

ayushtkn (changes requested):
Can you add a unit test covering this as well. ``TestBasicOzoneFileSystems`` would be a good place to add one

hadoop-ozone/ozonefs-common/src/main/java/org/apache/hadoop/fs/ozone/BasicOzoneFileSystem.java
cchenax: Ok，I will add the unit test

PR #2057, opened 2021-03-18
HDDS-4989. Decommission CLI should return details of nodes which fail

sodonnel:

With the current decommission / recommission / maintenance mode commands, you can pass a list of hosts to perform the operation on. If any of these hosts fail to enter the decommission / maintenance workflow, the command gives no feedback about the error. Some of the hosts can silently fail and the only way to know is to inspect the SCM log.

The most common way a host can fail, is if a node which is undergoing maintenance is instructed to go to decommission and vice versa as this is a transition which is not allowed.

This change will allow any failed nodes to feed back to the client. If the client detects that any of the nodes have failed, details will be written to stderr and the command exit code will be non-zero.

Note that even though the exit code is non-zero, the command may have partially worked.

Also note that the errors which are fed back are only around transitioning the node into the admin workflow - it is still possible for it to fail later for other reasons which will not be fed back to the client. This is because the client does not wait for the process to complete, but exits after confirmation the command has been processed by scm.

https://issues.apache.org/jira/browse/HDDS-4989

## How was this patch tested?

New unit tests and validated manually using docker

### DIFF
--- a/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/datanode/MaintenanceSubCommand.java
+++ b/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/datanode/MaintenanceSubCommand.java
@@ -19,6 +19,7 @@ package org.apache.hadoop.hdds.scm.cli.datanode;
 
 import org.apache.hadoop.hdds.cli.GenericCli;
 import org.apache.hadoop.hdds.cli.HddsVersionProvider;
+import org.apache.hadoop.hdds.scm.DatanodeAdminError;
 import org.apache.hadoop.hdds.scm.cli.ScmSubcommand;
 import org.apache.hadoop.hdds.scm.client.ScmClient;
 import picocli.CommandLine;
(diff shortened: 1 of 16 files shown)


hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/datanode/MaintenanceSubCommand.java
```diff
@@ -51,9 +52,20 @@
...
       System.out.println("Entering maintenance mode on datanode(s):\n" +
           String.join("\n", hosts));
+      if (errors.size() > 0) {
+        for (DatanodeAdminError error : errors) {
+          System.err.println("Error: " + error.getHostname() +": "
+              + error.getError());
+        }
+        // Throwing the exception will cause a non-zero exit status for the
+        // command.
+        throw new IOException(
```
fapifta, line 66: I am not sure if this is the best way, but at the moment as I understood GenericCLI we might not have any other way. We may file a follow up JIRA to utilize picocli facilities for returning an exit code instead of throwing an exception here.

fapifta:
The changes looks good to me, I have added a comment on the code, I think we should file a follow up JIRA, please let me know what do you think?

+1 for committing the changes.

swagle (approved):

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/datanode/MaintenanceSubCommand.java
sodonnel: Yes I agree. I had the same thought and realised that the way we have the GenericCLI setup, it prevents us returning a specific non-zero code. For this soft-failure (some worked some failed), I would like to have returned a specific code.

However, I did not want to get into refactoring Generic CLI as part of this change. The other problem, is that we are somewhat short of unit tests for the CLI commands, so any change we make to that area will need new tests added and some manual validation.

I will raise a Jira with the suggestion tomorrow before committing this. Thanks for the review!

PR #2058, opened 2021-03-18
HDDS-4993. Add guardrail for reserved buffer size when DN reads a chunk

adoroszlai:

Currently DN reserves a byte buffer for client's chunk read request based on whatever client specifies, without check.  Client can send a message forcing DN to allocate up to 2GB memory per request.

This change makes DN reject read requests if chunk length specified by client is over the maximum allowed chunk size (32MB).  Normally this should not happen, as chunk info is fetched before read from the server.

https://issues.apache.org/jira/browse/HDDS-4993

## How was this patch tested?

Added unit test.

https://github.com/adoroszlai/hadoop-ozone/actions/runs/664486050


elek (approved):
+1

Thanks for the patch. Looks reasonable adding to this additional validation.

adoroszlai:
Thanks @elek for the review.

PR #2062, opened 2021-03-18
HDDS-4553. ChunkInputStream should release buffer as soon as last byte in the buffer is read

hanishakoneru:

We currently wait for reading up till the Chunk EOF before releasing the buffers in ChunkInputStream (or when the stream is closed). Let's say a client reads first 3 MB of a chunk  of size 4MB and does not close the stream immediately. This would lead to the 3MB of data being cached in the ChunkInputStream buffers till the stream is closed.

After HDDS-4552, a chunk read from DN would return the chunk data as an array of ByteBuffers. After each ByteBuffer is read, it can be released. This would greatly help with optimizing memory usage of ChunkInputStream.

https://issues.apache.org/jira/browse/HDDS-4553

## How was this patch tested?

Added unit tests

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/ChunkInputStream.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/ChunkInputStream.java
@@ -21,7 +21,6 @@ package org.apache.hadoop.hdds.scm.storage;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import java.util.ArrayList;
-import java.util.stream.Collectors;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hadoop.fs.CanUnbuffer;
 import org.apache.hadoop.fs.Seekable;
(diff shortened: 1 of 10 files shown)


hanishakoneru:
@adoroszlai, I fixed the review comments you posted in HDDS-4552 here. Please take a look. Thanks.

adoroszlai:
> I fixed the review comments you posted in [HDDS-4552](https://issues.apache.org/jira/browse/HDDS-4552) [here](https://github.com/apache/ozone/pull/2062/commits/42f29eba3955a5aef36c26d1b17473d78abc402b)

Thanks.  That part looks good.  I'll check the rest later.

umamaheswararao:
@hanishakoneru Thank you for the patch. I have posted few minor comments. Please check when you get time. Thanks

hanishakoneru:
Thank you @umamaheswararao for the review. I have addressed your comments in the latest commit.

hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/storage/ChunkInputStream.java
```diff
@@ -628,12 +645,38 @@ private boolean chunkStreamEOF() {
     }
   }
 
+
+  /**
+   * Release a buffers upto the given index.
```
umamaheswararao, line 650: Tiny nit: make sure to change it before commit
Release a buffers --> Release the buffers

umamaheswararao (approved):
Thank latest changes looks good to me. 
Before please make sure to update tiny nit comment.

bshashikant (approved):
Looks good. Thanks @hanishakoneru for working on this.

hanishakoneru:
Thank you @umamaheswararao and @bshashikant for the reviews.

PR #2063, opened 2021-03-19
HDDS-4998. [SCM HA Security] Make storeValidCertificate method idempotent

bharatviswa504:

Make storeValidCertificate idempotent.

Steps done:
1. Move checkValidCertID to the caller.
2. Perform signAndStoreCertificate under lock. (As doing sign only under lock, so it is not a huge overhead, as anyway writing to DB is done already under lock. And this approach will change in HDDS-4998. This is a temporary fix, until we have proper solution which can work across SCM nodes. This is required because without this for the secure cluster with ratis enabled after a restartt SCM will fail to start.)

https://issues.apache.org/jira/browse/HDDS-4998

## How was this patch tested?

Existing tests.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultApprover.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultApprover.java
@@ -19,7 +19,6 @@
 
 package org.apache.hadoop.hdds.security.x509.certificate.authority;
 
-import org.apache.commons.lang3.RandomUtils;
 import org.apache.hadoop.hdds.security.exception.SCMSecurityException;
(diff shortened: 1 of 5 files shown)


bshashikant (approved):

bharatviswa504:

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultApprover.java
```diff
@@ -155,6 +156,12 @@ public  X509CertificateHolder sign(
 
   }
 
+  public long generateSerialId() {
+    // TODO: to make generation of serialId distributed.
+    // This issue will be fixed in HDDS-4999.
```
GlenGeng, line 161: Please check `SequenceIdGenerator`.

PR #2064, opened 2021-03-19
HDDS-5003. Introduce EC replication type

elek:

Introduce the EC replication type. This is required by multiple other patches, it can be easier to add it as a separated patch.

For example having this EC in the enum the EcPipelineProvider can be started to implement (though ReplicationConfig is also required long-term.

https://issues.apache.org/jira/browse/HDDS-5003

## How was this patch tested?

It's a simple change, any incompatibilities should be detected by unit tests.

Also tested with end2end key creation with https://github.com/elek/ozone/tree/ec (where this is the first commit)


dineshchitlangia (approved):
+1 pending CI

elek:
Thanks for the review @dineshchitlangia, I am merging it now...

PR #2066, opened 2021-03-21
HDDS-5006. Use Log4j to print error messages for Datanode startup

symious:

Use Log4j to print error messages for Datanode startup, instead of System.err

https://issues.apache.org/jira/browse/HDDS-5006

## How was this patch tested?

The updated result is as follows, which will be much easier for the users to find the root cause of Datanode failing to start:
Screenshot 2021-03-21 at 10 27 13 PM

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/HddsDatanodeService.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/HddsDatanodeService.java
@@ -587,4 +587,15 @@ public class HddsDatanodeService extends GenericCli implements ServicePlugin {
   public void setCertificateClient(CertificateClient client) {
     dnCertClient = client;
   }
+
+  @Override
+  public void printError(Throwable error) {
+    if (error.getMessage() == null || error.getMessage().length() == 0) {
+      //message could be null in case of NPE. This is unexpected so we can
+      //print out the stack trace.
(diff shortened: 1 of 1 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/HddsDatanodeService.java
```diff
@@ -587,4 +587,15 @@ public CertificateClient getCertificateClient() {
...
   }
+
+  @Override
+  public void printError(Throwable error) {
+    if (error.getMessage() == null || error.getMessage().length() == 0) {
+      //message could be null in case of NPE. This is unexpected so we can
+      //print out the stack trace.
+      LOG.error("Exception in HddsDatanodeService.", error);
+    } else {
+      LOG.error(error.getMessage().split("\n")[0]);
```
ayushtkn, line 598: Guess can keep this only for both the cases :
```
 LOG.error("Exception in HddsDatanodeService.", error);
```

Sometimes having trace of the exception really helps

symious:
@ayushtkn Thanks for the review, the new output is as follows.
Screenshot 2021-03-23 at 6 42 57 AM

ayushtkn (approved):
LGTM

adoroszlai (approved):
Thanks @symious for the improvement, @ayushtkn for the review.

PR #2068, opened 2021-03-22
HDDS-5011. Introduce ECReplicationConfig and Java based ReplicationConfig implementation

elek:

SCM proto file should be extended to use ECReplicationConfig which can be de-/serialized to a specific ReplicationConfiguration.

Note: this is the bare minimum version of HDDS-4882 #1973 which doesn't include the refactor of the existing proto/persistent fields but de-/serialize them to the new java pojos.

This PR is marked as draft as it depends on #2064 (this PR includes that patch, too)

https://issues.apache.org/jira/browse/HDDS-5011

## How was this patch tested?

With new unit tests and also with end2end testing based on branch https://github.com/elek/ozone/tree/ec where this (and #2064) are two patches from the commit list.


hadoop-hdds/interface-server/src/main/proto/ScmServerProtocol.proto
```diff
@@ -132,9 +132,13 @@ message AllocateScmBlockRequestProto {
   required uint64 size = 1;
   required uint32 numBlocks = 2;
   required ReplicationType type = 3;
-  required hadoop.hdds.ReplicationFactor factor = 4;
+  optional hadoop.hdds.ReplicationFactor factor = 4;
```
sodonnel, line 135: Is moving this field from required to optional OK from a forward / backwards compatibility point of view? Would an old client which has it set to required, complain if it received a message missing the field?

Old Client -> New Server:
Will only ever send the factor field | will be OK

New Server -> Old Client
Client can never understand EC. Other files would have factor set.

New Client -> Old Server
Old server cannot understand EC, must always send Factor for Ratis files.

Old Server -> New Client:
Will always sent factor

I think it will be OK, as there is a natural incompatibility between old client / server and new client / server which supports EC.

elek: > Would an old client which has it set to required, complain if it received a message missing the field?

Good question and I agree that it should be OK. This field will always be filled for STANDALONE / RATIS. Old clients remain compatible as long as they use only existing replication types. It is optional only for EC where we need an exception from the old client anyway, as old client couldn't use new replication types.

sodonnel (approved):
These changes LGTM.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/RatisReplicationConfig.java
```diff
@@ -0,0 +1,82 @@
...
+  @Override
+  public String toString() {
+    return "RATIS/" + replicationFactor;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(replicationFactor);
+  }
+}
```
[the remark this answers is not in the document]
elek: `RatisReplictionConfig` is not serialized anymore to any `***ReplicationConfig` proto structure. Only the replication factor is required which can be retrieved with either casting or using the `getLegacyFactor` helper method.

elek:
Planning to merge it tomorrow as there are no more concerns. @umamaheswararao let me know if you have any comments.

umamaheswararao (approved):
LGTM

elek:
Merging it now. Thanks the review @sodonnel and @umamaheswararao

PR #2070, opened 2021-03-22
HDDS-4976. Add container replica related commands to debug

guihecheng:

Add a subcommand to debug to get datanode local container info,
get as much information as possible, a bit different from `admin container info`

https://issues.apache.org/jira/browse/HDDS-4976

## How was this patch tested?

manual tests

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainerData.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainerData.java
@@ -18,6 +18,7 @@
 
 package org.apache.hadoop.ozone.container.keyvalue;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 
@@ -251,6 +252,7 @@ public class KeyValueContainerData extends ContainerData {
    * @return Protocol Buffer Message
    */
   @Override
+  @JsonIgnore
   public ContainerDataProto getProtoBufMessage() {
     ContainerDataProto.Builder builder = ContainerDataProto.newBuilder();
     builder.setContainerID(this.getContainerID());
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/OzoneDebug.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/OzoneDebug.java
@@ -21,6 +21,7 @@ package org.apache.hadoop.ozone.debug;
 import org.apache.hadoop.hdds.cli.GenericCli;
 import org.apache.hadoop.hdds.cli.HddsVersionProvider;
 
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import picocli.CommandLine;
 
 /**
@@ -32,8 +33,17 @@ import picocli.CommandLine;
         mixinStandardHelpOptions = true)
 public class OzoneDebug extends GenericCli {
 
+  private OzoneConfiguration ozoneConf;
+
   public OzoneDebug() {
     super(OzoneDebug.class);
+  }
+
+  public OzoneConfiguration getOzoneConf() {
+    if (ozoneConf == null) {
+      ozoneConf = createOzoneConfiguration();
+    }
+    return ozoneConf;
   }
 
   /**
--- a/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/container/ContainerCommands.java
+++ b/hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/container/ContainerCommands.java
@@ -16,10 +16,39 @@
  *  limitations under the License.
  */
 
-package org.apache.hadoop.ozone.debug;
+package org.apache.hadoop.ozone.debug.container;
+
+import com.google.common.base.Preconditions;
+import org.apache.hadoop.hdds.cli.GenericCli;
+import org.apache.hadoop.hdds.cli.HddsVersionProvider;
+import org.apache.hadoop.hdds.cli.SubcommandWithParent;
+import org.apache.hadoop.hdds.conf.ConfigurationSource;
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
+import org.apache.hadoop.hdds.protocol.datanode.proto.ContainerProtos;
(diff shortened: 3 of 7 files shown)


hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ContainerUtils.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Container related utility functions.
+ */
+public final class ContainerUtils {
+
+  /** Never constructed. **/
+  private ContainerUtils() {
+  }
+
+  public static String getScmId(String storageDir) throws IOException {
```
GlenGeng, line 47: After HDDS-2823 is merged, the scmId will be replace with clusterId. Need take care this conflict.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/LocalContainerInfo.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.hadoop.ozone.container.keyvalue.KeyValueContainerData;
+import org.apache.hadoop.ozone.container.ozoneimpl.ContainerController;
+import org.apache.hadoop.ozone.container.ozoneimpl.ContainerReader;
+
+import org.kohsuke.MetaInfServices;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.ParentCommand;
+
+@Command(name = "localcontainerinfo",
```
GlenGeng, line 50: seems containerreplica is what you mean ?

GlenGeng:
please diff ExportContainer between HDDS-2823 and master.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/ContainerUtils.java
guihecheng: ok, then this pr should hold until we merged scm-ha.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/LocalContainerInfo.java
guihecheng: sounds a better name, thx~

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/LocalContainerInfo.java
```diff
@@ -0,0 +1,165 @@
...
+      result.addProperty("DeleteTransactionId",
+          kvData.getDeleteTransactionId());
+      result.addProperty("BlockCommitSequenceId",
+          kvData.getBlockCommitSequenceId());
+    default:
+      break;
+    }
+
+    Gson gson2 = new GsonBuilder().setPrettyPrinting().create();
+    System.out.println(gson2.toJson(result));
```
adoroszlai, line 162: Would it be possible to use `JsonUtils.toJsonStringWithDefaultPrettyPrinter` to produce JSON output with less code?

Example:

https://github.com/apache/ozone/blob/85af50b84f2ecba81ff6b67095c4cd082da7b59f/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/InfoSubcommand.java#L72

adoroszlai (commented):
Thanks @guihecheng for working on this.  I would like to suggest some changes:

1. To prevent proliferation of "random" subcommands under `ozone debug`, please group container-related debug commands under `ozone debug container`.  See [ContainerCommands](https://github.com/apache/ozone/blob/85af50b84f2ecba81ff6b67095c4cd082da7b59f/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/ContainerCommands.java) for an example of how to do it.  The description of `ozone debug container` could explain that its subcommands operate on local container replicas and should be executed on datanodes only.  Existing `export-container` command would become `ozone debug container export`.
2. To be consistent with other similar commands, I think we should separate the new operation into two subcommands: `ozone debug container list` and `... info`.  Container ID should be a required `@Parameter` for `info`.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/LocalContainerInfo.java
guihecheng: @adoroszlai Actually I've considered this util, it seems that we may have to let the JsonUtils ignore some fields in `ContainerData` and `KeyValueContainerData` with a java annotation, which will add lines of code into the two class files.

guihecheng:
> Thanks @guihecheng for working on this. I would like to suggest some changes:
> 
> 1. To prevent proliferation of "random" subcommands under `ozone debug`, please group container-related debug commands under `ozone debug container`.  See [ContainerCommands](https://github.com/apache/ozone/blob/85af50b84f2ecba81ff6b67095c4cd082da7b59f/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/container/ContainerCommands.java) for an example of how to do it.  The description of `ozone debug container` could explain that its subcommands operate on local container replicas and should be executed on datanodes only.  Existing `export-container` command would become `ozone debug container export`.

Sounds a good idea, I remenber Sammi Chen mentioned this idea to too, we'll do that together.

> 2. To be consistent with other similar commands, I think we should separate the new operation into two subcommands: `ozone debug container list` and `... info`.  Container ID should be a required `@Parameter` for `info`.

OK, it seems cleaner to have separated commands.

GlenGeng:
@guihecheng You can continue the work now after rebase master.

guihecheng:
@GlenGeng updated, thanks~

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/debug/LocalContainerInfo.java
guihecheng: Hi @adoroszlai , now we are using this JsonUtils to dump more detailed output~

guihecheng:
ping @adoroszlai

adoroszlai (approved):
Thanks @guihecheng for updating the patch.

guihecheng:
Hi @elek , could you help review this one?

elek (approved):
+1. Thanks for the patch @guihecheng and the review @adoroszlai 

Let me merge it now....

PR #2071, opened 2021-03-23
HDDS-5018.[FSO] Add robot tests for new Filesystem layout

sadanand48:
Add robot test for the new FS Layout

https://issues.apache.org/jira/browse/HDDS-5018

## How was this patch tested?
Acceptance tests


mukul1987 (commented):
The changes generally look good to me. Lets wait for the CI results

adoroszlai:
1. `.env` file is missing from the new compose directory, so acceptance test is failing:
    ```
    The OZONE_RUNNER_VERSION variable is not set. Defaulting to a blank string.
    no such image: apache/ozone-runner:: invalid reference format
    ```
2. Can we find a way to avoid duplicating the compose environment just for testing with slightly different config?

sadanand48:
Thanks @adoroszlai for the comment. 
By not duplicating the compose environment do you mean adding this test to an existing environment (inside ozone/test.sh) or not duplicate the docker-compose.yaml file?

adoroszlai:
> By not duplicating the compose environment do you mean adding this test to an existing environment (inside ozone/test.sh) or not duplicate the docker-compose.yaml file?

I don't have a definitive answer for that (that's what I meant by "find a way").  Hopefully we can achieve it in the first way, ie. without any duplication.

One idea is to define relevant configs as environment variables directly in `test.sh`, instead of the static `docker-config` file.  With that, we can execute the same test suite with different configs by including `start/stop_docker_env` in the loop.

adoroszlai:
> One idea is to define relevant configs as environment variables directly in `test.sh`, instead of the static `docker-config` file. With that, we can execute the same test suite with different configs by including `start/stop_docker_env` in the loop.

Here's an implementation of this:
https://github.com/adoroszlai/hadoop-ozone/commit/dd2d58637da74f524b643b7b240bf53f306905d6

And the test run:
https://github.com/adoroszlai/hadoop-ozone/runs/2192093559?check_suite_focus=true

sadanand48:
> Here's an implementation of this:
> [adoroszlai@dd2d586](https://github.com/adoroszlai/hadoop-ozone/commit/dd2d58637da74f524b643b7b240bf53f306905d6)
> 
> And the test run:
> https://github.com/adoroszlai/hadoop-ozone/runs/2192093559?check_suite_focus=true

Thanks  a lot @adoroszlai  for demonstrating this. I have tried something similar . Please take a look.

mukul1987:
@sadanand48 can you please rebase the patch.

mukul1987 (approved):
+1, the changes looks good to me.

rakeshadr:
Thanks @sadanand48 for the contribution.
Thanks @mukul1987 for the reviews.

+1 LGTM, will merge it.

PR #2072, opened 2021-03-23
HDDS-5015. localId is not consistent across SCMs when setup a multi node SCM HA cluster.

GlenGeng:

We set up the three node SCM HA cluster for test purpose.
From ozone dbug ldb tool, we found that the localIDs are not same accross the three SCM.

**The root cause here is:**
when localId is not set in the sequenceId table, SCM will initialize it to be UniqueId.next(). When setup 3 SCM from scratch, each of them will individually set their localId to be their own UniqueId.next(), thus the sequenceId is diverged from the very beginning.
 
**Short term solution is:**
make the 3 empty SCM achieve an agreement about the localId.
 
**Long tem solutos is:**
Check HDDS-5016.
During bootstrap, the new SCM always downloads checkpoint from leader SCM, and replace their own scm.db with that of leader.
 
**The short term solution is safe:**
upgrade in-memory scm to bypass-ratis scm: not affected.
upgrade in-memory scm to single-node scm: not affected.
upgrade in-memory scm to three-node scm cluster: not support yet, waits for long-term solution.
setup a bypass-ratis scm: not affected.
setup a three-node scm cluster from scratch: fix by the short term solution.

https://issues.apache.org/jira/browse/HDDS-5015

## How was this patch tested?

CI and real env test.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIdGenerator.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SequenceIdGenerator.java
@@ -33,9 +33,11 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.lang.reflect.Proxy;
+import java.time.LocalDate;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


bshashikant (approved):

GlenGeng:
Thanks @bshashikant for the review! I will merge this short term solution. The long term solution has been tracked by a new HDDS-5016.

PR #2074, opened 2021-03-23
HDDS-5010 Fixing Prometheus duplicated TYPE definition

hgruck:
Currently, PrometheusMetricsSink adds **# TYPE** definition to each metrics line even though the metrics are repeatedly appearing in the endpoint due to differing in labels, e.g.
```
# TYPE hdds_dispatcher_counter counter
hdds_dispatcher_counter{type="PutBlock",hostname="host1"} 0
# TYPE hdds_dispatcher_counter counter
hdds_dispatcher_counter{type="ReadContainer",hostname="host1"} 0
```

Expected format is:
```
# TYPE hdds_dispatcher_counter counter
hdds_dispatcher_counter{type="PutBlock",hostname="host1"} 0
hdds_dispatcher_counter{type="ReadContainer",hostname="host1"} 0
```

https://issues.apache.org/jira/browse/HDDS-5010

## How was this patch tested?
Unit test added + manually.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/PrometheusMetricsSink.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/PrometheusMetricsSink.java
@@ -25,15 +25,13 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.regex.Pattern;
-
+import org.apache.commons.configuration2.SubsetConfiguration;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.metrics2.AbstractMetric;
 import org.apache.hadoop.metrics2.MetricType;
 import org.apache.hadoop.metrics2.MetricsRecord;
 import org.apache.hadoop.metrics2.MetricsSink;
 import org.apache.hadoop.metrics2.MetricsTag;
-
-import org.apache.commons.configuration2.SubsetConfiguration;
 
 /**
  * Metrics sink for prometheus exporter.
@@ -74,12 +72,9 @@ public class PrometheusMetricsSink implements MetricsSink {
             + " "
             + metrics.type().toString().toLowerCase();
 
-        if (!metricLines.containsKey(metricKey)){
-          metricLines.put(metricKey, new TreeMap<>());
-        }
-
-        metricLines.get(metricKey).put(prometheusMetricKeyAsString,
-            String.valueOf(metrics.value()));
+        metricLines.computeIfAbsent(metricKey,
+            any -> Collections.synchronizedSortedMap(new TreeMap<>()))
+            .put(prometheusMetricKeyAsString, String.valueOf(metrics.value()));
       }
     }
   }
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/PrometheusMetricsSink.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/PrometheusMetricsSink.java
@@ -21,8 +21,9 @@ import static org.apache.hadoop.hdds.utils.RocksDBStoreMBean.ROCKSDB_CONTEXT_PRE
 
 import java.io.IOException;
 import java.io.Writer;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/PrometheusMetricsSink.java
```diff
@@ -64,43 +66,48 @@ public void putMetrics(MetricsRecord metricsRecord) {
...
+        String prometheusMetricKeyAsString =
+            getPrometheusMetricKeyAsString(metricsRecord, key);
+
+        String metricKey = "# TYPE "
+            + key
+            + " "
+            + metrics.type().toString().toLowerCase();
+
+        if (!metricLines.containsKey(metricKey)){
+          metricLines.put(metricKey, new TreeMap<>());
```
elek, line 78:
```suggestion
          metricLines.put(metricKey, Collections.synchronizedSortedMap(new TreeMap<>()));
```

Fix me if I am wrong, but I think the sub maps should also be synchronized.

elek (changes requested):
Thanks the patch @hgruck. It looks good to me.

 (except one small comment: I think all the key specific maps also should be synchronized as `PrometheusMetricsSink` is shared between multiple client requests.)
 
 ((The integration test `TestNodeFailure` seems to have an intermittent problem, locally it was passed. Hope we will have a full green build with the next commit))

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/PrometheusMetricsSink.java
```diff
@@ -64,43 +66,48 @@ public void putMetrics(MetricsRecord metricsRecord) {
...
         }
-        prometheusMetricKey.append("}");
 
-        String prometheusMetricKeyAsString = prometheusMetricKey.toString();
-        builder.append(prometheusMetricKeyAsString);
-        builder.append(" ");
-        builder.append(metrics.value());
-        builder.append("\n");
-        metricLines.put(prometheusMetricKeyAsString, builder.toString());
+        metricLines.get(metricKey).put(prometheusMetricKeyAsString,
```
adoroszlai, line 81: Please use `computeIfAbsent` instead of `containsKey`, `put` and `get`.  Only individual calls are synchronized, so the sequence may lead to unexpected results.

```java
    metricLines.computeIfAbsent(metricKey, any -> synchronizedSortedMap(new TreeMap<>()))
        .put(prometheusMetricKeyAsString, builder.toString());
```

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/PrometheusMetricsSink.java
hgruck: Thanks a lot for the suggestion, I've made the changes based on @adoroszlai idea in a534cc2.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/PrometheusMetricsSink.java
hgruck: Good idea, I've added the changes in a534cc2.

elek (approved):
+1, thanks for the update

adoroszlai (approved):
Thanks @hgruck for updating the patch.

```
# TYPE hdds_dispatcher_counter counter
...
hdds_dispatcher_counter{type="PutBlock",hostname="490d015494f1"} 128
hdds_dispatcher_counter{type="PutSmallFile",hostname="490d015494f1"} 0
hdds_dispatcher_counter{type="ReadChunk",hostname="490d015494f1"} 44
hdds_dispatcher_counter{type="ReadContainer",hostname="490d015494f1"} 0
hdds_dispatcher_counter{type="UpdateContainer",hostname="490d015494f1"} 0
hdds_dispatcher_counter{type="WriteChunk",hostname="490d015494f1"} 256
```

Output from datanode LGTM.

elek:
Merging it now, thanks the contribution @hgruck and the review @adoroszlai

PR #2075, opened 2021-03-23
HDDS-5019. Inconsistent tmp bucket name on different nodes when Kerberos is enabled

zhengchenyu:

I setup yarn cluster on ozone, and kerberos is enable. NodeManager can't download resource package in ContainerLocalizer stage, show error like this

```
2021-03-22 19:56:42,004 WARN org.apache.hadoop.yarn.server.nodemanager.containermanager.localizer.ResourceLocalizationService: \{ ofs://test1/tmp/hadoop-yarn/staging/$user/.staging/job_1616412423610_0008/job.splitmetainfo, 1616414199202, FILE, null } failed: hadoop-yarn/staging/$user/.staging/job_1616412423610_0008/job.splitmetainfo: No such file or directory!
java.io.FileNotFoundException: hadoop-yarn/staging/$user/.staging/job_1616412423610_0008/job.splitmetainfo: No such file or directory!
```
 

The reason is that this wrong way to get user. For kerberos principal in different machine are not same, so UserGroupInformation.getCurrentUser().getUserName() are not same. Because kerberos principal's formate is "user/host@HADOOP.COM"
 

https://issues.apache.org/jira/browse/HDDS-5019

## How was this patch tested?

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OFSPath.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OFSPath.java
@@ -285,7 +285,7 @@ public class OFSPath {
    */
   public static String getTempMountBucketNameOfCurrentUser()
       throws IOException {
-    String username = UserGroupInformation.getCurrentUser().getUserName();
+    String username = UserGroupInformation.getCurrentUser().getShortUserName();
     return getTempMountBucketName(username);
   }
 


avijayanhwx (commented):
Thanks for fixing this @zhengchenyu. 

In the bucket create flow, is the bucket name created with the "full" name? Can you comment on how this affects existing deployments? 

cc @smengcl

smengcl:
Thanks @zhengchenyu for pointing this out.

This is an oversight. When OFS was implemented, Kerberos principal name (`getUserName()`) are actually used intentionally.
(For temp bucket names, to avoid any invalid characters in the user name, a [hash](https://github.com/apache/ozone/blob/4128813ab495dfd3941c0252e61e41ca7d1cf1ce/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/OFSPath.java#L283) of the user name is used instead.)

For the use case (different machines have different Kerberos principals but maps to the same short name), I agree that short name should be used instead of Kerberos principal.

I'm +1 on this. This also raises other questions that all other calls to `getUserName()` needs to be checked to avoid the same problem.

CC @xiaoyuyao

smengcl (approved):

smengcl:
> Thanks for fixing this @zhengchenyu.
> 
> In the bucket create flow, is the bucket name created with the "full" name? Can you comment on how this affects existing deployments?
> 
> cc @smengcl

@avijayanhwx Right now when Kerberos is enabled, tmp bucket name is the hash of Kerberos principal. With the change it becomes the hash of short user names which the Kerberos principal is mapped to.

avijayanhwx:
> > Thanks for fixing this @zhengchenyu.
> > In the bucket create flow, is the bucket name created with the "full" name? Can you comment on how this affects existing deployments?
> > cc @smengcl
> 
> @avijayanhwx Right now when Kerberos is enabled, tmp bucket name is the hash of Kerberos principal. With the change it becomes the hash of short user names which the Kerberos principal is mapped to.

Thanks for the clarification. This means that existing users will need to recreate/move the data from tmp directory to continue with their application. Is that correct?

smengcl:
> > > Thanks for fixing this @zhengchenyu.
> > > In the bucket create flow, is the bucket name created with the "full" name? Can you comment on how this affects existing deployments?
> > > cc @smengcl
> > 
> > 
> > @avijayanhwx Right now when Kerberos is enabled, tmp bucket name is the hash of Kerberos principal. With the change it becomes the hash of short user names which the Kerberos principal is mapped to.
> 
> Thanks for the clarification. This means that existing users will need to recreate/move the data from tmp directory to continue with their application. Is that correct?

Yes if users use `/tmp` to persist stuff (which they shouldn't, in theory) because their tmp bucket name changes with this patch.

zhengchenyu:
> Thanks for fixing this @zhengchenyu.
> 
> In the bucket create flow, is the bucket name created with the "full" name? Can you comment on how this affects existing deployments?
> 
> cc @smengcl
Sorry for miss this message. As smengcl said, different machines have different Kerberos principals but maps to the same short name. This PR only solve the problem "same user in different host can't process same file in tmp volume". Maybe another calls to getUserName() should be fix.

zhengchenyu:
As @smengcl guess, I found it's a common problem. 
I excute command "ozone sh volume create volume1" in host1. But there is no volume1 info When I execute "ozone sh volume list" in host2.

avijayanhwx:
Thank you for the response @smengcl & @zhengchenyu.

PR #2077, opened 2021-03-23
HDDS-4987. Import container should not delete container contents if container already exists

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-4987

Import container should also delete the whole container directory in case of import failure.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
@@ -466,7 +466,8 @@ public class KeyValueContainer implements Container<KeyValueContainerData> {
                 + " as the container descriptor (%s) has already been exist.",
             getContainerData().getContainerID(),
             getContainerFile().getAbsolutePath());
-        throw new IOException(errorMessage);
+        throw new StorageContainerException(errorMessage,
+            CONTAINER_ALREADY_EXISTS);
       }
       //copy the values from the input stream to the final destination
       // directory.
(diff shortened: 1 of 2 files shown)


hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/keyvalue/TestKeyValueContainer.java
```diff
@@ -206,8 +206,32 @@ public void testContainerImportExport() throws Exception {
...
+            keyValueContainerData.getLayOutVersion(),
+            keyValueContainerData.getMaxSize(), UUID.randomUUID().toString(),
+            datanodeId.toString());
+    container = new KeyValueContainer(containerData, CONF);
+
+    containerVolume = volumeChoosingPolicy.chooseVolume(volumeSet
+        .getVolumesList(), 1);
+    hddsVolumeDir = containerVolume.getHddsRootDir().toString();
+    container.populatePathFields(scmId, containerVolume, hddsVolumeDir);
+    try (FileInputStream fis = new FileInputStream(folderToExport)) {
```
GlenGeng, line 224: since you close `fis` explicitly, no need try with resource. Just

```
try {
  FileInputStream fis = new FileInputStream(folderToExport);
  fis.close();
}
```

GlenGeng (commented):
LGTM.

GlenGeng:
+1. Thanks @ChenSammi  for the contribution.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/KeyValueContainer.java
```diff
@@ -496,11 +497,17 @@ public void importContainerData(InputStream input,
       KeyValueContainerUtil.parseKVContainerData(containerData, config);
 
     } catch (Exception ex) {
+      if (ex instanceof StorageContainerException &&
```
GlenGeng, line 500: NIT. Following may be better to handle the exception cases.

```
} catch (StorageContainerException ex) {
  throw ex;
} catch (Exception ex) {
  //delete all the temporary data in case of any exception.
  ...
}
```

ChenSammi:
@GlenGeng , thanks for the code review.

PR #2078, opened 2021-03-23
HDDS-5009. Set the default output format based on the newest version …

sky76093016:

1.Set the default output format.
2.Add a note about the output format reminder of the new and old versions of async profiler.

https://issues.apache.org/jira/browse/HDDS-5009

## How was this patch tested?

Local test.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/ProfileServlet.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/server/http/ProfileServlet.java
@@ -107,6 +107,10 @@ import org.slf4j.LoggerFactory;
  * // Java events:
  * //    alloc
  * //    lock
+ * Note for version usage:
+ * The default output format of the newest async profiler is HTML.
+ * If the user is using an older version such as 1.5, HTML is not supported.
+ * Please specify the corresponding output format.
  */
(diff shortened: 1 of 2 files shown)


jojochuang (approved):
I think this is good. Confirmed that async profiler 1.8.5 does support "html" output, and async profiler 2.0 too. So this is compatible across v1 and v2.

adoroszlai (commented):
I'd like to suggest waiting for new `ozone-runner` image with Async Profiler 2.0 (see https://github.com/apache/ozone-docker-runner/pull/5), so that we can also upgrade `docker.ozone-runner.version` in this change here:

https://github.com/apache/ozone/blob/7b12f00ff56ca53bb47343505e40f336457dc387/hadoop-ozone/dist/pom.xml#L31

elek:
new ozone-runner (with updated async-profile) is available. Please change the suggested variable in the `pom.xml` to `20210329-1`

sky76093016:
Thanks for the reminder from @elek , I will modify it as soon as possible.

cku328 (approved):
Thanks @sky76093016 for the contribution.

adoroszlai:
@cku328 When squashing PR with only one commit, its message is used as is by default.  In this case one might need to tweak it to include Jira number etc.

PR #2082, opened 2021-03-25
HDDS-4982.Solve intellj warnings on DBTransactionBuffer

Xushaohong:
The root cause of this warning is due to misleadingly using generic classes instead of generic methods.
The generic class specifies the real type when instantiating,  and the generic method specifies the real type when it is being called.
There are two methods under DBTransactionBuffer, the method` addToBuffer` and `removeFromBuffer`.
According to where these methods call, they are actually designed to write keys into separate tables or remove them. 
The key and value of different table are different, and so are the types. This is where generic methods work.
Thus I modified the generic methods and erased the generic class.

https://issues.apache.org/jira/browse/HDDS-4982

## How was this patch tested?
CI

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/metadata/DBTransactionBuffer.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/metadata/DBTransactionBuffer.java
@@ -25,11 +25,11 @@ import java.io.IOException;
 /**
  * DB transaction that abstracts the updates to the underlying datastore.
  */
-public interface DBTransactionBuffer<KEY, VALUE> extends Closeable {
+public interface DBTransactionBuffer extends Closeable {
 
-  void addToBuffer(Table<KEY, VALUE> table, KEY key, VALUE value) throws
-      IOException;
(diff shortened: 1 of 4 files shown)


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/metadata/DBTransactionBuffer.java
```diff
@@ -25,11 +25,11 @@
...
-public interface DBTransactionBuffer<KEY, VALUE> extends Closeable {
+public interface DBTransactionBuffer extends Closeable {
 
-  void addToBuffer(Table<KEY, VALUE> table, KEY key, VALUE value) throws
-      IOException;
+  <KEY, VALUE> void addToBuffer(Table<KEY, VALUE> table, KEY key, VALUE value)
+      throws IOException;
 
-  void removeFromBuffer(Table<KEY, VALUE> table, KEY key) throws
+  <KEY, VALUE> void removeFromBuffer(Table<KEY, VALUE> table, KEY key) throws
```
GlenGeng, line 33: `throws IOException;`

GlenGeng (commented):
LGTM.

GlenGeng:
@bshashikant Please take a look.

GlenGeng:
+1

PR #2083, opened 2021-03-25
HDDS-4818. Add command to get status of Finalize in OM & SCM

guihecheng:

Add command to get status of Finalize in OM & SCM

https://issues.apache.org/jira/browse/HDDS-4818

## How was this patch tested?

a new ut
manual test

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
@@ -92,6 +92,11 @@ public abstract class BasicUpgradeFinalizer
     return new StatusAndMessages(status, returningMsgs);
   }
 
+  @Override
+  public synchronized Status getStatus() {
+    return versionManager.getUpgradeState();
+  }
+
   protected void preFinalizeUpgrade(T service) throws IOException {
     // No Op by default.
   }
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/hdds/upgrade/TestHDDSUpgrade.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/hdds/upgrade/TestHDDSUpgrade.java
@@ -410,7 +410,7 @@ public class TestHDDSUpgrade {
 
     // Wait for the Finalization to complete on the SCM.
     while (status.status() != FINALIZATION_DONE) {
-      status = scm.queryUpgradeFinalizationProgress("xyz", false);
+      status = scm.queryUpgradeFinalizationProgress("xyz", false, false);
     }
 
     Set<PipelineID> postUpgradeOpenPipelines =
@@ -1018,7 +1018,7 @@ public class TestHDDSUpgrade {
     while ((status.status() != FINALIZATION_DONE) &&
         (status.status() != ALREADY_FINALIZED)) {
       loadSCMState();
-      status = scm.queryUpgradeFinalizationProgress("xyz", true);
+      status = scm.queryUpgradeFinalizationProgress("xyz", true, false);
       if (status.status() == FINALIZATION_REQUIRED) {
         status = scm.finalizeUpgrade("xyz");
       }
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/client/ScmClient.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/client/ScmClient.java
@@ -340,5 +340,6 @@ public interface ScmClient extends Closeable {
       throws IOException;
 
   StatusAndMessages queryUpgradeFinalizationProgress(
-      String upgradeClientID, boolean force) throws IOException;
+      String upgradeClientID, boolean force, boolean readonly)
+      throws IOException;
 }
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocol.java
@@ -304,5 +304,6 @@ public interface StorageContainerLocationProtocol extends Closeable {
       throws IOException;
 
   StatusAndMessages queryUpgradeFinalizationProgress(
-      String upgradeClientID, boolean force) throws IOException;
+      String upgradeClientID, boolean force, boolean readonly)
+      throws IOException;
 }
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/UpgradeFinalizer.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/UpgradeFinalizer.java
@@ -177,6 +177,11 @@ public interface UpgradeFinalizer<T> {
   StatusAndMessages reportStatus(String upgradeClientId, boolean takeover)
       throws IOException;
 
+  /**
+   * Get a readonly status of the finalization.
+   * @return the status of the finalization
+   */
+  Status getStatus();
 
   /**
    * Runs the set of pre finalized state validations and actions that need to
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocolPB/StorageContainerLocationProtocolClientSideTranslatorPB.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocolPB/StorageContainerLocationProtocolClientSideTranslatorPB.java
@@ -746,12 +746,14 @@ public final class StorageContainerLocationProtocolClientSideTranslatorPB
 
   @Override
   public StatusAndMessages queryUpgradeFinalizationProgress(
-      String upgradeClientID, boolean force) throws IOException {
+      String upgradeClientID, boolean force, boolean readonly)
+      throws IOException {
     QueryUpgradeFinalizationProgressRequestProto req =
         QueryUpgradeFinalizationProgressRequestProto.
             newBuilder()
             .setUpgradeClientId(upgradeClientID)
             .setTakeover(force)
+            .setReadonly(readonly)
             .build();
 
     QueryUpgradeFinalizationProgressResponseProto response =
--- a/hadoop-hdds/interface-admin/src/main/proto/ScmAdminProtocol.proto
+++ b/hadoop-hdds/interface-admin/src/main/proto/ScmAdminProtocol.proto
@@ -431,6 +431,7 @@ message FinalizeScmUpgradeResponseProto {
 message QueryUpgradeFinalizationProgressRequestProto {
   required string upgradeClientId = 1;
   optional bool takeover = 2;
+  optional bool readonly = 3;
 }
 message QueryUpgradeFinalizationProgressResponseProto {
   required hadoop.hdds.UpgradeFinalizationStatus status = 1;
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
@@ -578,7 +578,7 @@ public final class StorageContainerLocationProtocolServerSideTranslatorPB
       throws IOException {
     StatusAndMessages progress =
         impl.queryUpgradeFinalizationProgress(request.getUpgradeClientId(),
-            request.getTakeover());
+            request.getTakeover(), request.getReadonly());
 
     UpgradeFinalizationStatus.Status protoStatus =
         UpgradeFinalizationStatus.Status.valueOf(progress.status().name());
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
@@ -668,8 +668,10 @@ public class SCMClientProtocolServer implements
 
   @Override
   public StatusAndMessages queryUpgradeFinalizationProgress(
-      String upgradeClientID, boolean force) throws  IOException {
-    return scm.queryUpgradeFinalizationProgress(upgradeClientID, force);
+      String upgradeClientID, boolean force, boolean readonly)
+      throws  IOException {
+    return scm.queryUpgradeFinalizationProgress(upgradeClientID, force,
+        readonly);
   }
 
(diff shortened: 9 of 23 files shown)


guihecheng:
Hi @avijayanhwx , I added a subcommand to om cli first, maybe you would like to have a look at it first and then, I'll do some changes and then go with SCM.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/FinalizeStatusSubCommand.java
```diff
@@ -0,0 +1,64 @@
...
+  )
+  private String omHost;
+
+  @Override
+  public Void call() throws Exception {
+    OzoneManagerProtocol client =
+        parent.createOmClient(omServiceId, omHost, false);
+    String upgradeClientID = "Upgrade-Client-" + UUID.randomUUID().toString();
+    UpgradeFinalizer.StatusAndMessages progress =
+        client.queryUpgradeFinalizationProgress(upgradeClientID, true);
```
avijayanhwx, line 60: Can we not "takeover" the finalization from another client, just to get status? If there is a finalization going on currently, then that will fail since the upgradeClientId has changed. Maybe we could pass in a readOnly flag, and use that in the server to just query status instead of taking over. What do you think?

avijayanhwx (commented):
Thanks for working on this @guihecheng. Left a comment inline.

avijayanhwx:
cc @errose28 for review.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/admin/om/FinalizeStatusSubCommand.java
guihecheng: @avijayanhwx oh, got it, I shall add some logic to diff a takeover and a read-only status, thx~

guihecheng: https://github.com/apache/ozone/blob/ed00db7a52bf1acce9a82b21642032b42c0fb223/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/UpgradeFinalizer.java#L169

As I read through the comments of the interface, I found that the assert on clientId might be optional by design?
Suppose we could just remove the assertion in `BasicUPgradeFinalizer` and let any client fetch the status of finalization? I'm not sure whether it will cause any problems?

avijayanhwx: @guihecheng It is optional at the interface level since we could have implementors (maybe recon or test finalizers that don't need it) I believe the takeover flag must remain since it used on the update path. The readOnly flag could just skip the assert on clientId. What do you think?

cc @fapifta who implemented this initially.

guihecheng: Hi @avijayanhwx , updated with a readonly flag, skip on assertion when readonly is true, thx~

errose28:
Thanks for working on this @guihecheng. The new OM command looks good, I think you can the PR out of draft mode.

guihecheng:
Thanks @errose28 , I'll push an update this week with a similar scm command and tests.

guihecheng:
@avijayanhwx @errose28 updated for review, thx~

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
```diff
@@ -145,12 +145,14 @@ public StatusAndMessages finalize(String upgradeClientID, T id)
...
   public synchronized StatusAndMessages reportStatus(
-      String upgradeClientID, boolean takeover
+      String upgradeClientID, boolean takeover, boolean readonly
   ) throws UpgradeException {
     if (takeover) {
       clientID = upgradeClientID;
     }
-    assertClientId(upgradeClientID);
+    if (!readonly) {
+      assertClientId(upgradeClientID);
```
fapifta, line 154: If we skip this assertion, then we need som changes later on this method.
The reportStatus logic was designed for one client only, and the plan was to track all the messages that should be supplied to the client, and when the client polls the status and reports it, we clean up the messages from the temporary in memory storage (msgs), if we allow multiple clients to poll the status then they start to race and steal messages from each other (see 5 lines down we do msgs.poll()). That is why if a takeover happens the original client fails due to this assertion. On the other hand if we do not clean the cumulated messages, then we will serve messages more than once to the original client.

Also if we just allow through a secondary client concurrently, then it will not be able to read all the messages.

In order for this to work as it was for the original client, and in the meantime to allow an other client to get the status messages, we will need an other in memory temporary storage, and keep the messages that are flushed to the original client in that, and serve the secondary client from that second in memory storage, so that it will be able to get the full status report. In this case it is still a question when we clean up these structures? At the end of the finalization? Some time later? At restart? I am hesitant because of the memory footprint, but this should not be more than a couple strings maybe a hundred...

fapifta:
Hi @guihecheng,

thank you for working on this, it seems to be a good improvement, I did throw the idea when working on the first implementation, as we thought about finalization as a not-too-long process, which will be monitored by an admin, but I can see the usefulness of this or the ease of understanding from the user's point of view compared to takeover.

I think there is a critical flaw remained in the implementation, I have added a detailed comment about it inline.

Besides that the changes looking good, however I have a linguistic suggestion, in the JIRA you have also mentione finalization-status, in the code we use finalizeStatus. I would go with finalizationStatus as the term on the CLI, and in the code as well, as we use the word finalization in other parts of the code. If you are up to changing this, I would be glad, though this is probably just a preference on my side.

guihecheng:
> Hi @guihecheng,
> 
> thank you for working on this, it seems to be a good improvement, I did throw the idea when working on the first implementation, as we thought about finalization as a not-too-long process, which will be monitored by an admin, but I can see the usefulness of this or the ease of understanding from the user's point of view compared to takeover.
> 
> I think there is a critical flaw remained in the implementation, I have added a detailed comment about it inline.

Oh, that's a really important point, I'll check the details.

> Besides that the changes looking good, however I have a linguistic suggestion, in the JIRA you have also mentione finalization-status, in the code we use finalizeStatus. I would go with finalizationStatus as the term on the CLI, and in the code as well, as we use the word finalization in other parts of the code. If you are up to changing this, I would be glad, though this is probably just a preference on my side.

Sure, I'll use finalizationStatus to rename the CLI. (At first I checked the finalizeupgrade CLI, so I picked that name.)

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/upgrade/BasicUpgradeFinalizer.java
guihecheng: @fapifta thanks for your comments, there is a problem indeed.
IMHO, the 'finalizationStatus' command just returns the status like a single word 'FINALIZATION_IN_PROGRESS' shall be enough, because we may need this command to do some scripting/autotesting work, or just give user a simple hint on the current status.
What about having to methods:

- reportStatus(void): just return the status which we can directly get from versionManager.
- reportStatusAndMsgs(String upgradeClientID, boolean takeover): the original 'reportStatus' method without any changes, also without the readonly flag we added here.

Then the 'finalizationStatus' command will use reportStatus, which is just enough.
And the 'finalizeUpgrade' command use reportStatusAndMsgs, the takeover behaviour remains untouched.

In this way, we don't bother handling the concurrent issue for the msgs queue.
Do you think it is workable?

fapifta: Hi @guihecheng,

ah, so the use case here is to poll the status from some kind of an automation as I understand. The proposal sounds reasonable to me, and it is way more easier than to ensure to properly serve the message queue contents to an unknown number of concurrent clients, I like it.

fapifta:
> Sure, I'll use finalizationStatus to rename the CLI. (At first I checked the finalizeupgrade CLI, so I picked that name.)

Thank you!

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/upgrade/TestOMUpgradeFinalizer.java
```diff
@@ -188,9 +198,11 @@ public void testFinalizationWithUpgradeAction() throws Exception {
     if (finalizer.isFinalizationDone()) {
       when(versionManager.getUpgradeState()).thenReturn(FINALIZATION_DONE);
     }
-    StatusAndMessages status = finalizer.reportStatus(CLIENT_ID, false);
+    StatusAndMessages status = finalizer.reportStatusAndMsgs(CLIENT_ID, false);
     assertEquals(FINALIZATION_DONE, status.status());
     assertFalse(status.msgs().isEmpty());
+    Status ret = finalizer.reportStatus();
+    assertEquals(status, ret);
```
fapifta, line 205: We need to assert equality of status.status(), and ret.
It might be more obvious if we rename the variable status to finalizerResponse or similar, but this is just an afterthought as I watch my code, to understand it from afar :) If you are up to do this rename, please rename the variable in other test methods as well.

fapifta:
Hi @guihecheng,

thank you for your dedication on working through this. I might misunderstand some parts of the suggestion. My false assumption was that the idea is to direct the query to a different method on the server side only, I did not expected a whole new proto request-response pair.

My first impression was that this might be an overkill, thinking it through, it is cleaner to have a new request-response and handle it, than having a flag in a message, so I am fine with this approach as well, though I leave it up for others to decide if adding a new request-response into the proto layer is something we can do anytime, as I am not sure if it has any negative consequences to have many message pair for similar purposes. If it does not, and no one raises a concern, then the patch looks good to me, after the junit failure is fixed. I left a comment on the problematic line.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/hdds/upgrade/TestHDDSUpgrade.java
```diff
@@ -1023,6 +1027,9 @@ public void testFinalizationWithFailuerInjectionHelper(
       }
     }
 
+    Status ret = scm.queryUpgradeFinalizationStatus();
+    Assert.assertEquals(FINALIZATION_DONE, ret);
```
fapifta, line 1031: Should the correct assertion to make be comparing status.status() to ret?

hadoop-ozone/ozone-manager/src/test/java/org/apache/hadoop/ozone/om/upgrade/TestOMUpgradeFinalizer.java
fapifta: We might also assert equality of FINALIZATION_DONE, and status.status{} here.

guihecheng: @fapifta thanks for the catch, I'll have better names for the variables in next push.

guihecheng:
@fapifta thanks for the detailed reviews, I also think that it is a cleaner way to have a separate proto pair, but a bit complex indeed.
Maybe @avijayanhwx could give some suggestions ? And I'll fix the problems in next push :)

avijayanhwx:
@guihecheng / @fapifta There are 3 status requests that we are trying to support here.

- Client who started finalization asks for status and messages
- Client who did not start finalization asks for status & messages with takeover = true
- Client just wants to know the general finalization status, irrespective of whether a finalization was started / completed or in progress.

I believe a single proto request with multiple arguments (or better a 'FinalizeUpgradeProgressRequestArgs' parameter in the proto request) is sufficient to handle the above cases. The UpgradeFinalizer interface can have a new getStatus API that just returns the status of finalization, along with the layout version (MLV, SLV).  Based on the args passed in, the request handler can call the correct API of the finalizer.

Maybe something like this?

> 
> message FinalizeUpgradeProgressRequestArgs {
>     optional string upgradeClientId = 1;
>    optional bool takeover = 2;
>     optional bool readOnly = 3;
> }

guihecheng:
> @guihecheng / @fapifta There are 3 status requests that we are trying to support here.
> 
> * Client who started finalization asks for status and messages
> * Client who did not start finalization asks for status & messages with takeover = true
> * Client just wants to know the general finalization status, irrespective of whether a finalization was started / completed or in progress.
> 
> I believe a single proto request with multiple arguments (or better a 'FinalizeUpgradeProgressRequestArgs' parameter in the proto request) is sufficient to handle the above cases. The UpgradeFinalizer interface can have a new getStatus API that just returns the status of finalization, along with the layout version (MLV, SLV). Based on the args passed in, the request handler can call the correct API of the finalizer.
> 
> Maybe something like this?
> 
> > message FinalizeUpgradeProgressRequestArgs {
> > optional string upgradeClientId = 1;
> > optional bool takeover = 2;
> > optional bool readOnly = 3;
> > }

So, there is an agreement that we shall not have new proto request & response pairs.
Instead, we add a flag into the existing request and change the "required" `upgradeClientId` to "optional".
And we shall put a new API "getStatus" only in the UpgradeFinalizer, with SLV & MLV (may be in a new class like StatusAndLayoutVersions).
Got it, thanks @avijayanhwx @fapifta

fapifta:
I think both solution can be fine, but what we are essentially talking about is the following:
- having one method with a couple of boolean flags as parameters, then some if conditions inside the method
- or having two different methods that serves their own purpose.

In this case this is a bit more complicated though, as we need to define proto messages and types in order to have different methods that serves their own purposes, so we have a lot of boilerplate here and there and with that the line of code overhead is way more bigger then just changing one proto message and add a few boolean flags, still I think it is cleaner, and I like it more.

I don't have a strong opinion though, and I can accept the more flags version as well, if the internal API remains clear, and based on the flags we map to the desired methods that are living only in the server side.
Even though, I don't have a strong opinion, we create precedent here, and my heart pulls me to the having different methods version, unless it has drawback I am ignorant of in the proto layer or regarding compatibility/perf or any other way more important stuff than code structure.

avijayanhwx (approved):
LGTM +1 (pending CI).

fapifta:
After CI is green, +1 from my side as well.

avijayanhwx:
@guihecheng Can we check if the CI failure is related?

guihecheng:
@avijayanhwx checked the failed case, but did not spot any related failures, it seems a compose suite `ozonesecure-mr` is unable to exit saftemode, Shall we try retrigger the CI ?

avijayanhwx:
@guihecheng Yes, please retrigger CI.

avijayanhwx:
@guihecheng Yes, please retrigger CI.

guihecheng:
@avijayanhwx @fapifta updated, thanks~

avijayanhwx:
Thank you for working on this patiently @guihecheng, and for the reviews @fapifta & @errose28. Merging this PR.

PR #2087, opened 2021-03-28
HDDS-5035. Use default config values to solve generated config file conflict

symious:

- Use default config values to initialize configurations

https://issues.apache.org/jira/browse/HDDS-5035

## How was this patch tested?

The user's assembled fat jar works fines now.

### DIFF
--- a/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/conf/TestGeneratedConfigurationOverwrite.java
+++ b/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/conf/TestGeneratedConfigurationOverwrite.java
@@ -42,9 +42,6 @@ public class TestGeneratedConfigurationOverwrite {
 
   private OzoneConfiguration conf;
 
-  @Rule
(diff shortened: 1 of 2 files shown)


symious:
@elek @adoroszlai Could you help to review this PR?

elek (commented):
Thanks the patch @symious 

It looks good to me. I took me some time to test it, I think a simple unit test would help a lot to demonstrate the fix

Is there any chance to add a simple unit test (at least for a few types)?

As far as I see `TestConfigurationReflectionUtil` is parametrized (which makes it hard to add anything) but worst case we can add a new test class.

symious:
@elek Thanks for the review, have added a unit test to simulate this case. Please have a check.

symious:
@elek Thanks for the review. 
Sorry for not adding the unit test in the first place, it took some time to find a way to simulate the scenario for the unit test.

elek:
>  Sorry for not adding the unit test in the first place, it took some time to find a way to simulate the scenario for the unit test.

Sure, no problem. **Now** I understand that it was a tricky request, but earlier I thought it's a simpler problem ;-)

PR #2090, opened 2021-03-29
HDDS-5033. SCM may not be able to know full port list of Datanode after Datanode is started.

GlenGeng:

When SCM HA is enabled, after restart DN, the SCM may not know the full ports of that DN.

**The issue is:** 
SCMNodeManager just record the DatanodeDetails once during register. But for DN, it won’t record the admin, server, client port into DatanodeDetails until its ratis server is up. Thus there is contention here: if the register request is reported before ratis server is up, SCM won’t know full port list of that DN.

**The solution is**
If SCM HA is enabled, OzoneContainer#start() will be called multi-times from VersionEndpointTask. The first call should do the initializing job, the successive calls should wait until OzoneContainer is initialized.

https://issues.apache.org/jira/browse/HDDS-5033

## How was this patch tested?

CI and integration test inside tencent.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/ozoneimpl/OzoneContainer.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/ozoneimpl/OzoneContainer.java
@@ -92,11 +92,6 @@ public class OzoneContainer {
   private final ReplicationServer replicationServer;
   private DatanodeDetails datanodeDetails;
 
-  /**
-   * If SCM HA is enabled, OzoneContainer#start() will be called multi-times
-   * from VersionEndpointTask. The first call should do the initializing job,
-   * the successive calls should wait until OzoneContainer is initialized.
-   */
   enum InitializingStatus {
     UNINITIALIZED, INITIALIZING, INITIALIZED
   }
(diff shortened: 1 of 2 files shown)


GlenGeng:
@bshashikant @nandakumar131 Please take a look at this bug fix of SCM HA. Thanks

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/ozoneimpl/OzoneContainer.java
```diff
@@ -257,10 +268,21 @@ private void stopContainerScrub() {
    * @throws IOException
    */
   public void start(String clusterId) throws IOException {
-    if (!isStarted.compareAndSet(false, true)) {
+    if (!initializingStatus.compareAndSet(
```
bshashikant, line 271: Please add some documentation here.

GlenGeng: Done. Please take another look !

bshashikant:
@GlenGeng , can you explain why OzoneContainer#start() needs to be called multiple times if SCM HA is enabled?

GlenGeng:
> @GlenGeng , can you explain why OzoneContainer#start() needs to be called multiple times if SCM HA is enabled for registering to the same SCM?

Please check 
```
public class SCMConnectionManager
    implements Closeable, SCMConnectionManagerMXBean {
      LoggerFactory.getLogger(SCMConnectionManager.class);

  private final ReadWriteLock mapLock;
  private final Map<InetSocketAddress, EndpointStateMachine> scmMachines;
```

If SCM HA is enabled, there will be 3 VersionEndpointTask created, one for each SCM.

GlenGeng:
> @GlenGeng , can you explain why OzoneContainer#start() needs to be called multiple times if SCM HA is enabled for registering to the same SCM?

Actually no, each call is for different SCM.

bshashikant (approved):

PR #2093, opened 2021-03-29
HDDS-4495. [FSO]Delete : Implement async cleanup of garbage and orphan sub-dirs/files

rakeshadr:

Implements a service to get the pending orphan directories and clean up its sub-paths(subdirs and subfiles). Sub-files will be added to existing KeyDeletinService where it will ask scm to clean its blocks, is any.

https://issues.apache.org/jira/browse/HDDS-4495

## How was this patch tested?

Added a few cases, will add more cases to it.

Raised followup jira to verify KDS HDDS-5042

### DIFF
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -255,12 +255,12 @@ public final class OMConfigKeys {
   public static final String OZONE_OM_LAYOUT_VERSION_V1 = "V1";
 
   /**
-   * Configuration properties for Dir Deleting Service.
+   * Configuration properties for Directory Deleting Service.
    */
   public static final String OZONE_DIR_DELETING_SERVICE_INTERVAL =
       "ozone.directory.deleting.service.interval";
   public static final String OZONE_DIR_DELETING_SERVICE_INTERVAL_DEFAULT
-      = "30s";
+      = "60s";
 
   public static final String OZONE_PATH_DELETING_LIMIT_PER_TASK =
       "ozone.path.deleting.limit.per.task";
--- a/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
+++ b/hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
@@ -253,4 +253,17 @@ public final class OMConfigKeys {
   public static final String OZONE_OM_LAYOUT_VERSION_DEFAULT = "V0";
 
   public static final String OZONE_OM_LAYOUT_VERSION_V1 = "V1";
+
+  /**
+   * Configuration properties for Dir Deleting Service.
+   */
+  public static final String OZONE_DIR_DELETING_SERVICE_INTERVAL =
+      "ozone.directory.deleting.service.interval";
+  public static final String OZONE_DIR_DELETING_SERVICE_INTERVAL_DEFAULT
+      = "30s";
+
+  public static final String OZONE_PATH_DELETING_LIMIT_PER_TASK =
+      "ozone.path.deleting.limit.per.task";
+  public static final int OZONE_PATH_DELETING_LIMIT_PER_TASK_DEFAULT = 10000;
+
 }
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
@@ -27,7 +27,6 @@ import org.apache.hadoop.fs.contract.ContractTestUtils;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.utils.db.Table;
 import org.apache.hadoop.ozone.MiniOzoneCluster;
-import org.apache.hadoop.ozone.OzoneConfigKeys;
 import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.TestDataUtil;
 import org.apache.hadoop.ozone.client.OzoneBucket;
@@ -80,7 +79,7 @@ public class TestDirectoryDeletingServiceV1 {
   @BeforeClass
   public static void init() throws Exception {
     OzoneConfiguration conf = new OzoneConfiguration();
-    conf.setInt(OzoneConfigKeys.OZONE_DIR_DELETING_SERVICE_INTERVAL, 5);
+    conf.setInt(OMConfigKeys.OZONE_DIR_DELETING_SERVICE_INTERVAL, 5);
     conf.setBoolean(OMConfigKeys.OZONE_OM_RATIS_ENABLE_KEY, omRatisEnabled);
     conf.setBoolean(OZONE_ACL_ENABLED, true);
     if (isBucketFSOptimized) {
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
@@ -1,0 +1,274 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.fs.ozone;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.contract.ContractTestUtils;
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
+import org.apache.hadoop.hdds.utils.db.Table;
+import org.apache.hadoop.ozone.MiniOzoneCluster;
+import org.apache.hadoop.ozone.OzoneConfigKeys;
+import org.apache.hadoop.ozone.OzoneConsts;
+import org.apache.hadoop.ozone.TestDataUtil;
+import org.apache.hadoop.ozone.client.OzoneBucket;
+import org.apache.hadoop.ozone.om.DirectoryDeletingService;
+import org.apache.hadoop.ozone.om.OMConfigKeys;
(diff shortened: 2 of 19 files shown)


hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
```diff
@@ -253,4 +253,17 @@ private OMConfigKeys() {
...
 
   public static final String OZONE_OM_LAYOUT_VERSION_V1 = "V1";
+
+  /**
+   * Configuration properties for Dir Deleting Service.
+   */
+  public static final String OZONE_DIR_DELETING_SERVICE_INTERVAL =
+      "ozone.directory.deleting.service.interval";
+  public static final String OZONE_DIR_DELETING_SERVICE_INTERVAL_DEFAULT
+      = "30s";
```
linyiqun, line 263: This default value here doesn't match that in ozone-default.xml

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/DirectoryDeletingService.java
```diff
@@ -0,0 +1,285 @@
...
+              }
+            }
+
+            // step-3: Since there is a boundary condition of 'numEntries' in
+            // each batch, check whether the sub paths count reached batch size
+            // limit. If count reached limit then there can be some more child
+            // paths to be visited and will keep the parent deleted directory
+            // for one more pass.
+            List<String> purgeDeletedDirs = new ArrayList<>();
+            if (count > 0) {
```
linyiqun, line 166: Here we at most purge one deleted dir, seems no need to use the List to store this.

linyiqun (commented):
Haven't taken a deep review for this,  but some minor comments below.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/DirectoryDeletingService.java
rakeshadr: Thanks @linyiqun for the reviews. Yes, you are perfectly correct. 

**Design Thoughts:** I've plans to enhance the implementations by adding more directories by recursively traversing to it sub-paths if there are less number of sub-path components in the tree. Please find the below TODO comment in the patch. I'm pasting the same thing here as well for our reference.

```
            if (count > 0) {
              // TODO: Now, there is only one entry in this list. Maintained
              //  list data structure becuase this can be extended to add
              //  more directories within the batchSize limit.
              purgeDeletedDirs.add(pendingDeletedDirInfo.getPath());
            }
```

Assume below is the FS dir structure at OM and batch size is 1000. Again, assume user has deleted dir 'a'. 
With recursive listing in place, `DirDeletingService` can collect dirs [a, b1, b2, b3, c1, c2, d1, d2, d3, e1, e2, e3] in first pass then delete these from `DirectoryTable` and files [d21.txt, d22.txt, e11.txt] will be added to `DeletedTable` so that KDS can take care its block(s) deletions. 
This way I can avoid adding the sub-directory child entries into `DeletedDirectoryTable` and save IO(OM Database) operations. Since this involves some complexity I am planning to implement it via another jira task. Probably, this along with other performance optimizations can be done after feature branch merge to master. Let's see how quickly we can complete the major functionality parts and reach to a logical completion of the feature:-)

By keeping all these thoughts in mind I've used list data structure:-)
Does this make sense to you and shall I keep list data structure?

```
     *                  buck-1
     *                    |
     *                    a
     *                    |
     *      -----------------------------------
     *     |           |                       |
     *     b1          b2                      b3
     *   -----       --------               ----------
     *   |    |      |    |   |             |    |     |
     *  c1   c2     d1   d2  d3             e1   e2   e3
     *                   |                  |
     *               --------               |
     *              |        |              |
     *           d21.txt   d22.txt        e11.txt
```

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
rakeshadr: Good catch, I will update:-)

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/DirectoryDeletingService.java
rakeshadr: FYI, I've created task HDDS-5048 to revisit this part later

linyiqun: Agreed to revisit this in another task.

>With recursive listing in place, DirDeletingService can collect dirs [a, b1, b2, b3, c1, c2, d1, d2, d3, e1, e2, e3]..

Current problem is that we delete dir from parent dir to child, it's not efficient because parent dir relied on subdir to be purged firstly. So then parent dir won't hit the count limitation. But this can lead the repeated traversing on parent dir.  A better way I am thinking we could use recursive way to collect all dirs and then stored in a stack structure. 
For above example, the dirs will be stored as[a, b1, c1, c2, b2, d1, d2, d3, b3, e1, e2, e3]. Then we pop the dir from the stack, the delete order is from e3, e2, ..., a. 

Based on above way, we could make sure there is no any subdir needed to be purged when purging a dir.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  @Test
+  public void testDeleteWithMultiLevels() throws Exception {
+    Path root = new Path("/rootDir");
+    Path appRoot = new Path(root, "appRoot");
+
+    for (int i = 1; i <= 3; i++) {
+      Path parent = new Path(appRoot, "parentDir" + i);
+      Path child = new Path(parent, "childFile");
```
linyiqun, line 194: Can we reset the ozone.path.deleting.limit.per.task to a small value and create dir/child files more than that value? So that we could test current purge logic works fine with the delete limitation per task.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
```diff
@@ -0,0 +1,273 @@
...
+    DirectoryDeletingService dirDeletingService =
+        (DirectoryDeletingService) cluster.getOzoneManager().getKeyManager()
+            .getDirDeletingService();
+
+    GenericTestUtils.waitFor(
+        () -> dirDeletingService.getDeletedDirsCount() >= 5,
+        1000, 10000);
+    GenericTestUtils.waitFor(
+        () -> dirDeletingService.getMovedFilesCount() >= 3,
+        1000, 10000);
```
linyiqun, line 245: Can we record the getDeletedDirsCount/getMovedFilesCount before the delete, and then get these value again and compare the delta value. This way would be the more accurate check.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
```diff
@@ -0,0 +1,273 @@
...
+import java.io.IOException;
+
+import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_ACL_ENABLED;
+import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_FS_ITERATE_BATCH_SIZE;
+import static org.junit.Assert.fail;
+
+/**
+ * Directory deletion service test cases.
+ */
+public class TestDirectoryDeletingServiceV1 {
```
linyiqun, line 59: Why class name is TestDirectoryDeletingServiceV1 not the TestDirectoryDeletingService? I don't see the name DirectoryDeletingServiceV1 is defined in this PR.

rakeshadr: @linyiqun  OK, I thought of following a pattern to search classes easily by any developer.

How about rename class like `TestDirectoryDeletingServiceWithFSOBucket`. Shortly, I am planning to put up a patch by renaming V1 items to FSOBucket or FSOB or welcome any better naming:-)

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
rakeshadr: Sure, will do it.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/DirectoryDeletingService.java
rakeshadr: @linyiqun Thanks for adding more details.

>A better way I am thinking we could use recursive way to collect all dirs and then stored in a stack structure.
Say, if user has deleted a very large parent dir with millions of sub-children with more depth(say 10 levels) and each level has more than batchSized dirs and files. I think, maintaining stack will have complexity by increasing heap memory consumption.

> Current problem is that we delete dir from parent dir to child, it's not efficient because parent dir relied on subdir to be purged firstly
I am deleting the parentdir once all its immediate sub-dirs are added to `DeletedDirectortTable` and sub-files to `DeletedTable`. This ensures that parentdir can be safely removed and the sub-dirs will become/promotted as parentdirs for the next DDService iteration.

Appreciate your feedback. Thanks again!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
linyiqun: TestDirectoryDeletingServiceWithFSOBucket looks good to me.

hadoop-ozone/common/src/main/java/org/apache/hadoop/ozone/om/OMConfigKeys.java
rakeshadr: Done!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
rakeshadr: Added new tests covering multiple batch deletion including dirs/files.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
rakeshadr: Done!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestDirectoryDeletingServiceV1.java
rakeshadr: Done!

rakeshadr:
Thanks again @linyiqun for the review comments. I've updated PR by addressing the comments.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/DirectoryDeletingService.java
rakeshadr: @linyiqun Does my above comment make sense to you. Thanks!

linyiqun: Comment makes the sense to me, @rakeshadr .

linyiqun (approved):
LGTM, +1.
Thanks for addressing the review comments!

rakeshadr:
Thanks a lot @linyiqun, I will merge it.

PR #2094, opened 2021-03-29
HDDS-5022. SCM get roles command should provide Ratis Leader/Follower…

ghuangups:
… information.

"ozone admin scm roles" command currently returns scm role info like this: [scm1:9865, scm2:9865, scm3:9865]. This change is to add: LEADER and FOLLOWER info in the return. The new result from the command is now look like this: [scm3:9865:FOLLOWER, scm2:9865:FOLLOWER, scm1:9865:LEADER] 

https://issues.apache.org/jira/browse/HDDS-5022

## How was this patch tested?

Manually tested locally by creating a SCM HA cluster with 3 SCM nodes and executed command: ozone admin scm roles from one of the SCM node. With this change, the command result was: [scm3:9865:FOLLOWER, scm2:9865:FOLLOWER, scm1:9865:LEADER].

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
@@ -18,24 +18,29 @@
 package org.apache.hadoop.hdds.scm.ha;
 
 import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 import java.util.Iterator;
+import java.util.Collection;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import org.apache.hadoop.hdds.HddsUtils;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.conf.OzoneConfiguration;
 import org.apache.hadoop.hdds.protocol.proto.SCMRatisProtocol.RequestType;
 import org.apache.hadoop.hdds.scm.AddSCMRequest;
 import org.apache.hadoop.hdds.scm.server.StorageContainerManager;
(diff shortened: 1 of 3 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
```diff
@@ -225,7 +225,7 @@ public void stop() throws IOException {
   @Override
   public List<String> getRatisRoles() {
     return division.getGroup().getPeers().stream()
-        .map(peer -> peer.getAddress() == null ? "" : peer.getAddress())
+        .map(peer -> peer.getAddress() == null ? "" : peer.getAddress().concat(peer.getAdminAddress() == null ? ":FOLLOWER" : ":LEADER"))
```
avijayanhwx, line 228: I am not sure if we should rely on the getAdminAddress() call to be null to determine if this node is LEADER or FOLLOWER. Can we do something similar to OM? The SCM that is returning the answer will be leader (since read calls go only to Leader), and the other nodes will be followers.

avijayanhwx: Also, can we use org.apache.ratis.proto.RaftProtos.RaftPeerRole#LEADER/FOLLOWER instead of hardcoded strings?

avijayanhwx (commented):
Thanks for working on this @ghuangups. Left a comment inline.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServerImpl.java
GlenGeng, line 228:
```
  private static RaftPeer.Builder raftPeerBuilderFor(DatanodeDetails dn) {
    return RaftPeer.newBuilder()
        .setId(toRaftPeerId(dn))
        .setAddress(toRaftPeerAddress(dn, Port.Name.RATIS_SERVER))
        .setAdminAddress(toRaftPeerAddress(dn, Port.Name.RATIS_ADMIN))
        .setClientAddress(toRaftPeerAddress(dn, Port.Name.RATIS));
  }
```

I consider that each raft peer has its own admin address, no matter if it is leader or follower.

Better use ```NetUtils.isLocalAddress()``` to determine the peer whose address equals local, and assign him the leader role.

PR #2096, opened 2021-03-30
HDDS-5047. Refactor Pipeline to use ReplicationConfig instead of factor/type

elek:

HDDS-5011 introduces Java ReplicationConfig classes which can be used as a replacement of replicationType and replicationFactor.

First task is replacing type/factor with ReplicationConfig in Pipeline and related managers (PipelineManager BackgroundPipelineCreatorV2, PipelineStateManager...)

We can do it on the master without the EC related stuff... (later we will add the small part which is required for EC...)

https://issues.apache.org/jira/browse/HDDS-5047

## How was this patch tested?

Main logic is not changed, just the signature of the interfaces (and implementations) all the related unit tests and integration tests are updated.


elek:
This pr depends on #2089

bshashikant (commented):
The approach looks good to me. I have one question regarding compatibility:
1) How does a new client can still work with old server?

elek:
> How does a new client can still work with old server?

Good question. The trick here is that we don't do any protobuf modifications. We serialize the protobuf content to a new Java object structure which makes it possible to handle additional parameters (for example: EC can handle more properties instead of just the one `replicationFactor`). 

image

This patch modifies the SCM to use `replicationConfig` everywhere, but the logic is the same as before. With this approach on the EC branch we can add a new replicationConfig implementation and all the pipeline managers will serve it.

elek:
> How does a new client can still work with old server?

Sorry, I didn't answer this exact question: there is no new client here. But EC client will add additional field and can work with old server as soon as EC replication type is not used.

So everything will work as before.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineManagerImpl.java
```diff
@@ -131,12 +133,12 @@ public void testCreatePipeline() throws Exception {
...
     Assert.assertTrue(pipelineManager.getPipelines().isEmpty());
     Pipeline pipeline1 = pipelineManager.createPipeline(
-        HddsProtos.ReplicationType.RATIS, HddsProtos.ReplicationFactor.THREE);
+        new RatisReplicationConfig(ReplicationFactor.THREE));
     Assert.assertEquals(1, pipelineManager.getPipelines().size());
     Assert.assertTrue(pipelineManager.containsPipeline(pipeline1.getId()));
 
     Pipeline pipeline2 = pipelineManager.createPipeline(
-        HddsProtos.ReplicationType.RATIS, HddsProtos.ReplicationFactor.ONE);
+        new RatisReplicationConfig(ReplicationFactor.THREE));
```
sodonnel, line 141: I think there is a mistake here - the old code had factor ONE. New code has THREE.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManager.java
```diff
@@ -237,39 +239,33 @@ void loadContainer(final ContainerInfo containerInfo) throws SCMException {
...
 
-    boolean bgCreateOne = (type == ReplicationType.RATIS) && replicationFactor
-        == ReplicationFactor.ONE && autoCreateRatisOne;
-    boolean bgCreateThree = (type == ReplicationType.RATIS) && replicationFactor
-        == ReplicationFactor.THREE;
+    boolean bgCreateOne = RatisReplicationConfig
+        .hasFactor(replicationConfig, ReplicationFactor.ONE)
+        && autoCreateRatisOne;
+    boolean bgCreateThree = RatisReplicationConfig
+        .hasFactor(replicationConfig, ReplicationFactor.ONE);
```
sodonnel, line 254: Is this last line of the change a mistake? The original code had THREE but this last line has ONE.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/BlockManagerImpl.java
```diff
@@ -149,20 +148,19 @@ public void stop() throws IOException {
   /**
    * Allocates a block in a container and returns that info.
    *
-   * @param size - Block Size
-   * @param type Replication Type
-   * @param factor - Replication Factor
```
sodonnel, line 154: NIT: Add in the param for ReplicationConfig (also owner is missing too, but not due to this change, so may as well add it in).

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocolPB/StorageContainerLocationProtocolClientSideTranslatorPB.java
```diff
@@ -158,15 +158,6 @@ private ScmContainerLocationResponse submitRpcRequest(
...
-  /**
-   * Asks SCM where a container should be allocated. SCM responds with the set
-   * of datanodes that should be used creating this container. Ozone/SCM only
-   * supports replication factor of either 1 or 3.
-   *
-   * @param type   - Replication Type
-   * @param factor - Replication Count
-   */
-  @Override
   public ContainerWithPipeline allocateContainer(
```
sodonnel, line 161: Just double checking - is this change intentional? I have not looked into the details of this class etc.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocolPB/ScmBlockLocationProtocolClientSideTranslatorPB.java
```diff
@@ -141,27 +144,42 @@ private SCMBlockLocationResponse handleError(SCMBlockLocationResponse resp)
    * set of datanodes that should be used creating this block.
    * @param size - size of the block.
    * @param num - number of blocks.
-   * @param type - replication type of the blocks.
-   * @param factor - replication factor of the blocks.
```
sodonnel, line 145: replicationType (and owner) are missing from the param spec.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocol/ScmBlockLocationProtocol.java
```diff
@@ -59,8 +60,29 @@
...
+  }
+
+  /**
+   * Asks SCM where a block should be allocated. SCM responds with the
+   * set of datanodes that should be used creating this block.
+   *
+   * @param size              - size of the block.
+   * @param numBlocks         - number of blocks.
+   * @param replicationConfig - replicationConfiguration
+   * @param excludeList       List of datanodes/containers to exclude during
```
sodonnel, line 78: Nit: Add owner to param spec - note it was missing before these changes, so not caused by this change.

hadoop-hdds/dev-support/checkstyle/checkstyle.xml
```diff
@@ -71,9 +71,6 @@
     <!--<module name="FileLength">-->
     <module name="FileTabCharacter"/>
 
-    <module name="LineLength">
-        <property name="fileExtensions" value="java"/>
-    </module>
```
sodonnel, line 76: Is this change intentional?

sodonnel:
@elek Thanks for this large and repetitive change. I think it is needed to enable EC to progress more easily.

This change looks mostly good to me. I have just a few comments inline for you to double check. If we address those I am +1.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelineManagerImpl.java
elek: Good point, thanks. Fixing it. (BTW, it seems to be the limitation of the unit test. Pipeline parameters are not checked, so it can be ONE/THREE, but let's keep it as it was)

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerStateManager.java
elek: yep. Thanks, fixing.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/block/BlockManagerImpl.java
elek: :+1:

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocolPB/StorageContainerLocationProtocolClientSideTranslatorPB.java
elek: Thanks. I think it was removed accidentally when I added the new method. Restored it (and also fixed the owner field here...)

hadoop-hdds/dev-support/checkstyle/checkstyle.xml
elek: Nope. My intellij removed it. Let me restore.

elek:
Thank you very much the review/comments @sodonnel. Especially as I know how ungrateful is reviewing such a big task.

 Fixes are pushed by a576fab7d861131cd4c6950c58475ad7244ad36d

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/BackgroundPipelineCreator.java
```diff
@@ -91,16 +96,20 @@ void triggerPipelineCreation() {
...
+    if (replicationConfig.getReplicationType()
+        == HddsProtos.ReplicationType.RATIS) {
+      return RatisReplicationConfig
+          .hasFactor(replicationConfig, ReplicationFactor.ONE) && (!autoCreate);
+    } else if (replicationConfig.getReplicationType()
+        == ReplicationType.STAND_ALONE) {
       // For STAND_ALONE Replication Type, Replication Factor 3 should not be
       // used.
-      return factor == HddsProtos.ReplicationFactor.THREE;
+      return ((StandaloneReplicationConfig) replicationConfig)
```
umamaheswararao, line 109: factor != ONE make sense here?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocol/ScmBlockLocationProtocol.java
```diff
@@ -59,8 +60,30 @@
...
+   * @param numBlocks         - number of blocks.
+   * @param replicationConfig - replicationConfiguration
+   * @param owner             - service owner of the new block
+   * @param excludeList       List of datanodes/containers to exclude during
+   *                          block
+   *                          allocation.
+   * @return allocated block accessing info (key, pipeline).
+   * @throws IOException
+   */
+  List<AllocatedBlock> allocateBlock(long size, int numBlocks,
```
umamaheswararao, line 85: Why owner fix included in this patch?

sodonnel (approved):
Latest version LGTM

umamaheswararao (approved):
Looks good to me.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/protocol/ScmBlockLocationProtocol.java
elek: Only javadoc fixes are included. Here it is showed only because I modified ReplicationConfig at the same line:

Old code
```
       ReplicationType type, ReplicationFactor factor, String owner,
``` 

New code:

```
       ReplicationType type, ReplicationFactor factor, String owner,
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/BackgroundPipelineCreator.java
elek: I tried to avoid modification in the algorithm, but I am fine with that. Just modified in d8d038af20c36cea571f7d6009dd9fbe1d78f579

elek:
Merging it now. Thanks the review @sodonnel and @umamaheswararao

PR #2098, opened 2021-03-31
HDDS-5022. SCM get roles command should provide Ratis Leader/Follower…

ghuangups:
… information.

"ozone admin scm roles" command currently returns scm role info like this: [scm1:9865, scm2:9865, scm3:9865]. This change is to add: LEADER and FOLLOWER info in the return. The new result from the command is now look like this: [scm3:9865:FOLLOWER, scm2:9865:FOLLOWER, scm1:9865:LEADER]

https://issues.apache.org/jira/browse/HDDS-5022

## How was this patch tested?

Manually tested locally by creating a SCM HA cluster with 3 SCM nodes and executed command: ozone admin scm roles from one of the SCM node. With this change, the command result was: [scm3:9865:FOLLOWER, scm2:9865:FOLLOWER, scm1:9865:LEADER].

Also manually executed admincli tests under ozonesecure.


GlenGeng:
LGTM. Thanks @ghuangups for the contribution. Wait for CI.

ghuangups:
Failed acceptance test seems unrelated?

<kw name="Execute AWSS3APICli" library="commonawslib">
<arg>get-object --bucket ${BUCKET} --key ${PREFIX}/copyrange/destination /tmp/part-result</arg>
<kw name="Execute" library="os">
<var>${output}</var>
<arg>aws s3api --endpoint-url ${ENDPOINT_URL} ${command}</arg>
<kw name="Run Keyword And Return" library="BuiltIn">
<arg>Execute and checkrc</arg>
<arg>${command}</arg>
<arg>0</arg>
<doc>Runs the specified keyword and returns from the enclosing user keyword.</doc>
<kw name="Execute and checkrc" library="os">
<arg>${command}</arg>
<arg>0</arg>
<kw name="Run And Return Rc And Output" library="OperatingSystem">
...
</kw>
<kw name="Log" library="BuiltIn">
...
</kw>
**<kw name="Should Be Equal As Integers" library="BuiltIn">
<arg>${rc}</arg>
<arg>${expected_error_code}</arg>
<doc>Fails if objects are unequal after converting them to integers.</doc>
<msg timestamp="20210331 13:39:44.665" level="INFO">Argument types are: <class 'int'> <type 'unicode'></msg>
<msg timestamp="20210331 13:39:44.665" level="FAIL">255 != 0</msg>
<status status="FAIL" endtime="20210331 13:39:44.665" starttime="20210331 13:39:44.664"/>
</kw>**
<status status="FAIL" endtime="20210331 13:39:44.665" starttime="20210331 13:39:41.772"/>
</kw>
<msg timestamp="20210331 13:39:44.665" level="INFO">Returning from the enclosing user keyword.</msg>
<status status="FAIL" endtime="20210331 13:39:44.666" starttime="20210331 13:39:41.770"/>
</kw>
<status status="FAIL" endtime="20210331 13:39:44.666" starttime="20210331 13:39:41.770"/>
</kw>
<status status="FAIL" endtime="20210331 13:39:44.666" starttime="20210331 13:39:41.769"/>
</kw>
<timeout value="5 minutes"/>
<status status="FAIL" endtime="20210331 13:39:44.666" starttime="20210331 13:39:32.016">255 != 0</status>
</test>

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServer.java
```diff
@@ -45,7 +46,7 @@ SCMRatisResponse submitRequest(SCMRatisRequest request)
   /**
    * Returns roles of ratis peers.
    */
-  List<String> getRatisRoles();
+  List<String> getRatisRoles() throws UnknownHostException;
```
avijayanhwx, line 49: nit. Good to through "IOException" here, rather than an implementation detail like UnknownHostException.

hadoop-ozone/dist/src/main/smoketest/admincli/scmha.robot
```diff
@@ -25,4 +25,4 @@ Test Timeout        5 minutes
 *** Test Cases ***
 Run scm roles
     ${output} =         Execute          ozone admin scm roles
-                        Should contain   ${output}  [scm:9865]
+                        Should Match Regexp   ${output}  [scm:9865:(LEADER|FOLLOWER)]
```
avijayanhwx, line 28: Is it possible to verify that we have 1 LEADER & 2 FOLLOWER strings in the response?

ghuangups: Thanks @avijayanhwx for the review and great comment. As discussed, this test was actually executed from ozonesecure with a 1 node SCM cluster. Therefore we will not have 1 LEADER and 2 FOLLOWER.

Also, had some discussion with @GlenGeng, there is a need to differentiate a single node SCM cluster from an SCM HA cluster with only 1 existing SCM node. In the later case, the command will print out something like this: [scm1:9865:LEADER] and therefore we'll know that this is a HA setup. Therefore the implementation of "getRatisRoles" remains unchanged from the previous version for now. Thanks again!

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMRatisServer.java
ghuangups: Thanks for the great input, code was updated.

GlenGeng:
+1. Thanks @ghuangups for the contribution. I will merge it.

ghuangups:
Thank you so much, @GlenGeng @avijayanhwx

PR #2099, opened 2021-03-31
HDDS-5049. Add timeout support for ratis requests in SCM HA.

bshashikant:
Added Timeout support for ratis requests.

https://issues.apache.org/jira/browse/HDDS-5049

## How was this patch tested?

Existing UT

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
@@ -467,6 +467,11 @@ public final class ScmConfigKeys {
   public static final TimeDuration OZONE_SCM_RATIS_MINIMUM_TIMEOUT_DEFAULT
       = TimeDuration.valueOf(1, TimeUnit.SECONDS);
 
+  public static final String OZONE_SCM_RATIS_REQUEST_TIMEOUT_KEY
+      = "ozone.scm.ratis.minimum.timeout";
+  public static final String OZONE_SCM_RATIS_REQUEST_TIMEOUT_DEFAULT
+      = "30s";
+
   // SCM Ratis Leader Election configurations
   public static final String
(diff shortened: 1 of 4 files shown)


bshashikant:
@GlenGeng , can you have a look?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
```diff
@@ -467,6 +467,11 @@
   public static final TimeDuration OZONE_SCM_RATIS_MINIMUM_TIMEOUT_DEFAULT
       = TimeDuration.valueOf(1, TimeUnit.SECONDS);
 
+  public static final String OZONE_SCM_RATIS_REQUEST_TIMEOUT_KEY
+      = "ozone.scm.ratis.minimum.timeout";
```
GlenGeng, line 471: Should we add the key to `SCMHAConfiguration` ? 
BTW, the key `OZONE_SCM_RATIS_SERVER_REQUEST_TIMEOUT_KEY`, `OZONE_SCM_RATIS_SERVER_RETRY_CACHE_TIMEOUT_KEY`, `OZONE_SCM_RATIS_MINIMUM_TIMEOUT_KEY`, `OZONE_SCM_LEADER_ELECTION_MINIMUM_TIMEOUT_DURATION_KEY` etc are not used any more. Better clean up them.

bshashikant: I think, we should be remove SCMHAConfiguration and pull all the required configs in ScmConfigKeys and remove the unsed ones. But, i would prefer to do this as a part of different jira altogether.

GlenGeng: Sure. Thanks for creating the jira.

GlenGeng:
+1. Thanks @bshashikant for the work. I will merge it.

PR #2100, opened 2021-03-31
HDDS-5052. [SCM HA Security] Handle leader changes between SCMInfo and getSCMSigned Cert in OM/SCM.

bharatviswa504:

This Jira is to handle leader change between getScmInfo and getScmSignedCert.

**Problem:**
Leader is SCM1 - Returned SCMID is SCM1ID
ScmInfo returns the leader SCMID which is SCM1ID.

Leader is SCM2 - SCM ID is SCM2ID
getSCMSignedCert, during generate certificate it has a check compare the scmId passed in CSR, is same as current SCM scmID

In this case when the leader change between these 2 calls OM will fail to get a Certificate.

**Proposed Solution:**
Discuss with Xiaoyu offline, and we decided to eliminate persist scmId and check SCMID during OM startup.

We persist SCM ClusteriD for both HA/non-HA in OzoneManager.

And in SCM we remove the SCMID check and keep the SCM ClusterID check. (ClusterID is unique for a single Ozone cluster)

In SCM previously sub-CA subject looks like **CN=scm-sub@host,OU=fetchedSCMID from primary,O=ClusterID**
After this change sub-CA subject will be **CN=scm-sub@host,OU=local SCMID,O=ClusterID**

For OM/DN the subject names are the same as before.

And as SCMID is used in Subject CN name, not touched that, so that the certificate subject will be in the same format before/after the upgrade.

https://issues.apache.org/jira/browse/HDDS-5052

## How was this patch tested?

Existing docker tests.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultApprover.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/authority/DefaultApprover.java
@@ -111,7 +111,7 @@ public class DefaultApprover extends BaseApprover {
     String csrClusterId = x500Name.getRDNs(BCStyle.O)[0].getFirst().getValue().
         toASN1Primitive().toString();
 
-    if (!scmId.equals(csrScmId) || !clusterId.equals(csrClusterId)) {
+    if (!clusterId.equals(csrClusterId)) {
       if (csrScmId.equalsIgnoreCase("null") &&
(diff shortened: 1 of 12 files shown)


bharatviswa504:
The last commit is the changes proposed in this PR. It is dependent on HDDS-4915.

bharatviswa504:
Rebased now. It is ready for review.

bshashikant (approved):
Looks good to me.

bharatviswa504:

PR #2101, opened 2021-03-31
HDDS-4828. SCM should go into "safe mode" until there is at least 1 pipeline to work with after finalization.

errose28:

After finalization of SCM, it will take at least a heartbeat for datanodes to finalize and update their layout versions. Only after this SCM will make pipelines. This feature will make SCM wait for at least one pipeline to be created before returning from finalization, so that writes can go through right after SCM is finalized.

This change also fixes the existing `TestHDDSUpgrade` integration test after SCM HA to work again and test this change.

HDDS-4828

## How was this patch tested?

Existing integration test modified to check the new behavior.


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/pipeline/BackgroundPipelineCreator.java
```diff
@@ -32,6 +32,8 @@
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
```
GlenGeng, line 36: Just a kind reminder, since SCM HA is merged back to master, `BackgroundPipelineCreator.java` is deprecated and replaced by `BackgroundPipelineCreatorV2.java`.

errose28: Thanks @GlenGeng. We are working on bring the SCM HA changes into the upgrade branch, then this PR will be updated. I will move it to a draft for now.

adoroszlai:
@errose28 findbugs fix (6822b1074ee72e008d2aa6c927bb82069477370b) needs to be merged into feature branch.

avijayanhwx (approved):
LGTM +1

PR #1722, opened 2020-12-17
HDDS-4600. SCM HA Config Phase 1: Use SCMNodeDetails as the entry to initialize SCM except for Ratis servers

amaliujia:
## What changes were proposed in this pull request?

Start to use SCMNodeDetails as the entry to retrieve address configs for servers (except Ratis server). Currently will load default config for servers.

Design doc: https://docs.google.com/document/d/1Q_ofDnvYYUGMQfZqdiUs5BMGQAy0YDuaM5AYuG8Ub6o/edit?usp=sharing

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-4600

## How was this patch tested?

UT

PR #2106, opened 2021-04-01
HDDS-5057. Remove backward direction dependency between HDDS->Ozone

elek:
JIRA: https://issues.apache.org/jira/browse/HDDS-5057

HDDS is the block-management layer and on the top of it we created Ozone components which uses this lower-level block-management layer.

This separation is not actively used (today we don't have any other upper layer which uses HDDS, only Ozone, but it may be possible).

This dependency is also reflected by the maven project hierarchy. HDDS projects don't depend on Ozone project, only Ozone projects depends on HDDS.

If something is required for both Ozone and HDDS components it should be added to the hdds-framework (or common if it's also required by clients).

This dependency hierarchy -- which makes the project easier to maintain -- is broken since #1722.

In this PR I suggest moving the required utilities back to HDDS and remove the backward direction hdds->ozone dependency.  

## How was this patch tested?

CI tests.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/ExitManager.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/ExitManager.java
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.hadoop.ozone.util;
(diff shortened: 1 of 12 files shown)


adoroszlai (approved):
Thanks @elek for the cleanup.

elek:
Merging it. Thanks the review @adoroszlai

PR #2107, opened 2021-04-01
HDDS-5058. Make getScmInfo retry for a duration.

bharatviswa504:

This Jira proposes to add a ceration duration to try getScmInfo, instead of retry forever with fixed sleep.

In a few docker tests CI run, we have seen this issue, after 15 retries Om init failed, as SCM is started later.

https://issues.apache.org/jira/browse/HDDS-5058

## How was this patch tested?

Manually tested with docker compose env.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
@@ -22,6 +22,7 @@ import org.apache.hadoop.hdds.annotation.InterfaceStability;
 import org.apache.ratis.proto.RaftProtos.ReplicationLevel;
 import org.apache.ratis.util.TimeDuration;
 
+import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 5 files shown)


bshashikant (approved):

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ScmConfigKeys.java
```diff
@@ -510,6 +510,11 @@
       "hdds.scm.ha.security.enable";
   public static final boolean OZONE_SCM_HA_SECURITY_SUPPORTED_DEFAULT = false;
 
+  public static final String OZONE_SCM_INFO_WAIT_DURATION = "ozone.scm.info" +
```
GlenGeng, line 513:
```
public static final String OZONE_SCM_INFO_WAIT_DURATION = 
    "ozone.scm.info.wait.duration";
```
So that one can easily search the key.

bharatviswa504: I will take care of this during commit.

bharatviswa504:
I will commit this shortly, as this will help in fixing occasional failures in docker tests.

One such run:
https://github.com/apache/ozone/runs/2243496103

bharatviswa504:

PR #2104, opened 2021-04-01
HDDS-4936. Change ozone groupId from org.apache.hadoop to org.apache.ozone

ChenSammi:
https://issues.apache.org/jira/browse/HDDS-4936

(this pr continues the work / discussion started in #2018)

PR #2111, opened 2021-04-06
HDDS-5063. Remove hadoop- prefixes from release artifacts (release br…

elek:

When we create a release artifact today it has the `hadoop-` prefix in the names. Would be better to remove to avoid confusion.

Note: full `hadoop-` prefix is removed by #2104 (thanks to @ChenSammi ), this patch is a very small subset of that patch just to fix the tar file name on the release branch.

https://issues.apache.org/jira/browse/HDDS-5063

## How was this patch tested?

```
mvn clean install -Dmaven.javadoc.skip=true -Dskip.npx -DskipTests -Psign,dist,src -Dtar -Dgpg.keyname=$CODESIGNINGKEY
cd hadoop-ozone/dist/target
tar tzf ozone-1.1.0-SNAPSHOT.tar.gz
#check the root dir prefix in the output
tar tzf ozone-1.1.0-SNAPSHOT-src.tar.gz
#check the root dir prefix in the output
```

### DIFF
--- a/hadoop-ozone/dist/dev-support/bin/dist-tar-stitching
+++ b/hadoop-ozone/dist/dev-support/bin/dist-tar-stitching
@@ -38,8 +38,8 @@ function run()
   fi
 }
 
-run tar -c -f "hadoop-ozone-${VERSION}.tar" "ozone-${VERSION}"
-run gzip -f "hadoop-ozone-${VERSION}.tar"
+run tar -c -f "ozone-${VERSION}.tar" "ozone-${VERSION}"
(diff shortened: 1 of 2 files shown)


prashantpogde (approved):
+1

prashantpogde:
Thank you @elek. Build failure is unrelated. I will merge this in 1.1.0 and try the CI in another PR.

adoroszlai:
@prashantpogde Please do not merge with failing checks.  Even if failures are unrelated, they may hide other issues: in this case HDDS-5069.

PR #2113, opened 2021-04-06
HDDS-5027. [SCM HA Security] Handle leader changes during bootstrap.

bharatviswa504:

This Jira is to handle when primary SCM is not leader SCM and bootstrap is failing due to this.

https://issues.apache.org/jira/browse/HDDS-5027

## How was this patch tested?
Existing docker ozonesecure-ha, as modified code to use RatisClient for propagation of SCM certs.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAInvocationHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAInvocationHandler.java
@@ -22,6 +22,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
 import com.google.common.base.Preconditions;
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
 import org.apache.hadoop.hdds.protocol.proto.SCMRatisProtocol.RequestType;
 import org.apache.hadoop.hdds.scm.metadata.Replicate;
 import org.apache.hadoop.util.Time;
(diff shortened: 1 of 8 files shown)


bshashikant:
Looks good. +1 . Please address the checkstyle before committing.

bharatviswa504:
Thank You @bshashikant for the review, Addressed checkstyle issues. (Looks like findbug issue is fixed by #2120

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAInvocationHandler.java
```diff
@@ -89,11 +90,29 @@ private Object invokeRatis(Method method, Object[] args)
...
+        method.getName(), method.getParameterTypes(), args);
+
+    // Scm Cert DB updates should use RaftClient.
+    // As rootCA which is primary SCM only can issues certificates to sub-CA.
+    // In case primary is not leader SCM, still sub-ca cert DB updates should go
+    // via ratis. So, in this special scenario we use RaftClient.
+    final SCMRatisResponse response;
+    if (method.getName().equals("storeValidCertificate") &&
+        args[args.length -1].equals(HddsProtos.NodeType.SCM)) {
+      response =
```
GlenGeng, line 103: How about remove `replicate` annotation for `storeValidCertificate`, and directly call HASecurityUtils.submitScmCertsToRatis() there ? It should be simpler and easier to maintain.

bharatviswa504: this is needed only for scm signed certs, for others like om/dn we still need to go via normal route.
This is the reason for this approach.

GlenGeng: Make sense. Thanks for the clarification.

bharatviswa504:
Findbug issue fix is taken care by #2120

PR #2114, opened 2021-04-06
HDDS-5062. Add a config to bypass clusterId validation for bootstrapping SCM.

bshashikant:
IN SCM HA, the primary node starts up the ratis server while other bootstrapping nodes will get added to the ratis group. Now, if all the bootstrapping SCM's get stopped, the primary node will now step down from leadership as it will loose majority. If the bootstrapping nodes are now bootstrapped again,  the bootsrapping node will try to first validate the cluster id from the leader SCM with the persisted cluster id , but as there is no leader existing, bootstrapping wil keep on failing and retrying until it shuts down. 

The issue can be very easily simulated in kubernetes deployments, where bootstrap and init cmds are run repeatedly on every restart.

The Jira aims to bypass the cluster id validation if a bootstrapping node already has a cluster id.
https://issues.apache.org/jira/browse/HDDS-5062

## How was this patch tested?
Added unit test

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
@@ -42,7 +42,6 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
-import java.io.File;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.List;
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
@@ -233,19 +233,16 @@ public class TestStorageContainerManagerHA {
 
   @Test
   public void testBootStrapSCM() throws Exception {
-    StorageContainerManager scm1 = cluster.getStorageContainerManagers().get(0);
     StorageContainerManager scm2 = cluster.getStorageContainerManagers().get(1);
     OzoneConfiguration conf2 = scm2.getConfiguration();
-    conf2.set(ScmConfigKeys.OZONE_SCM_PRIMORDIAL_NODE_ID_KEY,
-        scm1.getSCMNodeId());
-    conf2.set(ScmConfigKeys.OZONE_SCM_PRIMORDIAL_NODE_ID_KEY,
-        scm1.getSCMNodeId());
     boolean isDeleted = scm2.getScmStorageConfig().getVersionFile().delete();
     Assert.assertTrue(isDeleted);
     final SCMStorageConfig scmStorageConfig = new SCMStorageConfig(conf2);
     scmStorageConfig.setClusterId(UUID.randomUUID().toString());
     scmStorageConfig.getCurrentDir().delete();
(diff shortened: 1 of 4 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
```diff
@@ -228,4 +230,25 @@ public void testPrimordialSCM() throws Exception {
     Assert.assertTrue(
         StorageContainerManager.scmInit(conf2, scm2.getClusterId()));
   }
+
+  @Test
+  public void testBootStrapSCM() throws Exception {
+    StorageContainerManager scm1 = cluster.getStorageContainerManagers().get(0);
+    StorageContainerManager scm2 = cluster.getStorageContainerManagers().get(1);
+    OzoneConfiguration conf2 = scm2.getConfiguration();
+    conf2.set(ScmConfigKeys.OZONE_SCM_PRIMORDIAL_NODE_ID_KEY,
```
bharatviswa504, line 239: Same line duplicated

bharatviswa504: And also do we need this setting here for this test case?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
```diff
@@ -228,4 +230,25 @@ public void testPrimordialSCM() throws Exception {
...
+        scm1.getSCMNodeId());
+    conf2.set(ScmConfigKeys.OZONE_SCM_PRIMORDIAL_NODE_ID_KEY,
+        scm1.getSCMNodeId());
+    boolean isDeleted = scm2.getScmStorageConfig().getVersionFile().delete();
+    Assert.assertTrue(isDeleted);
+    final SCMStorageConfig scmStorageConfig = new SCMStorageConfig(conf2);
+    scmStorageConfig.setClusterId(UUID.randomUUID().toString());
+    scmStorageConfig.getCurrentDir().delete();
+    scmStorageConfig.initialize();
```
bharatviswa504, line 248: So test here is delete scm2 version file.
L246-249 created a new version file for SCM2 in same location with new clusterID.
Now we are check with/without OZONE_SCM_SKIP_BOOTSTRAP_VALIDATION_KEY?

bharatviswa504 (commented):
Overall LGTM.
Minor comments/questions.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
bshashikant: Yes

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/scm/TestStorageContainerManagerHA.java
bshashikant: Addressed in latest patch.

GlenGeng:
LGTM. Wait for CI pass.

bharatviswa504 (commented):
+1 LGTM

bshashikant:
The findbug issue reported will be fixed by #2120 . Merging it.

PR #2116, opened 2021-04-06
HDDS-5050. Add retry policy for ratis requests in SCM HA.

bshashikant:

Added failover and retry logic based on exception received from server.

https://issues.apache.org/jira/browse/HDDS-5050

## How was this patch tested?
Existing UT.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMBlockLocationFailoverProxyProvider.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMBlockLocationFailoverProxyProvider.java
@@ -22,6 +22,7 @@ import com.google.common.annotations.VisibleForTesting;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hdds.conf.ConfigurationException;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
+import org.apache.hadoop.hdds.scm.ha.SCMHAUtils;
 import org.apache.hadoop.hdds.scm.ha.SCMNodeInfo;
 import org.apache.hadoop.hdds.scm.protocolPB.ScmBlockLocationProtocolPB;
 import org.apache.hadoop.hdds.utils.LegacyHadoopConfigurationSource;
@@ -174,15 +175,6 @@ public class SCMBlockLocationFailoverProxyProvider implements
     }
   }
 
-  public RetryAction getRetryAction(int failovers) {
-    if (failovers < maxRetryCount) {
-      return new RetryAction(RetryAction.RetryDecision.FAILOVER_AND_RETRY,
-          getRetryInterval());
-    } else {
-      return RetryAction.FAIL;
-    }
-  }
-
   private synchronized long getRetryInterval() {
     // TODO add exponential backup
     return retryInterval;
@@ -257,8 +249,11 @@ public class SCMBlockLocationFailoverProxyProvider implements
       @Override
       public RetryAction shouldRetry(Exception e, int retry,
                                      int failover, boolean b) {
-        performFailoverToAssignedLeader(newLeader);
-        return getRetryAction(failover);
+        if (!SCMHAUtils.isRetriableWithNoFailoverException(e)) {
+          performFailoverToAssignedLeader(newLeader);
+        }
+        return SCMHAUtils.getRetryAction(failover, retry, e, maxRetryCount,
+            getRetryInterval());
       }
     };
     return retryPolicy;
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
@@ -22,6 +22,7 @@ import com.google.common.annotations.VisibleForTesting;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hdds.conf.ConfigurationException;
 import org.apache.hadoop.hdds.conf.ConfigurationSource;
+import org.apache.hadoop.hdds.scm.ha.SCMHAUtils;
 import org.apache.hadoop.hdds.scm.ha.SCMNodeInfo;
 import org.apache.hadoop.hdds.scm.protocolPB.StorageContainerLocationProtocolPB;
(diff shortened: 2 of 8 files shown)


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMBlockLocationFailoverProxyProvider.java
```diff
@@ -257,8 +249,11 @@ public RetryPolicy getSCMBlockLocationRetryPolicy(String newLeader) {
       @Override
       public RetryAction shouldRetry(Exception e, int retry,
                                      int failover, boolean b) {
-        performFailoverToAssignedLeader(newLeader);
-        return getRetryAction(failover);
+        if (!SCMHAUtils.isRetriableWithNoFailoverException(e)) {
+          performFailoverToAssignedLeader(newLeader);
+        }
+        return SCMHAUtils.getRetryAction(failover, retry, e, maxRetryCount,
```
GlenGeng, line 255: It is weird here that `SCMHAUtils.getRetryAction` does not handle conditions that `SCMHAUtils.isRetriableWithNoFailoverException(e)` is true.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
```diff
@@ -180,14 +181,22 @@ public synchronized void close() throws IOException {
...
+  public RetryPolicy.RetryAction getRetryAction(int failovers, int retry,
+      Exception e) {
+    if (SCMHAUtils.isRetriableWithNoFailoverException(e)) {
+      if (retry < maxRetryCount) {
+        return new RetryPolicy.RetryAction(
+            RetryPolicy.RetryAction.RetryDecision.RETRY, getRetryInterval());
+      } else {
+        return RetryPolicy.RetryAction.FAIL;
+      }
+    } else if (failovers < maxRetryCount) {
```
GlenGeng, line 193: as above. let's wrap `FAILOVER_AND_RETRY` and `FAIL` together for failover exception.

GlenGeng (commented):
LGTM.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
```diff
@@ -329,6 +331,9 @@ public ScmContainerLocationResponse processRequest(
             "Unknown command type: " + request.getCmdType());
       }
     } catch (IOException e) {
+      if (SCMHAUtils.isRetriableWithNoFailoverException(e)) {
```
GlenGeng, line 334: Do we need this `RetriableWithNoFailoverException` ? 
Say a `ServiceException` containing `ReconfigurationInProgressException` should return `true` by `isRetriableWithNoFailoverException`.

GlenGeng: Given client side has already done the check, the server side can just return a `ServiceException` containing the real exception.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
GlenGeng: Please change it too.

bharatviswa504 (commented):
Do we want to address the suggested leader in a new Jira?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
bshashikant: > Do we want to address the suggested leader in a new Jira?

Yes, it will be addressed in a separate jira.

bshashikant:
> Do we want to address the suggested leader in a new Jira?

Yes.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
bshashikant: > Given client side has already done the check, the server side can just return a `ServiceException` containing the real exception.

The check is required because for only certain exceptions , we throw ServiceException. For others, it sets an error code instead of throwing a service exception.

GlenGeng:
+1. Thanks for the clarification.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/protocol/StorageContainerLocationProtocolServerSideTranslatorPB.java
bshashikant: > Given client side has already done the check, the server side can just return a `ServiceException` containing the real exception.

The Service exception is not thrown for all exceptions back from server. Only for certain exceptions, it throws service exception. Only StorageContainerLocation server seems to be sending back service exception back to client for each failure. I have remove the check in StorageContainerLocationServer.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/scm/proxy/SCMContainerLocationFailoverProxyProvider.java
```diff
@@ -180,14 +181,22 @@ public synchronized void close() throws IOException {
     }
   }
 
-  public synchronized RetryPolicy.RetryAction getRetryAction(int failovers) {
-    if (failovers >= maxRetryCount) {
+  public RetryPolicy.RetryAction getRetryAction(int failovers, int retry,
+      Exception e) {
+    if (SCMHAUtils.isRetriableWithNoFailoverException(e)) {
```
GlenGeng, line 186: by the way, can this method replaced with `SCMHAUtils#getRetryAction` ?

bshashikant:
Committing this. Will take care of other suggested changes in https://issues.apache.org/jira/browse/HDDS-5051.

PR #2127, opened 2021-04-07
HDDS-5071. [SCM HA Security] Fix duration of sub-ca certs.

bharatviswa504:

Use max duration for sub-ca certs similar to self-signed root certs.
Fix description of config in ozone-default.xml
When max duration is set less than cert duration fail.

https://issues.apache.org/jira/browse/HDDS-5071

## How was this patch tested?

Added test


bshashikant (approved):

bshashikant:
Thanks @bharatviswa504 for the work.

PR #2129, opened 2021-04-07
HDDS-5026. Cancel failing PR workflow runs

adoroszlai:

Currently most CI checks run to completion even if other checks fail.  HDDS-4933 added cancellation for concurrent builds of the same check (eg. _acceptance_).  This change extends the separate _Cancelling_ workflow introduced in HDDS-4988, to abort PR builds with any checks already failing.  (Eg. avoid waiting 1.5 hours for acceptance if checkstyle has error).

https://issues.apache.org/jira/browse/HDDS-5026

## How was this patch tested?

Tested similar change in my fork.

### DIFF
--- a/.github/workflows/cancel-ci.yaml
+++ b/.github/workflows/cancel-ci.yaml
@@ -30,5 +30,12 @@ jobs:
           cancelFutureDuplicates: true
           token: ${{ secrets.GITHUB_TOKEN }}
           sourceRunId: ${{ github.event.workflow_run.id }}
-          notifyPRCancel: true
           skipEventTypes: '["push", "schedule"]'
+      - uses: potiuk/cancel-workflow-runs@a81b3c4d59c61e27484cfacdc13897dd908419c9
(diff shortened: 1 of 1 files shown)


elek (approved):
Thanks the patch @adoroszlai 

Original I had a feeling that it might be a too strict rule, and we may have more builds due to this change (in case of two failures in two long-running jobs, the second one may be hidden due to the stop and all the other tests should be repeated...)

But on the other hand I learned it during the weekend that Ozone is responsible for the 6.7 % of all the builds (and 3.9% percent used from the total available resources by Ozone)

image

image

And using own forks it's easy to pre-run all the tests and create the PR only if all the tests are passed (unfortunately it's harder to do when PR is already opened).

TLDR;  I am +1, let's try this out and monitor our build usage.

adoroszlai:
Thanks @elek for reviewing and committing this, and for collecting and sharing the numbers.

PR #2131, opened 2021-04-08
HDDS-5074. Bump Guava version

elek:
JIRA: https://issues.apache.org/jira/browse/HDDS-5074

Guava has a tmp directory related CVE ([CVE-2020-8908](https://github.com/advisories/GHSA-5mg8-w23w-74h3)).

Based on my best knowledge Ozone is not affected, but it's hard to explain this situation for all the automated tools. Let's just bump the version to the latest one... 

## How was this patch tested?

Full CI passed.

And checked if guava jar is really updated in the final distribution:

image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,7 +156,7 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <spotbugs.version>3.1.12</spotbugs.version>
     <dnsjava.version>2.1.7</dnsjava.version>
 
-    <guava.version>28.2-jre</guava.version>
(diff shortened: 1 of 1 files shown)


adoroszlai (approved):
Thanks @elek for the upgrade.

jojochuang:
LGTM
I just want to say that the description of that alert isn't accurate. Simply updating to guava 30 doesn't help. It requires removing the API call usage in the code. Fortunately we don't seem to use it in Ozone.

elek:
Thanks, the review @adoroszlai and @jojochuang. Agree with the comment, it's not a security fix. Just makes less  false-positive problems reported... Merging it now...

PR #2136, opened 2021-04-09
HDDS-5073. Use ReplicationConfig on client side 

elek:

HDDS-5011 introduced new ReplicationConfig with multiple implementations (Ratis/Standalone).

To make ozone sh client EC compatible we can use ReplicationConfig on the client side too: --replication parameter can be parsed from the string based on the ReplicationConfig. Today it can be THREE or ONE, but later ECReplicationConfig can support more sophisticated schemes (like 3:2)

https://issues.apache.org/jira/browse/HDDS-5073

## How was this patch tested?

With local docker-compose cluster + CI

No logic has been changed only the internal representation of the data.


sodonnel (approved):
LGTM. This follows on nicely from #2096.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/RatisReplicationConfig.java
```diff
@@ -35,6 +35,16 @@ public RatisReplicationConfig(ReplicationFactor replicationFactor) {
     this.replicationFactor = replicationFactor;
   }
 
+  public RatisReplicationConfig(String factorString) {
+    ReplicationFactor factor = null;
+    try {
+      factor = ReplicationFactor.valueOf(Integer.parseInt(factorString));
+    } catch (NumberFormatException ex) {
+      factor = ReplicationFactor.valueOf(factorString);
```
symious, line 43: Should we throw an exception explicitly here?

symious:
I saw there are some classes and tests still using the legacy replicationType and replicationFactor, for example, ObjectEndpiont.put(). 
The source branch is named "client-only" and if the work of updating other classes hasn't started, maybe I can help with the work?

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
```diff
@@ -82,17 +84,19 @@ protected void execute(OzoneClient client, OzoneAddress address)
...
-
     if (replicationType == null) {
       replicationType = ReplicationType.valueOf(
-          getConf().get(OZONE_REPLICATION_TYPE,
-              OZONE_REPLICATION_TYPE_DEFAULT));
+          getConf()
+              .get(OZONE_REPLICATION_TYPE, OZONE_REPLICATION_TYPE_DEFAULT));
     }
 
+    if (replication == null) {
```
mukul1987, line 93: why, do we want to convert replication factor to a string ? What are the expected strings which we expect after EC implementation ?

elek: Good question.

I think we can start with `d:p` pattern where `d` is the number of data parts and `p` is the number of the parity part. But we can also extend it later with codec. At first, we can support `3:2` and `6:4` and `10:3`

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/RatisReplicationConfig.java
elek: In general the expected values are `ONE` and `THREE` but this code adds support to parse `ozone.replication` which can be 1 and 3 today. Long-term it will enable to use `3:2` in the `ozone.replication` (but we also need to configure EC as default `replicationType`)

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/StorageInfo.java
```diff
@@ -133,7 +133,8 @@ public void setClusterId(String clusterId) {
   }
 
   private void verifyNodeType(NodeType type)
-      throws InconsistentStorageStateException {
+      throws
```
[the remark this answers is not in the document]
elek: yes, reverting it. Thanks for the notification.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/ReplicationFactor.java
```diff
@@ -71,6 +71,8 @@ public static ReplicationFactor fromProto(
     }
   }
 
+
```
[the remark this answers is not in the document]
elek: reverted, thanks.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/client/RatisReplicationConfig.java
elek: > Should we throw an exception explicitly here?

Currently, `IllegalArgumentException` is thrown with the message:

```
Exception in thread "main" java.lang.IllegalArgumentException: No enum constant org.apache.hadoop.hdds.protocol.proto.HddsProtos.ReplicationFactor.asd
```
I think the type (`IllegalArgumentException`) is fine, but I modified the error message with throwing explicit exception:

```
Exception in thread "main" java.lang.IllegalArgumentException: Invalid RatisReplicationFactor 'xxx'. Please use ONE or THREE
```

symious: Thanks for the update. At first, I thought ReplicationFactor is only for legacy configs and won't be used for EC schemas.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
mukul1987: Why do we want this config to be a string ? can cant we expose this as known set of enums ? cc @umamaheswararao @arp7

elek: That's a good question (thanks for it), and I think there are multiple parts.

1. This is only for `ozone sh` commands not for s3 (where we have storage classes) or for o3fs/ofs (where we plan to configure it cluster wide or bucket level. 

2. Originally I had the same idea. `ReplicationConfig` today can have multiple parameters. (For example `3:2` can be defined as `new ECReplicationConfig(3,2)`). I thought that it might be better to accept only enums (or pre-defined strings) which are further defined by the administrators (or hard-coded by the developers):
 
For example:

| replication type | enum / group name | replication config | 
|----------------------------|---|---|
| RATIS | ONE | `new RatisReplicationConfig(ONE) |
| RATIS | THREE | `new RatisReplicationConfig(THREE) |
| STANDALONE | ONE | `new StandaloneReplicationConfig(ONE) |
| EC | `3:2` | `new EcReplicationConfig(3:2)` |
| EC | `6:4` | `new EcReplicationConfig(6:4)` |
| EC | `10:3` | `new EcReplicationConfig(10:3)` |

But when we add the `codec` to `ECReplicationConfig` (`RS` vs `XOR`) it will create more and more entries in this table.

 1. I think (at least at long term) administrators should have the flexibility to define new groups (shouldn't be everything hardcoded in the code)
 2. I think it's an additional complexity to manage all of these mappings. 
 
As `ozone sh` is a low level cli tool, it seems to be easier to avoid this grouping and just enable to directly configure all possible parameters (especially as there is no limit on the SCM/OM side, we can handle different configuration without any problem)

3. Compatibility

I also felt that this approach is a more natural extension to the current scheme. We can introduce a new abstraction level (with new cli parameters like `--storage-type`) where we accept names instead of low level parameters:

| name | replication type |  replication config | 
|----------------------------|---|---|
| REDUCED | RATIS |  `new RatisReplicationConfig(ONE) |
| DEFAULT | RATIS | | `new RatisReplicationConfig(THREE) |
| STANDALONE  | `new StandaloneReplicationConfig(ONE) |
| EC_3_2 | EC  | `new EcReplicationConfig(3:2)` |

But I am not sure if we need this abstraction level, especially at the beginning of the feature branch.

4. Enum vs string 

When you asked about enum, we can talk about two parts: do we need pre-defined groups AND/OR do we need to store in enum the groups?

I learned with the port changes that enum is very dangerous. We introduced a new service on DN and added a new enum value for the service discovery API and introduced a lot of backward compatibility problem (new enum values couldn't be deserialized with old clients). String seems to be more flexible (we can validate it after protobuf conversion).

This example doesn't fully apply here, as it's a full client-only change. (on protobuf this string is not sent), but I learned that enums (in some cases) can be harder

TLDR: This is why I think the string based approach is better here, but these are somewhat subjective points, please add your opinion.

mukul1987: I feel keeping values as strings opens up the whole pandora box of validations to do.

Is there a list of terminal policies this config will translate to ? I mean with client to server communication, do we plan to use string to pass the code info ?

elek: No. For client to server communication we use protobuf structures where each of the fields (data/parity and fator/type) are serialized. String is used only to easily select the right `ReplicationConfiguration` instance.

Please also note that for `RatisReplicationConfig` and `StandaloneReplicationConfig` the string is converted to an enum.

For the client parameter we need a string as we don't know how the value is parsed. It depends on the `ReplicationConfig` implementation.

However, the string parsing in `ECReplicationConfig` can be more strict. We can fix which data/parity pairs are acceptable today (with either enum or some rules) I didn't do in this patch as it can make the testing harder in current phase. Today I can create pipeline where data=2 parity=1. It's not something good for production but for testing it requires only 3 nodes.

elek:
ping @mukul1987, @umamaheswararao

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
mukul1987: I am not comfortable with the idea of an open string-based config. ENUMS are not dangerous every time when we transfer them over the network.

arp7: I don't like the idea of an open string based config either. ENUMs are not inherently dangerous. Strings will be far worse.

The bigger issue is a lack of discipline in thinking through compatibility concerns.

elek: > ENUMS are not dangerous every time when we transfer them over the network.

Thanks the comment @mukul1987 and @arp7  

As we discussed during the EC meeting it's NOT about protobuf changes. It's a client side only change.

In general: we need multiple type of replication configs. For example:

```
new RatisReplicationConfig(ONE)
new RatisReplicationConfig(THREE)
new StandaloneReplicationConfig(ONE)
new StandaloneReplicationConfig(THREE)
new ECReplicationConfig(2,1,XOR)
new ECReplicationConfig(4,3,RS)
new ECReplicationConfig(6,4,RS)
new ECReplicationConfig(10,1,RS)
```

All of these configs are strongly typed both in Java code (OM/SCM) and protobuf. ONE/THREE are enums as before.


The only problem here is that Enum types / replication definitions  **depend on the replication config type**. EC may need different enum values. First we need to choose the replication config based on the type (EC/RATIS/STANDLAONE) and **after that** we can ask the selected `ReplicationConfig` to parse the string to `Enum`. This is what we do in this patch.

It's a separated discussion if we need to validate the input string in `ECReplicationConfig` based on predefined enums or based on smart validation rules. Both can work, and please add your opinion to https://issues.apache.org/jira/browse/HDDS-5247  

> The bigger issue is a lack of discipline in thinking through compatibility concerns.

Thanks the feedback @arp7 

Can you please share your compatibility concerns? 

This is a client side change and as you can see none of the smoketests are changed, so we can assume it's fully compatible. `ozone.replication` is also changed, but it accepts both integer and string existing values work as before.

elek:
@mukul1987 @umamaheswararao @arp7: Can we merge it? Do you have any more concerns?

mukul1987 (approved):
THe current patch looks ok, lets do a followup patch for the validation.

hadoop-ozone/tools/src/main/java/org/apache/hadoop/ozone/shell/keys/CopyKeyHandler.java
arp7: +1 from me also with client-only change (i.e. no strings on the wire)

elek:
Thanks the review @sodonnel, @symious @mukul1987 @arp7 

I am merging it now...

PR #1617, opened 2020-11-24
HDDS-4471. GrpcOutputStream length can overflow

wycccccc:
## What changes were proposed in this pull request?
Modified the type to prevent overflow.

## What is the link to the Apache JIRA
https://issues.apache.org/jira/browse/HDDS-4471

## How was this patch tested?
Not tested.

PR #2139, opened 2021-04-09
HDDS-5083. Bump version of common-compress

elek:
JIRA: https://issues.apache.org/jira/browse/HDDS-5083

This is similar to #2131. Based on my understanding Ozone is not affected by the reported [CVE-2018-11771](https://github.com/advisories/GHSA-hrmr-f5m6-m9pq)

But as it's very cheap to bump the version, it can be better to do it (and avoid the false-positive notification from the security scanners)

## How was this patch tested?

Full CI test.


mukul1987:
#1617 fixes a buffer overflow, I guess the PR number is wrong here :)

elek:
Ups, thanks for the notification. You are right, I mean #2131 (guava update)

adoroszlai (approved):
Thanks @elek for the bump.  Note that `commons-compress` no longer pulls `org.tukaani:xz`, so it may be removed from the license.

elek:
> so it may be removed from the license.

Good point. I am planning to create a simple check to void some mistakes. I think we can commit the current list of the jars and compare it during the build. (Similar to the `proto` dependency check).

Will create a PR soon based on this experiment: https://github.com/elek/ozone/commit/2dc476e418743af0eaafd5b667b076ea91495799 (if it works on the fork)

elek:
Merging it after the green build. Thanks the review @adoroszlai

PR #2140, opened 2021-04-09
HDDS-5084. Include HISTORY.md/SECURITY.md/CONTRIBUTING.md in the release artifacts.

elek:
JIRA: https://issues.apache.org/jira/browse/HDDS-5084

During the ozone-1.1.0 vote I realized that `HISTORY.md`/`SECURITY.md`/`CONTRIBUTING.md` files are missing from the `bin` and `src` artifacts. I think they include very useful information and would be better to include them in the release artifacts.

## How was this patch tested?

```
mvn clean install -Dmaven.javadoc.skip=true -DskipTests -Psign,dist,src -Dtar -Dgpg.keyname=$CODESIGNINGKEY
  cd hadoop-ozone/dist/target/
tar tzf hadoop-ozone-1.1.0-SNAPSHOT.tar.gz
tar tzf hadoop-ozone-1.1.0-src-SNAPSHOT.tar.gz
```

### DIFF
--- a/hadoop-ozone/dist/dev-support/bin/dist-layout-stitching
+++ b/hadoop-ozone/dist/dev-support/bin/dist-layout-stitching
@@ -67,6 +67,9 @@ run cp -p "${ROOT}/hadoop-ozone/dist/src/main/license/bin/LICENSE.txt" "LICENSE.
 run cp -pr "${ROOT}/hadoop-ozone/dist/src/main/license/bin/licenses" "licenses"
 run cp -p "${ROOT}/hadoop-ozone/recon/src/main/resources/webapps/recon/ozone-recon-web/LICENSE" "licenses/LICENSE-ozone-recon.txt"
 run cp -p "${ROOT}/README.md" .
+run cp -p "${ROOT}/HISTORY.md" .
+run cp -p "${ROOT}/SECURITY.md" .
+run cp -p "${ROOT}/CONTRIBUTING.md" .
 
 run mkdir -p ./share/ozone/web
 run mkdir -p ./bin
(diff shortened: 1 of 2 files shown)


mukul1987 (changes requested):
We should update the History.md, it is quite old right now. Including it in the current form in the release doesn't makes sense.

adoroszlai:
> We should update the History.md, it is quite old right now.

@mukul1987 This is a good suggestion, but I disagree with the other statement:

> Including it in the current form in the release doesn't makes sense.

1. It is not going to be part of the 1.1.0 release, so there is plenty of time to update it until the next one.
2. It's not only about HISTORY.md, but other two docs as well.
3. Writing prose for the history doc is quite distinct from updating a script to copy some files.  It may very well be updated by other people, not necessarily @elek.

So I think this change is fine in its scope.

mukul1987:
Fair point @adoroszlai. Can we please create a followup jira and mark it as a blocker for 1.2.0 release? I feel if we will update the history by the next release, then we should be good.

elek:
Thanks the suggestion @mukul1987, very good point.

It seems to be a small update, so I created the patch itself (please see #2149). And agree: as we have PRs for both problems in our radar we can merge the two PRs in any order.

mukul1987:
Thanks Marton. +1 for this patch as well.
I have already added +1 to the other patch.

Thanks for updating the file.

mukul1987 (approved):
+1,lgrm

ayushtkn (approved):

elek:
Thanks the review @mukul1987 @ayushtkn and @adoroszlai 

I am merging it after the green build.

PR #2142, opened 2021-04-09
HDDS-4992. SCM should not use pipelines with HEALTHY_READONLY datanodes

errose28:

For datanode heartbeats to SCM: Close pipelines a DN is involved in if its software or metadata layout version no longer matches SCM.

For datanode registration with SCM: Block registration if the DN's software layout version does not match SCM's. Allow nodes with mismatched metadata layout version in, but do not allow them in pipelines.

## What is the link to the Apache Jira?

HDDS-4992

## How was this patch tested?

Unit tests added.


hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestSCMNodeManager.java
```diff
@@ -113,6 +117,18 @@
   private File testDir;
   private StorageContainerManager scm;
 
+  private static final int MAX_LV = HDDSLayoutVersionManager.maxLayoutVersion();
+  private static final LayoutVersionProto LARGER_SLV_LAYOUT_PROTO =
+      toLayoutVersionProto(MAX_LV, MAX_LV + 1);
+  private static final LayoutVersionProto SMALLER_SLV_LAYOUT_PROTO =
+      toLayoutVersionProto(MAX_LV, MAX_LV - 1);
```
avijayanhwx, line 124: nit. Just to keep things theoretically correct, can we also set the metadata version to MAX_LV - 1. (Since MLV cannot be > SLV)

avijayanhwx (approved):
Thanks for working on this @errose28. One minor comment left inline.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/node/TestSCMNodeManager.java
errose28: Done

PR #2144, opened 2021-04-09
HDDS-5087. Ozone RPC client leaks KeyProvider instances.

xiaoyuyao:
Maintain a KeyProvider cache for Ozone RPC client and ensure Key provider is closed upon removal from the cache. 

https://issues.apache.org/jira/browse/HDDS-5087

## How was this patch tested?
UT added.

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneAtRestEncryption.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneAtRestEncryption.java
@@ -493,4 +493,11 @@ public class TestOzoneAtRestEncryption {
         offset + " and length " + readData.length,
         inputDataForComparison, readData);
   }
+
+  @Test
+  public void testGetKeyProvider() throws Exception {
+    KeyProvider kp1 = store.getKeyProvider();
+    KeyProvider kp2 = store.getKeyProvider();
+    Assert.assertEquals(kp1, kp2);
+  }
 }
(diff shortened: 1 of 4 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneAtRestEncryption.java
```diff
@@ -493,4 +493,11 @@ private static void assertReadContent(byte[] inputData, byte[] readData,
         offset + " and length " + readData.length,
         inputDataForComparison, readData);
   }
+
+  @Test
+  public void testGetKeyProvider() throws Exception {
+    KeyProvider kp1 = store.getKeyProvider();
+    KeyProvider kp2 = store.getKeyProvider();
+    Assert.assertEquals(kp1, kp2);
+  }
```
vivekratnavel, line 502: Can we add another test to check if the key provider gets closed when the RPC client is closed?

vivekratnavel (commented):
LGTM. Just a minor comment inline.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/client/rpc/TestOzoneAtRestEncryption.java
xiaoyuyao: I had thought about it but the guava cache removal call back is tricky to test. Spend some time in the morning and figure out a way to use mockito spy to test this indirectly.

vivekratnavel (approved):
Thanks for working on this. +1 LGTM

vivekratnavel:
We have to run the CI again to get a clean run before we can merge. Can you please push an empty commit to trigger CI?

xiaoyuyao:
Thanks @vivekratnavel  for the review. PR has been merged.

PR #2145, opened 2021-04-10
HDDS-4842. Add timestamp to Revoked Certs table in SCM DB

vivekratnavel:

Add timestamp to REVOKED_CERTS table in SCM db. 

Instead of modifying the definition of the existing column family, this patch introduces a new column family to make it easier for the finalize upgrade action to support upgrades from older versions.

The changes required to make it backward compatible is tracked here - https://issues.apache.org/jira/browse/HDDS-5088
The changes required to make the persistence happen via Ratis is tracked here - https://issues.apache.org/jira/browse/HDDS-5095

https://issues.apache.org/jira/browse/HDDS-4842

## How was this patch tested?

Unit tests

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/CertInfo.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/CertInfo.java
@@ -24,6 +24,7 @@ import org.apache.hadoop.hdds.security.x509.certificate.utils.CertificateCodec;
 import org.jetbrains.annotations.NotNull;
 
 import java.io.IOException;
+import java.io.Serializable;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.Comparator;
@@ -33,9 +34,10 @@ import java.util.Objects;
  * Class that wraps Certificate Info.
  */
 public class CertInfo implements Comparator<CertInfo>,
-    Comparable<CertInfo> {
+    Comparable<CertInfo>, Serializable {
 
   private X509Certificate x509Certificate;
+  // Timestamp when the certificate got persisted in the DB.
   private long timestamp;
 
   private CertInfo(X509Certificate x509Certificate, long timestamp) {
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/CertInfo.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/CertInfo.java
@@ -1,0 +1,169 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.hadoop.hdds.security.x509.certificate;
+
+import org.apache.hadoop.hdds.protocol.proto.HddsProtos;
+import org.apache.hadoop.hdds.security.exception.SCMSecurityException;
+import org.apache.hadoop.hdds.security.x509.certificate.utils.CertificateCodec;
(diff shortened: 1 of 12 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
```diff
@@ -75,6 +80,15 @@
           X509Certificate.class,
           new X509CertificateCodec());
 
+  public static final DBColumnFamilyDefinition<BigInteger, CertInfo>
+      REVOKED_CERTS_V2 =
```
GlenGeng, line 84: @bharatviswa504 Is `REVOKED_CERTS_V2` needed to be replicated across multi SCMs if SCM HA Is on?

bharatviswa504: Yes, it should be replicated across SCM's.

vivekratnavel: Thanks for bringing this up @GlenGeng and @bharatviswa504. I have created a separate jira to fix the persistence via Ratis - https://issues.apache.org/jira/browse/HDDS-5095 since we also need to fix the valid certs table.

vivekratnavel: It seems like a small change. So, I will update the same patch to get "REVOKED_CERTS_V2" replicated.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/CertInfo.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.Objects;
+
+/**
+ * Class that wraps Certificate Info.
+ */
+public class CertInfo implements Comparator<CertInfo>,
+    Comparable<CertInfo> {
+
+  private X509Certificate x509Certificate;
+  private long timestamp;
```
xiaoyuyao, line 39: Can you add the meaning/intended usage of timestamp here?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
xiaoyuyao: @vivekratnavel  and I have an offline discussion on this. Both crl sequence id and revoked certs replication will be taken care by HDDS-5095 via Replicate annotation on the CertificateStore#revokeCertificats

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/metadata/SCMDBDefinition.java
```diff
@@ -141,7 +155,7 @@ public String getLocationConfigKey() {
   @Override
   public DBColumnFamilyDefinition[] getColumnFamilies() {
     return new DBColumnFamilyDefinition[] {DELETED_BLOCKS, VALID_CERTS,
-        VALID_SCM_CERTS, REVOKED_CERTS, PIPELINES, CONTAINERS, TRANSACTIONINFO,
-        CRLS, CRL_SEQUENCE_ID, SEQUENCE_ID};
+        VALID_SCM_CERTS, REVOKED_CERTS_V2, PIPELINES, CONTAINERS,
```
xiaoyuyao, line 158: Should we keep the REVOKED_CERTS and add a separate take as V2 for compatibility? Otherwise, it might break on upgrade.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMCertStore.java
```diff
@@ -174,8 +175,12 @@ public void checkValidCertID(BigInteger serialID) throws IOException {
           // only if the revocation time has passed.
           if (now.after(revocationTime) || now.equals(revocationTime)) {
             for (X509Certificate cert : certsToRevoke) {
+              CertInfo certInfo = new CertInfo.Builder()
+                  .setX509Certificate(cert)
+                  .setTimestamp(now.getTime())
+                  .build();
               scmMetadataStore.getRevokedCertsTable()
-                  .putWithBatch(batch, cert.getSerialNumber(), cert);
+                  .putWithBatch(batch, cert.getSerialNumber(), certInfo);
```
xiaoyuyao, line 183: here we might provide some migration from revokecert table to revokecert v2 table.

xiaoyuyao (commented):
Thanks @vivekratnavel  for working on this. The patch LGTM overall. A few comments added inline...

xiaoyuyao:
@vivekratnavel TestSCMCertStore.testRevokeCertificates failure seems related.

xiaoyuyao:
Thanks @vivekratnavel  for the update. LGTM, +1.  I will merge it shortly,.

PR #2146, opened 2021-04-12
HDDS-5075. NPE during secure SCM initialization with HA code updated to an already existing cluster

bharatviswa504:

Fixed the issue of SCM not coming up after the upgrade, and also SCM Client commands failing due to NPE with setX509RootCACertificate. For more info refer to Jira for stack trace.

https://issues.apache.org/jira/browse/HDDS-5078

## How was this patch tested?

Tested it manually. Thank You @fapifta for reporting and also for trying out the fix.


xiaoyuyao (approved):
Thanks @bharatviswa504 for fixing this. LGTM, +1.

PR #2148, opened 2021-04-12
HDDS-5090. make Decommission work under SCM HA.

GlenGeng:

**The problem**
The decommission/maintenance info is saved in memory of SCM, and if SCM is restarted, it relearns this info during re-register of Datanode.

Only leader SCM handles the decommissionNodes(), recommissionNodes(), startMaintenanceNodes() request, and not replicate these info to follower SCM, thus when failover happens, the new leader SCM will lose this info, since they are saved in memory of previous leader SCM.

**Current status**
If a SCM is restarted, then upon re-registration the datanode will already be in DECOMMISSIONING or ENTERING_MAINTENANCE or IN_MAINTENANCE state. In that case, it needs to be added back into the monitor to track its progress.

For a registered node, the information stored in SCM is the source of truth. If SCM finds that the opState or opStateExpiryEpoch is different from what it saves in memory, it will send SetNodeOperationalStateCommand to update the Datanode.

**The solution**
leader SCM -hb> DN --hb-> follower SCM

- Leader SCM updates PersistedOpState of Datanode via heartbeat. Datanode update OpState in follower SCM via heartbeat.
- When follower SCM becomes leader, it calls continueAdminForNode for all datanode, so that the DECOMMISSIONING, ENTERING_MAINTENANCE, IN_MAINTENANCE datanode will be added back to the monitor.

**Disadvantage**
The same as now, if leader SCM records the info, notifies Datanode via heartbeat, but steps down before Datanode notifies follower SCM via heartbeat, that info will be lost in the new leader SCM.

As discussed with Stephen O'Donnell, we can live with the rare event of a decommission starting and SCM failing over before the state has made it to the DNs.

For details: https://docs.google.com/document/d/1N5PsUuLBGgvkYFQgDumvRZujc-9RcDwoE0SubZcLUzY/edit?usp=sharing

https://issues.apache.org/jira/browse/HDDS-5090

## How was this patch tested?

Tested in tencent internal integration environment, the operationalState can be replicated between multi SCMs, and survive failover of SCM.

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeDecommissionManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeDecommissionManager.java
@@ -252,6 +252,7 @@ public class NodeDecommissionManager {
     if (opState == NodeOperationalState.DECOMMISSIONING
         || opState == NodeOperationalState.ENTERING_MAINTENANCE
         || opState == NodeOperationalState.IN_MAINTENANCE) {
+      LOG.info("Continue admin for datanode {}", dn);
       monitor.startMonitoring(dn);
     }
   }
@@ -375,4 +376,20 @@ public class NodeDecommissionManager {
     return nodeManager.getNodeStatus(dn);
   }
 
+  /**
+   * Called in SCMStateMachine#notifyLeaderChanged when current SCM becomes
+   *  leader.
+   */
+  public void onBecomeLeader() {
+    nodeManager.getAllNodes().forEach(datanodeDetails -> {
+      try {
+        continueAdminForNode(datanodeDetails);
+      } catch (NodeNotFoundException e) {
+        // Should not happen, as the node has just registered to call this event
+        // handler.
+        LOG.warn("NodeNotFound when adding the node to the decommissionManager",
+            e);
+      }
+    });
+  }
 }
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeDecommissionManager.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeDecommissionManager.java
@@ -25,6 +25,7 @@ import org.apache.hadoop.hdds.scm.DatanodeAdminError;
 import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.hdds.scm.container.ContainerManagerV2;
 import org.apache.hadoop.hdds.scm.container.ReplicationManager;
+import org.apache.hadoop.hdds.scm.ha.SCMContext;
 import org.apache.hadoop.hdds.scm.node.states.NodeNotFoundException;
 import org.apache.hadoop.hdds.server.events.EventPublisher;
 import org.apache.hadoop.hdfs.DFSConfigKeys;
@@ -52,6 +53,7 @@ public class NodeDecommissionManager {
 
   private NodeManager nodeManager;
   //private ContainerManagerV2 containerManager;
+  private SCMContext scmContext;
   private EventPublisher eventQueue;
   private ReplicationManager replicationManager;
   private OzoneConfiguration conf;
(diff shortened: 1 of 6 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/SCMNodeManager.java
```diff
@@ -413,38 +413,56 @@ boolean opStateDiffers(DatanodeDetails dnDetails, NodeStatus nodeStatus) {
...
 
   /**
-   * If the operational state or expiry reported in the datanode heartbeat do
-   * not match those store in SCM, queue a command to update the state persisted
-   * on the datanode. Additionally, ensure the datanodeDetails stored in SCM
-   * match those reported in the heartbeat.
-   * This method should only be called when processing the
-   * heartbeat, and for a registered node, the information stored in SCM is the
-   * source of truth.
+   * This method should only be called when processing the heartbeat.
```
sodonnel, line 416: Could we add a test for this new logic - Command fired if leader, but not fired if follower?

I checked in TestSCMNodeManager, and there is a test "testSetNodeOpStateAndCommandFired" which I earlier set to ignore as it became invalid when I developed decommission. I then forgot to go back and fix it. 

You could use that tests as a starting point, and change it to call `processHeartBeat` with the correct DatanodeDetails to trigger a command.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/node/NodeDecommissionManager.java
```diff
@@ -375,4 +376,20 @@ private NodeStatus getNodeStatus(DatanodeDetails dn)
     return nodeManager.getNodeStatus(dn);
   }
 
+  /**
```
sodonnel, line 379: We should probably add a test for this - ie ensure that nodes are added to the decommission workflow when the onBecomeLeader() event is fired.

sodonnel (changes requested):
Thanks for working on this @GlenGeng. The change looks mostly good, but I think we should add a couple of unit tests where I have commented inline.

Also, it might be a good idea to add an Integration test or two. There are quite a few integration tests in the TestDecommissionAndMaintenance class.

Finally, when node registers for the first time on a restart SCM, the `NewNodeHandler` event is fired. Inside it, there is:

```
      if (datanodeDetails.getPersistedOpState()
          != HddsProtos.NodeOperationalState.IN_SERVICE) {
        decommissionManager.continueAdminForNode(datanodeDetails);
      }
```

So inside the above event handler, or inside `NodeDecommissionManager.continueAdminForNode` I think we will need the `isLeader()` check, and skip adding the DN to the Admin workflow if the SCM is not the leader.

GlenGeng (commented):
Thanks @sodonnel for the review. It's helpful!

test has been added. Please one more look.

>  I think we will need the isLeader() check, and skip adding the DN to the Admin workflow if the SCM is not the leader.

It is safe to add DN to the workflow for a follower SCM. The design of SCM HA is, all the event handler will keep running for a follower SCM, since 1) the SCMCommand sent out from a follower SCM will be ignored on Datanode, and 2) the scm.db is protected by ratis, any update to scm.db will fail with NotLeaderException 3) ReplicationManager and BackgrouondPipelineCreator has been stopped on a follower SCM. 

But I added the isLeader check to make behavior on follower SCM to be more straightforward.

> Also, it might be a good idea to add an Integration test or two.

I propose to replace `MiniOzoneCluster` with `MiniOzoneHAClusterImpl` in `TestDecommissionAndMaintenance`, and verify that the decommission info is correctly replicated across multi SCMs. I prefer to handle the work in a followup jira.

adoroszlai:
@GlenGeng I have updated the branch to avoid compile error in case of merge to `master`, where now another test is `@Ignore`-d.  Thus we cannot remove `import org.junit.Ignore;`.  This kind of conflict is not detected by Github, so it would not prevent merge.

GlenGeng:
Thanks @adoroszlai ! This is fine to me, since the it can pass check style

sodonnel (approved):
Updated tests look good to me. Feel free to commit the changes.

GlenGeng:
Thanks @sodonnel for the view. I will merge it.

PR #2149, opened 2021-04-12
HDDS-5091. Add project separation and first stable release to the HISTORY.md

elek:
JIRA: https://issues.apache.org/jira/browse/HDDS-5091

As @mukul1987 notified us in #2140 the `HISTORY.md` is not updated recently. I think the two main events since the last update are the first stable release and the project separation to TLP.

I added a paragraph about these events. 

## How was this patch tested?

This is a readme change. No test is required just :eye:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -60,4 +60,4 @@ In October 2019, the Ozone sources were moved out to the [apache/hadoop-ozone](h
 
 In March 2020, the first beta release ([Ozone 0.5.0](https://hadoop.apache.org/ozone/release/0.5.0-beta/) was published, after the earlier alpha releases from 2018 and 2019.
 
-In September/October 2020 -- just after the first stable release -- Hadoop community decided to move Ozone to a separated, top level Apache project. Apache Ozone is established by the board of the Apache Software Foundation during the October 2020 [board meeting](https://www.apache.org/foundation/records/minutes/2020/board_minutes_2020_10_21.txt).
+In September/October 2020 -- just after the first 1.0.0 stable release -- Hadoop community decided to move Ozone to a separated, top level Apache project. Apache Ozone is established by the board of the Apache Software Foundation during the October 2020 [board meeting](https://www.apache.org/foundation/records/minutes/2020/board_minutes_2020_10_21.txt).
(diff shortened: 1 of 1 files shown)


mukul1987 (approved):
+1, LGTM

mukul1987:
Thanks for updating the history Marton. I agree that we should try to keep updating it on every release. But thanks anyways for this update. I really appreciate this.

elek:
> I agree that we should try to keep updating it on every release. 

Thanks for reporting the problem. Yes, we can update it with every release, but at least at the end of the year with can add a quick overview. We may not need to mention all the features, but the big improvements (like SCM-HA) should definitely be added...

Will try to keep in my mind and not forget it...

bharatviswa504:
@elek 
Do we want to mention about 1.0 release which we declare production-ready/GA and removed beta and also about the OM HA feature released as part of 1.0.0?

elek:
>  Do we want to mention about 1.0 release which we declare production-ready/GA and removed beta and also about the OM HA feature released as part of 1.0.0?

Thanks for the idea @bharatviswa504, the first stable release was mentioned:

> In September/October 2020 -- just after the first stable release...

I just added the `1.0.0` version number to here.

Personally I don't feel that we need to add all the release and all the features to here, for me, this HISTORY is more about the huge milestones. 1.0.0 is definitely a very large milestone, but I would mention HA when we achieve full HA (both OM and SCM HA are released).

But I can be easily convinced, this is just my opinion. I am fine even if somebody would like to include [the official roadmap](https://cwiki.apache.org/confluence/display/OZONE/Ozone+Roadmap) at the end of this doc.

If you have any suggested sentence, let me know, and I will include it.

bharatviswa504:
>>but I would mention HA when we achieve full HA (both OM and SCM HA are released).
I am fine with it if we want to mention once we have complete HA.

My thought process here is as part of the 1.0.0 stable release OM HA is completed and which is a big feature. But it also makes sense to mention once we have complete HA(both OM and SCM). I also don't have strong opinion here.

bharatviswa504 (approved):

elek:
Thanks for the answer @bharatviswa504. I am merging it now, and we can improve it later.

I would say something like this at the end of the year:

```
During 2021 Ozone had 2 (3?) main releases with significant new features and many small bug and security fixes. Full HA setup become available with supporting HA for the two metadata server compnent (Ozone Manager HA from 1.0.0 and Storage Container Manger HA from 1.2.0). 
```

PR #2153, opened 2021-04-14
HDDS-5068. build errors due to missing JAXB-runtime jar dependency for openjdk 11 se standalone

neils-dev:

Added maven build dependency _jaxb-runtime_ to each _pom.xml_ file in ozone build that requires _jaxb-api_ dependency.  Changes made to compile and run with Java 11 see _**https://javaee.github.io/jaxb-v2/doc/user-guide/ch03.html#deployment-maven-coordinates**_.  In summary,
`If client application needs to include the runtime, e.g. running standalone on JavaSE jaxb-runtime should be also included.
`
Fixes java 11 compile issues for unit testing ie. _hadoop-ozone/tools_ as well as ozone runtime commands such as _getconf_.

Made small changes to fix unit test: _hadoop-hdds/container-service/TestContainerData.Yaml_.  With java 11 compile, unit test fails due to comparison between expected and actual container date/time.  Failure occurs because time in date/time is truncated (precision error) in comparison.  Change _to use_ timestamp epochs in milliseconds.

https://issues.apache.org/jira/browse/HDDS-5068

## How was this patch tested?

Build without errors : 
`mvn clean install -Phdds -DskipTests=true -Dmaven.javadoc.skip=true -Pdist -Dtar -DskipShade`
and,
`hadoop-ozone/dev-support/checks/unit.sh`
with clean hadoop-hdds/container-service tests after patch for TestContainerDataYaml

**Note:** observed _intermittent errors_ associated with  _hdds-container-service_: _TestSchemaOneBackwardsCompatibility_ (problem due to thread-pool dispatching a non-blocking thread when testcase expects a blocking thread;  thread is not finished when test checks condition hence intermittent error).
 
junit_test_fail_time_truncation_assert

with fixed errors :
junit_fix_pass_test

**Also** tested with ozone getconf command (passes with patch, see jira for error reported by **@elek** ):
ozone_getconf_command_with_patch


adoroszlai (commented):
Thanks @neils-dev for reporting and fixing this problem.  Generally looks good.  I have one question:

Now both `com.sun.xml.bind:jaxb-core:jar:2.3.0.1` and `org.glassfish.jaxb:jaxb-core:jar:2.3.0` are on the classpath (eg. `hadoop-ozone-datanode`).  Are both of these necessary or can one of them be excluded?

```
+- com.sun.xml.bind:jaxb-core:jar:2.3.0.1:compile
+- javax.xml.bind:jaxb-api:jar:2.3.0:compile
+- org.glassfish.jaxb:jaxb-runtime:jar:2.3.0:compile
|  +- org.glassfish.jaxb:jaxb-core:jar:2.3.0:compile
|  |  +- org.glassfish.jaxb:txw2:jar:2.3.0:compile
|  |  \- com.sun.istack:istack-commons-runtime:jar:3.0.5:compile
|  +- org.jvnet.staxex:stax-ex:jar:1.7.8:compile
|  \- com.sun.xml.fastinfoset:FastInfoset:jar:1.2.13:compile
```

@elek do we need any license update for the new dependencies?

```
@@ -1,3 +1,4 @@
+FastInfoset-1.2.13.jar
 accessors-smart-1.2.jar
 activation-1.1.1.jar
 animal-sniffer-annotations-1.18.jar
@@ -88,6 +89,7 @@
 httpclient-4.5.13.jar
 httpcore-4.4.13.jar
 ion-java-1.0.2.jar
+istack-commons-runtime-3.0.5.jar
 j2objc-annotations-1.3.jar
 jackson-annotations-2.12.1.jar
 jackson-core-2.12.1.jar
@@ -115,7 +117,9 @@
 javax.servlet-api-3.1.0.jar
 jaxb-api-2.3.0.jar
 jaxb-core-2.3.0.1.jar
+jaxb-core-2.3.0.jar
 jaxb-impl-2.3.0.1.jar
+jaxb-runtime-2.3.0.jar
 jcip-annotations-1.0-1.jar
 jersey-cdi1x-2.33.jar
 jersey-client-2.33.jar
@@ -223,8 +227,10 @@
 spring-jdbc-5.2.11.RELEASE.jar
 spring-tx-5.2.11.RELEASE.jar
 sqlite-jdbc-3.25.2.jar
+stax-ex-1.7.8.jar
 stax2-api-3.1.4.jar
 token-provider-1.0.1.jar
+txw2-2.3.0.jar
 weld-servlet-2.4.7.Final.jar
 woodstox-core-5.0.3.jar
 xz-1.0.jar
```

neils-dev:
Thanks **@adoroszlai** for comments on classpath jars resulting from pom build file changes for _jaxb_ upgrade.  Yes, unintentionally included is the _org.glassfish jaxb-core_.  Apparently it is pulled in from the _org.glassfish jaxb-runtime_ upgrade.  Both _jaxb-core_ jars are unnecessary.  New commit removes the non _glassfish jaxb-core_ - thanks!  In addition also removed older _jaxb-impl_ dependency as it is the old jaxb runtime which is replaced with the new _org.glassfish jaxb-runtime_.  FYI : The following two jiras from another apache project follows what we are doing here: https://issues.apache.org/jira/browse/TIKA-2743,https://issues.apache.org/jira/browse/TIKA-2778.  **NOTE**: we are also pulling in an additional jar that we have dependency on, the _javax.activation_ - currently in the base pom file we have the version 1.1.1 dependency, however we have **_also_** the 1.2.0 version (I believe from the jaxb-runtime).  Tried updating the base pom.xml file to:         
```
<dependency>
        <groupId>javax.activation</groupId>
        <artifactId>activation</artifactId>
        <version>1.2.0</version>
      </dependency>
```
however this results in a recon build problem (see attached image).  Therefore _**left**_ the _javax.activation_ version at current 1.1.1.

recon_build_failure_activation_1 2 0

elek (approved):
> @elek do we need any license update for the new dependencies?

Yes, I think we need to update the bin/LICENSE. The final change seems to be quite small:

```
38d37
< ./share/ozone/lib/FastInfoset-1.2.13.jar
92d90
< ./share/ozone/lib/istack-commons-runtime-3.0.5.jar
120c118
< ./share/ozone/lib/jaxb-runtime-2.3.0.1.jar
---
> ./share/ozone/lib/jaxb-impl-2.3.0.1.jar
229d226
< ./share/ozone/lib/stax-ex-1.7.8.jar
231d227
< ./share/ozone/lib/txw2-2.3.0.1.jar
```

To make it easier I uploaded the suggested license changes. @adoroszlai Can you please review that part. I am +1 to the original commit.

It was quite hard to find the original sources (and maven jars include references to the LICENSE but not the original text), so I decided to copy the CDDL test itself. (we have a few instances under different names which can be removed in the future)

hadoop-ozone/dist/src/main/license/bin/LICENSE.txt
```diff
@@ -228,12 +228,12 @@ BSD
 CDDL
 =====================
 
+   com.sun.istack:istack-commons-runtime
    com.sun.jersey:jersey-core
    com.sun.jersey:jersey-json
    com.sun.jersey:jersey-server
    com.sun.jersey:jersey-servlet
    com.sun.xml.bind:jaxb-core
```
adoroszlai, line 236: I think this one can be removed, too.

adoroszlai (commented):
Thanks @elek for the license update.  LGTM, except one more removal may be possible.

hadoop-ozone/dist/src/main/license/bin/LICENSE.txt
elek: Yep. We already have the line for the new jaxb-core: `org.glassfish.jaxb:jaxb-core`

adoroszlai (approved):

PR #2155, opened 2021-04-14
HDDS-5103. Fix Install Snapshot Mechanism in SCMStateMachine.

bshashikant:
Fixed install snapshot behaviour

https://issues.apache.org/jira/browse/HDDS-5103

## How was this patch tested?

Enabled Unit tests

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
@@ -52,14 +52,8 @@ public class InterSCMGrpcClient implements SCMSnapshotDownloader{
   private final InterSCMProtocolServiceGrpc.InterSCMProtocolServiceStub
       client;
 
-  private final long timeout;
-
-  public InterSCMGrpcClient(final String host, final ConfigurationSource conf) {
-    Preconditions.checkNotNull(conf);
-    int port = conf.getInt(ScmConfigKeys.OZONE_SCM_GRPC_PORT_KEY,
-        ScmConfigKeys.OZONE_SCM_GRPC_PORT_DEFAULT);
-    timeout =
-        conf.getObject(SCMHAConfiguration.class).getGrpcDeadlineInterval();
+  public InterSCMGrpcClient(final String host, final int port,
+      final long timeout) {
     NettyChannelBuilder channelBuilder =
         NettyChannelBuilder.forAddress(host, port).usePlaintext()
             .maxInboundMessageSize(OzoneConsts.OZONE_SCM_CHUNK_MAX_SIZE);
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
@@ -52,8 +52,16 @@ public class InterSCMGrpcClient implements SCMSnapshotDownloader{
   private final InterSCMProtocolServiceGrpc.InterSCMProtocolServiceStub
       client;
 
-  public InterSCMGrpcClient(final String host, final int port,
-      final long timeout) {
+  public InterSCMGrpcClient(final String host, final int leaderPort,
+      final ConfigurationSource conf) {
+    // if the leader grpc port details are not setup in the peer Map,
+    // fall back to default grpc port.
+    final int port = leaderPort == 0 ?
+        conf.getInt(ScmConfigKeys.OZONE_SCM_GRPC_PORT_KEY,
+            ScmConfigKeys.OZONE_SCM_GRPC_PORT_DEFAULT) :
+        leaderPort;
+    final long  timeout =
+        conf.getObject(SCMHAConfiguration.class).getGrpcDeadlineInterval();
     NettyChannelBuilder channelBuilder =
         NettyChannelBuilder.forAddress(host, port).usePlaintext()
             .maxInboundMessageSize(OzoneConsts.OZONE_SCM_CHUNK_MAX_SIZE);
@@ -89,7 +97,7 @@ public class InterSCMGrpcClient implements SCMSnapshotDownloader{
   }
 
   @Override
-  public void close() throws Exception {
+  public void close() {
     shutdown();
   }
 
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
@@ -22,7 +22,6 @@ import org.apache.hadoop.hdds.conf.ConfigurationSource;
 import org.apache.hadoop.hdds.protocol.scm.proto.InterSCMProtocolProtos;
 import org.apache.hadoop.hdds.protocol.scm.proto.InterSCMProtocolProtos.CopyDBCheckpointResponseProto;
 import org.apache.hadoop.hdds.protocol.scm.proto.InterSCMProtocolServiceGrpc;
-import org.apache.hadoop.hdds.scm.ScmConfigKeys;
 import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.ratis.thirdparty.io.grpc.ManagedChannel;
 import org.apache.ratis.thirdparty.io.grpc.netty.NettyChannelBuilder;
@@ -54,12 +53,7 @@ public class InterSCMGrpcClient implements SCMSnapshotDownloader{
 
   public InterSCMGrpcClient(final String host, final int leaderPort,
       final ConfigurationSource conf) {
-    // if the leader grpc port details are not setup in the peer Map,
-    // fall back to default grpc port.
-    final int port = leaderPort == 0 ?
-        conf.getInt(ScmConfigKeys.OZONE_SCM_GRPC_PORT_KEY,
-            ScmConfigKeys.OZONE_SCM_GRPC_PORT_DEFAULT) :
-        leaderPort;
+    final int port = leaderPort;
     final long  timeout =
         conf.getObject(SCMHAConfiguration.class).getGrpcDeadlineInterval();
     NettyChannelBuilder channelBuilder =
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMSnapshotProvider.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMSnapshotProvider.java
@@ -54,21 +54,14 @@ public class SCMSnapshotProvider {
   private final File scmSnapshotDir;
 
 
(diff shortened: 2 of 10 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -185,10 +186,16 @@ public void notifyNotLeader(Collection<TransactionContext> pendingEntries) {
   public CompletableFuture<TermIndex> notifyInstallSnapshotFromLeader(
       RaftProtos.RoleInfoProto roleInfoProto, TermIndex firstTermIndexInLog) {
 
-    String leaderNodeId = RaftPeerId.valueOf(roleInfoProto.getFollowerInfo()
-        .getLeaderInfo().getId().getId()).toString();
+    String leaderNode = roleInfoProto.getFollowerInfo()
```
bharatviswa504, line 189: In tests I see leaderNode coming ""
Received install snapshot notification from SCM leader:  with term index: (t:2, i:107)

GlenGeng:
Thanks @bshashikant. LGTM. 
BTW, does the ratis version in master contains the fix RATIS-1326 ?

bshashikant:
> Thanks @bshashikant. LGTM.
> BTW, does the ratis version in master contains the fix RATIS-1326 ?

This needs https://issues.apache.org/jira/browse/RATIS-1356 as well. Once this gets resolved, will update the ratis version.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
bshashikant: Addressed with latest ratis version bump

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMSnapshotProvider.java
```diff
@@ -103,18 +101,19 @@ public DBCheckpoint getSCMDBSnapshot(String leaderSCMNodeID)
...
     File targetFile = new File(snapshotFilePath + ".tar.gz");
 
-    // the client instance will be initialized only when first install snapshot
-    // notification from ratis leader will be received.
-    if (client == null) {
-      client = new InterSCMGrpcClient(
-          peerNodesMap.get(leaderSCMNodeID).getInetAddress().getHostAddress(),
-          conf);
-    }
+    // the downloadClient instance will be craeted as and when install snapshot
```
bharatviswa504, line 104: Minor: craeted -> created

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
```diff
@@ -52,13 +52,15 @@
...
-  public InterSCMGrpcClient(final String host, final ConfigurationSource conf) {
-    Preconditions.checkNotNull(conf);
-    int port = conf.getInt(ScmConfigKeys.OZONE_SCM_GRPC_PORT_KEY,
-        ScmConfigKeys.OZONE_SCM_GRPC_PORT_DEFAULT);
-    timeout =
+  public InterSCMGrpcClient(final String host, final int leaderPort,
+      final ConfigurationSource conf) {
+    // if the leader grpc port details are not setup in the peer Map,
+    // fall back to default grpc port.
+    final int port = leaderPort == 0 ?
```
bharatviswa504, line 59: Do we need this, because grpcPort is set in SCMNodeDetails during build?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -184,11 +186,20 @@ public void notifyNotLeader(Collection<TransactionContext> pendingEntries) {
...
+      return JavaUtils.completeExceptionally(new IOException("Failed " +
+          "notifyInstallSnapshotFromLeader due to missing leader info"));
+    }
+    String leaderAddress = roleInfoProto.getFollowerInfo()
+        .getLeaderInfo().getId().getAddress();
+    Optional<SCMNodeDetails> leaderDetails =
+        scm.getSCMHANodeDetails().getPeerNodeDetails().stream().filter(
+            p -> p.getRatisHostPortStr().equals(leaderAddress))
+            .findFirst();
+    Preconditions.checkNotNull(leaderDetails);
```
bharatviswa504, line 199: Minor: Here we need to check for checkState(leaderDetails.isPresent())
As findFirst returns Optional.empty for empty stream.

bharatviswa504 (commented):
Overall LGTM.
Few minor comments. Now I see leader address in logs.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -184,11 +186,20 @@ public void notifyNotLeader(Collection<TransactionContext> pendingEntries) {
   @Override
   public CompletableFuture<TermIndex> notifyInstallSnapshotFromLeader(
       RaftProtos.RoleInfoProto roleInfoProto, TermIndex firstTermIndexInLog) {
-
-    String leaderNodeId = RaftPeerId.valueOf(roleInfoProto.getFollowerInfo()
-        .getLeaderInfo().getId().getId()).toString();
+    if (!roleInfoProto.getFollowerInfo().hasLeaderInfo()) {
+      return JavaUtils.completeExceptionally(new IOException("Failed " +
```
bharatviswa504, line 190: Minor: Failed -> Failed to

pom.xml
```diff
@@ -79,7 +79,7 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <declared.ozone.version>${ozone.version}</declared.ozone.version>
 
     <!-- Apache Ratis version -->
-    <ratis.version>2.0.0</ratis.version>
+    <ratis.version>2.1.0-80f6e4b-SNAPSHOT</ratis.version>
```
GlenGeng, line 82: Thanks @bshashikant and @bharatviswa504 for the effect in RATIS-1326 and RATIS-1356.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
bshashikant: This is required for tests. With Multiple SCM setups, grpc ports for individual SCMs will be different. But, grpc client needs to connect leader grpc port which it does not know from the leaderInfo from Ratis. In actual production clusters, all grpc servers on all nodes will bind to the same grpc port which is not possible in MiniOzoneHAcluster.

bharatviswa504: If each SCM is having different Grpc port, it can get leader grpc port from nodeDetails Right?
If we have already set grpc port in tests.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
bshashikant: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
bshashikant: Done

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMSnapshotProvider.java
```diff
@@ -103,18 +101,19 @@ public DBCheckpoint getSCMDBSnapshot(String leaderSCMNodeID)
...
+    // request is received. No caching of the client as it should be a very rare
+    int port = peerNodesMap.get(leaderSCMNodeID).getGrpcPort();
+    SCMSnapshotDownloader downloadClient = new InterSCMGrpcClient(
+        peerNodesMap.get(leaderSCMNodeID).getInetAddress().getHostAddress(),
+        port, conf);
     try {
-      client.download(targetFile.toPath()).get();
-    } catch (InterruptedException | ExecutionException e) {
+      downloadClient.download(targetFile.toPath()).get();
```
GlenGeng, line 111: Use try with resource here ?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/InterSCMGrpcClient.java
bshashikant: fixed the tests. Thanks.

GlenGeng: The port here follows the HA style config, since it is in `OZONE_SCM_GRPC_PORT_KEY` of `nodeSpecificConfigKeys`, thus it can be handled in MiniOzoneCluster, right ?

bharatviswa504 (approved):
+1 LGTM

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMSnapshotProvider.java
bshashikant: For now, SCMSnapshotDownloader does not implement the AutoCloseable interface. Will take it up in refactoring in a subsequent patch.

bshashikant:
Thanks @bharatviswa504 / @GlenGeng for the review.  Committing this.

elek:
Hi, this patch modified Ozone to use a -SNAPSHOT dependency. Can we avoid using it? Is it absolutely necessary?

I would prefer to depend on released Ratis artifacts only as our last release was delayed with 4 months when we were blocked by the Ratis release.

Can we have a ratis 2.0.1 and depend on it?

bshashikant:
> Hi, this patch modified Ozone to use a -SNAPSHOT dependency. Can we avoid using it? Is it absolutely necessary?
> 
> I would prefer to depend on released Ratis artifacts only as our last release was delayed with 4 months when we were blocked by the Ratis release.
> 
> Can we have a ratis 2.0.1 and depend on it?

This is necessary as only the snapshot has the required fixes to make snapshot mechanism work for SCM HA.

elek:
> This is necessary as only the snapshot has the required fixes to make snapshot mechanism work for SCM HA.

Thanks for the answer. Still I would prefer to depend only on released artifacts from other projects (which is suggested originally by Nanda) I would ask for other's opinion on the community meeting.

Can you please help me with the required commit from Ratis side? Is it apache/ratis#460? If it's just one commit I will start a ratis discussion to release 2.0.1, and we can switch back to a released version...

PR #2157, opened 2021-04-15
HDDS-5107. Add trace for containerReport

JacksonYao287:

Add trace for containerReport to help debugging SCM OOM

HDDS-5107

## How was this patch tested?

no need to test

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerReportHandler.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerReportHandler.java
@@ -35,6 +35,7 @@ import org.apache.hadoop.hdds.scm.server.SCMDatanodeHeartbeatDispatcher
 import org.apache.hadoop.hdds.server.events.EventHandler;
 import org.apache.hadoop.hdds.server.events.EventPublisher;
 import org.apache.hadoop.ozone.common.statemachine.InvalidStateTransitionException;
+import org.apache.hadoop.util.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerReportHandler.java
```diff
@@ -118,6 +121,9 @@ public void onMessage(final ContainerReportFromDatanode reportFromDatanode,
     try {
       final List<ContainerReplicaProto> replicas =
           containerReport.getReportsList();
+      LOG.trace("begin processing ContainerReport of {} at {}," +
```
GlenGeng, line 124: Please use `LOG.info` instead.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerReportHandler.java
```diff
@@ -144,7 +150,9 @@ public void onMessage(final ContainerReportFromDatanode reportFromDatanode,
       LOG.error("Received container report from unknown datanode {}.",
           datanodeDetails, ex);
     }
-
+    LOG.trace("complete processing ContainerReport of {} at {}" +
+        "and processedContainerReportCount is {}", datanodeDetails.getUuid(),
+        Time.now(), processedContaienReportCount.addAndGet(1L));
```
GlenGeng, line 155: Please calculate the latency and add to the `LOF.info` as well.

GlenGeng (commented):
Thanks for the contribute. Some inline commens.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/ContainerReportHandler.java
JacksonYao287: ok, and a process latency is added too

PR #2158, opened 2021-04-16
HDDS-5089. On-demand disk checker for hdds volume

guihecheng:

Check hdds volume on io exceptions to detect bad disks quickly.
For now ozone only has a periodic disk checker which only check for disk errors at fixed interval.

https://issues.apache.org/jira/browse/HDDS-5089

## How was this patch tested?

manual tests
a new integration test

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/HddsVolumeUtil.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/HddsVolumeUtil.java
@@ -24,6 +24,8 @@ import org.apache.hadoop.ozone.OzoneConsts;
 import org.apache.hadoop.ozone.common.InconsistentStorageStateException;
 import org.apache.hadoop.ozone.container.common.DataNodeLayoutVersion;
 import org.apache.hadoop.ozone.container.common.volume.HddsVolume;
+import org.apache.hadoop.ozone.container.common.volume.MutableVolumeSet;
+import org.apache.hadoop.ozone.container.common.volume.VolumeSet;
 import org.apache.hadoop.util.ExitUtil;
 import org.apache.hadoop.util.Time;
 import org.slf4j.Logger;
@@ -233,5 +235,14 @@ public final class HddsVolumeUtil {
       return false;
     }
 
+  }
+
+  public static void onFailure(HddsVolume volume) {
+    if (volume != null) {
+      VolumeSet volumeSet = volume.getVolumeSet();
+      if (volumeSet != null && volumeSet instanceof MutableVolumeSet) {
+        ((MutableVolumeSet) volumeSet).checkVolumeAsync(volume);
+      }
+    }
   }
 }
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/HddsVolume.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/HddsVolume.java
@@ -78,6 +78,7 @@ public class HddsVolume
   private final VolumeInfo volumeInfo;
   private VolumeState state;
   private final VolumeIOStats volumeIOStats;
+  private final VolumeSet volumeSet;
 
   // VERSION file properties
   private String storageID;       // id of the file system
@@ -115,6 +116,7 @@ public class HddsVolume
     private String clusterID;
     private boolean failedVolume = false;
     private SpaceUsageCheckFactory usageCheckFactory;
+    private VolumeSet volumeSet;
 
     public Builder(String rootDirStr) {
       this.volumeRootStr = rootDirStr;
@@ -153,6 +155,11 @@ public class HddsVolume
       return this;
     }
 
+    public Builder volumeSet(VolumeSet volSet) {
+      this.volumeSet = volSet;
+      return this;
+    }
+
     public HddsVolume build() throws IOException {
       return new HddsVolume(this);
     }
@@ -172,6 +179,7 @@ public class HddsVolume
           .usageCheckFactory(b.usageCheckFactory)
           .build();
       this.committedBytes = new AtomicLong(0);
+      this.volumeSet = b.volumeSet;
 
       LOG.info("Creating Volume: {} of storage type : {} and capacity : {}",
           hddsRootDir, b.storageType, volumeInfo.getCapacity());
@@ -186,6 +194,7 @@ public class HddsVolume
       storageID = UUID.randomUUID().toString();
       state = VolumeState.FAILED;
       committedBytes = null;
+      volumeSet = null;
     }
   }
 
@@ -393,6 +402,10 @@ public class HddsVolume
 
   public VolumeIOStats getVolumeIOStats() {
     return volumeIOStats;
+  }
+
+  public VolumeSet getVolumeSet() {
+    return volumeSet;
   }
 
   public void failVolume() {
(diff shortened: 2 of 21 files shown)


bshashikant:
Thanks @guihecheng for the work. The changes look ok to me so far, though still going through the patch. Meanwhile, can you rebase the patch to the current master to see if the test failures get resolved?

guihecheng:
@bshashikant thanks for your comments, I believe that the following pr fixes the unit test failure case:
https://github.com/apache/ozone/pull/2159
I'll rebase it.

guihecheng:
@bshashikant updated thanks~

guihecheng:
Hi @bshashikant , I'm reviewing this patch myself, here is a summary:
# check points:
- block/chunk file file operation: covered
- container file operation: covered
- rocksdb operation: only open, maybe flush, compact, etc
- ratis opertation: not covered
# tests:
- unit-test / integration-test: hdfs has some utils that helps test volume/disk failures, we shall import them to help test
e.g. 
https://github.com/apache/hadoop/blob/6cb90005a7d0651474883ac4e1b6961ef74fe513/hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/server/datanode/DataNodeTestUtils.java#L149-L150

bshashikant:
Thanks @guihecheng . The patch looks ok. XceiverServerRatis#handleNodeLogFailure() is actually a raft log disk failure. It should be handled in similar fashion as well. Similarly, KeyValueContainer#flushAndSyncDB() failure also points to a disk gone bad , probably needs a similar handling.

It would be great, if u can add a unit or acceptance test to validate this.

guihecheng:
> Thanks @guihecheng . The patch looks ok. XceiverServerRatis#handleNodeLogFailure() is actually a raft log disk failure. It should be handled in similar fashion as well. Similarly, KeyValueContainer#flushAndSyncDB() failure also points to a disk gone bad , probably needs a similar handling.

@bshashikant Thanks for the hints, that is really helpful, I'll check them, maybe the most possible challenge is to get the container/volume context in these places.

> It would be great, if u can add a unit or acceptance test to validate this.

guihecheng:
@bshashikant updated with a integration test.
IMHO, #handleNodeLogFailure already handles ratis volume failure by closing pipelines, the failure is detected by ratis itself, and it is not a HddsVolume, so we don't touch it.
cc @ChenSammi

bshashikant:
> @bshashikant updated with a integration test.
> IMHO, #handleNodeLogFailure already handles ratis volume failure by closing pipelines, the failure is detected by ratis itself, and it is not a HddsVolume, so we don't touch it.
> cc @ChenSammi

Thanks @guihecheng . I think, closing the pipeline on a bad raft log disk will not go through as it needs to take a final snapshot and then remove the pipeline dir. What we also, need to ensure is ,  mark these disks are bad and remove it from ratis storage dir list, so that no new pipeline creation can happen on these disks. I am not to sure if ratis alredy handles it, but would be good to have a test so that, no new pipeline allocation happens using the bad disks.

guihecheng:
@bshashikant Thanks, I will check the ratis log disk handling soon.

guihecheng:
@bshashikant As I tested manually in my local boxes, I hit the following msg below, so I think that ratis is not able to manage bad disks, since it is still try to use a bad disk to create a pipeline.

As I know, ozone only passes `dfs.container.ratis.datanode.storage.dir` to ratis, and let ratis manage the directories itself(although we have a `ratisVolumeMap`, it is only used to build storage reports).
So I think we should directly add disk checker support to ratis instead of handling it here in ozone.

```
2021-05-06 18:19:48,166 [Command processor thread] ERROR org.apache.hadoop.ozone.container.common.statemachine.commandhandler.CreatePipelineCommandHandler: Can't create pipeline RATIS THREE PipelineID=08de41a6-5c9e-48d4-9789-4c09798ecffd
java.io.IOException: Input/output error
        at org.apache.hadoop.ozone.container.common.transport.server.ratis.XceiverServerRatis.addGroup(XceiverServerRatis.java:805)
        at org.apache.hadoop.ozone.container.common.statemachine.commandhandler.CreatePipelineCommandHandler.handle(CreatePipelineCommandHandler.java:92)
        at org.apache.hadoop.ozone.container.common.statemachine.commandhandler.CommandDispatcher.handle(CommandDispatcher.java:99)
        at org.apache.hadoop.ozone.container.common.statemachine.DatanodeStateMachine.lambda$initCommandHandlerThread$2(DatanodeStateMachine.java:506)
Caused by: java.io.IOException: Input/output error
        at java.io.UnixFileSystem.canonicalize0(Native Method)
        at java.io.UnixFileSystem.canonicalize(UnixFileSystem.java:172)
        at java.io.File.getCanonicalPath(File.java:620)
        at org.apache.ratis.server.storage.RaftStorageDirectoryImpl.analyzeStorage(RaftStorageDirectoryImpl.java:129)
        at org.apache.ratis.server.storage.RaftStorageImpl.analyzeAndRecoverStorage(RaftStorageImpl.java:95)
        at org.apache.ratis.server.storage.RaftStorageImpl.<init>(RaftStorageImpl.java:65)
        at org.apache.ratis.server.storage.RaftStorageImpl.<init>(RaftStorageImpl.java:51)
        at org.apache.ratis.server.impl.ServerState.<init>(ServerState.java:112)
        at org.apache.ratis.server.impl.RaftServerImpl.<init>(RaftServerImpl.java:193)
        at org.apache.ratis.server.impl.RaftServerProxy.lambda$newRaftServerImpl$4(RaftServerProxy.java:266)
        at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1604)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        ... 1 more
```

guihecheng:
@bshashikant As I found in ratis code, ratis tries for every storageDir passed in from ozone:
https://github.com/apache/ratis/blob/040bc52e19a5e36f5710ccd4fc1981e862e691e8/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java#L107-L117

Of course, ratis choose a dir based on some criteria:
https://github.com/apache/ratis/blob/040bc52e19a5e36f5710ccd4fc1981e862e691e8/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java#L149
It chooses the dir with least number of raftgroups for a new raftgroup.
For a bad volume, it still could be chosen by ratis, and later throws an exception in handling the directories just as the previous message shows, and ratis does not catch the IOException and throws out.
The simplest way should be let ratis catch IOException instead of only AccessDeniedException, remove the bad volume and try on other directories.

bshashikant:
> @bshashikant As I found in ratis code, ratis tries for every storageDir passed in from ozone:
> https://github.com/apache/ratis/blob/040bc52e19a5e36f5710ccd4fc1981e862e691e8/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java#L107-L117
> 
> Of course, ratis choose a dir based on some criteria:
> https://github.com/apache/ratis/blob/040bc52e19a5e36f5710ccd4fc1981e862e691e8/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java#L149
> It chooses the dir with least number of raftgroups for a new raftgroup.
> For a bad volume, it still could be chosen by ratis, and later throws an exception in handling the directories just as the previous message shows, and ratis does not catch the IOException and throws out.
> The simplest way should be let ratis catch IOException instead of only AccessDeniedException, remove the bad volume and try on other directories.

yes, Ratis we can change the code. Also, in ozone, before starting the ratis server, the disk checker can be be run on ratis volumes to see if its good right upfront in XceiverServerRatis as well. I am fine with either approaches.

bshashikant:
Thanks @guihecheng . Let'd do the ratis disk handling in a new pr. Thanks for the work.

bshashikant (approved):

PR #2165, opened 2021-04-20
HDDS-5123. Use the pre-created apache/ozone-testkrb5 image during secure acceptance tests

elek:

Today ozonesecure compose clusters (and ozonesecure-ha and ozonesecure-mr) use an adhoc keytab issuer. The issuer is download during the image creation and uses a third party go lang application to create the keytabs on-demand.

As discussed earlier, it would be faster to use a dedicated, pre-built container image which includes the pre-created keytabs instead of issuing them on-the fly (keytab generation is slow + container creation is slow)

For each of the tagged images we can export to current keytabs to hadoop-ozone/dist/src/main/compose/ which can be mounted to compose clusters.

It makes the overall acceptance test faster (instead of creating keytab, which is quite slow, we can start the cluster immediately). And we don't need to depend on an external utility app.

Pre-created keytabs are also more similar to production environment...

https://issues.apache.org/jira/browse/HDDS-5123

## How was this patch tested?

CI is passed on my fork: 
https://github.com/elek/ozone/actions/runs/763590947

The time between starting test.sh script and first robot test:

master: 3:30 (01:43:08 --01:46:38)
this patch: 2:10 (12:59:29 13:02:39)

(note: there are some variances between different builds, and in general the mentioned patch build was a slower one. It can be even faster).


hadoop-ozone/dist/src/main/smoketest/commonlib.robot
```diff
@@ -32,6 +32,6 @@ Kinit HTTP user
 
 Kinit test user
     [arguments]                      ${user}       ${keytab}
-    ${hostname} =       Execute                    hostname
+    ${hostname} =       Execute                    hostname | sed 's/.org//' | sed 's/[0-9]//'
```
adoroszlai, line 35: I think ACLs like `user:testuser2/${SCM}@EXAMPLE.COM` need to be tweaked, too.  With SCM HA they apply to `testuser2/scm1`, but this `sed` makes the test `kinit` with `testuser2/scm`.

This causes test failure:

```
Can follow link with read access                                      | FAIL |
'PERMISSION_DENIED User testuser2/scm@EXAMPLE.COM doesn't have READ permission to access volume 84112-target null null' does not contain 'key-in-readable-bucket'
```

https://github.com/apache/ozone/runs/2400748407#step:6:1636

adoroszlai (commented):
Thanks @elek for working on this improvement.

hadoop-ozone/dist/src/main/smoketest/commonlib.robot
adoroszlai: Thanks @elek for addressing my other comments.  I think this one still needs to be done.

```diff
--- hadoop-ozone/dist/src/main/smoketest/basic/links.robot
+++ hadoop-ozone/dist/src/main/smoketest/basic/links.robot
@@ -42,11 +42,11 @@
     Execute             ozone sh bucket link ${source}/readable-bucket ${target}/readable-link
     Execute             ozone sh bucket link ${source}/readable-bucket ${target}/unreadable-link
     Execute             ozone sh bucket link ${source}/unreadable-bucket ${target}/link-to-unreadable-bucket
-    Execute             ozone sh volume addacl --acl user:testuser2/${SCM}@EXAMPLE.COM:r ${target}
-    Execute             ozone sh volume addacl --acl user:testuser2/${SCM}@EXAMPLE.COM:rl ${source}
-    Execute             ozone sh bucket addacl --acl user:testuser2/${SCM}@EXAMPLE.COM:rl ${source}/readable-bucket
-    Execute             ozone sh bucket addacl --acl user:testuser2/${SCM}@EXAMPLE.COM:r ${target}/readable-link
-    Execute             ozone sh bucket addacl --acl user:testuser2/${SCM}@EXAMPLE.COM:r ${target}/link-to-unreadable-bucket
+    Execute             ozone sh volume addacl --acl user:testuser2/scm@EXAMPLE.COM:r ${target}
+    Execute             ozone sh volume addacl --acl user:testuser2/scm@EXAMPLE.COM:rl ${source}
+    Execute             ozone sh bucket addacl --acl user:testuser2/scm@EXAMPLE.COM:rl ${source}/readable-bucket
+    Execute             ozone sh bucket addacl --acl user:testuser2/scm@EXAMPLE.COM:r ${target}/readable-link
+    Execute             ozone sh bucket addacl --acl user:testuser2/scm@EXAMPLE.COM:r ${target}/link-to-unreadable-bucket

 Can follow link with read access
     Execute             kdestroy
```

adoroszlai (approved):

elek:
The build is green, merging it now. Thanks, @adoroszlai the review.

PR #2166, opened 2021-04-20
HDDS-5124. Use OzoneConsts.OZONE_TIME_ZONE instead of "GMT"

wzhallright:
jira: HDDS-5124. Use existing constants for org.apache.hadoop.ozone.s3.endpoint.ObjectEndpoint#addLastModifiedDate
https://issues.apache.org/jira/browse/HDDS-5124?filter=-2

The "GMT" already defined in OzoneConsts, so why not user it.
we can use Ozone.Consts.OZONE_TIME_ZONE instead of "GMT"


ayushtkn (commented):
It is there in ``RFC1123Util``, Give a check if you could use this there as well.
https://github.com/apache/ozone/blob/master/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/util/RFC1123Util.java#L95

wzhallright:
> It is there in `RFC1123Util`, Give a check if you could use this there as well.
> https://github.com/apache/ozone/blob/master/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/util/RFC1123Util.java#L95

thanks for review @ayushtkn.
Of course it can be used there, i will fix it.

ayushtkn (approved):
Thanx, Changes LGTM.
Can you change the PR title and the jira title to be a little generic,
Something like, Use OzoneConsts.OZONE_TIME_ZONE insted of "GMT"

PR #2167, opened 2021-04-20
HDDS-5122. SCM Reinitialization can end up leaking Ratis Segmented RaftLogWorker threads

bshashikant:

Fixed SCM reinitialization logic 

https://issues.apache.org/jira/browse/HDDS-5122

## How was this patch tested?

Modified UTs

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAUtils.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAUtils.java
@@ -74,10 +74,10 @@ public final class SCMHAUtils {
   }
 
   public static boolean isPrimordialSCM(ConfigurationSource conf,
-      String selfNodeId) {
+      String selfNodeId, String hostname) {
     String primordialNode = getPrimordialSCM(conf);
-    return isSCMHAEnabled(conf) && primordialNode != null && primordialNode
-        .equals(selfNodeId);
+    return isSCMHAEnabled(conf) && primordialNode != null && (primordialNode
+        .equals(selfNodeId) || primordialNode.equals(hostname));
   }
   /**
    * Get a collection of all scmNodeIds for the given scmServiceId.
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAUtils.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAUtils.java
@@ -74,10 +74,10 @@ public final class SCMHAUtils {
   }
 
   public static boolean isPrimordialSCM(ConfigurationSource conf,
-      String selfNodeId, String hostname) {
+      String selfNodeId, String hostName) {
     String primordialNode = getPrimordialSCM(conf);
     return isSCMHAEnabled(conf) && primordialNode != null && (primordialNode
-        .equals(selfNodeId) || primordialNode.equals(hostname));
+        .equals(selfNodeId) || primordialNode.equals(hostName));
   }
   /**
    * Get a collection of all scmNodeIds for the given scmServiceId.
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
@@ -251,6 +251,19 @@ public abstract class Storage {
   }
 
   /**
+   * Creates the Version file even if it exits,
+   * otherwise returns with IOException.
+   * @throws IOException
+   */
+  public void forceInitialize() throws IOException {
+    if (state != StorageState.INITIALIZED) {
+      initialize();
+    } else {
+      storageInfo.writeTo(getVersionFile());
+    }
+  }
+
+  /**
    * Persists current StorageInfo to file system..
    * @throws IOException
    */
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
@@ -251,7 +251,7 @@ public abstract class Storage {
   }
 
   /**
-   * Creates the Version file even if it exists
+   * Creates the Version file even if it exists.
    * @throws IOException
    */
   public void forceInitialize() throws IOException {
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
@@ -251,8 +251,7 @@ public abstract class Storage {
   }
 
   /**
-   * Creates the Version file even if it exits,
-   * otherwise returns with IOException.
+   * Creates the Version file even if it exists
    * @throws IOException
    */
   public void forceInitialize() throws IOException {
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMStorageConfig.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMStorageConfig.java
@@ -28,8 +28,11 @@ import java.io.IOException;
 import java.util.Properties;
 import java.util.UUID;
 
-import static org.apache.hadoop.ozone.OzoneConsts.*;
-
+import static org.apache.hadoop.ozone.OzoneConsts.SCM_HA;
+import static org.apache.hadoop.ozone.OzoneConsts.STORAGE_DIR;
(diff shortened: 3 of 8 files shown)


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -937,13 +943,20 @@ public static boolean scmInit(OzoneConfiguration conf,
...
-      if (SCMHAUtils.isSCMHAEnabled(conf)) {
-        SCMRatisServerImpl.reinitialize(clusterId, scmStorageConfig.getScmId(),
-            haDetails.getLocalNodeDetails(), conf);
+      final boolean isSCMHAEnabled = scmStorageConfig.isSCMHAEnabled();
+      if (SCMHAUtils.isSCMHAEnabled(conf) && !isSCMHAEnabled) {
+        SCMRatisServerImpl.initialize(scmStorageConfig.getClusterID(),
+            scmStorageConfig.getScmId(), haDetails.getLocalNodeDetails(),
+            conf);
+        scmStorageConfig.setSCMHAFlag(true);
+        scmStorageConfig.setPrimaryScmNodeId(scmStorageConfig.getScmId());
```
bharatviswa504, line 952: We set primaryScmID irrespective of HA scmStorageConfig.setPrimaryScmNodeId(scmStorageConfig.getScmId());

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
```diff
@@ -250,6 +250,19 @@ public void initialize() throws IOException {
     storageInfo.writeTo(getVersionFile());
   }
 
+  /**
+   * Creates the Version file even if it exits,
```
bharatviswa504, line 254: exits -> exists

And remove "otherwise returns with IOException."

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMStorageConfig.java
```diff
@@ -27,11 +28,7 @@
 import java.util.Properties;
 import java.util.UUID;
 
-import static org.apache.hadoop.ozone.OzoneConsts.PRIMARY_SCM_NODE_ID;
-import static org.apache.hadoop.ozone.OzoneConsts.SCM_CERT_SERIAL_ID;
-import static org.apache.hadoop.ozone.OzoneConsts.SCM_ID;
-import static org.apache.hadoop.ozone.OzoneConsts.STORAGE_DIR;
-
+import static org.apache.hadoop.ozone.OzoneConsts.*;
```
bharatviswa504, line 31: Unnecessary, expand *

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMStorageConfig.java
```diff
@@ -60,6 +57,13 @@ public void setScmId(String scmId) throws IOException {
     }
   }
 
+  public void setSCMHAFlag(boolean flag) {
+    if (!isSCMHAEnabled()) {
+      Preconditions.checkNotNull(flag);
```
bharatviswa504, line 62: No need of null check here.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -871,6 +874,7 @@ public static boolean scmBootstrap(OzoneConfiguration conf)
               getScmAddress(scmhaNodeDetails, conf), false);
         }
         scmStorageConfig.setPrimaryScmNodeId(scmInfo.getScmId());
+        scmStorageConfig.setSCMHAFlag(true);
```
bharatviswa504, line 877: Do we need to set this flag only when HA enabled to avoid user mistakes of running bootStrap on non-HA cluster?

bshashikant: BootStrap will fail if HA is not enabled right upfront. It won't execute this code at all.

bharatviswa504: Yes.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMStorageConfig.java
```diff
@@ -60,6 +57,13 @@ public void setScmId(String scmId) throws IOException {
     }
   }
 
+  public void setSCMHAFlag(boolean flag) {
```
GlenGeng, line 60: NIT The check of `isSCMHAEnabled` here is confusing. Boolean can not be null.
```
  public void setSCMHAFlag(boolean flag) {
    getStorageInfo().setProperty(SCM_HA, Boolean.toString(flag));
  }
```

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -937,13 +943,20 @@ public static boolean scmInit(OzoneConfiguration conf,
...
       clusterId = scmStorageConfig.getClusterID();
-      LOG.info("SCM already initialized. Reusing existing cluster id for sd={}"
-              + ";cid={};layoutVersion={}", scmStorageConfig.getStorageDir(),
-          clusterId, scmStorageConfig.getLayoutVersion());
-      if (SCMHAUtils.isSCMHAEnabled(conf)) {
-        SCMRatisServerImpl.reinitialize(clusterId, scmStorageConfig.getScmId(),
-            haDetails.getLocalNodeDetails(), conf);
+      final boolean isSCMHAEnabled = scmStorageConfig.isSCMHAEnabled();
+      if (SCMHAUtils.isSCMHAEnabled(conf) && !isSCMHAEnabled) {
+        SCMRatisServerImpl.initialize(scmStorageConfig.getClusterID(),
```
GlenGeng, line 948: NIT: Add an info trace to avoid the silent upgrade ?

GlenGeng (commented):
LGTM

hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/common/Storage.java
```diff
@@ -250,6 +250,19 @@ public void initialize() throws IOException {
...
+   * otherwise returns with IOException.
+   * @throws IOException
+   */
+  public void forceInitialize() throws IOException {
+    if (state != StorageState.INITIALIZED) {
+      initialize();
```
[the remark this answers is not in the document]
bshashikant: I think this is fine. State will be initialised only if the version file exists which implies that the current directory exists.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/StorageContainerManager.java
```diff
@@ -943,14 +943,18 @@ public static boolean scmInit(OzoneConfiguration conf,
       }
     } else {
       clusterId = scmStorageConfig.getClusterID();
+      if (!scmStorageConfig.checkPrimarySCMIdInitialized()) {
```
bharatviswa504, line 946: I am sorry for now we should setPrimaryScmNodeId only when HA, otherwise it will affect upgraded secure cluster.
As we assume that if primaryScmNodeId is set it means it is the cluster which has done init with new code.

bharatviswa504: This will be problem on clusters init runs every time during SCM start.

bshashikant: Done

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMHAUtils.java
```diff
@@ -74,10 +74,10 @@ public static String getPrimordialSCM(ConfigurationSource conf) {
   }
 
   public static boolean isPrimordialSCM(ConfigurationSource conf,
-      String selfNodeId) {
+      String selfNodeId, String hostname) {
```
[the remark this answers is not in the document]
bshashikant: It requires moving around classes for now. Will address in different jira.

bharatviswa504 (approved):
+1 LGTM

bshashikant:
Thanks @bharatviswa504 for the review.

PR #2168, opened 2021-04-21
HDDS-5125. Only test ozonesecure with SCM Ratis disabled

adoroszlai:

As [spotted](https://github.com/apache/ozone/pull/2052#discussion_r616570112) by @elek, `ozonesecure` acceptance test is run twice: with SCM Ratis enabled and disabled.  This doubles execution time.  Now that `ozonesecure-ha` acceptance test also has SCM HA (implies Ratis enabled), I don't think it's necessary to run `ozonesecure` (single SCM) with Ratis enabled.

https://issues.apache.org/jira/browse/HDDS-5125

## How was this patch tested?

Now _acceptance (secure)_ is only 1 hour:
https://github.com/adoroszlai/hadoop-ozone/actions/runs/768286122

### DIFF
--- a/hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
+++ b/hadoop-ozone/dist/src/main/compose/ozonesecure/docker-compose.yaml
@@ -104,6 +104,5 @@ services:
       KERBEROS_KEYTABS: scm HTTP testuser testuser2
       ENSURE_SCM_INITIALIZED: /data/metadata/scm/current/VERSION
       OZONE-SITE.XML_hdds.scm.safemode.min.datanode: "${OZONE_SAFEMODE_MIN_DATANODES:-1}"
-      OZONE-SITE.XML_ozone.scm.ratis.enable: "${OZONE_SCM_RATIS_ENABLE:-false}"
       OZONE_OPTS:
     command: ["/opt/hadoop/bin/ozone","scm"]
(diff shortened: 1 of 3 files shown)


elek (commented):
I am +1, but I was the one who complained about it.

Let's have the opinion of @bharatviswa504 here, as he [commented](https://github.com/apache/ozone/pull/2052#discussion_r617203023) this:

> (It would be good to have, but if we think it is expensive or any other way we can run this with ratis enable for single SCM cluster it would be good)

@bharatviswa504 Do you think we need to add anything to here which is not covered elsewhere?

bharatviswa504 (commented):
+1 LGTM.

I agree with description now we run secure-ha with ratis-enable which can give testing for SCM HA Security, but it would also been nice if we have this too as this runs ratis-enable with single node secure SCM. As said I am fine with it.

Question: Do we have some tests to run every day, there we can add this, so that CI resources are not used for this test scenario.

adoroszlai:
> Question: Do we have some tests to run every day, there we can add this, so that CI resources are not used for this test scenario.

We have scheduled builds twice a day, still using the same pool of resources.  Currently tests are exactly the same, but can be extended later.

adoroszlai:
Thanks @bharatviswa504 and @elek for the review.

bharatviswa504:
> > Question: Do we have some tests to run every day, there we can add this, so that CI resources are not used for this test scenario.
> 
> We have scheduled builds twice a day, still using the same pool of resources. Currently tests are exactly the same, but can be extended later.

Thanks @adoroszlai for the info, I have opened a jira https://issues.apache.org/jira/browse/HDDS-5134 for this.

PR #2169, opened 2021-04-21
HDDS-5128. Return specific error messages for S3-SDK

symious:

- Return WebApplicationException instead of OS3Exception in OzoneClientProducer.

https://issues.apache.org/jira/browse/HDDS-5128

## How was this patch tested?

Test passed with S3-SDK requests. Can test with the following Steps:
1. Remove credentials. 
```
mv ~/.aws ~/.aws_bak
unset AWS_ACCESS_KEY_ID
unset AWS_SECRET_ACCESS_KEY
```
2. List objects using S3 SDK via S3gateway.
```
AmazonS3 s3 = AmazonS3ClientBuilder.standard()
        .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration("http://localhost:9878", Regions.DEFAULT_REGION.getName()))
        .withPathStyleAccessEnabled(true)
        .build();
ListObjectsV2Result result = s3.listObjectsV2(bucket_name);
```
3. Before the patch the error message is as follows:
```
Server Error (Service: Amazon S3; Status Code: 500; Error Code: 500 Server Error; Request ID: null; S3 Extended Request ID: null; Proxy: null)
```
4. After the patch the error would be as follows:
```
The authorization header you provided is invalid. (Service: Amazon S3; Status Code: 404; Error Code: AuthorizationHeaderMalformed; Request ID: 2102408b-f97b-477f-8510-07e7008792be; S3 Extended Request ID: null; Proxy: null)
```

### DIFF
--- a/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneClientProducer.java
+++ b/hadoop-ozone/s3gateway/src/main/java/org/apache/hadoop/ozone/s3/OzoneClientProducer.java
@@ -21,6 +21,7 @@ import javax.annotation.PreDestroy;
 import javax.enterprise.context.RequestScoped;
 import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.core.Context;
 import java.io.IOException;
@@ -76,7 +77,7 @@ public class OzoneClientProducer {
   private ContainerRequestContext context;
 
   @Produces
-  public OzoneClient createClient() throws OS3Exception, IOException {
+  public OzoneClient createClient() throws WebApplicationException, IOException {
     client = getClient(ozoneConfiguration);
     return client;
   }
@@ -87,7 +88,7 @@ public class OzoneClientProducer {
   }
 
   private OzoneClient getClient(OzoneConfiguration config)
-      throws OS3Exception {
+      throws WebApplicationException {
     OzoneClient ozoneClient = null;
     try {
       SignatureInfo signatureInfo = signatureProcessor.parseSignature();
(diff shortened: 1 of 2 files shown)


elek (commented):
Thanks to create this patch @symious. I am very grateful, I think the current behavior (HTTP 500 in case of any error) is very confusing.

However, I tried to test the patch and still see HTTP 500.


With the example code I got:

```
Exception in thread "main" com.amazonaws.services.s3.model.AmazonS3Exception: Internal Server Error (Service: Amazon S3; Status Code: 500; Error Code: 500 Internal Server Error; Request ID: 07e9b8b3-1084-4722-a8c2-e49bc1f55e15; S3 Extended Request ID: NCWLQ4Zx0uBx; Proxy: null), S3 Extended Request ID: NCWLQ4Zx0uBx
```

With aws cli:

```
aws s3api create-bucket --endpoint http://localhost:9878 --bucket bucket999
  
An error occurred (500) when calling the CreateBucket operation (reached max retries: 4): Internal Server Error
```

Based on the stack trace I have the new code running:

```
s3g_1       | 	at java.base/java.lang.Thread.run(Thread.java:834)
s3g_1       | Apr 27, 2021 8:31:21 AM org.glassfish.jersey.internal.Errors logErrors
s3g_1       | WARNING: The following warnings have been detected: WARNING: Unknown HK2 failure detected:
s3g_1       | MultiException stack 1 of 1
s3g_1       | javax.ws.rs.WebApplicationException: HTTP 500 Internal Server Error
s3g_1       | 	at org.apache.hadoop.ozone.s3.OzoneClientProducer.wrapOS3Exception(OzoneClientProducer.java:182)
s3g_1       | 	at org.apache.hadoop.ozone.s3.OzoneClientProducer.getClient(OzoneClientProducer.java:146)
s3g_1       | 	at org.apache.hadoop.ozone.s3.OzoneClientProducer.createClient(OzoneClientProducer.java:82)
``` 

Do you have any idea where is my mistake?

symious:
For the real internal errors, I don't know if it's proper to return the root cause to end-users.
Say the user is trying to get data of a block, but the OM is having an issue connecting to the SCM, the internal error would be "OM can not connect to SCM", I don't know if the user would be confused about the term, or the user only needs to report to the ozone cluster PIC that an internal error is happening.

symious:
@elek Any update on this ticket? I think an S3gateway instance in MiniOzoneCluster would be useful for this kind of test case, may I ask if there are any second thoughts to not include S3gateway in MiniOzoneCluster?

elek (approved):
Thanks to explain it @symious. I hope HDDS-4440 will fix the full error handling. 

Until that it's good to have this fix to fix this specific case.

Let me merge it...

PR #2171, opened 2021-04-21
HDDS-5059. Increase number of client retries/ failovers to OMs

hanishakoneru:

Currently in an OM HA setup, client by default does 15 failovers before giving up. If there are 3 OMs and a leader has not been elected yet, then the client will try each OM in a round robin way, wait for 2 secs and then try all the OMs again. This gives the client around 10 seconds before it exhausts all its failover attempts and fails. 

This Jira proposes to increase the failover attempts to 100.

https://issues.apache.org/jira/browse/HDDS-5059

## How was this patch tested?

Existing unit tests

### DIFF
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2357,12 +2357,14 @@
   </property>
   <property>
     <name>ozone.client.failover.max.attempts</name>
-    <value>15</value>
+    <value>100</value>
     <description>
       Expert only. Ozone RpcClient attempts talking to each OzoneManager
       ipc.client.connect.max.retries (default = 10) number of times before
       failing over to another OzoneManager, if available. This parameter
-      represents the number of times the client will failover before giving up.
+      represents the number of times the client will failover before giving
+      up. This value is kept high so that client does not give up trying to
+      connect to OMs easily.
     </description>
   </property>
   <property>
--- a/hadoop-hdds/common/src/main/resources/ozone-default.xml
+++ b/hadoop-hdds/common/src/main/resources/ozone-default.xml
@@ -2357,7 +2357,7 @@
   </property>
   <property>
     <name>ozone.client.failover.max.attempts</name>
-    <value>100</value>
+    <value>500</value>
     <description>
       Expert only. Ozone RpcClient attempts talking to each OzoneManager
       ipc.client.connect.max.retries (default = 10) number of times before
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/ozone/OzoneConfigKeys.java
@@ -366,7 +366,7 @@ public final class OzoneConfigKeys {
   public static final String OZONE_CLIENT_FAILOVER_MAX_ATTEMPTS_KEY =
       "ozone.client.failover.max.attempts";
   public static final int OZONE_CLIENT_FAILOVER_MAX_ATTEMPTS_DEFAULT =
-      15;
+      500;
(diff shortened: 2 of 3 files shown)


bharatviswa504 (commented):
+1 LGTM

bharatviswa504:
Just a question, the client will totally get 24 seconds right (as 12 *2)? (As for first 3 OM's we don't wait if they are not leader)

hanishakoneru:
Thanks @bharatviswa504.
> Just a question, the client will totally get 24 seconds right (as 12 *2)? (As for first 3 OM's we don't wait if they are not leader)
We do not wait to failover between OMs (not just first time). So client will try om1, om2, om3, wait 2 secs, then try om1, om2, om3 again without any wait, then wait 2 secs again and this cycle is repeated. 

Changing to 100 also would not give us much difference though. We would get around 1 min before the client gives up. 

What do you think is a reasonable time for the client to give up? Or should the client just retry forever?
cc. @arp7

bharatviswa504:
> Thanks @bharatviswa504.
> 
> > Just a question, the client will totally get 24 seconds right (as 12 *2)? (As for first 3 OM's we don't wait if they are not leader)
> 
> We do not wait to failover between OMs (not just first time). So client will try om1, om2, om3, wait 2 secs, then try om1, om2, om3 again without any wait, then wait 2 secs again and this cycle is repeated.
> 
> Changing to 100 also would not give us much difference though. We would get around 1 min before the client gives up.
> 
> What do you think is a reasonable time for the client to give up? Or should the client just retry forever?
> cc. @arp7

Thanks, @hanishakoneru for the info. We can have a client max retry time out and derive retry count from that? (In SCM, this is what being planned)

hanishakoneru:
> We can have a client max retry time out and derive retry count from that? (In SCM, this is what being planned for obtaining retryCount)

Sounds good. There is an existing `ozone.om.client.rpc.timeout` with default value of 15 mins. I think we can use this value itself as the max retry time.

bharatviswa504:
@hanishakoneru I think the existing config is for a single rpc call max time out, and we use that during proxy object creation. (The above config was introduced, when only one OM is there our RPC clients are stuck for ever, now after 15 mts they give up, and finally failure response is returned)
So, do we want to introduce a new config for maxRetryTimeOut and use that and derive retry count from that?

hanishakoneru:
@bharatviswa504, in case of OM HA, the rpc timeout is for each proxy. If we reuse the same config, then the overall client rpc timeout would be the same for both non-HA and HA cases.

hanishakoneru:
@bharatviswa504, can you please review when you get a chance. In the latest iteration of the patch, I have removed consolidated logging of exceptions encountered during retry attemps. RetryInvocationHandler prints the exception message for each attempt (except the first on each proxy). Hence, I think we don't need to log them again. Thoughts?

hadoop-hdds/common/src/main/resources/ozone-default.xml
```diff
@@ -2357,12 +2357,14 @@
   </property>
   <property>
     <name>ozone.client.failover.max.attempts</name>
-    <value>15</value>
+    <value>100</value>
```
bharatviswa504, line 2360: Here also do we need to change to 500?

bharatviswa504 (commented):
Overall LGTM.

bharatviswa504:
> @bharatviswa504, can you please review when you get a chance. In the latest iteration of the patch, I have removed consolidated logging of exceptions encountered during retry attemps. RetryInvocationHandler prints the exception message for each attempt (except the first on each proxy). Hence, I think we don't need to log them again. Thoughts?

Yes, it makes sense to me.

bharatviswa504 (approved):
+1 LGTM

hanishakoneru:
Thanks @bharatviswa504 for the reviews. I will merge this shortly.

PR #2172, opened 2021-04-22
HDDS-5126. Recon should check new containers of a container report with batch

JacksonYao287:

in my test environment,  400000 containers exist. when bootstrap a new recon, every container will be checked and added to recondb.but , for now , recon check all the containers in a container report one by one , and each check will take a rpc call to scm. this is too slow and in my test environment , it leads to recon oom, because too many containers to be consumed are waitting in the message queue . It is better for recon to check new containers of a container report with batch.

https://issues.apache.org/jira/browse/HDDS-5126

## How was this patch tested?

unit test

### DIFF
--- a/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
+++ b/hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
@@ -48,7 +48,6 @@ import org.apache.hadoop.hdds.scm.pipeline.PipelineID;
 import org.apache.hadoop.hdds.scm.pipeline.PipelineManager;
 import org.apache.hadoop.hdds.utils.db.DBStore;
 import org.apache.hadoop.hdds.utils.db.Table;
-import org.apache.hadoop.ozone.common.statemachine.InvalidStateTransitionException;
 import org.apache.hadoop.ozone.recon.persistence.ContainerHistory;
 import org.apache.hadoop.ozone.recon.persistence.ContainerHealthSchemaManager;
 import org.apache.hadoop.ozone.recon.spi.ContainerDBServiceProvider;
@@ -114,7 +113,7 @@ public class ReconContainerManager extends ContainerManagerImpl {
   public void checkAndAddNewContainer(ContainerID containerID,
       ContainerReplicaProto.State replicaState,
       DatanodeDetails datanodeDetails)
-      throws IOException {
+      throws Exception {
     if (!containerExist(containerID)) {
       LOG.info("New container {} got from {}.", containerID,
           datanodeDetails.getHostName());
@@ -122,27 +121,89 @@ public class ReconContainerManager extends ContainerManagerImpl {
           scmClient.getContainerWithPipeline(containerID.getId());
       LOG.debug("Verified new container from SCM {}, {} ",
           containerID, containerWithPipeline.getPipeline().getId());
-      // If no other client added this, go ahead and add this container.
-      if (!containerExist(containerID)) {
-        addNewContainer(containerID.getId(), containerWithPipeline);
-      }
+      // no need call "containerExist" to check, because
+      // 1 containerExist and addNewContainer can not be atomic
+      // 2 addNewContainer will double check the existence
+      addNewContainer(containerWithPipeline);
     } else {
-      // Check if container state is not open. In SCM, container state
-      // changes to CLOSING first, and then the close command is pushed down
-      // to Datanodes. Recon 'learns' this from DN, and hence replica state
-      // will move container state to 'CLOSING'.
-      ContainerInfo containerInfo = getContainer(containerID);
-      if (containerInfo.getState().equals(HddsProtos.LifeCycleState.OPEN)
-          && !replicaState.equals(ContainerReplicaProto.State.OPEN)
-          && isHealthy(replicaState)) {
-        LOG.info("Container {} has state OPEN, but Replica has State {}.",
-            containerID, replicaState);
+      checkContainerStateAndUpdate(containerID, replicaState);
+    }
+  }
+
+  /**
+   * Check and add new containers in batch if not already present in Recon.
+   *
+   * @param containerReplicaProtoList list of containerReplicaProtos.
+   */
+  public void checkAndAddNewContainerBatch(
+      List<ContainerReplicaProto> containerReplicaProtoList) {
+    Map<Boolean, List<ContainerReplicaProto>> containers =
+        containerReplicaProtoList.parallelStream()
+        .collect(Collectors.groupingBy(c ->
+            containerExist(ContainerID.valueOf(c.getContainerID()))));
+
+    List<ContainerReplicaProto> existContainers = null;
+    if (containers.containsKey(true)) {
+      existContainers = containers.get(true);
+    }
+    List<Long> noExistContainers = null;
+    if (containers.containsKey(false)){
+      noExistContainers = containers.get(false).parallelStream().
+          map(ContainerReplicaProto::getContainerID)
+          .collect(Collectors.toList());
+    }
+
+    if (null != noExistContainers) {
+      List<ContainerWithPipeline> verifiedContainerPipeline =
+          scmClient.getExistContainerWithPipelinesInBatch(noExistContainers);
+      LOG.debug("{} new containers have been verified by SCM , " +
+              "{} containers not found at SCM",
+          verifiedContainerPipeline.size(),
+          noExistContainers.size() - verifiedContainerPipeline.size());
+      for (ContainerWithPipeline cwp : verifiedContainerPipeline) {
         try {
-          updateContainerState(containerID, FINALIZE);
-        } catch (InvalidStateTransitionException e) {
-          throw new IOException(e);
+          addNewContainer(cwp);
+        } catch (IOException ioe) {
+          LOG.error("Exception while checking and adding new container.", ioe);
         }
       }
+    }
+
+    if (null != existContainers) {
+      for (ContainerReplicaProto crp : existContainers) {
+        ContainerID cID = ContainerID.valueOf(crp.getContainerID());
+        ContainerReplicaProto.State crpState = crp.getState();
+        try {
+          checkContainerStateAndUpdate(cID, crpState);
+        } catch (Exception ioe){
+          LOG.error("Exception while " +
+              "checkContainerStateAndUpdate container", ioe);
+        }
+      }
+    }
+  }
+
+  /**
+   *  Check if container state is not open. In SCM, container state
+   *  changes to CLOSING first, and then the close command is pushed down
+   *  to Datanodes. Recon 'learns' this from DN, and hence replica state
+   *  will move container state to 'CLOSING'.
+   *
+   * @param containerID containerID to check
+   * @param state  state to be compared
+   * @throws IOException
+   */
+
+  private void checkContainerStateAndUpdate(ContainerID containerID,
+                                            ContainerReplicaProto.State state)
+      throws Exception {
+    ContainerInfo containerInfo = getContainer(containerID);
+    if (containerInfo.getState().equals(HddsProtos.LifeCycleState.OPEN)
+        && !state.equals(ContainerReplicaProto.State.OPEN)
+        && isHealthy(state)) {
+      LOG.info("Container {} has state OPEN, but given state is {}.",
+          containerID, state);
+      updateContainerState(containerID, FINALIZE);
     }
   }
 
@@ -154,12 +215,10 @@ public class ReconContainerManager extends ContainerManagerImpl {
 
   /**
    * Adds a new container to Recon's container manager.
-   * @param containerId id
    * @param containerWithPipeline containerInfo with pipeline info
    * @throws IOException on Error.
    */
-  public void addNewContainer(long containerId,
-                              ContainerWithPipeline containerWithPipeline)
+  public void addNewContainer(ContainerWithPipeline containerWithPipeline)
       throws IOException {
     ContainerInfo containerInfo = containerWithPipeline.getContainerInfo();
     try {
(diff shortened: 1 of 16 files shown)


JacksonYao287:
@avijayanhwx @GlenGeng  please take a review, thx!

bshashikant:
@avijayanhwx , can you please take a look at this?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
```diff
@@ -122,26 +122,87 @@ public void checkAndAddNewContainer(ContainerID containerID,
...
       LOG.debug("Verified new container from SCM {}, {} ",
           containerID, containerWithPipeline.getPipeline().getId());
-      // If no other client added this, go ahead and add this container.
-      if (!containerExist(containerID)) {
-        addNewContainer(containerID.getId(), containerWithPipeline);
-      }
+      // no need call "containerExist" to check, because
+      // 1 containerExist and addNewContainer can not be atomic
+      // 2 addNewContainer will double check the existence
+      addNewContainer(containerID.getId(), containerWithPipeline);
```
avijayanhwx, line 128: Looks like org.apache.hadoop.ozone.recon.scm.ReconContainerManager#addNewContainer does not use the first argument (containerId). Can we remove it?

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/scm/TestReconContainerManager.java
```diff
@@ -127,6 +129,33 @@ public void testCheckAndAddNewContainer() throws IOException {
...
+  public void testCheckAndAddNewContainerBatch() throws IOException {
+    List<ContainerReplicaProto> containerReplicaProtoList = new LinkedList<>();
+    ReconContainerManager containerManager = getContainerManager();
+    for (long i = 200L; i < 300L; i++) {
+      assertFalse(containerManager.containerExist(ContainerID.valueOf(i)));
+      ContainerReplicaProto.Builder ciBuilder =
+          ContainerReplicaProto.newBuilder();
+      ContainerReplicaProto crp = ciBuilder.
+          setContainerID(i).setState(OPEN).build();
+      containerReplicaProtoList.add(crp);
```
avijayanhwx, line 142: Can we also testing adding containers in non OPEN state?

avijayanhwx (commented):
Thank you for working on this @JacksonYao287. Approach looks good to me, I have couple of comments inline.

But, I still think this is not a scalable solution. We have to boostrap Recon with the SCM DB on startup (HDDS-2852) to solve this problem as well as something like HDDS-4355.

JacksonYao287:
@avijayanhwx  thanks for your review!

> But, I still think this is not a scalable solution. We have to boostrap Recon with the SCM DB on startup (HDDS-2852) to solve this problem as well as something like HDDS-4355.

yes,  I agree. but this patch can Speed ​​up processing the container report for recon. Batch is faster than one-by-one. when the number of containers in an ozone is not particularly large, i think this will have a Significant improvement.

BTW,  I will try to complete [HDDS-4177.Bootstrap Recon SCM Container DB](https://issues.apache.org/jira/browse/HDDS-4177) later.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
JacksonYao287: yes , done

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
```diff
@@ -122,26 +122,87 @@ public void checkAndAddNewContainer(ContainerID containerID,
...
+          .collect(Collectors.toList());
+    }
+
+    //for now , if any one container in noExistContainers is not found by SCM,
+    //an IOException will be throw and the whole noExistContainers will be drop.
+    //in some cases，this may slow the process for recon to learn new container,
+    //but it does not matter, just make it simple for the present
+    if (null != noExistContainers) {
+      List<ContainerWithPipeline> verifiedContainerPipeline =
+          scmClient.getContainerWithPipelineBatch(noExistContainers);
```
GlenGeng, line 165: The implementation of `getContainerWithPipelineBatch` in `SCMClientProtocolServer` is 

```
    StringBuilder strContainerIDs = new StringBuilder();
    for (Long containerID : containerIDs) {
      try {
        ContainerWithPipeline cp = getContainerWithPipelineCommon(containerID);
        cpList.add(cp);
        strContainerIDs.append(ContainerID.valueOf(containerID).toString());
        strContainerIDs.append(",");
      } catch (IOException ex) {
        AUDIT.logReadFailure(buildAuditMessageForFailure(
            SCMAction.GET_CONTAINER_WITH_PIPELINE_BATCH,
            Collections.singletonMap("containerID",
                ContainerID.valueOf(containerID).toString()), ex));
        throw ex;
      }
    }
```

Say the batch is `C1, C2, C3 C4`, `C2` is not existed in SCM, will C3 and C4 be ignored ?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
```diff
@@ -122,26 +122,87 @@ public void checkAndAddNewContainer(ContainerID containerID,
...
+   * Check and add new containers in batch if not already present in Recon.
+   *
+   * @param containerReplicaProtoList list of containerReplicaProtos.
+   * @throws IOException on Error.
+   */
+  public void checkAndAddNewContainerBatch(
+      List<ContainerReplicaProto> containerReplicaProtoList)
+      throws IOException {
+    Map<Boolean, List<ContainerReplicaProto>> containers =
+        containerReplicaProtoList.parallelStream()
```
GlenGeng, line 144: `stream().filter()` may be simpler and more straightforward  here.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
```diff
@@ -122,26 +122,87 @@ public void checkAndAddNewContainer(ContainerID containerID,
           scmClient.getContainerWithPipeline(containerID.getId());
       LOG.debug("Verified new container from SCM {}, {} ",
           containerID, containerWithPipeline.getPipeline().getId());
-      // If no other client added this, go ahead and add this container.
-      if (!containerExist(containerID)) {
-        addNewContainer(containerID.getId(), containerWithPipeline);
-      }
+      // no need call "containerExist" to check, because
+      // 1 containerExist and addNewContainer can not be atomic
+      // 2 addNewContainer will double check the existence
```
GlenGeng, line 127: If we finally adopt the batch solution, can we drop this singular method ?

hadoop-ozone/recon/src/test/java/org/apache/hadoop/ozone/recon/scm/TestReconContainerManager.java
JacksonYao287: added

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
JacksonYao287: thanks for the review!  maybe we can  maintain it here for the present, and remove it later in another patch

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
JacksonYao287: here , I want to divide "containerReplicaProtoList" into two parts , and both of the two parts will be used in the subsequent code。if i use "stream().filter()" instead, only one of the two parts got, so I have to call "stream().filter()" twice.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
JacksonYao287: yes ,  according to the current implementation of "getContainerWithPipelineBatch" , if any one container in noExistContainers is not found by SCM, an IOException will be throw and the whole noExistContainers will be drop.
but I think it has little effect here，and just make it simple for the present.
maybe we can make another patch to rewrite "getContainerWithPipelineBatch" in scm , which will just return the list of found containers, and do not throw an exception.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
symious: I think the original check is kind of optimistic locking here?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
symious: I think these codes can be simpler, seems too redundant for lambda here.

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
symious: Can we fallback to the original procedure here if the above issue happened?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerReportHandler.java
```diff
@@ -48,26 +46,16 @@ public ReconContainerReportHandler(NodeManager nodeManager,
...
-            reportFromDatanode.getDatanodeDetails());
-      } catch (IOException ioEx) {
-        LOG.error("Exception while checking and adding new container.", ioEx);
-      }
-      LOG.debug("Got container report for containerID {} ",
-          containerReplicaProto.getContainerID());
+    List<ContainerReplicaProto> containerReplicaProtoList =
+        reportFromDatanode.getReport().getReportsList();
+    try {
+      containerManager.checkAndAddNewContainerBatch(containerReplicaProtoList);
```
symious, line 54: Maybe we can set a batch size limit here to avoid the performance impact on SCM?

hadoop-ozone/recon/src/main/java/org/apache/hadoop/ozone/recon/scm/ReconContainerManager.java
JacksonYao287: @symious Thank for the review!  "addNewContainer" will call "ContainerStateMap#contains" ,and this is what "containerExist" exactly does, so no need here.

symious: There are still some steps that need to be taken before `addNewContainer` calls `ContainerStateMap$contains`, with the contains check here, the performance would be better.
Besides, in `ContainerStateManagerImpl`, the writeLock would be required to check if the container exists, it's not reasonable to involve the lock here.

symious: "check the existence and add" isn't meant to be an atomic operation here, it's just used to avoid some unnecessary operations if the other clients have updated the containerId already.

JacksonYao287:
@avijayanhwx @symious @GlenGeng  please take a review.
I rewrite this pr , there a two big  changes :
1 add a new SCM service "getExistContainerWithPipelinesInBatch" , which will ask scm that which containers in the given list can be found by scm and will never throw an exception at client-side. "ReconContainerManager#checkAndAddNewContainerBatch" now use this method instead of "getContainerWithPipelineBatch"  , which may throw an exception and drop the whole reports in a heartbeat.

2 fix a bug. for now , a closed container may be reported by datanode, and the pipeline of this container may be closed , recon can not sync this pipeline info from scm. so recon should have this ability and a no-open container to it's `containerStateMap` without the corresponding pipeling info.

avijayanhwx (commented):
I am OK with getting this patch as it is now, but would like to revisit/remove it when we do SCM DB boostrap on Recon. Generally, I am not for adding ad hoc APIs in SCM to support Recon. Slowly, the SCM API layer will become harder to maintain. @symious / @GlenGeng if you are ok with this approach as well, I can get this in.

symious:
@avijayanhwx Thanks for the review. I think we can keep the following check. 
```
      if (!containerExist(containerID)) {
        addNewContainer(containerID.getId(), containerWithPipeline);
      }
```
Since the inner check @JacksonYao287 mentioned would involve additional write lock operations.

JacksonYao287:
> @avijayanhwx Thanks for the review. I think we can keep the following check.
> 
> ```
>       if (!containerExist(containerID)) {
>         addNewContainer(containerID.getId(), containerWithPipeline);
>       }
> ```
> 
> Since the inner check @JacksonYao287 mentioned would involve additional write lock operations.

@avijayanhwx @GlenGeng  what is your opinion

> @avijayanhwx Thanks for the review. I think we can keep the following check.
> 
> ```
>       if (!containerExist(containerID)) {
>         addNewContainer(containerID.getId(), containerWithPipeline);
>       }
> ```
> 
> Since the inner check @JacksonYao287 mentioned would involve additional write lock operations.

@avijayanhwx  what is you opinion about this?

JacksonYao287:
> @avijayanhwx Thanks for the review. I think we can keep the following check.
> 
> ```
>       if (!containerExist(containerID)) {
>         addNewContainer(containerID.getId(), containerWithPipeline);
>       }
> ```
> 
> Since the inner check @JacksonYao287 mentioned would involve additional write lock operations.

@avijayanhwx please take a look, what is your opinion?

avijayanhwx (approved):
I have added a TODO in HDDS-4177 to revisit this logic. For now, I am +1 with this approach since it is causing an OOM issue.

PR #2173, opened 2021-04-22
HDDS-5127. Fix getServiceList when SCM HA is enabled

bharatviswa504:

Make HA config when SCM HA is enabled to figure out SCM client address.

https://issues.apache.org/jira/browse/HDDS-5127

## How was this patch tested?

Added test.

### DIFF
--- a/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
+++ b/hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
@@ -132,36 +132,6 @@ public final class HddsUtils {
   }
 
   /**
-   * Retrieve the socket address that should be used by clients to connect
-   * to the SCM for block service. If
-   * {@link ScmConfigKeys#OZONE_SCM_BLOCK_CLIENT_ADDRESS_KEY} is not defined
-   * then {@link ScmConfigKeys#OZONE_SCM_CLIENT_ADDRESS_KEY} is used. If neither
-   * is defined then {@link ScmConfigKeys#OZONE_SCM_NAMES} is used.
-   *
-   * @return Target {@code InetSocketAddress} for the SCM block client endpoint.
-   * @throws IllegalArgumentException if configuration is not defined.
-   */
-  public static InetSocketAddress getScmAddressForBlockClients(
-      ConfigurationSource conf) {
-    Optional<String> host = getHostNameFromConfigKeys(conf,
-        ScmConfigKeys.OZONE_SCM_BLOCK_CLIENT_ADDRESS_KEY,
-        ScmConfigKeys.OZONE_SCM_CLIENT_ADDRESS_KEY);
-
-    if (!host.isPresent()) {
-      // Fallback to Ozone SCM name
-      host = Optional.of(getSingleSCMAddress(conf).getHostName());
-    }
-
-    final int port = getPortNumberFromConfigKeys(conf,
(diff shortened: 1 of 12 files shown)


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
```diff
@@ -95,21 +97,38 @@ private HddsUtils() {
...
+              "set for SCM service-id " + scmNodeInfo.getServiceId() +
+              "node-id" + scmNodeInfo.getNodeId());
+        }
+        scmAddressList.add(
+            NetUtils.createSocketAddr(scmNodeInfo.getScmClientAddress()));
+      }
+      return scmAddressList;
+    } else {
+      Optional< String > host = getHostNameFromConfigKeys(conf,
+          ScmConfigKeys.OZONE_SCM_CLIENT_ADDRESS_KEY);
```
bshashikant, line 119: NIT: Optional< String > host -> Optional<String> host

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
```diff
@@ -95,21 +97,38 @@ private HddsUtils() {
...
 
-    final int port = getPortNumberFromConfigKeys(conf,
-        ScmConfigKeys.OZONE_SCM_CLIENT_ADDRESS_KEY)
-        .orElse(ScmConfigKeys.OZONE_SCM_CLIENT_PORT_DEFAULT);
+      if (!host.isPresent()) {
```
bshashikant, line 121: I think in case of non_HA setup , scm client address is not defined , client should still be bale to fall back to get the scm address from ozone.scm.names as before.

hadoop-ozone/client/src/test/java/org/apache/hadoop/ozone/client/TestHddsClientUtils.java
```diff
@@ -173,34 +222,6 @@ public void testBlockClientFallbackToScmNamesWithPort() {
     assertEquals(OZONE_SCM_BLOCK_CLIENT_PORT_DEFAULT, address.getPort());
   }
 
-  @Test
-  public void testClientFallbackToScmNamesNoPort() {
-    // When OZONE_SCM_CLIENT_ADDRESS_KEY is undefined, it should fallback
```
bshashikant, line 178: I think, once we restore back to failback logic to ozone.scm.names, these tests will still be valid.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
bharatviswa504: Refactored.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/HddsUtils.java
bharatviswa504: Done

hadoop-ozone/client/src/test/java/org/apache/hadoop/ozone/client/TestHddsClientUtils.java
bharatviswa504: Yes added them back

bharatviswa504:
As when re-work on review comments, seen that we need HDDS-5140 also, so included HDDS-5140 also in this. (It is just a cleanup change)

bshashikant (approved):

bharatviswa504:

PR #2179, opened 2021-04-23
HDDS-5118. Recover from failure during upgrade action.

errose28:

Remove the `upgradingToLayoutVersion` VERSION file marker, which was previously used to mark a finalization action as in progress in case of a crash during the action. There will be no special handling for this case, and documentation is updated to indicate that finalization actions that fail will be re-run on a subsequent finalize call. The pre-finalize validation action can be used for necessary cleanup on startup. Docs for other upgrade actions have also been updated, although their behavior is not modified.

HDDS-5118

## How was this patch tested?

Unit tests updated.


avijayanhwx (approved):
Thanks for working on this @errose28. LGTM +1.

PR #2181, opened 2021-04-24
HDDS-4515. Datanodes should be able to persist and load CRL

vivekratnavel:

 - Refactor common classes like CRLInfo, CRLInfoCodec, etc., to hadoop-hdds-common
 - Add a new RocksDB in Datanodes to persist CRL related information
 - Add a new DBStore to access the newly created Rocksdb

https://issues.apache.org/jira/browse/HDDS-4515

## How was this patch tested?

Unit tests and Docker-based cluster.


hadoop-hdds/container-service/src/test/java/org/apache/hadoop/hdds/datanode/metadata/TestDatanodeCRLStoreImpl.java
```diff
@@ -0,0 +1,126 @@
...
+    assertEquals(1L, crlInfo.getCrlSequenceID());
+    assertEquals(x509Certificate.getSerialNumber(),
+        crlInfo.getX509CRL().getRevokedCertificates()
+            .iterator().next().getSerialNumber());
+
+    // Test that restarting the store does not affect the data already persisted
+    dnCRLStore.stop();
+    dnCRLStore = new DatanodeCRLStoreImpl(conf);
+    assertEquals(5L, (long) dnCRLStore.getLatestCRLSequenceID());
+    assertEquals(1L, dnCRLStore.getPendingCRLs().size());
```
xiaoyuyao, line 116: Can we add a dnCRLStore.stop() at the end?

xiaoyuyao (commented):
LGTM, just one minor comment added inline...

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/hdds/datanode/metadata/TestDatanodeCRLStoreImpl.java
vivekratnavel: Sure. I have added this line in the latest patch. But, it will be taken care of by the destroyDbStore() regardless after each test.

xiaoyuyao (approved):
LGTM, +1.

PR #2183, opened 2021-04-26
HDDS-5099. Error with unit test for hdds.container-service TestSchemaOneBackwardsCompatibility

neils-dev:

Fixes for intermittent unit test failures found in  hdds.container-service _TestSchemaOneBackwardsCompatibility_ and in _TestBlockDeletingService_.  Current implementation for unit tests uses the _BackgroundService_ for block deletion.  This service uses a thread pool executor that uses worker thread tasks that run asynchronously.  The unit test callers for this service 'start' the _PeriodicalTask_ of the _BackgroundService_ and expect the workers to finish deleting blocks when they do checks for expected results.  The thread pool executor runs the worker tasks, _however_ they do not run to completion prior to returning control to the unit test callers.  This leads to an **intermittent error** due to results not ready when the unit test caller checks for expected results.

Patch _**extends**_ the _PeriodicalTask_ for the background service and runs the worker tasks async until completion prior to returning control to the unit test caller.  This is realized in the _BlockDeletingServiceTestImpl_ _**PeriodicalTaskTestImpl**_ implementation.

https://issues.apache.org/jira/browse/HDDS-5099

## How was this patch tested?

Patch was tested with unit tests : _**TestSchemaOneBackwardsCompatibility**_ and _**TestBlockDeletingService**_.

to see error, reproduce the error reliably (it is intermittent), can force failure though injecting fault into the 
_BlockDeletingService.java_ implementation - adding a small delay to the BackgroundTask call():
```
    public BackgroundTaskResult call() throws Exception {
      ContainerBackgroundTaskResult crr;
      final Container container = ozoneContainer.getContainerSet()
          .getContainer(containerData.getContainerID());
      container.writeLock();
      File dataDir = new File(containerData.getChunksPath());
      // Scan container's db and get list of under deletion blocks
      try (ReferenceCountedDB meta = BlockUtils.getDB(containerData, conf)) {
        TimeUnit.MILLISECONDS.sleep(100); 
```
**pre-patch**

`hadoop-hdds/container-service$ mvn -Dtest=TestSchemaOneBackwardsCompatibility#testDelete test` 
TestSchemaOneBackwardsCompatibility_prepatch

`hadoop-hdds/container-service$ mvn -Dtest=TestBlockDeletingService test
`
testBlockDeletingService_prepatch

**after patch is applied**

`hadoop-hdds/container-service$ mvn -Dtest=TestSchemaOneBackwardsCompatibility#testDelete test` 
TestSchemaOneBackwardsCompatibility_afterpatch

`hadoop-hdds/container-service$ mvn -Dtest=TestBlockDeletingService test`

TestBlockDeletingService_afterpatch

### DIFF
--- a/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
+++ b/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
@@ -398,7 +398,7 @@ public class TestBlockDeletingService {
   @Test
   public void testBlockDeletion() throws Exception {
     DatanodeConfiguration dnConf = conf.getObject(DatanodeConfiguration.class);
-    dnConf.setBlockDeletionLimit(3);
+    dnConf.setBlockDeletionLimit(2);
     this.blockLimitPerInterval = dnConf.getBlockDeletionLimit();
     conf.setFromObject(dnConf);
     ContainerSet containerSet = new ContainerSet();
@@ -448,7 +448,8 @@ public class TestBlockDeletingService {
       deleteAndWait(svc, 1);
 
       GenericTestUtils.waitFor(() ->
-              containerData.get(0).getBytesUsed() < 3, 100, 3000);
+              containerData.get(0).getBytesUsed() == containerSpace /
+                      3, 100, 3000);
       // After first interval 2 blocks will be deleted. Hence, current space
       // used by the container should be less than the space used by the
       // container initially(before running deletion services).
@@ -711,8 +712,8 @@ public class TestBlockDeletingService {
       deleteAndWait(service, 1);
 
       GenericTestUtils.waitFor(() ->
-              blockLimitPerInterval*blockSpace ==
-                      (totalContainerSpace-
+              blockLimitPerInterval * blockSpace ==
+                      (totalContainerSpace -
                               currentBlockSpace(containerData, containerCount)),
               100, 3000);
 
@@ -726,8 +727,8 @@ public class TestBlockDeletingService {
 
       long totalContainerBlocks = blocksPerContainer*containerCount;
       GenericTestUtils.waitFor(() ->
-              totalContainerBlocks*blockSpace ==
-                      (totalContainerSpace-
+              totalContainerBlocks * blockSpace ==
+                      (totalContainerSpace -
                               currentBlockSpace(containerData, containerCount)),
               100, 3000);
 
--- a/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
+++ b/hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
@@ -447,6 +447,8 @@ public class TestBlockDeletingService {
       // An interval will delete 1 * 2 blocks
       deleteAndWait(svc, 1);
 
+      GenericTestUtils.waitFor(() ->
+              containerData.get(0).getBytesUsed() < 3, 100, 3000);
       // After first interval 2 blocks will be deleted. Hence, current space
       // used by the container should be less than the space used by the
       // container initially(before running deletion services).
@@ -456,8 +458,8 @@ public class TestBlockDeletingService {
 
       // After deletion of all 3 blocks, space used by the containers
       // should be zero.
-      containerSpace = containerData.get(0).getBytesUsed();
-      Assert.assertTrue(containerSpace == 0);
+      GenericTestUtils.waitFor(() ->
+              containerData.get(0).getBytesUsed() == 0, 100, 3000);
 
       // Check finally DB counters.
       // Not checking bytes used, as handler is a mock call.
@@ -629,8 +631,11 @@ public class TestBlockDeletingService {
       // Hence, space used by the container of whose block has been
       // deleted should be zero.
       deleteAndWait(service, 1);
-      Assert.assertTrue((containerData.get(0).getBytesUsed() == 0)
-          || containerData.get(1).getBytesUsed() == 0);
+
+      GenericTestUtils.waitFor(() ->
+              (containerData.get(0).getBytesUsed() == 0 ||
+                      containerData.get(1).getBytesUsed() == 0),
+              100, 3000);
 
       Assert.assertFalse((containerData.get(0).getBytesUsed() == 0) && (
           containerData.get(1).getBytesUsed() == 0));
@@ -639,8 +644,10 @@ public class TestBlockDeletingService {
       // containers should be zero.
       deleteAndWait(service, 2);
 
-      Assert.assertTrue((containerData.get(0).getBytesUsed() == 0) && (
-          containerData.get(1).getBytesUsed() == 0));
+      GenericTestUtils.waitFor(() ->
+              (containerData.get(0).getBytesUsed() ==
+                      0 && containerData.get(1).getBytesUsed() == 0),
+              100, 3000);
     } finally {
       service.shutdown();
     }
@@ -702,9 +709,12 @@ public class TestBlockDeletingService {
       // Deleted space of 10 blocks should be equal to (initial total space
       // of container - current total space of container).
       deleteAndWait(service, 1);
-      Assert.assertEquals(blockLimitPerInterval * blockSpace,
-          (totalContainerSpace - currentBlockSpace(containerData,
-              containerCount)));
+
+      GenericTestUtils.waitFor(() ->
+              blockLimitPerInterval*blockSpace ==
+                      (totalContainerSpace-
+                              currentBlockSpace(containerData, containerCount)),
+              100, 3000);
 
       // There is only 5 blocks left to runDeletingTasks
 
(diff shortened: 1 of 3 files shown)


symious:
Since we have found the root cause of the intermittent exception in this case, can we simply wait a while before `future.get` in `BlockDeletingServiceTestImpl$start"?

neils-dev:
> Since we have found the root cause of the intermittent exception in this case, can we simply wait a while before `future.get` in `BlockDeletingServiceTestImpl$start"?

Thanks @symious - true, a delay in BlockDeleteingServiceTestImpl may work provided the delay is sufficient for the workers to complete the tasks; we can choose 10 - 200 ms however this value would vary from platform to platform.  Here we have the thread executor dispatch the completion futures and return when the worker threads are finished, ensuring that the results are available regardless of the platform/environment we are running on.  Another implementation can be through a latch and releasing the caller one the workers have finished, however that would require implementing the counter creating  the same behavior as the patch.

symious:
I think one point of these unit tests is to test the feature of `BackgroundService.PeriodicTask`, if we create a new PeriodicTask implementation for tests, it seems a little diverged from the goal of these unit tests.
In [HDDS-4231](https://issues.apache.org/jira/browse/HDDS-4231), the waiting for async results is removed, I think we can only focus on solving the intermittent exception here, no need to write the original code for tests.
Besides, the comment of `BackgroundService.PeriodicTask` is not accurate since HDDS-4231, could you help to update the comment?

neils-dev:
> I think one point of these unit tests is to test the feature of `BackgroundService.PeriodicTask`, if we create a new PeriodicTask implementation for tests, it seems a little diverged from the goal of these unit tests.
> In [HDDS-4231](https://issues.apache.org/jira/browse/HDDS-4231), the waiting for async results is removed, I think we can only focus on solving the intermittent exception here, no need to write the original code for tests.
> Besides, the comment of `BackgroundService.PeriodicTask` is not accurate since [HDDS-4231](https://issues.apache.org/jira/browse/HDDS-4231), could you help to update the comment?

Thanks for the followup @symious.  I understand your comment on extending the `PeriodicTask`, the goal of the unit test in this case is to test deleting blocks - both with the older schema and the block deleting service.  The goal of the unit test in these cases I believe is not to exercise the `PeriodicTask` of the `background service`.  Further, we are not using the background task in the intended '_periodic_' fashion in these cases as the comment in line 136 implies and how the test operates in a single fashion and _not_ periodic.  

On the [HDDS-4231] issues, the extended `PeriodicalTaskTestImpl` propagates this LOG.warn as in line 118 so will not change affected callers such as the `TestBlockDeletingService`.

symious:
I think the point HDDS-4231 is to remove the waiting for async deletes. So before HDDS-4231, this test won't have intermittent exception, since it will always wait for the deletion to complete.
Using a different deletion implementation seems a little confused here, better to use the original one for the goal of test.

lokeshj1703 (commented):
@neils-dev Thanks for working on this! I agree with @symious that we should not add our implementation of PeriodicalTask here. 
How about we add numTasksProcessed inside BackgroundService? In the BlockDeletingServiceTestImpl we can check the same. I think the problem can be solved if numOfProcessed is incremented only after task execution.

neils-dev:
Changes pushed to reflect comments.  Thanks **@symious**  and **@lokeshj1703** for insights and suggestions on the patch to fix the existing problems with the _BlockDeletionService_ related unit tests.  To exercise the existing _PeriodicTask_ implementation with our unit tests, I _**removed**_ the extended task from the _BlockDeletingServiceTestImpl_.  Instead a timeout is used (3 sec as was defined previously in the _BlockDeletingServiceTestImpl_), for tests involving the background service _BlockDeletionService_.  Here, using the _waitFor()_ test util we poll for the workers to finish.  This in line with existing tests in the _TestBockDeletionService_.  With this implementation we do **_not_** impact the production _backgroundservice_ or any of its subclasses with added variables both volatile and non volatile that can cause unnecessary processing and contention.  Thanks guys.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/testutils/BlockDeletingServiceTestImpl.java
```diff
@@ -80,6 +80,7 @@ public void start() {
           break;
         }
         Future<?> future = this.getExecutorService().submit(svc);
+
```
elek, line 83: This line may not be required to modified and this case this test will remain unchanged.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestSchemaOneBackwardsCompatibility.java
```diff
@@ -279,6 +288,8 @@ public void testDelete() throws Exception {
               refCountedDB.getStore().getMetadataTable();
       assertEquals(expectedRegularBlocks + expectedDeletingBlocks,
               (long)metadataTable.get(OzoneConsts.BLOCK_COUNT));
+    //} catch(IOException ex) {
```
elek, line 291: Do we need this commented out test? Or can we just remove this part?

elek (commented):
Thanks the patch @neils-dev. I have two very minor nit comments, but it looks good to me.

@lokeshj1703 / @symious: are you fine with the latest version?

symious:
> Thanks the patch @neils-dev. I have two very minor nit comments, but it looks good to me.
> 
> @lokeshj1703 / @symious: are you fine with the latest version?

@neils-dev Thanks for the update. LGTM after solving @elek 's comment.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
```diff
@@ -702,9 +709,12 @@ public void testBlockThrottle() throws Exception {
...
       deleteAndWait(service, 1);
-      Assert.assertEquals(blockLimitPerInterval * blockSpace,
-          (totalContainerSpace - currentBlockSpace(containerData,
-              containerCount)));
+
+      GenericTestUtils.waitFor(() ->
+              blockLimitPerInterval*blockSpace ==
+                      (totalContainerSpace-
+                              currentBlockSpace(containerData, containerCount)),
+              100, 3000);
```
lokeshj1703, line 717: Minor NIT: Space between `blockLimitPerInterval * blockSpace` and
`(totalContainerSpace -`

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
```diff
@@ -713,9 +723,14 @@ public void testBlockThrottle() throws Exception {
...
       deleteAndWait(service, 2);
-      Assert.assertEquals(blocksPerContainer * containerCount * blockSpace,
-          (totalContainerSpace - currentBlockSpace(containerData,
-              containerCount)));
+
+      long totalContainerBlocks = blocksPerContainer*containerCount;
+      GenericTestUtils.waitFor(() ->
+              totalContainerBlocks*blockSpace ==
+                      (totalContainerSpace-
+                              currentBlockSpace(containerData, containerCount)),
```
lokeshj1703, line 731: NIT. Same as above.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
```diff
@@ -447,6 +447,8 @@ public void testBlockDeletion() throws Exception {
       // An interval will delete 1 * 2 blocks
       deleteAndWait(svc, 1);
 
+      GenericTestUtils.waitFor(() ->
+              containerData.get(0).getBytesUsed() < 3, 100, 3000);
```
lokeshj1703, line 451: Unrelated to your changes. Can we set the block deletion limit to 2 at line 401?
`dnConf.setBlockDeletionLimit(2);`
Then here we should check `containerData.get(0).getBytesUsed() == 100`

lokeshj1703 (commented):
@neils-dev Thanks for updating the PR! The changes look good to me. I have few minor comments inline.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/testutils/BlockDeletingServiceTestImpl.java
neils-dev: Left over from previous additions to `BlockDeletingServiceTestImpl` - unnecessary and removed.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestSchemaOneBackwardsCompatibility.java
neils-dev: unnecessary, removed.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
neils-dev: change added.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
neils-dev: change added.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/common/TestBlockDeletingService.java
neils-dev: Thanks @lokeshj1703 - `dfConf.setBlockDeletionLimit` corrected from 3 to 2 for test.  Setting conditional check to equiv 100 (`containerData.get(0).getBytesUsed() == containerSpace / 3`).

neils-dev:
Changes from reviewers comments added in latest commit - thanks @lokeshj1703 , @symious and @elek !

lokeshj1703 (approved):
@neils-dev Thanks for updating the PR! The changes look good to me. +1.

PR #2185, opened 2021-04-26
HDDS-5097. [FSO] Cleanup integration tests and reduce the execution time

rakeshadr:

This task is to reduce the test execution time of integration tests.
 - refactored the costliest test and reduced the execution time.
 - avoided the tests which are not relevant to PREFIX layout
 - moved independent tests to already existing test class to reuse the same cluster.
 
https://issues.apache.org/jira/browse/HDDS-5097

## How was this patch tested?

Existing test cases.


rakeshadr:
**Integration (filesystem-hdds)**:
 
`Build times: ~94 mins (before patch) , 63 mins (after this patch), ~56 mins (master)`

For the above`(after this patch)` section, I've referred the [finished integration test run](https://github.com/apache/ozone/pull/2185/checks?check_run_id=2444322738), recently.

I hope the feature branch test case optimization gives a better test execution time reduction. Appreciate your feedback on this patch, thanks!

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneDirectory.java
```diff
@@ -1,183 +0,0 @@
...
-    conf.setInt(FS_TRASH_INTERVAL_KEY, 1);
-    TestOMRequestUtils.configureFSOptimizedPaths(conf,
-            true, OMConfigKeys.OZONE_OM_METADATA_LAYOUT_PREFIX);
-    cluster = MiniOzoneCluster.newBuilder(conf)
-            .setNumDatanodes(3)
-            .build();
-    cluster.waitForClusterToBeReady();
-    // create a volume and a bucket to be used by OzoneFileSystem
-    OzoneBucket bucket = TestDataUtil.createVolumeAndBucket(cluster);
-    volumeName = bucket.getVolumeName();
```
mukul1987, line 161: How is this tested now ? with the test removed ?

mukul1987 (approved):

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/fs/ozone/TestOzoneDirectory.java
rakeshadr: Thanks @mukul1987 for the reviews. 

I've moved `TestOzoneDirectory#testMultiLevelDirs` and `TestOzoneFileOps#testCreateFile` to `TestOzoneFileSystemWithFSO.java` test class. Both were starting/stopping cluster and taking nearly ~25secs each, total ~50secs. IMHO, these two are running on `MiniOzoneCluster` and can be tested using TestOzoneFileSystemWithFSO or any other class which starts `MiniOzoneCluster`, where its finishing the each run in milliseconds (<100 milliseconds in my local env).

mukul1987: ok sure, LGTM

rakeshadr:
Thanks @mukul1987 for the reviews and approving it. I'll merge it to the branch.

PR #2186, opened 2021-04-26
HDDS-4729. Add token support for container admin operations

adoroszlai:

This change introduces container tokens.  These are similar to existing "block tokens", but applied and required for container operations.  They are issued by SCM in response to the new `getContainerToken` request.

https://issues.apache.org/jira/browse/HDDS-4729

## How was this patch tested?

Added unit test.  Updated existing integration tests (including fixing and enabling `TestSecureContainerServer`).

`HEAD~2`: https://github.com/adoroszlai/hadoop-ozone/actions/runs/786346758
`HEAD~1`: https://github.com/adoroszlai/hadoop-ozone/actions/runs/786799698
`HEAD` (in progress): https://github.com/adoroszlai/hadoop-ozone/actions/runs/791801535


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/token/ShortLivedTokenVerifier.java
```diff
@@ -0,0 +1,135 @@
...
+  protected void verify(T tokenId, ContainerCommandRequestProto cmd)
+      throws SCMSecurityException {
+    // NOP
+  }
+
+  @Override
+  public void verify(String user, Token<?> token,
+      ContainerCommandRequestProto cmd) throws SCMSecurityException {
+
+    if (!conf.isBlockTokenEnabled()) {
```
xiaoyuyao, line 73: This is not block token specific, do we have a generic conf here for scm token?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/token/ShortLivedTokenVerifier.java
```diff
@@ -0,0 +1,135 @@
...
+      throw new SCMSecurityException("Certificate client not available " +
+          "to validate token");
+    }
+
+    T tokenId = createTokenIdentifier();
+    try {
+      tokenId.readFields(new DataInputStream(new ByteArrayInputStream(
+          token.getIdentifier())));
+    } catch (IOException ex) {
+      throw new BlockTokenException("Failed to decode token : " + token);
```
xiaoyuyao, line 91: This is not block token specific, can we have a generic SCM token exception here?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/token/ShortLivedTokenVerifier.java
adoroszlai: Thanks @xiaoyuyao for taking a look.

I thought having a single config item for both block token and container token would be enough.  Is there a reason to enable only one of them, but not the other one?

Regarding the names (config item, exception), I kept "block token"-related ones for backward compatibility.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/security/x509/SecurityConfig.java
```diff
@@ -330,6 +330,9 @@ public Duration getMaxCertificateDuration() {
     return this.maxCertDuration;
   }
 
+  /**
+   * Whether to require short-lived tokens for block and container operations.
+   */
   public boolean isBlockTokenEnabled() {
```
xiaoyuyao, line 336: Block token are container token are all validated on datanodes but issued by different services, can we define separate  configurations to enable/disable them differently?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/HddsDispatcher.java
```diff
@@ -109,10 +109,8 @@ public HddsDispatcher(ConfigurationSource config, ContainerSet contSet,
     this.containerCloseThreshold = conf.getFloat(
         HddsConfigKeys.HDDS_CONTAINER_CLOSE_THRESHOLD,
         HddsConfigKeys.HDDS_CONTAINER_CLOSE_THRESHOLD_DEFAULT);
-    this.isBlockTokenEnabled = conf.getBoolean(
-        HddsConfigKeys.HDDS_BLOCK_TOKEN_ENABLED,
-        HddsConfigKeys.HDDS_BLOCK_TOKEN_ENABLED_DEFAULT);
-    this.tokenVerifier = tokenVerifier;
+    this.tokenVerifier = tokenVerifier != null ? tokenVerifier
```
xiaoyuyao, line 112: NIT: can we move new NoopTokenVerifier() to the caller in OzoneContainer?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/HddsDispatcher.java
```diff
@@ -425,14 +423,14 @@ ContainerCommandResponseProto createContainer(
     return handler.handle(requestBuilder.build(), null, null);
   }
 
-  private void validateBlockToken(
+  private void validateToken(
       ContainerCommandRequestProto msg) throws IOException {
-    if (isBlockTokenEnabled && tokenVerifier != null &&
-        HddsUtils.requireBlockToken(msg.getCmdType())) {
+
+    if (tokenVerifier != null) {
```
xiaoyuyao, line 429: Do we still need the null check with the NoopTokenVerifier?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/client/CertificateClient.java
```diff
@@ -278,4 +281,14 @@ void storeRootCACertificate(String pemEncodedCert, boolean force)
    */
   long getLatestCrlId() throws IOException;
 
+  default void assertKeyPair() throws OzoneSecurityException {
+    try {
+      Objects.requireNonNull(getPublicKey());
+      Objects.requireNonNull(getPrivateKey());
+      Objects.requireNonNull(getCertificate());
```
xiaoyuyao, line 288: NIT: can we rename it to like assertValidKeyCerts to reflect the code?

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/security/x509/SecurityConfig.java
adoroszlai: I don't think introducing a new config property for this adds value.  I'm not sure why we even have a config property for enabling block tokens.  I think they should have been enabled in all secure clusters, ie. by `ozone.security.enabled=true`.  But now we have `hdds.block.token.enabled`.  Do we really need another config property just to force all users who want security to set this new config, too?

We would also need another one for container token token lifetime.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/HddsDispatcher.java
adoroszlai: I think checking for `null` here is more defensive.  Also, `HddsDispatcher` is created by several tests, I wanted to avoid changing those.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/security/x509/SecurityConfig.java
xiaoyuyao: Technical wise, I think that we could live without a separate config. But there are several reasons for having the token on datanode optional:
1. performance: the additional validation adds the cost for each block I/O
2. backward compatibility: old client will be broken without upgrade.
3. deployment with flexibility  

bq. We would also need another one for container token token lifetime.
We can use a default one for the short lived tokens if customer does not need to specify. This is common in  hadoop delegation tokens.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/HddsDispatcher.java
xiaoyuyao: I found that too from later code changes in tests. Good to be defensive here.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/impl/HddsDispatcher.java
adoroszlai: Removed.

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/security/x509/certificate/client/CertificateClient.java
adoroszlai: Renamed to `assertValidKeysAndCertificate`.

adoroszlai:
Thanks @xiaoyuyao, I have updated the patch based on your comments.  Can you please take another look?

xiaoyuyao (commented):
Thanks @adoroszlai for the update. The change LGTM. I just have some questions added inline wrt. mixed usage of container token/block token configuration for the client code.

adoroszlai:
@xiaoyuyao Thanks for taking another look.

> I just have some questions added inline wrt. mixed usage of container token/block token configuration for the client code.

Can you please point me to these questions?  I cannot find them.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
```diff
@@ -790,6 +795,23 @@ public boolean getReplicationManagerStatus() {
         .collect(Collectors.toList());
   }
 
+  @Override
+  public Token<?> getContainerToken(ContainerID containerID)
+      throws IOException {
+    String remoteUser = getRemoteUserName();
+    getScm().checkAdminAccess(remoteUser);
+
+    if (!blockTokenEnabled) {
```
xiaoyuyao, line 804: should we use the container token enabled key to determine whether return a valid container token or not?

hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/ContainerOperationClient.java
```diff
@@ -86,6 +91,8 @@ public ContainerOperationClient(OzoneConfiguration conf) throws IOException {
       replicationFactor = HddsProtos.ReplicationFactor.ONE;
       replicationType = HddsProtos.ReplicationType.STAND_ALONE;
     }
+    blockTokenEnabled = conf.getBoolean(HDDS_BLOCK_TOKEN_ENABLED,
```
xiaoyuyao, line 94: same as above?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
adoroszlai: Thanks, fixed.

adoroszlai:
Thanks @xiaoyuyao for spotting the enabled flag mismatch.  Patch is updated, please take another look.

xiaoyuyao:
Thanks @adoroszlai  for the update. Latest change LGTM, +1.

adoroszlai:
Thanks @xiaoyuyao for the reviews.

PR #1885, opened 2021-02-02
HDDS-4504. Datanode deletion config should be based on number of blocks

aryangupta1998:
## What changes were proposed in this pull request?

Currently datanode's deletion config is based on number of containers on which deletion should be performed(default value 10). The deletion is only performed on those containers.

It should rather be based on number of blocks to delete per interval rather than number of containers to process. If a datanode has 100000 containers and every container has 10 blocks to delete then deletion of all these blocks can take a lot of time with the current configuration. It would take 100000/10 = 10000 intervals for deletion to complete.

## What is the link to the Apache JIRA

https://issues.apache.org/jira/browse/HDDS-4504

## How was this patch tested?

Tested Manually

PR #2188, opened 2021-04-27
HDDS-5147. Intermittent test failure in TestContainerDeletionChoosingPolicy#testRandomChoosingPolicy

elek:

The `TestContainerDeletionChoosingPolicy#testRandomChoosingPolicy` is reported to be intermittent.

> Observed in this CI run: https://github.com/apache/ozone/pull/2179/checks?check_run_id=2423463857

> Bundle is attached to issue.

It seems that #1885 introduced a few problems. (It was new to me, so I try to explain here the changes)

```
-data.incrPendingDeletionBlocks(random.nextInt(numContainers) + 1);
+data.incrPendingDeletionBlocks(20);
```
It's better to use fixed size of containers as it's very hard to predict the test behaviors without that.

```
 for (int i = 0; i < result1.size(); i++) {
```

That's the root cause of the problem. As we have block size based selection, the container size may be less than the requested block size.

```
List<ContainerBlockInfo> result1 = blockDeletingService
              .chooseContainerForBlockDeletion(50, deletionPolicy);
```

Here we should use block number s(50) instead of the container numbers (was 10 earlier)

```
for (int j = 0; j < 100; j++) {
```

As result1 and result2 returns with 3 containers it's better to check multiple times the randomness.

https://issues.apache.org/jira/browse/HDDS-5147

## How was this patch tested?

Unit test is passed 7000 times locally. Without the patch it fails during the first 50-100 executions.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/statemachine/background/BlockDeletingService.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/statemachine/background/BlockDeletingService.java
@@ -177,8 +177,6 @@ public class BlockDeletingService extends BackgroundService {
             .filter(e -> isDeletionAllowed(e.getValue().getContainerData(),
                 deletionPolicy)).collect(Collectors
             .toMap(Map.Entry::getKey, e -> e.getValue().getContainerData()));
-    System.out.println("size " + containerDataMap.size());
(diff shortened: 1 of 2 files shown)


errose28:
Thanks for the quick fix Marton! This LGTM +1.

lokeshj1703 (approved):
@elek Thanks for working on this! The changes look good to me. +1.

lokeshj1703:
@elek Thanks for the contribution! @errose28 Thanks for the review! I have committed the PR to master branch.

PR #2189, opened 2021-04-27
HDDS-5152. Fix Suggested leader in Client.

bharatviswa504:

When hostname is having "." the regex is not matching properly with suggested leader and fail over is failing with below exception.
```
Caused by: java.util.NoSuchElementException: No value present
	at java.util.Optional.get(Optional.java:135)
	at org.apache.hadoop.hdds.scm.proxy.SCMBlockLocationFailoverProxyProvider.performFailoverToAssignedLeader(SCMBlockLocationFailoverProxyProvider.java:166)
	at org.apache.hadoop.hdds.scm.proxy.SCMBlockLocationFailoverProxyProvider$1.shouldRetry(SCMBlockLocationFailoverProxyProvider.java:273)
	at 
```

https://issues.apache.org/jira/browse/HDDS-5152

## How was this patch tested?

Added tests

### DIFF
--- a/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/ratis/TestServerNotLeaderException.java
+++ b/hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/ratis/TestServerNotLeaderException.java
@@ -61,6 +61,22 @@ public class TestServerNotLeaderException {
     snle = new ServerNotLeaderException(message);
     Assert.assertEquals(null,
         snle.getSuggestedLeader());
+
+    message = "Server:7fdd7170-75cc-4e11-b343-c2657c2f2f39 is not the " +
+        "leader.Suggested leader is Server:localhost:98634:8988 \n" +
+        "at org.apache.hadoop.hdds.ratis.ServerNotLeaderException" +
+        ".convertToNotLeaderException(ServerNotLeaderException.java:96)";
+    snle = new ServerNotLeaderException(message);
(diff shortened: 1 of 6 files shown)


hadoop-hdds/common/src/test/java/org/apache/hadoop/hdds/ratis/TestServerNotLeaderException.java
```diff
@@ -0,0 +1,58 @@
...
+    snle = new ServerNotLeaderException(message);
+    Assert.assertEquals("scm5-3.scm5.root.hwx.site:9863",
+        snle.getSuggestedLeader());
+
+    // Test hostname with out "."
+    message = "Server:7fdd7170-75cc-4e11-b343-c2657c2f2f39 is not the " +
+        "leader.Suggested leader is Server:localhost:98634 \n" +
+        "at org.apache.hadoop.hdds.ratis.ServerNotLeaderException" +
+        ".convertToNotLeaderException(ServerNotLeaderException.java:96)";
+    snle = new ServerNotLeaderException(message);
```
bshashikant, line 53: can we add a msg where suggested leader is specified as ":port" ?

bharatviswa504: Added it.

bshashikant (approved):

bharatviswa504:

PR #2191, opened 2021-04-28
HDDS-5155. Add revokeCertificate to SCMSecurityProtocolServer.

xiaoyuyao:

Expose Revoke API at Hadoop RPC layer. 

https://issues.apache.org/jira/browse/HDDS-5155

## How was this patch tested?

UT has coverage on the revoke processing layer. Will add acceptance test when the CLI for revoke is added in next patch.


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMSecurityProtocolServer.java
```diff
@@ -333,6 +341,30 @@ public long getLatestCrlId() {
...
 
+  @Override
+  public long revokeCertificates(List<Long> certIds, int reason,
+      long revocationTime) throws IOException {
+    String remoteUser = getRpcRemoteUsername();
+    storageContainerManager.checkAdminAccess(remoteUser);
+
+    Future<Optional<Long>> revoked = scmCertificateServer.revokeCertificates(
+        certIds.stream().map(id -> BigInteger.valueOf(id))
+            .collect(Collectors.toList()),
```
vivekratnavel, line 352: Is there a reason why we need to keep this API signature `List<Long> certIds`? If we change it to `List<BigInteger> certIds`, we will not need this Long -> BigInt conversion

xiaoyuyao: Good question. Most of the caller and existing APIs use String for certid and convert it to BigInteger when talking to the store layer. We can follow that and covert it only once (String->BigInteger) when talking to the store layer.

vivekratnavel (approved):
+1 LGTM.

PR #2192, opened 2021-04-28
HDDS-5150 when datanode restart,reload the container data will check the chunksfile

cchenax:
when datanode restart,reload the container data will check the chunksfile

https://issues.apache.org/jira/projects/HDDS/issues/HDDS-5150?filter=addedrecently

## How was this patch tested?


adoroszlai (commented):
Thanks @cchenax for proposing this improvement.  However, I don't think metadata load should be prevented just because chunk dir is missing.  The DB may be useful anyway, to find out exactly what is missing.  Also, returning early from this method with a log message would just hide errors.

Ozone already has a tool (container scanner) to detect missing data and more problems.

errose28:
Looks like this is now covered by #3052

PR #2193, opened 2021-04-28
HDDS-5159. Make periodic disk check interval configurable.

guihecheng:

Make datanode periodic disk check interval configurable, originally it is fixed 15min.

https://issues.apache.org/jira/browse/HDDS-5159

## How was this patch tested?

extended existing ut
manual test

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeConfiguration.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeConfiguration.java
@@ -41,8 +41,12 @@ public class DatanodeConfiguration {
       "hdds.datanode.replication.streams.limit";
   static final String CONTAINER_DELETE_THREADS_MAX_KEY =
       "hdds.datanode.container.delete.threads.max";
+  static final String PERIODIC_DISK_CHECK_INTERVAL_MINUTES_KEY =
+      "hdds.datanode.periodic.disk.check.interval.minutes";
 
   static final int REPLICATION_MAX_STREAMS_DEFAULT = 10;
+
+  static final long PERIODIC_DISK_CHECK_INTERVAL_MINUTES_DEFAULT = 15;
 
   /**
    * The maximum number of replication commands a single datanode can execute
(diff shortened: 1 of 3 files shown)


guihecheng:
@ChenSammi @bshashikant could you please help review this one too?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/MutableVolumeSet.java
```diff
@@ -123,14 +124,19 @@ public MutableVolumeSet(String dnUuid, String clusterID,
...
+    long periodicDiskCheckIntervalMinutes =
+        dnConf.getPeriodicDiskCheckIntervalMinutes();
     this.periodicDiskChecker =
       diskCheckerservice.scheduleWithFixedDelay(() -> {
         try {
           checkAllVolumes();
         } catch (IOException e) {
           LOG.warn("Exception while checking disks", e);
         }
-      }, DISK_CHECK_INTERVAL_MINUTES, DISK_CHECK_INTERVAL_MINUTES,
```
avijayanhwx, line 133: @guihecheng Are there other usages of DISK_CHECK_INTERVAL_MINUTES? If not, can that constant be removed as well?

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeConfiguration.java
```diff
@@ -110,6 +114,15 @@ public void setBlockDeletionLimit(int limit) {
     this.blockLimitPerInterval = limit;
   }
 
+  @Config(key = "periodic.disk.check.interval.minutes",
+      defaultValue = "15",
+      type = ConfigType.LONG,
```
avijayanhwx, line 119: Do we need a LONG? Can we just use an INT or TIME instead?

avijayanhwx (commented):
Thanks for working on this @guihecheng. Quick comments inline.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/volume/MutableVolumeSet.java
guihecheng: Sure, its private, will remove it.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/statemachine/DatanodeConfiguration.java
guihecheng: @avijayanhwx I think it's because the method `scheduleWithFixedDelay` requires a LONG parameter, if we make the config item INT, then we may have some convertions (INT -> LONG).

guihecheng:
@avijayanhwx updated with the unused const removed, thanks~

avijayanhwx (approved):

avijayanhwx:
Thank you for working on this @guihecheng.

PR #2195, opened 2021-04-29
HDDS-5166. Remove duplicate assignment of OZONE_OPTS for freon and sh

symious:

The option of "OZONE_OPTS" was assigned twice for Freon and SH, so the error of "duplicate jdwp options" prompts. Need the remove the duplicate assignment of "OZONE_OPTS".

https://issues.apache.org/jira/browse/HDDS-5166

## How was this patch tested?

Manually test. The following command works normal now.
```
 export OZONE_FREON_OPTS=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5011
 bin/ozone freon randomkeys --num-of-volumes=1 --num-of-buckets 1 --num-of-keys 10  --replication-type=RATIS --factor=THREE

 export OZONE_SH_OPTS=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5011
 bin/ozone sh volume ls /
```

### DIFF
--- a/hadoop-ozone/dist/src/shell/ozone/ozone
+++ b/hadoop-ozone/dist/src/shell/ozone/ozone
@@ -132,7 +132,6 @@ function ozonecmd_case
     freon)
       OZONE_CLASSNAME=org.apache.hadoop.ozone.freon.Freon
       OZONE_FREON_OPTS="${OZONE_FREON_OPTS}"
-      OZONE_OPTS="${OZONE_OPTS} ${OZONE_FREON_OPTS}"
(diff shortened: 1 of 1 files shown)


adoroszlai (approved):
Thanks @symious for reporting the problem and fixing it.

Debug output (enabled by `OZONE_SHELL_SCRIPT_DEBUG=true`) also confirms the duplication and the fix (`DEBUG: Final OZONE_OPTS: ...`).

symious:
@adoroszlai Thanks for the review.

PR #2196, opened 2021-04-29
HDDS-5149 when source datanode download container tar from target datanode,but the target datanode container file missing,import error

cchenax:
fix if the metadatafile or chunksfile missing import export container error

https://issues.apache.org/jira/projects/HDDS/issues/HDDS-5149?filter=addedrecently

## How was this patch tested?
unittest

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/TarContainerPacker.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/TarContainerPacker.java
@@ -53,7 +53,6 @@ import static java.util.stream.Collectors.toList;
  */
 public class TarContainerPacker
     implements ContainerPacker<KeyValueContainerData> {
-
   private static final Logger LOG =
       LoggerFactory.getLogger(TarContainerPacker.class);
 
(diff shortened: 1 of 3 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/keyvalue/TarContainerPacker.java
```diff
@@ -152,6 +156,20 @@ public void pack(Container<KeyValueContainerData> container,
...
+      } else if (!new File(containerData.getChunksPath()).exists()) {
+        LOG.warn("Chunkfile {} not exist",
+            containerData.getDbFile().toPath().toString());
+        return;
+      } else if (!container.getContainerFile().exists()) {
+        LOG.warn("Containerfile {} not exist",
+            containerData.getDbFile().toPath().toString());
+        return;
+      }
+
```
adoroszlai, line 172: Would it make sense to utilize `Container.scanMetadata()` instead?  More complete checks without introducing duplication.

hadoop-hdds/container-service/src/test/java/org/apache/hadoop/ozone/container/keyvalue/TestKeyValueContainer.java
```diff
@@ -238,6 +239,55 @@ public void testContainerImportExport() throws Exception {
...
+    String hddsVolumeDir = containerVolume.getHddsRootDir().toString();
+
+    container.populatePathFields(scmId, containerVolume, hddsVolumeDir);
+    long bytes = Files.size(folderToExport.toPath());
+    Assert.assertTrue(bytes <= 45);
+
+    try (FileInputStream fis = new FileInputStream(folderToExport)) {
+      container.importContainerData(fis, packer);
+    } catch (Exception ex) {
+      assertTrue(ex instanceof NullPointerException);
```
adoroszlai, line 287: * This passes even if `importContainerData` does not throw, which is not expected.
* Some exception other than `NullPointerException`  may better describe the error.
* `Assert.assertThrows` or `LambdaTestUtils.intercept` could provide more details if the assertion fails (eg. if not the expected kind of exception was thrown).

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/DownloadAndImportReplicator.java
```diff
@@ -121,10 +121,17 @@ public void replicate(ReplicationTask task) {
...
-
-        importContainer(containerID, path);
-        LOG.info("Container {} is replicated successfully", containerID);
-        task.setStatus(Status.DONE);
+        // if tar is null, the tar size is 45 bytes
+        if (bytes <= 45) {
+          task.setStatus(Status.FAILED);
+          LOG.warn("Container {} is downloaded with size {}, " +
+              "if size less than 45 bytes the tar file is null",
+              containerID, bytes);
```
adoroszlai, line 129: I think these kind of tar-specific details should be handled by `TarContainerPacker`.

adoroszlai (changes requested):
Thanks @cchenax for working on this.  The idea that container export/import may perform more sanity checks sounds good to me.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/replication/DownloadAndImportReplicator.java
cchenax: hello,I think it should be check in the DownloadAndImportReplicator,because of  the source datanode will through outputstream to the tar,if tar is null,the task status should be set failed,the tar size check in TarContainerPacker I think is not useful.

adoroszlai: Please consider the case if someone wanted to introduce a new implementation of `ContainerPacker`, eg. based on zip.  This check would not make sense, or at least would need different number of minimum bytes, for the other implementation.

adoroszlai:
Continued in #2376.

PR #2065, opened 2021-03-19
HDDS-5004. Upgrade Jersey2 dependency

adoroszlai:
## What changes were proposed in this pull request?

Upgrade Jersey 2 to most recent release for security.  2.33 makes JAX-B optional, so we need to explicitly declare it as dependency.

https://issues.apache.org/jira/browse/HDDS-5004

## How was this patch tested?

CI:
https://github.com/adoroszlai/hadoop-ozone/actions/runs/668462717

PR #2199, opened 2021-04-29
HDDS-5144. Create github check to alert when dependency tree is changed

elek:
This is the fixed version (handles ratis snapshots) of #2177 

JIRA: https://issues.apache.org/jira/browse/HDDS-5144

`hadoop-ozone/dist/src/main/license/bin/LICENSE.txt` and `hadoop-ozone/dist/src/main/license/bin/LICENSE.txt` should be modified when the transitive dependencies are changed.

However, it's very easy to miss any dependency change when we bump versions. It's possible that we change only versions but we also introduce new jar files via transitive dependencies.

This happened (for example) in #2139 and #2065 

To avoid any mistake I suggest the same approach what we use the protobuf compatibility check:

1. commit the current list of the jar files to the repository
2. create a check which compares the actual value with the committed one

If new dependency is introduced, the LICENSE file can be updated and the committed jar list also can be updated.

## How was this patch tested?

Tested on my fork. 

Example of failing builds:

https://github.com/elek/ozone/runs/2407951331?check_suite_focus=true


hadoop-ozone/dist/src/main/license/update-jar-report.sh
```diff
@@ -30,4 +30,4 @@ cd "$OZONE_DIST_DIR"
 
 #sed expression removes the version. Usually license is not changed with version bumps
 #jacoco and test dependencies are excluded
-find . -type f -name "*.jar" | cut -c3- | perl -wpl -e 's/-[0-9]+(.[0-9]+)*(-SNAPSHOT)?+//g' | grep -v jacoco | grep -v hadoop-hdds-test-utils | sort > "$SCRIPTDIR"/$REPORT_NAME
+find . -type f -name "*.jar" | cut -c3- | perl -wpl -e 's/-[0-9]+(.[0-9]+)*(-SNAPSHOT)?+//g' | perl -wpl -e 's/[0-9a-z]+-SNAPSHOT//g' | grep -v jacoco | grep -v hadoop-hdds-test-utils | sort > "$SCRIPTDIR"/$REPORT_NAME
```
adoroszlai, line 33: Nit: the hash in snapshot jars could be handled by replacing `(-SNAPSHOT)?` with `(-([0-9a-f]+-)?SNAPSHOT)?`.  One less `perl` command, no ugly `-` suffix in jar names.

adoroszlai (approved):

hadoop-ozone/dist/src/main/license/update-jar-report.sh
elek: Good idea, thanks for the suggestion. Pushed the fix in b9b7dae2c, will merge it once we have a green build.

adoroszlai:
Thanks @elek for updating the patch.  Merging it after a green build.

PR #2200, opened 2021-04-29
HDDS-5165. OM DB checkpoint servlet not accessible in a secure cluster

vivekratnavel:

Update OM DB checkpoint servlet to enforce access restrictions only when both ACL and SPNGEO auth are enabled. Also, updated the rules to allow recon by default.

https://issues.apache.org/jira/browse/HDDS-5165

## How was this patch tested?

Unit tests and tested in a real cluster.

### DIFF
--- a/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/DBCheckpointServlet.java
+++ b/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/DBCheckpointServlet.java
@@ -33,6 +33,8 @@ import java.util.Collection;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.apache.hadoop.hdds.conf.OzoneConfiguration;
+import org.apache.hadoop.hdds.recon.ReconConfig;
 import org.apache.hadoop.hdds.utils.db.DBCheckpoint;
 import org.apache.hadoop.hdds.utils.db.DBStore;
 
@@ -48,6 +50,7 @@ import org.apache.commons.lang3.StringUtils;
 import static org.apache.hadoop.ozone.OzoneConfigKeys.OZONE_ADMINISTRATORS_WILDCARD;
 import static org.apache.hadoop.ozone.OzoneConsts.OZONE_DB_CHECKPOINT_REQUEST_FLUSH;
 
+import org.apache.hadoop.security.UserGroupInformation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -64,10 +67,13 @@ public class DBCheckpointServlet extends HttpServlet {
   private transient DBCheckpointMetrics dbMetrics;
 
   private boolean aclEnabled;
+  private boolean isSpnegoEnabled;
   private Collection<String> ozAdmins;
+  private String reconSPN;
 
   public void initialize(DBStore store, DBCheckpointMetrics metrics,
-      boolean omAclEnabled, Collection<String> ozoneAdmins)
+                         boolean omAclEnabled, Collection<String> ozoneAdmins,
+                         boolean isSpnegoEnabled, OzoneConfiguration conf)
       throws ServletException {
 
     dbStore = store;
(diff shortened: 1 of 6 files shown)


hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMDbCheckpointServlet.java
```diff
@@ -83,114 +89,184 @@
...
+        getNumCheckpoints() > initialCheckpointCount);
+  }
+
+  @Test
+  public void testSpnegoEnabled() throws Exception {
+    cluster.shutdown();
+    conf.setBoolean(OZONE_ACL_ENABLED, true);
+    conf.set(OZONE_ADMINISTRATORS, "");
+    conf.set(OZONE_OM_HTTP_AUTH_TYPE, "kerberos");
+    conf.set(OZONE_RECON_KERBEROS_PRINCIPAL_KEY, "recon/host1@REALM");
```
avijayanhwx, line 202: There seems to be some common code between the 2 tests. Can they be moved to common method?

avijayanhwx (commented):
Approach looks OK to me. 1 comment inline. Can we fix the findbugs issue?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/DBCheckpointServlet.java
```diff
@@ -79,14 +86,27 @@ public void initialize(DBStore store, DBCheckpointMetrics metrics,
...
     this.ozAdmins = ozoneAdmins;
+    this.isSpnegoEnabled = isSpnegoEnabled;
+    ReconConfig reconConfig = conf.getObject(ReconConfig.class);
+    String reconPrincipal = reconConfig.getKerberosPrincipal();
+    if (!reconPrincipal.isEmpty()) {
+      UserGroupInformation ugi =
+          UserGroupInformation.createRemoteUser(reconPrincipal);
+      reconSPN = ugi.getShortUserName();
+    } else {
+      reconSPN = StringUtils.EMPTY;
```
xiaoyuyao, line 97: Do we have guarantee that SCM configuration always have recon related configuration like recon SPN?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/DBCheckpointServlet.java
```diff
@@ -64,10 +67,14 @@
...
 
   private boolean aclEnabled;
+  private boolean isSpnegoEnabled;
   private Collection<String> ozAdmins;
+  private String reconSPN;
 
+  @SuppressWarnings("checkstyle:hiddenfield")
   public void initialize(DBStore store, DBCheckpointMetrics metrics,
-      boolean omAclEnabled, Collection<String> ozoneAdmins)
+                         boolean omAclEnabled, Collection<String> ozoneAdmins,
```
xiaoyuyao, line 76: Can we refactor the initialize() with just a Collection<String> allowedUsers? and let the callers to combine ozoneAdmin and recon SPN without additional checkstyle warning suppress?

xiaoyuyao (commented):
Thanks @vivekratnavel for working on this. The PR looks good to me overall. A few minor suggestions added inline...

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMDbCheckpointServlet.java
```diff
@@ -83,114 +93,171 @@
...
+    Assert.assertTrue(
+        omMetrics.getDBCheckpointMetrics().
+            getLastCheckpointStreamingTimeTaken() > 0);
+    Assert.assertTrue(omMetrics.getDBCheckpointMetrics().
+        getNumCheckpoints() > initialCheckpointCount);
+  }
+
+  @Test
+  public void testSpnegoEnabled() throws Exception {
+    cluster.shutdown();
```
xiaoyuyao, line 212: Should we have cluster.shutdown() in testDoGet() to ensure we don't leak cluster instances running if the test execution order is changed?

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMDbCheckpointServlet.java
```diff
@@ -83,114 +93,171 @@
...
+    when(userPrincipalMock.getName()).thenReturn("recon/localhost@REALM");
+
+    when(requestMock.getUserPrincipal()).thenReturn(userPrincipalMock);
+    when(responseMock.getOutputStream()).thenReturn(servletOutputStream);
+
+    omDbCheckpointServletMock.doGet(requestMock, responseMock);
 
+    // Recon user should be able to access the servlet and download the
+    // snapshot
+    Assert.assertTrue(tempFile.length() > 0);
```
xiaoyuyao, line 260: the test cluster instance needs to be shutdown at the end of the test.

xiaoyuyao (commented):
Thanks for the update. I have one comment on the unit test cluster shutdown. Otherwise, LGTM.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMDbCheckpointServlet.java
vivekratnavel: Yes, it is handled in the `@AfterClass` method - https://github.com/apache/ozone/pull/2200/files#diff-816030b8fda0e307628b659606cf1d37b2117996c83caab10513b1956f72bc13R128

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMDbCheckpointServlet.java
vivekratnavel: Yes, it is handled in the `@AfterClass` method - https://github.com/apache/ozone/pull/2200/files#diff-816030b8fda0e307628b659606cf1d37b2117996c83caab10513b1956f72bc13R128

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMDbCheckpointServlet.java
xiaoyuyao: That should be sufficient for the current two test cases. My suggestion is to move the shutdown with @After for each test case. This way, new test cases added won't need to put cluster.shutdown() any more.

xiaoyuyao (approved):
+1 pending CI.

hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOMDbCheckpointServlet.java
vivekratnavel: Thanks @xiaoyuyao! I have updated the latest patch to fix this.

vivekratnavel:
Thank you @xiaoyuyao and @avijayanhwx for the reviews.

PR #2201, opened 2021-04-30
HDDS-5137. Update licenses for non-rolling upgrade new dependencies.

guihecheng:

Update licenses for non-rolling upgrade new dependencies.

https://issues.apache.org/jira/browse/HDDS-5137

## How was this patch tested?

lincense files only, no test.


guihecheng:
@elek @avijayanhwx license files updated, thanks~

elek (approved):
+1, Thanks the patch @guihecheng 

 * Neither of these are ASLv2 --> NOTICE.txt is not required to be changed
 * Only binary package (tar.gz) contains new jars --> only binary LICENSE should be changed
 * These are the only added jar files.

PR #2202, opened 2021-04-30
HDDS-5171. Make the report max limit configurable

wzhallright:

Make the report max limit configurable.

https://issues.apache.org/jira/browse/HDDS-5171

## How was this patch tested?
The original UT can cover this.


hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HddsServerUtil.java
```diff
@@ -325,6 +327,16 @@ public static int getScmRpcRetryCount(ConfigurationSource conf) {
...
+  /**
+   * Get the max limit when get all available reports.
+   * @param conf
+   * @return max limit
+   */
+  public static int getReportMaxLimit(ConfigurationSource conf) {
+    return conf.getInt(HDDS_DATANODE_REPORT_MAX_LIMIT,
+        HDDS_DATANODE_REPORT_MAX_LIMIT_DEFAULT);
+  }
+
```
adoroszlai, line 339: Nit: the name `getReportMaxLimit` seems too generic when placed in the utility class.  Can you please either move it to some datanode-specific class (e.g. `StateContext` itself), or rename it to be more specific?

adoroszlai (commented):
Thanks @wzhallright for proposing this change.  I have two questions:

1. What is the motivation (use case) to make the report limit configurable?
2. Adding the new config as an exception in `TestOzoneConfigurationFields` implies that it's a developer-only setting.  Is this correct, i.e. did you intend to hide it from end users?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HddsServerUtil.java
wzhallright: Thanks @adoroszlai for review.
`public List<GeneratedMessage> getReports(InetSocketAddress endpoint,
                                           int maxLimit) {
    if (maxLimit < 0) {
      throw new IllegalArgumentException("Illegal maxLimit value: " + maxLimit);
    }
    List<GeneratedMessage> reports = getNonIncrementalReports();
    if (maxLimit <= reports.size()) {
      return reports.subList(0, maxLimit);
    } else {
      reports.addAll(getIncrementalReports(endpoint,
          maxLimit - reports.size()));
      return reports;
    }
  }`
1) In this code, the max reports size of getNonIncrementalReports() return is 3, if the maxLimit is Integer.MAX_VALUE, so `maxLimit < 0` and `maxLimit <= reports.size()` is meaningless.
I think the maxLimit need to become configurable.
2)I will add it to ozone-default.xml instead of TestOzoneConfigurationFields.
3)Move getReportMaxLimit to StateContext.java.

wzhallright:
@adoroszlai would you help to take another look again?

hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/HddsServerUtil.java
adoroszlai: Thanks for the answers, and updating the patch.

> 1. In this code, the max reports size of getNonIncrementalReports() return is 3, if the maxLimit is Integer.MAX_VALUE, so maxLimit < 0 and maxLimit <= reports.size() is meaningless.  I think the maxLimit need to become configurable.

The checks are not necessarily meaningless, may simply be defensive programming.  Currently the only production caller passes `Integer.MAX_VALUE`.  But the method is public, and new calls may be introduced where different limits might make sense.  In this case they can be kept without making the limit used by `StateContext` in `getReports()` configurable.

Or they may be unnecessary and could be simply removed.

I was looking for a use case of this configuration from user/administrator point of view, not based on the code.  If we don't know why it may be configured to different values (i.e. use cases), and how users should come up with those, then I think it may be better to avoid adding the config.

But that's just my 2 cents.

adoroszlai:
/pending a use case of this configuration from user/administrator point of view

PR #2204, opened 2021-04-30
HDDS-5170. Race condition in NodestateManager#addNode allows datanodes with lower MLV to be used in pipelines.

errose28:

Fix race condition in NodeStateManager#addNode that gave the background pipeline creator thread a small window of time to use newly added healthy readonly nodes in pipelines.

HDDS-5170

## How was this patch tested?

Condition was discovered because of flaky unit test `TestSCMNodeManager#testScmLayoutOnRegister`. Without the patch, the test fails after 15-30 consecutive runs. With the patch, I was able to do 400 consecutive runs and did not see any failures.

`TestSCMNodeManager#testScmLayoutOnHeartbeat` was also checked just to be safe. It is not flaky both before and after the patch based on 200 consecutive runs (which took about half an hour).


avijayanhwx (approved):
LGTM +1.

PR #2206, opened 2021-05-03
HDDS-5145. Extend Pipline/ReplicationConfig refactor with ECReplicationConfig

elek:

HDDS-5047 / #2096 started to use `ReplicationConfig` in `Pipeline` and pipeline related SCM service on master. `ReplicationConfig` was introduced in HDDS-5011, but it has two versions one for master #(2089) and one for the EC branch.(#2068)

Merging master after HDDS-5047 requires small modification in the code to make HDDS-5047 compatible the ec version of HDDS-5011:

During the proto serialization / deserialization we should use optional ECReplicationConfig (if exists).

https://issues.apache.org/jira/browse/HDDS-5145

## How was this patch tested?

With CI, The EC branch has compilation failure without this patch.


sodonnel (approved):
Changes LGTM. Thanks for fixing this.

PR #2207, opened 2021-05-03
HDDS-5176. Increase default block cache capacity for Datanodes

lokeshj1703:

Currently Ozone datanodes use a default block cache capacity of 64MB. We can increase the default block cache capacity to a higher value like 512MB..1GB.

https://issues.apache.org/jira/browse/HDDS-5176

## How was this patch tested?

Existing UT.


mukul1987 (approved):
+1, LGTM

PR #2214, opened 2021-05-04
HDDS-5184. Use separate DB profile for Datanodes.

lokeshj1703:

Currently datanodes shares a common DB profile with other components like SCM and OM etc. The jira aims to create a different profile for datanodes which also makes sure that column family options are shared by containers.

https://issues.apache.org/jira/browse/HDDS-5184

## How was this patch tested?

Existing UT, PR adds a UT to test columnfamilyoptions are shared.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/db/DatanodeDBProfile.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/db/DatanodeDBProfile.java
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


dineshchitlangia (approved):
+1, thanks for the improvement @lokeshj1703

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/utils/db/DatanodeDBProfile.java
```diff
@@ -0,0 +1,144 @@
...
+      if (config == null) {
+        return blockBasedTableConfig;
+      }
+
+      long cacheSize = (long) config
+          .getStorageSize(HDDS_DATANODE_METADATA_ROCKSDB_CACHE_SIZE,
+              HDDS_DATANODE_METADATA_ROCKSDB_CACHE_SIZE_DEFAULT,
+              StorageUnit.BYTES);
+      blockBasedTableConfig
+          .setBlockCache(new LRUCache(cacheSize * SizeUnit.MB));
```
mukul1987, line 140: we should not multiply by MB here ? as cache size is already in bytes ? :D

lokeshj1703: Added a UT in new commit.

bshashikant: @lokeshj1703 , why multiplying with MB is required? can you please explain?

lokeshj1703: This was a bug @bshashikant. I have changed it in latest commits.

jojochuang (approved):
LGTM
@mukul1987 anything other comments?

mukul1987 (approved):
+1, LGTM

PR #2218, opened 2021-05-06
HDDS-5186. If primordial SCM id is set, a non-HA cluster can not be initialized.

mukul1987:
SCM init should work when primordial node is specified in the config and SCM HA is not enabled.

https://issues.apache.org/jira/browse/HDDS-5186

## How was this patch tested?
Existing Tests


bharatviswa504 (approved):
+1 LGTM

bharatviswa504:
Thank You @mukul1987 for the contribution.

PR #2219, opened 2021-05-06
HDDS-5191. Increase default pvc storage size

yoowonsuk:

Containers are 5Gb by default, however pvc is 2Gb.
it needs to be changed.

HDDS-5191

## How was this patch tested?

manual tests

### DIFF
--- a/hadoop-ozone/dist/src/main/k8s/definitions/ozone/definitions/persistence.yaml
+++ b/hadoop-ozone/dist/src/main/k8s/definitions/ozone/definitions/persistence.yaml
@@ -29,7 +29,7 @@ description: Add real PVC based persistence
           accessModes: [ "ReadWriteOnce" ]
           resources:
             requests:
-              storage: 2Gi
+              storage: 20Gi
 - type: Add
   trigger:
      metadata:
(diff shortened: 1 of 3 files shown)


elek (commented):
Thank you to report this issue @yoowonsuk 

I agree with proposed change but having two comments:

1. I think you can further increase this storage to 15Gi or 20Gi. If you use a storage where the space is really limited 5Gi may not be enough (especially as datanode can complain that there is not enough space for the containers if more than one container is opened at the same time).

 2. Can you please change all the related files:
 
 ```
examples/ozone/datanode-statefulset.yaml
72:          storage: 2Gi

definitions/ozone/definitions/persistence.yaml
32:              storage: 2Gi

examples/ozone-ha/datanode-statefulset.yaml
72:          storage: 2Gi
 ```
 
 (`examples` files are generated from the definitions, but you shouldn't be worried just update both the definition and examples)

elek (approved):
+1. Thanks for the update @yoowonsuk

elek:
Merging it now. Thanks for the contribution @yoowonsuk

PR #2224, opened 2021-05-08
HDDS-5173. Divide snapshot related work into notifyInstallSnapshotFromLeader and reinitialize for SCMStateMachine.

GlenGeng:

As mentioned in https://issues.apache.org/jira/browse/RATIS-1370
During `notifyInstallSnapshotFromLeader`, `StateMachineUpdater` may call `applyTransactions` when StateMachine is in PAUSED state.
Just divide snapshot related work into `notifyInstallSnapshotFromLeader` and `reinitialize` for `SCMStateMachine`.

During `notifyInstallSnapshotFromLeader`, SCM just downloads snapshot but not modify StateMachine, since StateMachineUpdater is in RUNNING state, may call applyTransactions during this period.
During `reinitialize`, SCM can safely reload the StateMachine, such as rocksdb and in-memory state. During this period, StateMachineUpdater is in RELOAD state, thus there will be no contention between SCM and StateMachineUpdater.

https://issues.apache.org/jira/browse/HDDS-5173

## How was this patch tested?

CI

### DIFF
--- a/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
+++ b/hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
@@ -27,12 +27,14 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.InvalidProtocolBufferException;
 import org.apache.hadoop.hdds.scm.block.DeletedBlockLog;
 import org.apache.hadoop.hdds.scm.block.DeletedBlockLogImplV2;
 import org.apache.hadoop.hdds.scm.server.StorageContainerManager;
 import org.apache.hadoop.hdds.utils.TransactionInfo;
+import org.apache.hadoop.hdds.utils.db.DBCheckpoint;
 import org.apache.hadoop.util.concurrent.HadoopExecutors;
 import org.apache.ratis.proto.RaftProtos;
 import org.apache.hadoop.hdds.scm.exceptions.SCMException;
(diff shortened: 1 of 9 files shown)


GlenGeng:
@bharatviswa504  @bshashikant could you please take a look at this PR ?

GlenGeng:
Hey @adoroszlai , a question about CI result, why the last two commits are passed, but the `checks` is not ?

adoroszlai:
> a question about CI result, why the last two commits are passed, but the `checks` is not ?

Do you mean why the checks passed on your branch, but not in the PR?  It can happen for two reasons:

1. PR checks are run against the "merge" commit, which includes recent changes from the target branch (`master`).  This source of difference can be eliminated (temporarily) by merging `master` into your branch, instead of triggering CI with empty commit.
2. Flaky tests.

I think in this case it is the second one.  `TestRatisPipelineProvider` has been failing occasionally, and it passes for me locally on your branch, as well as after a merge from `master`.

GlenGeng:
Thanks @adoroszlai for the explanation. I will rebase the dev branch with latest master, to see if `TestRatisPipelineProvider ` can pass on the branch.

GlenGeng:
> Please prefer merge over rebase for open PRs. 

Sure. Will take care of this in future.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/ha/SCMStateMachine.java
```diff
@@ -76,6 +78,11 @@
   private final boolean isInitialized;
   private ExecutorService installSnapshotExecutor;
 
+  // The atomic variable RaftServerImpl#inProgressInstallSnapshotRequest
+  // ensures serializable between notifyInstallSnapshotFromLeader()
+  // and reinitialize().
+  private DBCheckpoint installingDBCheckpoint;
+
```
bshashikant, line 85: Once, the installSnaoshot completes, i guess, we need to set the "IntsallingCheckpoint" filed to null. Also, before the downloadSnaoshot icall is invoked, we should add a precondition check that installingCheckpoint is null always.

bshashikant (commented):
Looks good overall.

PR #2230, opened 2021-05-10
HDDS-4927. Determine over and under utilized datanodes in Container Balancer.

siddhantsangwan:

`ContainerBalancer` will identify over, under, and within threshold utilized nodes at the start of each iteration. Based on this, it will determine whether balancing should continue. Unit tests to test this functionality.

https://issues.apache.org/jira/browse/HDDS-4927

## How was this patch tested?

Added unit test `TestContainerBalancer`


hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -59,74 +80,305 @@ public ContainerBalancer(
...
+        overUtilizedNodes.size() + underUtilizedNodes.size());
+
+    if (numDatanodesBalanced + numDatanodesToBalance > maxDatanodesToBalance) {
+      LOG.info("Approaching Max Datanodes To Balance limit in Container " +
+          "Balancer. Stopping Balancer.");
+      stop();
+      return false;
+    } else {
+      sourceNodes.addAll(overUtilizedNodes);
+      sourceNodes.addAll(underUtilizedNodes);
```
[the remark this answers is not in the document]
linyiqun: >So here the term 'source nodes' has been used for nodes that need to be balanced. 

Okay, so the meaning of source node is a little different with before.

lokeshj1703:
@siddhantsangwan The PR shows commits from HDDS-4925 as well. Can you please take a look?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -59,74 +80,321 @@ public ContainerBalancer(
...
+    this.clusterRemaining = 0L;
+
+    this.overUtilizedNodes = new ArrayList<>();
+    this.underUtilizedNodes = new ArrayList<>();
+    this.aboveAverageUtilizedNodes = new ArrayList<>();
+    this.belowAverageUtilizedNodes = new ArrayList<>();
+    this.sourceNodes = new ArrayList<>();
+
     LOG.info("Starting Container Balancer...");
+    LOG.info(toString());
```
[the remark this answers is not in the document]
siddhantsangwan: Okay.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -59,74 +80,321 @@ public ContainerBalancer(
...
+
+  /**
+   * Initializes an iteration during balancing. Recognizes over, under,
+   * below-average,and under-average utilizes nodes. Decides whether
+   * balancing needs to continue or should be stopped.
+   *
+   * @return true if successfully initialized, otherwise false.
+   */
+  private boolean initializeIteration() {
+    List<DatanodeUsageInfo> nodes;
```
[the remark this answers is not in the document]
siddhantsangwan: Yes, changing that.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -59,74 +80,321 @@ public ContainerBalancer(
...
     // under utilized nodes have utilization(that is, used / capacity) less
     // than lower limit
-    double lowerLimit = avgUtilisation - threshold;
+    double lowerLimit = clusterAvgUtilisation - threshold;
 
     // over utilized nodes have utilization(that is, used / capacity) greater
     // than upper limit
-    double upperLimit = avgUtilisation + threshold;
+    double upperLimit = clusterAvgUtilisation + threshold;
+
```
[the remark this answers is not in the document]
siddhantsangwan: A negative lower limit and upper limit greater than 1 will not lead to errors(as checked by the unit test). But in that case we could return false early since the cluster is balanced.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -59,74 +80,321 @@ public ContainerBalancer(
...
+      nodes = nodeManager.getMostOrLeastUsedDatanodes(true);
+    } catch (NullPointerException e) {
+      LOG.error("Container Balancer could not retrieve nodes from Node " +
+          "Manager.", e);
+      stop();
+      return false;
+    }
+
+    try {
+      clusterAvgUtilisation = calculateAvgUtilization(nodes);
```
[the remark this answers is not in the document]
siddhantsangwan: Makes sense. But shouldn't we return false then?

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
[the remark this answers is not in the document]
GlenGeng: Question:

say we have a 10 DN cluster, the usage of all of them is 95%, then one empty DN is added to rebalance the cluster. Given the threshold is 10%, it seems the balancer will not work in this case, since that 10 DN will not achieve upperLimit.

Have we consider corner case like this ?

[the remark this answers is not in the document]
siddhantsangwan: In that case, the newly added DN will be under utilized. Balancer will recognize this and then try to move data into it from the other 10 DNs that are now above average utilized. However, the exact algorithm for choosing particular source DN and containers to move is under progress.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -59,74 +80,321 @@ public ContainerBalancer(
...
+  /**
+   * Performs binary search to determine if the specified listToSearch
+   * contains the specified node.
+   *
+   * @param listToSearch List of DatanodeUsageInfo to be searched.
+   * @param node DatanodeUsageInfo to be searched for.
+   * @return true if the specified node is present in listToSearch, otherwise
+   * false.
+   */
+  private boolean containsNode(
```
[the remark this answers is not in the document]
siddhantsangwan: That's a great suggestion. The current implementation is assuming that the order of nodes in terms of utilization is important so balancer can focus on nodes that are most over or under utilized first. That's why a sorted list is being used.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/balancer/TestContainerBalancer.java
```diff
@@ -0,0 +1,213 @@
...
+   */
+  private List<DatanodeUsageInfo> determineExpectedSourceNodes(
+      double threshold) {
+    double lowerLimit = averageUtilization - threshold;
+    double upperLimit = averageUtilization + threshold;
+
+    // use node utilizations to determine over and under utilized nodes
+    List<DatanodeUsageInfo> expectedSourceNodes = new ArrayList<>();
+    for (int i = 0; i < numberOfNodes; i++) {
+      if (nodeUtilizations.get(numberOfNodes - i - 1) > upperLimit) {
```
lokeshj1703, line 153: Can we use nodesInCluster directly to determine the source nodes?

lokeshj1703 (commented):
@siddhantsangwan Thanks for working on the PR! I have added a few comments inline.

siddhantsangwan:
Changed the term `source` nodes to `unBalanced` nodes to avoid ambiguity. Source node can mean 'node from which data is leaving' as expected. @linyiqun

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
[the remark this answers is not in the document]
siddhantsangwan: Good point.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -45,9 +48,27 @@
   private long maxSizeToMove;
   private boolean balancerRunning;
   private List<DatanodeUsageInfo> sourceNodes;
-  private List<DatanodeUsageInfo> targetNodes;
+  private List<DatanodeUsageInfo> overUtilizedNodes;
+  private List<DatanodeUsageInfo> underUtilizedNodes;
+  private List<DatanodeUsageInfo> aboveAverageUtilizedNodes;
+  private List<DatanodeUsageInfo> belowAverageUtilizedNodes;
```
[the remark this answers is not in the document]
siddhantsangwan: Replacing these with a single list `withinThresholdUtilizedNodes`.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/balancer/TestContainerBalancer.java
siddhantsangwan: By using a separate method, varying thresholds can be passed as argument to determine the expected list for each threshold.

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/balancer/TestContainerBalancer.java
```diff
@@ -0,0 +1,213 @@
...
+    replicationManager = Mockito.mock(ReplicationManager.class);
+
+    balancerConfiguration = new ContainerBalancerConfiguration();
+    balancerConfiguration.setThreshold("0.1");
+    balancerConfiguration.setMaxDatanodesToBalance(10);
+    balancerConfiguration.setMaxSizeToMove(500L);
+    conf.setFromObject(balancerConfiguration);
+
+    this.numberOfNodes = 10;
+    generateUtilizations(numberOfNodes);
```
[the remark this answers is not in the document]
siddhantsangwan: Done

hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/container/MockNodeManager.java
```diff
@@ -139,6 +137,37 @@ public MockNodeManager(boolean initializeFakeNodes, int nodeCount) {
         initializeFakeNodes, nodeCount);
   }
 
+  public MockNodeManager(List<DatanodeUsageInfo> nodes)
```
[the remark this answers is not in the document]
siddhantsangwan: Extracted initialization out to an initialization block.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -59,74 +80,321 @@ public ContainerBalancer(
...
+  }
+
+  /**
+   * Initializes an iteration during balancing. Recognizes over, under,
+   * below-average,and under-average utilizes nodes. Decides whether
+   * balancing needs to continue or should be stopped.
+   *
+   * @return true if successfully initialized, otherwise false.
+   */
+  private boolean initializeIteration() {
```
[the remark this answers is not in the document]
siddhantsangwan: Using `calculateAvgUtilization`.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -40,96 +44,341 @@
...
+    this.maxDatanodesToBalance = config.getMaxDatanodesToBalance();
     this.maxSizeToMove = config.getMaxSizeToMove();
 
-    LOG.info("Starting Container Balancer...");
+    this.clusterCapacity = 0L;
+    this.clusterUsed = 0L;
+    this.clusterRemaining = 0L;
+
+    this.overUtilizedNodes = new ArrayList<>();
+    this.underUtilizedNodes = new ArrayList<>();
```
JacksonYao287, line 114: maybe it is better to put these initialize operations in constructor, and the start function will just do the start work

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -205,11 +202,10 @@ private boolean initializeIteration() {
         numDatanodesBalanced += 1;
       }
     }
-
-    // calculate total number of nodes that have been balanced
+    // calculate total number of nodes that have been balanced so far
     numDatanodesBalanced =
-        numDatanodesBalanced + metrics.getNumDatanodesBalanced().get();
-    metrics.setNumDatanodesBalanced(new LongMetric(numDatanodesBalanced));
+        metrics.addToNumDatanodesBalanced(numDatanodesBalanced);
```
lokeshj1703, line 207: NIT: Change fn name to `incroNumDatanodesBalanced`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancerConfiguration.java
```diff
@@ -23,80 +23,95 @@
...
       "Balancer will not balance more number of datanodes than this limit.")
   private int maxDatanodesToBalance = 5;
 
-  @Config(key = "size.moved.max", type = ConfigType.LONG,
-      defaultValue = "10737418240L", tags = {ConfigTag.BALANCER},
-      description = "The maximum size of data in Bytes that will be moved " +
-          "by the Container Balancer.")
-  private long maxSizeToMove = 10737418240L;
+  @Config(key = "size.moved.max", type = ConfigType.SIZE,
+      defaultValue = "10", tags = {ConfigTag.BALANCER},
```
lokeshj1703, line 53: Please see how ConfigType.SIZE is used in other places. I think the default value should be `10GB`. It is more flexible config type and it will not be limited to GB alone. The description and javadoc for corresponding fns can be updated to remove specificity to GB.

lokeshj1703 (commented):
@siddhantsangwan Thanks for updating the PR! I have added a few more minor comments based on recent changes.

siddhantsangwan:
Thanks for the reviews. I have addressed the comments.

lokeshj1703 (commented):
@siddhantsangwan Thanks for updating the PR! The changes look good to me. +1. 

Can you create another jira for https://github.com/apache/ozone/pull/2230#discussion_r630797680 or handle it in next jira?

siddhantsangwan:
> @siddhantsangwan Thanks for updating the PR! The changes look good to me. +1.
> 
> Can you create another jira for [#2230 (comment)](https://github.com/apache/ozone/pull/2230#discussion_r630797680) or handle it in next jira?

I have handled that case in the `containsNode` method. Do you mean changing the comparator itself?

lokeshj1703:
If hypothetically you have 5 nodes with same utilisation then containsNode might not return true with the current change. Because binary search can return index of any node with same utilisation.

`return index >= 0 && listToSearch.get(index).equals(node);`

This logic would return false since the returned datanode might have a different id.

Comparator would need to handle the case of same utilisation.

siddhantsangwan:
> Because binary search can return index of any node with same utilisation.

Yes, this case can be handled in another Jira. Thanks for pointing this out.

siddhantsangwan:
@JacksonYao287 and @linyiqun can you please review the changes? Any comments are welcome.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancerMetrics.java
```diff
@@ -0,0 +1,127 @@
...
+    "Container Balancer running in SCM", context = "SCM")
+public final class ContainerBalancerMetrics {
+
+  @Metric(about = "The total amount of used space in GigaBytes that needs to " +
+      "be balanced.")
+  private LongMetric totalSizeToBalanceGB;
+
+  @Metric(about = "The amount of Giga Bytes that have been moved to achieve " +
+      "balance.")
+  private LongMetric gigaBytesMoved;
```
GlenGeng, line 36: NIT: `gigaBytesMoved` to `dataSizeBalancedGB`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancerMetrics.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Metric(about = "Number of containers that Container Balancer has moved" +
+      " until now.")
+  private LongMetric numContainersMoved;
```
GlenGeng, line 40: NIT: `numContainersMoved` to `movedContainerNum`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancerMetrics.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Metric(about = "The total number of datanodes that need to be balanced.")
+  private LongMetric totalNumDatanodesToBalance;
```
GlenGeng, line 43: NIT: `totalNumDatanodesToBalance` to `datanodeNumToBalance`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancerMetrics.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Metric(about = "Number of datanodes that Container Balancer has balanced " +
+      "until now.")
+  private LongMetric numDatanodesBalanced;
```
GlenGeng, line 47: NIT: `numDatanodesBalanced` to `datanodeNumBalanced`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancerMetrics.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Metric(about = "Utilisation value of the current maximum utilised datanode.")
+  private double maxUtilizedDatanodeRatio;
```
GlenGeng, line 50: NIT: `maxUtilizedDatanodeRatio` to `maxDatanodeUtilizedRatio`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancerMetrics.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.hadoop.metrics2.annotation.Metric;
+import org.apache.hadoop.metrics2.annotation.Metrics;
+
+@Metrics(name = "ContainerBalancer Metrics", about = "Metrics related to " +
+    "Container Balancer running in SCM", context = "SCM")
+public final class ContainerBalancerMetrics {
+
+  @Metric(about = "The total amount of used space in GigaBytes that needs to " +
+      "be balanced.")
+  private LongMetric totalSizeToBalanceGB;
```
GlenGeng, line 32: NIT `totalSizeToBalanceGB` -> `dataSizeToBalanceGB`

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -40,96 +44,345 @@
...
+        stat.getCapacity().get().doubleValue();
   }
 
+  /**
+   * Stops ContainerBalancer.
+   */
   public void stop() {
     LOG.info("Stopping Container Balancer...");
-    balancerRunning = false;
+    balancerRunning.set(false);
```
GlenGeng, line 320: NIT. remove line 319, one info line is sufficient here, since no actual work needs to be done here.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -40,96 +44,345 @@
...
+  }
+
+  /**
+   * Calculates the utilization, that is used space divided by capacity, for
+   * the given datanodeUsageInfo.
+   *
+   * @param datanodeUsageInfo DatanodeUsageInfo to calculate utilization for
+   * @return Utilization value
+   */
+  private double calculateUtilization(DatanodeUsageInfo datanodeUsageInfo) {
```
GlenGeng, line 308: NIT. make it to be a static helper function.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -40,96 +44,345 @@
...
     }
-    return aggregatedStats.getScmUsed().get().doubleValue() /
-        aggregatedStats.getCapacity().get().doubleValue();
+    clusterCapacity = aggregatedStats.getCapacity().get();
+    clusterUsed = aggregatedStats.getScmUsed().get();
+    clusterRemaining = aggregatedStats.getRemaining().get();
+
+    try {
+      return clusterUsed / (double) clusterCapacity;
+    } catch (ArithmeticException e) {
```
GlenGeng, line 294: NIT. better not handle `ArithmeticException`, instead check `nodes.size() != 0` at the entrance of function.

GlenGeng (commented):
Thanks @siddhantsangwan for the work. Just some inline comments.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
siddhantsangwan: Updated PR. Please take a look.

GlenGeng:
LGTM. We can merge this PR for now, and start our future development based on it.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/container/balancer/ContainerBalancer.java
```diff
@@ -40,96 +44,347 @@
...
 
   /**
-   * Start ContainerBalancer. Current implementation is incomplete.
+   * Starts ContainerBalancer. Current implementation is incomplete.
    *
    * @param balancerConfiguration Configuration values.
    */
-  public void start(ContainerBalancerConfiguration balancerConfiguration) {
-    this.balancerRunning = true;
+  public boolean start(ContainerBalancerConfiguration balancerConfiguration) {
```
JacksonYao287, line 102: maybe it is better to move these configuration initialization operations to constructor, start just do the start work without any parameter

lokeshj1703: This was initially decided in order to support configuration change using restart. Admin could restart balancer and balancer was supposed to load the new configuration values.

lokeshj1703: @JacksonYao287 if there are other suggestions for this function we can take them up in #2278.  If it is ok I will commit this PR.

JacksonYao287: if we put container balancer inside SCM ,  the configuration is just the one loaded by scm.  so ,the configuration can not be reloaded unless restarting scm

GlenGeng: According to our design decision, params will be based from command line, which will be implemented by https://github.com/apache/ozone/pull/2278, thereby the conf related code will be removed in future.

lokeshj1703: I see. We will not need the configuration param in that case. I see that change is made in #2278.

lokeshj1703:
@siddhantsangwan Thanks for the contribution! @linyiqun @GlenGeng @JacksonYao287 Thanks for the reviews! I have committed the PR to master branch.

PR #2231, opened 2021-05-10
HDDS-5203. Allow suppressing deprecation warning for HADOOP_ variables

adoroszlai:

HDDS-4525 deprecated `HADOOP_*` variables in Ozone in favor of corresponding `OZONE_*` variables.  This change allows suppressing the deprecation warning by setting `OZONE_DEPRECATION_WARNING=false`.

https://issues.apache.org/jira/browse/HDDS-5203

## How was this patch tested?

Added new acceptance test case for `OZONE_DEPRECATION_WARNING=false`.

https://github.com/adoroszlai/hadoop-ozone/runs/2544900937#step:6:870

### DIFF
--- a/hadoop-ozone/dist/src/main/smoketest/cli/envvars.robot
+++ b/hadoop-ozone/dist/src/main/smoketest/cli/envvars.robot
@@ -66,6 +66,21 @@ Picks up deprecated vars if valid
                         Should contain   ${output}   WARNING: HADOOP_HOME
                         Should contain   ${output}   WARNING: HADOOP_CONF_DIR
 
+Warning for deprecated vars can be suppressed
+    Set Environment Variable       OZONE_DEPRECATION_WARNING    false
+    Set Environment Variable       HADOOP_HOME           /opt/hadoop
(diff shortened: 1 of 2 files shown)


dineshchitlangia:
@adoroszlai Why do we need to suppress the warning ? How does it make the upgrade easier?
My understanding is that since this is a warning it should cause any failures/error/process termination.

adoroszlai:
@dineshchitlangia Thanks for taking a look.

> How does it make the upgrade easier?  My understanding is that since this is a warning it should cause any failures/error/process termination.

Correct, it's only a warning, does not cause any failure.

We deprecated the `HADOOP_*` variables in favor of `OZONE_*` ones mainly to be able to define different values for a co-existing Hadoop + Ozone installation.  In an existing Ozone install where one doesn't need that flexibility, we can live with `HADOOP_*`, but the warning printed for each CLI command may be annoying anyway.  The ability to suppress the warning makes upgrade easier because otherwise you would have to rename all the other variables just to get rid of the warning.

PR #2233, opened 2021-05-10
HDDS-5207. Intermittent failure in TestRatisPipelineProvider#testCreatePipelineWithFactorThree

adoroszlai:

`testCreatePipelineWithFactorThree` fails intermittently with two 3-nodes pipeline being assigned to the same set of nodes, while only a single pipeline per datanode is allowed.  The problem is that `MockNodeManager` ignores this config and has a separate variable, which tests need to set.  Similar setting in other test:

https://github.com/apache/ozone/blob/8e8a01dfd20899cd9dc95ab3de744b636a807ab2/hadoop-hdds/server-scm/src/test/java/org/apache/hadoop/hdds/scm/pipeline/TestPipelinePlacementPolicy.java#L91-L92

https://issues.apache.org/jira/browse/HDDS-5207

## How was this patch tested?

100x repeated `TestRatisPipelineProvider`:
https://github.com/adoroszlai/hadoop-ozone/runs/2549825702

Regular CI:
https://github.com/adoroszlai/hadoop-ozone/actions/runs/829491474


GlenGeng:
+1. 
Thanks @adoroszlai for the effort to stabilize the CI.

bshashikant (approved):

bshashikant:
Thanks @adoroszlai for the fix and @GlenGeng for the review.

adoroszlai:
Thanks @GlenGeng and @bshashikant for the review.

PR #2234, opened 2021-05-10
HDDS-5164. Improve client and server logging.

hanishakoneru:

On the OM server side, NotLeaderException and LeaderNotReadyExceptions can be suppressed. Otherwise, OM1 log is flooded with NotLeaderExceptions as clients always try OM1 before moving to the next OM. Instead we can change these exception logs to DEBUG.

Some BlockOutputStream and BlockOutputStreamEntryPool logs should be DEBUG level instead of INFO. For example, block allocation with an exclude list need not be logged at INFO level.

https://issues.apache.org/jira/browse/HDDS-5164

## How was this patch tested?

No new tests. Changing only LOG levels.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -106,6 +106,8 @@ import org.apache.hadoop.ozone.audit.OMAction;
 import org.apache.hadoop.ozone.common.Storage.StorageState;
 import org.apache.hadoop.ozone.om.exceptions.OMException;
 import org.apache.hadoop.ozone.om.exceptions.OMException.ResultCodes;
+import org.apache.hadoop.ozone.om.exceptions.OMLeaderNotReadyException;
+import org.apache.hadoop.ozone.om.exceptions.OMNotLeaderException;
 import org.apache.hadoop.ozone.om.ha.OMHANodeDetails;
 import org.apache.hadoop.ozone.om.ha.OMNodeDetails;
 import org.apache.hadoop.ozone.om.helpers.DBUpdates;
@@ -883,6 +885,10 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
         false)) {
       rpcServer.refreshServiceAcl(conf, OMPolicyProvider.getInstance());
     }
+
+    rpcServer.addSuppressedLoggingExceptions(OMNotLeaderException.class,
+        OMLeaderNotReadyException.class);
+
     return rpcServer;
   }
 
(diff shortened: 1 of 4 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerProtocolServerSideTranslatorPB.java
```diff
@@ -193,6 +193,15 @@ private OMResponse submitReadRequestToOM(OMRequest request)
     }
   }
 
+  private ServiceException createLeaderErrorException(
```
bharatviswa504, line 196: Minor: Unnecessary code change. I see it is just moved up.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -883,6 +885,10 @@ private static StorageContainerLocationProtocol getScmContainerClient(
         false)) {
       rpcServer.refreshServiceAcl(conf, OMPolicyProvider.getInstance());
     }
+
+    rpcServer.addSuppressedLoggingExceptions(OMNotLeaderException.class,
```
bharatviswa504, line 889: Good find, so now on server side Rpc does not log these errors.

bharatviswa504 (commented):
Over all LGTM.
One minor comment

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerProtocolServerSideTranslatorPB.java
hanishakoneru: Yes functionally no change. Moved it up for aesthetics and improving code readability.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
hanishakoneru: Yes, otherwise we see a lot of noise - "om1 is not the leader" exceptions, on the server side logs if the 1st OM is not the leader.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/protocolPB/OzoneManagerProtocolServerSideTranslatorPB.java
bharatviswa504: I am fine with it, but one thing with this is we will lose which commit originally added this. (As sometimes I check why this code is added/changed to understand the reason for this code change. This is just my personal opinion.)

bharatviswa504:
Thank You @hanishakoneru for the fix.

hanishakoneru:
Thank you @bharatviswa504 

>I am fine with it, but one thing with this is we will lose which commit originally added this. (As sometimes I check why this code is added/changed to understand the reason for this code change. This is just my personal opinion.)

Makes sense. I did not consider this aspect. I will keep this in mind in future. Thanks.

PR #2235, opened 2021-05-11
HDDS-5208. bump rocksdb version to 6.20.3

JacksonYao287:

now , the rocksdb version used in ozone is 6.8.1, which is build on Apr 16 , 2020.
this PR bumps rockdsb to 6.20.3, which is build on May 5, 2021, and includes many new bug-fix

https://issues.apache.org/jira/browse/HDDS-5208

## How was this patch tested?

no need to add additional tests

### DIFF
--- a/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerRocksDBLogging.java
+++ b/hadoop-ozone/integration-test/src/test/java/org/apache/hadoop/ozone/om/TestOzoneManagerRocksDBLogging.java
@@ -91,7 +91,7 @@ public class TestOzoneManagerRocksDBLogging {
       throws TimeoutException, InterruptedException {
     GenericTestUtils.waitFor(() -> logCapturer.getOutput()
             .contains("db_impl.cc"),
-        1000, 10000);
+        1000, 30000);
   }
 
 }
--- a/pom.xml
+++ b/pom.xml
@@ -1602,7 +1602,7 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
       <dependency>
         <groupId>org.rocksdb</groupId>
         <artifactId>rocksdbjni</artifactId>
-        <version>6.8.1</version>
+        <version>6.20.3</version>
       </dependency>
       <dependency>
         <groupId>org.xerial</groupId>


JacksonYao287:
@elek please take a look

JacksonYao287:
@elek hi, in my local test , integration(ozone) test  is passed, but it fail here. is it a flaky case?

adoroszlai:
> in my local test , integration(ozone) test is passed, but it fail here. is it a flaky case?

`TestOzoneManagerRocksDBLogging` has not been flaky recently, plus it looks related.  I guess the problem could be that output may be changed in the new version.

(BTW, please trigger CI with empty commit instead of closing+reopening the PR.  This makes it easier to see previous CI results.)

JacksonYao287:
@adoroszlai i make the wait time longer ,and CI is passed.  it seems we should wait longer for the log if we use the rocksdb of this version.

elek (approved):
thanks the patch @JacksonYao287 

Retriggered the build to be sure that the (possible) intermittency is disappeared. If it passes, I am +1 to merge it.

adoroszlai:
> i make the wait time longer ,and CI is passed. it seems we should wait longer for the log if we use the rocksdb of this version.

Thanks for trying that.

The test expects the output `db_impl.cc`, which first appears when `DUMPING STATS`:

```
... [Thread-302] INFO  rocksdb.RocksDB (DBStoreBuilder.java:log(304)) - [/db_impl/db_impl.cc:932] ------- DUMPING STATS -------
```

Now with the new version there is a 15-sec delay.  @avijayanhwx Would it make sense to check for `rocksdb.RocksDB` instead?  I think it would appear in all lines coming from RocksDB:

https://github.com/apache/ozone/blob/6b24c188df038f1ab404acace20a4060fe2114fd/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/DBStoreBuilder.java#L64-L65

avijayanhwx:
> > i make the wait time longer ,and CI is passed. it seems we should wait longer for the log if we use the rocksdb of this version.
> 
> Thanks for trying that.
> 
> The test expects the output `db_impl.cc`, which first appears when `DUMPING STATS`:
> 
> ```
> ... [Thread-302] INFO  rocksdb.RocksDB (DBStoreBuilder.java:log(304)) - [/db_impl/db_impl.cc:932] ------- DUMPING STATS -------
> ```
> 
> Now with the new version there is a 15-sec delay. @avijayanhwx Would it make sense to check for `rocksdb.RocksDB` instead? I think it would appear in all lines coming from RocksDB:
> 
> https://github.com/apache/ozone/blob/6b24c188df038f1ab404acace20a4060fe2114fd/hadoop-hdds/framework/src/main/java/org/apache/hadoop/hdds/utils/db/DBStoreBuilder.java#L64-L65

Yes, I am ok with that.

JacksonYao287:
> Retriggered the build to be sure that the (possible) intermittency is disappeared. If it passes, I am +1 to merge it.
thanks @elek for the retrigger。

JacksonYao287:
> Now with the new version there is a 15-sec delay. @avijayanhwx Would it make sense to check for `rocksdb.RocksDB` instead? I think it would appear in all lines coming from RocksDB:

@adoroszlai , in this test case , it seems that server rocksdb benchmarks are carried out and then the corresponding benchmark stats are dumped. so I think maybe in this version of rocksdb , more benchmarks are involved ,  and thus it takes longer time to complete

adoroszlai:
Thanks @JacksonYao287 for the patch, @elek for the review.  We can address the delay separately.

PR #2236, opened 2021-05-11
HDDS-5205. Make admin check work for SCM HA cluster

bharatviswa504:

Make admin commands like safe mode etc., work on SCM HA cluster.

Note: When working this identified for two of the calls getContainerWithPipeline and getContainerWithPipelineBatch checkAdmin is called with null, not sure why it is called if we do not do any checks. For now, left as it is.

https://issues.apache.org/jira/browse/HDDS-5205

## How was this patch tested?

Added a smoke test.


hadoop-ozone/dist/src/main/compose/ozonesecure-ha/test.sh
```diff
@@ -34,6 +34,9 @@ execute_robot_test ${SCM} freon
 execute_robot_test ${SCM} basic/links.robot
 
 execute_robot_test ${SCM} s3
+
+export SCM=scm2.org
+execute_robot_test ${SCM} security-ha
```
adoroszlai, line 39: Instead of introducing a new test suite (`security-ha`) with only a single test case that is not even HA-specific, I think we should run the complete `admincli` test suite (probably for both scm1 and scm2).

```suggestion
execute_robot_test ${SCM} admincli

export SCM=scm2.org
execute_robot_test ${SCM} admincli
```

Only one test case in each `admincli` test file needs to be changed for this to work: `... on unknown host` fails because `--scm` parameter is ignored for SCM HA cluster.  I think it is a bug that should be fixed in the long term.  But for now these test cases can be disabled (commented out), as they are not too important.  I think it is more important to have full `admincli` test coverage for HA than to run this marginal test case for non-HA.

https://github.com/apache/ozone/blob/49538aaa11f41bf80f565f9ce4f2c49ebac5fd51/hadoop-ozone/dist/src/main/smoketest/admincli/container.robot#L75-L77

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
```diff
@@ -404,13 +403,12 @@ private boolean hasRequiredReplicas(ContainerInfo contInfo) {
...
   public void deleteContainer(long containerID) throws IOException {
-    String remoteUser = getRpcRemoteUsername();
     boolean auditSuccess = true;
     Map<String, String> auditMap = Maps.newHashMap();
     auditMap.put("containerID", String.valueOf(containerID));
-    auditMap.put("remoteUser", remoteUser);
+    auditMap.put("remoteUser", getRpcRemoteUsername());
     try {
-      getScm().checkAdminAccess(remoteUser);
+      getScm().checkAdminAccess(Server.getRemoteUser());
```
adoroszlai, line 411: Also note that `TestStorageContainerManager` integration test is failing, because `Server.getRemoteUser()` returns `null` (no mock set up for this static method, as opposed to `SCMClientProtocolServer.getRpcRemoteUsername()`), so access is granted.

https://github.com/apache/ozone/pull/2236/checks?check_run_id=2553193751#step:4:4223

adoroszlai (changes requested):

hadoop-ozone/dist/src/main/compose/ozonesecure-ha/test.sh
bharatviswa504: Done. For now, commented them.

hadoop-hdds/server-scm/src/main/java/org/apache/hadoop/hdds/scm/server/SCMClientProtocolServer.java
bharatviswa504: Fixed it

hadoop-ozone/dist/src/main/compose/ozonesecure-ha/test.sh
bharatviswa504: For ha --scm should it pass multiple scms? What is the expected behavior, I can open a Jira for this.

bharatviswa504:
Thank You @adoroszlai for the review. I have addressed review comments.

hadoop-ozone/dist/src/main/compose/ozonesecure-ha/test.sh
adoroszlai: I guess SCM service ID and SCM host should be exclusive, only one of these should be allowed.

https://github.com/apache/ozone/blob/03620da6158534ac618354737a8b35d09a25fa33/hadoop-hdds/tools/src/main/java/org/apache/hadoop/hdds/scm/cli/ScmOption.java#L46-L53

adoroszlai (approved):
Thanks @bharatviswa504 for updating the patch, LGTM.

bharatviswa504:

